PR #2, opened 2021-12-23
Add a proxy and a web UI

imobachgs:
* Implement a `yastd-proxy` to access the D-Bus service from the browser.
* Bring the web UI from https://github.com/imobachgs/installer-80 and adapt it to the new proxy.

installer-80


.rubocop.yml
```diff
@@ -15,4 +15,4 @@ Metrics/LineLength:
 Metrics/BlockLength:
   # rspec is known as DSL with big blocks
   Exclude:
-    - lib/yast2/dbus/installer.rb
+    - yastd/lib/yast2/dbus/installer.rb
```
jreidinger, line 18: just note that comment is wrong. it is not about rspec, but about dbus DSL

yastd-proxy/config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: test
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
jreidinger, line 9: why the fetch needs block?

yastd-proxy/config/initializers/filter_parameter_logging.rb
```diff
@@ -0,0 +1,6 @@
+# Be sure to restart your server when you modify this file.
+
+# Configure sensitive parameters which will be filtered from the log file.
+Rails.application.config.filter_parameters += [
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
```
jreidinger, line 5: nice. Would be nice to have something similar for yast logger

jreidinger (approved):

PR #3, opened 2021-12-24
Implement changes notification in  the UI

imobachgs:
* Replace the Rails-based implementation with an EventMachine + Sinatra one.
* Notify properties changes over D-Bus.
* Refresh the UI when something changes.
* Reorder the ports: 3000 for UI, 3001 for HTTP/JSON, 3002 for WS.


dgdavid (approved):
LGTM

PR #4, opened 2021-12-24
Update installer status

imobachgs:
* Probing and installation do not block the D-Bus server anymore. For the time being we are using an `EventMachine::PeriodicTimer`, but probably there is a better way.
* Add a method `Installer#on_status_change` in order to inject a callback which is called when the status change. There was some kind of a circular dependency between `DBus::Installer` and `Installer` (I only was able to call the `#StatusChanged` signal through the `DBus::Installer).
* The `DBus::InstallerClient` does not exist anymore because of the previous change.
* Status changes are now properly propagated to the web UI.

*TODO: the status update is not handled in the web UI yet.**

## Some videos

Partitioning the disk (you need to uncomment the [inst_prepdisk client call](https://github.com/yast/the-installer/blob/update-status/yastd/lib/yast2/installer.rb#L113):

partitioning-a-disk

Delivering properties and status changes to the browser:

fake-installation


README.md
```diff
@@ -11,6 +11,13 @@ service. At first sight, we have identified these components:
 * A user interface. For this experiment, we have decided to use a
   [Cockpit](https://cockpit-project.org/) module.
 
+## Requirements
+
+To build and run this software you need a few tools. To install them on openSUSE
+Tumbleweed just type:
+
+    $ sudo zypper in gcc gcc-c++ make openssl-devel ruby-devel npm
```
jreidinger, line 19: :+1: to document it

jreidinger:
lets merge it now as it can do partitioning and software installation

PR #7, opened 2022-01-03
Add Progress to bootloader install

jreidinger:


yastd/lib/yast2/installation_progress.rb
```diff
@@ -22,23 +22,70 @@
...
         # TODO: localization
         "Installing packages (#{@remaining_pkgs} remains)",
-        1, 0, @total_pkgs, @total_pkgs - @remaining_pkgs
+        1, substeps: @total_pkgs, current_substep: @total_pkgs - @remaining_pkgs
+      )
+    end
+
+  private
+
+    def report(title, step, substeps: 0, current_substep: 0)
```
dgdavid, line 85: Why not making it public and use it directly in the installer.rb?

I mean,

```ruby

...

def install
  ...
  Yast::Installation.destdir = "/mnt"
  # Imaging we have a collection of steps or so where partitioning could be...
  # 
  # class Partitioning < InstallationStep
  #   def start_label
  #     _("Partitioning target disk")
  #   end
  # 
  #   def perform
  #     Yast::WFM.CallFunction("inst_prepdisk", [])
  #   end
  #  
  #   ...
  steps = [partitioning, package_installation, bootloader_installation]

  # So...

  progress = InstallationProgress.new(@dbus_obj, steps.size, logger: logger)
  
  steps.each_with_index |step, index| do
    progress.report(step.start_label, index)
    step.perform
    progress.report(step.end_label, index + 1)
  end

  ...
end

...

```

jreidinger: well, I am still not sure whose responsibility it is. Also progress needs to know about number of steps so it can report it. So question is who owns responsibility for proper messaging. Still not sure what would be the best solution

dgdavid: > Also progress needs to know about number of steps so it can report it

Right. That's why I put the "steps.size" in the `InstallationProgress` constructor. To tell it the number of steps when creating it, instead of having an internal constant that must be in sync with the installer.rb.

Anyway, let's continue playing around. Thank you!

PR #11, opened 2022-01-26
D-Bus API for the minimal use case

mvidner:
- https://trello.com/c/4ayVcpxC/2796-5-d-installer-define-a-d-bus-api-for-the-minimal-use-case


yastd/dbus_api.md
```diff
@@ -0,0 +1,29 @@
+
+## get the list of available languages
+
+Iface: o.o.YaST.Installer1.Language
+
+stupid api:
+  AvailableLanguages -> array(string)
```
jreidinger, line 7: what about using ObjectManager interface for it?

yastd/dbus_api.md
```diff
@@ -0,0 +1,29 @@
+
+## get the list of available languages
```
jreidinger, line 2: just to clarify it is available languages for installation or for clients?

yastd/dbus_api.md
```diff
@@ -0,0 +1,29 @@
...
+
+Iface: o.o.YaST.Installer1.Language
+
+stupid api:
+  AvailableLanguages -> array(string)
+
+minimal working api:
+  needed: human readable name (how localized?), identifier
```
jreidinger, line 10: I also expect some setter/getter for selected language ( maybe languages to allow install more locales variants? )

yastd/dbus_api.md
```diff
@@ -0,0 +1,29 @@
...
+
+identifiers: maybe LanguageTag https://www.rubydoc.info/github/yast/yast-packager/master/LanguageTag
+- move it to yast-yast2
+- link to the standard from yard 
+- see https://tools.ietf.org/html/rfc4647 Matching of Language Tags
+- see https://lists.opensuse.org/archives/list/yast-devel@lists.opensuse.org/message/D52PSZ7TRID2RVM6CE6K2C2RUNNGOS6Z/
```
jreidinger, line 16: well dbus api design document suggest using enums. What about having just simple plain enums here?

yastd/dbus_api.md
```diff
@@ -0,0 +1,29 @@
...
+
+return an array of objects (object paths) representing the languages?
+
+(later: keyboard layouts; territories? timezones? console fonts?)
```
jreidinger, line 20: if you plan to extend it for timezone or keyboard layouts, then I think naming interface language is not propriete.

yastd/dbus_api.md
mvidner: Huh, good point. This is for the installed system.
The UI at installation time should be in the same language, but the D-Bus API is a lower layer and does not care what its caller is showing to the user.

yastd/dbus_api.md
mvidner: Enums... This example will illustrate why I dislike enums on D-Bus:
```console
$ for p in Capabilities Ip4Connectivity Ip6Connectivity Metered State; do 
  printf "$p: "; \
  busctl --system get-property org.freedesktop.NetworkManager \
    /org/freedesktop/NetworkManager/Devices/3 \
    org.freedesktop.NetworkManager.Device $p; \
  done
Capabilities: u 1
Ip4Connectivity: u 4
Ip6Connectivity: u 3
Metered: u 4
State: u 100
```

I mean, while D-Bus has _many_ provisions to make interfaces _discoverable_, using enums is a sure way to kill that discoverability. If we use enums I will argue for adding an interface to expose the number-name mapping.

Anyway, I digress. Enjms are surely unfit for languages/locales, which are objects (simple objects, or complex strings), not members of an enumeration.

yastd/dbus_api.md
mvidner: Right, they should get their own interfaces. I was just looking at what yast2-country is doing.

yastd/dbus_api.md
mvidner: Yes
- SelectedLanguages
- DefaultLanguage
- 
(maybe Locale is better than Language)

yastd/dbus_api.md
mvidner: ObjectManager is an implementation detail about the tree of objects and their interfaces and properties.
Your suggestion implies the existence of the objects
```
/org/opensuse/YaST/Installer1/Language/Available/af_ZA
/org/opensuse/YaST/Installer1/Language/Available/ar_EG
/org/opensuse/YaST/Installer1/Language/Available/bg_BG
...
```

It's one of the options, I still don't know if it's the best one

yastd/dbus_api.md
jreidinger: @imobachgs can you please clarify this? I think our initial goal is to have API for installer. So what exactly should do that list of available languages?

yastd/dbus_api.md
jreidinger: yeah, for sure iterating over all that objects can be performance kill, so maybe returnning array of structs with open ended map as the last element or array of map can be better solution that is extendable and also good performance wise.....But as mentioned we should first clarify what is an exact goal of this API.

yastd/dbus_api.md
imobachgs: The idea is to set the language for the installed system. We most probably should sync the UI language too, but that's out of D-Bus control by now.

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+Iface: o.o.YaST.Installer1.Language
+
+methods:
+
+  ToInstall(array(string)) -> void
+    Set list of languages to install
+    Example:
+      ToInstall(["cs_CZ", "de_DE"]) -> () # only lang codes from AvailableLanguages is supported
+
+Properties (all read only):
```
imobachgs, line 13: :+1:

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+
+  AvailableLanguages -> array(array(string, string, dict(string, variant)))
+    List of all available languages to install on target system.
+    Example:
+      AvailableLanguages -> [["cs_CZ", "Czech", {}]] # it is lang code, human readable lang name and dict for future extensions to provide more data
+
+  MarkedForInstall -> array(string)
```
imobachgs, line 20: Marked or Selected?

imobachgs (commented):
Having `Available*` methods, like `AvailableLanguages`, makes sense to me. However, I am wondering if we should use a D-Bus object for each language (or product, or whatever...) instead.

Thinking about the languages:

* `/o/o/YaST/Languages/English` (or `/en_US`).
* `/o/o/YaST/Languages/Czech`

And those objects could define properties like the display name, or any other stuff we would want to add in the future. The same for products (`/o/o/YaST/Products/SLES`) and so on. Even they could implement a "Selectable" interface (just thinking aloud).

Please, bear in mind that I have little knowledge about D-Bus, so I am not sure whether it is a reasonable practice or just a stupid idea 😄

yastd/dbus_api.md
jreidinger: it really depends on installer behavior. If all marked for install will be used or not.

jreidinger:
@imobachgs Well, issue with that object approach can be that it needs multiple calls just to get initial set of languages ( or products ). I expect bigger problem for languages as e.g. opensuse support more of them and you need to query all of them to get data.
Another reason which we discus with @mvidner is that having everything as object makes API and implementation more complex, so for initial API we would like to try both approaches ( and pure data for simplier one, and object for more complex ones ) to get some experience and also see that performance.
What I am also not sure ( @mvidner feel free to correct me ) is how behaves PropertiesChanged with object. E.g. if product version or name is different after some change ( e.g. registration ) if PropertiesChanged is fired when content of object is changed and not object path itself.

yastd/dbus_api.md
imobachgs: OK, I see. It is just that I saw `SelectedBaseProduct` and I thought about using the same pattern.

imobachgs:
I expected to avoid multiple calls by implementing the [ObjectManager interface](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager), but I am really not sure if it serves our purposes.

imobachgs:
> I expected to avoid multiple calls by implementing the [ObjectManager interface](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager), but I am really not sure if it serves our purposes.

Oh, it was already mentioned in https://github.com/yast/the-installer/pull/11#discussion_r792537052 😄 Sorry for overlooking that!

jreidinger:
We tried to do some measuring with d-feet how will object manager works when there are more objects, but sadly all objects that implements ObjectManager has for me just 1-5 elements and result is that speed is OK like 50ms for single call.

yastd/dbus_api.md
```diff
@@ -3,27 +3,167 @@
 
 Iface: o.o.YaST.Installer1.Language
 
-stupid api:
-  AvailableLanguages -> array(string)
+methods:
 
-minimal working api:
-  needed: human readable name (how localized?), identifier
+  ToInstall(array(string)) -> void
```
mvidner, line 8: An important thing about **string**:
Even if it is the implementation type at the D-Bus layer, we should design the API in terms of specialized constant strings. 
Here, a locale identifier as provided by glibc-locale.rpm

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+    Example:
+      InstallProduct("SLES") -> () # only name from available BaseProducts is supported
+
+Properties (all read only):
+
+  AvailableBaseProducts -> array(array(string, string, dict(string, variant)))
+    List of all available base product to install on target system.
+    Note: List is sorted according to defined display order
+    Example:
+      AvailableBaseProducts -> [["SLES", "SUSE Linux Enterprise Server", {}]] # it is product name, human readable name and dict for future extensions to provide more data
```
mvidner, line 54: Again, a struct like for languages.
How about this naming:
- ProductKey or ProductIdentifier
- product_label (where **label** is a "string visible by the user"; I try to avoid **name** because it can mean "label" or "key")
- details

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+
+  InstallProduct(string) -> void
+    Set product for install.
+    TODO: do we need version or arch or any other parameter? should we add generic dict(sv) to it?
+    Example:
+      InstallProduct("SLES") -> () # only name from available BaseProducts is supported
+
+Properties (all read only):
+
+  AvailableBaseProducts -> array(array(string, string, dict(string, variant)))
```
mvidner, line 50:
```suggestion
  AvailableBaseProducts -> array(struct(string, string, dict(string, variant)))
```

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+Iface: o.o.YaST.Installer1.Storage
+
+methods:
+
+  MarkForUse(array(object)) -> void
+    set objects for use of installation. it means erase content of that devices
+    example:
+      MarkForUse([disk1,disk2partition2]) -> ()
+
+  MarkForShrink(array(object)) -> void
```
mvidner, line 77:
```suggestion
  MarkForShrinking(array(object)) -> void
```

yastd/dbus_api.md
jreidinger: we, we should be consistent on API side. so if we use here identifier from glibc-locale, we should use same identifier in list of available languages

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+    set objects to allow shrink of them. it means keep content and reduce its free space.
+    example:
+      MarkForShrink([disk1,disk2partition2]) -> ()
+
+Properties (all read only):
+
+  Disks -> array(object)
```
mvidner, line 84:
```suggestion
  Disks -> array(o.o.YaST.Installer1.Storage.Drive)  # an object_path whose object implements this interface
```

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+    List of all disks.
+    Example:
+      Disks -> [disk1, disk2]
+
+  Partitions -> array(object)
```
mvidner, line 89:
```suggestion
  Partitions -> array(o.o.YaST.Installer1.Storage.Partition)
```

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+    List of all partitions.
+    Example:
+      Disks -> [disk1partition1, disk1partition2, disk2partition1]
+
+  DevicesToUse -> array(object)
+    Devices that will be fully used by installation
+    Example:
+      DevicesToUse -> [disk1,disk2partition2]
+
+  DevicesToShrink -> array(object)
```
mvidner, line 99:
```suggestion
  DevicesToShrink -> array(o.o.YaST.Installer1.Storage.BlockDevice)
```
?

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+  Vendor -> string
+    Vendor of device or empty string if not known like "Fujitsu"
+
+  Model -> string
+    Device model or empty string if not known
+
+  Removable -> boolean
+    if device is removable like usb sticks
+
+  Partitions -> array(object)
```
mvidner, line 153:
```suggestion
  Partitions -> array(o.o.YaST.Installer1.Storage.Partition)
```

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+    partitions on given drive
+
+Iface: o.o.YaST.Installer1.Storage.Partition
+
+Inspired by Udisks2.Partition
+
+Properties (all read only):
+
+  Drive -> object
```
mvidner, line 162:
```suggestion
  Drive -> o.o.YaST.Installer1.Storage.Drive
```

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+methods:
+
+  ToInstall(array(string)) -> void
+    Set list of languages to install
+    Example:
+      ToInstall(["cs_CZ", "de_DE"]) -> () # only lang codes from AvailableLanguages is supported
+
+Properties (all read only):
+
+  AvailableLanguages -> array(array(string, string, dict(string, variant)))
```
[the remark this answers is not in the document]
jreidinger: yeap, struct makes sense here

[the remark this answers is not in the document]
mvidner: Another thing, currently ruby-dbus implements *struct* as an *Array* but I can imagine a mapping from a *Struct* (or any class, with some declaratory helpers).

Which reminds me that *OpenStruct* should pass for a D-Bus dictionary (feature wish)

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+- see https://tools.ietf.org/html/rfc4647 Matching of Language Tags
+- see https://lists.opensuse.org/archives/list/yast-devel@lists.opensuse.org/message/D52PSZ7TRID2RVM6CE6K2C2RUNNGOS6Z/
+
+## get the list of base products
+
+Iface: o.o.YaST.Installer1.Software
+
+methods:
+
+  InstallProduct(string) -> void
```
mvidner, line 42: Oh, methods taking a long time!
Probably not this one, but D-Bus does not like them. There is a global timeout of like 25 seconds per call. You need to learn from PackageKit and use signals for things taking a longer time.

mvidner:
> how behaves PropertiesChanged with object. E.g. if product version or name is different after some change ( e.g. registration ) if PropertiesChanged is fired when content of object is changed and not object path itself.

Object mutation is not covered by the signal. Object identity is, as it's object path changes.

yastd/dbus_api.md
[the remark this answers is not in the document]
jreidinger: openstruct usage is basically discouraged. So I am not sure if it should be used.

About Struct usage it will be nice, but I think it is hard to get from dbus signature names of variables...and mapping from array to struct is quite easy `response = MyStruct.new(*dbus_call)`

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+
+Properties (all read only):
+
+  AvailableBaseProducts -> array(array(string, string, dict(string, variant)))
+    List of all available base product to install on target system.
+    Note: List is sorted according to defined display order
+    Example:
+      AvailableBaseProducts -> [["SLES", "SUSE Linux Enterprise Server", {}]] # it is product name, human readable name and dict for future extensions to provide more data
+
+  SelectedBaseProduct -> string
```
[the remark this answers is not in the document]
jreidinger: I am not sure about mixing dbus types and semantic of that string. when writting api I consider using var names syntax like `string ProductKey` or in params like `SelectBaseProduct(string ProductKey)`

yastd/dbus_api.md
```diff
@@ -0,0 +1,169 @@
...
+  Device -> string
+    Block device name in /dev like "/dev/sda"
+
+  Size -> uint64
+    Size of devices in bytes
+
+  ReadOnly -> boolean
+    if device is read only
+
+Iface: o.o.YaST.Installer1.Storage.BlockDevice
```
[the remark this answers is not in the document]
jreidinger: Copy paste error. Vim is mighty :)

jreidinger:
@mvidner Hi, PR should be ready now for final review.

yastd/dbus_api.md
```diff
@@ -0,0 +1,202 @@
...
+
+-  MarkForShrinking(array(o.o.YaST.Installer1.StorageBlockDevice Device)) -> void
+    set objects to allow shrink of them. it means keep content and reduce its free space.
+    example:
+
+      MarkForShrink([disk1,disk2partition2]) -> ()
+
+#### Properties (all read only):
+
+-  Disks -> array(o.o.YaST.Installer1.Storage.Drive)  # an object\_path whose object implements this interface
```
mvidner, line 134:
```suggestion
-  Drives -> array(o.o.YaST.Installer1.Storage.Drive)  # an object\_path whose object implements this interface
```

For consistency, the method name and the interface of the return value should match, like everywhere else in this API

yastd/dbus_api.md
```diff
@@ -0,0 +1,202 @@
...
+    List of all disks.
+    Example:
+
+      Disks -> [disk1, disk2]
```
mvidner, line 138:
```suggestion
      Drives -> [disk1, disk2]
```

jreidinger (approved):

PR #15, opened 2022-01-31
Switch from Chakra UI to PatternFly + EOS Design System

dgdavid:
Until now, we have been using [chakra](https://chakra-ui.com/) as a component library. Mainly because it allowed us to start prototyping [the $INSTALLER:80](https://github.com/imobachgs/installer-80) quickly and… because it was one of the ~~_new toys to play with_~~ new and promising UI libraries to explore at that time. And, among others like [Base Web](https://baseweb.design/), [Mantine](https://mantine.dev), or [whichever of those awesome libraries](https://dev.to/chetan_atrawalkar/25-top-react-ui-component-library-1m68), it still looking as a great choice, at least looking at it with the glasses of someone touching Web development from time to time :wink:

However, since we will try to go a step further with `[the|D]-installer` project, betting on [PatternFly](https://www.patternfly.org) and [EOS Design System](https://www.eosdesignsystem.com/) feels like the right path since they are great too. 

Following EOS DS, the project could provide a beautiful, consistent and cohesive UI/UX while using PatternFly can not only easy a future integration of some components into the [Cockpit](https://cockpit-project.org/) dashboard, but it also offers a couple of interesting components that can helps in an adavanced stage of the project like the [wizard](https://www.patternfly.org/v4/components/wizard) - think in guided proposals -, the [about modal](https://www.patternfly.org/v4/components/about-modal) - maybe for presenting detailed information about a role or a product-, the [dual list selector](https://www.patternfly.org/v4/components/dual-list-selector), the [progress stepper](https://www.patternfly.org/v4/components/progress-stepper), etc. 

Nothing that couldn't be created when required or even that might be already solved in other UI libraries, of course. But, I'd go for them and even to encourage the collaboration with their teams.

---

Please, note that this PR is somehow a 1:1 migration from the previous code using chakra to a new one using PatternFly, without concerning too much about component abstraction, optimization, etc. What is more, the overall approach  needs some polishing (like start applying https://suse.eosdesignsystem.com/writing/ux-writing) for integrating better all pieces, but it can/should be addressed later.

---

| After | Before |
|-|-|
| Prototype using Chakra UI | Prototype using PatternFly + EOS DS |


web/src/App.js
```diff
@@ -60,75 +52,79 @@ function App() {
...
+        <StackItem>
+          <TextContent>
+            <Text component={TextVariants.h1}>Welcome to D-Installer</Text>
+          </TextContent>
+        </StackItem>
+
+        <StackItem>
+          <Category title="Language" icon={LanguagesSelectionIcon}>
+            <LanguageSelector
+              value={l10n.language || "Select language"}
```
dgdavid, line 80: :warning: NOTE that we should start applying https://suse.eosdesignsystem.com/writing/ux-writing here. Looks to me that it'd help to have a clear and concise UI. I.e., better UX.

dgdavid: I mean, not only there. In general.

imobachgs (approved):

PR #22, opened 2022-02-07
[yastd] Only consider available base products

imobachgs:
Only consider available base products for installation.


dgdavid (approved):

yastd/lib/yast2/software.rb
```diff
@@ -53,7 +53,7 @@ def probe
       Yast::Pkg.TargetLoad
       Yast::Pkg.SourceRestore
       Yast::Pkg.SourceLoad
-      @products = Y2Packager::Product.all
+      @products = Y2Packager::Product.available_base_products
```
dgdavid, line 56: 👍

PR #23, opened 2022-02-10
[web] Add tests for the React application

imobachgs:
Add some testing for the web UI.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: ./web
+
+    strategy:
+      matrix:
+        node-version: [16.x]
```
jreidinger, line 16: isn't TW already using note17?

web/package.json
```diff
@@ -48,5 +48,15 @@
...
     "prettier": "^2.5.1"
+  },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{js,jsx}",
+      "!src/lib/cockpit.js",
+      "!src/reportWebVitals.js",
+      "!src/serviceWorker.js",
+      "!src/setupProxy.js",
+      "!<rootDir>/node_modules/"
```
jreidinger, line 59: this line is not much clear for me ( even after I google jest rootDir ). Can you add comment with explanation?

web/src/App.test.js
```diff
@@ -1,9 +1,52 @@
...
-test("renders the App", () => {
-  render(<App />);
-  const title = screen.getByText(/Welcome to D-Installer/i);
-  expect(title).toBeInTheDocument();
+jest.mock("./lib/InstallerClient");
+jest.mock("./Installer", () => {
+  return {
+    __esModule: true,
+    default: () => {
+      return <div>Installer Component</div>;
```
jreidinger, line 12: so jest code can use JSX? cool

web/src/App.test.js
```diff
@@ -1,9 +1,52 @@
...
+      };
+    });
+  });
+
+  it("shows an error", async () => {
+    authRender(<App />);
+    userEvent.type(screen.getByLabelText(/Username/i), "john");
+    userEvent.type(screen.getByLabelText(/Password/i), "something");
+    userEvent.click(screen.getByRole("button", { name: /Login/ }));
+    await screen.findByText(/Authentication failed/i);
```
jreidinger, line 50: this awaits is not so clear for me. Does it mean that in failed case it will wait timeout and then make test failed with error message that given text is not found?

web/src/LanguageSelector.test.js
```diff
@@ -0,0 +1,54 @@
...
+it("displays the proposal", async () => {
+  installerRender(<LanguageSelector />);
+  await screen.findByText("English");
+});
+
+describe("when the user changes the language", () => {
+  let setOptionFn;
+
+  beforeEach(() => {
+    // if defined outside, the mock is cleared automatically
```
jreidinger, line 32: is this comment for code here or for definition of variable at 29?

jreidinger: ah, variable is used at test, so needed to be outside.

web/src/Storage.test.js
```diff
@@ -0,0 +1,95 @@
...
+  beforeEach(() => {
+    const finalProposal = [
+      { mount: "/", type: "Btrfs", device: "/dev/sdb1", size: 100000000 },
+      { mount: "/home", type: "Ext4", device: "/dev/sdb2", size: 10000000000 }
+    ];
+
+    const getStorageFn = jest
+      .fn()
+      .mockResolvedValue(finalProposal) // default return value
+      .mockResolvedValueOnce(initialProposal); // first call
```
jreidinger, line 71: just curious, is method call order important here or not?

.github/workflows/ci.yml
imobachgs: TW ships Node 16 and 17. However, Node 16 is the active LTS version. See [Node releases](https://nodejs.org/en/about/releases/) to find out more details.

web/src/App.test.js
imobachgs: Yes, that's it. The react-testing-library documentation contains [an interesting comparison](https://testing-library.com/docs/dom-testing-library/cheatsheet/#queries) of `findBy`, `getBy` and `queryBy`. You can find more information about [findBy](https://testing-library.com/docs/dom-testing-library/api-async#findby-queries) queries.

web/src/Storage.test.js
imobachgs: The order matters, but not in this case. The `*Once` version is called first, and then it falls back to the rest. If we had several `*Once` versions, it would call them in the correct order (and the general one at the end).

jreidinger (approved):

web/src/Storage.js
```diff
@@ -64,13 +64,13 @@ export default function Storage() {
   }, []);
 
   return (
-    <div>
+    <>
```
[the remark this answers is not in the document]
dgdavid: There is an important reason: [Fragments](https://reactjs.org/docs/fragments.html) were introduced to avoid adding extra nodes to the DOM.

PR #24, opened 2022-02-15
[web] Limit the memory usage when building

imobachgs:
Just trying...

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -22,8 +22,8 @@
     "web-vitals": "^0.2.2"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


mvidner:
This may help in some cases but I've tried 2048, 1024, 1500, 1400 and neither works on a 4 GiB RAM guest.

imobachgs:
This PR is now obsolete. After migrating to Vite (see #32) I was able to run the development environment (and even build the production bundle) in a 4G machine. :-)

PR #26, opened 2022-02-16
initial new dbus service code

jreidinger:
trello: https://trello.com/c/4ayVcpxC/2796-5-d-installer-define-a-d-bus-api-for-the-minimal-use-case

Start with new service with more serious API which should be according to api design doc - https://github.com/yast/d-installer/blob/master/yastd/dbus_api.md

Still work in progress, but lets merge it to allow parallelize work.


yastd/d-installer.gemspec
```diff
@@ -46,5 +46,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fast_gettext", "~> 2.2.0"
   spec.add_dependency "nokogiri", "~> 1.13.1"
   spec.add_dependency "rexml", "~> 3.2.5"
-  spec.add_dependency "ruby-dbus", "~> 0.16.0"
+  spec.add_dependency "ruby-dbus", "~> 0.17.0"
```
imobachgs, line 49: :+1:

yastd/lib/yast2/dbus/service.rb
```diff
@@ -38,7 +38,7 @@ class Service
       SERVICE_NAME = "org.opensuse.YaST"
 
       # @return [String] D-Bus object path
-      OBJECT_PATH = "/org/opensuse/YaST/Installer"
+      OBJECT_PATH = "/org/opensuse/YaST/Installer1"
```
imobachgs, line 41: Now that we are using separate branches, you could use the same path. It is true that perhaps we could run both APIs in parallel (I do not know), but I do not think it is a valid use case by now.

yastd/lib/yast2/dbus/service.rb
```diff
@@ -50,7 +50,7 @@ def initialize(logger = nil)
       # Exports the installer object through the D-Bus service
       def export
         service = bus.request_service(SERVICE_NAME)
-        installer_obj = Yast2::DBus::Installer.new(
+        installer_obj = Yast2::DBus::Installer1.new(
```
imobachgs, line 53: Should not be Installer?

joseivanlopez (approved):

yastd/lib/yast2/dbus/installer.rb
```diff
@@ -23,163 +23,67 @@
...
-          logger.info "GetDisks"
-
-          disks = installer.disks.map do |disk|
-            {
-              "name"  => disk.name,
-              "model" => disk.model,
-              "size"  => disk.size.to_human_string
-            }
-          end
+      LANGUAGE_INTERFACE = "org.opensuse.YaST.Installer1.Language"
```
imobachgs, line 46: Not directly related to this PR, but I wonder if we could organize the interfaces in a different way. Otherwise, this file is going to grow a lot :-)

yastd/share/dbus.conf
```diff
@@ -11,6 +11,8 @@
     <allow send_destination="org.opensuse.YaST" />
     <allow send_destination="org.opensuse.YaST"
            send_interface="org.freedesktop.DBus.Introspectable"/>
+    <allow send_destination="org.opensuse.YaST.Installer1" />
+    <allow receive_sender="org.opensuse.YaST.Installer1" />
```
imobachgs, line 15: Still needed?

imobachgs (commented):
It looks much better using the ruby-dbus API :-)

yastd/lib/yast2/dbus/service.rb
jreidinger: well, this is intentional change. It is dbus convention to always include version number in object path...so you can provide multiple versions on dbus. BTW I think it would be nicer to have `/org/opensuse/DInstaller1` and we should probably also change service name.

yastd/lib/yast2/dbus/service.rb
jreidinger: yeap, thanks for catch

yastd/lib/yast2/dbus/installer.rb
jreidinger: yeap, @mvidner any hints? Maybe mixins for each interface?

yastd/share/dbus.conf
jreidinger: well, it is needed as it follows correct versioned interface. I just do not want to kill old one if you switch between branches.

yastd/lib/yast2/dbus/service.rb
imobachgs: Oh, I did not know that. For instance, DBus, NetworkManager, PackageKit or UPower do not follow that approach; but, on the other hand, FirewallD, UDisks or PolicyKit do it.

Interesting topic. And yes, we could think about renaming the service and the path as you propose.

yastd/share/dbus.conf
imobachgs: Yes, you are right, I did not know about the versioning. I wonder if we should drop the lines below:

```xml
    <allow send_destination="org.opensuse.YaST.Installer" />
    <allow receive_sender="org.opensuse.YaST.Installer" />
```

imobachgs: We will need to review the D-Bus policy in the future anyway, so let's keep it as it is by now. Thanks!

imobachgs (approved):

PR #28, opened 2022-02-17
RPM packaging

imobachgs:
This PR contains many changes to simplify the process of packaging the service and the web UI. 

## d-installer service

* Packaged as [rubygem-d-installer](https://build.opensuse.org/package/show/YaST:Head/rubygem-d-installer).
* Add a [gemspec](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/yastd/d-installer.gemspec). We have not released the software as a gem yet.
* Add the [D-Bus configuration](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/yastd/share/dbus.conf) and a [systemd service definition](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/web/package/_service).
* Include a [gem2rpm.yml](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/yastd/package/gem2rpm.yml) file to build the spec file from the sources.

## Web interface

* Packaged as [d-installer-web](https://build.opensuse.org/package/show/YaST:Head/d-installer-web).
* Follow the approach used by Cockpit modules.
* Add a [_service](https://github.com/yast/d-installer/blob/packaging/web/package/_service) to instruct OBS to fetch the sources, the dependencies and so on.
* Add a [spec file](https://github.com/yast/d-installer/blob/packaging/web/package/d-installer-web.spec).

## Helper scripts

* Update [deploy.sh](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/deploy.sh) to install from packages.
* Added a [setup.sh](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/setup.sh) to set-up the development environment.

## Documentation

The process to release a new version (or to update the packages) is documented in the [PACKAGING.md](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/deploy.shPACKAGING.md) file. Additionally, the [README.md](https://github.com/yast/d-installer/blob/7cc29b4ee1bff94ff1440892187a561f5938baec/README.md) has been updated and improved.

## Open questions

* What about the versioning policy? At this point, we are explicitly setting the version manually, but OBS offers alternative mechanisms to set the version (using tags, for instance).
* License: I am using `GPL-2.0-only` because that's what YaST uses. However, we are including a 3rd party `cockpit.js` file that uses `LGPL-2.1-or-later`. Is that correct?


web/package/d-installer-web.spec
```diff
@@ -0,0 +1,59 @@
...
+
+Name:           d-installer-web
+Version:        0
+Release:        0
+Summary:        Web-based user interface for D-Installer
+License:        GPL-2.0-only
+URL:            https://github.com/yast/d-installer
+# source_validator insists that if obscpio has no version then
+# tarball must neither
+Source:         d-installer-web.tar
```
jreidinger, line 27: why not compressed?

web/package/d-installer-web.spec
```diff
@@ -0,0 +1,59 @@
...
+Web-based user interface for the experimental YaST D-Installer.
+
+
+%prep
+%autosetup -p1 -n %{name}
+rm -f package-lock.json
+local-npm-registry %{_sourcedir} install --with=dev --legacy-peer-deps || ( find ~/.npm/_logs -name '*-debug.log' -print0 | xargs -0 cat; false)
+
+%build
+#cp -r %{_datadir}/cockpit/devel/lib src/lib
```
jreidinger, line 47: what is reason for this? just debug line?

web/package/d-installer-web.spec
```diff
@@ -0,0 +1,59 @@
...
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+
+Name:           d-installer-web
+Version:        0
```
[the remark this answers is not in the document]
imobachgs: We use the `set_version` service, which sets the version when building the package. One of the open questions that I would like to discuss during the review is about versioning taking advantage of OBS services.

web/package/d-installer-web.spec
jreidinger: so maybe we should document it, as it is different then what we have in yast, so it can confuse people.

web/package/d-installer-web.spec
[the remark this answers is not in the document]
jreidinger: you mean like having e.g. time based version? It would be ok for me, but question is what we will do in potentionally maintenance branches.

PR #31, opened 2022-02-22
Split dbus api

jreidinger:
follow up of #29 

Manually tested and now interface do what it should.


joseivanlopez (approved):
LGTM. We can continue from here.

PR #32, opened 2022-02-23
[web] Replace Create React App with Vite.js

imobachgs:
In our case, [Create React App](https://create-react-app.dev/) takes quite some time and memory to build the production bundle and to start the development server. Perhaps we are doing something wrong, but we decided to give a try to [Vite](https://vitejs.dev) instead.


web/README.md
```diff
@@ -7,21 +7,14 @@ and it was bootstrapped with [Create React App](https://github.com/facebook/crea
```
dgdavid, line 7: It still true that the application was bootstraped with Create React App... but not sure how relevant or confusing could that sentence be here.

dgdavid (approved):
LGTM.

PR #34, opened 2022-02-23
Users dbus api

jreidinger:
Ready for review. Tested using d-feet. Commit will be done as part of manager PR.

### DIFF
--- a/service/lib/dinstaller/dbus/users.rb
+++ b/service/lib/dinstaller/dbus/users.rb
@@ -18,6 +18,7 @@
 # find current contact information at www.suse.com.
 
 require "dbus"
+require "dinstaller/users"
 
 module DInstaller
   module DBus
@@ -39,7 +40,7 @@ module DInstaller
       end
 
       dbus_interface USERS_INTERFACE do
-        dbus_attr_reader :root_password_set, "b"
+        dbus_reader :root_password_set, "b"
 
         dbus_reader :root_ssh_key, "s", dbus_name: "RootSSHKey"
 
@@ -47,14 +48,14 @@ module DInstaller
 
         dbus_method :SetRootPassword, "in Value:s, in Encrypted:b" do |value, encrypted|
           logger.info "Setting Root Password"
-          assign_root_password(value, encrypted)
+          backend.assign_root_password(value, encrypted)
 
           self[DBus::PROPERTY_INTERFACE].PropertiesChanged(USERS_INTERFACE, {"RootPasswordSet" => !value.empty?}, [])
         end
 
         dbus_method :SetRootSSHKey, "in Value:s" do |value|
           logger.info "Setting Root ssh key"
-          assign_root_ssh_key(value)
+          backend.root_ssh_key=(value)
 
           self[DBus::PROPERTY_INTERFACE].PropertiesChanged(USERS_INTERFACE, {"RootSSHKey" => value}, [])
         end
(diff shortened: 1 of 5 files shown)


service/lib/dinstaller/dbus/users.rb
```diff
@@ -0,0 +1,96 @@
...
+        @logger = logger
+        @root_password_set = false
+
+        super(PATH)
+      end
+
+      dbus_interface USERS_INTERFACE do
+        dbus_reader :root_password_set, "b"
+
+        dbus_reader :root_ssh_key, "s", dbus_name: "RootSSHKey"
```
imobachgs, line 45: np: What about calling it RootSSHPublicKey?

service/lib/dinstaller/dbus/users.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        dbus_reader :first_user, "(ssba{sv})"
+
+        dbus_method :SetRootPassword, "in Value:s, in Encrypted:b" do |value, encrypted|
+          logger.info "Setting Root Password"
+          backend.assign_root_password(value, encrypted)
+
+          self[DBus::PROPERTY_INTERFACE].PropertiesChanged(USERS_INTERFACE, {"RootPasswordSet" => !value.empty?}, [])
```
imobachgs, line 53: Signaling that the password is set is fine; I wonder if `SetRootPassword` should return something.

imobachgs (commented):
The approach looks OK.

service/lib/dinstaller/dbus/users.rb
```diff
@@ -0,0 +1,96 @@
...
+      def root_password_set
+        backend.root_password?
+      end
+
+    private
+
+      attr_reader :logger
+
+      def backend
+        @backend = ::DInstaller::Users.instance.tap { |i| i.logger = @logger }
```
joseivanlopez, line 91: Why is backend assinged everytime? I would suggest to simply return the singleton instance.

service/lib/dinstaller/dbus/users.rb
jreidinger: well, only think that comes to my mind is encrypted password, but I am not sure if we have it, as it depends on default encryption method. What else you expect from it?

service/lib/dinstaller/dbus/users.rb
jreidinger: my fault. It should be `||=` that is also reason why I use tap to assign logger.

service/lib/dinstaller/dbus/users.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        FUSER_SIG = "in FullName:s, in UserName:s, in Password:s, in AutoLogin:b, in data:a{sv}"
+        dbus_method :SetFirstUser, FUSER_SIG do |full_name, user_name, password, auto_login, data|
+          logger.info "Setting first user #{full_name}"
+          backend.assign_first_user(full_name, user_name, password, auto_login, data)
+
+          PropertiesChanged(USERS_INTERFACE, {"FirstUser" => first_user}, [])
+        end
+
+
```
ancorgs, line 71: At which point in time do we expect to add Rubocop? :wink:

service/lib/dinstaller/dbus/users.rb
```diff
@@ -0,0 +1,96 @@
...
+    class Users < ::DBus::Object
+      PATH = "/org/opensuse/DInstaller/Users1".freeze
+      private_constant :PATH
+
+      USERS_INTERFACE = "org.opensuse.DInstaller.Users1".freeze
+      private_constant :USERS_INTERFACE
+
+      def initialize(logger)
+        @logger = logger
+        @root_password_set = false
```
ancorgs, line 37: This looks like a leftover from a previous implementation. That attribute is now delegated to the backend, isn't?

service/lib/dinstaller/dbus/users.rb
imobachgs: My main concern is not RuboCop, but RSpec ;-)

service/lib/dinstaller/dbus/users.rb
imobachgs: Perhaps if it was successful or not. But I guess that it is something we should discuss for other endpoints too, so let's keep it as it is by now.

ancorgs (approved):
LGTM

PR #42, opened 2022-02-25
initial API proposal for manager

jreidinger:
Goal of this PR is to write top level manager API that control probing and installing. It is also responsible for providing reasonable progress information back to user.


service/lib/dinstaller/dbus/manager.rb
```diff
@@ -41,21 +41,34 @@ def initialize(installer, logger)
...
-
         super(PATH)
       end
 
       dbus_interface MANAGER_INTERFACE do
-        dbus_method :probe, "out result:u" do
-          # TODO
-          0
+        dbus_method :probe, "" do
+          # TODO: do it assynchronous. How? ractors will have problem with sharing yast data.
```
imobachgs, line 49: np: asynchronous

We need to do some research here. By now, we have been using threads and, although it is far from ideal, it worked. But sure we need a better solution. Ideally, we could have an actors-like solution where every subsystem is represented by an actor and they share nothing. But that's not realistic at this point.

jreidinger: sadly with way how we using yast I worry it will be very challenging

mvidner:
About asynchronous calls and D-Bus:
(This is probably different from what you've started to discuss, but I want to summarize it anyway)

In the old times (before 2009) the bus itself (dbus-daemon) would have a method call timeout (about 30 seconds) so things taking a long time had to be designed as void methods + signals bearing the reply.

But that has [changed][1] (suprisingly to me), and now [the D-Bus API Design Guildelines say](https://dbus.freedesktop.org/doc/dbus-api-design.html#using-the-features) that a normal call+reply should be used, possibly disabling a timeout implemented in the language binding:

[1]: https://bugs.freedesktop.org/show_bug.cgi?id=76112

> Note that D-Bus client bindings may implement synthetic timeouts of several tens of seconds, unless explicitly disabled for a call. For very long-running operations, you should disable the client bindings’ timeout and make it clear in the client’s UI that the application has not frozen and is simply running a long operation.

> An anti-pattern to avoid in this situation is to start a long-running operation when a method call is received, then to never reply to the method call and instead notify the client of completion of the operation via a signal. This approach is incorrect as signal emissions do not have a one-to-one relationship with method calls — the signal could theoretically be emitted multiple times, or never, which the client would not be able to handle.

Anyway, ruby-dbus does not currently impose a timeout on method calls, but there is an issue (https://github.com/mvidner/ruby-dbus/issues/61) asking for it.

mvidner:
@jreidinger please remember to fill the PR description. It provides CONTEXT for people who happen to not have a copy of your brain. A link to Bugzilla/Trello is a good start.

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -41,21 +41,34 @@ def initialize(installer, logger)
...
+        # 1 : probing
+        # 2 : probed
+        # 3 : installing
+        # 4 : installed
+        dbus_reader :status, "u"
+
+        # Progress has struct with values:
+        # s message
+        # t total major steps to do
+        # t current major step
```
mvidner, line 68: starting at 0 or 1?

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -41,21 +41,34 @@ def initialize(installer, logger)
...
+        # t total minor steps. Can be zero which means no minor steps
+        # t current minor step
+        dbus_reader :progress, "(stttt)"
```
mvidner, line 71: So a D-Bus client can poll the progress by reading the property, or listen for PropertiesChanged signals, right?

InstallationProgress has `@dbus_obj&.Progress(...args...)`. How is that related? Which class (interface) is `@dbus_obj`?

service/lib/dinstaller/dbus/manager.rb
jreidinger: yeah, the most tricky quetion in programming. It starts with 0.

service/lib/dinstaller/dbus/manager.rb
jreidinger: Exactly. I see that pattern often.
InstallationProgress will be dropped. Plan is to use  Progress.new in dbus Manager object and assign callback to it that raise signal. And attach progress to data structure read from that progress.object.

service/lib/dinstaller/dbus/manager.rb
mvidner:
```suggestion
        # t current major step (0-based)
```

service/lib/dinstaller/manager.rb
```diff
@@ -138,53 +120,36 @@ def storage_proposal
...
-
-      progress = InstallationProgress.new(@dbus_obj, logger: logger)
-      progress.partitioning do |_|
+      Thread.new do
+        change_status(InstallerStatus::INSTALLING)
+        progress.init_progress(3, "Partitioning")
+        Yast::Installation.destdir = "/mnt"
+        # lets propose it here to be sure that software proposal reflects product selection
+        # FIXME: maybe repropose after product selection change?
+        # first make bootloader proposal to be sure that required packages is installed
```
joseivanlopez, line 129: Typo: are installed.

service/lib/dinstaller/manager.rb
```diff
@@ -100,19 +74,27 @@ def options
...
-      probe_storage
-      @software.probe
-      true
-    rescue StandardError => e
-      logger.error "Probing error: #{e.inspect}"
-      false
-    ensure
-      change_status(InstallerStatus::IDLE)
+      Thread.new do
+        sleep(1) # do sleep to ensure that dbus service is already attached
```
joseivanlopez, line 79: I guess this would not be needed if probing is launched by a client calling to `Manager#probe` dbus method, right?

service/lib/dinstaller/installer_status.rb
```diff
@@ -23,29 +23,16 @@
 module DInstaller
   # This class represents the installer status
   class InstallerStatus
```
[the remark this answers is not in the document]
joseivanlopez: Idea: What about having something like `DInstaller::StatusManager` class, which defines possible statuses, implement logic for changing status, for calling callbacks and so on? And `DInstaller::Manager` would simply instanciate a `StatusManager` and would request a status change when needed.

service/lib/dinstaller/progress.rb
```diff
@@ -0,0 +1,108 @@
...
+  #
+  # @example module interaction
+  #
+  #   def step2(progress)
+  #     progress.init_minor_steps(2, "Doing subtask1")
+  #     subtask1
+  #     progress.next_minor_step("Doing subtask2")
+  #     subtask2
+  #     progress.next_minor_step("Finished subtask2")
+  class Progress
```
joseivanlopez, line 50: Maybe it would too much, but I also imagine this as a kind of `ProgressManager`, generating steps when requested:

~~~
steps_manager = StepsManager.new(steps: 3)
steps_manager.next_step("Doing step1") #=> DInstaller::Step
do_step1(steps_manager)
steps_manager.next_step("Doing step2")
do_step2(steps_manager)
steps_manager.next_step("Doing step3")
do_step3(steps_manager)
...

def do_step1(steps_manager)
  current_step = steps_manager.current_step
  sub_steps_manager = current_step.sub_steps(2) #=> DInstaller::StepsManager
  sub_steps_manager.next_step("Doing" subtask1")
  subtask1
  sub_steps_manager.next_step("Doing" subtask2")
  subtask2
~~~ 

This would allow to generate a tree of steps: step1 -> step 1.1 -> step 1.1.1. With this, we could also have different callbacks for substeps, etc. But as said, maybe is too much.

service/lib/dinstaller/installer_status.rb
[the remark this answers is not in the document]
jreidinger: Well, I am not sure. I try to not touch this part :) Just change it from string to unsigned.

service/lib/dinstaller/manager.rb
jreidinger: yeap, but my goal is to get to probed status as soon as possible. That is why I do it when starting service.

service/lib/dinstaller/progress.rb
jreidinger: Well, having tree of steps is something that is hard to imagine for me how to display it in UI. I try to keep it simple now. So top level manager knows about top level steps and then module knows about module specific substeps if needed. ( from my current testing some steps like language probe is super fast, that it is even hard to notice it ).

imobachgs (approved):

PR #44, opened 2022-02-28
Set up proposed layout

dgdavid:
## TL;DR

Set up the [proposed layout](https://yast.opensuse.org/blog/2022-02-23/start-2022#dinstaller) for D-Installer.

| Login Screen | Installation Summary | Popup | Progress screen |
|-|-|-|-|
| Screenshot_2022-03-06-01-43-53-477_com android chrome | Screenshot_2022-03-06-00-41-00-585_com android chrome | Screenshot_2022-03-06-01-05-25-295_com android chrome | Screenshot_2022-03-05-21-34-12-247_com android chrome |

<p align="center"><em>How D-Installer looks in a 6.55" Android smartphone (using dummy data)</em></p>

---

| Login Screen | Installation Summary | Popup | Progress screen |
|-|-|-|-|
| Screen Shot 2022-03-06 at 00 58 48 | Screen Shot 2022-03-06 at 01 04 58 | Screen Shot 2022-03-06 at 01 04 43 | Screen Shot 2022-03-06 at 00 59 11 |

<p align="center"><em>How D-Installer looks in a desktop Firefox browser but emulating a landscaped device (1080x810) using the Responsive Design Mode and dummy data</em></p>

---

| Login Screen | Installation Summary | Popup | Progress screen |
|-|-|-|-|
| Screenshot 2022-03-06 at 01-02-24 D-Installer | Screenshot 2022-03-06 at 01-02-09 D-Installer | Screenshot 2022-03-06 at 01-01-59 D-Installer | Screenshot 2022-03-06 at 01-03-15 D-Installer |

<p align="center"><em>How D-Installer looks in Firefox in a 1920x1080 (16:9) laptop screen (using dummy data)</em></p>

## The Details

As discussed internally, going for a tiny, single-column, mobile first layout will not only help to have a multi-device installer, but ~~force~~ encourage us to develop a simple, yet powerful, UI.

Thus, this PR can be considered as a starting point for laying out the bases to achieve that end as well as a post collecting both, taken decisions and problems found.

### Fixed header

Having a single column layout somehow means having vertical scroll. It is something (almost) not present in current YaST UI, but quite common in web and mobile applications. For sure, we are committed to do our best for having not so long _screens_, but we have to ensure some context and actions are always visible/at hand for the user. For that reason, a fixed header has been implemented. The idea is to hold there a possible main menu, section title and maybe other actions (theme switcher, expert actions).

In the future, we could even consider making the header disappear when the user scroll down and making it appear again when scrolling up. A quite common patter nowadays, though.

https://www.codemzy.com/blog/react-sticky-header-disappear-scroll

### Fixed footer

Slightly related with the fixed header, there is a fixed footer to display either, actions or (short) messages. Or event both at the same time.

Although it is nothing hard to achieve in web development (we even see it all the time thanks to the infamous cookies warnings), we found an interesting problem in mobile browsers: the viewport if positioned partially offscreen because the URL Bar—which is hidden when user scrolls down—move it out when it is on the screen.

Fortunately, looks like most of mobile browsers has fixed the issue for elements that are `position: fixed`. At least, it is the case for Chrome, Firefox, and Brave on Android. Although not for DuckDuckGo :cry:

<details>
<summary>:point_right: Click to show/hide screenshots.</summary>

---
| URL bar **NOT** hidden | URL bar **hidden** |
|-|-|
| Screenshot_2022_03_06_00_40_21_562_com_duckduckgo_mobile_android | Screenshot_2022_03_06_00_40_29_859_com_duckduckgo_mobile_android |

<p align="center"><em>How the fixed footer looks in DuckDuckGo browser on
Android devices for both address bar statuses, displayed and hidden</em></p>

</details>

If interested, know more by reading below links

  * https://stackoverflow.com/a/57748562
  * https://developers.google.com/web/updates/2016/12/url-bar-resizing
  * https://css-tricks.com/the-trick-to-viewport-units-on-mobile/

And these others to follow the progress of DuckDuckGo issue

  * https://github.com/duckduckgo/Android/issues/638
  * https://github.com/duckduckgo/Android/issues/914

### Thinking on i18n

Despite the fact the code is not ready for translations yet, thinking on i18n it's important for us. But i18n goes beyond the content itself. It is all about 

> develop your content, application, specification, and so on, in a way that
> ensures it will work well for, or can be easily adapted for, users from any
> culture, region, or language. — https://www.w3.org/standards/webdesign/i18n

Our experience with YaST tell us that an installer has a wide target audience, reason why we don't want to elude having i18n in its full sense into account. Thus, we have started using [CSS Logical Properties](https://www.w3.org/TR/css-logical-1) instead directional ones. I.e., _margin-block-end_ instead of _margin-right_ and so on.

Yes, they still a [working draft](https://www.w3.org/2020/Process-20200915/#maturity-levels)  but latest browsers versions on both, desktop and mobile devices, have a good support for them. (check https://caniuse.com/css-logical-props). Moreover, PatternFly (the component library we are using) currently support [**only the latest versions** of Chrome, Firefox, Safari, and Edge browsers](https://www.patternfly.org/v4/get-started/about/#supported-browsers).

<details>
<summary>:point_right:  Click to show/hide screenshots and see the difference between using or not those logical properties.</summary>

---

  - **NOT USING** CSS logical props

    | `direction: rtl` | `direction: rtl`and `writing-mode: sideway-lr` |
    |-|-|
    | before-direction-rtl | before-direction-rtl-wm-sideway-lr |

 - **USING** CSS logical props

    | `direction: rtl` | `direction: rtl`and `writing-mode: sideway-lr` |
    |-|-|
    | Screen Shot 2022-03-06 at 02 09 01 | Screen Shot 2022-03-06 at 02 09 12 |

<p align="center"><em>NOTE that those `direction` and `writing-mode` usages can make not sense without a right font/script. But they have been used just for illustration how the layout is _more_ in the right place when using CSS Logical Properties.</em></p>

</details>

#### Caveats

Beyond the browsers support, there are three foreseeable handicaps on the way to achieve a good i18n

  * We are not (web) experts on that regard (yet :wink:)
  * PatternFly does not use CSS Logical Properties at all
  * SVG Icons does not follow [writing mode](https://www.w3.org/TR/css-writing-modes-3/) (?)

Still, none of them should be an obstacle for continuing along this path because 

  * We ~~must~~ can learn about it and there are a plenty of resources to do so
    * https://chenhuijing.com/blog/css-for-i18n/
    * https://www.w3.org/International/questions/qa-html-dir 
    * https://web.dev/learn/design/internationalization
    * https://moderncss.dev/developing-for-imperfect-future-proofing-css-styles
    * https://rtlstyling.com/posts/rtl-styling
    * ...
  * We can override PatternFly CSS rules by now (and look for the best structural way to do it) until CSS Logical Properties reach the recommendation stage and, eventually, collaborate with upstream to adopt them.

Certainly, although it looks unlikely at its status of implementation, ~~shit happens~~ anything can happen and the CSS Logical Properties working draft could end up not reaching the [candidate recommendation](https://www.w3.org/2020/Process-20200915/#RecsCR) stage. But even in such case, the impact should not be so huge. A matter of _rolling back_ to directional properties, maybe with a script similar to this one for converting [to-logical](https://gist.github.com/nyurik/d438cb56a9059a0660ce4176ef94576f).

In short, betting for CSS Logical Properties looks safe and helps to have the CSS _i18n ready_ from the very beginning.

## What's next?

This set of changes is ready to be reviewed and merged (as long as nobody is against). However, a lot of (UI) work remains ahead, such as

  - Improve the layout component and its usage

    For this first iteration, the layout is imported in each wrapped component and it receives the section title, footer messages, and footer actions via props. Enough for prototyping, but perhaps another mechanism should be chosen for using the layout in a single point and letting it know everything it needs. A _layout context_? Don't know at this moment, but it is something to be discussed, especially when implementing the navigation among sections ([React Router](https://reactrouter.com)?).

  - Decides if fixed footer should be displayed in the login screen too

    Placing the Login button in the fixed footer (and re-enabling the return keystroke for triggering the form send).

  - Set a minimum size/orientation supported (?)

    Admittedly, having a responsive, mobile first UI is nice. But making the proposed single column layout useful in the landscape mode of small devices is hard (to not say impossible).

    <details>
    <summary>:point_right:  Click to show/hide screenshots taken in a 6.55" screen size Android smartphone</summary>

    ---

    | Portrait | Landscape |
    |-|-|
    | Screenshot_2022-03-06-00-41-00-585_com android chrome | Screenshot_2022-03-06-00-41-18-465_com android chrome |

    _It works_ in both orientations, but it does not feel so neat and tidy in landscape mode.

    </details>

    Should we block the landscape mode below a certain screen size, encouraging the user to rotate the device? See https://code-boxx.com/lock-screen-orientation

  - Investigate if tree shaking or any other _purge_ technique can be applied to (S)CSS

    For not including not needed parts of either, EOS or PatternFly, in the final build.

    * https://purgecss.com/getting-started.html
    * https://markmurray.co/blog/tree-shaking-css-modules/

  - Define reasonable layout width for each breakpoint and re-evaluate if still using PatternFly utilty classes for it or go for a custom media query instead. 

  - Extract some common components

    Like the one for centering the content inside the layout: a PatternFly Bullseye with a class name for fixing the ["percentage-sized children should only respect containers with an explicit height"](https://bugs.webkit.org/show_bug.cgi?id=26559#c31) _issue_ (we only set the _min-block-size_ for filling the full screen even if the content is shorter)

  - Tests (jest)

    In a next iteration, once the layout has been approved (and hopefully more defined)

  - General CSS improvements.

    * **Improve variables naming** (improves variables in general)
    * Defines how to organize PatternFly overrides rules
    * ...

### DIFF
--- a/web/src/layout.scss
+++ b/web/src/layout.scss
@@ -5,54 +5,56 @@
 @use "eos-ds/dist/scss/eos-base/variables/branding";
 
 .layout {
+  --header-height: 8ex;
+  --footer-height: calc(2 * var(--header-height));
+
   @extend .pf-u-box-shadow-md;
 
   background-color: white;
   margin-left: auto;
   margin-right: auto;
-  --header-height: 8ex;
-  --footer-height: calc(2 * var(--header-height));
 }
 
 .layout__header {
   @extend .pf-u-box-shadow-md-bottom;
 
-  background-color: branding.$eos-bc-pine-500;
-  color: white;
-
-  height: var(--header-height);
   // Note: "position: fixed" + "width: inherit" does not work here. Sticky is enough for the header
   position: sticky;
   z-index: 100;
   top: 0;
 
+  background-color: branding.$eos-bc-pine-500;
+  color: white;
+
+  height: var(--header-height);
+  padding: 0.5rem;
+
   display: flex;
   align-items: center;
   gap: 0.5rem;
-  padding: 0.5rem;
 }
 
-.layout__header__left-action {
+.layout__header-left-action {
 }
 
-.layout__header__action-icon {
+.layout__header-action-icon {
   fill: white;
-  width: 1.5rem;
   height: 1.5rem;
+  width: 1.5rem;
   vertical-align: middle;
 }
 
-.layout__header__section-title {
+.layout__header-section-title {
   font-size: 1.5rem;
 }
 
-.layout__header__section-title-icon {
+.layout__header-section-title-icon {
   fill: white;
   // Sadly, we can't use font-size with EOS Icons
-  width: 1em;
   height: 1em;
+  width: 1em;
   vertical-align: sub;
-  margin: 0;
+  margin-right: 0.5em;
 }
 
 .layout__content {
@@ -63,17 +65,19 @@
 
 .layout__footer {
   @extend .pf-u-box-shadow-sm-top;
-  background-color: white;
 
   position: fixed;
   bottom: 0;
+
+  background-color: white;
+
   height: var(--footer-height);
   width: inherit;
+  padding: 1rem;
+
   display: grid;
   grid-template-columns: 2fr 1fr;
   gap: 1rem 1rem;
-  padding: 1rem;
-  
 }
 
 .layout__footer-info-area {
--- a/web/src/layout.scss
+++ b/web/src/layout.scss
@@ -9,8 +9,8 @@
 
   background: white;
   margin: 0 auto;
-  --header-height: 48px;
-  --footer-height: 96px;
+  --header-height: 8ex;
+  --footer-height: calc(2 * var(--header-height));
 }
 
 .layout__header {
--- a/web/src/layout.scss
+++ b/web/src/layout.scss
@@ -1,0 +1,91 @@
+// See https://github.com/patternfly/patternfly/pull/4297
+@use "@patternfly/react-styles/css/utilities/BoxShadow/box-shadow.css";
+@use "@patternfly/react-styles/css/utilities/Text/text.css";
+
+@use "eos-ds/dist/scss/eos-base/variables/branding";
+
+.layout {
+  @extend .pf-u-box-shadow-md;
+
+  background: white;
+  margin: 0 auto;
+  --header-height: 48px;
+  --footer-height: 96px;
+}
+
+.layout__header {
+  @extend .pf-u-box-shadow-md-bottom;
+
+  background-color: branding.$eos-bc-pine-500;
+  color: white;
+
+  height: var(--header-height);
+  // Note: "position: fixed" + "width: inherit" does not work here. Sticky is enough for the header
+  position: sticky;
+  z-index: 100;
+  top: 0;
+
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  padding: 0.5rem;
+}
+
+.layout__header__left-action {
+}
+
+.layout__header__action-icon {
+  fill: white;
+  width: 1.5rem;
+  height: 1.5rem;
+  vertical-align: middle;
+}
+
+.layout__header__section-title {
+  font-size: 1.5rem;
+}
+
+.layout__header__section-title-icon {
+  fill: white;
+  // Sadly, we can't use font-size with EOS Icons
+  width: 1em;
+  height: 1em;
+  vertical-align: sub;
+  margin: 0;
+}
+
+.layout__content {
+  padding: 1rem;
+  padding-bottom: calc(2rem + var(--footer-height));
+  min-height: calc(100vh - var(--header-height));
+}
+
+.layout__footer {
+  @extend .pf-u-box-shadow-sm-top;
+  background: white;
+
+  position: fixed;
+  bottom: 0;
+  height: var(--footer-height);
+  width: inherit;
+  display: grid;
+  grid-template-columns: 2fr 1fr;
+  gap: 1rem 1rem;
+  padding: 1rem;
+  
+}
+
+.layout__footer-info-area {
+  align-self: center;
+  overflow-y: auto;
+}
+
+// TODO: Keep alert icons vetically centered here?
+// .layout__footer-info-area .pf-c-alert__icon {
+//   align-self: center;
+// }
+
+.layout__footer-actions-area {
+  align-self: center;
+  text-align: center;
+}
--- a/web/src/layout.scss
+++ b/web/src/layout.scss
@@ -5,14 +5,16 @@
 @use "eos-ds/dist/scss/eos-base/variables/branding";
 
 .layout {
-  --header-height: 8ex;
-  --footer-height: calc(2 * var(--header-height));
+  --header-block-size: 8ex;
+  --footer-block-size: calc(2 * var(--header-block-size));
 
   @extend .pf-u-box-shadow-md;
 
   background-color: white;
-  margin-left: auto;
-  margin-right: auto;
+  margin-inline: auto;
+
+  // Needed for "telling" the fixed footer its (inherited) inline-size.
+  inline-size: 100%;
 }
 
 .layout__header {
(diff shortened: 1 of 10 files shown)


mvidner (commented):
Thanks!

I don't know the coding languages involved, so instead let me try taking
a fresh **outsider user's look**:

"Welcome to D-Installer"
- what, why should I log in if this is fresh hardware yet to be installed?
- who is behind this? (IMHO we should include a small but omnipresent branding
  bit: made by SUSE)

Regarding your choices about **layout**, it sounds good.

About **i18n**...

Are you thinking about right-to-left and bidirectional writing
already? Wow, OK, we should use logical directions instead of hardcoding
lefts and rights.

IMHO more important is to set good expectations about the potential difference
between the web UI language and the target system language. Maybe we want to
force them to be the same. OTOH we do have many target localizations and none
of the web yet.

dgdavid:
> "Welcome to D-Installer"
> 
>     * what, why should I log in if this is fresh hardware yet to be installed?

Really good question. I'll ask @imobachgs, :pray: 

> 
>     * who is behind this? (IMHO we should include a small but omnipresent branding
>       bit: made by SUSE)

Completely agree. I even added a background (that one with a lot of horizontal lines in the middle) from _internal_ https://brand.suse.com/brand-system/imagery... but I discarded it because I do not know about implications and so on. That's why there is only gray at the background and none SUSE logos around. Any idea is welcome!

imobachgs:
> > "Welcome to D-Installer"
> > ```
> > * what, why should I log in if this is fresh hardware yet to be installed?
> > ```

That's a good question. The problem is that we need to log in to Cockpit, so we need a user and a password. We might need to research whether it is possible a password-less authentication but only for local installations. If you want connect remotely, I would say that you better have to enter a user and a password. Actually, we have [a Trello card](https://trello.com/c/dxt8I2CN/5078-d-installer-research-skip-the-authentication-when-connecting-from-localhost) for this.

web/src/layout.scss
```diff
@@ -0,0 +1,118 @@
...
+  grid-template-columns: 2fr 1fr;
+  gap: 1rem 1rem;
+}
+
+.layout__footer-info-area {
+  align-self: center;
+  overflow-y: auto;
+}
+
+// TODO: Keep alert icons vetically centered here?
```
imobachgs, line 110: typo: vertically

imobachgs (approved):

dgdavid:
BTW


> 
> About **i18n**...
> 
> Are you thinking about right-to-left and bidirectional writing already?

Yes, I am. I see no reason for delaying it :wink:

kwwii:
- I don't understand how you point this at a machine, or which machine will be installed. 
-- Why would I be installing linux on my cell phone? What is the use-case here, and why is it predominate to the desktop installation? 
- The user name and password feel more like a login than an entry field for the settings of the machine to install.
- I suggest putting the warning text in one centered line with the Install button below (also centered and at full width)

imobachgs:
Hi Ken,

> * I don't understand how you point this at a machine, or which machine will be installed.
>       -- Why would I be installing linux on my cell phone? What is the use-case here, and why is it predominate to the desktop installation?

I would say that the main use case is remote installation. It is not that you are installing anything in your phone, but you can connect to the machine you are installing remotely without needing to use SSH or VNC. You can even use your mobile phone, that's all.

>     * The user name and password feel more like a login than an entry field for the settings of the machine to install.

Yes, it is. If you are going to install remotely, you need to log in (like when using SSH) into the system you want to install. Ideally, we should skip the login screen for a local installation, but unfortunately, we are not at that point yet (we eventually will). We have a [Trello card](https://trello.com/c/dxt8I2CN/) for that.

>     * I suggest putting the warning text in one centered line with the Install button below (also centered and at full width)

Would be that text just above the button or at the top of the screen?

Thanks for your feedback!

dgdavid:
For the record, 

In order to avoid confusion, I have updated a bit the description changing

>  will not only help to have a multi-device installer

by 

> will not only help to have an installer UI accessible from differently-sized devices

Any suggestion for improving the message is more than welcome.

PR #45, opened 2022-03-03
Storage D-Bus

joseivanlopez:
Add initial D-Bus API for storage stuff. For now, it only covers a very simplistic storage proposal and the list of actions.

### DIFF
--- a/service/lib/dinstaller/dbus/manager.rb
+++ b/service/lib/dinstaller/dbus/manager.rb
@@ -31,7 +31,7 @@ module DInstaller
       PATH = "/org/opensuse/DInstaller/Manager1"
       private_constant :PATH
 
-      MANAGER_INTERFACE = "org.opensuse.DInstaler.Manager1"
+      MANAGER_INTERFACE = "org.opensuse.DInstaller.Manager1"
       private_constant :MANAGER_INTERFACE
 
       attr_reader :logger
@@ -51,11 +51,11 @@ module DInstaller
       end
 
       dbus_interface MANAGER_INTERFACE do
-        dbus_method :probe, "" do
+        dbus_method :Probe, "" do
           @backend.probe
         end
 
-        dbus_method :commit, "" do
+        dbus_method :Commit, "" do
           @backend.commit
         end
 
--- a/service/lib/dinstaller/dbus/service.rb
+++ b/service/lib/dinstaller/dbus/service.rb
@@ -24,20 +24,19 @@ require "dinstaller/dbus/manager"
 require "dinstaller/dbus/language"
 require "dinstaller/dbus/software"
 require "dinstaller/dbus/users"
+require "dinstaller/dbus/storage/proposal"
+require "dinstaller/dbus/storage/actions"
 require "dinstaller/manager"
 
 module DInstaller
   module DBus
-    # YaST D-Bus service (org.opensuse.DInstaller)
+    # D-Bus service (org.opensuse.DInstaller)
     #
-    # It connects to the system D-Bus and answers requests on `/org/opensuse/DInstaller/*`.
+    # It connects to the system D-Bus and answers requests on objets below
+    # `/org/opensuse/DInstaller`.
     #
     # @example Running the server
-    #   s = DInstaller::DBus::Service.new
-    #   s.export
-    #   some_main_loop { s.dispatch }
-    #
-    # @see Yast2::DBus::Installer
+    #   DInstaller::DBus::Service.new.run
     class Service
       # @return [String] service name
       SERVICE_NAME = "org.opensuse.DInstaller"
@@ -75,11 +74,18 @@ module DInstaller
 
       # @return [Array<::DBus::Object>]
       def dbus_objects
-        @dbus_objects ||= [manager_bus, language_dbus, software_dbus, users_dbus]
+        @dbus_objects ||= [
+          manager_dbus,
+          language_dbus,
+          software_dbus,
+          users_dbus,
+          storage_proposal_dbus,
+          storage_actions_dbus
+        ]
       end
 
-      def manager_bus
-        @manager_bus ||= DInstaller::DBus::Manager.new(@logger)
+      def manager_dbus
+        @manager_dbus ||= DInstaller::DBus::Manager.new(@logger)
       end
 
       def language_dbus
@@ -92,6 +98,20 @@ module DInstaller
 
       def users_dbus
         @users_dbus ||= DInstaller::DBus::Users.new(@logger)
+      end
+
+      def storage_proposal_dbus
+        @storage_proposal_dbus ||= DInstaller::DBus::Storage::Proposal.new(@logger)
+      end
+
+      def storage_actions_dbus
+        @storage_actions_dbus ||= DInstaller::DBus::Storage::Actions.new(@logger)
+      end
+
+      def installer
+        # TODO: this god object should not be needed anymore when all dbus API is adapted
+        # just that probe should  be kept to get installer probed ASAP
+        @installer ||= DInstaller::Manager.instance.tap(&:probe)
       end
     end
   end
--- a/service/lib/dinstaller/dbus/service.rb
+++ b/service/lib/dinstaller/dbus/service.rb
@@ -107,12 +107,6 @@ module DInstaller
       def storage_actions_dbus
         @storage_actions_dbus ||= DInstaller::DBus::Storage::Actions.new(@logger)
       end
-
-      def installer
-        # TODO: this god object should not be needed anymore when all dbus API is adapted
-        # just that probe should  be kept to get installer probed ASAP
-        @installer ||= DInstaller::Manager.instance.tap(&:probe)
-      end
     end
   end
 end
--- a/service/lib/dinstaller/dbus/storage/actions.rb
+++ b/service/lib/dinstaller/dbus/storage/actions.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "dbus"
+require "dinstaller/storage/actions"
+
+module DInstaller
+  module DBus
+    module Storage
+      # D-Bus object to get the list of actions to perform in the system
+      class Actions < ::DBus::Object
+        PATH = "/org/opensuse/DInstaller/Storage/Actions1"
+        private_constant :PATH
+
+        INTERFACE = "org.opensuse.DInstaller.Storage.Actions1"
+        private_constant :INTERFACE
+
+        def initialize(logger)
+          @logger = logger
+
+          super(PATH)
+        end
+
+        dbus_interface INTERFACE do
+          # TODO: emit a signal when actions change (e.g., when a proposal is calculated). Right
+          #   now, actions changes can only be detected by subcribing to PropertiesChange signal
+          #   from Storage::Proposal dbus object.
+          dbus_reader :all, "aa{sv}"
+        end
+
+        def all
+          backend.all.map { |a| to_dbus(a) }
+        end
+
+      private
+
+        def to_dbus(action)
+          { "Text" => backend.text_for(action), "Subvol" => backend.subvol_action?(action) }
+        end
+
+        def backend
+          @backend ||= ::DInstaller::Storage::Actions.instance
+        end
+      end
+    end
+  end
+end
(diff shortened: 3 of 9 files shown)


service/lib/dinstaller/dbus/storage/guided_proposal.rb
```diff
@@ -0,0 +1,78 @@
...
+
+          dbus_method :Calculate, "in settings:a{sv}, out result:u" do |settings|
+            backend.calculate(settings)
+
+            self.ProposalChanged
+
+            backend.proposal.failed? ? 1 : 0
+          end
+
+          dbus_signal :ProposalChanged
```
jreidinger, line 59: this should not be needed. It should be done automatic by ruby-dbus. See other dbus that generate properties signal.

service/lib/dinstaller/dbus/storage/guided_proposal.rb
```diff
@@ -0,0 +1,78 @@
...
+          dbus_method :AvailableDevices, "out devices:as" do
+            backend.available_devices
+          end
+
+          dbus_method :Calculate, "in settings:a{sv}, out result:u" do |settings|
+            backend.calculate(settings)
+
+            self.ProposalChanged
+
+            backend.proposal.failed? ? 1 : 0
```
jreidinger, line 56: the result enum needs to be for sure documented.

service/lib/dinstaller/manager.rb
```diff
@@ -167,54 +170,11 @@ def probe_languages
...
-      @status_callbacks = []
-      @status = InstallerStatus::ERROR # it should start with probing, so just temporary status
-      @logger = logger || Logger.new($stdout)
-      @progress = Progress.new
-      @software = Software.new(@logger)
-      # Set stage to initial, so it will act as installer for some cases like
-      # proposing installer instead of reading current one
-      Yast::Stage.Set("initial")
+      logger.info "Probing storage and performing proposal"
+      Y2Storage::StorageManager.instance.probe
```
jreidinger, line 176: is possible to get some progress from that probing?

ancorgs: Well, potentially libstorage-ng can trigger a callback for every step of the process. But it's not exactly progress reporting (nothing like a percentage or so), it's more a live log of the actions libstorage-ng is performing. See https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/callbacks/probe.rb#L57

service/lib/dinstaller/dbus/storage/guided_proposal.rb
joseivanlopez: Ok, let's use *PropertiesChanged*.

service/lib/dinstaller/manager.rb
jreidinger: @ancorgs well, progress also does not have percentages, but it has substeps there. So if you know number of steps, then reporting is fine. If not, then we can add ability to progress API to have like living substeps that just change message and does not do counting of substeps.

service/d-installer.gemspec
```diff
@@ -35,6 +35,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.5.0"
 
+  spec.add_development_dependency "byebug"
```
jreidinger, line 38: well, I am not sure if we should add byebug to devel dependencies because debug tool is developer preference and also for TW there are already debug gem that can replace byebug and has some additional features. rake and packaging tasks is something that needs everyone who use rake there. But debugger is tool that developer use.

service/lib/dinstaller/dbus/storage/commit_actions.rb
```diff
@@ -0,0 +1,68 @@
...
+            actions_text(backend.all_actions)
+          end
+
+          dbus_method :GetMainActions, "out actions:as" do
+            actions_text(backend.main_actions)
+          end
+
+          dbus_method :GetSubvolumeActions, "out actions:as" do
+            actions_text(backend.subvolume_actions)
+          end
```
jreidinger, line 53: Well, question. Why not use for that properties and having signal properties changed to catch when actions changed? Now you need to poll actions to see it changes? Or is there other way to see that actions are changed?

service/lib/dinstaller/dbus/storage/guided_proposal.rb
```diff
@@ -0,0 +1,77 @@
...
+
+          dbus_method :AvailableDevices, "out devices:as" do
+            backend.available_devices
+          end
+
+          # result: 0 success; 1 error
+          dbus_method :Calculate, "in settings:a{sv}, out result:u" do |settings|
+            backend.calculate(settings)
+
+            PropertiesChanged(INTERFACE, settings, [])
```
jreidinger, line 55: Maybe I am wrong. But this looks like wrong usage of PropertiesChanged signal. Or is there really direct mapping of "LVM" and "CandidateDevices" strings and its values and assigned?

I expect that it looks like
```
PropertiesChanged(INTERFACE, { "LVM" => lvm, "CandidateDevices" => candidate_devices}, [])
```

service/lib/dinstaller/manager.rb
```diff
@@ -49,9 +49,7 @@ class InvalidValue < StandardError; end
     DEFAULT_LANGUAGE = "en_US"
 
     # TODO: move to own module classes
-    attr_reader :disks, :languages
-    # TODO: move to own module classes
-    attr_reader :disk
+    attr_reader :languages
```
jreidinger, line 52: :+1:

service/lib/dinstaller/manager.rb
```diff
@@ -135,6 +123,8 @@ def install
         proposal = ::Bootloader::ProposalClient.new.make_proposal({})
         logger.info "Bootloader proposal #{proposal.inspect}"
         @software.propose
+        # TODO: install required storage tools
```
jreidinger, line 126: well, where it have to be? in live CD? if so, then we need different software backend as software used here for propose targets target system and not live one.

service/lib/dinstaller/manager.rb
```diff
@@ -152,6 +142,21 @@ def install
...
   private
 
+    def initialize
+      Yast::Mode.SetUI("commandline")
+      Yast::Mode.SetMode("installation")
+      @languages = []
+      @products = []
+      @status_callbacks = []
+      @status = InstallerStatus::ERROR # it should start with probing, so just temporary status
+      @logger = logger || Logger.new($stdout)
```
jreidinger, line 152: this looks wrong. logger will be always nil when initialized. We need to call instance.tap and set logger or maybe consider having singleton Logger object that we init very early and can modify it during runtime? What do you think?

service/lib/dinstaller/storage/commit_actions.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def main_actions
+        actions = actions.reject { |a| a.device_is?(:btrfs_subvolume) }
+        sort_actions(actions)
+      end
+
+      def subvolume_actions
+        actions = actions.select { |a| a.device_is?(:btrfs_subvolume) }
+        sort_actions(actions)
+      end
```
jreidinger, line 43: regarding that signal. Can we get some observer that recognize if actions changes?

service/lib/dinstaller/storage/guided_proposal.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    private
+
+      def generate_proposal_settings(settings)
+        proposal_settings = Y2Storage::ProposalSettings.new_for_current_product
+
+        proposal_settings.use_lvm = settings["LVM"] if settings.keys.include?("LVM")
+
+        if settings.keys.include?("CandidateDevices")
+          proposal_settings.candidate_devices = settings["CandidateDevices"]
```
jreidinger, line 73: It is probably intention to have here settings based on DBus call convention (that allows that easy signal that just return values that changed), but I do not like much this coupling of DBus property names and backend hash keys.

service/d-installer.gemspec
joseivanlopez: Maybe we need a development enviroment or something similar, because using the byebug package from the system does not work. AFAIS, *bundle exec* only uses gems from the gemspec.

service/lib/dinstaller/dbus/storage/commit_actions.rb
joseivanlopez: Yes, this is a WIP API. I was thinking about it at night and I am going to change it. I was also thinking about a callbacks/observer mechanish. Let's try it.

service/lib/dinstaller/dbus/storage/guided_proposal.rb
joseivanlopez: Yes, there is a direct mapping. I mean, *settings* should be real properties. I guess we will need to implement some validations, but I think it is not necessary now.

service/lib/dinstaller/manager.rb
joseivanlopez: I am not sure. This is something we need to discuss.

service/lib/dinstaller/manager.rb
joseivanlopez: TBH, I haven't take care about the logger yet. Anyway, I haven't touched this code, ony moved it.

service/lib/dinstaller/storage/commit_actions.rb
joseivanlopez: Yes, I will implement some observer/callbacks system.

service/lib/dinstaller/storage/guided_proposal.rb
joseivanlopez: Well, *settings* is the value of a parameter from a dbus method, so I expect it uses dbus properties names. We need some conversion at some point.

joseivanlopez:
~~~ruby
class Actions < DBus::Object
  def initialize
    Proposal.callbacks.after(:Calculate) do
      # do something, for example, emit a signal
      PropertiesChange(INTERFACE, { "Actions" => actions }, [])
    end

    dbus_interface INTERFACE do
      dbus_reader :actions
    end
  end 
end

class Proposal < DBus::Object
  def callbacks
    @callbacks ||= CallbacksManager.new
  end

  dbus_interface "name" do
    dbus_method :Calculate do
      # calculate ...
      callbacks.after(:Calculate).call
    end
  end
end
~~~

service/lib/dinstaller/storage/commit_actions.rb
joseivanlopez: After discussing it, we decided to not implement observers yet. Let's discuss first how to deal with threads.

service/lib/dinstaller/storage/guided_proposal.rb
jreidinger: yeap, I just this that such conversion should be done between in dbus and not in backend as backend should be dbus agnostic. but just my opinion

service/lib/dinstaller/dbus/storage/actions.rb
```diff
@@ -0,0 +1,65 @@
...
+        private_constant :INTERFACE
+
+        def initialize(logger)
+          @logger = logger
+
+          super(PATH)
+        end
+
+        dbus_interface INTERFACE do
+          # TODO: emit a signal when actions change (e.g., when a proposal is calculated). Right
```
jreidinger, line 43: another easy way without subscribing is to pass this dbus object to Proposal and call signal emit there.

service/lib/dinstaller/storage/guided_proposal.rb
joseivanlopez: Ah, good point!

service/lib/dinstaller/dbus/storage/actions.rb
joseivanlopez: Shouldn't it be similar to implementing the observer pattern? Anyway, I would do nothing for now. As we agreed, we need to take a decision about the general approach to follow.

service/lib/dinstaller/dbus/service.rb
```diff
@@ -103,12 +103,6 @@ def storage_proposal_dbus
       def storage_actions_dbus
         @storage_actions_dbus ||= DInstaller::DBus::Storage::Actions.new(@logger)
       end
-
-      def installer
-        # TODO: this god object should not be needed anymore when all dbus API is adapted
-        # just that probe should  be kept to get installer probed ASAP
-        @installer ||= DInstaller::Manager.instance.tap(&:probe)
```
jreidinger, line 94: yeap, I overlook method when I do merge of upstream where imo remove its usage from language. And it is also reason why probe is not called. We found it during coding session for adapting software in frontend and it is part of https://github.com/yast/d-installer/pull/49

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -52,11 +52,11 @@ def initialize(logger)
       end
 
       dbus_interface MANAGER_INTERFACE do
-        dbus_method :probe, "" do
+        dbus_method :Probe, "" do
           @backend.probe
         end
 
-        dbus_method :commit, "" do
+        dbus_method :Commit, "" do
```
jreidinger, line 59: good work :+1:

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -73,11 +73,7 @@ def calculate(settings = {})
       def generate_proposal_settings(settings)
         proposal_settings = Y2Storage::ProposalSettings.new_for_current_product
 
-        proposal_settings.use_lvm = settings["LVM"] if settings.keys.include?("LVM")
-
-        if settings.keys.include?("CandidateDevices")
-          proposal_settings.candidate_devices = settings["CandidateDevices"]
-        end
+        settings.each { |k, v| proposal_settings.send("#{k}=", v) }
```
jreidinger, line 76: just NP, I would use `public_send`

jreidinger (approved):
so far looks good for me.

PR #53, opened 2022-03-08
Add a class to handle network settings

imobachgs:


service/lib/dinstaller/manager.rb
```diff
@@ -141,6 +142,10 @@ def software
       @software ||= Software.instance.tap { |s| s.logger = @logger }
     end
 
+    def network
+      @network ||= Network.instance.tap { |s| s.logger = @logger }
```
jreidinger, line 146: as said also in @joseivanlopez PR. I think we should consider global logger. This is quite annoying to repeat it and also it would be nice to have it automatic.

service/lib/dinstaller/network.rb
```diff
@@ -0,0 +1,50 @@
...
+    #
+    # @param _progress [Progress] Progress reporting class
+    def probe(_progress)
+      logger.info "Probing network"
+      Yast.import "Lan"
+      Yast::Lan.read_config
+      settings = Y2Network::ProposalSettings.instance
+      settings.refresh_packages
+      settings.apply_defaults
+    end
```
jreidinger, line 41: logger accessor missing here.

imobachgs: Good point.

service/lib/dinstaller/manager.rb
imobachgs: Actually, we could use a mixing for that.

jreidinger (approved):

PR #54, opened 2022-03-08
Replace the Singleton pattern with Dependency Injection

imobachgs:
This PR tries to address two things that I consider design problems. Of course, I might be wrong so, please, share your POV too.

* We are abusing the singleton pattern. For instance, all D-Bus object classes are accessing to the corresponding model by using a singleton instance. It looks little obscure to me. Would not be better to directly inject the dependency?
* The D-Bus interface is guiding the whole application. I would rather prefer to decouple our business logic (`DInstaller::Manager`, `DInstaller::Software`, and friends) from the D-Bus layer. I think that building the `DInstaller::Manager` **outside** and passing the instance to the D-Bus service to build the interface on top is a cleaner approach.

More changes introduced by this PR:

* Class `InstallationStatus` is replaced by `StatusManager`. This new class takes care of managing status and callbacks when the status is changed.
* Error messages are now provided by a new D-Bus property from *Manager1* instead of having it as a progress message.
* Added contributing file.


joseivanlopez (approved):
I totally agree. I don't like how we are using singleton neither. From my point of view, dbus objects should only take care of definining the DBus API. The DBus API implementation should be only a matter of calling a backend API, nothing else. Somehow we were already doing that with the sigleton objects. But injecting the backend object looks a more flexible approach. I vote to go in that direction.

jreidinger:
In general it looks good. Only small part I do not like is that Manager start to be god object that has to know about everything that can limit reusability of other parts. But probably not so big issue. With singleton approach I feel manager is more on level with other backend classes.

imobachgs:
Yes, if we are concerned about the Manager, we can extract other classes initialization.

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -20,77 +20,100 @@
...
+      # @see DInstaller::Manager#status
       def status
-        @backend.status.id
+        backend.status.id
       end
 
+      # @see DInstaller::Manager#progress
       def progress
-        prg = @backend.progress
+        progress = backend.progress
```
jreidinger, line 81: well, reason I use prg is that it was too much progress for me. progress as method name, backend method and also variable names feel too confusing for me.

joseivanlopez: Ok, I can revert it if you feel more confortable with `prg`. In general, I prefer to use complete works to make the code more readable, but it is true that threre is too many `progress` there.

service/lib/dinstaller/status.rb
```diff
@@ -19,20 +19,47 @@
...
+    #
+    # Callbacks are called.
+    #
+    # @param id [Integer] new status
+    def change(id)
+      @id = id
+      on_change_callbacks.each(&:call)
     end
 
-    ERROR = new(0)
```
ancorgs, line 32: I like this old approach more (in which `InstallerStatus::ERROR` returns an status object) than the correct one (in which `Status::ERROR` returns a number).

service/lib/dinstaller/dbus/manager.rb
jreidinger: or use prefix like `backend_progress` or `progress_o` to emphasize that it is progress object.

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -90,6 +101,32 @@ def progress
...
+      # @return [Logger]
+      attr_reader :logger
+
+      # @return [DInstaller::Manager]
+      attr_reader :backend
+
+      # Adds callback to be called when the status changes
+      #
+      # The callback will emit a signal
+      def add_status_callback
```
jreidinger, line 116: well, in general I am not sure about names. I would use something like `register_status_callback` as this method sounds like you add callback to this method and not registering if to backend.

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -20,68 +20,79 @@
...
         @logger = logger
-        @backend = ::DInstaller::Manager.instance.tap(&:probe)
-        @backend.progress.add_on_change_callback do
-          PropertiesChanged(MANAGER_INTERFACE, { "Progress" => progress }, [])
-        end
-        @backend.add_status_callback do
-          PropertiesChanged(MANAGER_INTERFACE, { "Status" => @backend.status.id }, [])
-        end
+        @backend = backend
+
```
imobachgs, line 41: np: ETOOMANYEMPTYLINES :-)

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -20,68 +20,79 @@
...
 
-        # Enum list for statuses. Possible values:
-        # 0 : error ( it can be read from progress message )
-        # 1 : probing
-        # 2 : probed
-        # 3 : installing
-        # 4 : installed
+        # Current status
+        #
+        # Possible values:
```
imobachgs, line 56: What about pointing to the `Status` class? I fear that these values might get outdated sooner than later.

service/lib/dinstaller/dbus/storage/proposal.rb
```diff
@@ -33,7 +32,15 @@ class Proposal < ::DBus::Object
...
 
-        def initialize(logger, actions)
+        # Constructor
+        #
+        # @param backend [DInstaller::Storage::Proposal]
+        # @param logger [Logger]
+        # @param actions [DInstaller::DBus::Storage::Actions] D-Bus object representing the storage
+        #   actions to perform in the system. It is needed to raise signals when a new proposal is
+        #   calculated.
+        def initialize(backend, logger, actions)
```
imobachgs, line 42: It would look better for me to inject the dependencies (`backend` and `actions`), and the logger at the end. Or we could even consider moving logger to a named argument. But it is just a matter of opinion, feel free to ignore this comment.

service/lib/dinstaller/dbus/storage/proposal.rb
```diff
@@ -46,47 +53,64 @@ def initialize(logger, actions)
...
+
           # result: 0 success; 1 error
           dbus_method :Calculate, "in settings:a{sv}, out result:u" do |settings|
             backend.calculate(to_proposal_properties(settings))
 
             PropertiesChanged(INTERFACE, settings, [])
-            @actions.refresh
+            actions.refresh
 
             backend.success? ? 0 : 1
```
imobachgs, line 64: Not related to this PR, but I wonder whether it might be better that `backend.calculate` already returns the success value.

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -20,68 +20,79 @@
...
+        backend.status_manager.status.id
+      end
+
+      # Error messages from the error status
+      #
+      # @return [Array<String>]
+      def error_messages
+        return [] unless backend.status_manager.error?
+
+        backend.status_manager.status.messages
```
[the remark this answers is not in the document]
joseivanlopez: Right now it only includes errors from the Error status. And Error status is storing only the error message from the lastest exception. But in the API I have considered the possibility of returning more than one error message, just in case we need to start reporting serveral errors.

service/lib/dinstaller/manager.rb
```diff
@@ -19,75 +19,63 @@
...
-    # * Network configuration
-    #
-    # The initialization of these subsystems should probably live in a different place.
+    # @param logger [Logger]
+    def initialize(logger)
+      @logger = logger
+      @status_manager = StatusManager.new(Status::Error.new) # temporary status until probing starts
+      @progress = Progress.new
+
+      initialize_yast
```
imobachgs, line 60: 👍🏻

service/lib/dinstaller/status_manager.rb
```diff
@@ -0,0 +1,132 @@
...
+module DInstaller
+  # Manager for installation status
+  #
+  # It allows to change the installation status and to execute callbacks when the status changes.
+  #
+  # @example
+  #   status_manager = StatusManager.new(Status::Error.new)
+  #   status_manager.on_change { puts "changing status" }
+  #   status_manager.change(Status::Installing.new)
+  class StatusManager
```
imobachgs, line 31: 👍🏻

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -28,32 +27,59 @@ module DInstaller
...
       def success?
         raise NoProposalError unless proposal
 
         !proposal.failed?
       end
 
+      # Calculates a new proposal
+      #
+      # @param settings [Hash] settings to calculate the proposal
+      #   (e.g., { "LVM" => true, "CandidateDevices" => ["/dev/sda"]})
```
imobachgs, line 82: I wonder if we should use downcase keys and keep the conversation from D-Bus style in the corresponding D-Bus class. However, it is something we could do later (when we start writing RSpec tests).

imobachgs (commented):
In general, it looks good. Just a few comments that, most of them, can be ignored :-)

service/lib/dinstaller/storage/proposal.rb
joseivanlopez: Hehe, that is exactly what it does! Sorry, documentation is wrong.

service/lib/dinstaller/dbus/manager.rb
joseivanlopez: Yes, this is something I was thinking about. For now, I decided to keep it as it was (using the id). I am going to add a TODO note.

service/lib/dinstaller/dbus/storage/proposal.rb
joseivanlopez: Yes, using a named param is something I commented with @ancorgs. But actually we are not using name params in other places yet. So, again, I keep doing the same.

service/lib/dinstaller/dbus/storage/proposal.rb
joseivanlopez: Ummm, I am not sure. I prefered to decide in the D-Bus class what value to return in case of success or failure.

imobachgs: That's fine. What does not look that fine is to ask for calculation and get the success/failure value in a separate method.

service/lib/dinstaller/dbus/storage/proposal.rb
imobachgs: The named param is something optional, my main point was about the arguments order.

PR #57, opened 2022-03-08
Adapt storage UI

imobachgs:
Adapt the Storage components to the new Storage D-Bus API.

### DIFF
--- a/web/src/Proposal.jsx
+++ b/web/src/Proposal.jsx
@@ -1,26 +1,15 @@
 import React from "react";
 import {
-  TableComposable,
-  Thead,
-  Tr,
-  Th,
-  Tbody,
-  Td
-} from "@patternfly/react-table";
-
-import filesize from "filesize";
+  List,
+  ListItem
+} from "@patternfly/react-core";
 
 const Proposal = ({ data = [] }) => {
   // FIXME: use better key for tr, mount can be empty
-  const renderProposal = () => {
-    return data.map(p => {
+  const renderActions = () => {
+    return data.map((p, i) => {
       return (
-        <Tr key={p.mount}>
-          <Td dataLabel="Mount Point">{p.mount}</Td>
-          <Td dataLabel="Type">{p.type}</Td>
-          <Td dataLabel="Device">{p.device}</Td>
-          <Td dataLabel="Size">{filesize(p.size)}</Td>
-        </Tr>
+        <ListItem key={i}>{p.text}</ListItem>
       );
     });
   };
@@ -30,17 +19,7 @@ const Proposal = ({ data = [] }) => {
   }
 
   return (
-    <TableComposable variant="compact">
-      <Thead>
-        <Tr>
-          <Th>Mount point</Th>
-          <Th>Type</Th>
-          <Th>Device</Th>
-          <Th>Size</Th>
-        </Tr>
-      </Thead>
-      <Tbody>{renderProposal()}</Tbody>
-    </TableComposable>
+    <List>{renderActions()}</List>
   );
 };
 
--- a/web/src/Storage.jsx
+++ b/web/src/Storage.jsx
@@ -17,7 +17,7 @@ const reducer = (state, action) => {
       return { ...state, target: action.payload, error: false };
     }
 
-    case "UPDATE_PROPOSAL": {
+    case "UPDATE_ACTIONS": {
       return { ...state, actions: action.payload };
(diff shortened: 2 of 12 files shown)


web/src/Proposal.jsx
```diff
@@ -1,26 +1,15 @@
 import React from "react";
```
dgdavid, line 1: NP: I'd rename this component to ProposalActions or similar.

web/src/Proposal.jsx
```diff
@@ -1,26 +1,15 @@
...
 import {
-  TableComposable,
-  Thead,
-  Tr,
-  Th,
-  Tbody,
-  Td
-} from "@patternfly/react-table";
-
-import filesize from "filesize";
```
dgdavid, line 11: Remember to remove the _filesize_ dependency, no longer needed.

web/src/Storage.jsx
```diff
@@ -31,47 +33,57 @@ export default function Storage() {
...
-      const [_, changes] = args;
-      if (Object.keys(changes).includes("Disk")) {
-        client
-          .getStorage()
-          .then(proposal =>
-            dispatch({ type: "UPDATE_PROPOSAL", payload: proposal })
-          );
+      const [iface, properties] = args;
+
+      if (iface !== "org.opensuse.DInstaller.Storage.Actions1") {
```
dgdavid, line 63: Maybe we could have a JS module maping ifaces and exporting them as constants. I mean, a _ifaces.js_ file or so

```js
const ACTIONS = "org.opensuse.DInstaller.Storage.Actions1";
const  WHATERVER = "org.opensuse.DInstaller.Storage.Actions1";

...

export {
  ACTIONS,
  WHATEVER,
  ...
};
```

Then, 

```js

import { ACTIONS as ACTIONS_IFACE } from 'ifaces';

...

if (iface !== ACTIONS_IFACE) {
 
...

```

web/src/Storage.jsx
```diff
@@ -31,47 +33,57 @@ export default function Storage() {
...
+        return;
       }
+
+      const newActions = properties.All.v.map(action => {
+        const { Text: textVar, Subvol: subvolVar } = action.v;
+        return { text: textVar.v, subvol: subvolVar.v };
+      });
```
dgdavid, line 70: I'd move that mapping to an util function.

web/src/lib/InstallerClient.test.js
```diff
@@ -10,16 +10,14 @@ const DBUS_PATH = "/org/opensuse/YaST/Installer";
 const DBUS_IFACE = "org.opensuse.YaST.Installer";
 const LANGUAGE_IFACE = "org.opensuse.DInstaller.Language1";
 const SOFTWARE_IFACE = "org.opensuse.DInstaller.Software1";
+const STORAGE_PROPOSAL_IFACE = "org.opensuse.DInstaller.Storage.Proposal1";
+const STORAGE_ACTIONS_IFACE = "org.opensuse.DInstaller.Storage.Actions1";
```
dgdavid, line 14: They could be in the proposed _ifaces.js_ module.

dgdavid (approved):
Just minor suggestions, but looks good.

joseivanlopez:
LGTM for the backend part. For now, I think it is enough. Let's see if we need something more complex in the future.

web/src/Storage.jsx
imobachgs: I prefer to wait until we improve `InstallerClient`. In the component, you shouldn't need to deal with D-Bus variants.

web/src/Storage.jsx
imobachgs: The idea is that this callback does not expose D-Bus internals (so you should not get the interface name and so on). Let's wait until we refactor `InstallerClient`.

web/src/Proposal.jsx
imobachgs: Good point.

web/src/Proposal.jsx
imobachgs: Yes. And I would change the wording about storage in general: e.g., replacing "Target" with "Partitioning" or something like that. Let's do that in an open PR.

web/src/lib/InstallerClient.test.js
imobachgs: Something for the `InstallerClient` refactoring.

PR #58, opened 2022-03-09
Fix linting errors and code formatting

imobachgs:
Make ESLint and Prettier happy. Actually, ESLint caught a potential error using the wrong `cockpit.variant`. It was not a problem because `cockpit` lives in `window`.


dgdavid (approved):
LGTM :see_no_evil:

web/src/InstallationProgress.jsx
```diff
@@ -51,31 +42,25 @@ function InstallationProgress() {
...
+  };
 
   // FIXME: this is an example. Update or drop it.
   const Actions = () => {
     return (
-      <Button
-        isDisabled
-        onClick={() => console.log("User want to see the summary!") }
-      >
+      <Button isDisabled onClick={() => console.log("User want to see the summary!")}>
```
dgdavid, line 59: BTW, this is a leftover… but it can be dropped when removing all _layout examples_.

PR #60, opened 2022-03-10
Refactor client

imobachgs:
Refactor the client into smaller classes.

```javascript
import InstallerClient from "./lib/client";
client = new InstallerClient();
client.manager.GetStatus();


// or even...
const { manager } = new InstallerClient();
manager.GetStatus();
```

It is still a WIP (we want to get rid of `Client` class), but it is good enough by now.


dgdavid (approved):
LGTM

PR #63, opened 2022-03-10
Improve status

jreidinger:
Fixes in this PR:

- show progress when probing instead of crash
- show error message when failed to connect to dbus service

Not Fixes:

- does not automatically redraw when dbus is on, because it cannot register signal. ( maybe timer will be needed? )

### DIFF
--- a/web/src/Installer.jsx
+++ b/web/src/Installer.jsx
@@ -27,11 +27,13 @@ import InstallationProgress from "./InstallationProgress";
 
 function Installer() {
   const client = useInstallerClient();
-  const [isInstalling, setIsInstalling] = useState(false);
+  // set initial state to true to avoid async calls to dbus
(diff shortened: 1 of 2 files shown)


web/src/Installer.jsx
```diff
@@ -27,24 +27,40 @@ import InstallationProgress from "./InstallationProgress";
...
+
 
   useEffect(async () => {
-    const status = await client.manager.getStatus();
-    setIsInstalling(status === 3);
+    try {
+      const status = await client.manager.getStatus();
+      setIsProgress(status === 3 || status == 1);
+    } catch (err) {
+      console.log("Error");
```
imobachgs, line 41: You could use "console.error" instead.

imobachgs (approved):

dgdavid (approved):

PR #64, opened 2022-03-10
use single thread backend service

jreidinger:
only way how I can achieve some result when using single thread backend.


jreidinger:
already done as part of users PR.

PR #65, opened 2022-03-11
Add an API to listen for changes

imobachgs:
This PR improves the overall installer clients approach by

* making use of [mixins](https://javascript.info/mixins)
* dropping the `InstallerClient` in favor of `createClient` [factory function](https://www.javascripttutorial.net/javascript-factory-functions/)
* **adding an API for listening changes in different modules**
  
  * _StorageClient#onActionsChange_, for reacting to storage proposal actions changes.
  * _ManagerClient#onChange_, for listening to installation progress changes.

  Both make use of the `onObjectChanged` _withDBus_ mixin method, which allows registering a method to be triggered when properties change for a given D-Bus path hiding the details of calling the `onPropertyChanged` cockpit method.

Additionally, the PR also improves the way of dealing with the installation progress status by adding a specific module for those statuses.

### DIFF
--- a/web/src/InstallationProgress.jsx
+++ b/web/src/InstallationProgress.jsx
@@ -34,11 +34,9 @@ function InstallationProgress() {
   const [progress, setProgress] = useState({});
 
   useEffect(() => {
-    return client.onPropertyChanged((_path, _iface, signal, args) => {
-      const iface = "org.opensuse.DInstaller.Manager1";
-      const [input_iface, changed] = args;
-      if (input_iface === iface && "Progress" in changed) {
-        const [title, steps, step, substeps, substep] = changed.Progress.v.map(pr => pr.v);
+    return client.manager.onChange(changes => {
+      if ("Progress" in changes) {
+        const [title, steps, step, substeps, substep] = changes.Progress;
         setProgress({ title, steps, step, substeps, substep });
       }
     });
--- a/web/src/Storage.jsx
+++ b/web/src/Storage.jsx
@@ -57,7 +57,7 @@ export default function Storage() {
   }, []);
 
   useEffect(() => {
-    return client.storage.onActionsChanged(changes => {
+    return client.storage.onActionsChange(changes => {
       const { All: newActions } = changes;
       dispatch({ type: "UPDATE_ACTIONS", payload: newActions });
     });
--- a/web/src/Storage.jsx
+++ b/web/src/Storage.jsx
@@ -57,19 +57,8 @@ export default function Storage() {
   }, []);
 
   useEffect(() => {
-    // TODO: abstract D-Bus details
-    return client.onPropertyChanged((_path, _iface, _signal, args) => {
-      const [iface, properties] = args;
-
-      if (iface !== "org.opensuse.DInstaller.Storage.Actions1") {
-        return;
-      }
-
-      const newActions = properties.All.v.map(action => {
-        const { Text: textVar, Subvol: subvolVar } = action.v;
-        return { text: textVar.v, subvol: subvolVar.v };
-      });
-
+    return client.storage.onActionsChanged(changes => {
+      const { All: newActions } = changes;
       dispatch({ type: "UPDATE_ACTIONS", payload: newActions });
     });
   }, []);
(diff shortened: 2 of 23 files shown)


web/src/InstallationProgress.jsx
```diff
@@ -34,11 +34,9 @@ function InstallationProgress() {
   const [progress, setProgress] = useState({});
 
   useEffect(() => {
-    return client.onPropertyChanged((_path, _iface, signal, args) => {
-      const iface = "org.opensuse.DInstaller.Manager1";
-      const [input_iface, changed] = args;
-      if (input_iface === iface && "Progress" in changed) {
-        const [title, steps, step, substeps, substep] = changed.Progress.v.map(pr => pr.v);
+    return client.manager.onChange(changes => {
```
jreidinger, line 37: maybe I would go one step further and have method `onProgress` that just get that progress as param?

web/src/Storage.jsx
```diff
@@ -57,19 +57,8 @@ export default function Storage() {
...
-      if (iface !== "org.opensuse.DInstaller.Storage.Actions1") {
-        return;
-      }
-
-      const newActions = properties.All.v.map(action => {
-        const { Text: textVar, Subvol: subvolVar } = action.v;
-        return { text: textVar.v, subvol: subvolVar.v };
-      });
-
+    return client.storage.onActionsChange(changes => {
```
jreidinger, line 60: so is it changes here or directly new actions?

web/src/lib/client/manager.js
```diff
@@ -46,4 +51,27 @@ export default class ManagerClient extends Client {
     const proxy = await this.proxy(MANAGER_IFACE);
     return proxy.Status;
   }
+
+  /**
+   * Register a callback to run when properties in the Actions object change
```
jreidinger, line 56: I think it is not about actions....and I would probably split it to two methods.

web/src/lib/client/mixins.js
```diff
@@ -0,0 +1,94 @@
...
+  },
+
+  /**
+   * Register a callback to run when some D-Bus signal is emitted
+   *
+   * @param {function} handler - callback function
+   */
+  onSignal(signal, handler) {
+    const { remove } = this._client.subscribe(
+      { interface: "org.opensuse.YaST.Installer", member: signal },
```
jreidinger, line 79: this looks like old interface

web/src/lib/client/storage.js
```diff
@@ -59,4 +64,21 @@ export default class StorageClient extends Client {
...
+   *
+   * @param {function} handler - callback function
+   */
+  onActionsChange(handler) {
+    return this.onObjectChanged(ACTIONS_PATH, changes => {
+      const newActions = changes.All.v.map(action => {
+        const { Text: textVar, Subvol: subvolVar } = action.v;
+        return { text: textVar.v, subvol: subvolVar.v };
+      });
+      handler({ All: newActions });
```
jreidinger, line 79: why not pass just new actions as parameter?

web/src/InstallationProgress.jsx
imobachgs: It might be possible, I am not against that. But we might end up with one `onXXX` method for each property. Alternatively, we could have an argument to filter by property name. But as it is rather straight-forward, we decided to keep it as it is until the API evolves a little bit (based on our use cases).

web/src/Storage.jsx
imobachgs: These are the changes from the `Actions` path.

web/src/lib/client/mixins.js
imobachgs: Yes, you are right. We just moved this method from the old API. I have fixed it now, although it is still not used.

web/src/lib/client/storage.js
imobachgs: If we add a new property in the future, we would need to update all the handlers to cope with the change. Passing the actions under the `All` property makes this code future-proof.

web/src/InstallationProgress.jsx
imobachgs: Actually, we could do both cases: adding the "name" filtering and defining `onProgress` on top of that.

jreidinger (approved):

PR #66, opened 2022-03-14
[web] Few UI improvements

dgdavid:
This is an _intermediate_ PR that introduces few UI improvements. _Intermediate_ because it pushed things that should be improved later (like example code).

Those improvements are

* A component to inform about a DBus connection problem
  <details>

  ---

  | Before | After |
  |-|-|
  | Screenshot 2022-03-14 at 10-04-29 D-Installer | Screen Shot 2022-03-14 at 12 41 38-fullpage |

  </details>

  It can be extracted to a more general _ApplicationError_ component based on props. But let's wait until we have another blocking error and think about the real action to offer ("retry", "go back, etc)

* Extract the UI logic for centered the content to a `Center` component.
* Slightly improves the substep progress
  <details>

  ---

  | Before | After |
  |-|-|
  | Screenshot 2022-03-14 at 10-19-29 D-Installer | Screenshot 2022-03-14 at 11-02-05 D-Installer |
  </details>
  This improvement is arguably, of course. But it's a bit weird having a percentage without a description. In the near future, we could add information about the substep and still improving that screen.

* Add screen for _wiring up the UI_

  It helps to avoid flickering in the login screen, but the whole UI deserves more love to avoid flickering everywhere as much as possible.

  <details>
  <summary>Click to show/hide screenshot</summary>

  ---
 
  
  Screen Shot 2022-03-14 at 12 42 48-fullpage

  </details>

  Note that, intentionally, the initial value for _isLoggedIn_ has been initializated to `null`. It is something to re-think in next iterations, enriching the status or whatever other option.

### DIFF
--- a/web/src/App.test.jsx
+++ b/web/src/App.test.jsx
@@ -49,9 +49,11 @@ describe("when username and password are wrong", () => {
 
   it("shows an error", async () => {
     authRender(<App />);
+    const button = await screen.findByRole("button", { name: /Login/ });
+
     userEvent.type(screen.getByLabelText(/Username/i), "john");
     userEvent.type(screen.getByLabelText(/Password/i), "something");
-    userEvent.click(screen.getByRole("button", { name: /Login/ }));
+    userEvent.click(button);
     await screen.findByText(/Authentication failed/i);
   });
 });
--- a/web/src/App.test.jsx
+++ b/web/src/App.test.jsx
@@ -49,11 +49,13 @@ describe("when username and password are wrong", () => {
 
   it("shows an error", async () => {
     authRender(<App />);
-    const button = await screen.findByRole("button", { name: /Login/ });
+
+    await screen.findByText(/Username/i);
 
     userEvent.type(screen.getByLabelText(/Username/i), "john");
     userEvent.type(screen.getByLabelText(/Password/i), "something");
-    userEvent.click(button);
+    userEvent.click(screen.getByRole("button", { name: /Login/ }));
+
     await screen.findByText(/Authentication failed/i);
   });
 });
--- a/web/src/DBusError.jsx
+++ b/web/src/DBusError.jsx
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) [2022] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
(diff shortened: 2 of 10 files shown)


web/src/DBusError.jsx
```diff
@@ -0,0 +1,59 @@
...
+);
+
+function DBusError() {
+  return (
+    <Layout sectionTitle="DBus Error" SectionIcon={SectionIcon} FooterActions={ReloadAction}>
+      <Center>
+        <EmptyState>
+          <EmptyStateIcon icon={DisconnectionIcon} />
+          <Title headingLevel="h4" size="lg">
+            Cannot connect to DBus
```
imobachgs, line 47: Please, use *D-Bus* instead of *DBus*.

web/src/DBusError.jsx
```diff
@@ -0,0 +1,59 @@
...
+          </Title>
+          <EmptyStateBody>
+            D-Installer UI cannot connect with D-Installer DBus service. Please check if it is
```
imobachgs, line 50: I would write "Could not connect to the D-Bus service. Please, check whether it is running."

You could even include the name of the service, so the user knows what to search for.

web/src/InstallationProgress.jsx
```diff
@@ -49,6 +50,9 @@ function InstallationProgress() {
 
   // FIXME: this is an example. Update or drop it.
   const Messages = () => {
+    if (status !== 3)
```
imobachgs, line 53: Now that we merge #65, you could use the constants we defined for statuses.

web/src/LoadingEnvironment.jsx
```diff
@@ -0,0 +1,47 @@
...
+} from "eos-icons-react";
+
+function LoadingEnvironment() {
+  return (
+    <Layout sectionTitle="D-Installer">
+      <Center>
+        <EmptyState>
+          <EmptyStateIcon icon={LoadingIcon} />
+          <Title headingLevel="h4" size="lg">
+            Loading environment, please wait.
```
imobachgs, line 39: What about *installation environment* instead of just *environment*? Just a suggestion...

web/src/context/auth.jsx
```diff
@@ -43,7 +43,7 @@ function authReducer(state = initialState, action) {
   }
 }
 
-const initialState = { loggedIn: false };
+const initialState = { loggedIn: null };
```
imobachgs, line 46: Yes, you do not know whether you are logged in or not. Good point.

web/src/App.test.jsx
```diff
@@ -49,9 +49,11 @@ describe("when username and password are wrong", () => {
 
   it("shows an error", async () => {
     authRender(<App />);
+    const button = await screen.findByRole("button", { name: /Login/ });
+
     userEvent.type(screen.getByLabelText(/Username/i), "john");
     userEvent.type(screen.getByLabelText(/Password/i), "something");
-    userEvent.click(screen.getByRole("button", { name: /Login/ }));
+    userEvent.click(button);
```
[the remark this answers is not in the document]
dgdavid: I'll check if it works. I actually used it as first line in order to _wait_ for the login screen.

web/src/InstallationProgress.jsx
dgdavid: Right!

web/src/App.test.jsx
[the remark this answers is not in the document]
dgdavid: I changed the approach

imobachgs (approved):

PR #68, opened 2022-03-15
Add screen/component to be displayed when installation finishes

dgdavid:
As part of the `InstallationProgess` by now, this PR adds a new component for displaying feedback to the user when the installation finishes.

The example summary has been borrowed from the [openSUSE control file](https://github.com/yast/skelcd-control-openSUSE/blob/master/control/control.openSUSE.xml#L580-L584), but of course it's just an example.

About the `InstallationProgress` component itself, it has been restructured a bit as a consequence of minor issues found when adding a quite basic automated test.

---

Screen Shot 2022-03-14 at 16 29 31-fullpage

---

Related trello card (internal link): https://trello.com/c/7frozuK2

### DIFF
--- a/web/src/InstallationFinished.jsx
+++ b/web/src/InstallationFinished.jsx
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) [2022] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import {
+  Button,
+  Title,
+  Text,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody,
+  EmptyStateSecondaryActions
+} from "@patternfly/react-core";
+
+import Layout from "./Layout";
+import Center from "./Center";
+
+import {
+  EOS_TASK_ALT as InstallationFinishedIcon,
+  EOS_CHECK_CIRCLE as SectionIcon
+} from "eos-icons-react";
+
+const FinishAction = () => (
+  <Button
+    isLarge
+    variant="primary"
+    onClick={() => console.log("FIXME: trigger a reboot machine action here!")}
+  >
+    Finish
+  </Button>
+);
+
+function InstallationFinished() {
+  return (
+    <Layout
+      sectionTitle="Installation Finished"
+      SectionIcon={SectionIcon}
+      FooterActions={FinishAction}
+    >
+      <Center>
+        <EmptyState>
+          <EmptyStateIcon icon={InstallationFinishedIcon} className="success-icon" />
+          <Title headingLevel="h2" size="4xl">
+            Congratulations!
+          </Title>
+          <EmptyStateBody className="pf-c-content">
+            <div>
+              <Text>The installation on your machine is complete.</Text>
+              <Text>After clicking on "Finish" you can log in to the sytem.</Text>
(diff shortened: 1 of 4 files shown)


web/src/InstallationFinished.jsx
```diff
@@ -0,0 +1,88 @@
...
+    <Button isLarge variant="primary" onClick={onFinish}>
+      Finish
+    </Button>
+  </>
+);
+
+function InstallationFinished() {
+  const client = useInstallerClient();
+  const onRestartAction = () => client.manager.startProbing();
+  const onFinishAction = () => console.log("FIXME: trigger a reboot machine action here!");
```
jreidinger, line 56: probably should be manager action on dbus?

web/src/Installer.jsx
```diff
@@ -19,43 +19,71 @@
...
 import DBusError from "./DBusError";
 import Overview from "./Overview";
+import ProbingProgress from "./ProbingProgress";
 import InstallationProgress from "./InstallationProgress";
+import InstallationFinished from "./InstallationFinished";
+import LoadingEnvironment from "./LoadingEnvironment";
+
+import statuses from "./lib/client/statuses";
+
+const { PROBING, PROBED, INSTALLING, INSTALLED } = statuses;
```
jreidinger, line 34: one more status is error.

web/src/InstallationFinished.jsx
dgdavid: Sure, but unless I'm wrong it is not available yet. That's why I put an example there. We can delete if by now if you want.

web/src/Installer.test.jsx
```diff
@@ -0,0 +1,120 @@
+import React from "react";
+
+import { act, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { authRender } from "./test-utils";
+import { createClient } from "./lib/client";
+import { PROBING, PROBED, INSTALLING, INSTALLED } from "./lib/client/statuses";
```
dgdavid, line 7: BTW, it's not clear to me if "status" is a countable or uncountable noun.

Uncountable according to

* https://www.collinsdictionary.com/dictionary/english/status, 5th entry
* https://www.macmillandictionary.com/dictionary/american/status, 3rd entry
* https://dictionary.cambridge.org/dictionary/english/status#cbed-1

Countable according to 

*  https://www.wordreference.com/definition/statuses

Both according to 

* https://english.stackexchange.com/questions/877/what-is-the-plural-form-of-status

:man_shrugging:

shundhammer: I highly recommend "status values".

web/src/InstallationFinished.jsx
```diff
@@ -0,0 +1,88 @@
...
+  EOS_CHECK_CIRCLE as SectionIcon
+} from "eos-icons-react";
+
+const Actions = ({ onRestart, onFinish }) => (
+  <>
+    <Button isLarge variant="secondary" onClick={onRestart}>
+      Restart Installation
+    </Button>
+    <Button isLarge variant="primary" onClick={onFinish}>
+      Finish
```
imobachgs, line 48: By now, we are not implementing a Reboot action. What about just a "Close" button? It could close the tab.

web/src/InstallationFinished.jsx
imobachgs, line 56: Yes, please, let's get rid of it by now.

web/src/InstallationFinished.jsx
```diff
@@ -0,0 +1,88 @@
...
+      <Center>
+        <EmptyState>
+          <EmptyStateIcon icon={InstallationFinishedIcon} className="success-icon" />
+          <Title headingLevel="h2" size="4xl">
+            Congratulations!
+          </Title>
+          <EmptyStateBody className="pf-c-content">
+            <div>
+              <Text>The installation on your machine is complete.</Text>
+              <Text>After clicking on "Finish" you can log in to the system.</Text>
```
imobachgs, line 73: We could ask the user to "reboot" the installed system.

web/src/Installer.test.jsx
imobachgs: Or just "status", as it is the name of a module.

imobachgs (commented):
Oh, this PR improves the `Installer` component quite a lot. :-)

web/src/InstallationFinished.jsx
dgdavid: Good idea, but sadly `window.close()` does not work because

> Scripts may not close windows that were not opened by script.

I had try some dirty tricks, but none of them worked. So, I removed the button and adapted the feedback test. See the PR screenshot.

PR #71, opened 2022-03-16
Show more information about disks in the target selector

ancorgs:
Simple change to make the `<select>` on this form display something else than just the raw device name.

disk_labels

### DIFF
--- a/service/lib/dinstaller/dbus/storage/proposal.rb
+++ b/service/lib/dinstaller/dbus/storage/proposal.rb
@@ -52,7 +52,7 @@ module DInstaller
 
           dbus_reader :candidate_devices, "as"
 
-          dbus_reader :available_devices, "as"
+          dbus_reader :available_devices, "a(ssa{sv})"
 
           # result: 0 success; 1 error
           dbus_method :Calculate, "in settings:a{sv}, out result:u" do |settings|
@@ -67,7 +67,9 @@ module DInstaller
 
         # @see DInstaller::Storage::Proposal
         def available_devices
-          backend.available_devices.map(&:name)
+          backend.available_devices.map do |dev|
+            [dev.name, backend.available_device_label(dev), {}]
+          end
         end
 
         # @see DInstaller::Storage::Proposal
--- a/service/lib/dinstaller/dbus/storage/proposal.rb
+++ b/service/lib/dinstaller/dbus/storage/proposal.rb
@@ -52,6 +52,13 @@ module DInstaller
 
           dbus_reader :candidate_devices, "as"
 
+          # The first string is the name of the device (as expected by #Calculate for
+          # the setting CandidateDevices), the second one is the label to represent that device in
+          # the UI when further information is needed.
+          #
+          # TODO: this representation is a temporary solution. In the future we should likely
+          # return more complex structures, probably with an interface similar to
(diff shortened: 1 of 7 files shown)


service/lib/dinstaller/storage/proposal.rb
```diff
@@ -120,6 +136,13 @@ def disk_analyzer
         storage_manager.probed_disk_analyzer
       end
 
+      # Helper to generate a disk label
+      #
+      # @return [Helpers::Disk]
```
joseivanlopez, line 141: The return type should be `Y2Storage::Dialogs::GuidedSetup::Helpers::Disk`.

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -43,6 +44,21 @@ def available_devices
...
+      # The label has the form: "NAME, SIZE, [USB], INSTALLED_SYSTEMS".
+      #
+      # Examples:
+      #
+      #   "/dev/sda, 250.00 GiB, Windows, OpenSUSE"
+      #   "/dev/sdb, 8.00 GiB, USB"
+      #
+      # @param device [Y2Storage::Device]
+      # @return [String]
+      def available_device_label(device)
```
joseivanlopez, line 58: NP: Why not `device_label` ?

service/lib/dinstaller/dbus/storage/proposal.rb
```diff
@@ -67,7 +67,9 @@ def initialize(backend, actions, logger)
 
         # @see DInstaller::Storage::Proposal
         def available_devices
```
joseivanlopez, line 69: Please, document the return value.

joseivanlopez: BTW, what do you think about returning the label as part of the hash?

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -43,6 +44,21 @@ def available_devices
         disk_analyzer.candidate_disks
       end
 
+      # Label that should be used to represent the given disk in the UI
+      #
+      # The label has the form: "NAME, SIZE, [USB], INSTALLED_SYSTEMS".
```
jreidinger, line 49: well, to be honest this looks really unflexible that label is driven by backend. I kind of expect that backend gives information and it is up to UI to decide what and how it will be showed.

joseivanlopez: I agree. On the other hand, I believe it would be enough for this initial prototype.

ancorgs: We agreed to do it this way during the planning session. Otherwise we would have to expose a lot of details about the disks. And we wanted something quick without growing the API a lot with representation of quite complex objects.

But I can add a note to reconsider that later. It's true that at some points we will need to represent the disks with something that is closer to any of these:

- https://github.com/storaged-project/blivet/blob/3.5-devel/doc/dbus.rst
- http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Block.html

But so far we decided to go for a simple approach.

jreidinger (approved):

joseivanlopez (approved):
LGTM

PR #74, opened 2022-03-17
Monitor D-Bus

imobachgs:
If the D-Bus service goes down, the UI refreshes the interface and displays the D-Bus error page, so you know that something went wrong.


dgdavid (approved):
Nice improvement, thanks!

PR #77, opened 2022-03-18
Add users management

jreidinger:
This pull request aims to add to UI ability to configure first user and also root password. It is more to set playground so test coverage is far from perfect and also contain some rough edges, but merge to master as it is already big enough.

Some interesting changes:

- umount /run when writings users otherwise users cache is involved and cause strange effects
- remove all threads otherwise yast segfault when loading perl module in different thread
- handle exceptions in write manager method otherwise write silently failed without any notification to user

Some screenshots:

user in overview
users_overview
first user setting
users_details


jreidinger:
Note for me. Currently users finish failing to set password and also other issues that it cannot find given user tux. Reason is caches living in /mnt/run if it is umounted, everything works as expected. Maybe issue is visible because in insts-sys such login management is not available, so happens only on liveCD?

service/lib/dinstaller/users.rb
```diff
@@ -60,22 +62,44 @@ def first_user
...
+        issues = writer.write
+        logger.error(issues.inspect) unless issues.empty?
+      end
+    end
+
   private
 
     attr_reader :logger
 
+    def without_run(&block)
```
imobachgs, line 96: `without_run` sounds confusing unless you know what we are talking about. What about something `without_run_mount` or something similar?

service/bin/d-installer
```diff
@@ -32,10 +32,11 @@ require "dinstaller/dbus/service"
 
 logger = Logger.new($stdout)
 manager = DInstaller::Manager.new(logger)
-manager.probe # we can do it within the service, in a different thread, or...
 
 EM.run do
   service = DInstaller::DBus::Service.new(manager, logger)
   service.export
+  manager.probe # probeb after export, so we can report on DBus
```
imobachgs, line 39: typo: probe

service/test/dinstaller/users_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    context "if some issue occurs" do
+      let(:issues) { [double("issue")] }
+
+      it "logs the issue" do
+        expect(logger).to receive(:error).with(/issue/)
+        subject.write(progress)
+      end
+    end
+
+    it "do write without /run bind mounted" do
```
imobachgs, line 142: writes

imobachgs (approved):
Just minor comments. Please, feel free to merge the PR once you fix them.

PR #78, opened 2022-03-18
Save language settings

imobachgs:
* Save language settings.
* Restore the SCR handle, even if something went wrong.

### DIFF
--- a/service/lib/dinstaller/language.rb
+++ b/service/lib/dinstaller/language.rb
@@ -42,10 +42,15 @@ module DInstaller
       @languages = Yast::Language.GetLanguagesMap(true)
     end
 
+    def install(_progress)
+      Yast::Language.Save
+    end
+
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
LGTM

service/lib/dinstaller/language.rb
```diff
@@ -42,10 +42,15 @@ def probe(_progress)
...
 
+    def install(_progress)
+      Yast::Language.Save
+    end
+
     def language=(name)
       raise Errors::InvalidValue unless languages.include?(name)
 
       Yast::Language.Set(name)
+      Yast::Language.languages = Yast::Language.RemoveSuffix(name)
```
jreidinger, line 53: really confusing attribute of Language. I have to check it if it is not array.

jreidinger (approved):

PR #80, opened 2022-03-21
WIP: Experimenting how to make the layout more scalable

dgdavid:
This is an **incomplete** idea to make the layout more scalable, mounting it only once instead of every time a new component/screen is displayed.

It is highly inspired by https://gregberge.com/blog/react-scalable-layout but using the technique explained in https://www.sabinthedev.com/blog/using-a-react-components-function-from-its-parent instead of [React Portals](https://reactjs.org/docs/portals.html#gatsby-focus-wrapper).

  > In a React application, composition must always be preferred.
  > The layout is no exception.

At the time of creating this draft PR, only for sharing the idea, there are some issues and questions around:

* Only the SectionTitle has been adapted to use the "slot technique". Enough for discussing the topic.
* React documentation says that
  * [useRef](https://reactjs.org/docs/hooks-reference.html#useref) _doesn’t notify you when its content changes_, which I don't know now if it is a problem for this use case.
  * [useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) _should be avoided in most cases_, but maybe this is a valid use case for it. Each _slot placeholder_ is exposing its _setContent_ to the component in charge of transfer the content for it. Not exactly a _parent_ component, but it works nicely, at least in the few manual tests done.

### DIFF
--- a/web/src/DBusError.jsx
+++ b/web/src/DBusError.jsx
@@ -22,13 +22,13 @@
 import React from "react";
 import { Button, Title, EmptyState, EmptyStateIcon, EmptyStateBody } from "@patternfly/react-core";
 
-import Layout from "./Layout";
-import Center from "./Center";
-
 import {
   EOS_ANNOUNCEMENT as SectionIcon,
   EOS_ENDPOINTS_DISCONNECTED as DisconnectionIcon
 } from "eos-icons-react";
+
+import Center from "./Center";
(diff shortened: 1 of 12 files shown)


dgdavid:
Finally we have re-evaluated [reac-teleporter](https://github.com/gregberge/react-teleporter) (updated recently) and go for it. Check https://github.com/yast/d-installer/pull/216.

PR #83, opened 2022-03-22
Only offer openSUSE Tumbleweed for installation

imobachgs:
MicroOS installation does not work properly because there are several missing pieces (importing the control file, system roles support, etc.). By now, just offer to install openSUSE Tumbleweed.

### DIFF
--- a/service/lib/dinstaller/software.rb
+++ b/service/lib/dinstaller/software.rb
@@ -60,17 +60,18 @@ module DInstaller
       progress.next_minor_step("Initialize sources")
       Yast::Pkg.SourceRestore
       Yast::Pkg.SourceLoad
-      progress.next_minor_step("Making initial proposal")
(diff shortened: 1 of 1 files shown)


service/lib/dinstaller/software.rb
```diff
@@ -59,14 +61,16 @@ def probe(progress)
       Yast::Pkg.SourceRestore
       Yast::Pkg.SourceLoad
       progress.next_minor_step("Making initial proposal")
-      @products = Y2Packager::Product.available_base_products
+      @products = Y2Packager::Product.available_base_products.select do |product|
+        SUPPORTED_PRODUCTS.include?(product.name)
+      end
```
jreidinger, line 66: in general looks good, but please add more logging, so we know what is going on.

jreidinger (approved):

jreidinger (approved):

PR #84, opened 2022-03-22
initial react component for ssh key

jreidinger:
## Problem

Add to root form also ssh key.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/web/src/RootSSHKey.jsx
+++ b/web/src/RootSSHKey.jsx
@@ -1,0 +1,31 @@
+import React, { useState, useEffect } from "react";
+import { FileUpload } from '@patternfly/react-core';
+
+export default function RootSSHKey {
+    const (value, setValue) = useState("");
+    const (fileame, setFilename) = useState("");
+    const (loading, setLoading) = useState(false);
+
+    useEffect(async () => {
+        const key = await client.users.RootSSHKey();
+        setValue(key);
+      }, []);
+
+    return (
+        <FileUpload
+          id="simple-text-file"
+          type="text"
+          value={value}
+          filename={filename}
+          filenamePlaceholder="Drag and drop a file or upload one"
+          onFileInputChange={(_event, file) => setFilename(file.name)}
+          onDataChange={value => setValue(value )}
+          onTextChange={value => setValue(value )}
+          onReadStarted={() => setLoading(true)}
+          onReadFinished={() => setLoading(false)}
+          onClearClick={() => setFilename(""); setValue("")}
+          isLoading={loading}
+          browseButtonText="Upload"
+        />
+      );
+}
--- a/web/src/lib/client/users.js
+++ b/web/src/lib/client/users.js
@@ -50,6 +50,16 @@ export default class UsersClient {
   }
 
   /**
+   * Return string with ssh key or empty string
+   *
+   * @return {Promise.<String>}
+   */
+  async rootSSHKey() {
+    const proxy = await this.proxy(USERS_IFACE);
+    return proxy.RootSSHKey;
+  }
+
+  /**
    * Set the languages to install
    *
    * @param {object} user - object with full name, user name, password and boolean for autologin
@@ -71,7 +81,16 @@ export default class UsersClient {
     return proxy.SetRootPassword(password, false);
   }
 
-  // TODO: ssh key for root
+  /**
+   * Set the root SSH Key
+   *
+   * @param {String} key - plain text root ssh key. Empty string means disabled
+   * @return {Promise.<String|undefined>}
+   */
+   async setRootSSHKey(key) {
+    const proxy = await this.proxy(USERS_IFACE);
+    return proxy.SetRootSSHKey(key);
+  }
 }
 
 applyMixin(UsersClient, withDBus);


web/src/RootSSHKey.jsx
```diff
@@ -0,0 +1,26 @@
...
+export default function RootSSHKey(props) {
+  const [loading, setLoading] = useState(false);
+
+  return (
+    <FileUpload
+      id="SSHKey"
+      type="text"
+      value={props.value}
+      filenamePlaceholder="Drag and drop an SSH public key or upload one"
+      onDataChange={value => {
```
imobachgs, line 13: np: I think you can just write:

```
onChange={props.valueChanged}
```

web/src/RootUser.jsx
```diff
@@ -32,21 +33,24 @@ const reducer = (state, action) => {
...
 
   useEffect(async () => {
     const rootPassword = (await client.users.isRootPassword()) ? hiddenPassword : "";
+    const SSHKey = await client.users.rootSSHKey();
     dispatch({
       type: "LOAD",
       payload: {
-        rootPassword
+        rootPassword,
+        SSHKey
```
imobachgs, line 53: I am not sure about capitalization here. It is not wrong, but I would go for something like `sshKey`.

web/src/RootUser.jsx
```diff
@@ -92,7 +114,7 @@ export default function RootUser() {
   return (
     <>
       <Button variant="link" onClick={open}>
-        {rootLabel()}
+        {rootLabel().concat(SSHKeyLabel())}
```
imobachgs, line 117: It is OK. If you prefer, you can do `${rootLabel()} ${SSHKeyLabel()}` (so you can avoid having to add that tricky whitespace to the end of `rootLabel()`.

web/src/lib/client/users.js
```diff
@@ -49,6 +49,16 @@ export default class UsersClient {
     return proxy.RootPasswordSet;
   }
 
+  /**
+   * Return string with ssh key or empty string
+   *
+   * @return {Promise.<String>}
+   */
+  async rootSSHKey() {
```
imobachgs, line 57: We are using `get*` for that kind of functions. We can change it, for sure, but we should be consistent. So I would use getRootSSHKey (and I am still not sure about capitalization).

imobachgs (approved):
In general, it looks OK. I am not sure about the overall approach of putting the password and the public key in the same form and using an "empty" value to remove them (instead of using a dedicated button/link). But that's something we can change later.

I did not find any blocking issue, so feel free to merge.

web/src/RootSSHKey.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default function RootSSHKey(props) {
+  const [loading, setLoading] = useState(false);
+
+  return (
+    <FileUpload
+      id="SSHKey"
+      type="text"
+      value={props.value}
+      filenamePlaceholder="Drag and drop an SSH public key or upload one"
```
dgdavid, line 12: NP: I'd use "Upload an SSH key (drag and deop supported)" or similar. But mentioning the "Upload" action first.

web/src/RootSSHKey.jsx
```diff
@@ -0,0 +1,26 @@
...
+        props.valueChanged(value);
+      }}
+      onTextChange={value => {
+        props.valueChanged(value);
+      }}
+      onReadStarted={() => setLoading(true)}
+      onReadFinished={() => setLoading(false)}
+      onClearClick={() => props.valueChanged("")}
+      isLoading={loading}
+      browseButtonText="Upload"
```
dgdavid, line 23: NP: Upload or Search. Not sure wich is better/common.

web/src/RootSSHKey.jsx
```diff
@@ -0,0 +1,26 @@
+import React, { useState } from "react";
+import { FileUpload } from "@patternfly/react-core";
+
+export default function RootSSHKey(props) {
```
dgdavid, line 4: NP: a matter of taste, but I prefer destructuring directly here. I mean, going for `{ value, valueChange }` instead of `props`. 

Another note, for those callbaks I usually see the _convention_ **on**ValueChanged.

web/src/RootUser.jsx
jreidinger: Well, I try to use consistently SSH key as acronym. So all big. If we go with lower form it needs to be changed on multiple places and e.g. `ssh` does not look bad, but e.g. `RootSshKey` looks really strange for me.

web/src/RootSSHKey.jsx
jreidinger: will try, thanks

web/src/RootUser.jsx
jreidinger: will try, thanks

web/src/RootSSHKey.jsx
jreidinger: well, I use text from example on pattern fly side  and just slightly adapt it - https://www.patternfly.org/v4/components/file-upload/

web/src/RootSSHKey.jsx
jreidinger: Same here. I keep example from pattern fly as it was for me obviously bad.

web/src/RootUser.jsx
imobachgs: Oh, sorry, the formatting makes my suggestion not clear:

```javascript
`${rootLabel()} ${SSHKeyLabel()}`
```

I was referring to [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

web/src/RootSSHKey.jsx
dgdavid: I guessed was taken from the example, but still thinking starting with the action first would be better.

web/src/RootSSHKey.jsx
dgdavid: I see. And it is true that the action is a "Browse & Upload"... maybe the Upload label is not that bad. No strong opinion.

web/src/RootUser.jsx
```diff
@@ -57,32 +61,50 @@ export default function RootUser() {
...
     // TODO: handle errors
-    if (rootPassword !== hiddenPassword && rootPassword !== "") {
+    if (rootPassword !== hiddenPassword) {
       await client.users.setRootPassword(rootPassword);
     }
+    const remembered_password = rootPassword === "" ? "" : hiddenPassword;
+    client.users.setRootSSHKey(SSHKey);
     // TODO use signals instead
-    dispatch({ type: "ACCEPT", payload: { rootPassword: hiddenPassword } });
+    dispatch({ type: "ACCEPT", payload: { rootPassword: remembered_password } });
```
dgdavid, line 70: NOTE: I got rid of the remembered_ password in another PR. From my point of view, it's not necessary to show a fake remembered password to the user. Displaying the form for setting the new password is enough. We can discuss it later.

imobachgs (approved):

web/src/RootSSHKey.jsx
jreidinger: well, both are actions. Question is if we want to suggest more browsing or dragging.

dgdavid: True. I'm more with browsing first... because it supposed that UI is _mobile-first_. But let's postpone the decision for later PRs.

Thanks a lot!

PR #86, opened 2022-03-23
Add tests for Language classes

imobachgs:
Extend testing and documentation for Language related classes.

### DIFF
--- a/service/lib/dinstaller/language.rb
+++ b/service/lib/dinstaller/language.rb
@@ -37,15 +37,25 @@ module DInstaller
       @languages = []
     end
 
+    # Probes the available system languages
+    #
+    # @param _progress [Progress] Progress reporting object
     def probe(_progress)
       logger.info "Probing languages"
       @languages = Yast::Language.GetLanguagesMap(true)
     end
 
+    # Writes the language settings
+    #
+    # @param _progress [Progress] Progress reporting object
     def install(_progress)
       Yast::Language.Save
     end
 
+    # Sets the language
+    #
+    # @param name [String] Language code
+    # @raise Errors::InvalidValue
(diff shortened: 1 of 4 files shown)


service/test/dinstaller/dbus/language_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  describe "#available_languages" do
+    it "returns the list of available languages" do
+      expect(subject.available_languages).to eq(
+        [["en_US", "English (US)", {}], ["de_DE", "Deutsch", {}]]
+      )
+    end
+  end
+
+  describe "#marked_for_install" do
+    it "returns the languages that are selected for installation" do
```
dgdavid, line 53: (stupid) Question: Can be a list of languages marked for installation.

service/lib/dinstaller/language.rb
```diff
@@ -37,15 +37,25 @@ def initialize(logger)
       @languages = []
     end
 
+    # Probes the available system languages
+    #
+    # @param _progress [Progress] Progress reporting object
     def probe(_progress)
```
dgdavid, line 43: Question: why those methods receive a param that are not used?

dgdavid (approved):
Looks good, thanks

service/test/dinstaller/dbus/language_test.rb
imobachgs: Not yet, but I think it is planned for the future.

service/lib/dinstaller/language.rb
imobachgs: For consistency with the rest of `#probe` and `#install` methods.

PR #87, opened 2022-03-23
Relax dependency on ruby-dbus

imobachgs:
Using a conservative approach on RubyGems dependencies prevents the service to use newer ruby-dbus versions. For gems that are out of our control it is OK, but in this case let's relax the dependency so we can use anything above version 0.17.

### DIFF
--- a/service/d-installer.gemspec
+++ b/service/d-installer.gemspec
@@ -47,5 +47,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fast_gettext", "~> 2.2.0"
   spec.add_dependency "nokogiri", "~> 1.13.1"
   spec.add_dependency "rexml", "~> 3.2.5"
-  spec.add_dependency "ruby-dbus", "~> 0.17.0"
+  spec.add_dependency "ruby-dbus", ">= 0.17.0"
 end


wfeldt:
Currently we have `rubygem(ruby:3.1.0:ruby-dbus:0.17) >= 0.17` but rubygem-dbus provides (e.g.) 'rubygem(ruby:3.1.0:ruby-dbus:0.18) = 0.18.0'.

Which will never match regardless the `>=`, afaict. IMHO the dep should be be `rubygem(ruby:3.1.0:ruby-dbus:0) >= 0.17`.

Not sure, does the patch achieve that?

imobachgs:
The dependency is automatically generated while building the package (I guess it is extracted from the gem). From `rpm -q --requires` I get `rubygem(ruby:3.1.0:ruby-dbus) >= 0.17.0`, which looks fine.

I updated the package in the [devel project](https://build.opensuse.org/project/show/YaST:Head:D-Installer) (to give it a try) and it worked just fine. It installed ruby-dbus 0.18 from the same project.

wfeldt (approved):
ok, then let's try

PR #92, opened 2022-03-24
Finish installation steps

joseivanlopez:
## Problem

After finishing the installation, the target system keeps mounted in */mnt*. This makes impossible to re-launch the installer over the same target device. Moreover, installation logs are not copied to the target system.

* https://trello.com/c/CQFLj155/2907-3-d-installer-umount-and-clean-up-after-installation

## Solution

Call to finish YaST clients to copy logs and unmount the system. 

## Testing

- Tested manually

### DIFF
--- a/service/lib/dinstaller/manager.rb
+++ b/service/lib/dinstaller/manager.rb
@@ -71,7 +71,7 @@ module DInstaller
     # rubocop:disable Metrics/AbcSize
     def install
       status_manager.change(Status::Installing.new)
-      progress.init_progress(6, "Partitioning")
+      progress.init_progress(7, "Partitioning")
       Yast::Installation.destdir = "/mnt"
       # lets propose it here to be sure that software proposal reflects product selection
       # FIXME: maybe repropose after product selection change?
@@ -101,6 +101,9 @@ module DInstaller
         progress.next_step("Saving Language Settings")
         language.install(progress)
       end
+
+      progress.next_step("Finishing installation")
+      finish_installation
 
       progress.next_step("Installation Finished")
       status_manager.change(Status::Installed.new)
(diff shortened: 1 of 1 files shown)


service/lib/dinstaller/manager.rb
```diff
@@ -176,6 +179,21 @@ def probe_steps
       status_manager.change(Status::Probed.new)
     end
 
+    # Performs required steps after installing the system
+    #
+    # For now, this only unmounts the installed system and copies installation logs. Note that YaST
+    # performs many more steps like copying configuration files, creating snapshots, etc. Adding
+    # more features to D-Installer could require to recover some of that YaST logic.
+    def finish_installation
+      progress.init_minor_steps(2, "Copying logs")
```
dgdavid, line 188: NP: I'd say here are 3 minor steps, namely: "Copying logs", "Unmounting target system", "Target system correctly unmounted". Although the latest one is like a _success message_ instead of a minor step.

dgdavid: > NP: I'd say here are 3 minor steps, namely: "Copying logs", "Unmounting target system", "Target system correctly unmounted". Although the latest one is like a _success message_ instead of a minor step.
> 
> So I suggest either, to use a 3 anyway 

Never mind, I was wrong https://github.com/yast/d-installer/blob/913186d48ef6d6fae6c8e4c97ecc7c943421a8fd/service/lib/dinstaller/progress.rb#L43 since 2 means 3: 0, 1, and 2 :confused: 


> or to remove the latest _next_minor_step_ call.

But the latest step still looking more like feedback about the 2 step that a step itself.

Anyway, PR LGTM.

dgdavid (approved):
After https://github.com/yast/d-installer/pull/92#discussion_r834447016, it LGTM. Thanks

service/lib/dinstaller/manager.rb
joseivanlopez: I don't like the current API. I would prefer something like:

~~~ruby
progress = Progress.new

progress.init_steps(3)
progress.step("Doing step 1")
progress.step("Doing step 2")
progress.init_sub_steps(2)
progress.sub_step("Doing step 2.1")
progress.sub_step("Doing step 2.2")
progress.finish_sub_steps
progress.step("Doing step 3")
progress.finish
~~~

dgdavid: > I don't like the current API. I would prefer something like:
> ...

I guess we can think about it and make needed changes if we found/decide/agree on better/different approach.

PR #94, opened 2022-03-25
Initialize and save repositories

imobachgs:
## Problem

Until now, D-Installer relies on the repositories that are defined in the system where it is running. However, in the new live image, there are no repositories at all, so D-Installer is responsible for setting them up.

## Solution

This PR includes support to use the repository defined in the `/etc/install.inf` (`ZyppRepoURL`). If no repository is defined, it just uses openSUSE Tumbleweed OSS repository as fallback.

Additionally, D-Installer will copy the repositories to the installed system.

### DIFF
--- a/service/test/dinstaller/software_test.rb
+++ b/service/test/dinstaller/software_test.rb
@@ -1,0 +1,117 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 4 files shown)


service/test/dinstaller/software_test.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    it "creates a packages proposal" do
+      expect(Yast::Packages).to receive(:Proposal)
+      subject.probe(progress)
+    end
+
+    context "when a base URL is defined via the InstURL module" do
+      let(:base_url) { "dvd://" }
+
+      it "register the repository" do
```
joseivanlopez, line 76: NP: registers

service/test/dinstaller/software_test.rb
```diff
@@ -0,0 +1,117 @@
...
+        expect(Yast::Pkg).to receive(:SourceCreateBase).with(base_url, "/")
+        expect(Yast::Pkg).to receive(:SourceSaveAll)
+        subject.probe(progress)
+      end
+    end
+
+    context "when no base URL is defined" do
+      let(:base_url) { "" }
+
+      it "register the repository" do
```
joseivanlopez, line 86: NP: registers

joseivanlopez (approved):
LGTM

PR #106, opened 2022-03-30
Update changes files

imobachgs:
It is time to release the 1st public version, so let's update the changes files.


teclator (approved):
LGTM

PR #107, opened 2022-03-30
Install btrfsprogs

imobachgs:
At this point, the btrfsprogs package is not installed. We should do something like [this](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/clients/inst_disk_proposal.rb#L164), but for this first release this fix is good enough.


dgdavid (approved):
As I know you're going to write a short summary about this change before merging, it LGTM.

Thanks Imo!

PR #109, opened 2022-04-01
[web] Do not use a line as a section separator

dgdavid:
Because the headings are enough for now, although I have doubts if the explicit separator will be needed in the future (because we put some menu options / buttons or whatever per section or so)

| Before | After |
|-|-|
| Summary using explicit separator | Summary not using explicit separator |


imobachgs (approved):

ancorgs:
I agree it looks better

PR #117, opened 2022-04-05
[WIP] [RFC] [PoC] Automated screenshots

dgdavid:
A mix between a PoC and RFC for automated screenshots using [viteshot](https://viteshot.com/)

A bit tricky at this moment because there are some warnings and the development server must be running (?) for taking the screenshots with `npm run screenshots`

But the idea is to have a way to automating and sending screenshots of specific components, easing the process to make a PR and keep up-to-date the screenshots shown in the README, always using the same sizes and so on (see [_reviewing screenshots_ and _screenshot consistency_](https://viteshot.com/docs/managing-screenshots#reviewing-screenshots)).

Useful links:

* https://viteshot.com/
* https://viteshot.com/docs/managing-screenshots#option-4-upcoming-screendiff-platform
* https://playwright.dev/
* https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json


imobachgs:
I like the idea. It would help to keep the screenshots up-to-date. It think it is worth to give it a try.

teclator:
I also like it.

dgdavid:
I'm still thinking that would be a good thing for the project, but now that we switched to webpack we should look for a similar solution in the webpack world.

Thus, closing this PR.

PR #118, opened 2022-04-05
Add install confirmation

joseivanlopez:
## Problem

The *Install* button has the focus by default, and actioning it runs the installation without asking for confirmation. So, an unintentional enter keystroke raises the installation.

## Solution

Add a confirmation popup. The focus is on the *Cancel* button to avoid starting installation with unintentional double enter keystroke.

## Testing

- Added new unit tests
- Tested manually

## Screenshots

Screenshot from 2022-04-05 15-19-57

### DIFF
--- a/web/src/Overview.jsx
+++ b/web/src/Overview.jsx
@@ -19,10 +19,10 @@
  * find current contact information at www.suse.com.
  */
 
-import React from "react";
+import React, { useState } from "react";
 import { useInstallerClient } from "./context/installer";
 
-import { Button, Flex, FlexItem } from "@patternfly/react-core";
+import { Button, Modal, ModalVariant, Flex, FlexItem, Text } from "@patternfly/react-core";
 
 import Layout from "./Layout";
 import Category from "./Category";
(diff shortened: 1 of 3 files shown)


web/src/Overview.jsx
```diff
@@ -58,10 +58,41 @@ function Overview() {
...
+          title="Confirm Installation"
+          isOpen={isOpen}
+          showClose={false}
+          variant={ModalVariant.small}
+          actions={[
+            <Button key="accept" variant="primary" onClick={install}>
+              Install
+            </Button>,
+            <Button key="back" variant="primary" onClick={close} autoFocus>
+              Back
```
dgdavid, line 83: This is the first time we use "Back" in the D-Installer. Usually we use "Cancel". I'd consider rewording the message and change this by _Cancel_, using a "link" variant.

web/src/Overview.jsx
```diff
@@ -58,10 +58,41 @@ function Overview() {
...
+            </Button>
+          ]}
+        >
+          <Text>
+            If you continue, partitions on your hard disk will be modified according to the
+            installation settings in the previous dialog.
+          </Text>
+          <Text>
+            Go back and check the settings if you are unsure.
```
dgdavid, line 92: What about?

```suggestion
            If unsure, please cancel and check the settings.
```

joseivanlopez: And "Please, cancel and check the setting if you are unsure" ?

dgdavid (approved):
LGTM, thanks.

PR #119, opened 2022-04-05
Use cockpit.js from cockpit-project/cockpit#17207

imobachgs:
Now upstream cockpit.js behaves as an ES6 module (see https://github.com/cockpit-project/cockpit/pull/17207), so we can simply get the module from upstream without patching it which is great 😄

Thanks to Cockpit.js maintainers.

### DIFF
--- a/web/src/lib/cockpit.js
+++ b/web/src/lib/cockpit.js
@@ -29,7 +29,6 @@ try {
 /* injected by tests */
 var mock = mock || { }; // eslint-disable-line no-use-before-define, no-var
 
-(function() {
 const cockpit = { };
 event_mixin(cockpit, { });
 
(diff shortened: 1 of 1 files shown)


dgdavid (approved):

martinpitt:
Heh, the branch is not merged yet, but I'll work on landing this one way or the other.

BTW, our recommended approach is to checkout cockpit's [shared components](https://github.com/cockpit-project/cockpit/tree/main/pkg/lib) like [starter-kit](https://github.com/cockpit-project/starter-kit/blob/fde873e0bdeeddfa14bee46cf20c9cfec74e7abf/Makefile#L184) does. We use the same in c-podman, c-machines, and our other pages. That avoids literal code copies, thus makes it easier to bump to new versions, while retaining the ability to pin down the API and avoid surprises (kinda poor developer's npmjs.com module :grin: ). Then you can also use other shared components, like the PO and rsync webpack plugins, the Terminal widget, or the file selector.

imobachgs:
Thanks for the hint! We follow that approach in https://github.com/openSUSE/cockpit-wicked, based on the starter-kit. By now, we are just using `cockpit.js`, but we will consider using the shared components approach because it contains quite some interesting stuff.

Issue #120, opened 2022-04-06
Skip authentication when installing from localhost


imobachgs:
Introducing a user/password when running the installation locally is annoying and confusing. We need to find a way to autologin into the system so the user does not need to do it.

## Links

* [Cockpit's authentication documentation](https://github.com/cockpit-project/cockpit/blob/main/doc/authentication.md)
* [cockpit-desktop](https://cockpit-project.org/guide/latest/cockpit-desktop.1.html) might be what we need.

imobachgs:
#127 is already included in our [testing image](https://build.opensuse.org/package/show/YaST:Head:D-Installer/d-installer-live). [sr#969213](https://build.opensuse.org/request/show/969213) adjusts the image to use `cockpit-desktop. We can consider this feature as implemented.

PR #127, opened 2022-04-11
D-Installer UI as a Cockpit module

imobachgs:
Once we decided to rely on Cockpit, it makes sense to develop this UI as a regular module. Cockpit allows to:

* Skip the authentication from localhost (using the `cockpit-desktop` helper). Handy for #120.
* Loading only a single module (e.g., `/cockpit/@localhost/d-installer/index.html`).

So we do not find any limitation that justifies building the web UI in a different way.

This PR is based on [Cockpit's starter kit](https://github.com/cockpit-project/starter-kit/tree/b2379f78e203aab0028d8548b39f5f0bd2b27d2a), although we decided to drop some parts and introduce some small changes.

## Relevant changes

* `src/lib` is reserved now for Cockpit files. Our client has been moved to `src/client`.
* The code has been adapted to the new ESLint rules.
* While adapting the code to ESLint rules, we found out that we were using async functions in `useEffect` [in the wrong way](https://devtrium.com/posts/async-functions-useeffect).
* We renamed the package to `cockpit-d-installer`.

## Pending

- [ ] There are some problems with the styles, although most of them are already solved.
- [x] Development documentation must be updated.
- [ ] Move components to `src/components`. On a different PR.
- [ ] Some clean-up is still needed, as we rely on Cockpit for authentication. On a different PR ([see auth-clean-up branch](https://github.com/yast/d-installer/compare/cockpit-module...auth-clean-up?expand=1)).


joseivanlopez (approved):
LGTM

Issue #125, opened 2022-04-07
Wrong language selector behaviour


imobachgs:
The language selector does not work properly. Picking a language works, but:

* the language link changes its content before the confirm button is clicked.
* clicking the cancel button causes the selection to be reverted "English (US)".

Of course, that's not the behavior a user expects.

mvidner:
PR 130 has been merged

PR #130, opened 2022-04-11
Fix the behavior of the modal language selector

mvidner:
## Problem

As #125 says:

> The language selector does not work properly. Picking a language works, but:
> 
> -    the language link changes its content before the confirm button is clicked.
> -    clicking the cancel button causes the selection to be reverted "English (US)".
> 
> Of course, that's not the behavior a user expects.

- https://trello.com/c/MSURGLiw

## Solution

- Learned ReactJS :rofl: https://reactjs.org/tutorial/tutorial.html
- The state transition logic in `reducer` needed fixing.
## Testing

- Tested manually
- Then added new test cases
  - Tip: `npm test -- -o` runs only the tests that have changes against git. <https://jestjs.io/docs/cli>

## Screenshots

(Statically it looks the same)

### DIFF
--- a/web/src/LanguageSelector.jsx
+++ b/web/src/LanguageSelector.jsx
@@ -38,19 +38,19 @@ const reducer = (state, action) => {
       return { ...state, ...action.payload };
     }
     case "ACCEPT": {
-      return { ...state, isFormOpen: false };
+      return { ...state, isFormOpen: false, current: state.formCurrent };
     }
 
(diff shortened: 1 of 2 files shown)


dgdavid:
> * Learned ReactJS

https://www.youtube.com/watch?v=_UUFu8zjRxE :stuck_out_tongue_closed_eyes:

PR #132, opened 2022-04-12
Initial Yaml config

jreidinger:
## Problem

Dependency on product control file. Also testing ground for allowing installation of more distribution from same medium.

- https://github.com/yast/d-installer/issues/122
- https://trello.com/c/XXf1aA8u/2932-3-d-installer-define-a-mechanism-for-installer-configuration

## Solution

Define own configuration file and as POC define installation URL there.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -35,13 +35,13 @@ List of supported distros that can be offered in installer.
 ### conditions
 
 List of specific conditions. Conditions are specific keys that based on data can modify
-installatio config like distro or arch specific options. If multiple conditions
+installation config like distro or arch specific options. If multiple conditions
 are specified like arch and distro together, then both have to be satisfied.
 Below are possible keys.
 
 #### distro
 
-List of distros for which condition apply. Any of it can apply. Possible keys are from list
+List of distros for which condition apply. Any of them can apply. Possible keys are from list
 supported distros in software section. TODO: regex of glob support?
 
 #### arch
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -1,13 +1,13 @@
-YaML Config
+YAML Config
 -----------
 
-Why YaML
+Why YAML
 ========
 
 With our previous experience we discard at the start XML due to its poor readability.
 JSON format we also consider has serious disadvantage - lack of comments.
-So we decide to go with YaML which is still comfortable to read and also machine read.
-Only disadvantage we found is lack of mature schema support for YaML.
+So we decide to go with YAML which is still comfortable to read and also machine read.
+Only disadvantage we found is lack of mature schema support for YAML.
 
 Rules for Config
 ================
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -28,7 +28,7 @@ Software manager related options. It is map with following keys:
 
 Array of url for installation repositories.
 
-### distros
+### distributions
 
 List of supported distros that can be offered in installer.
 
@@ -39,7 +39,7 @@ installation config like distro or arch specific options. If multiple conditions
 are specified like arch and distro together, then both have to be satisfied.
 Below are possible keys.
 
-#### distro
+#### distribution
 
 List of distros for which condition apply. Any of them can apply. Possible keys are from list
 supported distros in software section. TODO: regex of glob support?
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -9,6 +9,12 @@ JSON format we also consider has serious disadvantage - lack of comments.
 So we decide to go with YaML which is still comfortable to read and also machine read.
 Only disadvantage we found is lack of mature schema support for YaML.
 
+Rules for Config
+================
+
+1. No default values in code. All values have to be defined in yaml config.
+2. No objects, just built-in data types.
+
 Structure
 =========
 
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -1,0 +1,48 @@
+YaML Config
+-----------
+
+Why YaML
+========
+
+With our previous experience we discard at the start XML due to its poor readability.
+JSON format we also consider has serious disadvantage - lack of comments.
+So we decide to go with YaML which is still comfortable to read and also machine read.
+Only disadvantage we found is lack of mature schema support for YaML.
+
+Structure
+=========
+
+Top level config element is map. Each key is described here in following sections.
+
+### software
+
+Software manager related options. It is map with following keys:
+
+#### installation\_repositories
+
+Array of url for installation repositories.
+
+### distros
+
+List of supported distros that can be offered in installer.
+
+### conditions
+
+List of specific conditions. Conditions are specific keys that based on data can modify
+installatio config like distro or arch specific options. If multiple conditions
+are specified like arch and distro together, then both have to be satisfied.
+Below are possible keys.
+
+#### distro
+
+List of distros for which condition apply. Any of it can apply. Possible keys are from list
+supported distros in software section. TODO: regex of glob support?
+
+#### arch
+
+List of architectures where apply condition.
+
+#### value
+
+map with same structure as top level config element that overwrites default values. Only exception
+is conditions which cannot be nested.
--- a/service/lib/dinstaller/software.rb
+++ b/service/lib/dinstaller/software.rb
@@ -145,10 +145,10 @@ module DInstaller
 
     def add_base_repo
       # TODO properly pass config around. Here it is just for quick testing
-      config = Config.new
+      config = Config.new(logger)
       config.parse_file("x86_64", "Tumbleweed")
       config.data["software"]["installation_repositories"].each do |repo|
-        Yast::Pkg.SourceCreate(repo)
+        Yast::Pkg.SourceCreate(repo, "/") # TODO: having that dir also in config?
       end
 
       Yast::Pkg.SourceSaveAll
--- a/service/lib/dinstaller/software.rb
+++ b/service/lib/dinstaller/software.rb
@@ -21,9 +21,9 @@
 
 require "yast"
 require "dinstaller/package_callbacks"
+require "dinstaller/config"
 require "y2packager/product"
 
-Yast.import "InstURL"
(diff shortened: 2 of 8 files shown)


doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+So we decide to go with YaML which is still comfortable to read and also machine read.
+Only disadvantage we found is lack of mature schema support for YaML.
+
+Rules for Config
+================
+
+1. No default values in code. All values have to be defined in yaml config.
+2. No objects, just built-in data types.
+
+Structure
```
dgdavid, line 18: Maybe would be nice to have a YAML schema or similar, but not sure if there is some solid tool/standard nowadays. Some links that can help:

* https://json-schema-everywhere.github.io/yaml
* https://github.com/dreftymac/awesome-yaml
* https://github.com/datatxt/awesome-yaml

doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+Array of url for installation repositories.
+
+### distros
+
+List of supported distros that can be offered in installer.
+
+### conditions
+
+List of specific conditions. Conditions are specific keys that based on data can modify
+installatio config like distro or arch specific options. If multiple conditions
```
dgdavid, line 38:
```suggestion
installation config like distro or arch specific options. If multiple conditions
```

doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+are specified like arch and distro together, then both have to be satisfied.
+Below are possible keys.
+
+#### distro
+
+List of distros for which condition apply. Any of it can apply. Possible keys are from list
```
dgdavid, line 44:
```suggestion
List of distros for which condition apply. Any of them can apply. Possible keys are from list
```

doc/yaml_config.md
jreidinger: well, I found https://asdf-standard.readthedocs.io/en/latest/schemas/yaml_schema.html but it also does not looks like solid/standard.

ancorgs: I also feel some kind of schema should be needed. The idea is to have a common structure for:

- The file that holds the configuration of the installer (which is being defined here)
- The options that are passed to the installer via boot parameters
- Some kind of future substitute for the AutoYaST profile

For that we need some kind of schema validation. Maybe not exactly at the yaml level, but we need something.

doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### installation\_repositories
+
+Array of url for installation repositories.
+
+### distros
+
+List of supported distros that can be offered in installer.
+
+### conditions
```
joseivanlopez, line 35: Some tools like [codefresh](https://codefresh.io/docs/docs/codefresh-yaml/condition-expression-syntax/) use a slightly different approach for conditions, for example:

~~~yaml
partitioning:
  lvm: true
  disks: [/dev/sda]
  when:
    arch: x86
    dist: TW

partitioning:
  lvm: true
  disks: [/dev/sda]
  when:
    arch: s390
~~~

 This avoid the problem of having nested conditions in *#value* as you mentioned.

jreidinger: Well, it means that we need to process that when everywhere. On other hand here it will be if condition is satisfied, it will overwrite default. And I also hope it will be more readable to have
```yaml
partitioning:
  lvm: true
  disks: [/dev/sda]

conditions:
  - arch: aarch64
  - value:
    partitioning:
       lvm: false
```

so it should be clearly visible what will change on specific architectures or distros.

jreidinger: @joseivanlopez btw I think your example won't work as you use same key name for hash. So it has to be like `tw_partitioning` and `s390_partitioning`

joseivanlopez: How that *conditions* scale when there are more than one set of conditions? Let's say we have to apply different options for: a) x86 and b) x86 and TW. Something like:

~~~yaml
partitioning:
  lvm: true
  when:
    all:
      arch: x86
    value:
      lvm: false
      when:
        all:
          dist: TW
        value:
          lvm: true
~~~

ancorgs:
I would need to see some more examples of the "conditions" thingy to really understand it

jreidinger:
@ancorgs well, usage I have in my mind is e.g. for inst-url:

```yaml
software:
  installation_repositories:
  - http://download.opensuse.org/tumbleweed/repo/oss/

conditions:
  - distro: leap15_4
    value:
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap/15.4/repo/oss/
 - distro: micro5_2
   arch: x86_64
   value:
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap-micro/5.2/product/repo/Leap-Micro-5.2-x86_64-Media/
 - distro: micro5_2
   arch: aarch64
   value:
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap-micro/5.2/product/repo/Leap-Micro-5.2-aarch64-Media/
```
so here there are generic urls for Leap and TW...for micro it exists only for aarch and x86_64 so condition is together with arch. and also url is different for different archs.

doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+
+### software
+
+Software manager related options. It is map with following keys:
+
+#### installation\_repositories
+
+Array of url for installation repositories.
+
+### distros
```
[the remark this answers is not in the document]
jreidinger: I welcome better name...just `distribution` is a bit long for me :)

doc/yaml_config.md
jreidinger: it is order dependent. So if you want to have different ones, then use at first the generic one and later specific one. I think it is quite commonly used in configuration. Basically I see there three different strategies:

1. stop on first matched conditions.
2. use the last match
3. apply all matched configs in order

For me advantage of 3 is that it does not restrict use to repeat some settings. So e.g. when we have condition for network based on arch, then software based on distro and then partitioning based on distro and arch, we can easily split it to different sections where it makes sense and really have it separated as needed.

doc/yaml_config.md
jreidinger: Well, we can also do validation ourself and there are already tools to verify if yaml is valid https://github.com/adrienverge/yamllint .
So some script that verify if everything is defined and with correct types can be also written manually ( but of course some DSL for it would be better ).

ancorgs:
I see. So "value" is used to specify the value that should be used to replace an element or section. I understand "value" is used to replace the most inner element, `software.installation_repositories` in the case of the example. What if I want to want to redefine the complete `software` section?

On the other hand, what about using "replace" instead of "value". That would allow us to have others like "add" in the future. Something like:

```yaml
software:
  installation_repositories:
  - http://download.opensuse.org/tumbleweed/repo/oss/

conditions:
  - distro: leap15_4
    replace:
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap/15.4/repo/oss/
 - distro: leap15_4
   arch: aarch64
   add:
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap/15.4/repo/additional-aarch64/
```

Anyways, I can't help the feeling we are reinventing the wheel here. This sounds like a problem that must be already solved by others.

ancorgs:
> What if I want to want to redefine the complete `software` section?

In that regard, this would look more understandable to me:

```yaml
software:
  installation_repositories:
  - http://download.opensuse.org/tumbleweed/repo/oss/
  whatever_flag: true
conditions:
  - distro: leap15_4
    replace: # Replace the software section with this content (i.e kills whatever_flag)
      software:
        installation_repositories:
        - http://download.opensuse.org/distribution/leap/15.4/repo/oss/
  - distro: leap15_5
    replace: # Replace the repositories only (i.e. leave whatever_flag untouched)
      software.installation_repositories:
        - http://download.opensuse.org/distribution/leap/15.4/repo/oss/
```

jreidinger:
@ancorgs idea with killing flag is kind of against rule to not have default values. So you cannot kill flag. You have to specify it. I want to avoid scenario we have with control file, where defaults are in yast2.rpm, then something is in leanos, then something in sles and so on. So you cannot kill flag and replace is not needed. If flag is not there, it means keep it as it was.

jreidinger:
conclusion of discussion is that we will use the simple approach and if needed we can extend it in future. E.g. by adding beside `value` also something like `append` for bigger lists. just for record yaml discussion about something like append https://github.com/yaml/yaml/issues/35

doc/yaml_config.md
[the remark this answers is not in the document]
jreidinger: agreed one is distribution to avoid confusion

doc/yaml_config.md
```diff
@@ -0,0 +1,54 @@
...
+
+With our previous experience we discard at the start XML due to its poor readability.
+JSON format we also consider has serious disadvantage - lack of comments.
+So we decide to go with YAML which is still comfortable to read and also machine read.
+Only disadvantage we found is lack of mature schema support for YAML.
+
+Rules for Config
+================
+
+1. No default values in code. All values have to be defined in yaml config.
```
imobachgs, line 15: That's fine, but we should not crash when something is missing. At least, we should try to display a proper error message.

service/lib/dinstaller/software.rb
```diff
@@ -146,9 +144,13 @@ def import_gpg_keys
     end
 
     def add_base_repo
-      base_url = Yast::InstURL.installInf2Url("")
-      base_url = FALLBACK_REPO if base_url.empty?
-      Yast::Pkg.SourceCreateBase(base_url, "/")
+      # TODO properly pass config around. Here it is just for quick testing
+      config = Config.new
```
imobachgs, line 148: I know it is just temporary code, but I would prefer to inject this configuration file through the constructor. So the manager could read the config (and eventually validate it) and pass it to each "module" (software, language, and so on).

imobachgs: OK, that's just what you commented above. But let's change it before merging, please :-)

imobachgs (approved):

PR #135, opened 2022-04-12
Add D-Bus questions

joseivanlopez:
## Problem

Sometimes the service needs to ask some information to the clients. For example, the password to activate a LUKS while probing the storage devices. D-Installer does not provide how to require such information from clients.  

- https://github.com/yast/d-installer/issues/123
- https://gist.github.com/joseivanlopez/a7568408a6bf9defcaa7a92faea9654d `dinstaller-questions-dbus.md`

## Solution

D-Installer now provides a D-Bus mechanism to ask questions. The questions are exported on D-Bus in a ObjectManager tree. 

Note: ObjectManager is not supported ruby-dbus yet. The implementation is directly provided by the `DInstaller::DBus::Questions` object.

Example:

~~~
/org/opensuse/DInstaller/Questions1
    /org/opensuse/DInstaller/Questions1/1
    /org/opensuse/DInstaller/Questions1/2
    /org/opensuse/DInstaller/Questions1/3
~~~

Note that question objects are exported and unexported dynamically, see the following code snippet: 

~~~ruby
questions_manager = DInstaller::QuestionsManager.new(logger)

DInstaller::DBus::Questions.new(questions_manager, logger) # This configures QuestionsManager callbacks to work with D-Bus

question1 = DInstaller::Question.new("Do you want to continue?", options: [:yes, :no], default_option: :no)
question2 = DInstaller::LuksQuestion.new("/dev/sda1", label: "mydata", size: "3 GiB")

# Questions are exported on DBus here (#on_add callbacks)
questions_manager.add(question1)
questions_manager.add(question2)

# Process waits until all questions are answered. Bus dispatches messages while waiting (#on_wait callbacks) 
questions_manager.wait

# Get answers
continue = question1.answer == :yes

# Question is unexported from D-Bus (#on_delete callbacks)
questions_manager.delete(question1)
~~~

## Testing

- Added new unit tests
- Tested manually

## Screenshots

Screenshot from 2022-04-26 00-52-18

### DIFF
--- a/service/lib/dinstaller/dbus/question.rb
+++ b/service/lib/dinstaller/dbus/question.rb
@@ -1,0 +1,179 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "dbus"
+require "pathname"
+require "dinstaller/dbus/questions"
+require "dinstaller/question"
+require "dinstaller/luks_question"
+
+module DInstaller
+  module DBus
+    # Class to represent a question on D-Bus
+    #
+    # Questions are dynamically exported on D-Bus. All questions are exported as children of
+    # {DBus::Questions} object to mimic ObjectManager behavior. Note that ruby-dbus does not support
+    # ObjectManager yet.
+    #
+    # Clients should provide an answer for each question.
+    class Question < ::DBus::Object
+      # This module contains the D-Bus interfaces that question D-Bus object can implement. The
+      # interfaces that a question implements are dynamically determined while creating a new
+      # D-Bus question.
+      module Interfaces
+        # Generic interface for a question
+        module Question
+          QUESTION_INTERFACE = "org.opensuse.DInstaller.Question1"
+          private_constant :QUESTION_INTERFACE
+
+          # @!method backend
+          #   @note Classes including this mixin must define a #backend method
+          #   @return [DInstaller::Question]
+
+          # Unique id of the question
+          #
+          # @return [Integer]
+          def id
+            backend.id
+          end
+
+          # Text of the question
+          #
+          # @return [String]
+          def text
+            backend.text
+          end
+
+          # Options the question admits as answer
+          #
+          # @note Clients are responsible of generating the proper label for each option.
+          #
+          # @return [Array<String>]
+          def options
+            backend.options.map(&:to_s)
+          end
+
+          # Default option a client should offer as answer
+          #
+          # @return [String]
+          def default_option
+            backend.default_option.to_s
+          end
+
+          # Answer selected for a client
+          #
+          # @return [String]
+          def answer
+            backend.answer.to_s
+          end
+
+          # Selects an option as answer
+          #
+          # @param option [String]
+          def answer=(option)
+            backend.answer = option.to_sym
+          end
+
+          def self.included(base)
+            base.class_eval do
+              dbus_interface QUESTION_INTERFACE do
+                dbus_reader :id, "u"
+                dbus_reader :text, "s"
+                dbus_reader :options, "as"
+                dbus_reader :default_option, "s"
+                dbus_accessor :answer, "s"
+              end
+            end
+          end
+        end
+
+        # Interface to request a LUKS password
+        module LuksPassword
+          LUKS_PASSWORD_INTERFACE = "org.opensuse.DInstaller.Question.LuksPassword1"
+          private_constant :LUKS_PASSWORD_INTERFACE
+
+          # @!method backend
+          #   @note Classes including this mixin must define a #backend method
+          #   @return [DInstaller::Question]
+
+          # Given password
+          #
+          # @return [String]
+          def luks_password
+            backend.password || ""
+          end
+
+          # Sets a password
+          #
+          # @param value [String]
+          def luks_password=(value)
+            backend.password = value
+          end
+
+          def self.included(base)
+            base.class_eval do
+              dbus_interface LUKS_PASSWORD_INTERFACE do
+                dbus_accessor :luks_password, "s", dbus_name: "Value"
+              end
+            end
+          end
+        end
+      end
+
+      # Defines the interfaces to implement according to the backend type
+      INTERFACES_TO_INCLUDE = {
+        DInstaller::Question     => [Interfaces::Question],
+        DInstaller::LuksQuestion => [Interfaces::Question, Interfaces::LuksPassword]
+      }.freeze
+      private_constant :INTERFACES_TO_INCLUDE
+
+      # Constructor
+      #
+      # @param backend [DInstaller::Question]
+      # @param logger [Logger]
+      def initialize(backend, logger)
+        @backend = backend
+        @logger = logger
+
+        super(build_path)
+
+        add_interfaces
+      end
+
+    private
+
+      # @return [DInstaller::Question]
+      attr_reader :backend
+
+      # Builds the question path under {DBus::Questions} path (e.g., o.o.DInstaller/Questions1/1)
+      #
+      # @return [String]
+      def build_path
+        Pathname.new(Questions.path).join(backend.id.to_s).to_s
+      end
+
+      # Adds interfaces to the question
+      def add_interfaces
+        INTERFACES_TO_INCLUDE[backend.class].each { |i| singleton_class.include(i) }
+      end
+    end
+  end
+end
--- a/service/lib/dinstaller/dbus/question.rb
+++ b/service/lib/dinstaller/dbus/question.rb
@@ -20,8 +20,6 @@
 # find current contact information at www.suse.com.
 
 require "dbus"
-require "pathname"
-require "dinstaller/dbus/questions"
 require "dinstaller/question"
 require "dinstaller/luks_question"
 
@@ -30,8 +28,7 @@ module DInstaller
     # Class to represent a question on D-Bus
     #
     # Questions are dynamically exported on D-Bus. All questions are exported as children of
-    # {DBus::Questions} object to mimic ObjectManager behavior. Note that ruby-dbus does not support
-    # ObjectManager yet.
+    # {DBus::Questions} object.
     #
     # Clients should provide an answer for each question.
     class Question < ::DBus::Object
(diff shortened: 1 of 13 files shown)


jreidinger:
Can you try to add some generic UI code ( or even CLI pseudo code ) that will present such methods? E.g. for that example question `question1 =DInstaller::Question.new("Do you want to continue?", options: [:yes, :no])` how it should be presented?

imobachgs (commented):
How questions are represented looks good to me. However, I can foresee some confusion when using this API in a multi-process environment, as each process will have its own QuestionsManager. However, that's not the problem we need to solve right now. We will take care of fixing those problems later.

service/lib/dinstaller/dbus/question.rb
```diff
@@ -0,0 +1,172 @@
...
+          # @return [String]
+          def answer
+            backend.answer.to_s
+          end
+
+          # Selects an option as answer
+          #
+          # @param option [String]
+          def answer=(option)
+            backend.answer = option.to_sym
```
jreidinger, line 87: I kind of expect here validation if answer is from options before passing to_sym otherwise we can be easy DOSed by passing random strings as answer.

joseivanlopez: Yes, I already had it on mind. I plan to start adding validations meanwhile I write unit tests. But I did prefer to not do it until validating the general approach.

service/lib/dinstaller/dbus/question.rb
```diff
@@ -0,0 +1,172 @@
...
+    # Class to represent a question on D-Bus
+    #
+    # Questions are dynamically exported on D-Bus. All questions are exported as children of
+    # {DBus::Questions} object to mimic ObjectManager behavior. Note that ruby-dbus does not support
+    # ObjectManager yet.
+    #
+    # Clients should provide an answer for each question.
+    class Question < ::DBus::Object
+      # This module contains the interfaces a question can implement. The interfaces that a question
+      # implements are dynamically determined while creating a new question.
```
jreidinger, line 39: I think you should explicitelly mention that any class that use this modules has to be ::DBus::Object otherwise it failed when calling dbus interface.

service/lib/dinstaller/dbus/question.rb
```diff
@@ -0,0 +1,172 @@
...
+      def add_interfaces
+        interfaces.each { |i| singleton_class.include(i) }
+      end
+
+      # Interfaces the question implements
+      #
+      # @return [Array<Module>]
+      def interfaces
+        interfaces = [Interfaces::Question]
+        interfaces << Interfaces::LuksPassword if backend.is_a?(DInstaller::LuksQuestion)
```
jreidinger, line 167: To be honest I do not like this part as it means that for each specialized question you need define it also here, so it is spread over multiple files. Can it live more closely? or at least add such mapping as constant to a bit more prominent place and document steps that needs to be done to have own question.

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        register_callbacks
+
+        super(PATH)
+      end
+
+      # Paths of all currently exported questions
+      #
+      # @return [Array<String>]
+      def all
```
jreidinger, line 90: I have to say that naming is really confusing here. I kind of expect that if I have `Questions.new.all` that it will return list of Question and not list of string...maybe call it `paths` ?

joseivanlopez: Umm, that `#all` is only the implementation of the dbus reader method.

service/lib/dinstaller/luks_question.rb
```diff
@@ -0,0 +1,74 @@
...
+      @device = device
+      @label = label
+      @size = size
+
+      super(generate_text, options: [:skip, :decrypt])
+    end
+
+    # Password to activate the LUKS device
+    #
+    # @return [String, nil]
```
jreidinger, line 42: please document what means nil here. Also what means empty password. ( I have suspicious that something should mean to not activate luks, but maybe it is in answer and not in password.

joseivanlopez: Yes, it is in answer. The `nil` password means nothing has been provided yet. Setting an empty password should be prevented by the client, I guess.

service/lib/dinstaller/manager.rb
```diff
@@ -54,6 +58,7 @@ class Manager
     def initialize(logger)
       @logger = logger
       @status_manager = StatusManager.new(Status::Error.new) # temporary status until probing starts
+      @questions_manager = QuestionsManager.new
```
jreidinger, line 61: I kind of expect that sooner or later we will need logger there to get some ideas what is going on.

service/lib/dinstaller/questions_manager.rb
```diff
@@ -0,0 +1,121 @@
...
+    # Adds a question
+    #
+    # Callbacks are called after adding the question, see {#on_add}.
+    #
+    # @param question [Question]
+    def add(question)
+      questions << question
+      on_add_callbacks.each { |c| c.call(question) }
+
+      questions
```
jreidinger, line 48: why it return questions and not self?

service/lib/dinstaller/questions_manager.rb
```diff
@@ -0,0 +1,121 @@
...
+      loop do
+        on_wait_callbacks.each(&:call)
+        sleep(0.1)
+        break if questions_answered?
+      end
+    end
+
+    # Registers a callback to be called when a new question is added
+    #
+    # @param block [Proc]
```
jreidinger, line 78: I kind of missing here block definition. Because on_add and on_delete have parameter and on_wait does not have it.

jreidinger: btw usually it is not documented as block, but using yield tag together with its parts for param and return https://www.rubydoc.info/gems/yard/file/docs/Tags.md#yieldparam

service/lib/dinstaller/dbus/questions.rb
jreidinger: well, it makes sense from dbus POV where questions objects are strings aka object paths, but in ruby world it is quite strange. So add least add some `@warn` or `@note` to documentation.

service/lib/dinstaller/dbus/question.rb
```diff
@@ -0,0 +1,172 @@
...
+        # Generic interface for a question
+        module Question
+          QUESTION_INTERFACE = "org.opensuse.DInstaller.Question1"
+          private_constant :QUESTION_INTERFACE
+
+          # Unique id of the question
+          #
+          # @return [Integer]
+          def id
+            backend.id
```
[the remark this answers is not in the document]
joseivanlopez: I am not sure if I got you. We have follow the very same pattern for all classes defining a D-Bus object. They receive a `backend` object as parameter in the constructor. And such a `backend` is used to implement the business logic, keeping the D-Bus object class as a mere interfaces definition. Here, `backend` is again used in the interfaces modules.

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -0,0 +1,136 @@
...
+        @exported_questions = []
+
+        register_callbacks
+
+        super(PATH)
+      end
+
+      # Paths of all currently exported questions
+      #
+      # @return [Array<String>]
```
mvidner, line 89:
```suggestion
      # @return [Array<::DBus::ObjectPath>]
```

... which `is_a? String`:
https://github.com/mvidner/ruby-dbus/blob/master/lib/dbus/object_path.rb#L15

service/lib/dinstaller/questions_manager.rb
joseivanlopez: Good poit!

service/lib/dinstaller/luks_question.rb
```diff
@@ -0,0 +1,84 @@
...
+    # @param label [String, nil] label of the device
+    # @param size [String, nil] size of the device (e.g., "5 GiB")
+    def initialize(device, label: nil, size: nil)
+      @device = device
+      @label = label
+      @size = size
+
+      # Clients should answer either skip activation (:skip) or decrypt (:decrypt) with the given
+      # password (see {#password}).
+      super(generate_text, options: [:skip, :decrypt])
```
jreidinger, line 47: NP ( or maybe not NP :) We need one central place where we document all possible options symbols. Otherwise I am pretty sure we will forget to translate ( or make proper Labels) some of them in frontends

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -0,0 +1,134 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "dbus"
+require "dinstaller/dbus/question"
+
+module DInstaller
+  module DBus
+    # This class mimics the basic functionality of ObjectManager. Note that ruby-dbus does not
+    # support ObjectManager yet.
```
jreidinger, line 28: this is one thing that is not clear for me, even if ruby-dbus does not implement auto export of object manager, you can still implement that interface yourself, not?

And according to https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager method names and interfaces are different.

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -0,0 +1,134 @@
...
+      # @warn This method is used to implement a D-Bus reader method. Do not use it to recover all
+      #   the {Question} objects.
+      #
+      # @return [Array<::DBus::ObjectPath>]
+      def all
+        exported_questions.map(&:path)
+      end
+
+      dbus_interface QUESTIONS_INTERFACE do
+        dbus_reader :all, "as"
```
jreidinger, line 93: signature looks wrong for me, as I expect here `ao`

service/lib/dinstaller/luks_question.rb
joseivanlopez: Maybe we can comment these values in the documentation of the constructor.

service/lib/dinstaller/dbus/questions.rb
joseivanlopez: The commet states "mimics the basic functionality of ObjectManager", but not its API. The question is if we want to provide the same API as ObjectManager or not. I didn't do it because we have to change clients anyway if we finally replace *o.o.DInstaller.Questions* by ObjectManager. So, for now, I created the most minimal API I thought.

service/lib/dinstaller/luks_question.rb
jreidinger: question is which constructor? Each question subclass? I think it would be better for frontend developers to have it in one single place.

service/lib/dinstaller/dbus/questions.rb
jreidinger: I do not get why we would need to replace it if we change implementation of ObjectManager interface from hand crafted one to one provided by ruby-dbus? It is same interface, so different implementation should not make different results.

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -106,14 +86,40 @@ def all
...
+      # be returned by the org.freedesktop.DBus.Properties.GetAll() method for that combination of
+      # object path and interface. If an interface has no properties, the empty hash is returned.
+      #
+      # @param question [DBus::Question]
+      # @return [Hash]
+      def interfaces_and_properties(question)
+        get_all_method = self.class.make_method_name("org.freedesktop.DBus.Properties", :GetAll)
+
+        question.intfs.keys.each_with_object({}) do |interface, hash|
+          hash[interface] = question.send(get_all_method, interface).first
```
jreidinger, line 111: I would just add here TODO to use public method when it will be available.

joseivanlopez: In fact, it is a public method, so let's use `#public_send`.

jreidinger (approved):

PR #136, opened 2022-04-19
Keep `package-lock.json`

imobachgs:
We consider keeping the `package-lock.json` in the version control as a good practice. However, the Makefile removes that file to make sure that the dependencies are updated, which partially defeats the `package-lock.json` purpose.

This PR adapts the Makefile to not modify the `package-lock.json` when running `make`. Additionally, it adds a `clean_all` target to remove the `node_modules` directory in addition to the regular clean-up.


dgdavid (approved):
LGTM

PR #137, opened 2022-04-19
Simplify webpack config file

dgdavid:
## Problem

Apart from ESlint complains, the webpack.config.file contains some stuff _apparently_ not needed:

* A kind of hack while processing a "patternfly-4-cockpit.scss" file
* An additional loader for including SVGs assets

## Solution

To apply ESlint fixes (first commit) and to remove not needed stuff.

## Testing

Manually tested that `npm run build` still producing a valid output to _dist/_ directory and that browser can navigate there.

## Suggestions for reviewers

Better review it commit by commit.

### DIFF
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -127,11 +127,10 @@ module.exports = {
           filename: 'fonts/[name][ext]',
         }
       },
+      // Load SVG files
       {
         test: /\.svg/,
-        use: {
-          loader: "svg-url-loader",
-        },
+        type: 'asset/inline',
       }
     ]
   },
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -1,8 +1,8 @@
 const fs = require("fs");
 const path = require("path");
 
-const copy = require("copy-webpack-plugin");
-const extract = require("mini-css-extract-plugin");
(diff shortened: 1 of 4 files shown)


web/webpack.config.js
```diff
@@ -88,47 +88,9 @@ module.exports = {
         use: "babel-loader",
         test: /\.(js|jsx)$/
       },
-      /* HACK: remove unwanted fonts from PatternFly's css */
-      {
-        test: /patternfly-4-cockpit.scss$/,
```
dgdavid: Please, see the commit description.

Now I have doubts if we should start importing that file instead. Comments are welcome. 

Of course, we can continue working and change our mind later.

imobachgs: I meant https://github.com/cockpit-project/cockpit/tree/267/pkg/lib/patternfly.

dgdavid: Note, there [are more options](https://github.com/patternfly/patternfly/tree/main/src/patternfly) when importing a SCSS patternfly file, like just [patternfly.scss](https://github.com/patternfly/patternfly/blob/main/src/patternfly/patternfly.scss) but it includes too much for our current use.

imobachgs (approved):

Issue #126, opened 2022-04-07
Updating to ruby-dbus 0.18.0.beta2 breaks the installer


imobachgs:
Updating to ruby-dbus 0.18.0.beta2 causes the installer to break. I found out that, for some properties, no values are returned at all. For instance, on interface `org.opensuse.DInstaller.Language1`, I am not able to get a value for `AvailableLanguages` and `MarkedForInstall` properties. Playing with `busctl`, I get this error:

```
‣ Type=method_call  Endian=l  Flags=0  Version=1 Cookie=523  Timestamp="Thu 2022-04-07 15:24:01.541983 UTC"
  Sender=:1.1515  Destination=:1.1590  Path=/org/opensuse/DInstaller/Language1  Interface=org.freedesktop.DBus.Properties  Member=Get
  UniqueName=:1.1515
  MESSAGE "ss" {
          STRING "org.opensuse.DInstaller.Language1";
          STRING "MarkedForInstall";
  };

‣ Type=error  Endian=l  Flags=0  Version=1 Cookie=195  ReplyCookie=523  Timestamp="Thu 2022-04-07 15:24:01.643540 UTC"
  Sender=:1.1590  Destination=:1.1515
  ErrorName=org.freedesktop.DBus.Error.Failed  ErrorMessage="missing keyword: :member_type; caused by 1 sender=:1.1515 -> dest=:1.1590 serial=523 reply_serial= path=/org/opensuse/DInstaller/Language1; interface=org.freedesktop.DBus.Properties; member=Get error_name="
  UniqueName=:1.1590
  MESSAGE "sas" {
          STRING "missing keyword: :member_type; caused by 1 sender=:1.1515 -> dest=:1.1590 serial=523 reply_serial= path=/org/opensuse/DInstaller/Language1; interface=org.freedesktop.DBus.Properties; member=Get error_name=";
          ARRAY "s" {
                  STRING "/home/imobach/Proyectos/yast/repos/d-installer/service/vendor/bundle/ruby/3.1.0/gems/ruby-dbus-0.18.0.beta2/lib/dbus/data.rb:555:in `initialize'";
                  STRING "/home/imobach/Proyectos/yast/repos/d-installer/service/vendor/bundle/ruby/3.1.0/gems/ruby-dbus-0.18.0.beta2/lib/dbus/data.rb:545:in `new'";
[...]

PR #142, opened 2022-04-21
Adapt JavaScript clients to ruby-dbus 0.18.0-beta.5

imobachgs:
## Problem

The UI does not work with the latest changes in ruby-dbus. Related to #126.

## Solution

Adapt the JavaScript code to work with ruby-dbus 0.18.0.beta4.

## Testing

- [x] *Adapted unit test*
- [x] *Tested manually*

## To do

- [x] Check changes in storage. It works, but the shape of the data we get through D-Bus looks a bit tricky. The problem is fixed in https://github.com/mvidner/ruby-dbus/pull/111.
- [ ] Release once beta.5 is available.

### DIFF
--- a/service/Gemfile.lock
+++ b/service/Gemfile.lock
@@ -9,7 +9,7 @@ PATH
       fast_gettext (~> 2.2.0)
       nokogiri (~> 1.13.1)
       rexml (~> 3.2.5)
-      ruby-dbus (>= 0.17.0)
+      ruby-dbus (>= 0.18.0.beta5)
 
 GEM
   remote: https://rubygems.org/
@@ -51,7 +51,7 @@ GEM
       rspec-support (~> 3.11.0)
     rspec-support (3.11.0)
     ruby-augeas (0.5.0)
-    ruby-dbus (0.17.0)
+    ruby-dbus (0.18.0.beta5)
       rexml
     simplecov (0.21.2)
       docile (~> 1.1)
--- a/service/d-installer.gemspec
+++ b/service/d-installer.gemspec
@@ -47,5 +47,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency "fast_gettext", "~> 2.2.0"
   spec.add_dependency "nokogiri", "~> 1.13.1"
   spec.add_dependency "rexml", "~> 3.2.5"
-  spec.add_dependency "ruby-dbus", ">= 0.17.0"
+  spec.add_dependency "ruby-dbus", ">= 0.18.0.beta5"
 end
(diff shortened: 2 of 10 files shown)


dgdavid (approved):

imobachgs:
Adapting the changes in the storage area, I found some interesting differences.

Version 0.17.0:

```
method return time=1650977446.409355 sender=:1.200 -> destination=:1.212 serial=136 reply_serial=46
   array [
      dict entry(
         string "All"
         variant             array [
               variant                   array [
                     dict entry(
                        string "Text"
                        variant                            string "Suprimir la partición /dev/vdb2 (19.99 GB)"
                     )
                     dict entry(
                        string "Subvol"
                        variant                            boolean false
                     )
                     dict entry(
                        string "Delete"
                        variant                            boolean true
                     )
                  ]
...
```

The list of actions (`All`) seems to be nested too deep, and that's fixed on beta4. However, action attributes seem to be wrapped with an extra "variant".

```
method return time=1650976950.558941 sender=:1.108 -> destination=:1.185 serial=1593 reply_serial=46
   array [
      dict entry(
         string "All"
         variant             array [
               array [
                  dict entry(
                     string "Text"
                     variant                         variant                            string "Suprimir la partición /dev/vdb2 (19.99 GB)"
                  )
                  dict entry(
                     string "Subvol"
                     variant                         variant                            boolean false
                  )
                  dict entry(
                     string "Delete"
                     variant                         variant                            boolean true
                  )
               ]
...
```

I am not sure whether it is right or not. @mvidner please, could you have a look?

Thanks!

imobachgs:
To add more context, they look fine when I get the value through a PropertiesChanged signal:

```
  array [
      dict entry(
         string "All"
         variant             array [
               variant                   array [
                     dict entry(
                        string "Text"
                        variant                            string "Suprimir la partición /dev/vdd2 (19.99 GB)"
                     )
                     dict entry(
                        string "Subvol"
                        variant                            boolean false
                     )
                     dict entry(
                        string "Delete"
                        variant                            boolean true
                     )
                  ]
...
```

mvidner:
Yes, the double variant is a bug. I can reproduce it with
```console
$ bundle exec spec/service_newapi.rb &
$ dbus-send --print-reply --dest=org.ruby.service /org/ruby/MyInstance \
    org.freedesktop.DBus.Properties.Get string:org.ruby.SampleInterface string:MyDict
method return time=1650979704.258429 sender=:1.1382 -> destination=:1.1385 serial=24 reply_serial=2
   variant       array [
         dict entry(
            string "one"
            variant                variant                   int32 1
         )
         dict entry(
            string "two"
            variant                variant                   string "dva"
         )
         dict entry(
            string "three"
            variant                variant                   array [
                     variant                         int32 3
                     variant                         int32 3
                     variant                         int32 3
                  ]
         )
      ]
```

dgdavid (approved):

Issue #123, opened 2022-04-06
Add a mechanism to communicate with the user from the service


imobachgs:
The service should be able to communicate with the user to:

* Ask for information (passwords, registration codes, etc.).
* Report errors (during probing, installation, etc.).

`storage-ng` and `libzypp` callbacks are typical use cases.

Additionally, the mechanism we implement should be kind of stateful, so if I got a question and I reload the browser, I should see the same question again.

## Out of scope

* Reimplementing the callbacks to use this mechanism.
* Autoinstallation and time-out questions.

PR #145, opened 2022-04-22
Add support to tweak Cockpit SSL settings

imobachgs:
## Problem

As described in #123, Cockpit uses SSL for remote access, so unless you install a known certificate, your browser will complain. This PR adds the basic support to allow disabling SSL or using your own certificate.

## Solution

The `d-installer.yaml` configuration file contains now a few settings:

```yaml
web:
  ssl: null
  ssl_cert: null
  ssl_key: null
```

If those values are set to `null`, Cockpit's configuration will not be updated. Otherwise, `d-installer` adjusts the configuration and downloads the certificates to `/etc/cockpit/ws-certs.d`.

## What is missing?

As it is right now, this configuration is not useful for our users. You need to modify the configuration file yourself and restart `d-installer` to make it work. Now, we need to add a mechanism to specify those arguments when booting the installation image (as we do with many YaST options). Something pretty basic could be:

```
web.ssl=1 web.ssl_cert=http://192.168.122.1/mycert.crt web.ssl_key=http://192.168.122.1/mycert.key
```

As we discussed, such a mechanism would allow for tweaking other settings. Additionally, we could implement support for loading and merging a configuration file from elsewhere.

However, I would like to keep those changes in a separate PR.

## Testing

- [x] Added a new unit test
- [x] Tested manually


service/etc/d-installer.yaml
```diff
@@ -7,3 +7,8 @@ software:
 distributions:
   # TODO replace simple id with better description of distributions
   - Tumbleweed
+
+web:
+  ssl: null
+  ssl_cert: null
+  ssl_key: null
```
jreidinger, line 14: please document that new keys in markdown ( maybe I should add a note to file? )

jreidinger (approved):

PR #150, opened 2022-04-26
Activate storage devices

joseivanlopez:
## Problem

While probing storage devices, the D-Installer service might need some information from the client. For example, when multipath devices are detected, the service needs to know  whether to activate multipath. Another example is when there is an encrypted device. The service needs the encryption password to open the device. 

* https://trello.com/c/7ZyVukNN/2950-5-d-installer-add-support-for-storage-activation-callbacks
* https://github.com/yast/d-installer/issues/148

## Solution

The service makes use of the [new questions mechanism](https://github.com/yast/d-installer/pull/135) to implement storage activation callbacks. The web client uses a popup to show the questions and provides an answer to the D-Installer service. 

## Testing

- Added a new unit test

  Except for #onQuestionAdded and #onQuestionRemoved _subscriptions_ after @dgdavid agreeing with @imobachgs to address them in another PR along with other not yet tested subscriptions methods

- Tested manually

## Screenshots

Screenshot from 2022-05-11 14-08-57
Screenshot from 2022-05-11 14-09-12
Screenshot from 2022-05-11 14-05-48


web/src/questions-utils.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {String} text - string to be capitalized
+ * @returns {String} capitalized text
+ */
+const label = text => `${text[0].toUpperCase()}${text.slice(1)}`;
+
+/**
+ * Returns a the options of given question as collection of Popup.Actions, using the defaultOption
+ * as the Popup.PrimaryAction
+ *
+ * TODO: Make it a component?
```
dgdavid, line 38: I know, it should be not a `TODO` but a component directly. What is more, I did the component too.

<details>
<summary>Click to show/hide the <em>QuestionActoins.jsx</em> component code</summary>

---

```jsx
/*
 * Copyright (c) [2022] SUSE LLC
 *
 * All Rights Reserved.
 *
 * This program is free software; you can redistribute it and/or modify it
 * under the terms of version 2 of the GNU General Public License as published
 * by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
 * more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program; if not, contact SUSE LLC.
 *
 * To contact SUSE LLC about this file by physical or electronic mail, you may
 */

import React from "react";
import Popup from "./Popup";
import { partition } from "./utils";

const label = action => `${action[0].toUpperCase()}${action.slice(1)}`;

export default function QuestionActions({ actions, defaultAction, actionCallback, conditions }) {
  let [[primaryAction], secondaryActions] = partition(actions, a => a === defaultAction);

  // Ensure there is always a primary action
  if (!primaryAction) [primaryAction, ...secondaryActions] = secondaryActions;

  return (
    <>
      <Popup.PrimaryAction
        key={primaryAction}
        onClick={() => actionCallback(primaryAction)}
        isDisabled={conditions?.disable?.[primaryAction]}
      >
        {label(primaryAction)}
      </Popup.PrimaryAction>
      {
        secondaryActions.map(action =>
          <Popup.SecondaryAction
            key={action}
            onClick={() => actionCallback(action)}
            isDisabled={conditions?.disable?.[action]}
          >
            {label(action)}
          </Popup.SecondaryAction>
        )
      }
    </>
  );
}
```
</details>

But due to the way how the `Popup` component works, it should not include the `<Popup.Actions>` because it is the component that [`Popup` uses to distinguish/extract the actions from the content](https://github.com/yast/d-installer/blob/66faebebe4f413eff704cc3171debc0c5d945e87/web/src/Popup.jsx#L197)

Thus, when using it as it is

```jsx

render (
  <Popup>
    <Text>The question text</Text>
    <QuestionActions />
  </Popup>
)
```

it does not work as expected because `<Popup.Actions>` is not an immediate child of Popup anymore :man_shrugging: 

Changing it to 

```jsx
render (
  <Popup>
    <Text>The question text</Text>
    <Popup.Actions>
      <QuestionActions />
    </Popup.Actions>
  </Popup>
)
```

will work and even we can argue that it is the expected. However, I decided to push it as it is now (before writing that explanation) because

* it renders the `Popup.Actions` as an immediate child of `Popup`

  ```jsx
  return (
    <Popup>
      <Text>The Question</Text>
      { buildQuestionActions(question, actionCallback) }
    </Popup>
  );
```
* as usual, I'd like to hear opinions from you

That said, at this moment I'd vote for making it a component. I don't know what I'll think tomorrow :stuck_out_tongue_winking_eye:

File: web/src/questions-utils.js [hunk 2]
```diff
@@ -0,0 +1,76 @@
...
+import Popup from "./Popup";
+import { partition } from "./utils";
+
+/**
+ * Returns given text capitalized
+ *
+ * @param {String} text - string to be capitalized
+ * @returns {String} capitalized text
+ */
+const label = text => `${text[0].toUpperCase()}${text.slice(1)}`;
```
dgdavid, line 32: Not needed by now, but this should be improved (or changed) to work with i18n too. Not to mention, that a "notNow" option will be transformed to "NotNow" instead of "Not now".

joseivanlopez: Yes, IMHO this deserves a TODO comment.

web/src/questions-utils.js
joseivanlopez: Having that `<QuestionActions>` component makes a lot of sense to me. Now we are using it only for the generation of actions in the GenericQuestion. But, actually, we can use it for the LuksActivationQuestion, too.

dgdavid: We're using it (actually its _method version_) in the LuksActivationQuestion already :) 

My doubt with the component vs the method was simply because is a bit weird to have the `QuestionAction` components wrapped inside de `Popup.Actions` (a requirement by the `Popup` design. 

A _dumb_ doubt honestly, because there is nothing wrong having either, whatever element inside the `Popup.Actions` or the _Popup_ shortcuts actions (just `Button`s) outside the `Popup.Actions`_wrapper_

In short, I'll move `buildQuestionActions` to `QuestionActions`. Everything will have more sense once we start "grouping" components instead of having all of them in _src/_

web/src/Questions.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const answerQuestion = useCallback(question => {
+    client.questions.answer(question);
+    removeQuestion(question.id);
+  }, [client.questions, removeQuestion]);
+
+  useEffect(() => {
+    client.questions.getQuestions()
+      .then(setPendingQuestions)
+      .catch(e => console.log("Something went wrong retrieving pending questions", e));
```
imobachgs, line 54: Please, log to "error".

web/src/Questions.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+      await screen.findByText("A Generic question mock");
+    });
+  });
+
+  describe("when a question is removed", () => {
+    beforeEach(() => {
+      pendingQuestions = [genericQuestion];
+    });
+
+    it("quits it from the queue", async () => {
```
imobachgs, line 82: s/quits/removes/

web/src/client/mixins.js
```diff
@@ -26,14 +26,14 @@ const withDBus = {
    * @param {string} iface - D-Bus iface
    * @return {Object} a cockpit DBusProxy
    */
-  async proxy(iface) {
+  async proxy(iface, path) {
```
imobachgs, line 29: Please, document the `path` argument. And do not forget to mention that you can leave it undefined.

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+};
+
+/**
+ * Builds a question from the given D-Bus question
+ *
+ * @param {Object} dbusQuestion
+ * @returns {Object}
+*/
+function buildQuestion(dbusQuestion) {
+  const getIfacesAndProperties = (dbusQuestion) => Object.values(dbusQuestion)[0];
```
imobachgs, line 49: code, blank, code, blank... :-) I do not see the point at all. I would get it if the function definition took several lines, but that's not the case. But it is just my POV.

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+
+  const getIfaces = (dbusQuestion) => Object.keys(getIfacesAndProperties(dbusQuestion));
+
+  const fetchValue = (ifaceProperties, key) => {
+    const dbusValue = ifaceProperties[key];
+
```
imobachgs, line 55: np: Another extra blank line.

web/src/client/questions.js
imobachgs, line 49: I am not sure if I would define the function here. If you were using a closure over `dbusQuestion` it would be fine, but that's not the case (you are passing it anyway as an argument). I am not saying that it is wrong, it is just it simply looks strange to me.

service/lib/dinstaller/storage/callbacks/activate_multipath.rb
```diff
@@ -0,0 +1,73 @@
...
+        attr_reader :questions_manager
+
+        # @return [Logger]
+        attr_reader :logger
+
+        # Question to ask for multipath activation
+        #
+        # @return [Question]
+        def question
+          text = "The system seems to have multipath hardware. " \
```
mvidner, line 65: BTW what is our approach to translations? 
I see "I18n support" in the Unplanned list of https://github.com/orgs/yast/projects/1/views/3, right?

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+   * Answer with the information in the given question
+   *
+   * @param {Object} question
+   */
+  async answer(question) {
+    const path = DBUS_CONFIG.questions.path + "/" + question.id;
+
+    if (question.type === QUESTION_TYPES.luksActivation) {
+      const proxy = this._client.proxy(DBUS_CONFIG.question.ifaces.luksActivation, path);
+      await proxy.wait();
```
imobachgs, line 119: I would expect `proxy.wait` to be [already called](https://github.com/yast/d-installer/pull/150/files#diff-2d2ba2cb24f885ee0c6fab7c64735a70ecb224aec2c08bf55aa0b88be2c4f6aaR37). But perhaps I lost track...

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+      proxy.Password = question.password;
+    }
+
+    const proxy = this._client.proxy(DBUS_CONFIG.question.ifaces.question, path);
+    await proxy.wait();
```
imobachgs, line 124: Same than above.

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+        member
+      },
+      (_path, _iface, _signal, args) => {
+        const [path, changes, invalid] = args;
+        handler(path, changes, invalid);
+      }
+    );
+    return remove;
+  }
+
```
imobachgs, line 147: Please, document this one.

web/src/client/questions.js
```diff
@@ -0,0 +1,161 @@
...
+  onQuestionAdded(handler) {
+    return this.onObjectsChanged("InterfacesAdded", (path, ifacesAndProperties) => {
+      const question = buildQuestion({ [path]: ifacesAndProperties });
+      handler(question);
+    });
+  }
+
+  onQuestionRemoved(handler) {
```
imobachgs, line 155: Please, document this one.

imobachgs (changes requested):
First of all, thanks a lot for such great work. I have added a few notes but, in general, it looks fine.

A short screencast would be a nice addition to PR's description.

service/lib/dinstaller/luks_activation_question.rb
```diff
@@ -29,22 +29,28 @@ module DInstaller
...
   #
   #   question.answer = :skip    # in case you do not want to activate the device
   #
   #   question.answer = :decrypt # in case you want to decrypt with the given password
-  class LuksQuestion < Question
+  class LuksActivationQuestion < Question
+    # Current attempt to decrypt the device
+    #
+    # @return [Integer]
+    attr_reader :attempt
```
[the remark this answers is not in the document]
joseivanlopez: Because the client needs it to know whether the same question was asked again. For example, when the provided encryption password does not work, a new identical questions is created, but with attempt+1.

service/lib/dinstaller/storage/callbacks/activate_multipath.rb
joseivanlopez: The approach is not decided yet. For now, we don't take care of translations. But yes, it is time to start thinking about it.

service/lib/dinstaller/dbus/question.rb
```diff
@@ -137,8 +145,8 @@ def self.included(base)
 
       # Defines the interfaces to implement according to the backend type
       INTERFACES_TO_INCLUDE = {
-        DInstaller::Question     => [Interfaces::Question],
-        DInstaller::LuksQuestion => [Interfaces::Question, Interfaces::LuksPassword]
+        DInstaller::Question               => [Interfaces::Question],
+        DInstaller::LuksActivationQuestion => [Interfaces::Question, Interfaces::LuksActivation]
```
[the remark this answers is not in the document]
joseivanlopez: Probably yes. I was thinking on that possibility, something like `DInstaller::Questions::Generic`, `DInstaller::Questions::LuksActivation`, `DInstaller::Questions::Manager`. I would reconsider it after this PR.

service/test/dinstaller/storage/callbacks/activate_luks_test.rb
```diff
@@ -0,0 +1,80 @@
...
+require "dinstaller/storage/callbacks/activate_luks"
+require "dinstaller/questions_manager"
+require "storage"
+
+describe DInstaller::Storage::Callbacks::ActivateLuks do
+  subject { described_class.new(questions_manager, logger) }
+
+  let(:questions_manager) { DInstaller::QuestionsManager.new(logger) }
+
+  let(:logger) { instance_double(Logger, warn: nil, info: nil, error: nil) }
```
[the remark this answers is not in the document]
joseivanlopez: I was mocking the logger because I don't want to see log messages in middle of the rspec output. Is there any other way to avoid it?

web/src/GenericQuestion.jsx
```diff
@@ -0,0 +1,47 @@
...
+import QuestionActions from "./QuestionActions";
+
+export default function GenericQuestion({ question, answerCallback }) {
+  const actionCallback = (option) => {
+    question.answer = option;
+    answerCallback(question);
+  };
+
+  return (
+    <Popup isOpen aria-label="A question from DBus">
```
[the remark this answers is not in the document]
dgdavid: You're completely right. This is a kind of leftover of first lines :) Anyway, hard to choose an aria-label here (please, note that generic questions does not show a title)

What about _D-Installer question_? Or do you prefer just _Installer question_, _Active question_, or even _Pending question_ instead?

web/src/LuksActivationQuestion.test.jsx
```diff
@@ -0,0 +1,130 @@
...
+        attempt: 1,
+        options: ["decrypt", "skip"],
+      };
+    });
+
+    it("calls the callback after setting both, answer and password", async() => {
+      const { user } = renderQuestion();
+
+      const passwordInput = await screen.findByLabelText("Encryption Password");
+
```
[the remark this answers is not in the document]
dgdavid: No reason :)

web/src/client/questions.js
joseivanlopez: Note that questions are dinamically generated and removed. We need to create a new proxy everytime we are dealing we a new question.

service/test/dinstaller/storage/callbacks/activate_luks_test.rb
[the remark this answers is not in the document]
imobachgs: Yes, what about just setting the "level" to "warn" or "error" or... Perhaps there is a better way, but IMHO this mock does not bring any value at all.

web/src/client/questions.js
dgdavid: I see your point. What is more, I'd move those functions outside the buildQuestion... a matter of taste. Let's remove the blank lines by now.

web/src/client/questions.js
imobachgs: Oh, my fault, you are doing it right.

web/src/client/questions.js
dgdavid: That's is why I commented before that I'd move those methods outside. Now that there are two opinions in the same side... lets move them.

web/src/QuestionActions.jsx
```diff
@@ -0,0 +1,78 @@
...
+ * NOTE: We use the Popup component for displaying a question, wrapping its actions inside a
+ * Popup.Actions component, which must be an immediate child of Popup. That's why we use
+ * React.Fragment (aka <>) here for wrapping the actions instead of directly using the Popup.Actions.
+ *
+ * @param {object} props - component props
+ * @param {Array.<String>} props.actions - the actions to be shown
+ * @param {String} [props.defaultAction] - the action to be shown as primary
+ * @param {function} props.actionCallback - the function to be called when user clicks on action
+ * @param {Object} [props.conditions={}] - an object holding conditions, like when an action is disabled
+ */
```
mvidner, line 49: Good docs, but it is not rendered (I'm using `npx jsdoc -r src`) and I have no idea how to fix that

web/src/utils.js
```diff
@@ -25,7 +25,7 @@
  *
  * @param {Array} collection - the collection to be filtered
  * @param {function} filter - the function to be used as filter
- * @return {[Array, Array]} the new array with grouped elements
+ * @return {Array[]} a pair of arrays, [passing, failing]
```
dgdavid, line 28: Nice!

We _fixed_ it by using `<Array.<Array, Array>` but the change has not been uploaded yet. That said, I like more your approach. Thank you!

mvidner:
BTW thanks for the excellent API docs in the comments

mvidner (approved):
Thanks!
(I've only read it, not tested)

web/src/QuestionActions.jsx
dgdavid: I'll try to explain it better later (or maybe to make a PR _fixing_ it), but as far as I can see it is not part of the rendered output because it is using `export default function Component() {}`. The Popup component, for example, is part of the output but it is using a different approach for _making it public_, 

```js
const Popup = () => {}
...
export default Popup
```

A quick test reveled that using the same technique for the QuestionActions includes it in the output. So, I suspect that it is something that can be changed using jsdoc options (I have to read about it).

web/src/Questions.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+import React from "react";
+
+import { act, screen, waitFor } from "@testing-library/react";
+import { installerRender } from "./test-utils";
+import { createClient } from "./client";
+
+import Questions from "./Questions";
+
+jest.mock("./client");
+jest.mock("./GenericQuestion", () => () => "A Generic question mock");
```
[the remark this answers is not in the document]
dgdavid: And, as @imobachgs pointed out in the IRC channle, maybe 

```js
const FakeComponent = () => "A fake component"
jest.mock("src/WithoutMagic", () => FakeComponent);
```

it is more clear. However, have a look to _Overview.test.jsx_. I prefer having one-liners there and, for consistency, in every unit test we need that mocking.

PR #152, opened 2022-05-03
Update testing-libary/user-event package

dgdavid:
## Problem

[@testing-library/user-event](https://testing-library.com/docs/user-event/intro) package recommends updating to the latest version due to important bug fixes and new features.

> We recommend updating your projects to this version, as it includes important bug fixes and new features.

## Solution

Update such a package to its latest version and adapt test suite codebase to make it work. Basically, start using `userEvent.setup()` inside the `installerRender` util for [_applying workarounds and mock the UI layer to simulate user interactions_](https://testing-library.com/docs/user-event/setup).

## Testing

- Adapted unit tests

### DIFF
--- a/web/src/test-utils.js
+++ b/web/src/test-utils.js
@@ -20,6 +20,7 @@
  */
 
 import React from "react";
+import userEvent from "@testing-library/user-event";
 import { render } from "@testing-library/react";
 
 import { InstallerClientProvider } from "./context/installer";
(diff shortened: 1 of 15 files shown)


web/src/test-utils.js
```diff
@@ -30,8 +31,9 @@ const InstallerProvider = ({ children }) => {
   return <InstallerClientProvider client={client}>{children}</InstallerClientProvider>;
 };
 
-const installerRender = (ui, options = {}) => {
-  return render(ui, { wrapper: InstallerProvider, ...options });
-};
+const installerRender = (ui, options = {}) => ({
+  user: userEvent.setup(),
+  ...render(ui, { wrapper: InstallerProvider, ...options })
+});
```
dgdavid, line 37: I do not have a strong preference here right now. We can also provide a separated `user` util and use it where needed instead of including it here. What do you think?

```js
const setupUser = () => userEvent.setup();

const installerRender = (ui, options = {}) => {
  return render(ui, { wrapper: InstallerProvider, ...options });
};

export { 
  setupUser,
  installerRender
}
```

And then, 

```js
import { setupUser, installerRender } from 'test-utils';

it("looks better for you", async () => {
  installerRender(<DInstaller />);  
  
  const user = setupUser();
 
  await user.click("Install");
});
```

teclator (approved):
LGTM

PR #161, opened 2022-05-17
Handle correctly cmdline boolean values

teclator:
## Problem

In - https://github.com/yast/d-installer/pull/158
 we extended the installer configuration supporting to override the default values through the kernel command line options but as we do not have an schema or the type of the attributes that we are defining the override could be problematic in some cases. Current problem is the **web.ssl** option which is boolean and which default value is not initialized.

## Solution

By now we have decided to just convert **false** and **true** strings to **Booleans**

## Testing

- *Added a new unit test*

### DIFF
--- a/service/lib/dinstaller/cmdline_args.rb
+++ b/service/lib/dinstaller/cmdline_args.rb
@@ -48,7 +48,7 @@ module DInstaller
         # Omit config_url from Config options
         next args.config_url = value if key == "config_url"
 
-        # TODO: Add some kind of schema or 'knowdlege' of attribute types to convert them properly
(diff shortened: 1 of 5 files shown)


service/lib/dinstaller/cmdline_args.rb
```diff
@@ -48,6 +48,9 @@ def self.read_from(path)
         # Omit config_url from Config options
         next args.config_url = value if key == "config_url"
 
+        # TODO: Add some kind of schema or 'knowdlege' of attribute types to convert them properly
```
dgdavid, line 51: typo: knowledge

dgdavid (approved):
LGTM

PR #164, opened 2022-05-18
POC for separate users service

jreidinger:
## Problem

POC for having separate dbus processes

## Solution

WIP

## Testing

- *Added a new unit test*
- *Tested manually*

## Tasks

- [x] separate service
- [x] code to call users from main service
- [ ] code to call Package.Available from users service to main service
- [ ] testing

### DIFF
--- a/service/bin/d-installer-users
+++ b/service/bin/d-installer-users
@@ -1,0 +1,41 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+#
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+# TEMPORARY overwrite of Y2DIR to use DBus for communication with dependent yast modules
+ENV["Y2DIR"] = File.expand_path("../lib/dinstaller/dbus/y2dir", __dir__)
+
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+
+require "rubygems"
+require "bundler/setup"
+
+require "eventmachine"
+require "dinstaller/dbus/users_service"
+
+logger = Logger.new($stdout)
+
+EM.run do
+  service = DInstaller::DBus::UsersService.new(logger)
+  service.export
+  EventMachine::PeriodicTimer.new(0.1) { service.dispatch }
+end
--- a/service/lib/dinstaller/dbus/users_service.rb
+++ b/service/lib/dinstaller/dbus/users_service.rb
@@ -20,25 +20,19 @@
 # find current contact information at www.suse.com.
 
 require "dbus"
-require "dinstaller/dbus/manager"
-require "dinstaller/dbus/language"
-require "dinstaller/dbus/software"
 require "dinstaller/dbus/users"
-require "dinstaller/dbus/storage/proposal"
-require "dinstaller/dbus/storage/actions"
-require "dinstaller/dbus/questions"
 
 module DInstaller
   module DBus
     # D-Bus service (org.opensuse.DInstaller)
     #
     # It connects to the system D-Bus and answers requests on objects below
-    # `/org/opensuse/DInstaller`.
-    class Service
+    # `/org/opensuse/DInstaller/Users`.
+    class UsersService
       # Service name
       #
       # @return [String]
-      SERVICE_NAME = "org.opensuse.DInstaller"
+      SERVICE_NAME = "org.opensuse.DInstaller.Users"
       private_constant :SERVICE_NAME
 
       # System D-Bus
@@ -46,15 +40,8 @@ module DInstaller
       # @return [::DBus::Connection]
       attr_reader :bus
 
-      # Installation manager
-      #
-      # @return [DInstaller::Manager]
-      attr_reader :manager
-
-      # @param manager [Manager] Installation manager
       # @param logger [Logger]
-      def initialize(manager, logger = nil)
-        @manager = manager
+      def initialize(logger = nil)
         @logger = logger || Logger.new($stdout)
         @bus = ::DBus::SystemBus.instance
       end
@@ -85,45 +72,12 @@ module DInstaller
       # @return [Array<::DBus::Object>]
       def dbus_objects
         @dbus_objects ||= [
-          manager_dbus,
-          language_dbus,
-          software_dbus,
           users_dbus,
-          storage_proposal_dbus,
-          storage_actions_dbus,
-          questions_dbus
         ]
       end
 
-      def manager_dbus
-        @manager_dbus ||= DInstaller::DBus::Manager.new(manager, logger)
-      end
-
-      def language_dbus
-        @language_dbus ||= DInstaller::DBus::Language.new(manager.language, logger)
-      end
-
-      def software_dbus
-        @software_dbus ||= DInstaller::DBus::Software.new(manager.software, logger)
-      end
-
       def users_dbus
-        @users_dbus ||= DInstaller::DBus::Users.new(manager.users, logger)
-      end
-
-      def storage_proposal_dbus
-        @storage_proposal_dbus ||= DInstaller::DBus::Storage::Proposal.new(
-          manager.storage.proposal, storage_actions_dbus, logger
-        )
-      end
-
-      def storage_actions_dbus
-        @storage_actions_dbus ||=
-          DInstaller::DBus::Storage::Actions.new(manager.storage.actions, logger)
-      end
-
-      def questions_dbus
-        @questions_dbus ||= DInstaller::DBus::Questions.new(manager.questions_manager, logger)
+        @users_dbus ||= DInstaller::DBus::Users.new(Users.new(logger), logger)
       end
     end
   end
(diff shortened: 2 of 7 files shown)


service/bin/d-installer-users
```diff
@@ -0,0 +1,41 @@
...
+
+require "rubygems"
+require "bundler/setup"
+
+require "eventmachine"
+require "dinstaller/dbus/users_service"
+
+logger = Logger.new($stdout)
+
+EM.run do
```
mvidner, line 37: NP: We should use `EventMachine` everywhere for consistency

service/bin/d-installer-users
```diff
@@ -0,0 +1,41 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+# TEMPORARY overwrite of Y2DIR to use DBus for communication with dependent yast modules
```
mvidner, line 24:
```suggestion
# FIXME: for monkey-patching Autologin.rb (yast2-pam.rpm) to make it work over D-Bus
```

service/lib/dinstaller/dbus/users_service.rb
```diff
@@ -0,0 +1,84 @@
...
+require "dbus"
+require "dinstaller/dbus/users"
+
+module DInstaller
+  module DBus
+    # D-Bus service (org.opensuse.DInstaller)
+    #
+    # It connects to the system D-Bus and answers requests on objects below
+    # `/org/opensuse/DInstaller/Users`.
+    class UsersService
```
mvidner, line 31:
```
 service/lib/dinstaller/dbus/clients/users.rb
 service/lib/dinstaller/dbus/users.rb
 service/lib/dinstaller/dbus/users_service.rb
```

This is getting complicated, I'll think about naming, namespacing and commenting the architecture so that it makes sense for newcomers

mvidner:
<details>
<summary>I was curious, here's the diff of changes for Autologin:</summary>

```diff
--- Autologin.rb        2022-05-19 10:54:52.063810791 +0200
+++ AutologinD.rb       2022-05-19 10:54:18.944404665 +0200
@@ -45,10 +45,6 @@
 
       Yast.import "Package"
       Yast.import "Popup"
-      Yast.import "Pkg"
-      Yast.import "PackageCallbacks"
-      Yast.import "Installation"
-      Yast.import "Stage"
 
       # User to log in automaticaly
       @user = ""
@@ -180,8 +176,7 @@
     #
     # @return Boolean
     def supported?
-      pkg_lazy_init
-      supported = DISPLAY_MANAGERS.any? { |dm| Pkg.IsSelected(dm) || Pkg.IsProvided(dm) }
+      supported = DISPLAY_MANAGERS.any? { |dm| Package.Available(dm) }
 
       if supported
         log.info("Autologin is supported")
@@ -192,26 +187,6 @@
       supported
     end
 
-    # Initialize the pkg subsystem
-    def pkg_lazy_init
-      # do not initialize the package manager when running in inst-sys,
-      # it is initialized by the installation framework (bsc#1135295)
-      return if Stage.initial || @pkg_initialized
-
-      # We don't strictly need any package callbacks here, but libzypp might
-      # report an error, and then there would be no user feedback.
-      PackageCallbacks.InitPackageCallbacks
-      # FIXME: Mode.test workaround for the old testsuite to not break the other modules
-      Pkg.TargetInitialize(Installation.destdir) unless Mode.test
-
-      # Add the installed system to the libzypp pool
-      # FIXME: Mode.test workaround for the old testsuite to not break the other modules
-      Pkg.TargetLoad unless Mode.test
-
-      @pkg_initialized = true
-    end
-
-
     publish :variable => :user, :type => "string"
     publish :variable => :pw_less, :type => "boolean"
     publish :variable => :used, :type => "boolean"
```
</details>

service/bin/d-installer-users
mvidner: Hm, now that i've seen the diff, my description does not seem true.

service/bin/d-installer-users
jreidinger: well, it is just copied code from `bin/d-installer` so it is consistent with other service

service/bin/d-installer-users
jreidinger: well, for now just for testing I use code from my pull request. But it looks not good simplification, so I see in the end how to do it in this POC

jreidinger:
@mvidner real change with test case can be seen at https://github.com/yast/yast-pam/pull/29

jreidinger:
@mvidner btw the real change to use dbus should be in Package Module overwrite which should implement that available.

jreidinger:
Output of benchmark comparing one method call for all resolvables versus calling it for each resolvable (see perf.rb file):

```
                      user     system      total        real
ask multi:        0.286406   0.019677   0.306083 ( 48.372429)
single call:      0.062197   0.003379   0.065576 (  8.968695)
```

joseivanlopez:
NP: if finally we go with several services, then we should revisit  our code structure. I would prefer to have more isolation for each service,for example `DInstaller::Software::DBus::Service`, and everything for software (lib classes to commnunicate with YaST a.k.a. backend, dbus objects, etc) under `DInstaller::Software`.

doc/multi_process_backend.md
```diff
@@ -0,0 +1,33 @@
+## Multi Process Backend
+
+It is attempt to have backend consisting of several dbus services that communicates between.
+The goal of attempt is to verify that having separate services can speed up probing phase and maybe even slightly installation one.
```
imobachgs, line 4: Another important goal is to be responsive (via D-Bus) even if d-installer is busy. For instance, while reading metadata from repositories or installing packages, the D-Bus interface does not respond. We would like to avoid such a problem (at least with long-term operations).

doc/multi_process_backend.md
```diff
@@ -0,0 +1,61 @@
...
+  It basically checks which of supported Display Managers are available.
+- ProductFeatures which should be in dinstaller replaced by Config. It can mean
+  that each process needs to evaluate dinstaller config. Ignored in POC for now.
+
+What was agreed:
+
+- for POC it uses Y2DIR with modified dependent modules that uses dbus to
+  provide info. It is now located at `service/lib/dinstaller/dbus/y2dir`
+  where is modified yast sources
+- Having `DInstaller::DBus::Clients for dbus clients that is needed to
```
joseivanlopez, line 50: Mission closing `: ``DInstaller::DBus::Clients`

doc/multi_process_backend.md
```diff
@@ -1,61 +1,56 @@
...
-  two APIs - one for Clients that discover new questions and reply to them and
-  one for Services which asks questions and then get notified that result is
-  available.
-- The next step is to have separate service for Software as it can have
-  the biggest impact due to its slow probe and installation. So we can do next
-  evaluation when we have it and do software probing asynchronous.
+- We agreed on using a separate process for questions. The API should allow to
+  asking questions and replying to them.
+- Software is the most time-consuming aspect of the installation, so we should
+  aim to move it to a separate process.
```
jreidinger, line 56: nice improvement of document :+1:

service/bin/d-installer-users
```diff
@@ -0,0 +1,41 @@
...
+ENV["Y2DIR"] = File.expand_path("../lib/dinstaller/dbus/y2dir", __dir__)
+
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+
+require "rubygems"
```
joseivanlopez, line 29: Do we really need to require "rubygems" and "bundler/setup" here?

service/lib/dinstaller/dbus/clients.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
joseivanlopez, line 1: Missing license text.

service/lib/dinstaller/dbus/clients/dinstaller.rb
```diff
@@ -0,0 +1,59 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "dbus"
+
+module DInstaller
+  module DBus
+    module Clients
+      # Client class to call methods from main dinstaller service
+      class DInstaller
```
joseivanlopez, line 28: Why is this client called `DInstaller`? I would expect something like `Software`.

service/lib/dinstaller/dbus/service.rb
```diff
@@ -32,7 +32,7 @@ module DInstaller
   module DBus
     # D-Bus service (org.opensuse.DInstaller)
     #
-    # It connects to the system D-Bus and answers requests on objets below
+    # It connects to the system D-Bus and answers requests on objects below
     # `/org/opensuse/DInstaller`.
     class Service
```
joseivanlopez, line 37: From now on, due to we are going to provide different services, I would use more specific service names, for example: `DInstaller::DBus::Services::Manager`.

service/lib/dinstaller/dbus/y2dir/modules/Autologin.rb
```diff
@@ -0,0 +1,207 @@
...
+      Disable() if @used && Popup.YesNo(question)
+      @used
+    end
+
+    # Check if autologin is supported with the currently selected or installed
+    # packages.
+    #
+    # @return Boolean
+    def supported?
+      supported = DInstaller::DBus::Clients::DInstaller.provisions_selected?(DISPLAY_MANAGERS).any?
```
joseivanlopez, line 181: Are we copying the whole module only because this line?

service/lib/dinstaller/dbus/clients/dinstaller.rb
jreidinger: well, this client should represent main manager service. So it is the client for main DInstaller service

service/lib/dinstaller/dbus/service.rb
jreidinger: Well, it will make sense when we provide only manager there like probe, install and status. But for now it contains much more methods that is not just manager one.

service/lib/dinstaller/dbus/y2dir/modules/Autologin.rb
jreidinger: yes. It is how Y2DIR works. it is like self-update or LOAD_PATH in ruby. It prefers loading different file. For me it was easier then doing some dynamic monkey patching, as this needs to be done early due to some side effect when doing require/import. So this is safe way to do it without risking too early execution.

joseivanlopez:
As agreed, we are going to merge this PR as it is in the *multi-process* feature branch and we will continue working there.

Issue #155, opened 2022-05-11
Generate the JavaScript documentation


imobachgs:
Although we document the JavaScript code with [JSDoc](https://jsdoc.app/), we are not building that documentation. If we want to build the docs, we need to run jsdoc manually. And what is even worse, the format of our documentation might be wrong.

Ideally, we should define a new [script](https://github.com/joseivanlopez/d-installer/blob/d42c1d19de54a9059a6a89e9135767f3da906334/web/package.json#L7-L13) in the `package.json` to generate the documentation.

## Acceptance criteria

- [x] Running a single command generates the documentation
- [ ] Formatting issues are fixed (_mvidner_: which issues? Errors in tag syntax, reported by jsdoc? Currently we have none.)
- [x] The documentation is generated as part of our CI
- [ ] Update tasks documentation (_mvidner_: we're not sure what we meant with this :-/ anyway, added a short note to web/README.md )

dgdavid:
Notes: Some weeks ago I was playing a bit with related things. Namely, 

* JSDoc theme/templates
  - [docdash](https://github.com/clenemt/docdash) a JSDoc 3 template inspired in (former?) [lodash documentation website](https://lodash.com/docs/4.17.15). See [available options](https://github.com/clenemt/docdash#options)
  - [minami](https://github.com/nijikokun/minami), another clean and minimal JSDoc theme/template
* Other tools probably out of the scope at this moment
  - [Docusaurus.io](https://docusaurus.io) - _beautiful documentation site in no time_. See [showcases](https://docusaurus.io/showcase), specially the [Ionic documentation](https://ionicframework.com/docs/) and [Ionic components documentation](https://ionicframework.com/docs/components)
  - [docz](https://www.docz.site/docs/introduction)
  - [React Styleguidist](https://github.com/styleguidist/react-styleguidist)
  -  [Storybook](https://storybook.js.org/) and similar tools ([Ladle](https://ladle.dev/docs/), [Vitebook](https://github.com/vitebook/vitebook), ).Out of scope at this moment, I guess.

mvidner:
Fixed in #186.

```sh
$ cd web
$ npm run jsdoc

> jsdoc
> npx jsdoc -c jsdoc.conf.json && echo Tip: && echo xdg-open jsdoc.out/index.html

Tip:
xdg-open jsdoc.out/index.html
$ xdg-open jsdoc.out/index.html
```

PR #170, opened 2022-05-26
Display machine ip

mchf:
## Problem

D-Installer should display target's ip. The ip has to be read from NetworkManager via d-bus

- *Bugzilla link*
- *openQA link*

## Solution

An link-style button is shown in the footer of s-installer's main page. The link contains first ip (whatever it is ;-), after clicking shows popup with the rest of ips.

## Testing

- *Tested manually*

## Screenshots


web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+};
+
+export default function NetworkManager() {
+  const client = useInstallerClient();
+  const [state, dispatch] = useReducer(reducer, initIpData);
+
+  useEffect(() => {
+    const first_ip = async () => {
```
dgdavid, line 44: NP: in JavaScript camelCase is preferred over snake_case style. I.e., use _firstIp_ instead.

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,62 @@
...
+        type: "READ",
+        payload: data
+      });
+    };
+
+    first_ip();
+  }, [client.network]);
+
+  return (
+    <>
```
dgdavid, line 58: NP: since you're not returning multiple sibling elements, you don't need the [React.Fragment](https://reactjs.org/docs/fragments.html) here.

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,62 @@
...
+      // for now just string expected as the payload
+      // for now state struct holds just one member
+      return { ip: action.payload };
+    }
+  }
+};
+
+export default function NetworkManager() {
+  const client = useInstallerClient();
+  const [state, dispatch] = useReducer(reducer, initIpData);
```
dgdavid, line 41: Np: Reducer is ok, but I'd use `useState` instead of `useReducer` here, just because the component it's dealing with a simple piece of state.

web/src/client/network.js
```diff
@@ -0,0 +1,44 @@
...
+  constructor(dbusClient) {
+    this._client = dbusClient;
+  }
+
+  /**
+   * Return the first ip
+   *
+   * @return {Promise.<Object>}
+   */
+  async first_ip() {
```
dgdavid, line 36: NP: let's use `firstIp` instead.

web/src/Target.jsx
```diff
@@ -0,0 +1,56 @@
...
+ * find current contact information at www.suse.com.
+ */
+
+import React, { useState } from "react";
+import Popup from "./Popup";
+import { Button, Text } from "@patternfly/react-core";
+
+import NetworkManager from "./NetworkManager";
+
+export default function Ip() {
```
dgdavid, line 28: NP: It does not matter which name you put in the default export, but we tend to use the same as chosen for the file. Maybe `IpAddresses` suits better for both, the component and the file. But it is just a suggestion. I really sucks at naming stuff :)

BTW,  did you know that `default` exports are not named export? So you can import them with the name of your preference, although usually we pick the original name. More about _exports_ and _imports_ at https://javascript.info/import-export#a-word-against-default-exports and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

dgdavid (commented):
@mchf I have left a few comments there, most of them NPs. BTW, it starts looking great :)

web/src/NetworkManager.jsx
mvidner: Reference:
- https://reactjs.org/docs/hooks-reference.html#usestate
- https://reactjs.org/docs/hooks-reference.html#usereducer

web/Makefile
```diff
@@ -91,6 +91,11 @@ devel-install: $(WEBPACK_TEST)
 	rm -f ~/.local/share/cockpit/$(PACKAGE_NAME)
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
+custom-install: $(WEBPACK_TEST)
+	mkdir -p ~/.local/share/cockpit
+	rm -f ~/.local/share/cockpit/$(PACKAGE_NAME)
+	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
```
jreidinger, line 97: what is purpose of this? It looks same as devel-install

mvidner: Yes, it is just another name for `devel-install`. Why?

web/src/Layout.jsx
```diff
@@ -25,6 +25,7 @@ import "./layout.scss";
 import logo from "./assets/suse-horizontal-logo.svg";
 
 import About from "./About";
+import Ip from "./Target";
```
mvidner, line 28: The name "ip" is OK for a local variable, but I think it is too ambiguous here. What about "TargetMachineAddress"

web/src/Target.jsx
mvidner: Ip what? Target what?
I don't know the React naming conventions, but the words you chose are way too generic. And you have "resolved" David's note without answering :)

web/src/client/network.js
```diff
@@ -0,0 +1,84 @@
...
+      };
+    });
+
+    return {
+      addresses: addrs,
+      hostname: await this.hostname()
+    };
+  }
+
+  async connections() {
```
mvidner, line 46: We know that the NM API is complex.
Please add links to NM reference documentation so that in a few weeks/months we can understand what's going on.
Like this for example:

```suggestion
/**
 * Return object paths implementing o.fd.NM.Connection.Active
 *  https://developer-old.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Connection.Active.html
 */
  async connections() {
```

mvidner: BTW if you care about the rendered APi docs (optional), this should work, while #155 is not done:

```console
cd web
npx jsdoc src
xdg-open out/index.html
```

mvidner:
@mchf the CI should be green now. You can run the tests locally with
`cd web; npm test`

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,102 @@
...
+        { firstIp } ({state.hostname})
+      </Button>
+
+      <Popup
+        isOpen={isOpen}
+        title={state.hostname}
+        aria-label="IP Addresses"
+      >
+        <Text>
+          { ips.map((ip) => <p key={ip}> {ip} </p>) }
```
dgdavid, line 94: Note: Patternfly's `<Text>` component  is already _tranlated_ to an HTML `<p>` element by default. See https://www.patternfly.org/v4/components/text#text

Thus, I'd use 

```jsx
{ ips.map(ip => <Text key={ip}>{ip}</Text>) }
``` 

Although maybe it's better to use a plain [`<List>`](https://www.patternfly.org/v4/components/list#plain). No strong opinion at this moment.

dgdavid: NP: as per [HTML specification](https://html.spec.whatwg.org/#the-p-element), `<p>` element is categorized as [_flow content_](https://html.spec.whatwg.org/#flow-content-2) and [_palpable content_](https://html.spec.whatwg.org/#palpable-content-2) and can only contain [_phrasing content_](https://html.spec.whatwg.org/#phrasing-content-2). A bit messy way to say that a `<p>` must not contain another `<p>`.

web/src/client/index.js
```diff
@@ -27,16 +27,22 @@ import StorageClient from "./storage";
...
 
 const SERVICE_NAME = "org.opensuse.DInstaller";
+const NM_SERVICE_NAME = "org.freedesktop.NetworkManager";
 
 const createClient = () => {
   const client = cockpit.dbus(SERVICE_NAME, {
     bus: "system",
     superuser: "try"
   });
+  const nmClient = cockpit.dbus(NM_SERVICE_NAME, {
```
imobachgs, line 42: To be honest, I am not sure which is the best way to organize those clients. But most probably, I would move this `nmClient` variable inside the `NetworkClient`.

The difference between `client` and `nmClient` is that the former is used by many classes, so we wanted to share the same instance. But I am not sure whether it was a good idea.

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,102 @@
...
+      });
+    };
+
+    config();
+  }, [client.network]);
+
+  return state;
+}
+
+export default function DetailsPopup() {
```
dgdavid, line 72: NP: I'd keep export names _in sync_ with file name. As I already said, actually it _does not matter_ which name you use for the default export since you can import it using the name of your preference, but just to follow the same style as the other components.

web/Makefile
mchf: sorry ... some experiment which i forgot to remove

mvidner:
Oops, my and Imo's commits from yesterday afternoon got lost. I'll fix it.
@mchf please be careful with force-pushing.

web/src/NetworkManager.jsx
mchf: So, React, d-bus and now patternfly ... I'm like Columbus in India ;-)

web/src/client/index.js
mchf: for now I'm not able to have strong opinion here ... ENOTENOUGHEXPERIENCE ;-)

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,95 @@
...
+      </Button>
+
+      <Popup
+        isOpen={isOpen}
+        title={state.hostname}
+        aria-label="IP Addresses"
+      >
+        <Text>
+          { ips.map((ip) => <Text key={ip}> {ip} </Text>) }
+        </Text>
```
dgdavid, line 88:
```suggestion
        { ips.map((ip) => <Text key={ip}> {ip} </Text>) }
```

web/src/client/network.js
```diff
@@ -0,0 +1,118 @@
...
+
+export default class NetworkClient {
+  constructor(dbusClient) {
+    this._client = dbusClient;
+  }
+
+  /**
+   * Returns IP config overview - addresses and hostname
+   *
+   * @return {Promise.<Map>} address key stores list of addresses,
```
dgdavid, line 34: NP: I'd say the promise returns an [array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) of [objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), not a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

web/src/client/network.js
```diff
@@ -0,0 +1,118 @@
...
+    const data = await this.#addresses();
+    const addrs = data.map(a => {
+      return {
+        address: a.address.v,
+        prefix: a.prefix.v
+      };
+    });
+
+    return {
+      addresses: addrs,
```
dgdavid, line 47: Hint: actually you can use `const addresses = data.map....` above and then use directly here as below

```suggestion
      addresses,
```

See https://javascript.info/object#property-value-shorthand

web/src/NetworkManager.jsx
```diff
@@ -0,0 +1,95 @@
...
+      return state;
+    }
+  }
+};
+
+function formatIp(address, prefix) {
+  return address + "/" + prefix;
+}
+
+export default function DetailsPopup() {
```
dgdavid, line 52: Please, use another name.

I'm so bad at naming, but 

* this component does not looks like as a "NetworkManager" to me. I see it more like a "IpAddresses" or "TargetIpAdresses" or whatever.
* in any case I'd use the same name for both, the file and the component/default export function.

web/src/Target.jsx
```diff
@@ -0,0 +1,31 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import NMDetailsPopup from "./NetworkManager";
+
+export default function TargetAddressOverview() {
+  return (
+    <>
```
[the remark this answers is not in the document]
mchf: May be I'm a bit over engineering here, but ... idea was to have it as an abstraction layer if network backend changes from NM to something else.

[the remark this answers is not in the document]
mchf: in the end ... keeps it simple ... removed intermediate step.

web/src/TargetIpsPopup.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+
+  const open = () => setIsOpen(true);
+  const close = () => setIsOpen(false);
+  const client = useInstallerClient();
+  const [state, setState] = useState(initIpData);
+  const ips = state.addresses.map((addr) => formatIp(addr.address, addr.prefix));
+  const firstIp = ips.length > 0 ? ips[0] : "";
+
+  useEffect(() => {
```
imobachgs, line 46: You do not need to track whether the component is mounted. And sure you do not need an `async` function here. There might be better ways to do it, but I would change the approach to:

1. Extend `NetworkClient` to listen for changes in NetworkManager. I would expect to be possible to listen for a path or a signal to detect such changes.
2. Allow registering a callback for 1). Let's call it `onChange`.
3. Register the callback in the `useEffect`.

```javascript
useEffect(() => {
  return client.network.onChange(changes => {
    // you can call the setState here.
  });
}, []); // the dependencies array is empty so this useEffect executes only once
```

As the dependency array is empty, React will `client.network.onChange` only once, to register the callback. Beware that the `onChange` function should return a function to unsubscribe (see [onObjectChange as example](https://github.com/yast/d-installer/blob/98b9d2ea8344b8dafae0be8e16c01611db48081b/web/src/client/mixins.js#L63)). React will call that function when the component is unmounted, so no more *Can't perform a React state update... on an unmounted component* errors.

If listening for changes in NetworkManager is too hard, we could ignore the changes by now.

web/src/client/network.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /*
+   * Returns list of active NM connections
+   *
+   * Private method.
+   * See NM API documentation for details.
+   * https://developer-old.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html
+   *
+   * @return {Promis.<Array>}
```
imobachgs, line 72: typo: Promise

web/src/TargetIpsPopup.jsx
```diff
@@ -0,0 +1,70 @@
...
+  hostname: ""
+};
+
+function formatIp(address, prefix) {
+  return address + "/" + prefix;
+}
+
+export default function TargetIpsPopup() {
+  const [isOpen, setIsOpen] = useState(false);
+
```
mvidner, line 38:
```suggestion
```

web/src/TargetIpsPopup.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const open = () => setIsOpen(true);
+  const close = () => setIsOpen(false);
```
mvidner, line 40:
```suggestion
  const close = () => setIsOpen(false);

```

Just a small whitespace change to bundle the `isOpen` logic together.

mvidner (approved):

PR #172, opened 2022-05-26
Microos poc

jreidinger:
## Problem

Create a POC for installaling microOS from the same tumbleweed medium.

## Solution

The first step is to create patch branch that is not intended to merge. It is just to proof that concept works.
With given changes it works when there are no repositories on system with d-installer. ( something we probably need to solve if we want to support multiple OSes ).

## Testing

- *Tested manually*

## Screenshots


jreidinger:
close as already part of multi product pr

PR #166, opened 2022-05-20
Add CLI

joseivanlopez:
## Feature

This PR implements the first version of a CLI for the DInstaller service. Note, this CLI is a quick and initial proposal. The idea is to have somenthing to present in the openSUSE Conference 22 in order to show how the web UI reacts to changes performed from the CLI. All the commands will be revisited later. Moreover, we will reevaluate whether to continue using ruby or to reimplement the CLI with a different (compiled) language. 

This CLI provides the following commands:

~~~
dinstallerctl help COMMAND
dinstallerctl status
dinstallerctl install
dinstallerctl language SUBCOMMAND
dinstallerctl software SUBCOMMAND
dinstallerctl storage SUBCOMMAND
dinstallerctl rootuser SUBCOMMAND
dinstallerctl user SUBCOMMAND
~~~

In general, evething that can be done with the web UI can also be done with the CLI, excepts answering questions. This CLI will be extended to support installation from a config file.

Related to:

* https://trello.com/c/qjBEtfg0/2961-8-d-installer-build-a-cli-to-install-the-system-from-an-autoyast-like-profile
* https://github.com/yast/d-installer/issues/163

## Testing

- Added unit tests
- Tested manually

## Screenshots

d-installer-cli

PR #180, opened 2022-06-07
refresh the UI when detected DBUS object changes

teclator:
## Problem

The UI gets refreshed when the status or the storage actions change, no matter where the change comes from (busctl, the new CLI [#166](https://github.com/yast/d-installer/pull/166), etc.). However, that's not the case for other configuration values. For instance, if you change the language via D-Bus, the web UI does not get updated.

We should implement support to refresh the interface in those situations.

- https://trello.com/c/ygoE9Xdx/2970-2-d-installer-refresh-the-web-ui-when-a-configuration-option-changes

## Solution

All the components are now subscribed to DBus object changes refreshing the UI accordingly. 

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

- See the demo [here](https://youtu.be/jTZ8XwvHBfQ?t=115)

### DIFF
--- a/web/src/RootPassword.test.jsx
+++ b/web/src/RootPassword.test.jsx
@@ -21,7 +21,7 @@
 
 import React from "react";
 
-import { screen, waitFor, within } from "@testing-library/react";
+import { act, screen, waitFor, within } from "@testing-library/react";
 import { installerRender } from "./test-utils";
 import { createClient } from "./client";
 
@@ -31,6 +31,7 @@ jest.mock("./client");
 
 let isRootPasswordSetFn;
 let setRootPasswordFn = jest.fn();
+let onUsersChangeFn = jest.fn();
 const getRootSSHKeyFn = () => "";
 const removeRootPasswordFn = jest.fn();
 
@@ -42,7 +43,8 @@ beforeEach(() => {
         isRootPasswordSet: isRootPasswordSetFn,
         setRootPassword: setRootPasswordFn,
         removeRootPassword: removeRootPasswordFn,
-        getRootSSHKey: getRootSSHKeyFn
+        getRootSSHKey: getRootSSHKeyFn,
+        onUsersChange: onUsersChangeFn
       }
     };
   });
@@ -145,5 +147,30 @@ describe("when an error happens while changing the password", () => {
     await user.click(cancelButton);
 
     await screen.findByText(/Something went wrong/i);
+  });
+});
+
+describe("when the Users change", () => {
+  let callbacks;
+
+  beforeEach(() => {
+    callbacks = [];
+    onUsersChangeFn = cb => callbacks.push(cb);
+  });
+
+  describe("and the RootPassword has been modified", () => {
+    it("updates the proposal root password description", async () => {
+      installerRender(<RootPassword />);
+      let rootPassword = await screen.findByText(/Root password/i);
+      within(rootPassword).getByRole("button", { name: "is not set" });
+
+      const [cb] = callbacks;
+      act(() => {
+        cb({ rootPasswordSet: true });
+      });
+
+      rootPassword = await screen.findByText(/Root password/i);
+      within(rootPassword).getByRole("button", { name: "is set" });
+    });
   });
 });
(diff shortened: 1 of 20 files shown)


web/src/Storage.jsx
```diff
@@ -71,9 +71,10 @@ export default function Storage() {
         candidateDevices: [disk]
       } = await client.storage.getStorageProposal();
       const actions = await client.storage.getStorageActions();
+      const target = disk || disks[0]?.id;
```
dgdavid, line 74: NP: Not sure if worth it adding a reminder about _"`disks` is a collection of plain JavaScript objects while `disk` is just an id"_.

dgdavid: Thinking it twice, maybe using _diskId_ instead _disk_ [while destructuring  `candidateDevices`](https://github.com/yast/d-installer/pull/180/files#diff-7e1258ac61a3996cefbc2a9726cd5962b3767c14e55817d7f5351cd97db2c2a5R71) could help more to future readers since that line will become


```suggestion
      const target = diskId || disks[0]?.id;
```

web/src/client/storage.js
```diff
@@ -25,6 +25,7 @@ import { applyMixin, withDBus } from "./mixins";
 const STORAGE_PROPOSAL_IFACE = "org.opensuse.DInstaller.Storage.Proposal1";
 const STORAGE_ACTIONS_IFACE = "org.opensuse.DInstaller.Storage.Actions1";
 const ACTIONS_PATH = "/org/opensuse/DInstaller/Storage/Actions1";
+const TARGET_PATH = "/org/opensuse/DInstaller/Storage/Proposal1";
```
dgdavid, line 28: NP:

```suggestion
const STORAGE_PROPOSAL_PATH = "/org/opensuse/DInstaller/Storage/Proposal1";
```

dgdavid: Or

```suggestion
const PROPOSAL_PATH = "/org/opensuse/DInstaller/Storage/Proposal1";
```

I'd go for that and even change the STORAGE_XYZ_IFACE ones for not using the STORAGE prefix.

web/src/client/storage.js
```diff
@@ -81,6 +82,18 @@ export default class StorageClient {
       handler({ All: newActions });
     });
   }
+
+  /**
+   * Register a callback to run when properties in the Target object change
+   *
+   * @param {function} handler - callback function
+   */
+  onTargetChange(handler) {
```
dgdavid, line 91: NP:

```suggestion
  onStorageProposalChange(handler) {
```

web/src/client/users.js
```diff
@@ -122,6 +124,24 @@ export default class UsersClient {
...
   }
+
+  /**
+   * Register a callback to run when properties in the Users object change
+   *
+   * @param {function} handler - callback function
+   */
+  onUsersChange(handler) {
+    return this.onObjectChanged(USERS_PATH, changes => {
+      if (changes.RootPasswordSet) {
```
imobachgs, line 135: I wonder if we can get more than one change simultaneously (for instance, using the CLI).

teclator: is that really possible?, wouldn't be two different events? I mean we could have both for example with different clients using the CLI.. but would not expect to be a problem...

web/src/RootPassword.test.jsx
```diff
@@ -147,3 +149,28 @@ describe("when an error happens while changing the password", () => {
...
+describe("when the Users change", () => {
+  let callbacks;
+
+  beforeEach(() => {
+    callbacks = [];
+    onUsersChangeFn = cb => callbacks.push(cb);
+  });
+
+  describe("and the RootPassword has been modified", () => {
+    it("updates the proposal root password description", async () => {
```
dgdavid, line 162: NP: I'd use "refreshes the UI" or something similar for all of them :)

dgdavid (approved):
LGTM, thank you

web/src/client/users.js
imobachgs: `changes` is plural, so I expect you can have more than one change. Let's say you have a "PropertiesChanged" with many properties... I am not saying today, with our current UI, it is possible. But programmatically, I would expect it is possible.

teclator: Yep, not with the current UI and CLI implementation but it could happen.

PR #182, opened 2022-06-08
[WIP] Simplify DBusError tests

imobachgs:
## Problem

`DBusError` tests:

* rely on installer context
* run on async

Running on installer context is not needed (not even desired) and async is only needed for the last test.

## Solution

* Allow rendering a component without the installer context.
* Use async only where it is needed.


imobachgs:
Replaced by #183.

PR #184, opened 2022-06-08
extend config with security and more software settings

jreidinger:
## Problem

Extend yaml config with patterns for given product and also with more detailed security settings. It is needed for support of microos from yaml.

## Testing

- *Tested manually*

### DIFF
--- a/doc/yaml_config.md
+++ b/doc/yaml_config.md
@@ -28,6 +28,40 @@ Software manager related options. It is map with following keys:
 
 Array of url for installation repositories.
 
+#### mandatory\_patterns
+
+Array of patterns that have to be selected.
+
+#### optional\_patterns
+
+Array of patterns that should be selected but can be deselected or skipped if not available.
+
+### security
+
+Options related to security
+
+#### lsm
+
+Default linux security module. Currently supported values are `selinux`, `apparmor` and `none`.
+
+#### available\_lsms
+
+Map for available linux security modules. If only one module is
+available it means that lsm is not configurable.
+
+##### patterns
+
+Required patterns for given lsm to be selected.
+
+##### policy
+
+Default policy. Only applicable for selinux lsm.
+
+
+#### selinux\_policy
+
+Default policy for selinux. It is ignored if lsm is not selinux.
+
 ### distributions
 
 List of supported distros that can be offered in installer.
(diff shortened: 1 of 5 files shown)


doc/yaml_config.md
```diff
@@ -28,6 +28,40 @@ Software manager related options. It is map with following keys:
...
+#### lsm
+
+Default linux security module. Currently supported values are `selinux`, `apparmor` and `none`.
+
+#### available\_lsms
+
+Map for available linux security modules. If only one module is
+available it means that lsm is not configurable.
+
+##### patterns
```
ancorgs, line 52: Just a side note not meant as a review of this: I was a big fan of the idea of having just one configuration file to shape the installation process instead of separate concepts for AutoYaST profile and control.xml. But these kind of thinks make me wonder.

doc/yaml_config.md
```diff
@@ -28,6 +28,40 @@ Software manager related options. It is map with following keys:
...
+
+Required patterns for given lsm to be selected.
+
+##### policy
```
ancorgs, line 56: Is this not the same than "selinux_policy" below?

jreidinger: yeah, my fault. I consider both options and looks like I forget it there. Thanks for notice it.

doc/yaml_config.md
jreidinger: To be honest me too. I think it is kind of two separate things:

1. product specific settings, some are optional, but some are mandatory
2. user choices, which is basically capture of user intentions. Can be direct ones like CLI calls with specific value or meta ones like tuning storage.

jreidinger:
@ancorgs regarding that coverage it looks like bug in commenter of coveralls as I see proper coverage in build info at coveralls https://coveralls.io/builds/49884177

service/lib/dinstaller/security.rb
```diff
@@ -23,10 +23,45 @@
...
+# instead use DIinstaller::Config
+# TODO: add ability to set product features in LSM::Base
+module Y2Security
+  module LSM
+    # modified LSM Base class to use dinstaller config
+    class Base
+      def product_feature_settings
+        return @product_feature_settings unless @product_feature_settings.nil?
+
+        value = ::DInstaller::Config.current.data["security"]["available_lsms"][id.to_s]
```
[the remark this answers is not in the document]
imobachgs: We had a task in the [GitHub project](https://github.com/orgs/yast/projects/1). I turned it into #189.

ancorgs (approved):
LGTM

service/lib/dinstaller/software.rb
```diff
@@ -169,5 +183,26 @@ def find_products
...
+    REPOS_BACKUP = "/etc/zypp/repos.d.dinstaller.backup"
+    private_constant :REPOS_BACKUP
+
+    REPOS_DIR = "/etc/zypp/repos.d"
+    private_constant :REPOS_DIR
+
+    # ensure that repos backup is there and repos.d is empty
+    def store_original_repos
+      # Backup was already created, so just remove all repos
+      if File.directory?(REPOS_BACKUP)
```
ancorgs, line 196: Maybe we should log the store/restore actions. I kind of expect that to become relevant for debugging at some point.

jreidinger: makes sense for me.

PR #185, opened 2022-06-09
Unify d-installer executables

imobachgs:
* `d-installer users` runs the users service.
* Just `d-installer` or `d-installer manager` runs the main service.
* Related to #164.


service/lib/dinstaller/dbus/service_runner.rb
```diff
@@ -0,0 +1,97 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "eventmachine"
+require "dinstaller/manager"
+require "dinstaller/dbus/service"
+require "dinstaller/dbus/users_service"
```
jreidinger, line 25: I am not sure if it is good idea to open everything in every service. Especially giving problems yast have with some Yast.imports ... so maybe require it dynamically based on param?

service/lib/dinstaller/dbus/service_runner.rb
```diff
@@ -0,0 +1,97 @@
...
+        service
+      end
+
+      # Set up a service
+      #
+      # @param name [String] Service name (ie, "users")
+      # @param logger [Logger] Service logger
+      # @return [#export,#dispatch] Class that implements #export and #dispatch methods.
+      def setup_service(name, logger)
+        klass = SERVICES_MAP[name]
```
jreidinger, line 90: setting `Y2DIR` env is missing in this approach.

jreidinger: and what should be also considered, that maybe we will need different y2dirs for different services as it can collide between itself. But it is something that future will show

imobachgs: Yes, I agree.

service/lib/dinstaller/dbus/service_runner.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      # Set up a service
+      #
+      # @param name [String] Service name (ie, "users")
+      # @param logger [Logger] Service logger
+      # @return [#export,#dispatch] Class that implements #export and #dispatch methods.
+      def setup_service(name, logger)
+        require "dinstaller/dbus/#{name}_service"
+        klass = DInstaller::DBus.const_get("#{name.capitalize}Service")
+        klass.new(logger)
```
joseivanlopez, line 86: Maybe we could move the logic of `setup_manager` to the constructor of `Dinstaller::DBus::Service` and initialize all the services in the very same way.

imobachgs: To be honest, I do not have a clear view of how the `Manager` class will evolve now that we are moving some pieces outside. IMHO `Manager` does too much (sets up Cockpit, initializes other services, etc.).

I see your point, and perhaps it is the right approach. But by now, I prefer to inject the already initialized manager into the `Service`, so the `Service` has nothing to do (just to serve as an API on top). This piece of logic lives in a kind of ancillary class (`ServiceRunner`) and it should be pretty straightforward to move it later.

But I might be wrong.

imobachgs: The thing I do not feel comfortable with is that, in other cases, the services are taking care of initializing the backends. And I do not know if I like that.

PR #186, opened 2022-06-09
Add: npm run jsdoc

mvidner:
## Problem

https://github.com/yast/d-installer/issues/155

> Although we document the JavaScript code with [JSDoc](https://jsdoc.app/), we are not building that documentation. If we want to build the docs, we need to run jsdoc manually. And what is even worse, the format of our documentation might be wrong.

> Ideally, we should define a new [script](https://github.com/joseivanlopez/d-installer/blob/d42c1d19de54a9059a6a89e9135767f3da906334/web/package.json#L7-L13) in the package.json to generate the documentation.

## Solution

- add a "jsdoc" script to `package.json`.
- instead of CLI options, prefer a configuration file
- in web/README.md, say  how to use it: `npm run jsdoc`

TODOs

- [x] automatic deployment to surge.sh looks easy https://github.com/yavisht/deploy-via-surge.sh-github-action-template but probably can wait to another PR
- [x] probably need to declare `jsdoc` in `package.json`

## Testing

- This is developer tooling. Simple enough not to need tests: either it works or it doesn't

## Screenshots

d-installer-jsdoc

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -44,6 +44,7 @@
     "jed": "^1.1.1",
     "jest": "^27.5.1",
     "jest-transform-stub": "^2.0.0",
+    "jsdoc": "^3.6.10",
     "mini-css-extract-plugin": "^2.5.3",
     "po2json": "^1.0.0-alpha",
     "qunit": "^2.9.3",
--- a/web/package.json
+++ b/web/package.json
@@ -9,6 +9,7 @@
     "build": "webpack",
     "eslint": "eslint --ext .js --ext .jsx src/",
     "eslint:fix": "eslint --fix --ext .js --ext .jsx src/",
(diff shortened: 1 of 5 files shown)


imobachgs (approved):

web/package.json
```diff
@@ -9,6 +9,7 @@
     "build": "webpack",
     "eslint": "eslint --ext .js --ext .jsx src/",
     "eslint:fix": "eslint --fix --ext .js --ext .jsx src/",
+    "jsdoc": "npx jsdoc -c jsdoc.conf.json; echo Tip:; echo xdg-open out/index.html",
```
mvidner, line 12: BTW it doesn't work without `npx` for me, and I'd like to understand why

jreidinger (approved):
in general looks good. Maybe I would just change `out` to something like `jsdoc` or `jsdoc.out`

Issue #168, opened 2022-05-25
Move software management to a separate process


imobachgs:
Software management is the most time-consuming part of D-installer. It takes some time to read the repositories metadata at the beginning and, of course, most of the installation time is spent installing packages. And while D-Installer is busy dealing with packages, the D-Bus interface is kind of blocked.

In our road to a multi-processes architecture (see #153), software management should be the next step. The idea is to extract this logic into a separate process.

Bear in mind that having a separate process for a time-consuming layer might affect how the installer status is determined. For instance, you cannot enable the "Install" button until the software layer is ready.

Open question: should we keep a separate "status" for each service we extract?

PR #187, opened 2022-06-09
Refactor main Service

imobachgs:
Let's consider this one as preparation work for #168.

* `DBus::Service` is now called `DBus::ManagerService`.
* The `Manager` is instantiated within the service (like `UsersService`).
* Simplify the `ServiceRunner`, as now all services have a similar API.
* The configuration is injected from the `ServiceRunner`.

## To do

* Should we use `#export` or `#start`?

### DIFF
--- a/service/lib/dinstaller/dbus/service_runner.rb
+++ b/service/lib/dinstaller/dbus/service_runner.rb
@@ -21,7 +21,6 @@
 
 require "eventmachine"
 require "dinstaller/manager"
-require "dinstaller/dbus/service"
 
 module DInstaller
   module DBus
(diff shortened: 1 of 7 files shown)


service/lib/dinstaller/dbus/service_runner.rb
```diff
@@ -62,28 +58,21 @@ def run
...
-        manager = DInstaller::Manager.new(logger)
-        manager.setup
-        service = DInstaller::DBus::Service.new(manager, logger)
-        manager.probe # probe after export, so we can report on DBus
-        manager.progress.on_change { service.dispatch } # make single thread more responsive
-        service
+      # Configuration
+      def config
+        Config.load unless Config.current
+        Config.current
```
jreidinger, line 64: maybe we should add that load to current method? So it lazy loads and user of config do not need to care?

imobachgs: It makes sense. But I would like to move the decision to #189. I will make a note there.

jreidinger (approved):

PR #188, opened 2022-06-09
Configurable storage volumes

ancorgs:
## Problem

The `Y2Storage::GuidedSetup` tries to accommodate a set of volumes. That's configurable in YaST (and is usually defined by the product, the installation role and other criteria). But in D-Installer there was no way to specify the set of volumes, so it always created the fallback set of volumes that is hard-coded at YaST.

## Solution

Now `d-installer.yaml` can contain a 'volumes' section that is identical to the [corresponding section](https://github.com/yast/yast-installation/blob/master/doc/control-file.md#the-volumes-subsection) of YaST configuration (just using Yaml instead of XML, of course).

The code still uses the fallback legacy list if the list of volumes is missing in the configuration file, but this pull request also syncs the current `d-installer.yaml` with the Tumbleweed configuration [available at openSUSE's skelcd](https://github.com/yast/skelcd-control-openSUSE/blob/master/control/control.openSUSE.xml#L249) repo.

## Testing

- Tested manually with:
  - The hard-coded fallback
  - The TW-like configuration mentioned above
  - A configuration equivalent to MicroOS
- Included the first unit tests for `DInstaller::Storage::Proposal`

### DIFF
--- a/service/etc/d-installer.yaml
+++ b/service/etc/d-installer.yaml
@@ -12,3 +12,72 @@ web:
   ssl: null
   ssl_cert: null
   ssl_key: null
+
+storage:
+  volumes:
+  - mount_point: "/"
+    fs_type: btrfs
+    desired_size: 10 GiB
+    min_size: 5 GiB
+    max_size: unlimited
+    weight: 30
+
+    # There must always be a root
+    proposed_configurable: false
+
+    snapshots: true
+    snapshots_percentage: 250
+    snapshots_configurable: true
+    # Disable snapshots if there is not enough room
+    disable_order: 3
+
+    btrfs_default_subvolume: "@"
+    subvolumes:
+    - path: home
+    - path: opt
+    - path: root
+    - path: srv
+    - path: usr/local
+    # Unified var subvolume - https://lists.opensuse.org/opensuse-packaging/2017-11/msg00017.html
+    - path: var
+      copy_on_write: false
+
+    # Architecture specific subvolumes
+    - path: boot/grub2/arm64-efi
+      archs: aarch64
+    - path: boot/grub2/arm-efi
+      archs: arm
+    - path: boot/grub2/i386-pc
+      archs: x86_64
+    - path: boot/grub2/powerpc-ieee1275
+      archs: ppc,!board_powernv
+    - path: boot/grub2/s390x-emu
+      archs: s390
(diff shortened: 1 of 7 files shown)


service/etc/d-installer.yaml
```diff
@@ -12,3 +12,72 @@ web:
...
+      archs: riscv64
+
+  - mount_point: "/home"
+    fs_type: xfs
+    desired_size: 40 GiB
+    min_size: 10 GiB
+    max_size: unlimited
+    weight: 60
+
+    proposed: false
```
jreidinger, line 68: I think we should at least link current existing documentation, as some keys looks strange for me.

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -120,11 +125,22 @@ def calculate(settings = {})
       def generate_proposal_settings(settings)
         proposal_settings = Y2Storage::ProposalSettings.new_for_current_product
 
+        config_volumes = read_config_volumes
+        proposal_settings.volumes = config_volumes unless config_volumes.empty?
```
jreidinger, line 129: what means here empty config volumes? That product does not require any partitioning?

service/etc/d-installer.yaml
```diff
@@ -12,3 +12,72 @@ web:
...
+    # There must always be a root
+    proposed_configurable: false
+
+    snapshots: true
+    snapshots_percentage: 250
+    snapshots_configurable: true
+    # Disable snapshots if there is not enough room
+    disable_order: 3
+
+    btrfs_default_subvolume: "@"
```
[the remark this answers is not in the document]
ancorgs: Not all. Default values and other aspects are explained in the documentation for the YaST control file, that is now linked.

service/etc/d-installer.yaml
ancorgs: Of course, the plan was to link to the corresponding YaST documentation. Now it's done.

service/lib/dinstaller/storage/proposal.rb
ancorgs: I added a comment (and a test case) to clarify that.

jreidinger (approved):

ancorgs:
@jreidinger This would be the corresponding section for microos
https://paste.opensuse.org/49605740

PR #191, opened 2022-06-10
JSDoc: make FooClient classes visible to it

mvidner:
## Problem

```js
export default class Foo { ... }
```
works for JS but sadly not for JSDoc.

The rendered docs has just 1 entry instead of all the classes:
> ### Classes
> exports

## Solution

Use

```js
class Foo {
  ...
}
export default Foo;
```

> ### Classes
> 
> LanguageClient
> ManagerClient
> Monitor
> NetworkClient
> QuestionsClient
> SoftwareClient
> StorageClient
> UsersClient

The same is probably true for `function`

## Testing

- *Tested manually*: `npm run jsdoc` before and after

### DIFF
--- a/web/src/client/language.js
+++ b/web/src/client/language.js
@@ -23,7 +23,10 @@ import { applyMixin, withDBus } from "./mixins";
 
 const LANGUAGE_IFACE = "org.opensuse.DInstaller.Language1";
 
-export default class LanguageClient {
+/**
+ * Language client
+ */
+class LanguageClient {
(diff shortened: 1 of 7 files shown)


dgdavid (approved):
Looks good. It's the same fot components too... but it is ok to adapt them later (once they are more documented). Thanks!

mvidner:
> It's the same for components too...

 OK, I see, like `QuestionActions`. Will do those next :)

PR #192, opened 2022-06-12
Leap micro

jreidinger:
## Problem

Allow to install from one medium TW, Leap Micro and Leap itself.

## Solution

With usage of POC and preparation pull requests put it together and implement it.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/service/lib/dinstaller/manager.rb
+++ b/service/lib/dinstaller/manager.rb
@@ -140,7 +140,7 @@ module DInstaller
     #
     # @return [Software]
     def software
-      @software ||= Software.new(logger, config)
+      @software ||= Software.new(logger, config, self)
     end
 
     # Language manager
--- a/service/etc/d-installer.yaml
+++ b/service/etc/d-installer.yaml
@@ -1,99 +1,174 @@
-software:
-  installation_repositories:
-    - https://download.opensuse.org/tumbleweed/repo/oss/
-    - https://download.opensuse.org/tumbleweed/repo/non-oss/
-    - https://download.opensuse.org/update/tumbleweed/
-  mandatory_patterns:
-    - enhanced_base # only pattern that is shared among all roles on TW
-  optional_patterns: null # no optional pattern shared
-
-security:
-  lsm: apparmor
-  available_lsms:
-    apparmor:
-      patterns:
-        - apparmor
-    selinux:
-      patterns:
-        - selinux
-      policy: permissive
-    none:
-      patterns: null
-
-distributions:
-  # TODO replace simple id with better description of distributions
+products:
+  # TODO replace simple id with better description
   - Tumbleweed
+  - Leap Micro
 
 web:
   ssl: null
   ssl_cert: null
   ssl_key: null
 
-storage:
-  volumes:
-  - mount_point: "/"
-    fs_type: btrfs
-    desired_size: 10 GiB
-    min_size: 5 GiB
-    max_size: unlimited
-    weight: 30
 
-    # There must always be a root
-    proposed_configurable: false
+Tumbleweed:
+  software:
+    installation_repositories:
+      - https://download.opensuse.org/tumbleweed/repo/oss/
+      - https://download.opensuse.org/tumbleweed/repo/non-oss/
+      - https://download.opensuse.org/update/tumbleweed/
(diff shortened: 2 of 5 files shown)


service/lib/dinstaller/storage/manager.rb
```diff
@@ -58,7 +58,7 @@ def install(_progress)
       #
       # @return [Storage::Proposal]
       def proposal
-        @proposal ||= Proposal.new(logger, config)
+        @proposal ||= Proposal.new(logger, config, actions)
```
joseivanlopez, line 61: Maybe we can use the callbacks approach instead of passing objects:

~~~
@proposal ||= Proposal.new(logger, config).tap do |proposal|
  proposal.on_change { actions.changed! }
end
~~~

joseivanlopez: Well, it depends who is the resposible of orchestrating what happens when a proposal (or whatever object) changes. For example, if we consiser that `Manager` should do that work, then I would also do something like:

~~~
@software ||= Software.new(logger, config).tap do |software|
  software.on_product_change { self.probe } 
end
~~~

But, if we consider that `Proposal` should know when to call `Actions` callbacks, then I would keep as it is.

jreidinger: For me it is more about who depends on what. If actions depends on proposal or other way. So for me nature way is to pass proposal to actions ( or use callback when creating actions ) as actions depends on proposal and not vice versa.

Regarding that if we should use for depedencies callbacks or passing directly object. It does not matter too much for me, only if there is more uses for callbacks, then callbacks is easier to share.

dgdavid:
> better UI for switching distro

I have been doing some test:

| `<Radio>` wrapped into a plain `<Card>` | `<Radio>` wrapped into a isSelectable `<Card>` | `<Tile>` |
|-|-|-|
| Screen Shot 2022-06-15 at 19 43 27-fullpage | Screen Shot 2022-06-15 at 19 42 27-fullpage | Screen Shot 2022-06-15 at 20 08 52-fullpage |
| It is probably the one I'd choose for this PR. We can change or improve it in the future | There are some troubles with navigation by keyboard, hovering and focus. Nothing that we cannot manage to fix in some way, but I wouldn't invest time on it at this point | A bit better than the selected card, but it is a Beta component. See https://www.patternfly.org/v4/components/tile (note: I changed font-sizes through the developers console before take the screenshot)  |


`Tile` styling can be emulated with a `Card` using the _isSelectedRaise_ prop, in Beta too.

service/lib/dinstaller/storage/manager.rb
imobachgs: In general, I prefer the callback approach for the same reason that @joseivanlopez pointed out: decoupling pieces. For instance, the software layer *does not depend on the manager* to do its work. Actually, now that we have moved the software logic to a separate service, I would expect to just raise a signal and forget about the rest.

Of course, I know that this PR is not based on the *multi-process* branch but on *master*. So unless you plan to keep working on this, I would like to adapt these changes and continue working from there on top of *multi-process* changes.

service/lib/dinstaller/manager.rb
```diff
@@ -133,7 +133,11 @@ def install
     #
     # @return [Software]
     def software
-      @software ||= DBus::Clients::Software.new
+      return @software if @software
+
+      @software = DBus::Clients::Software.new
+      @software.on_product_selected { probe }
```
joseivanlopez, line 139: Cool! I think the *software* command should configure this callback too.

joseivanlopez: Something like:

~~~ruby
software_client.on_product_selected { manager_client.probe } 
~~~

imobachgs: I prefer to just raise a signal instead of calling back the manager, so the software component is as decoupled as possible from the manager service.

joseivanlopez: You are totally right! The signal is already emitted and the manager service reacts to it. Thanks!

joseivanlopez (approved):
LGTM

PR #195, opened 2022-06-13
CLI: add config command

joseivanlopez:
## Problem

In https://github.com/yast/d-installer/pull/179, the command *dinstallerctl install* was extended to optionally receive a config YAML file with the configuration for the installation (language, product, etc). During the review meeting, it was suggested to allow loading the configuration in a separte command. This would make possible to adapt some settings after loading the configuration. For example, if you have a common *config.yaml* file for several machines but you want to modify the language in a specific machine. 

* https://github.com/yast/d-installer/issues/193

## Solution

Provide a new command *dinstallerctl config* which allows to load and to dump the YAML configuration.

~~~bash
dinstallerctl config load config.yaml
dinstallerctl language selected es_ES
dinstallerctl config dump
dinstallerctl install
~~~

## Testing

- Added new unit tests
- Tested manually

### DIFF
--- a/cli/lib/dinstaller_cli/commands/config.rb
+++ b/cli/lib/dinstaller_cli/commands/config.rb
@@ -39,7 +39,7 @@ module DInstallerCli
         say_error("error: invalid configuration")
       end
 
-      desc "dump", "Dump the current installation config"
+      desc "dump", "Dump the current installation config to stdout"
       def dump
         say(current_config.dump)
       end
--- a/cli/lib/dinstaller_cli/commands/config.rb
+++ b/cli/lib/dinstaller_cli/commands/config.rb
@@ -1,0 +1,150 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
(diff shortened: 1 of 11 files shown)


joseivanlopez:
NOTE: CI task for jsdoc is failing when running from a fork.

cli/lib/dinstaller_cli/commands/config.rb
```diff
@@ -0,0 +1,150 @@
...
+    class Config < Thor
+      desc "load <config>", "Load a config file and apply the configuration"
+      def load(config_source)
+        config = InstallConfigReader.new(config_source).read
+        configure_installation(config)
+      rescue DInstallerCli::InstallConfigReader::Error
+        say_error("error: invalid configuration")
+      end
+
+      desc "dump", "Dump the current installation config"
```
mvidner, line 42:
```suggestion
      desc "dump", "Dump the current installation config to stdout"
```

cli/lib/dinstaller_cli/install_config.rb
```diff
@@ -55,6 +57,28 @@ def initialize
...
+        "product"   => product,
+        "languages" => languages,
+        "disks"     => disks,
+        "user"      => {
+          "name"      => user&.name,
+          "fullname"  => user&.fullname,
+          "autologin" => user&.autologin,
+          "password"  => user&.password
+        },
+        "root"      => {
```
mvidner, line 75: This couples the top `InstallConfig` class too tightly to its components. I think `InstallConfig::Root` should have its own `to_h` and then here:

```suggestion
        "root"      => root&.to_h || {},
```

and the same for `User` of course

mvidner (commented):
I see code gets moved across files.
That's a thing I dislike about git, I never know whether it was moved as is, or with some further changes.
@joseivanlopez can you please point out if you changed the code after moving it?

(For my own code moves, if I remember to do it, I prefer to commit the exact move, even with syntax errors, and then do an explicit fixup commit.)

joseivanlopez:
> I see code gets moved across files. That's a thing I dislike about git, I never know whether it was moved as is, or with some further changes. @joseivanlopez can you please point out if you changed the code after moving it?

The moved code and tests were not modified, only moved. I would take it into account next time for making the review easier. Sorry.

mvidner:
> I would take it into account next time for making the review easier. Sorry.

Don't be sorry, it's git's fault :)

mvidner (approved):
Thanks!

PR #196, opened 2022-06-14
Updated package-lock.json

teclator:
## Problem

Just update the package-lock.json file as otherwise the build of the package will fail.


dgdavid (approved):
Thanks @teclator, let's merge it.

PR #198, opened 2022-06-15
[RFC] D-Bus service is responsive during probing and installation

imobachgs:
## Problem

When a time-consuming task is running (like *probing*  or *installation*), the D-Bus service does not dispatch any other call. Hence you cannot ask for the current status, progress or any other value. The UI gets the status/progress information from signals.

We could mitigate this problem by running time-consuming tasks in a separate thread. However, as YaST is not thread-safe, we might face some problems in the future if conflicting tasks run simultaneously.

## Solution

We have introduced a new class (`TaskRunner`) that takes care of running one, and only one, of those potentially conflicting tasks at a time. In case you try to run more than one of those tasks simultaneously, the second one is rejected with a DBus error.

## To do

- [ ] Identify potentially conflicting tasks.
- [ ] Adapt the UI not to wait for signals to get the status.
- [ ] Testing

## Related work

This PR adds some responsiveness to the D-Bus service. Additionally, we are working on a multi-process approach to get some parallelism.

### DIFF
--- a/service/lib/dinstaller/dbus/manager.rb
+++ b/service/lib/dinstaller/dbus/manager.rb
@@ -20,6 +20,7 @@
 # find current contact information at www.suse.com.
 
 require "dbus"
+require "dinstaller/dbus/task_runner"
 
 module DInstaller
   module DBus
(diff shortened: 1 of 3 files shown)


imobachgs:
I am closing this pull request. We can use it as a reference if we decide to continue with this (or a similar) approach in the future. Thanks!

PR #201, opened 2022-06-16
Move the software API to a separate service `org.opensuse.DInstaller.Software`

imobachgs:
## Problem

This PRs moves the software management logic to a separate service, `org.opensuse.DInstaller.Software`. See #168 for further information.

## Solution

This is the first PR to move the software API to a separate `org.opensuse.DInstaller.Software`. It sets the basic infrastructure:

* Move the software API to a separate `SoftwareService`.
* Extend `bin/d-installer` to run all services at once or just one of them. More work is expected to improve this executable soon.
* Support for using a different `Y2DIR` per service.
* Adapt the UI to use the new software service.

## Testing

- [ ] *Added a new unit test*
- [x] *Tested manually*

## Tasks (other PRs)

- Keep track of the status
- API to add resolvables to the packages proposal (if needed)

PR #202, opened 2022-06-17
Software proposal

imobachgs:
Follow-up of #201.

* Add a D-Bus API to add/get/remove package proposals resolvables (`/org/opensuse/DInstaller/Software/Proposal1`).
* Add replacement modules for `Package`, `PackageProposal` and `SpaceCalculation`.

## Open questions

* We need to add support for `Package#Available` and `Package#Installed` to D-Bus. Where? Should we merge all the calls in a single D-Bus object? Or should we add another one in addition to the `Proposal1` object?


service/lib/dinstaller/dbus/y2dir/manager/modules/PackagesProposal.rb
```diff
@@ -0,0 +1,64 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+require "dinstaller/dbus/clients/software"
+
+# :nodoc:
+module Yast
+  # Replacement for the Yast::PackagesProposal module
+  class PackagesProposalClass < Module
+    include Yast::Logger
```
dgdavid, line 27: NP: leftover?

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -0,0 +1,135 @@
...
+# find current contact information at www.suse.com.
+
+require "dbus"
+
+module DInstaller
+  module DBus
+    module Software
+      # D-Bus object to manage software installation
+      class Manager < ::DBus::Object
+        PATH = "/org/opensuse/DInstaller/Software1"
```
joseivanlopez, line 29: Shouldn't this be `/org/opensuse/DInstaller/Software/Manager1`? Similar to we did with storage.

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -0,0 +1,135 @@
...
+            backend.finish
+          end
+
+          # Progress has struct with values:
+          #   s message
+          #   t total major steps to do
+          #   t current major step (0-based)
+          #   t total minor steps. Can be zero which means no minor steps
+          #   t current minor step
+          dbus_reader :progress, "(stttt)"
```
joseivanlopez, line 94: Similar to status, we would need a common interface for progress. Maybe with a mixin.

service/lib/dinstaller/dbus/software/proposal.rb
```diff
@@ -0,0 +1,81 @@
...
+require "yast"
+require "dbus"
+Yast.import "PackagesProposal"
+
+module DInstaller
+  module DBus
+    module Software
+      # Software proposal D-Bus representation
+      #
+      # This class allows to change sofware proposal settings through D-Bus.
```
joseivanlopez, line 31: Typo: software

service/lib/dinstaller/dbus/y2dir/manager/modules/Package.rb
```diff
@@ -0,0 +1,50 @@
...
+
+# :nodoc:
+module Yast
+  # Replacement for the Yast::Package module
+  #
+  # @see https://github.com/yast/yast-yast2/blob/b8cd178b7f341f6e3438782cb703f4a3ab0529ed/library/packages/src/modules/Package.rb
+  class PackageClass < Module
+    def main
+      puts "Loading mocked module #{__FILE__}"
+      @client = DInstaller::DBus::Clients::Software.new
```
joseivanlopez, line 30: This file shoud require this client.

joseivanlopez (commented):
I see we still have a `DBus::Clients::DInstaller` providing `#provision_selected?` and `#provision_selected?` methods. I think we can drop that service.

service/lib/dinstaller/dbus/software/manager.rb
imobachgs: I do not see the need to have a `manager` path at all. For instance, we have `/org/opensuse/DInstaller/Language1`. I am not against it, though, but we should define what we prefer.

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -0,0 +1,135 @@
...
+
+          register_progress_callback
+
+          super(PATH)
+        end
+
+        # rubocop:disable Metrics/BlockLength
+        dbus_interface SOFTWARE_INTERFACE do
+          dbus_reader :available_base_products, "a(ssa{sv})"
+          attr_writer :available_base_products
```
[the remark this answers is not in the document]
imobachgs: I do not know. Just moved from the original file.

service/lib/dinstaller/dbus/y2dir/manager/modules/Package.rb
imobachgs: I have decided to drop this change. Why? Well, because when it is used by storage we have not read the repositories yet. By now, let's consider that the package is always available.

service/lib/dinstaller/dbus/software/manager.rb
joseivanlopez: I am not sure, but I would try to follow the same approach for the different services. So, for storage service, I would expect something like:

* `/org/opensuse/DInstaller/Storage1`
* `/org/opensuse/DInstaller/Storage/Proposal1`
* `/org/opensuse/DInstaller/Storage/Actions1`

imobachgs: You are right. I will get rid of that client.

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -0,0 +1,135 @@
...
+          register_progress_callback
+
+          super(PATH)
+        end
+
+        # rubocop:disable Metrics/BlockLength
+        dbus_interface SOFTWARE_INTERFACE do
+          dbus_reader :available_base_products, "a(ssa{sv})"
+
+          dbus_watcher :available_base_products
```
joseivanlopez, line 52: I would say the watcher is not needed.

joseivanlopez (approved):
LGTM

PR #209, opened 2022-06-21
Installation phases

joseivanlopez:
## Problem

Our D-Bus services currently have an status (error, probing, probed, installing, installed). But that information is not that useful for clients. Note that the manager service can call to probe, to propose, etc, for each service at different times. For example, probing languages could be required at the very beginning, but probing software at the very end when the installation starts. That means that the current status used for the services stands for basically nothing. We need a more convenient way of expresing the status of the installation process and the status of the different services.

* https://trello.com/c/7a24Yxog/2987-5-d-installer-define-the-new-status-handling-workflow

## Solution

The *Installation Phase* concept was introduced. The manager service orchestrates the whole installation process, and it points to a specific installation phase value. The manager can be in one of these three installation phases: *startup*, *config* or *install*. Check the documentation at *doc/dbus_api.md* for more details about the meaning of each installation phase value. Note that an installation phase does imply nothing about the rest of services. The manager service decides what to do with each service in each installation phase. For example, manager could decide to probe languages during startup phase, and to probe storage during the config phase.

Now, each D-Bus service has a status to indicate whether the service is idle or busy. Normally, a service would switch from idle to busy when it starts a long term D-Bus action. Note that at this moment, a service is blocked meanwhile it executes a long task. This makes the status info useless, because the service does not respond until it finishes its long task. Because of that, the manager service registers the status of the rest of services as well. So clients can directly ask to manager about the status of the services.

This PR includes many other improvements:

* A new API for the `Progress` class.
* Mixins for common D-Bus APIs (`Progress` and `ServiceStatus`).

## Testing

* Added new unit tests
* Tested manually


doc/dbus_api.md
```diff
@@ -288,3 +288,37 @@ when the question is answered.
...
+## Status
+
+Each service exports an *status* object, for example `/org/opensuse/DInstaller/Software/Status1`.
+
+### org.opensuse.DInstaller.Status1
+
+#### Properties
+
+- Id -> unsigned 32-bit integer (r)
+  Id of the status (0 error, 1 probing, etc).
```
jreidinger, line 302: Just here I expect extensive list of enums. It is API documentation, so all values should be specified here.

joseivanlopez:
For the `Progress` "backend", I am thinking in a new API. The current API looks like:

~~~
progress = Progress.new
progress.on_chage { puts progress.message }
progress.init_progress(3, "Doing step1")    # steps 0/3
step1
progress.next_step("Doing step2")           # steps 1/3
step2
progress.next_step("Doing step3")           # steps 2/3
step3
progress.next_step("Finished action")       # steps 3/3, progress is done
~~~

And I woul like to have something like:

~~~
progress = Progress.new(3)
progress.on_chage { puts progress.message }
progress.on_finish { puts "finished"  }

progress.step("Doing step1") { step1 }
progress.step("Doing step2") { step2 }
progress.step("Doing step3") { step3 }
~~~

doc/dbus_api.md
```diff
@@ -288,3 +288,42 @@ when the question is answered.
...
+
+## Status
+
+The main object of a service implements the following interface:
+
+### org.opensuse.DInstaller.Status1
+
+#### Properties
+
+- All -> array(dict(string, variant)) (r)
```
imobachgs, line 301: Perhaps "All" is too generic. However, I would say we are safe, as all `org.freedsktop.DBus.Properties` methods get an interface name as the first argument, so it should not be a problem. I wonder, however, how D-Feet is going to show that property.

imobachgs: Oh, as @joseivanlopez pointed out the properties are shown under their own interface. So that's fine.

doc/dbus_api.md
```diff
@@ -288,3 +288,42 @@ when the question is answered.
...
+  All possible statuses:
+  [
+    {"id" => 0, "label" => "error"},
+    {"id" => 1, "label" => "probing"},
+    {"id" => 2, "label" => "probed"},
+    {"id" => 3, "label" => "installing"},
+    {"id" => 4, "label" => "installed"},
+  ]
```
jreidinger, line 309: maybe we should also mention when service is busy...so in probing and installing service is blocked ( aka busy ). In other states not...maybe there should be boolean for it?

jreidinger: well, even if it won't be blocked, still I think for UI it will be interesting info that service doing some long term job and maybe should display progress in such case?

joseivanlopez: Ok, let's evaluate it as part of https://trello.com/c/7a24Yxog/2987-d-installer-define-the-new-status-handling-workflow.

doc/dbus_api.md
```diff
@@ -285,6 +285,113 @@ when the question is answered.
...
+This is the initial phase. The manager service will start in this phase and it will not change to
+another phase until the client asks for performing the next phase.
+
+* *Config* Phase
+
+The installation is configured during this phase. Configuring the installation means that everything
+needed from the system is read and the required default proposal are calculated. In YaST terms, the
+*config* phase implies to probe some modules like storage, language, etc, and to perform their
+proposals. Note that not all modules have to be probed/proposed. Probing some modules could be
+delayed to the next *install* phase.
```
imobachgs, line 359: Could probing languages/network happen during the Startup phase? just to clarify.

doc/dbus_api.md
```diff
@@ -285,6 +285,113 @@ when the question is answered.
...
+### org.opensuse.DInstaller.Manager1
+
+#### Properties
+
+- InstallationPhases -> array(array(dict(string, variant))) (r)
+
+  All possible phases:
+~~~
+  [
+    {"id" => 0, "label" => "none"},
```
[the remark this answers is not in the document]
joseivanlopez: Discussed and removed.

doc/dbus_api.md
joseivanlopez: Yes, sure. We can move to the startup phase everything we consider. The idea with the config phase is that you could "click" on install after the manager has done the config phase, that is, the manager has finished to configure/calculate everything that could be modified by a client.

service/lib/dinstaller/manager.rb
```diff
@@ -43,106 +44,115 @@ module DInstaller
...
-    # @return [Software]
+    # @return [DBus::Clients::Software]
     def software
-      @software ||= DBus::Clients::Software.new
+      @software ||= DBus::Clients::Software.new.tap do |client|
+        client.on_product_selected do |selected|
+          config.pick_product(selected)
+          config_phase
+        end
+        client.on_service_status_change do |status|
```
imobachgs, line 152: I have discovered that having two calls to `on_signal` will not work. The [second one](https://github.com/joseivanlopez/d-installer/blob/06ec911c9003f16a51c78c0bb950a1b5f7608ca1/service/lib/dinstaller/dbus/clients/with_service_status.rb#L50) will overwrite [the first one](https://github.com/joseivanlopez/d-installer/blob/06ec911c9003f16a51c78c0bb950a1b5f7608ca1/service/lib/dinstaller/dbus/clients/software.rb#L164). So we might need to implement this in a slightly different way (registering the callbacks and listening just once).

imobachgs (commented):
Really good work, congrats and thank you :-) I have a few notes and I will have a closer look at the tests, but everything looks quite well.

imobachgs (approved):

PR #210, opened 2022-06-21
adapt to new ruby-dbus

jreidinger:
## Problem

new partitioning is now shown after product is changed

## Solution

Adapt to new ruby dbus and test that it fixes issue.

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/dinstaller/dbus/manager.rb
+++ b/service/lib/dinstaller/dbus/manager.rb
@@ -65,7 +65,8 @@ module DInstaller
         dbus_reader :status, "u"
 
         # Error messages when the status is 0
-        dbus_reader :error_messages, "as"
+        dbus_reader :error_messages, "as",
+          emits_changed_signal: ::DBus::EmitsChangedSignal.new(:invalidates)
 
         # Progress has struct with values:
         #   s message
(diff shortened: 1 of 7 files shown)


imobachgs (approved):

service/lib/dinstaller/dbus/manager.rb
```diff
@@ -65,7 +65,8 @@ def initialize(backend, logger)
         dbus_reader :status, "u"
 
         # Error messages when the status is 0
-        dbus_reader :error_messages, "as"
+        dbus_reader :error_messages, "as",
+          emits_changed_signal: ::DBus::EmitsChangedSignal.new(:invalidates)
```
mvidner, line 69: @jreidinger 

```rb
 dbus_reader :error_messages, "as", emits_changed_signal: :invalidates
```

is the intended usage

mvidner: ... and your way will `raise` as soon as the file is parsed

mvidner: ... actually it will not because a `dbus_reader` *without* an assignment method (`error_messages=`) means no `dbus_watcher` is created and the `e_c_s:` option is never used.

PR #211, opened 2022-06-22
Add a product selection component

dgdavid:
## Problem

As stated in https://github.com/yast/d-installer/issues/200, a preliminary step/screen for selecting the desired product is wanted.

## Solution

* Added [react-router](https://reactrouter.com/) to have proper routing. We can have a separate page for product selection and, in the future, for storage, users, and so on.
* Service:
  - Do not pre-select any product if many are available.
  - Postpone the repository probing until the installation starts. At this point, we are no reading anything from the repositories.
* UI:
  - Get rid of the product section on the overview page (WIP).
  - Redirect to the new product selection page if no product is selected.
  - Add a `context/software` to keep the list of products and the selected one. The reason is that several components need to access this information, and passing through `useOutletContext` did not look right.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

<details>
  <summary>Multi-product scenario</summary>

Please, ignore the "Target" section because the machine where I recorded the demo did not have an available device for installation. The "Probing" page should disappear when we adapt the code to the new status API.

multi-product
</details>

<details>
  <summary>Single-product scenario</summary>

The selector is not visible, and the user does not go through the product selection page.

single-product

</details>


web/src/ProductSelection.jsx
```diff
@@ -0,0 +1,121 @@
...
+  const ContinueButton = () => {
+    return (
+      <Button isLarge variant="primary" onClick={accept}>
+        Continue
+      </Button>
+    );
+  };
+
+  const buildOptions = () => {
+    const options = products.map((p) => (
```
jreidinger, line 90: One feature that I think is a must have is to skip this selection if there is just one product. I expect it will be common case when customer use his own handcrafted iso for install or wants to have it as PXE entries. Also maybe for some products  we want specialized installers.....Of course second question is how to react with system roles, but it is topic for future.

dgdavid: Of course, this component/screen will not be shown when there is only one product.

web/src/Main.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+  it("renders the Questions component and the content", async () => {
+    installerRender(<Main />);
+
+    await screen.findByText("Questions Mock");
+    await screen.findByText("Content");
+  });
+});
+
+describe("when a product is selected", () => {
```
dgdavid, line 63: Did you mean _when a product **is not** selected_?

web/src/Main.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+  beforeEach(() => {
+    product = null;
+  });
+
+  it("renders the Questions component and the content", async () => {
```
dgdavid, line 68: Actually, it _renders a Navigate component_

dgdavid: And what about not mocking `<Navigate /> ` but `<ProductSelection />` and testing that it _redirects to the product selection screen_ instead?

web/src/Layout.jsx
```diff
@@ -66,6 +66,17 @@ function Layout({ MenuIcon, sectionTitle, SectionIcon, FooterActions, children }
     );
   };
 
+  const renderHeaderRightActions = () => {
```
dgdavid, line 69: NOTE: I know it was me who choose those names back in the time. But they (_LeftAction_, _RightActions_) does not look good so good. In fact, _LeftAction_ is not in use yet, but it could became in something like _MainMenu_, _AppMenu_ or simply _Menu_ (have I heard _HamburguerMenu_? :stuck_out_tongue_closed_eyes:? On the other hand, _RightActions_ are more like _PageActions_ or _SectionActions_, which is not completely true for the `Overview` page which has the _Install_ action in the footer.

Handsdown, these small things need some love. The Layout in general needs more love now that we somehow introduced the concept of pages/sections for real.

Something for another PR, of course.

web/src/App.jsx
```diff
@@ -80,6 +66,8 @@ function App() {
 
   useEffect(() => {
     return client.manager.onChange(changes => {
+      console.log("manager has changed", changes);
```
dgdavid, line 69: Left over?

web/src/LoadingEnvironment.jsx
```diff
@@ -27,14 +27,14 @@ import Center from "./Center";
 
 import { EOS_THREE_DOTS_LOADING_ANIMATED as LoadingIcon } from "eos-icons-react";
 
-function LoadingEnvironment() {
+function LoadingEnvironment({ text = "Loading installation environment, please wait." }) {
```
dgdavid, line 30: NP: maybe we can rename this component to _Loading_

web/src/Overview.jsx
```diff
@@ -100,11 +100,28 @@ function Overview() {
...
+      <Button
+        isSmall
+        variant="plain"
+        icon={<ModeEditIcon />}
+        aria-label="Change selected product"
+        onClick={() => navigate("products")}
+      />
+    );
+  };
+
```
dgdavid, line 118: NP: Now that we have _useSoftware_, we could extract this out of the Overview component. I mean, similar to what was done with InstallButton.

web/src/Overview.test.jsx
```diff
@@ -80,9 +95,29 @@ beforeEach(() => {
   });
 });
 
+// FIXME: it must not display the action if there is only one product
```
dgdavid, line 98: Leftover

web/src/ProductSelectionPage.jsx
```diff
@@ -0,0 +1,124 @@
...
+    return (
+      <Layout
+        sectionTitle="D-Installer"
+        SectionIcon={SectionIcon}
+        FooterActions={SelectButton}
+      >
+        <Center>Loading available products...</Center>
+      </Layout>
+    );
+  }
```
dgdavid, line 88: I guess this can be changed. What about

```suggestion
  if (!products) return <LoadingEnvironment text="Loading available products, please wait..." />
```

Of course, it's needed to import the LoadingEnvironment component (or simply Loading if we endup renaming it)

dgdavid: Note that makes no sense to show the "SelectButton" action while "Loading available products"

dgdavid (commented):
Just few notes from my side, but in general LGTM. Not sure if we need a review from someone else, or we can just continue from here. Anyway, I can't approve it because I'm who created the PR initially.

web/src/Layout.jsx
imobachgs: Yes, we should find a better solution for the layout because, at this point, we are having some trouble with the footer. However, I would expect those problems to disappear as soon as we merge the new status API.

web/src/LoadingEnvironment.jsx
imobachgs: Let's do it as part of our changes related to the status API.

dgdavid (commented):
LGTM

PR #217, opened 2022-07-11
Move `<Questions>` component to `<App>`

dgdavid:
Since a question could arrive at any moment (except when the connection with D-Bus is lost), there is no reason for having the `Questions` component not mounted when the user is selecting a product or accepting a license, for example.

Thus, moving it from `Main` to `App` ensures it is almost always there, apart from avoiding some React complains because of trying to update an unmounted component since it is not removing the subscriptions when it will be unmounting. Maybe a bug, let's discuss it.

### DIFF
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -31,6 +31,7 @@ import ProbingProgress from "./ProbingProgress";
 import InstallationProgress from "./InstallationProgress";
 import InstallationFinished from "./InstallationFinished";
 import LoadingEnvironment from "./LoadingEnvironment";
+import Questions from "./Questions";
 
 import { PROBING, PROBED, INSTALLING, INSTALLED } from "./client/status";
 
@@ -94,6 +95,7 @@ function App() {
   return (
     <Layout>
       <Title>D-Installer</Title>
+      <Questions />
       <Content />
       <AdditionalInfo>
         <About />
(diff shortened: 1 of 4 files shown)


dgdavid:
As commented, the component is not triggering the _unsubscription_ callbacks when it will be unmounted.

https://github.com/yast/d-installer/blob/c244b6d6a0d9f3ae72f4adf7da17805bf7205285/web/src/Questions.jsx#L51-L59


If I'm not wrong, it could be addressed doing something like

```diff
  useEffect(() => {
+   const unsubscriptionCallbacks = [];
+
    client.questions.getQuestions()
      .then(setPendingQuestions)
      .catch(e => console.error("Something went wrong retrieving pending questions", e));
-   client.questions.onQuestionAdded(addQuestion);
+   unsubscriptionCallbacks.push(client.questions.onQuestionAdded(addQuestion));
-   client.questions.onQuestionRemoved(removeQuestion);
+   unsubscriptionCallbacks.push(client.questions.onQuestionRemoved(removeQuestion));
+
+ return () => { unsubscriptionCallbacks.forEach(cb => cb()) };
  }, [client.questions, addQuestion, removeQuestion]);


  if (pendingQuestions.length === 0) return null;
```

Which, IMHO, it's compatible with the relocation of `<Questions>` from `<Main>` to `<App>`.

dgdavid:
As discussed offline, let's fix the subscriptions bug and keep the `<Questions>` component where it is by now.

Closing.

PR #218, opened 2022-07-11
Fix Questions component

dgdavid:
## Problem

`Questions` component is not canceling its subscriptions to D-Bus signals, which makes React complain because of a potential memory leak.

## Solution

Execute all unsubscription callbacks when the component is unmounted.

## Testing

- *Tested manually*


joseivanlopez (approved):
LGTM

PR #219, opened 2022-07-13
Adapt the web UI to the new status/progress API

imobachgs:
## Problem

Adapt the web UI to the new status and progress API (see #209).

## Solution

* Adapt several components to the new API.
* Drop the `ProbingProgress` component.
* Extend the D-Bus storage proposal object with status information.

## Testing

- Added and updated unit tests
- Tested manually (work in progress)


web/src/Storage.jsx
```diff
@@ -93,8 +99,40 @@ export default function Storage() {
...
+  }, [client.storage]);
+
+  // FIXME: this useEffect should be removed after moving storage to its own service.
+  useEffect(() => {
+    client.manager.getStatus().then(status => {
+      dispatch({ type: "CHANGE_STATUS", payload: status });
+    });
+  }, [client.manager]);
+
+  // FIXME: this useEffect should be removed after moving storage to its own service.
```
dgdavid, line 121: NP: maybe we should left a comment in the _backend_ part as a reminder...

web/src/client/mixins.js
```diff
@@ -87,6 +71,75 @@ const withDBus = {
...
+      return {
+        total:    proxy.TotalSteps,
+        current:  proxy.CurrentStep[0],
+        message:  proxy.CurrentStep[1],
+        finished: proxy.Finished
+      };
+    },
+
+    /**
+     * Register a callback to run when the status changes
```
dgdavid, line 125:
```suggestion
     * Register a callback to run when the progress changes
```

dgdavid (approved):
LGTM

web/src/Storage.jsx
imobachgs: Actually, we do not need to do any changes in the backend related to this code. I would keep the comment just here.

web/src/InstallerSkeleton.jsx
```diff
@@ -20,24 +20,24 @@
...
+
+const MAX = 100;
+
+const InstallerSkeleton = ({ lines = 2, ...props }) => (
+  [...Array(lines)].map((_, i) => {
+    const width = Math.floor(Math.random() * MAX);
+    return (
+      <Skeleton
+        className="installer-skeleton"
+        key={i}
```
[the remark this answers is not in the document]
imobachgs: Good point.

web/src/Overview.jsx
```diff
@@ -102,7 +102,7 @@ function Overview() {
     <Category key="language" title="Language" icon={LanguagesSelectionIcon}>
       <LanguageSelector />
     </Category>,
-    <Category key="targer" title="Target" icon={HardDriveIcon}>
+    <Category key="target" title="Storage" icon={HardDriveIcon}>
```
dgdavid, line 105: :+1: 

I like it more. Maybe we can change the `key` value too.

PR #221, opened 2022-07-14
Small UI fixes

imobachgs:
## Problem

Some React components where not correctly unmounted after selecting the product for the first time. The problem was that the components were redrawn because the route AND the phase changed at the same time.

## Solution

Move the phase handling one level below, to the Main component.

## Testing

- *Added a new unit test*
- *Tested manually*


joseivanlopez (approved):
LGTM

PR #222, opened 2022-07-15
Install LSM patterns

imobachgs:
## Problem

Required patterns for the default LSM are not selected, hence the system might not be usable (for instance, Leap Micro will not boot).

## Solution

Add the needed patterns required for the "selected" LSM patterns.

## Testing

- Added a new unit test
- Tested manually


joseivanlopez (approved):
LGTM

PR #231, opened 2022-07-22
Questions in a separate service/process

mvidner:
## Problem

The user-facing callbacks, aka **Questions** ([dbus-api](https://github.com/yast/d-installer/blob/master/doc/dbus_api.md#questions)) have so far been limited to a single process/service. However, a major use case are the packager callbacks, and there is a separate **o.o.DInstaller.Software** service.

All the questions should live in one service. Originally we thought that it should be a separate process hosting *only* the questions, but so far that hasn't been necessary, the main service is able to host them.

At least a D-Bus call to *add a new question* is needed.

- https://trello.com/c/JY14vDza

## Solution

The main idea of the solution is: keep the existing `CanAskQuestions` mixin, and let it be used the same way, with the questions service being either in the same process or not.

This is achieved by providing `def questions_manager { DInstaller::DBus::Clients::QuestionsManager.new }` with the same API as `DInstaller::QuestionsManager`.

## Testing

- [x] Added new unit tests
- [x] Tested manually

## Screenshots

(No new UI)

### DIFF
--- a/service/lib/dinstaller/dbus/clients/questions_manager.rb
+++ b/service/lib/dinstaller/dbus/clients/questions_manager.rb
@@ -48,9 +48,11 @@ module DInstaller
         # @yieldparam question [Question] added question
         #
         # @param question [Question]
-        # @return [Boolean] whether the question was added
+        # @return [Boolean] whether the question was added FIXME
         def add(question)
-          @dbus_object.New(question.text, question.options.map(&:to_s), Array(question.default_option&.to_s))
+          q_path = @dbus_object.New(question.text, question.options.map(&:to_s), Array(question.default_option&.to_s))
+          p q_path
+          service[q_path]
           # question identity, the api does not really fit, New really should return an obj path
           # or whatever
         end
@@ -61,28 +63,31 @@ module DInstaller
         #
         # @yieldparam question [Question] deleted question
         #
-        # @param question [Question]
-        # @return [Boolean] whether the question was deleted
+        # @param question [Question] FIXME the yardocs are wrong here
         def delete(question)
-          return false unless include?(question)
-
-          questions.delete(question)
-          on_delete_callbacks.each { |c| c.call(question) }
-
-          true
+          @dbus_object.Delete(question.path)
         end
 
         # Waits until all questions are answered
         #
         # Callbacks are periodically called while waiting, see {#on_wait}.
-        def wait
-          logger.info "Waiting for questions to be answered"
+        def wait(questions)
+          puts "In Wait"
+          # so what is the minimum we must do? wait should receive a list of questions to wait for (object paths) and ignore the others
+          # register the InterfacesAdded callback...
+          # stupid but simple way: poll the answered property of the questions.first object, sleep 0.5 s, repeat
+          #
+          question = questions.first # FIXME: use them all
+          q_proxy_iface = question["org.opensuse.DInstaller.Question1"]
+          10.times do
+            puts "asking answer"
+            answer = q_proxy_iface["Answer"]
+            puts "  it is #{answer.inspect}"
+            break unless answer.empty?
 
-          loop do
-            on_wait_callbacks.each(&:call)
-            sleep(0.1)
-            break if questions_answered?
+            sleep(0.5)
           end
+          question.define_singleton_method(:answer) { q_proxy_iface["Answer"] }
         end
 
         # Registers a callback to be called while waiting for questions be answered
@@ -93,9 +98,6 @@ module DInstaller
         end
 
       private
-
-        # @return [Logger]
-        attr_reader :logger
 
         # Callbacks to be called when waiting for answers
         #
--- a/service/lib/dinstaller/dbus/clients/questions_manager.rb
+++ b/service/lib/dinstaller/dbus/clients/questions_manager.rb
@@ -20,7 +20,7 @@
 # find current contact information at www.suse.com.
 
 require "dinstaller/dbus/clients/base"
-require "dinstaller/dbus/clients/with_service_status"
+require "dinstaller/dbus/clients/question"
 
 module DInstaller
   module DBus
@@ -45,7 +45,7 @@ module DInstaller
         #
         # FIXME: I don't see a need for callbacks here. Keep them anyway?
         #
-        # @param question [Question]
+        # @param question [DInstaller::Question]
         # @return [DBus::Clients::Question] FIXME: I don't understand the concept of not asking a duplicate question
         def add(question)
           q_path = @dbus_object.New(
@@ -53,9 +53,7 @@ module DInstaller
             question.options.map(&:to_s),
             Array(question.default_option&.to_s)
           )
-          # FIXME: we're using ::DBus::ProxyObject
-          # but there should be a DInstaller::DBus::Clients::Question to wrap it
-          service[q_path]
+          DBus::Clients::Question.new(q_path)
         end
 
         # Deletes the given question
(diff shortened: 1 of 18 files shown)


mvidner:
One thing that is beyond me: when I put the Software service `testing_question` in the product selection, no UI showed to present it to the user. It only started working when I triggered it (with another testing D-Bus call) at the main installation summary.

Now that I think of it, probably the product selection react component needs to include the questions component? Haven't checked yet.

imobachgs:
[..]

> Now that I think of it, probably the product selection react component needs to include the questions component? Haven't checked yet.

Yes, we decided to not put the questions on the product selection page because we did not want other questions (like decrypting storage devices and so on) to come on top of the product selection. We consider product selection the very first step.

Of course, as everything related to D-Installer, we can discuss and decide something different. :-)

mvidner:
`DInstaller::QuestionsManager` has instance methods [`add`][add] and `delete` which can fail.
`delete` obviously fails when its argument has not been `add`ed previously, that's fine
`add`fails when the added question is "duplicate" but there are 2 problems with this
1. [`CanAskQuestion#ask`][ask] does not care about the failure
2. the criterion for duplication is `Question#id` but this `id` is unique for each Question instance, even with the same text+answers, or even if you happened to use the same `add` call site repeatedly

I don't understand what scenario `add` is trying to guard against.

Counterpoint: `add` should always succeed.

@joseivanlopez as you designed this, what do you think?

[add]: https://github.com/joseivanlopez/d-installer/blob/d34fdd24dfacafd01cd57884935445cced4ba3a2/service/lib/dinstaller/questions_manager.rb#L50
[ask]: https://github.com/joseivanlopez/d-installer/blob/de979b25937f78220e5db803d330f116f82a0de4/service/lib/dinstaller/can_ask_question.rb#L39

service/lib/dinstaller/dbus/clients/questions_manager.rb
```diff
@@ -0,0 +1,96 @@
...
+        def wait(questions)
+          # TODO: detect if no UI showed up to display the questions and time out?
+          # for example:
+          # (0..Float::INFINITY).each { |i| break if i > 100 && !question.displayed; ... }
+
+          # We should register the InterfacesAdded callback... BEFORE adding to avoid races.
+          # Stupid but simple way: poll the answer property, sleep, repeat
+          loop do
+            questions = questions.find_all { |q| !q.answered? }
+            break if questions.empty?
```
jreidinger, line 83: why not `break if questions.all?(&:answered?)`

service/lib/dinstaller/dbus/clients/software.rb
```diff
@@ -97,6 +97,10 @@ def finish
           dbus_object.Finish
         end
 
+        def testing_question
```
jreidinger, line 100: leftover of testing method?

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -78,6 +78,8 @@ def initialize(backend, logger)
           dbus_method(:Propose) { propose }
           dbus_method(:Install) { install }
           dbus_method(:Finish) { finish }
+
+          dbus_method(:TestingQuestion) { backend.testing_question }
```
jreidinger, line 82: same here?

service/lib/dinstaller/dbus/questions.rb
```diff
@@ -73,6 +76,31 @@ def managed_objects
...
+          )
+          backend.add(backend_q)
+          path_for(backend_q)
+        end
+
+        dbus_method :Delete, "in question:o" do |question_path|
+          dbus_q = @service.get_node(question_path)&.object
+          raise ArgumentError, "Object path #{question_path} not found" unless dbus_q
+          raise ArgumentError, "Object #{question_path} is not a Question" unless
+            dbus_q.is_a? DInstaller::DBus::Question
```
jreidinger, line 97: please do not use trailing if/unless when it is not single line

service/lib/dinstaller/manager.rb
```diff
@@ -83,12 +86,25 @@ def config_phase
       security.probe
       network.probe
 
+      if ENV["DINSTALLER_TEST_QUESTIONS"] == "1"
+        testing_question
+        software.testing_question
+      end
```
jreidinger, line 92: testing leftover?

service/lib/dinstaller/questions_manager.rb
```diff
@@ -63,26 +63,31 @@ def add(question)
...
     # Callbacks are periodically called while waiting, see {#on_wait}.
-    def wait
+    # @param questions [Array<Question>]
+    def wait(questions)
       logger.info "Waiting for questions to be answered"
 
       loop do
         on_wait_callbacks.each(&:call)
+        questions = questions.find_all { |q| !q.answered? }
+        break if questions.empty?
```
jreidinger, line 88: same here. I would use single line more readable `break if questions.all?(&:answered?)`

service/lib/dinstaller/software.rb
```diff
@@ -68,10 +72,23 @@ def select_product(name)
       return if name == @product
       raise ArgumentError unless @products[name]
 
+      # TODO: if a question is asked here, there is no web UI to handle it.
+      # Is it a problem?
+      # testing_question if ENV["DINSTALLER_TEST_QUESTIONS"] == "1"
```
jreidinger, line 77: testing code leftover?

service/test/dinstaller/manager_test.rb
```diff
@@ -84,6 +85,11 @@
   end
 
   describe "#config_phase" do
+    before do
+      allow(subject).to receive(:testing_question)
+      allow(software).to receive(:testing_question)
```
jreidinger, line 90: Well, now you get me. I am no longer sure if that testing question is testing leftover or not. But in general I do not like it :)

service/lib/dinstaller/dbus/clients/questions_manager.rb
mvidner: Because our D-Bus clients do not cache properties and this way we don't make calls about questions that we already know are answered.

service/lib/dinstaller/software.rb
mvidner: Yes, but a good opportunity to ask about the TODO:

What should happen when some code asks a Question and the web UI has no component to display it?
Currently the code deadlocks waiting for an answer that will not come.
Arguably it will only happen during development.
We could add a `displayed?` property to Questions and `raise` an error if a Question is not displayed within 20 seconds.

service/lib/dinstaller/dbus/clients/questions_manager.rb
jreidinger: so please document it. or maybe use better variable name.

mvidner:
With commit https://github.com/yast/d-installer/pull/231/commits/b8c197e050c2a5a7f71bc77395a6f6144409ee72 I am removing the testing questions. When we add some actual question, that code can serve as usage example of the API.

jreidinger (approved):

joseivanlopez:
> `DInstaller::QuestionsManager` has instance methods [`add`](https://github.com/joseivanlopez/d-installer/blob/d34fdd24dfacafd01cd57884935445cced4ba3a2/service/lib/dinstaller/questions_manager.rb#L50) and `delete` which can fail. `delete` obviously fails when its argument has not been `add`ed previously, that's fine `add`fails when the added question is "duplicate" but there are 2 problems with this
> 
> 1. [`CanAskQuestion#ask`](https://github.com/joseivanlopez/d-installer/blob/de979b25937f78220e5db803d330f116f82a0de4/service/lib/dinstaller/can_ask_question.rb#L39) does not care about the failure
> 2. the criterion for duplication is `Question#id` but this `id` is unique for each Question instance, even with the same text+answers, or even if you happened to use the same `add` call site repeatedly
> 
> I don't understand what scenario `add` is trying to guard against.

The only important thing is to avoid duplicate questions, which does not make sense. Failing "helps" to identify whether the question was already added. But, as you suggested, `#add` could simply success when the question already exists (without adding the question again).

> 
> Counterpoint: `add` should always succeed. (`DInstaller::DBus::Clients::QuestionsManager#add` already does)
> 
> @joseivanlopez as you designed this, what do you think?

PR #234, opened 2022-07-25
Dispatch message queue on progress in the software service

### LINKED ISSUE #233: Progress installation is not refreshed after reloading the UI
The packages installation progress does not get refreshed if the user reloads the UI. Most likely, the problem is that client tries to introspect the software service, but the service is busy. Listening for signals is fine, but trying to set up the connection is not.

In case we can confirm the problem, there are two solutions:

* Dispatch pending D-Bus calls after each package is installed.
* Avoid introspection for that use case.

imobachgs:
On progress, the software service dispatches the message queue. It fixes #233.

### DIFF
--- a/service/lib/dinstaller/dbus/software_service.rb
+++ b/service/lib/dinstaller/dbus/software_service.rb
@@ -42,6 +42,7 @@ module DInstaller
       # @param logger [Logger]
       def initialize(config, logger = nil)
         @backend = DInstaller::Software.new(config, logger)
+        @backend.on_progress_change { dispatch }
         @logger = logger || Logger.new($stdout)
         @bus = ::DBus::SystemBus.instance
       end
@@ -80,7 +81,6 @@ module DInstaller
       def dbus_software_manager
         return @dbus_software_manager if @dbus_software_manager
 
-        @backend.on_progress_change { dispatch }
         @dbus_software_manager = DInstaller::DBus::Software::Manager.new(@backend, logger)
         @dbus_software_manager
       end
--- a/service/lib/dinstaller/dbus/software_service.rb
+++ b/service/lib/dinstaller/dbus/software_service.rb
@@ -71,9 +71,18 @@ module DInstaller
       # @return [Array<::DBus::Object>]
       def dbus_objects
         @dbus_objects ||= [
-          DInstaller::DBus::Software::Manager.new(@backend, logger),
+          dbus_software_manager,
           DInstaller::DBus::Software::Proposal.new(logger)
         ]
+      end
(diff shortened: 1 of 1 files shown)


service/lib/dinstaller/dbus/software_service.rb
```diff
@@ -71,10 +71,19 @@ def service
...
+          dbus_software_manager,
           DInstaller::DBus::Software::Proposal.new(logger)
         ]
       end
+
+      # @return [DInstaller::DBus::Software::Manager]
+      def dbus_software_manager
+        return @dbus_software_manager if @dbus_software_manager
+
+        @backend.on_progress_change { dispatch }
```
joseivanlopez, line 83: NP: Maybe it would be better to configure this in the constructor, while instantiating the backend.

joseivanlopez: Another option is to define a `#start` method containing that kind of configurations, similar to manager service.

imobachgs: OK, it makes sense. After all, we are calling the `register_*` methods in the constructor.

service/lib/dinstaller/dbus/software_service.rb
```diff
@@ -42,6 +42,7 @@ class SoftwareService
       # @param logger [Logger]
       def initialize(config, logger = nil)
         @backend = DInstaller::Software.new(config, logger)
+        @backend.on_progress_change { dispatch }
```
joseivanlopez, line 45: BTW, shouln't logger be initialized before? I mean, I would try to use the very same logger for both, backend and service.

service/lib/dinstaller/dbus/software_service.rb
```diff
@@ -71,10 +72,18 @@ def service
...
         @dbus_objects ||= [
-          DInstaller::DBus::Software::Manager.new(@backend, logger),
+          dbus_software_manager,
           DInstaller::DBus::Software::Proposal.new(logger)
         ]
       end
+
+      # @return [DInstaller::DBus::Software::Manager]
+      def dbus_software_manager
+        return @dbus_software_manager if @dbus_software_manager
```
joseivanlopez, line 82: NP: I guess this could simply be `@dbus_software_manager ||= DInstaller::DBus::Software::Manager.new(backend, logger)`

imobachgs: Yes, you are right, it was a leftover.

joseivanlopez (approved):
LGTM

PR #236, opened 2022-07-26
Support nil value for resolvables list

imobachgs:
In the original `PackagesProposal` module, it is possible to [use `nil` as the `resolvables` parameter](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/PackagesProposal.rb#L122). So we should keep this behavior in the new module.


dgdavid (approved):
LGTM

PR #240, opened 2022-07-27
Move language packages selection to the software service

### LINKED ISSUE #239: Changing the locale causes the installation to fail
Changing the locale, causes libzypp to get blocked by the `org.opensuse.DInstaller` software (see [language.rb](https://github.com/yast/d-installer/blob/f5be795b1077ea3a3cbaf9501cc1d167d03479c7/service/lib/dinstaller/language.rb#L60)).

imobachgs:
See #239.

## Problem

When the user changes the language, the installation will not work. The problem is that, [accidentally](https://github.com/yast/d-installer/blob/master/service/lib/dinstaller/language.rb#L60), the manager will set up the packages database, acquiring the libzypp lock. Needless to say that, as a consequence, the software service will be unable to work with libzypp.

## Solution

* The language module does not initialize the packages database anymore.
* The software service listens for language changes and saves the value for later.
* When the installation starts, the software layer is responsible for setting the libzypp locales.
* As software needs to listen for language changes (in the main `org.opensuse.DInstaller` service), we got a chicken and egg problem (the manager also relies on `org.opensuse.DInstaller.Software`). Splitting the language handling to a separate service (`org.opensuse.DInstaller.Language`) fixed the issue. This new service will be responsible for everything related to yast2-country (we might consider renaming the service in the future).


service/test/dinstaller/software_test.rb
```diff
@@ -133,6 +133,12 @@
       subject.propose
     end
 
+    it "selects the language packagees" do
```
dgdavid, line 136: typo

```suggestion
    it "selects the language packages" do
```

dgdavid:
CI/Backend tests is complaining about

```
Errno::ENOENT:
        No such file or directory - connect(2) for /var/run/dbus/system_bus_socket
```

dgdavid (approved):
Pretty nice work, @imobachgs.

Thanks a lot!

mvidner:
Just a reminder, after merging this you need to propagate the updated D-Bus config on your testing machine:

```
sudo cp service/share/dbus.conf /usr/share/dbus-1/system.d/org.opensuse.DInstaller.conf
```

PR #244, opened 2022-08-02
Fix packaging

imobachgs:
Submission to Factory is not working because of the sources validator. See [bsc#992030](https://build.opensuse.org/request/show/992030) as an example.

* When a `obscpio` file is used, the validator assumes that the tarball will be generated at runtime [source code](https://github.com/openSUSE/obs-service-source_validator/blob/master/20-files-present-and-referenced#L136). That's fine, but the validator expects the file to have the same name. So we are keeping the same filename to not confuse the source validator.
* The changes file should be named `cockpit-d-installer.changes`.
* `node_modules.sums` is not included in the list of sources.


teclator (approved):
LGTM

teclator:
NP: It we are renaming everything to **d-installer** looks like the package name should be also **d-installer** instead of **cockpit-d-installer** :P or the other way around... rename the repository to **cockpit-d-installer**

imobachgs:
> NP: It we are renaming everything to **d-installer** looks like the package name should be also **d-installer** instead of **cockpit-d-installer** :P or the other way around... rename the repository to **cockpit-d-installer**

That's the point: if we rename the file, the validator will not work. So let's keep it as it is. After all, the name of the sources is not that important.

teclator (approved):
LGTM

PR #245, opened 2022-08-19
Object manager upstreamed

mvidner:
## Problem

The [ObjectManager interface for D-Bus services][om] was missing from ruby-dbus when we needed it so Ivan added it here. Now that the feature is upstreamed, redundant or conflicting code needs to be removed/adapted.

- #135 implemented here
- https://github.com/mvidner/ruby-dbus/pull/122 upstreamed

[om]: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager

## Solution

Somewhat amazingly, the solution is to add a single `include` and remove a bunch of code and it seems to work. And update the dependency version.

(There is some testing code which is added and then reverted)

## Testing

- [ ] ~~Added a new unit test~~; well, removed one here because the method is upstreamed
- [x] Re-tested manually; was surprised by an actual Question about trusting the ALP repo key :smile:

### DIFF
--- a/cli/Gemfile.lock
+++ b/cli/Gemfile.lock
@@ -9,7 +9,7 @@ PATH
       fast_gettext (~> 2.2.0)
       nokogiri (~> 1.13.1)
       rexml (~> 3.2.5)
-      ruby-dbus (~> 0.18.1)
+      ruby-dbus (~> 0.19.0)
 
 PATH
   remote: .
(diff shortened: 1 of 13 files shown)


mvidner:
BTW during testing the Storage component was crashing because of Issues that could not be displayed, and I worked around it with `env LIBSTORAGE_IGNORE_PROBE_ERRORS=1`

mvidner:
@joseivanlopez has noted that we can probably remove `exported_questions` and use one of the new upstream methods, probably [`descendant_objects`](https://www.rubydoc.info/gems/ruby-dbus/0.19.0/DBus/Node#descendant_objects-instance_method)

joseivanlopez (approved):
LGTM

PR #246, opened 2022-08-24
Better error reporting for the storage proposal

imobachgs:
## Problem

When D-Installer does not find any suitable storage device for installation, the user does not get any feedback.

## Solution

Display a better error when not suitable storage devices were found or when something else happened.

Bonus: display more information about the selected storage devices.

## Screenshots

Captura desde 2022-08-24 13-44-57


dgdavid (approved):
LGTM

Issue #247, opened 2022-08-26
No storage proposal when a probing error occurs


imobachgs:
When any issue is found during storage probing, D-Installer is not able to come up with a storage proposal. It might happen that blkid returns an error code (for instance, running on [Iguana](https://github.com/aaannz/dracut-iguana/)) but libstorage-ng is still able to do its job.

Trello: https://trello.com/c/jssdI31v/

imobachgs:
See https://github.com/openSUSE/libstorage-ng/pull/898.

imobachgs:
Once libstorage-ng 4.5.45 is available, the problem should be gone. yast2-storage-ng 4.5.9 would improve the situation in case an error still occurs.

Issue #207, opened 2022-06-18
Move storage/bootloader management to a separate process


imobachgs:
While working on #201, we found out that the software layer depends on the storage API too. As the manager is blocked during installation, the software layer cannot talk to the storage layer. So our best bet would be to split the storage layer too. As bootloader and storage are quite related, we can put both of them together below an `org.opensuse.DInstaller.Storage` D-Bus API.

imobachgs:
Implemented in #248 to contribute to fixing #247

Issue #169, opened 2022-05-25
Add a D-Bus API for registering questions


imobachgs:
While working in the multi-process architecture (see #153), we found out that from the client's point of view, it might be pretty inefficient to check for questions in each separate service.

We decided to move the "questions" handling to a single point (it might be the Manager or a separate process). Each of the processes might register a new question via D-Bus, and wait until it is solved.

imobachgs:
Implemented in #231. Moved to a separate process in #248.

PR #248, opened 2022-08-26
Questions and storage

imobachgs:
## Problem

While working on #247 we found out that the manager might get blocked for quite some time while waiting for the user's answer. So we decided to implement the overdue move storage/bootloader to a separate process (#207).

However, this change introduces a bi-directional dependency between the storage and the manager (as storage needs to access the questions API). So we decided to move that API to a separate process too (extending #169).

## Solution

* Move bootloader/storage layer to `org.opensuse.DInstaller.Storage`.
* Move the questions API to `org.opensuse.DInstaller.Questions`.

## Caveats

As `DInstaller::Security` is only used to set up the Linux Security Modules and it is closely related to bootloader, we decided to move its usage to this storage/bootloader process.

## Testing

- Added a new unit test
- Tested manuall


service/bin/d-installer
```diff
@@ -43,7 +43,7 @@ def start_service(name)
   service_runner.run
 end
 
-ORDERED_SERVICES = [:language, :software, :users, :manager].freeze
+ORDERED_SERVICES = [:questions, :language, :software, :storage, :users, :manager].freeze
```
jreidinger, line 46: when I see it question is what is future of that containerized dinstaller? if we want to keep container per process we need to update way how it is run. ( but as we do not have yet docker-compose or something similar, probably not big issue )

service/lib/dinstaller/dbus/clients/storage.rb
```diff
@@ -0,0 +1,110 @@
...
+        # @return [Array<String>] name of the devices
+        def candidate_devices
+          dbus_proposal["org.opensuse.DInstaller.Storage.Proposal1"]["CandidateDevices"]
+        end
+
+        # Actions to perform in the storage devices
+        #
+        # @return [Array<String>]
+        def actions
+          dbus_proposal["org.opensuse.DInstaller.Storage.Proposal1"]["Actions"].map do |a|
```
jreidinger, line 88: to my taste there are too many repeating of `"org.opensuse.DInstaller.Storage.Proposal1"` what about having constant for it?

service/lib/dinstaller/dbus/questions_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    class QuestionsService
+      SERVICE_NAME = "org.opensuse.DInstaller.Questions"
+      private_constant :SERVICE_NAME
+
+      # D-Bus connection
+      #
+      # @return [::DBus::Connection]
+      attr_reader :bus
+
+      # @param config [Config] Configuration object
```
jreidinger, line 41: this won't work as it is `_config` and not `config`

service/lib/dinstaller/helpers.rb
```diff
@@ -0,0 +1,49 @@
...
+
+module DInstaller
+  # This module contains some reusable utility methods
+  #
+  # We might consider turning some of them into proper classes if needed.
+  module Helpers
+    # Run a block in the target system
+    #
+    # @param block [Proc] Block to run on the target system
+    def on_target(&block)
```
jreidinger, line 32: this is something  that I can imagine to have in WFM itself, just with additional param `target`

service/lib/dinstaller/storage/manager.rb
```diff
@@ -20,41 +20,73 @@
...
+        progress.step("Selecting Linux Security Modules") { security.probe }
       end
 
       # Prepares the partitioning to install the system
       def install
-        add_packages
-        Yast::WFM.CallFunction("inst_prepdisk", [])
+        start_progress(4)
+        progress.step("Preparing bootloader proposal") do
+          # first make bootloader proposal to be sure that required packages are installed
```
jreidinger, line 62: hmm, how this will work with separate processes? it first says which packages bootloader needs and then writes bootloader, but where it wait till it is installed?

service/bin/d-installer
imobachgs: Yes, we need to think a little bit about it. Actually, the questions service does not even depend on YaST at all.

service/lib/dinstaller/storage/manager.rb
imobachgs: At that point, it adds the packages to the PackagesProposal (through D-Bus, thanks to our [overriden PackagesProposal](https://github.com/yast/d-installer/blob/master/service/lib/dinstaller/dbus/y2dir/manager/modules/PackagesProposal.rb).

Later, when the [bootloader is about to be installed](https://github.com/yast/d-installer/pull/248/files/5c7d9beeb80892a958e2138c0904250cb8a46532#diff-808c5e766d91d2840280e4f4ee5248f8a434ab7d78cd6e9a21103cc317e8eb72R84), the packages should be there.

service/lib/dinstaller/helpers.rb
imobachgs: It sounds like a good idea. I will make a note.

service/share/dbus.conf
```diff
@@ -23,12 +25,24 @@
     <allow send_destination="org.opensuse.DInstaller.Language" />
     <allow receive_sender="org.opensuse.DInstaller.Language" />
 
+    <allow send_destination="org.opensuse.DInstaller.Questions" />
+    <allow send_destination="org.opensuse.DInstaller.Questions"
+           send_interface="org.freedesktop.DBus.Introspectable"/>
+    <allow send_destination="org.opensuse.DInstaller.Questions" />
```
mvidner, line 31: Hm, this line is duplicate, overall this is equivalent to

    <allow send_destination="org.opensuse.DInstaller.Questions" />
    <allow   receive_sender="org.opensuse.DInstaller.Questions" />

Issue #249, opened 2022-09-01
It is not possible to unlock a LUKS device


imobachgs:
After moving questions to a separate service (see #248), it is not possible to unlock LUKS devices anymore. The problem is there is no way to ask a LUKS question over D-Bus, so those questions are always handled like generic ones.

mvidner:
Fixed by the above PR

PR #250, opened 2022-09-06
LUKS activation over D-Bus

mvidner:
## Problem

#249 says: After moving questions to a separate service (see https://github.com/yast/d-installer/pull/248), it is not possible to unlock LUKS devices anymore. The problem is there is no way to ask a LUKS question over D-Bus, so those questions are always handled like generic ones. 

## Solution

Teach the Questions service about `NewLuksActivation` and add the appropriate client code.

## Testing

Added new unit tests

Tested manually:

1. set up a toy LUKS partition in the testing VM
```sh
IMG=/dev/sda3 # it is the swap partition. make sure to use a scratch one too in your testing!
swapoff $IMG
cryptsetup --force-password luksFormat $IMG  # force accepting a simple password
```

2. now run the installer and click through 

3. Once the installer successfully unlocks the partition, we need to close it manually to be able to rerun the test: `cryptsetup close cr-auto-1`

See also `man cryptsetup`

## Screenshots

d-installer-luks-question

### DIFF
--- a/service/lib/dinstaller/dbus/manager_service.rb
+++ b/service/lib/dinstaller/dbus/manager_service.rb
@@ -23,7 +23,6 @@ require "dbus"
 require "dinstaller/manager"
 require "dinstaller/cockpit_manager"
 require "dinstaller/dbus/manager"
-require "dinstaller/dbus/language"
 require "dinstaller/dbus/storage/proposal"
 
 module DInstaller
@@ -102,17 +101,12 @@ module DInstaller
       # @return [Array<::DBus::Object>]
       def dbus_objects
         @dbus_objects ||= [
-          manager_dbus,
-          language_dbus
+          manager_dbus
         ]
       end
 
       def manager_dbus
         @manager_dbus ||= DInstaller::DBus::Manager.new(manager, logger)
-      end
-
-      def language_dbus
-        @language_dbus ||= DInstaller::DBus::Language.new(manager.language, logger)
       end
     end
   end
--- a/service/lib/dinstaller/dbus/clients/question.rb
+++ b/service/lib/dinstaller/dbus/clients/question.rb
@@ -46,6 +46,11 @@ module DInstaller
 
         # TODO: what other methods are useful?
 
+        # @return [String] Question text
+        def text
+          @dbus_iface["Text"].to_s
+        end
+
         # @return [Symbol] no answer yet = :""
         def answer
           @dbus_iface["Answer"].to_sym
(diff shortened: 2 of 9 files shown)


mvidner:
@imobachgs on IRC you said that "questions_manager looks weird when used over D-Bus". Can you please explain a bit?

imobachgs:
> @imobachgs on IRC you said that "questions_manager looks weird when used over D-Bus". Can you please explain a bit?

Sure. My concerns are visible in the [CanAskQuestion](https://github.com/yast/d-installer/blob/master/service/lib/dinstaller/can_ask_question.rb#L42) module. As a client, you pass a question object to the `add` method (`Question` class) and, if you are using D-Bus, you get an `DBus::Question` instance (`asked_question`).

We need to make sure that `Question` and `DBus::Question` implement a common interface (that should be enough). Or we could work with IDs, so `add` returns some kind of ID and `wait` receive such an ID. In that way, as a client, you do not worry which kind of question you are working with.

mvidner:
>  make sure that Question and DBus::Question implement a common interface (that should be enough).

OK. I did cheat in #231 and only implemented a bare minimum interface in DI::DBus::Clients::Question (as opposed to DI::Question)
 https://github.com/yast/d-installer/pull/231/files#diff-0b3ec193f1637a485a5350ffced5f13432d20b7166fb8855e781e7020f000c7cR47

Do you really mean DI::DBus::Question instead?

imobachgs:
> > make sure that Question and DBus::Question implement a common interface (that should be enough).
> 
> OK. I did cheat in #231 and only implemented a bare minimum interface in DI::DBus::Clients::Question (as opposed to DI::Question) https://github.com/yast/d-installer/pull/231/files#diff-0b3ec193f1637a485a5350ffced5f13432d20b7166fb8855e781e7020f000c7cR47
> 
> Do you really mean DI::DBus::Question instead?

No, you are right, I meant DBus::Clients::Question.

service/lib/dinstaller/dbus/manager_service.rb
```diff
@@ -102,18 +101,13 @@ def service
...
-          language_dbus
+          manager_dbus
         ]
       end
 
       def manager_dbus
         @manager_dbus ||= DInstaller::DBus::Manager.new(manager, logger)
       end
-
-      def language_dbus
```
imobachgs, line 114: Good catch!

imobachgs (approved):

PR #251, opened 2022-09-14
Add missing attempt param to LUKS activation questions

dgdavid:
## Problem

LuksActivationQuestion is not displaying any hint about a wrong password entered. See https://github.com/yast/d-installer/issues/165#issuecomment-1245428344

## Solution

Add the missing `attempt` param when registering the question in D-Bus.

## Testing

- Adapted unit tests.


ancorgs (approved):
I implemented the same change in my local copy (I needed it to check other stuff) and I can confirm this works.

Issue #165, opened 2022-05-20
LUKS activation 'cosmetic' issues


teclator:
## Description

There is some issues that we have detected while installing with some encrypted device present in the system.

![LuksActivation](https://user-images.githubusercontent.com/7056681/169482408-4c988177-206f-4124-922c-0e7f60e9aa80.png)

1. Maybe the question could be reworded as it ask if you want something but it is not a **Yes / No** answer, instead of it it provides a input for the encryption password. What about something like _**"Please, enter the encryption password if you want to decrypt/open it"**_
2. The default action is **Skip** which looks good when there is no password provided but it is still the **default** and **primary action** when a password is provided. What is worst, If the user press **Enter** the dialog is reload just like a reload action. I would expect it to do nothing or submit the default action which in case of password provided would expect it to be decrypt.

teclator:
@dgdavid was already taking some notes when I show him the errors and these are his notes with some ideas for fixing them:

> Observations from Knut:
> 
> The question would be better if it doesn't contain a Yes/No question. Something like "Please, enter the encryption password if you want to decrypt/open it" instead of "Do you want to decrypt it?"
> 
> It feels weird having the "Skip" as the primary action. If I'm not wrong, he prefers having "Decrypt", even though it is disabled until user enters a password
> 
> Note that the user still being able to jumpt to the skip action with the keyboard using just one TAB while "Decrypt" action is disabled
> 
> Hitting Enter is buggy: instead of send the "Decrypt" answer (when possible) it refreshes the app like if the user actually pressed F5
> 
> Which happens because the form wrapping the input is being [implicitely submitted](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission)
> 
> implicit submission mechanism must do nothing if the form has more than one field that blocks implicit submission, and must submit the form element from the form element itself otherwise.
> 
> There are a couple of options to fix that:
> 
> Set the defaultOption as a type=submit.
> 
> PROBLEMS:
> 
> a) there is not a defaultOption for LuksActivation, which became "Skip" (the first option) the "Primary" one... and would be weird mapping the expected "submit" to the "skip" action
> b) a Question is not always a form, but it will always relay in QuestionActions for building the actions (aka options) for answering the question. Not a big deal, the QuestionActions could receive a prop to know if it should use type=submit for defaultOption.
> Add an onSubmit method that prevents the implicit submission if none password has been entered but calls actionCallback("decrypt") otherwise
> 
[...]

PR #252, opened 2022-09-14
Usability improvements in the LUKS decrypt dialog

ancorgs:
## Problem

The problem is already described at #165, the dialog for decrypting a LUKS device has several usability problems including:

- The wording is a bit weird (kind of inconsistent with the labels of the buttons).
- Clicking "enter" does not submit the password, instead it causes a page reload.
- The default option is to skip decryption which:
  - Hurts usability with keyboard (and you are already using the keyboard if you entered the password, isn't it?)
  - Makes too easy to skip decryption even after you have already entered a password

## Solution

- Simpler wording, based on the YaST one (see YaST screenshot at #165)
- "Decrypt" is enforced to be the default option.
- Ensure pressing "enter" works as expected, that is:
  - Doing nothing if there is no password in the corresponding input (since the "Decrypt" button is disabled).
  - Submitting the entered password if there is any written.

## Still pending

The buttons don't look like in the other dialogs. They are aligned to the left.

## Potentially controversial code changes

- type=submit
- hardcoded string

## Testing

Tested manually

## Screenshots

decrypt

### DIFF
--- a/service/lib/dinstaller/luks_activation_question.rb
+++ b/service/lib/dinstaller/luks_activation_question.rb
@@ -75,7 +75,7 @@ module DInstaller
     #
     # @return [String]
     def generate_text
-      "The device #{device_info} is encrypted. Do you want to decrypt it?"
+      "The device #{device_info} is encrypted."
     end
 
     # Device information to include in the question
--- a/web/src/LuksActivationQuestion.jsx
+++ b/web/src/LuksActivationQuestion.jsx
@@ -20,7 +20,7 @@
  */
 
 import React, { useState } from "react";
-import { Alert, Form, FormGroup, Stack, StackItem, Text, TextInput } from "@patternfly/react-core";
+import { ActionGroup, Alert, Form, FormGroup, Stack, StackItem, Text, TextInput } from "@patternfly/react-core";
 import Popup from "./Popup";
 import QuestionActions from "./QuestionActions";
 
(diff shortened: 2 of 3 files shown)


ancorgs:
After performing all these changes (eg. putting the buttons inside the form) I realized the enter key works properly in the "New root password" popup. I want to check what the difference because it could mean there is a less intrusive way to fix the original problem in the decryption dialog.

dgdavid:
> After performing all these changes (eg. putting the buttons inside the form) I realized the enter key works properly in the "New root password" popup. I want to check what the difference because it could mean there is a less intrusive way to fix the original problem in the decryption dialog.

Excuse me @ancorgs . Blame to my bad memory. I completely forgot that @joseivanlopez and me were working on the implicit submission issue  few months ago and solved it by using the [`form` attribute](https://html.spec.whatwg.org/#the-button-element) in buttons to avoid the implicit submission. Indeed, it was the less intrusive solution we found.

dgdavid:
> using the [`form` attribute](https://html.spec.whatwg.org/#the-button-element) in buttons to avoid the implicit submission

See https://github.com/yast/d-installer/commit/45d3efc4bc6ee389fc5596e9fc6c5802d5c602bb

dgdavid:
> The buttons don't look like in the other dialogs. They are aligned to the left.

**TL;DR**

This is because these buttons are no longer placed in the Modal footer after you nested them inside an `ActionGroup`.

**Long Story**

To follow EOS recommendations and place modal buttons at the right, we simply put a `flex-direction: row-reverse` CSS rule

https://github.com/yast/d-installer/blob/66c1bec8a926b34dd1010da3922c3ddac4b88c5d/web/src/patternfly.scss#L93-L96

It is still working, but after you nested `Popup.Actions` the `Popup` component is not able to extract them using the `React.Children` API because it is no longer a direct child of the Popup.

https://github.com/yast/d-installer/blob/66c1bec8a926b34dd1010da3922c3ddac4b88c5d/web/src/Popup.jsx#L196-L198

I.e., `actions` is an empty array now and these buttons are not placed in the [`PatternFly Modal`](https://www.patternfly.org/v4/components/modal)  _\<footer\>_ but in its body. Something to document better or even to improve, I guess.

ancorgs:
> [...] I realized the enter key works properly in the "New root password" popup. I want to check what the difference because it could mean there is a less intrusive way to fix the original problem in the decryption dialog.

After checking how things are done there, I came up with #253 which looks to me like a less intrusive change (and more consistent with other dialogs in the UI).

I'm closing this pull request in favor of that other one.

PR #253, opened 2022-09-15
Usability improvements in the LUKS decrypt dialog

ancorgs:
## Problem

The problem is already described at #165, the dialog for decrypting a LUKS device has several usability problems including:

- The wording is a bit weird (kind of inconsistent with the labels of the buttons).
- Clicking "enter" does not submit the password, instead it causes a page reload.
- The default option is to skip decryption which:
  - Hurts usability with keyboard (and you are already using the keyboard if you entered the password, isn't it?)
  - Makes too easy to skip decryption even after you have already entered a password

## Solution

- Simpler wording, based on the YaST one (see YaST screenshot at #165)
- "Decrypt" is enforced to be the default option.
- Ensure pressing "enter" works as expected, that is:
  - Doing nothing if there is no password in the corresponding input (since the "Decrypt" button is disabled).
  - Submitting the entered password if there is any written.

## Other pull requests

This replaces #252. In my opinion, this is less intrusive.

## Testing

Tested manually

### DIFF
--- a/web/src/LuksActivationQuestion.jsx
+++ b/web/src/LuksActivationQuestion.jsx
@@ -39,11 +39,19 @@ const renderAlert = (attempt) => {
 export default function LuksActivationQuestion({ question, answerCallback }) {
   const [password, setPassword] = useState(question.password || "");
   const conditions = { disable: { decrypt: password === "" } };
+  const defaultAction = "decrypt";
 
   const actionCallback = (option) => {
     question.password = password;
     question.answer = option;
     answerCallback(question);
+  };
+
+  const triggerDefaultAction = async (e) => {
+    e.preventDefault();
+    if (!conditions.disable?.[defaultAction]) {
+      actionCallback(defaultAction);
+    }
   };
 
   return (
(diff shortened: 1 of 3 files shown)


web/src/LuksActivationQuestion.jsx
```diff
@@ -39,13 +39,21 @@ const renderAlert = (attempt) => {
...
   const conditions = { disable: { decrypt: password === "" } };
+  const defaultAction = "decrypt";
 
   const actionCallback = (option) => {
     question.password = password;
     question.answer = option;
     answerCallback(question);
   };
 
+  const triggerDefaultAction = async (e) => {
```
dgdavid, line 50: NP: since y the function does not contain an `await` statement, I think `async` is not needed here. Although technically it _isn't wrong_ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description.

dgdavid (approved):
LGTM, just a neat picking comment.

web/src/LuksActivationQuestion.jsx
ancorgs: oops, copy&paste :-)

ancorgs:
This failing check looks pretty bogus to me, especially since this pull request includes no modifications in the backend at all: "_Coveralls - backend — Coverage decreased (-0.0%) to 73.023% _"

dgdavid (approved):

PR #260, opened 2022-10-17
Add basic support for network configuration

dgdavid:
## Problem

So far, D-Installer does not offer a way to either, see how the network will end up configured or a way to change it.

- https://trello.com/c/h1qF8pue (internal link)

## Solution

Add a bare minimum and interactive summary in the installation overview.

We have been playing with several approaches (see one of them [here](https://gist.github.com/dgdavid/7fe0193688ce5d8d4709c34126b2abde)) to be able to interact with the [NetworkManager D-Bus API](https://networkmanager.dev/docs/api/latest/spec.html) while keeping the summary simpler enough. Consequently, this first _proposal_ only displays ETHERNET and WiFi **active** connections, allowing to change only the formers. 

The basic support to interact with the wireless networks will be added as part of https://trello.com/c/SFpwJaU7/ (internal link).

## Caveat

Of course, nothing is written in stone (yet :P) and this UI could be subject to change in favor of a better UX in the next iterations. Thus, take this PR as a starting point, which has helped a lot to understand how we want to interact with Network Manager DBus API and to have a big picture that will help continue the implementation of network features in D-Installer.

## Testing

- Added a few new unit tests (still missing others, to be written once the UI is more consolidated)
- Tested manually

## Screenshots

| Overview | Setting up a wired connection |
|-|-|
| Network overview | Editing a wired connection
 |


web/src/IpSettingsForm.jsx
```diff
@@ -0,0 +1,184 @@
...
+    return (
+      <HelperText>
+        <HelperTextItem variant="error">{errors[field]}</HelperTextItem>
+      </HelperText>
+    );
+  };
+
+  return (
+    <Popup isOpen height="medium" title={`Edit "${connection.id}" connection`}>
+      {/* FIXME: use a real onSubmit callback */}
```
dgdavid, line 128: Left over

web/src/IpSettingsForm.jsx
```diff
@@ -0,0 +1,184 @@
...
+        <AddressesDataList
+          addresses={addresses}
+          updateAddresses={setAddresses}
+          allowEmpty={method === "auto"}
+        />
+      </Form>
+
+      <Popup.Actions>
+        <Popup.Confirm form="edit-connection" type="submit" />
+        {/* FIXME: use a real onClick callback */}
```
dgdavid, line 179: Left over

web/src/Network.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+    return client.network.listen("connectionUpdated", onConnectionUpdated);
+  }, [client.network]);
+
+  if (!connections.length) return null;
+
+  const activeWiredConnections = connections.filter(c => c.type === CONNECTION_TYPES.ETHERNET);
+  const activeWifiConnections = connections.filter(c => c.type === CONNECTION_TYPES.WIFI);
+
+  return (
```
dgdavid, line 71: When there are none active connection

* Return an explanatory message (i.e., "Network cannot be configured", "No network connections available"), or
* Display a list of devices that can be configured to have an active connection

Something to do in next iterations, after having a more clear vision about the desired UI for that _Network Overview_ component (please, try to avoid making it too complex; we could have a _Network Section_ component for an advanced network setup).

web/src/NetworkWifiStatus.jsx
```diff
@@ -0,0 +1,40 @@
...
+import React from "react";
+import { CONNECTION_STATE } from "./client/network";
+
+import ConnectionsDataList from "./ConnectionsDataList";
+
+/**
+ * D-Installer component to show status of wireless network connections
+ *
+ * @todo evaluate if it should be "merged" into NetworkWiredStatus
+ * @todo display link for setting up a WiFi connection when possible
```
dgdavid, line 31: See https://trello.com/c/SFpwJaU7/ (internal link)

web/src/NetworkWifiStatus.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+import React from "react";
+import { CONNECTION_STATE } from "./client/network";
+
+import ConnectionsDataList from "./ConnectionsDataList";
+
+/**
+ * D-Installer component to show status of wireless network connections
+ *
+ * @todo evaluate if it should be "merged" into NetworkWiredStatus
```
dgdavid, line 30: :point_up_2:

web/src/NetworkWiredStatus.jsx
```diff
@@ -0,0 +1,45 @@
...
+import React, { useState } from "react";
+import IpSettingsForm from "./IpSettingsForm";
+import ConnectionsDataList from "./ConnectionsDataList";
+
+import { CONNECTION_STATE } from "./client/network";
+
+/**
+ * D-Installer component to show status of wired network connections
+ *
+ * @todo evaluate if it should be "merged" into NetworkWifiStatus
```
dgdavid, line 31: :point_up_2:

web/src/Overview.jsx
```diff
@@ -31,12 +31,14 @@ import Category from "./Category";
...
 import Storage from "./Storage";
 import Users from "./Users";
+import Network from "./Network";
 import Popup from "./Popup";
 
 import {
   EOS_SOFTWARE as OverviewIcon,
   EOS_TRANSLATE as LanguagesSelectionIcon,
   EOS_VOLUME as HardDriveIcon,
+  EOS_SETTINGS_ETHERNET as NetworkIcon,
```
dgdavid, line 41: A better icon is more than welcome. But remember that we are using https://eos-icons.com/, so have a look there for consistence. (Spoiler: although it is listed there, EOS_LAN actually is not available; in fact, would be nice to use it to visually identify listed wired connections)

web/src/TargetIpsPopup.jsx
```diff
@@ -20,62 +20,80 @@
...
-  if (addresses.length === 0) return null;
-
-  if (ips.length === 1) return <Text component="small" className="host-ip">{label}</Text>;
-
   return (
     <>
-      <Button variant="link" onClick={open}>
-        { label }
+      <Button variant="link" onClick={open} isDisabled={ips.length === 1}>
+        {firstIp} {hostname && <Text component="small">({hostname})</Text>}
```
dgdavid, line 84: :warning:  **NOTE** 

Despite we have a _Network overview_ in the installation summary now, we have kept this component because it is interesting to have a way to quickly identify the machine we have dealing with. For example, think of someone performing multiple installations at the same time… they must be able to distinguish between these machines no matter which section is displayed (product selection, storage section, etc)

Of course, we can do a lot of things to improve the current presentation, but let's book some time to think/discuss about it later.

web/src/app.scss
```diff
@@ -187,3 +187,31 @@ p {
...
+}
+
+.medium-modal-popup {
+  min-block-size: 55vh;
+  max-block-size: 55vh;
+}
+
+.large-modal-popup {
+  min-block-size: 85vh;
+  max-block-size: 85vh;
```
dgdavid, line 216: These height sizes were added to have _height-fixed_ popups, which is useful to avoid having a popup that changes its height while adding or removing dynamic fields, as happens now with the Connection Settings.

If you think other sizes are more appropriated, please speak up :wink:

web/src/client/network.js
```diff
@@ -23,13 +23,99 @@
...
+};
+
+// TODO: document
+const CONNECTION_TYPES = {
+  ETHERNET: "802-3-ethernet",
+  WIFI: "802-11-wireless"
+};
+
+/**
+ * @typedef {object} IPAddress
```
dgdavid, line 61: Question: should we move types definitions to a separated file?

web/src/client/network.test.js
```diff
@@ -0,0 +1,278 @@
...
+        prefix: { v: "22", t: "s" }
+      }
+    ]
+  }
+};
+
+const ipConfigProxy = path => {
+  return addressesData[path];
+};
+
```
dgdavid, line 152: We are testing a bare minimum here, reason why there is a lot of room for improvements in this unit test.

dgdavid (commented):
Please, take this PR as a starting point for having a network overview and adding a basic support for its configuration. We know that there are plenty of improvements and nice features that can be added…. but we need to sort and planning them before continue to avoid still walking in circles :wink:

web/src/AddressesDataList.jsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  return (
+    <Stack className="data-list-form" hasGutter>
+      <StackItem>
+        <Split hasGutter>
+          <SplitItem isFilled>Addresses</SplitItem>
+          <SplitItem>
+            <Button variant="primary" className="btn-sm" onClick={() => addAddress()}>
+              <PlusIcon />
```
dgdavid, line 126: I have tested to change the icon and use test instead.

What do you think? Which you prefer? Why?

| Using icons | Using text (no address yet) | Using text (at least one address already)
|-|-|-|
| Using icons | Using texts 1/2 | Using texts 2/2 |


```diff
diff --git a/web/src/AddressesDataList.jsx b/web/src/AddressesDataList.jsx
index f83a7b9..baba9f9 100644
--- a/web/src/AddressesDataList.jsx
+++ b/web/src/AddressesDataList.jsx
@@ -41,9 +41,6 @@ import {
   SplitItem
 } from "@patternfly/react-core";
 
-import PlusIcon from "@patternfly/react-icons/dist/js/icons/plus-icon";
-import MinusIcon from "@patternfly/react-icons/dist/js/icons/minus-icon";
-
 let index = 0;
 
 export default function AddressesDataList({
@@ -81,7 +78,7 @@ export default function AddressesDataList({
       return (
         <DataListAction>
           <Button variant="secondary" className="btn-sm" onClick={() => deleteAddress(id)}>
-            <MinusIcon />
+            Remove
           </Button>
         </DataListAction>
       );
@@ -116,6 +113,8 @@ export default function AddressesDataList({
     );
   };
 
+  const newAddressButtonLabel = addresses.length === 0 ? "Add an address" : "Add another address";
+
   return (
     <Stack className="data-list-form" hasGutter>
       <StackItem>
@@ -123,7 +122,7 @@ export default function AddressesDataList({
           <SplitItem isFilled>Addresses</SplitItem>
           <SplitItem>
             <Button variant="primary" className="btn-sm" onClick={() => addAddress()}>
-              <PlusIcon />
+              {newAddressButtonLabel}
             </Button>
           </SplitItem>
         </Split>

Comment (joseivanlopez) [reply web/src/AddressesDataList.jsx:129]: IMHO, using buttons with text makes the UI is much more clear .

File: web/src/client/network.js [hunk 13]
```diff
@@ -212,7 +212,8 @@ class NetworkManagerAdapter {
     // FIXME: find a better way to add gateway only if there are addresses. If not, a DBusError will
     // be raises "gateway cannot be set if there are no addresses configured".
     if (newSettings.ipv4["address-data"].v.length !== 0) {
-      newSettings.ipv4.gateway = cockpit.variant("s", connection.ipv4.gateway);
+      const gateway = newSettings.ipv4.method === "manual" ? newSettings.ipv4.gateway : "";
```
teclator, line 215: I guess it is not needed here

dgdavid: Why? Could you please tell a bit more?

teclator: Sorry, my fault, I was thinking it was in other part of the code looks good ;)

web/src/AddressesDataList.jsx
dgdavid: Another option is to preserve the icon 
Screenshot from 2022-10-19 10-54-10
while adding the text

joseivanlopez: Personally I prefer without icons. It does not add value to me. I mean, having the icon does not make me to understand the meaning of the action better. BTW, what about using red color for "Remove" ?

teclator: For me just icons was fine

web/src/client/network.js
teclator: Even with DHCP we could set the gateway and also maintain the additional addresses, for example if you do not get a DHCP lease, you will use the gateway defined... so maybe we should just allow to set the field unconditionally as we do with additional IP addresses.

teclator: Therefore I would do `newSettings.ipv4.gateway = cockpit.variant("s", newSettings.ipv4.gateway)`  unconditionally.

teclator: Therefore I would not touch the gateway here at all, and just use what is in newSettings.ipv4

web/src/AddressesDataList.jsx
dgdavid: > Personally I prefer without icons. It does not add value to me. I mean, having the icon does not make me to understand the meaning of the action better. 

Again, I completely agree. The experiment of preserving the icons was done to, somehow, keep a certain level of "consistence" with cockpit UI. Nothing else. But from my point of view, texts makes a better UX. The user does not have to think about the goal of that "+" button there, specially when the form contains more than once, similar, widgets.

> [...] what about using red color for "Remove" ?

I have no a strong opinion about that at this moment, but putting it in red could be a bit "aggressive" or disrupting to me. Maybe we can change to red color when the user over the button.

web/src/AddressesDataList.jsx
```diff
@@ -0,0 +1,138 @@
...
+    };
+
+    const cells = [
+      <DataListCell key={`address-${id}-local`}>
+        <TextInput
+          defaultValue={address}
+          onChange={value => updateAddress(id, "address", value)}
+          placeholder="Ip Address"
+          aria-label="Ip Address"
+        />
```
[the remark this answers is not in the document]
dgdavid: Fixed by https://github.com/yast/d-installer/pull/270

web/src/client/network.js
imobachgs: Not needed yet.

web/src/client/network/index.js
```diff
@@ -0,0 +1,244 @@
...
+   * Returns the active connections
+   *
+   * @returns {Promise<ActiveConnection[]>}
+   */
+  async activeConnections() {
+    return this.adapter.activeConnections();
+  }
+
+  /**
+   * Returns the connection witht the given ID
```
teclator, line 209:
```suggestion
   * Returns the connection with the given ID
```

web/src/client/network/network_manager.js
```diff
@@ -0,0 +1,267 @@
...
+          }
+        ))
+        ),
+        dns: cockpit.variant("au", connection.ipv4.nameServers.map(stringToIPInt)),
+        method: cockpit.variant("s", connection.ipv4.method)
+      }
+    };
+
+    // FIXME: find a better way to add gateway only if there are addresses. If not, a DBusError will
+    // be raises "gateway cannot be set if there are no addresses configured".
```
teclator, line 125:
```suggestion
    // raise "gateway cannot be set if there are no addresses configured".
```

joseivanlopez (approved):
LGTM

Very good work guys! Congrats.

BTW, having the list of IPs in the summary looks a bit weird, but I think we still have to refine the UI. So far so good. Thanks!

PR #269, opened 2022-10-18
Force network manager as we do not support other backends

teclator:
## Problem

When we started with the installer we [added support for basic network setup](https://trello.com/c/xPThihCG/85-2-d-installer-basic-network-set-up) using the defaults from the control file which in case of ALP is not present anymore.

## Solution

As currently we will not support other backends apart of NetworkManager we will enforce the selected backend as NetworkManager and will continue relying in the y2network save_network client  in order to copy the config and enable the service at the end of the installation


imobachgs:
It looks good. It just a matter of adapting the unit tests.

imobachgs (approved):

PR #270, opened 2022-10-19
Validate IPs

imobachgs:
Add IP and netmask/prefix validation to #260. The confirmation button is not disabled yet.
Under the hood, we use https://github.com/whitequark/ipaddr.js/ as it is rather small and we already used it in cockpit-wicked.

<details>
<summary>Validating the IP address</summary>

Invalid IP
</details>


dgdavid (approved):
LGTM, thanks a lot

PR #272, opened 2022-10-20
Network D-Bus API

imobachgs:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/service/Gemfile
+++ b/service/Gemfile
@@ -26,4 +26,5 @@ gemspec
 
 group :development do
   gem "byebug"
+  gem "rubocop", "1.24.1"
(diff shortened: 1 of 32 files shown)


imobachgs:
We experimented with the idea of building a minimal Ruby-based D-Bus interface on top of the NetworkManager API. The plan was to avoid reimplementing the same logic in our web UI and the CLI. However, far from making things easier, it added quite some complexity. Consequently, we decided to drop this approach in favor of just #260.

PR #281, opened 2022-10-27
Fix typo in command in README.md

agam778:
## Problem

A typo was there in README.md

## Solution

Fixed the typo


imobachgs (approved):
Good catch! Thank you.

PR #292, opened 2022-11-03
Add support for configuring WiFi devices

teclator:
## Problem

In #260 we added basic support for configuring the network, the idea is to extend it supporting also wireless devices.

- https://trello.com/c/SFpwJaU7/3139-5-d-installer-add-support-for-configuring-wifi-devices

## Solution

We have added support for configuring WiFi networks at least allowing the user to connect to a network without security or using **WPA-PSK** authentication as well as configuring a **Hidden network**.

## Testing

- Added some unit tests **(but it is still not well covered)**.
- Tested manually

## Screenshots

<details>
<summary>Click to show/hide some screenshots</summary>

---

  | Network summary with Wi-Fi capabilities | Available networks | Connecting to a network |
  |-|-|-|
  | Installation summary holding a network section with Wi-Fi capabilities | Available Wi-Fi networks for connecting to | Connecting to a Wi-Fi network |

  | Form for setting a connection | Form for setting a hidden connection | Detecting errors BEFORE trying to connect |
  |-|-|-|
  | Form for setting a new Wi-Fi connection | Form for setting a hidden connection | Detecting errors BEFORE trying to connect |


web/src/WirelessSelector.jsx
```diff
@@ -0,0 +1,236 @@
...
+    { value: "wpa-psk", label: "WPA & WPA2 Personal" },
+    { value: "wpa-eap", label: "WPA & WPA2 Enterprise" }
+  ];
+
+  const selectorOptions = security_options.map(security => (
+    <FormSelectOption key={security.value} value={security.value} label={security.label} />
+  ));
+
+  const connectNetwork = async () => {
+    await client.network.connectTo(ssid, { security, password });
```
dgdavid, line 67: @teclator,

At this point, we have to send the hidden option when needed, right?

web/src/WifiSelector.jsx
```diff
@@ -0,0 +1,276 @@
...
+    });
+  });
+
+  const isStateChanging = (network) => {
+    const state = network.connection?.state;
+    return state === ConnectionState.ACTIVATING || state === ConnectionState.DEACTIVATING;
+  };
+
+  const renderFilteredNetworks = (networks) => {
+    return networks.map(n => {
```
imobachgs, line 171: Please, use a meaningful name. For a small function, is perfectly fine, but this one has around 60 lines :-)

web/src/WifiSelector.jsx
```diff
@@ -0,0 +1,276 @@
...
+              Connect to hidden network
+            </Button>
+          </Center> }
+      </>
+    );
+  };
+
+  return (
+    <Popup isOpen={isOpen} height="large" title="Connect to a Wi-Fi network">
+      { renderFilteredNetworks(networksFromValues(networks)) }
```
imobachgs, line 266: Why not use a proper component? OK, I know, components are functions, but I would expect to have something like `Networks` and `HiddenNetworkForm`. Do you need to share all that state between them?

In any case, about the `renderFilteredNetworks`, does that function do any filtering? If that's not the case, I do not think you need to mention that in the name. It isn't very clear.

web/src/WifiSelector.jsx
```diff
@@ -0,0 +1,276 @@
...
+    const loadNetworks = async () => {
+      const knownSsids = [];
+
+      return client.network.accessPoints()
+        .sort((a, b) => b.strength - a.strength)
+        .reduce((networks, ap) => {
+          // Do not include networks without SSID
+          if (!ap.ssid || ap.ssid === "") return networks;
+          // Do not include "duplicates"
+          if (knownSsids.includes(ap.ssid)) return networks;
```
imobachgs, line 93: What about filtering networks you cannot configure? For instance, we cannot set up a WEP network, right? Alternatively, we could show them, but we should not allow trying to connect.

imobachgs (changes requested):

web/src/WifiConnectionForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+  FormGroup,
+  FormSelect,
+  FormSelectOption,
+  TextInput
+} from "@patternfly/react-core";
+import { useInstallerClient } from "./context/installer";
+
+const security_options = [
+  { value: "", label: "None" },
+  { value: "wpa-psk", label: "WPA & WPA2 Personal" }
```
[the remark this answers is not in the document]
teclator: In this case we missed a class **enum** for adding things like the label for example.

web/src/WifiConnectionForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+];
+
+const selectorOptions = security_options.map(security => (
+  <FormSelectOption key={security.value} value={security.value} label={security.label} />
+));
+
+const securityFrom = (supported) => {
+  if (supported.includes("WPA2"))
+    return "wpa-psk";
```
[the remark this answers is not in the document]
teclator: Similar to above and it also shows that is very related to the adapter... so maybe it should be declared there.

web/src/TargetIpsPopup.jsx
```diff
@@ -30,64 +30,45 @@ import { formatIp } from "./client/network";
...
 
   const open = () => setIsOpen(true);
   const close = () => setIsOpen(false);
 
   return (
     <>
-      <Button variant="link" onClick={open} isDisabled={ips.length === 1}>
-        {firstIp} {hostname && <Text component="small">({hostname})</Text>}
+      <Button variant="link" onClick={open} isDisabled={addresses.length === 1}>
+        {formatIp(firstIp)} {hostname && <Text component="small">({hostname})</Text>}
```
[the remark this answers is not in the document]
teclator: Better to move now it than later ;)

web/src/WifiSelector.jsx
dgdavid: Imo, you're right.

Somehow, we forget to extract this to small components after so many changes to get the Wi-Fi selector working. I have [tried to improve the situation a bit](https://github.com/yast/d-installer/pull/292/commits/beb0e4a7ffcf41c5d41e1c4facb9c546e2028472), but I fear a better improvement/re-organization must wait until next iterations in the network area.

What is more, I'd like to create a custom _SelectionList_ component (Wi-Fi and Product selector are quite similar, though) and reorganize not only that part, but the whole src/ directory. 

Hope, however, it looks good/acceptable enough for now.

web/src/HiddenNetworkForm.jsx
```diff
@@ -0,0 +1,67 @@
...
+import WifiConnectionForm from "./WifiConnectionForm";
+
+function HiddenNetworkForm({ network, expanded, onClick, onCancel, onSubmitCallback }) {
+  return (
+    <>
+      <Card className={[
+        "selection-list-item",
+        expanded && "selection-list-focused-item",
+        !expanded && "collapsed"
+      ].join(" ")}
```
dgdavid, line 42: NOTE: better to use a util function to improve this and avoid _"false"_ CSS classes. 

See https://github.com/jangouts/jangouts/blob/2081d8f344b016a884b58f651a46aa69ece5070a/src/utils/common.js#L8-L18

```js
/**
 * Helper to simplify building the className prop conditionally
 *
 * @param {String} CSS classes to join
 * @returns {String} final CSS classes properly joined, exluding "false"s
 */
function classNames(...classes) {
  return classes.filter((item) => !!item).join(' ');
}
```

web/src/WifiConnectionForm.jsx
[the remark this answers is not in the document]
imobachgs: Yes, that's true. My point is that using a "wpa-psk" looks fragile. But let's address it later.

web/src/WifiConnectionForm.jsx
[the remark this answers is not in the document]
imobachgs: Sure it should not be declared here :-) but as said before, we can live with this.

web/src/NetworkListItem.jsx
```diff
@@ -0,0 +1,117 @@
...
+    case ConnectionState.DEACTIVATING:
+      return 'Disconnecting';
+    case ConnectionState.DEACTIVATED:
+      return 'Disconnected';
+    default:
+      return "";
+  }
+};
+
+function NetworkListItem ({ network, isChecked, isFocused, showAsChecked, showSpinner, onSelect, onCancel }) {
```
imobachgs, line 60: It would be nice to document the arguments.

web/src/NetworksList.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+function NetworksList({ networks, activeNetwork, selectedNetwork, onSelectionCallback, onCancelSelectionCallback }) {
+  const isStateChanging = (network) => {
+    const state = network.connection?.state;
+    return state === ConnectionState.ACTIVATING || state === ConnectionState.DEACTIVATING;
+  };
+
+  return networks.map(n => {
+    const isChecked = n.ssid === selectedNetwork?.ssid;
+    const showAsChecked = !selectedNetwork && n.ssid === activeNetwork?.ssid;
```
imobachgs, line 35: np: Would not be better something like `isActive`? I meant, the component will decide whether to show it. It sounds more "semantic" (perhaps `isSelected` and `isActive`). Just a suggestion, feel free to ignore it.

web/src/NetworksList.jsx
```diff
@@ -0,0 +1,55 @@
...
+    const showSpinner = (isChecked && n.settings && !n.connection) || isStateChanging(n);
```
imobachgs, line 36: The same as above: the `NetworkListItem` should decide which is the right representation (a spinner, a different color, etc.).

imobachgs (approved):
It looks much better now. Just a few suggestions, but feel free to ignore them.

web/src/NetworksList.jsx
dgdavid: Completely agree. Not an _np_ at at all.

Believe it or not, I thought the same after reading the code, a couple of hours after writing it.

Definitely, being in a rush is not a good thing ;)

Thanks a lot for taking your time for carefully reviewing that PR.

web/src/NetworkListItem.jsx
dgdavid: Ok, let's do it

web/src/HiddenNetworkForm.jsx
dgdavid: Done, see https://github.com/yast/d-installer/pull/292/commits/4491ee2c0651c589f7a1e6add535549e7b825be1

web/src/ConnectionsDataList.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+  name: "WiFi 1",
+  type: ConnectionTypes.WIFI,
+  addresses: [{ address: "192.168.69.200", prefix: 24 }]
+};
+
+let conns = [];
+
+describe("ConnectionsDataList", () => {
+  describe("when no connections are given", () => {
+    it("renders nothing", async () => {
```
imobachgs, line 47: It is unnecessary to make this test async (IMHO, it is wrong). Actually, `ConnectionDataList` is a pretty simple component: no async, no `useInstallerClient`...

Please, have a look at #313 which should simplify the test. Of course, do not hesitate to ask if anything remains unclear.

web/src/ConnectionsDataList.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+      const { container } = installerRender(<ConnectionsDataList conns={conns} />, { usingLayout: false });
```
imobachgs, line 48: You can simply pass an empty array here. So you do not need to reassign `conns` in the test below.

web/src/WifiHiddenNetworkForm.test.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+import WifiHiddenNetworkForm from "./WifiHiddenNetworkForm";
+
+const beforeDisplayingFn = jest.fn();
+
+jest.mock("./WifiConnectionForm", () => () => "WifiConnectionForm mock");
+
+describe("WifiHiddenNetworkForm", () => {
+  describe("when it is visible", () => {
+    it("renders the WifiConnectionForm", async () => {
```
imobachgs, line 35: No need for async in `WifiHiddenNetworkForm` tests (except for the last one which tests user interaction). Please, check #313 for an example.

web/src/WifiHiddenNetworkForm.test.jsx
```diff
@@ -0,0 +1,70 @@
...
+      installerRender(<WifiHiddenNetworkForm visible />);
```
imobachgs, line 36: Please, use `plainRender` as `useInstallerClient` is not needed.

web/src/WifiNetworkListItem.test.jsx
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const fakeSettings = {
+  wireless:  {
+    password: "notSecret"
+  }
+};
+
+describe("NetworkListItem", () => {
+  it("renders an input radio for selecting the network", async () => {
```
imobachgs, line 46: In general, there is no need for async (except where user interaction is tested).

web/src/WifiNetworkListItem.test.jsx
```diff
@@ -0,0 +1,130 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+
+import { screen } from "@testing-library/react";
+import { installerRender } from "./test-utils";
```
imobachgs, line 25: I think you can use `plainRender`.

web/src/WifiNetworksList.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+const otherNetwork = {
+  ssid: "My Neighbour Network",
+  security: ["WPA2"],
+  strengh: 35
+};
+
+const networksMock = [myNetwork, otherNetwork];
+
+describe("WifiNetworksList", () => {
```
imobachgs, line 46: No need for async.

web/src/WifiNetworksList.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+
+import { screen, waitFor } from "@testing-library/react";
+import { installerRender } from "./test-utils";
```
imobachgs, line 25: You can use `plainRender`.

imobachgs (changes requested):
I have reviewed the tests more closely and found out that they use async too much. Additionally, if you do need to access the `useInstallerClient` hook, you can use `plainRender` instead.

web/src/ConnectionsDataList.test.jsx
teclator: Yep, originally it was an empty array, then I expected to write more tests and contexts and moved it to a variable, but finally it is not needed at all.

imobachgs (approved):

PR #293, opened 2022-11-03
Use `cockpit.proxies` and reduce the number of async methods

imobachgs:
This PR refactors the network client to:

* use `cockpit.proxies`, making it easier to track the changes (using events on `proxies`) and to read the properties;
* reduce the number of `async` methods in the network client.

We can use the same approach for objects like `/org/freedesktop/NetworkManager`, but let's do it as a separate PR.

### DIFF
--- a/web/src/client/network/network_manager.js
+++ b/web/src/client/network/network_manager.js
@@ -24,7 +24,7 @@
 import { DBusClient } from "../dbus";
 import cockpit from "../../lib/cockpit";
 import { intToIPString, stringToIPInt } from "./utils";
-import { NetworkEventTypes, ConnectionState } from "./index";
+import { NetworkEventTypes } from "./index";
 import { createAccessPoint, SecurityProtocols } from "./model";
 
 /**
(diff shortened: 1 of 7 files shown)


web/src/client/network/network_manager.js
```diff
@@ -24,22 +24,25 @@
...
-const NM_IFACE = "org.freedesktop.NetworkManager";
-const NM_SETTINGS_IFACE = "org.freedesktop.NetworkManager.Settings";
-const NM_CONNECTION_IFACE = "org.freedesktop.NetworkManager.Settings.Connection";
-const NM_ACTIVE_CONNECTION_IFACE = "org.freedesktop.NetworkManager.Connection.Active";
-const NM_IP4CONFIG_IFACE = "org.freedesktop.NetworkManager.IP4Config";
+const SERVICE_NAME = "org.freedesktop.NetworkManager";
+const IFACE = "org.freedesktop.NetworkManager";
+const SETTINGS_IFACE = "org.freedesktop.NetworkManager.Settings";
+const CONNECTION_IFACE = "org.freedesktop.NetworkManager.Settings.Connection";
+const ACTIVE_CONNECTION_IFACE = "org.freedesktop.NetworkManager.Connection.Active";
```
dgdavid, line 42: NP: if using "AP_IFACE" instead "ACCESS_POINTS_IFACE"… then use "AC_IFACE" instead "ACTIVE_CONNECTION_IFACE". For consistence. That said, I prefer reading longer versions. Just a neat-picking.

dgdavid (approved):

Issue #285, opened 2022-10-29
Currently the installer allows users to install OS without setting root and user accounts


gautamkrishnar:
Currently, the installer allows users to install OS without setting root and user accounts

https://user-images.githubusercontent.com/8397274/198835505-2bf7c999-8e83-4df5-9b90-67adefb3e329.mov

The resulting tumbleweed installation is totally unusable as a result of this:
![Screenshot 2022-10-29 at 7 37 25 PM](https://user-images.githubusercontent.com/8397274/198835990-159e8900-a6b7-4a38-a576-2c287fe0273c.png)

### Expected behavior
- The installer should make it mandatory to set a root/user password before installing the OS
- The install button should be disabled otherwise

imobachgs:
Thanks for the report. We have some related issues like #177 and #140. I will keep all of them open, but I expect they to be fixed altogether.

Issue #177, opened 2022-06-04
Unable to login to target installation


jengelh:
The D-Installer defaults define neither a root password nor a user with password. When clicking on "Install", there is no warning about this.

![1](https://user-images.githubusercontent.com/8861948/171993744-60bbc9e0-46c8-465d-b8dd-901959cb4d08.png)

The resulting system is unusable, as no login was assigned and there is no X11 autologin.

![2](https://user-images.githubusercontent.com/8861948/171993996-161eda17-af55-4176-961d-0974db39ca02.png)

imobachgs:
Hi @jengelh,

Yes, that's true. The fact is that we plan to add some mechanism to warn the user about this kind of stuff (not only with users). But we are working on other stuff we would like to solve first.

Thanks for the report!

PR #299, opened 2022-11-11
Pre-installation checks

imobachgs:
## Problem

D-Installer does not perform any pre-installation check. So if you click the "Install" button, it will blindly try to install the system. It does not matter whether you have provided all the required information.

See #285 or #177 as examples.

## Solution

This is the first version of a mechanism to report the issues that might block the installation. We might add some additional information to these *validation errors* in the future (severity, scope, etc.), but we need to start with something simple enough.

The implementation has two parts:

* A new [org.opensuse.DInstaller.Validation1](https://github.com/yast/d-installer/blob/7a03da68e37b657cdda4a665926bb4acf2619933/service/lib/dinstaller/dbus/interfaces/validation.rb#L51) that exposes the validation errors. The D-Bus layer takes care of keeping this information in sync, so the only thing that the backend classes need to provide is a `#validate` method (like [Manager#validate](https://github.com/yast/d-installer/blob/7a03da68e37b657cdda4a665926bb4acf2619933/service/lib/dinstaller/storage/manager.rb#L102) or [Users#validate](https://github.com/yast/d-installer/blob/7a03da68e37b657cdda4a665926bb4acf2619933/service/lib/dinstaller/storage/proposal.rb#L136)).
* Support at UI level to display these validation errors:
  - In the users section, the errors are shown after the user clicks the Install button.
  - In storage, the errors are always visible.

### User and storage checks

[demo-d-installer-checks.webm](https://user-images.githubusercontent.com/15836/201545826-e7017073-cff5-458e-a7ec-bef1e01ceaee.webm)

### Multiple errors

When there are multiple errors, they are shown in a pop-over. Of course, in the example below, the second is a testing error. Bear in mind that, by now, this situation should not happen.

multiple-errors-description

## TODO

- [x] Improve how errors are presented
- [x] Document the `Validation1` interface
- [x] Extract the `InstallButton` component to a separate file
- [x] Fix unit tests
- [x] Rename IsValid to just Valid.


service/lib/dinstaller/dbus/interfaces/validation.rb
```diff
@@ -0,0 +1,92 @@
...
+      #   end
+      #
+      #   class Demo < DInstaller::DBus::BaseObject
+      #     include DInstaller::DBus::Interfaces::Validation
+      #
+      #     dbus_interface "org.opensuse.DInstaller.Demo1" do
+      #       dbus_reader :errors, "as", dbus_name: "Errors"
+      #       dbus_method :Foo, "out result:u" do
+      #         # do some stuff
+      #         update_validation # run this method is the validation can change
```
joseivanlopez, line 43: Typo: *if* the validation ...

service/lib/dinstaller/dbus/storage/manager.rb
```diff
@@ -73,6 +76,10 @@ def finish
 
         # @return [DInstaller::Software::Manager]
         attr_reader :backend
+
+        def register_proposal_callbacks
+          backend.proposal.add_on_change_listener { update_validation }
```
joseivanlopez, line 81: NP: This line should be adapted https://github.com/yast/d-installer/blob/master/service/lib/dinstaller/storage/proposal.rb#L62

joseivanlopez:
```suggestion
          backend.proposal.on_calculate { update_validation }
```

joseivanlopez (commented):
Only minor things. For the UI part ... I did my best to understand it :), but it looks ok. Great work!

doc/dbus_api.md
```diff
@@ -333,6 +333,20 @@ The main object of a service implements the following interface:
...
+
+The main object of a service may implement the validation interface. It reports
+any issue that might block the installation.
+
+### org.opensuse.DInstaller.Validation1
+
+- ValidationErrors: array of strings (as)
+  List of validation errors.
+
+- Valid: boolean (b)
```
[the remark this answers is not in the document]
imobachgs: I do not know whether they are expensive, but it simply looked unnecessary to me as `valid?` can be defined based on `#errors` (so it should not get outdated).

service/lib/dinstaller/dbus/interfaces/validation.rb
```diff
@@ -60,28 +60,28 @@ def validation_properties
...
         def valid?
-          validation_errors.empty?
+          errors.empty?
         end
 
         def self.included(base)
           base.class_eval do
             dbus_interface VALIDATION_INTERFACE do
-              dbus_reader :validation_errors, "as", dbus_name: "ValidationErrors"
+              dbus_reader :errors, "as", dbus_name: "Errors"
```
joseivanlopez, line 84: I would say dbus_name: param is not needed now.

joseivanlopez (approved):
LGTM

service/lib/dinstaller/dbus/interfaces/validation.rb
imobachgs: Yes, you are right, removed...

PR #300, opened 2022-11-11
Simplify the D-Bus policy config 

mvidner:
## Problem

In https://bugzilla.suse.com/show_bug.cgi?id=1202059 we are asking for a security review of our D-Bus service.

But the relevant bus policy file is hard to read.

## Solution

Simplify the policy file while keeping it equivalent.

## Testing

- [x] Manually tested that the D-Bus policy simplification is still working:

#### Owning the bus name:

Non-root should fail
- `bundle exec bin/d-installer questions`
  - Expect error /not allowed to own the service "org.opensuse.DInstaller.Questions"/

Root should succeed
- `sudo !!`
  - Expect no error

#### Introspection call:
Non-root should fail
- `busctl introspect org.opensuse.DInstaller.Questions /org/opensuse/DInstaller/Questions1`
  - Expect error /Access denied/

Root should succeed:
- `sudo !!`
  - Expect /PropertiesChanged/

#### Question API call:
Non-root should fail
```
busctl \
  call \
  org.opensuse.DInstaller.Questions \
  /org/opensuse/DInstaller/Questions1 \
  org.opensuse.DInstaller.Questions1 \
  New sasas "should I stay or should I go" 2 yes no 1 yes
```

Root should succeed:
- `sudo !!`
  - Expect /^o "\/org\/opensuse/

## Screenshots

Not affected

### DIFF
--- a/service/share/dbus.conf
+++ b/service/share/dbus.conf
@@ -3,6 +3,7 @@
  "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
 
 <busconfig>
+  <!-- only root can own the services -->
   <policy user="root">
     <allow own="org.opensuse.DInstaller" />
     <allow own="org.opensuse.DInstaller.Language" />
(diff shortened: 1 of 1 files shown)


imobachgs (approved):

imobachgs (approved):
It still LGTM :-)

PR #301, opened 2022-11-14
Use `/etc/d-installer.yaml` outside of a Git clone

imobachgs:
To make testing easier, `d-installer` reads the configuration file from a relative path (`etc/d-installer.yaml`). So when you are working with a Git clone, it reads the file from `service/etc/d-installer.yaml`).

The problem is that it checks that the file exists, but it does not check whether it is part of a Git clone or not. This PR should fix that problem, preventing the file in `/usr/lib64/ruby/gems/3.1.0/gems/d-installer-0.5/etc/d-installer.yaml` from being read.


dgdavid (approved):
LGTM, thank you

PR #305, opened 2022-11-14
[web] Display only one primary action when editing a connection

dgdavid:
## Problem

The popup shown when editing a connection ends up displaying three buttons that looks like _primary_ actions, which might be a bit confusing.

## Solution

Try to use only one primary action per dialog, as we did in https://github.com/yast/d-installer/pull/292.

## Screenshots

| After | Before |
|-|-|
| Dialog with more than one primary action | Dialog with only one primary action |


imobachgs (approved):
LGTM

PR #268, opened 2022-10-18
Extend storage proposal D-Bus API

ancorgs:
## Problem

The D-Bus API for calculating a storage proposal is very limited. It only allows to select lvm and the candidate devices for the installation. The D-Bus API nees to provide more options to allow clients to configure the volumes to create.

* https://github.com/yast/d-installer/issues/264

## Solution

The D-Bus API was extended and the backend classes adapted to support volume definitions.

## Testing

* Manually tested
* Added unit tests
   
## Examples

~~~

# Calculate a proposal without indicating volumes at all

$ busctl call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage/Proposal1 org.opensuse.DInstaller.Storage.Proposal1 Calculate a{sv} 
1 CandidateDevices as 1 /dev/vdc
> u 0

$ busctl get-property org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage/Proposal1 org.opensuse.DInstaller.Storage.Proposal1 Volumes
> aa{sv} 2 
13 "MountPoint" s "/" "Optional" b false "DeviceType" s "partition" "Encrypted" b false "FsTypes" av 5 s "Ext2" s "Ext3" s "Ext4" s "Btrfs" s "XFS" "FsType" s "Btrfs"
"MinSize" x 18790481920 "MaxSize" i -1 "FixedSizeLimits" b false "AdaptiveSizes" b true "Snapshots" b true "SnapshotsConfigurable" b true "SnapshotsAffectSizes" b true
13 "MountPoint" s "swap" "Optional" b true "DeviceType" s "partition" "Encrypted" b false "FsTypes" av 1 s "Swap" "FsType" s "Swap"
"MinSize" i 1073741824 "MaxSize" x 2147483648 "FixedSizeLimits" b false "AdaptiveSizes" b false "Snapshots" b false "SnapshotsConfigurable" b false "SnapshotsAffectSizes" b false


# Calculate a proposal giving only some details about the volumes to create

$ busctl call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage/Proposal1 org.opensuse.DInstaller.Storage.Proposal1 Calculate a{sv} 
2 CandidateDevices as 1 /dev/vdc Volumes aa{sv} 2 
2 MountPoint s /home MinSize u 1024 
2 MountPoint s / Snapshots b false 
> u 0

[...]

PR #306, opened 2022-11-15
Force to use only one candidate device

imobachgs:
## Problem

After merging #268, there is a mismatch between the UI and the backend. By default, the backend selects many `candidate_devices` (as shown in the screenshot below).

Captura desde 2022-11-15 06-52-39

However, the UI only allows selecting a single device. Initially, I thought about extending the UI, but as you cannot specify any volume yet, there is no point in doing selecting more than one device. Actually, the result of the current behavior, is that `/dev/sdc` GPT gets removed for no good reason.

## Solution

I would force the use of a single device by now, so we can release the new version as soon as possible. After all, we are expected to work on the storage UI for the next-next release, so we can address this problem in a proper way.

I am proposing a solution but, to be honest, I am not sure if it is the "best" way. Please, let me know if you think we should do something different.

Captura desde 2022-11-15 07-03-27

## Testing

- Set a unit test to pending 😞 
- Tested manually

## Screenshots

### DIFF
--- a/service/lib/dinstaller/storage/proposal.rb
+++ b/service/lib/dinstaller/storage/proposal.rb
@@ -135,6 +135,12 @@ module DInstaller
         @settings.freeze
         proposal_settings = to_y2storage_settings(@settings)
 
+        # FIXME: by now, use only one disk
+        if proposal_settings.candidate_devices.nil?
+          first_disk = available_devices.first&.name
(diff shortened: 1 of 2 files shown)


joseivanlopez:
Closing in favor of https://github.com/yast/d-installer/pull/309.

PR #295, opened 2022-11-07
Use D-Bus activation instead of explicit service starting

mvidner:
## Problem

- https://github.com/yast/d-installer/issues/287

> The D-Installer backend is composed of several D-Bus services. A service might have dependencies on other services, so they are started sequentially (as we know the dependencies in advance). We decided to use a [simplistic approach](https://github.com/yast/d-installer/blob/master/service/bin/d-installer#L52-L58) (waiting 1 second between one service and the next), but it is time to evolve this solution to something more robust. If some service takes more time than expected to start, it might cause the whole start-up sequence to fail.

The startup failure then shows as "**Cannot connect to D-Bus**" in the web UI, as reported in #280 and #286

## Solution

D-Bus has an [activation mechanism][dba] exactly for this, now we use it.

- As we now ship a service running as root that users could potentially call and misuse, an audit is required. ~~Filed [bsc#1205246](https://bugzilla.suse.com/show_bug.cgi?id=1205246)~~ Duplicate of [bsc#1202059](https://bugzilla.suse.com/show_bug.cgi?id=1202059) AUDIT-0: ruby3.1-rubygem-d-installer: review of D-Bus services org.opensuse.DInstaller (and friends)

Also I've kept the manual running mechanism, by increasing the inter-service delay to 5 seconds and adding a tiny process manager with a help:
```console
$ sudo bundle exec bin/d-installer -h
bin/d-installer -h|--help
bin/d-installer -s|--status
bin/d-installer -k|--kill
bin/d-installer language|manager|questions|software|storage|users
```

[dba]: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-starting-services

- [x] development: how to terminate it, Ctrl-C no longer easy. See also https://github.com/yast/d-installer/issues/206
- [x] adjust README talking about manually starting it
- [x] clarify relationship between systemd and D-Bus `*.service`
- [x] syslog: adjust logger not to buffer and to omit the timestamp
- [x] integrate the setup into `setup.sh`

## Testing

- The changes are at the application+integration level so there are no unit tests
- Manually tested these deployment use cases:
    - development from git checkout: by running `setup.sh`
    - production, on [YaST:Head:D-Installer/d-installer-live/images](https://build.opensuse.org/package/binaries/YaST:Head:D-Installer/d-installer-live/images): did not actually test those, only built a *.gem and then a rubygem-d-installer.rpm and checked that the *.service files are present

## Screenshots

*If the fix affects the UI attach some screenshots here.*

PR #307, opened 2022-11-15
Activate the manager service without systemd

mvidner:
Followup to #295 

## Problem

We are starting one of the handful of services in a different way, via systemd instead of via dbus-daemon.

There is no reason other than I wanted to try if it works. It just adds complexity.

## Solution

Remove the unnecessary difference from the other services. dbus-daemon starts it just fine.

## Testing

- [x] manual test: take a clean VM; checkout the repo; `./setup.sh`; open browser to see no error; verify with `d-installer -s`

## Screenshots

Not affected


ancorgs (approved):
LGTM, the simpler the better

Issue #296, opened 2022-11-08
Enable vdagent on the d-installer live media (allows copy-paste etc)


lkocman:
One thing that should be easy to implement and would improve "testing experience" is enablement of the spice-vdagent on the live image. Users would be able to copy paste and resize screen easily.

lkocman:
This could be a good first issue

PR #308, opened 2022-11-16
Add files to build the Live ISO in the build service

### LINKED ISSUE #113: The remember password dialog at the very first experience with installer feels odd
An observation I was quite surprised when I experienced the installer for the first time. 

The very first dialog - I wanted to add my expected user/password for the host being installed. My fault I suppose

And the second odd thing is remembering the credentials for system install from livecd which I do not plan to do repeatedly. Perhaps we could skip that. We should minimize the 

d_installer_save_login
amount of user input (e.g. clicks), and this one seems to be unnecessary.

imobachgs:
This PR adds the files that we use to [build the ISO](https://build.opensuse.org/package/show/YaST:Head:D-Installer/d-installer-live) in the build service.

It closes #113 and #296.

## Testing

- You can grab the ISO with these changes from [this branch](https://build.opensuse.org/package/show/home:IGonzalezSosa:branches:YaST:Head:D-Installer/d-installer-live).

## To do

- [x] Extend the README with the missing information
- [x] Add a service to download sources/configuration


joseivanlopez (approved):
LGTM

PR #309, opened 2022-11-16
Use calculated settings

joseivanlopez:
## Problem

The new D-Bus API for the storage proposal (see https://github.com/yast/d-installer/pull/268) allows asking for some info, for example, the candidate devices or whether LVM was used for calculating the proposal. Some of that values could be nil if the proposal was calculated with incomplete data:

~~~
 busctl call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage/Proposal1 org.opensuse.DInstaller.Storage.Proposal1 Calculate a{sv} 1 LVM b true 
~~~

With the previous call, asking for `#candidate_devices` would return nil.

## Solution

Do not use the incomplete settings from the call to the D-Bus `Calculate` method. Instead of that, use settings with data from the actual Y2Storage proposal settings, which always contains all the information used to calculate the proposal.

This PR also limits the default proposal calculated during the config phase. Now such a proposal is calculated with a single candidate device (the first availabe device). Note that the UI only allows selecting a single device, and the storage proposal could use more than one candidate device if none is indicated. This is a temporary fix meanwhile the UI is adapted to use more than one device.

## Testing

* Adapted new unit tests
* Tested manually

### DIFF
--- a/service/lib/dinstaller/storage/proposal.rb
+++ b/service/lib/dinstaller/storage/proposal.rb
@@ -44,11 +44,6 @@ module DInstaller
     class Proposal
       include WithProgress
 
-      # Settings that were used to calculate the proposal
-      #
-      # @return [ProposalSettings, nil]
-      attr_reader :settings
-
       # Constructor
       #
       # @param logger [Logger]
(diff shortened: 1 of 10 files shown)


service/lib/dinstaller/storage/proposal.rb
```diff
@@ -103,37 +98,35 @@ def device_label(device)
...
-        return [] unless proposal
+      # @return [ProposalSettings]
+      def calculated_settings
+        return nil unless proposal
 
-        generator = VolumesGenerator.new(specs_from_proposal,
-          planned_devices: proposal.planned_devices)
-        volumes = generator.volumes(only_proposed: true)
+        settings = proposal.settings.dup
+        settings.volumes = settings.volumes.select(&:proposed?)
```
ancorgs, line 116: Shouldn't this filtering be part of the converter?

service/lib/dinstaller/storage/proposal.rb
```diff
@@ -103,37 +98,35 @@ def device_label(device)
...
 
-        # FIXME: setting this should be a responsibility of VolumesGenerator or any other component,
-        # but this is good enough until we implement fine-grained control on encryption
-        volumes.each { |v| v.encrypted = proposal.settings.use_encryption }
-
-        volumes
+        to_dinstaller_settings(settings,
+          default_specs: proposal.settings.volumes, devices: proposal.planned_devices)
```
ancorgs, line 119: Hmmm, I need to think closely about this but I was expecting:

```ruby
to_dinstaller_settings(proposal.settings, default_specs: default_specs, devices: proposal.planned_devices)
```

ancorgs (changes requested):
I think there is a certain responsibility mix. See my two comments.

ancorgs (approved):
LGTM

Issue #189, opened 2022-06-10
Improve configuration handling


imobachgs:
## The `Config` class

At this point, the Config class is basically a container for a regular Hash. When you want to read a given param, you basically access the hash through the `#data` method. No validation, type conversions and so on.

We might consider using something similar to [SectionWithAttributes](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/autoinst_profile/section_with_attributes.rb#L120). We could also have a look to [some meta-programming technique as the one used in yast2-network](https://github.com/yast/yast-network/blob/master/src/lib/cfa/interface_file.rb#L145-L151).

Additionally, we might want to change how [.load and .current methods](https://github.com/yast/d-installer/pull/187#pullrequestreview-1002318468) work.

## Configuration workflow

See https://github.com/yast/d-installer/pull/312#pullrequestreview-1184327235.

imobachgs:
It is not realistic to commit to this change, at least at short or mid term. Time to close it.

PR #312, opened 2022-11-16
Storage service applies the product's configuration

### LINKED ISSUE #311: Security/storage settings are not changed when a product is selected
The storage service does not get the product configuration when a product is selected. I guess it happens since we extracted bootloader/storage/security to a separate process (see #207). It causes many side-effects, like not being able to select the proper selinux packages.

imobachgs:
The storage service takes the selected product into account before performing the probing. It fixes #311.

PS: we should do something similar for the network service, but there is no network configuration in the YAML file, so we can keep it as it is now.

### DIFF
--- a/service/lib/dinstaller/storage/manager.rb
+++ b/service/lib/dinstaller/storage/manager.rb
@@ -30,6 +30,7 @@ require "dinstaller/with_progress"
 require "dinstaller/can_ask_question"
 require "dinstaller/security"
 require "dinstaller/dbus/clients/questions_manager"
+require "dinstaller/dbus/clients/software"
 require "dinstaller/helpers"
 
 Yast.import "PackagesProposal"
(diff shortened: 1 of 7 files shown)


dgdavid (approved):

service/lib/dinstaller/storage/manager.rb
```diff
@@ -50,6 +51,7 @@ def initialize(config, logger)
       # Probes storage devices and performs an initial proposal
       def probe
         start_progress(4)
+        config.pick_product(software.selected_product)
```
joseivanlopez, line 54: I don't like to have this here. Note that this class does not need the config at all. The config is used by `Storage::Proposal` and `Security`, so I would expect that classes retrieve the current config somehow. I think we need to re-think the config approach completely.

joseivanlopez: https://github.com/yast/d-installer/issues/189

imobachgs: Yes, I agree. I would keep the changes to the bare minimum, but it is clear that we need to evolve the configuration handling. I will add this information to #189.

dgdavid (approved):

PR #313, opened 2022-11-17
Improve ConnectionsDataList test

imobachgs:
* Do not rely on async unless it is strictly needed.
* Test that the `onSelect` callback is called.
* Do not use a `beforeEach` function.


teclator (approved):
LGTM

PR #316, opened 2022-11-17
Only show the link for connecting to a WiFI network if wireless is enabled

teclator:
## Problem

In #292 we added support for configuring a WiFi connection but it shows the link to connect to a network even when wireless is currently not enabled which is not correct at all.

## Solution

Do not show the link to connect to a Wireless when it is not enabled.

## Testing

- Added unit test for the backend, the Network.jsx component still lacks of it as there is some problem with the useEffect and maybe waitFor with a callback needs to be added.
- Tested manually

## Screenshots
| Wireless Enabled | Wireless Disabled |
| --- | --- |
| localhost_9090_cockpit_@localhost_d-installer_index html (9) | localhost_9090_cockpit_@localhost_d-installer_index html (10) |

D-Installer_NetworkSettings

### DIFF
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Dec 15 10:14:22 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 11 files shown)


dgdavid (approved):
Looks good, thank you!

PR #317, opened 2022-11-18
Check whether the root SSH public key is set

imobachgs:
Setting the root SSH public key is enough to install. However, we need to make sure that the network and the ssh service are enabled.


ancorgs (approved):
LGTM

PR #321, opened 2022-11-18
Storage proposal UI

joseivanlopez:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


web/src/components/storage/ProposalSettingsForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+            isChecked={state.encryption}
+            onChange={onEncryptionChange}
+          />
+        }
+      >
+        <FormGroup fieldId="encryptionPassword" label="Encryption password">
+          <TextInput
+            id="encryptionPassword"
+            type="password"
+            isDisabled={!state.encryption}
```
dgdavid, line 105: NP: perhaps we should move the `isDisabled` prop to the [`Fieldset`](https://github.com/yast/d-installer/blob/37b2276ad431a8dd38d7647136037eb53d514259/web/src/components/core/Fieldset.jsx#L54), making any future element in this group disabled too when the "Encrypt devices" is disabled. Note that the legend will be not disabled, so it's safe enough for this use case. Read more at https://html.spec.whatwg.org/#the-fieldset-element

imobachgs (approved):

PR #325, opened 2022-11-21
[web] Restructure src by creating a module per installer area

dgdavid:
## Problem

As D-Installer evolves, the src directory becomes messy and out of control.

## Solution

To create a [module](https://javascript.info/modules-intro) per _installation area_ by using [_re-exports_](https://javascript.info/import-export#re-export) and to use aliases for easing the way components are imported instead of using complex relative paths.

## Other interesting links

  * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#default_exports_versus_named_exports
  * https://webpack.js.org/configuration/resolve/
  * https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring
  * https://www.typescriptlang.org/tsconfig#paths
  * **https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/**


web/src/components/core/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export { default as About } from "./About";
+export { default as Category } from "./Category";
+export { default as FormLabel } from "./FormLabel";
```
dgdavid, line 24: Question: Should be `FormLabel`  in the _layout_ module instead?

web/src/components/language/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export { default as LanguageSelector } from "./LanguageSelector";
```
dgdavid, line 22: Maybe `LanguageSelector` can be named just `Selector` now that it is part of the _language_ module. On the other hand, then it might or might not be imported as _LanguageSelector_ or so in some places.

web/src/components/layout/DBusError.test.jsx
```diff
@@ -22,11 +22,11 @@
 import React from "react";
 
 import { screen } from "@testing-library/react";
-import { plainRender } from "./test-utils";
+import { plainRender } from "@/test-utils";
 
-import DBusError from "./DBusError";
+import { DBusError } from "@components/layout";
```
dgdavid, line 27: Maybe it is time for renaming `DBusError` component. What do you think?

web/src/components/layout/index.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export * from "./Layout";
+export { default as Layout } from "./Layout";
+export { default as Center } from "./Center";
+export { default as DBusError } from "./DBusError";
+export { default as LoadingEnvironment } from "./LoadingEnvironment";
```
dgdavid, line 26: I think this is PR is a good opportunity for making `LoadingEnvironment` a more generic `Loading` component. In fact, it already receives a[ _text_](https://github.com/yast/d-installer/blob/2de401429030d7b22febc58aad4758910aabe303/web/src/LoadingEnvironment.jsx#L29) prop that can be accordingly adapted too. What do you think?

web/src/components/network/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+export { default as Network } from "./Network";
+export { default as AddressesDataList } from "./AddressesDataList";
+export { default as ConnectionsDataList } from "./ConnectionsDataList";
+export { default as DnsDataList } from "./DnsDataList";
+export { default as IpAddressInput } from "./IpAddressInput";
+export { default as IpPrefixInput } from "./IpPrefixInput";
+export { default as IpSettingsForm } from "./IpSettingsForm";
+export { default as NetworkWifiStatus } from "./NetworkWifiStatus";
+export { default as NetworkWiredStatus } from "./NetworkWiredStatus";
```
dgdavid, line 30: I guess we can get rid of `Network` prefixes now. Do you agree?

web/src/components/network/index.js
```diff
@@ -0,0 +1,37 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export { default as Network } from "./Network";
```
dgdavid, line 22: What about renaming this compononent to `NetworkOverview`. Tha way, we can go for a `NetworkOverview` or `NetworkSummary` for the summary and `NetworkPage` for the section page. Simlarly, we could have `StorageOverview` and `StoragePage`, `LanguageOverview` and `LanguagePage` and so on. What is more, they could be simply an `Overview` and `Page` components inside their modules...

web/src/components/layout/index.js
```diff
@@ -0,0 +1,27 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export * from "./Layout";
```
joseivanlopez, line 22: Shouldn't this be `@components/layout/Layout` ?

web/src/components/network/index.js
joseivanlopez: Yes, I vote for it: `Network/Overview`, `Network/Page`, etc. But maybe in a separate PR.

web/src/components/core/index.js
joseivanlopez: Umm, I would say a `FormLabel` is not related to the layout. For now, I would keep it as part of `core`.

web/src/components/language/index.js
joseivanlopez: I would do all those changes about renaming things as part of a separate PR.

joseivanlopez (approved):
LGTM

web/src/components/layout/index.js
dgdavid: As discussed _off-line_, let's keep it as it is by now. Somehow, it's more clear that it is re-exporting components locally to that directory/area.

PR #327, opened 2022-11-22
[service] Fix Leap Micro version

imobachgs:
Fix Leap Micro version in the description (forgot to update it on #318).


ancorgs (approved):
:+1:

PR #330, opened 2022-11-22
Enlarge the temporary hack for TPM-based encryption

ancorgs:
Expanding #326 

Jeos-Firstboot is needed to trigger Olaf's scripts.


ancorgs:
Manually tested. It works as expected. jeos-firstboot is executed and does not touch the root password.

imobachgs (approved):

PR #332, opened 2022-11-23
[web] Improve Section component (former Category)

dgdavid:
## Problem

The `Category` component needs some improvements now that we have _pages_ too, not only the overview.

## Solution

To change _a bit_ the component for fulfilling current needs.

Starting with a rename, the `Section` component now allows
* To be used without icon (maybe useful for a _settings_ pages with a lot of sections)
* To specify a description for helping the user to understand what is the section about
* To include a visual separation between its title and the rest of its content
* To put an action right after the title, usually for opening a settings popup or similar.
* To change the icon for mentioned action (a _cog_ by default)
* To include a tooltip to be shown when user hovering the section action

## Screenshots

Below, see a few examples by using just fake sections

| Not displaying the tooltip | Displaying the tooltip |
|-|-|
| Screenshot from 2022-11-25 01-19-41 | Screenshot from 2022-11-25 01-19-38 |

---

Related to https://github.com/yast/d-installer/pull/321


imobachgs (approved):

joseivanlopez (approved):
LGTM

As discussed offline, it would be better to use composition instead of passing props, for example:

~~~
<Section>
  <SectionAction />
</section>
~~~

And the same for the `Fieldset` created in https://github.com/yast/d-installer/pull/333:

~~~
<Fieldset>
  <Legend />
</Fieldset>
~~~

Agreed to change it in a refact PR.

PR #336, opened 2022-11-25
Get the changes file from the repo to build the web UI

imobachgs:
Add the changelog to the list of files that are retrieved from the repo when building the web UI package (see #335).


dgdavid (approved):
Ouch! I didn't realize when reviewing #335. Let's merge it before anyone else notice it :stuck_out_tongue_winking_eye:

PR #337, opened 2022-11-28
Added validation of first_user creation

teclator:
## Problem

Currently it is allowed to add a root user or system users as the first user which is or could be problematic.

- https://trello.com/c/pwc8fQL0/3148-2-d-installer-do-not-allow-defining-root-as-a-first-user

## Solution

Rely on the **backend** validations at least for adding the first users not allowing **system known users** and **root** user. Any error issue found when trying to add the user will be reported by the **CLI** and by the **web** dialog.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

| Overview | Dialog |
| --- | --- |
| localhost_9090_cockpit_@localhost_d-installer_index html (11) | localhost_9090_cockpit_@localhost_d-installer_index html (12) |

### Service logs

`
I, [2022-11-28T09:08:48.705694 #28659]  INFO -- users: Setting first user 
I, [2022-11-28T09:08:48.706057 #28659]  INFO -- users: First user fatal issues detected: ["There is a conflict between the entered\nusername and an existing username.\nTry another one.", "No password entered.\nTry again."]
`

### DIFF
--- a/web/src/client/users.js
+++ b/web/src/client/users.js
@@ -29,6 +29,12 @@ const USERS_IFACE = "org.opensuse.DInstaller.Users1";
 const USERS_PATH = "/org/opensuse/DInstaller/Users1";
 
 /**
+* @typedef {object} UserResult
+* @property {boolean} result - whether the action succeeded or not
+* @property {string[]} issues - issues found when applying the action
+*/
+
+/**
  * @typedef {object} User
  * @property {string} fullName - User full name
  * @property {string} userName - userName
@@ -81,7 +87,7 @@ class UsersBaseClient {
    * Sets the first user
    *
    * @param {User} user - object with full name, user name, password and boolean for autologin
-   * @return {Promise<object>} returns an object with the result and the issues found if error
+   * @return {Promise<UserResult>} returns an object with the result and the issues found if error
    */
   async setUser(user) {
     const proxy = await this.client.proxy(USERS_IFACE);
--- a/web/src/client/users.test.js
+++ b/web/src/client/users.test.js
@@ -91,7 +91,7 @@ describe("#getRootSSHKey", () => {
 });
 
 describe("#setUser", () => {
-  it("sets the values of the first user and returns whether succeded or not an errors found", async () => {
+  it("sets the values of the first user and returns whether succeeded or not an errors found", async () => {
     const client = new UsersClient(dbusClient);
     const result = await client.setUser({
       fullName: "Jane Doe",
@@ -110,7 +110,7 @@ describe("#setUser", () => {
       usersProxy.SetFirstUser = jest.fn().mockResolvedValue(setFirstUserResult);
     });
 
-    it("returns an object with the result as 1 and the issues found", async () => {
+    it("returns an object with the result as false and the issues found", async () => {
       const client = new UsersClient(dbusClient);
       const result = await client.setUser({
         fullName: "Jane Doe",
--- a/web/src/components/users/FirstUser.test.jsx
+++ b/web/src/components/users/FirstUser.test.jsx
@@ -61,15 +61,15 @@ it("allows defining a new user", async () => {
   const button = within(firstUser).getByRole("button", { name: "is not defined" });
   await user.click(button);
 
-  await screen.findByRole("dialog");
+  const dialog = await screen.findByRole("dialog");
 
-  const fullNameInput = screen.getByLabelText("Full name");
+  const fullNameInput = within(dialog).getByLabelText("Full name");
   await user.type(fullNameInput, "Jane Doe");
 
-  const usernameInput = screen.getByLabelText("Username");
+  const usernameInput = within(dialog).getByLabelText(/Username/);
   await user.type(usernameInput, "jane");
 
-  const passwordInput = screen.getByLabelText("Password");
+  const passwordInput = within(dialog).getByLabelText(/Password/);
   await user.type(passwordInput, "12345");
 
   const confirmButton = screen.getByRole("button", { name: /Confirm/i });
@@ -94,11 +94,11 @@ it("doest not allow to confirm the settings if the user name and the password ar
   const button = within(firstUser).getByRole("button", { name: "is not defined" });
   await user.click(button);
 
-  await screen.findByRole("dialog");
+  const dialog = await screen.findByRole("dialog");
 
-  const usernameInput = screen.getByLabelText("Username");
+  const usernameInput = within(dialog).getByLabelText(/Username/);
   await user.type(usernameInput, "jane");
-  const confirmButton = screen.getByRole("button", { name: /Confirm/i });
+  const confirmButton = within(dialog).getByRole("button", { name: /Confirm/i });
   expect(confirmButton).toBeDisabled();
 });
 
@@ -108,9 +108,9 @@ it("does not change anything if the user cancels", async () => {
   const button = within(firstUser).getByRole("button", { name: "is not defined" });
   await user.click(button);
 
-  await screen.findByRole("dialog");
+  const dialog = await screen.findByRole("dialog");
 
-  const cancelButton = screen.getByRole("button", { name: /Cancel/i });
+  const cancelButton = within(dialog).getByRole("button", { name: /Cancel/i });
   await user.click(cancelButton);
 
   expect(setUserFn).not.toHaveBeenCalled();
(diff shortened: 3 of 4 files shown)


service/lib/dinstaller/users.rb
```diff
@@ -84,9 +84,15 @@ def assign_first_user(full_name, user_name, password, auto_login, _data)
       user = Y2Users::User.new(user_name)
       user.gecos = [full_name]
       user.password = Y2Users::Password.create_plain(password)
-      config.attach(user)
-      config.login ||= Y2Users::LoginConfig.new
-      config.login.autologin_user = auto_login ? user : nil
+      fatal_issues = user.issues.map.select(&:error?)
+
+      if fatal_issues.empty?
```
imobachgs, line 89: I would use a shortcut:

```ruby
return fatal_issues unless fatal_issues.empty?

config.attach(user)
config.login ||= Y2Users::LoginConfig.new
config.login.autologin_user = auto_login ? user : nil
[]
```

service/lib/dinstaller/dbus/users.rb
```diff
@@ -90,13 +91,18 @@ def initialize(backend, logger)
...
-          update_validation
-          0
+          if issues.empty?
+            dbus_properties_changed(USERS_INTERFACE, { "FirstUser" => first_user }, [])
+            update_validation
+          else
+            logger.info "First user fatal issues detected: #{issues.map(&:message)}"
+          end
+
+          [{ "result" => issues.empty? ? 0 : 1, "issues" => issues.map(&:message) }]
```
imobachgs, line 105: Why not use a pair with the result and the issues, instead of a hash?

imobachgs: And why not using a boolean instead of a number? D-Bus support a BOOLEAN type.

web/src/client/users.js
```diff
@@ -92,7 +92,11 @@ class UsersBaseClient {
       user.autologin,
       {}
     );
-    return result === 0;
+
+    return {
+      result: result.result.v,
```
imobachgs, line 97: I would rather prefer to use a boolean if possible.

imobachgs (changes requested):
The approach looks good, but I am not convinced about the typing:

1) I think you can return multiple values, so you can return the result and the list of issues without needing the put them into a dictionary.
2) In any case, I would expect a boolean instead of a number. If for any reason you need a number (although I think that's not the case) you can convert it at client level.

service/lib/dinstaller/dbus/users.rb
teclator: > Why not use a pair with the result and the issues, instead of a hash?

Do not have an strong opinion, it was suggested by @joseivanlopez and was fine for me.

teclator: > And why not using a boolean instead of a number? D-Bus support a BOOLEAN type.

The same here, I followed what looked like the convention in this class returning integers instead of booleans, we were also discussing that in some situation it will be not just true or false but the returned values could represent different kind of issues but as we are not in that point yet it makes sense to be a boolean.

service/lib/dinstaller/dbus/users.rb
```diff
@@ -90,13 +90,18 @@ def initialize(backend, logger)
         end
 
         dbus_method :SetFirstUser,
-          FUSER_SIG + ", out result:u" do |full_name, user_name, password, auto_login, data|
+          FUSER_SIG + ", out result:(bas)" do |full_name, user_name, password, auto_login, data|
```
jreidinger, line 93: I am missing where this change is documented.

web/src/client/users.js
```diff
@@ -70,29 +70,30 @@ class UsersBaseClient {
...
     return proxy.RootPasswordSet;
   }
 
   /**
-   * Sets the languages to install
+   * Sets the first user
    *
    * @param {User} user - object with full name, user name, password and boolean for autologin
-   * @return {Promise<boolean>} whether the operation was successful or not
+   * @return {Promise<object>} returns an object with the result and the issues found if error
```
imobachgs, line 84: Please, annotate the type. I would expect this to be a common pattern (a result object returning the whether it was successful or not and a list of errors). So you could even create a `@type` for it.

web/src/client/users.test.js
```diff
@@ -89,30 +91,35 @@ describe("#getRootSSHKey", () => {
...
-  describe("when setting the user fails", () => {
-    beforeEach(() => (usersProxy.SetFirstUser = jest.fn().mockResolvedValue(1)));
+  describe("when setting the user fails because some issue", () => {
+    beforeEach(() => {
+      setFirstUserResult = [false, ["There is an error"]];
+      usersProxy.SetFirstUser = jest.fn().mockResolvedValue(setFirstUserResult);
+    });
 
-    it("returns false", async () => {
+    it("returns an object with the result as 1 and the issues found", async () => {
```
imobachgs, line 113: s/1/false/

web/src/components/users/FirstUser.test.jsx
```diff
@@ -86,6 +88,20 @@ it("allows defining a new user", async () => {
   });
 });
 
+it("doest not allow to confirm the settings if the user name and the password are not provided", async () => {
+  const { user } = installerRender(<FirstUser />);
+  const firstUser = await screen.findByText(/A user/);
+  const button = within(firstUser).getByRole("button", { name: "is not defined" });
+  await user.click(button);
+
+  await screen.findByRole("dialog");
```
imobachgs, line 97: I would be consistent with the `within` usage above. See https://github.com/yast/d-installer/blob/27cfff5f127945629d8ed65509584d74a6739d3f/web/src/components/core/About.test.jsx#L36-L38.

It is not that the current solution does not work, but I would say it is more correct (and less fragile) to make sure that you interact with the dialog.

web/src/components/users/FirstUser.test.jsx
```diff
@@ -103,6 +119,45 @@ it("does not change anything if the user cancels", async () => {
...
+    setUserFn = jest.fn().mockResolvedValue(setUserResult);
+  });
+
+  it("shows the issues found", async () => {
+    const { user } = installerRender(<FirstUser />);
+    const firstUser = await screen.findByText(/A user/);
+    const button = within(firstUser).getByRole("button", { name: "is not defined" });
+    await user.click(button);
+
+    await screen.findByRole("dialog");
```
imobachgs, line 134: Same (use within to interact with the dialog).

imobachgs (changes requested):

service/lib/dinstaller/dbus/users.rb
```diff
@@ -90,13 +90,20 @@ def initialize(backend, logger)
         end
 
         dbus_method :SetFirstUser,
-          FUSER_SIG + ", out result:u" do |full_name, user_name, password, auto_login, data|
+          # It returns an Struct with the first field with the result of the operation as a boolean
+          # and the second parameter as an array of issues found in case of failure
```
mvidner, line 94: Well, thanks for an opportunity to explore multiple return values, but...
In this case I'd just use a single "list of problems" return value, where naturally an empty list means no problem = success.
(Consistency across the whole API also matters, I haven't checked that)

imobachgs: Yes, it might be simpler. Having that boolean value makes the return value more "semantic" to me. But I am not against taking a different approach. In any case, if we decide to go with just a list of errors, I would keep the JavaScript API as it is.

imobachgs: Additionally, let's say that you could create the user but return a list of warnings (not errors at all). In that case, checking the first value is pretty straightforward (and you do not need to check the kind of errors you got).

teclator: > Well, thanks for an opportunity to explore multiple return values, but... In this case I'd just use a single "list of problems" return value, where naturally an empty list means no problem = success. (Consistency across the whole API also matters, I haven't checked that)

Fine with me, in fact that is currently what is implemented in the assign_first_user backend method. @imobachgs what do you think?

teclator: Sorry I read the cache version and did saw that you already replied :)

imobachgs: > > Well, thanks for an opportunity to explore multiple return values, but... In this case I'd just use a single "list of problems" return value, where naturally an empty list means no problem = success. (Consistency across the whole API also matters, I haven't checked that)
> 
> Fine with me, in fact that is currently what is implemented in the assign_first_user backend method. @imobachgs what do you think?

I do not think it is needed, we do not need to expose exactly the same API that we have internally.

imobachgs (approved):

PR #339, opened 2022-11-29
Multi-arch support

jreidinger:
## Problem

for some products we need different configuration of repositories. Also some products are not available on all architectures, so they should not be offered there.

For details see https://github.com/yast/d-installer/issues/279

trello: https://trello.com/c/C3apZtHv/3231-3-d-installer-support-for-dynamic-configuration

## Solution

Allow architecture specific configuration in d-installer.yml file. It depends on ArchFilter from https://github.com/yast/yast-yast2/pull/1280

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ to have a closer look, then clone and configure the project as explained in the 
 
 * [multi-product](https://build.opensuse.org/package/binaries/YaST:Head:D-Installer/d-installer-live:default/images):
   it can be used to install different products, like *openSUSE Tumbleweed*, *Leap*, *Leap Micro* or
-  an experimental version of the *Adaptable Linux Platform Host OS*.
+  an experimental version of the *SUSE ALP ContainerHost OS*.
 * [ALP only](https://build.opensuse.org/package/binaries/YaST:Head:D-Installer/d-installer-live:ALP/images):
-  it only contains the definition for the experimental *Adaptable Linux Platform Host OS*, although
+  it only contains the definition for the experimental *SUSE ALP ContainerHost OS*, although
   the rest of the content is pretty much the same than the multi-product version.
 
 ### Manual Configuration
--- a/Rakefile
+++ b/Rakefile
@@ -77,7 +77,7 @@ task package: [] do
     sh "gem build #{name}.gemspec"
     gem = find_gem(dir).first
     gem2rpm = File.join(package_dir, "gem2rpm.yml")
-    sh "gem2rpm --config #{gem2rpm} #{gem} > package/#{package_name}.spec"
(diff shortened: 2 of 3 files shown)


service/test/fixtures/d-installer-archs.yaml
```diff
@@ -0,0 +1,291 @@
+products:
```
ancorgs, line 1: Do we need such a comprehensive configuration file just for testing archs? I looses focus and can tempt others to (ab)use the same fixture for other unrelated tests (like storage volumes). I would include the bare minimal content that relates to the test.

jreidinger: well, I try to be consistent with test for single product which is also quite big - https://github.com/yast/d-installer/blob/master/service/test/fixtures/d-installer-single.yaml
I will be more then happy to reduce size of fixture.

imobachgs: You can reduce the size, especially the storage one, but remember that all the keys should be included AFAIK.

service/lib/dinstaller/config.rb
```diff
@@ -76,19 +77,32 @@ def data
...
+    # list of available base products for current architecture
+    def products
+      return @products if @products
+      return [] unless @pure_data["products"]
+
+      # cannot use `data` here to avoid endless loop as in data we use
+      # pick_product that select product from products
+      @products = @pure_data["products"].select do |_key, value|
+        value["archs"].nil? ||
+          Yast2::ArchFilter.from_string(value["archs"]).match?
```
imobachgs, line 97: I am not against the current solution, but I wonder whether it might be better (or even possible) to removed/adjust the elements in a generic way. For instance, you are filtering the products here, so the `#products` method needs to know about that. What if it already gets a filtered representation?

service/test/fixtures/d-installer-archs.yaml
jreidinger: ok, I will try to reduce it as much as possible.

service/lib/dinstaller/config.rb
imobachgs: My point is that we could do it one level up, taking cmdline and other stuff into consideration. However, I am not sure whether it would work in all cases. Perhaps something that we can explore after the December Prototype.

imobachgs (approved):

ancorgs (changes requested):
We need to change the ALP repos. Check Slack

ancorgs (approved):
Now with the new ALP repos, it LGTM (and turns to be a nice real-life test for the new multi-arch support).

PR #357, opened 2022-12-05
Storage improvements

joseivanlopez:
## Problem

A new storage proposal UI was added in https://github.com/yast/d-installer/pull/321, but some unit tests were still missing.

## Solution

Add missing unit tests and improve storage client to avoid sending undefined settings on D-Bus.

## Testing

- Added new unit tests
- Tested manually

### DIFF
--- a/web/src/client/storage.test.js
+++ b/web/src/client/storage.test.js
@@ -25,6 +25,8 @@ import { StorageClient } from "./storage";
 // NOTE: should we export them?
 const STORAGE_PROPOSAL_IFACE = "org.opensuse.DInstaller.Storage.Proposal1";
 
+const calculateFn = jest.fn();
+
 const dbusClient = new DBusClient("");
 const storageProposalProxy = {
   wait: jest.fn(),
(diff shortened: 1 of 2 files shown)


web/src/client/storage.test.js
```diff
@@ -67,7 +87,76 @@ describe("#getProposal", () => {
...
     expect(proposal.volumes[1].mountPoint).toEqual("/test2");
   });
 });
+
+describe("#calculate", () => {
+  it("calculates a default proposal when no settings are given", async () => {
+    const client = new StorageClient(dbusClient);
+    await client.calculateProposal({});
+
+    expect(calculateFn).toHaveBeenCalledWith({});
```
dgdavid, line 115: NP: actually, we're testing that a new proposal was requested…. but not _the default_ one.

web/src/client/storage.js
```diff
@@ -107,24 +105,47 @@ class StorageBaseClient {
...
-        MountPoint: cockpit.variant("s", volume.mountPoint),
-        Encrypted: cockpit.variant("b", volume.encrypted),
-        FsType: cockpit.variant("s", volume.fsType),
-        MinSize: cockpit.variant("x", volume.minSize),
-        MaxSize: cockpit.variant("x", volume.maxSize),
-        FixedSizeLimits: cockpit.variant("b", volume.fixedSizeLimits),
-        Snapshots: cockpit.variant("b", volume.snapshots)
-      };
+    // Builds a new object without undefined attributes
+    const cleanObject = (object) => {
```
dgdavid, line 109: Question: is this method useful in other clients? If so, I'd move to a better location. Let's discuss it later.

dgdavid (approved):
Just two minor comments, but it looks good to me. Thank you!

dgdavid (changes requested):
Ouch! Tests failing :cry:

web/src/client/storage.test.js
joseivanlopez: A defaut proposal means a proposal without settings.

web/src/client/storage.js
joseivanlopez: I don't know. I would move it later if we find it useful in other places.

dgdavid (approved):
LGTM

PR #358, opened 2022-12-07
[service] Copy logs just before unmouting filesystem

dgdavid:
## Problem

Currently, the installation logs are copied too early to the target system. 

- https://trello.com/c/ihX3kxvR (internal link)

## Solution

To copy them just before unmounting the file systems.

## Testing

- Unit tests adapted
- Tested manually
   
  ```
  ➜ grep "Calling YaST client" ~/yast-installation-logs/YaST2/y2log
  2022-12-09 16:07:23 <1> localhost.localdomain(1601) [Interpreter] storage/manager.rb:71 Calling YaST client inst_prepdisk
  2022-12-09 16:07:26 <1> localhost.localdomain(1601) [Interpreter] storage/manager.rb:76 Calling YaST client inst_bootloader
  2022-12-09 16:14:46 <1> localhost.localdomain(1595) [Interpreter] dinstaller/network.rb:46 Calling YaST client save_network
  2022-12-09 16:15:06 <1> localhost.localdomain(1601) [Interpreter] storage/manager.rb:91 Calling YaST client snapshots_finish
  2022-12-09 16:15:07 <1> localhost.localdomain(1601) [Interpreter] storage/manager.rb:94 Calling YaST client copy_logs_finish
  2022-12-09 16:15:08 <1> localhost.localdomain(1601) [Interpreter] installation/copy_logs_finish.rb:94 Calling YaST client save_y2logs
  ```

### DIFF
--- a/service/lib/dinstaller/manager.rb
+++ b/service/lib/dinstaller/manager.rb
@@ -86,7 +86,7 @@ module DInstaller
     def install_phase
       installation_phase.install
 
-      start_progress(9)
+      start_progress(8)
 
       progress.step("Reading software repositories") do
         software.probe
@@ -107,7 +107,6 @@ module DInstaller
         progress.step("Writing Network Configuration") { network.install }
         progress.step("Saving Language Settings") { language.finish }
         progress.step("Writing repositories information") { software.finish }
-        progress.step("Copying logs") { copy_logs }
         progress.step("Finishing storage configuration") { storage.finish }
       end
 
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

joseivanlopez:
I would not move generic logic like copying logs to a specific service like storage. I understand we are doing it because we need to copy logs just before unmounting. But storage process should not be aware of that. I think we could find a better solution, otherwise we will start having dependencies between servicies without a good reason, and it would make everything more complex.

A quick solution could consist on splitting `storage#Finish` into two different calls: `storage#Finish` and `storage#Unmount`. And manager copies logs in middle of both calls:

~~~
progress.step("Finishing storage configuration") { storage.finish }
progress.step("Copying logs") { copy_logs }
progress.step("Unmounting devices") {  storage.unmount }
~~~

PR #359, opened 2022-12-07
Fix packaging

imobachgs:
* Fix dependency on D-Bus.
* Do not include the version in the D-Installer sources name.
* Fix cockpit-d-installer changelog.


dgdavid (approved):
LGTM

PR #360, opened 2022-12-12
[service] switch selinux policy for ALP to enforcing

fcrozat:
Switch selinux policy to enforcing for ALP


dgdavid (approved):
Sorry for the delay, @fcrozat.

Let's merge it.

PR #362, opened 2022-12-14
[service] Fix generic questions handling

imobachgs:
## Problem

Asking generic questions may cause issues, especially when they do not come from the storage service.

## Solution

* Add a missing require
* Do not assume that the Luks interface is always there (it causes ruby-dbus to crash).

## Testing

- *Added a new unit test*
- *Tested manually*


dgdavid (approved):
LGTM, but I have a _question_: does this change deserve an entry in the changelog?

joseivanlopez (commented):
Maybe it makes sense to have a specific client for `LuksActivationQuestion`. What do you think? It is a bit weird to have methods like `#password` that only applies when the dbus object implements certain iface.

joseivanlopez:
> LGTM, but I have a _question_: does this change deserve an entry in the changelog?

I think there would be a following PR which add questions for software issues.

joseivanlopez (approved):

imobachgs:
> Maybe it makes sense to have a specific client for `LuksActivationQuestion`. What do you think? It is a bit weird to have methods like `#password` that only applies when the dbus object implements certain iface. Anyway, I would add a TODO or FIXME to improve it in a separate PR.

I agree. We can address it later.

PR #363, opened 2022-12-14
Temporary hack to unlock the encryption using the TPM if possible 

ancorgs:
## Current status (before merging this pull request)

D-Installer already provides full-disk encryption (FDE). In all cases, unlocking of the devices during boot will be done by entering the passphrase chosen by the user during installation.

FDE

If the product being installed is based on ALP, then LUKS2 will be used. That makes it possible for the user to configure the system to unlock the encrypted devices during boot using a secret stored in the TPM (instead of the passphrase). That can be done manually after the system installation using the `fde-tools` developed by Olaf Kirch, but D-Installer offers no assistance on that regard.

Note there are certain conditions that need to be met in order to be able to use `fde-tools` to configure automatic LUKS unlocking based on the TPM:

- The system must be based on ALP, the needed tools are currently not available on other (open)SUSE distributions.
- The system must boot using UEFI.
- The system must contain a capable TPMv2 unit (TPM 1.X is not enough).
- The system must use LUKS2 encryption directly at partition level, that is, LVM cannot be used.

## Content of the patch

If we merge this PR, D-Installer will automatically take care of setting TPM unlocking of the devices.

For that, all the conditions mentioned above must be met. If not, the installer will do nothing about the TPM. On the one hand the system must use UEFI and contain a TPMv2 chip. On the other hand, the user must choose "Encrypt Devices" but must not choose "Use LVM". 

The process works like this:

- D-Installer prepares everything so TPM unlocking is automatically configured during the first boot of the installed system (since the final step of the configuration CANNOT be done during installation).
- When the installation finishes, the user must reboot and must ensure the system is booting directly from the disk containing the fresh new ALP system.
- The encryption passphrase will be asked to the user in that first boot.
- When the system finishes booting, unlocking via TPM would be transparently configured and the passphrase will not be needed in subsequent reboots.

## Caveats

- The first boot in the new system will still require the passphrase to be entered. This is quite unavoidable at the current state of the art.
- There is no trace in the UI about this change (intentionally, since this is a quick&dirty implementation). The behavior must be documented.
- It's crucial that the user boots directly from the new system device in the first boot. That must be documented.  If the system reboots again to the ISO and then the option "Boot from hard disk" is used, `fde-tools` will not be able to automatically configure the TPM unlocking during the first boot.

boot

## Testing

- Tested manually on a virtual machine with a virtual TPMv2
- Still pending: testing in a real bare-metal machine with a real TPMv2 chip
- Still pending: regression tests to ensure this does not affect any other scenario (installing non-ALP distributions, installing without encryption and so on).

### DIFF
--- a/service/etc/d-installer.yaml
+++ b/service/etc/d-installer.yaml
@@ -161,10 +161,13 @@ ALP:
     optional_patterns: null # no optional pattern shared
     mandatory_packages:
       - device-mapper # Apparently needed if devices at /dev/mapper are used at boot (eg. FDE)
+      - fde-tools # Needed for FDE with TPM, hardcoded here temporarily
+      - libtss2-tcti-device0 # Same than fde-tools
     optional_packages: null
     base_product: ALP
 
   security:
+    tpm_luks_open: true
     lsm: selinux
     available_lsms:
       # apparmor:
--- a/service/lib/dinstaller/storage/manager.rb
+++ b/service/lib/dinstaller/storage/manager.rb
@@ -84,7 +84,6 @@ module DInstaller
         on_target do
           progress.step("Writing Linux Security Modules configuration") { security.write }
           progress.step("Installing bootloader") do
-            hack_olaf_password
             ::Bootloader::FinishClient.new.write
           end
           progress.step("Configuring file systems snapshots") do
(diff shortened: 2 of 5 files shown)


dgdavid (approved):
LGTM

PR #371, opened 2022-12-15
Implement the ImportGpgKey callback

imobachgs:
## Problem

Although users are not expected to use their own repositories at this time, it is not uncommon that we need to try some package that lives in a development repository in OBS. In those cases, you might need to import the GPG key.

## Solution

This PR implements support to ask the user about importing a GPG key. The presentation could be improved, but let's do that after December's prototype.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

libzypp-callbacks-import-pgp-key

### DIFF
--- a/service/test/dinstaller/software/callbacks/signature_test.rb
+++ b/service/test/dinstaller/software/callbacks/signature_test.rb
@@ -71,8 +71,21 @@ describe DInstaller::Software::Callbacks::Signature do
       end
     end
 
-    context "when the repo information is not available"
-    it "includes a generic message containing the filename" do
+    context "when the repo information is not available" do
+      before do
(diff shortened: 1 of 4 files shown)


service/test/dinstaller/software/callbacks/signature_test.rb
```diff
@@ -71,12 +71,63 @@
...
+
+      it "returns true" do
+        expect(subject.import_gpg_key(key, 1)).to eq(true)
+      end
+    end
+
+    context "when the user answers :Skip" do
+      let(:answer) { :Skip }
+
+      it "returns true" do
```
dgdavid, line 119: _returns false_

joseivanlopez (approved):
LGTM

PR #374, opened 2022-12-15
Do not show WiFi configuration link if there are no WiFi devices

teclator:
## Problem

In #316 we added logic to hide the link for connecting to a WiFi network but only in case the WiFi was enable based on rfkill but if there are no WiFi devices at all then it does not makes sense to show the link.

## Solution

Do not show the link for connecting to a WiFi network if there are no WiFi devices.

Update: and also do not show an empty network section when there is no connection detected and no WiFi device available.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

| With a network connection | Whitout network connection |
| --- | --- |
| Screenshot from 2023-01-02 11-29-04 | Screenshot from 2023-01-02 11-26-53 |
| Screenshot from 2023-01-02 11-33-33 | Screenshot from 2023-01-02 11-36-22 |

### DIFF
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jan  2 11:46:37 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not show the link to configure wifi networks when there are no
+  wifi devices and also inform the user when no network connection
+  was detected. (gh#yast/d-installer#323).
+- Version 0.6.3
+
+-------------------------------------------------------------------
 Mon Jan  2 10:03:18 UTC 2023 - David Diaz <dgonzalez@suse.com>
 
 - Ensure custom fonts are including in the build
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -3,7 +3,7 @@ Mon Jan  2 11:46:37 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
 
 - Do not show the link to configure wifi networks when there are no
   wifi devices and also inform the user when no network connection
-  was detected. (gh#yast/d-installer#323).
+  was detected (gh#yast/d-installer#323).
 - Version 0.6.3
 
 -------------------------------------------------------------------
--- a/web/src/client/network/network_manager.js
+++ b/web/src/client/network/network_manager.js
@@ -40,6 +40,8 @@ const SERVICE_NAME = "org.freedesktop.NetworkManager";
 const IFACE = "org.freedesktop.NetworkManager";
 const SETTINGS_IFACE = "org.freedesktop.NetworkManager.Settings";
 const CONNECTION_IFACE = "org.freedesktop.NetworkManager.Settings.Connection";
+const DEVICE_IFACE = "org.freedesktop.NetworkManager.Device";
+const DEVICES_NAMESPACE = "/org/freedesktop/NetworkManager/Devices";
 const ACTIVE_CONNECTION_IFACE = "org.freedesktop.NetworkManager.Connection.Active";
 const ACTIVE_CONNECTION_NAMESPACE = "/org/freedesktop/NetworkManager/ActiveConnection";
 const IP4CONFIG_IFACE = "org.freedesktop.NetworkManager.IP4Config";
@@ -47,6 +49,7 @@ const IP4CONFIG_NAMESPACE = "/org/freedesktop/NetworkManager/IP4Config";
 const ACCESS_POINT_IFACE = "org.freedesktop.NetworkManager.AccessPoint";
 const ACCESS_POINT_NAMESPACE = "/org/freedesktop/NetworkManager/AccessPoint";
 const SETTINGS_NAMESPACE = "/org/freedesktop/NetworkManager/Settings";
+const NM_DEVICE_TYPE_WIFI = 2;
 
 const ApFlags = Object.freeze({
   NONE: 0x00000000,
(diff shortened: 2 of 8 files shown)


web/src/client/network/network_manager.js
```diff
@@ -500,19 +502,29 @@ class NetworkManagerAdapter {
...
-      hostname: settings?.Hostname || ""
-    };
+  /*
+  * Returns whether the system is able to scan wifi networks based on rfkill and the presence of
+  * some wifi device
+  *
+  * @return {boolean}
+  */
+  wirelessScanSupported() {
+    const wEnabled = !!(this.proxies.manager?.WirelessEnabled && this.proxies.manager?.WirelessHardwareEnabled);
```
imobachgs, line 512: It looks a little hard to read to me. I would rather prefer something like (not tested):

```javascript
if (this.proxies.manager === undefined) return false;

const { WirelessEnabled: wirelessEnabled, WirelessHardwareEnabled: wirelessHardwareEnabled} = this.proxies.manager;
const wifiDevices = Object.values(this.proxies.devices).filter(d => d.DeviceType === NM_DEVICE_TYPE_WIFI);
return WirelessEnabled && wirelessHardware && wifiDevices.length > 0;
```

So return earlier if the manager is unavailable, and make the conditions more readable. Additionally, why do you use "wireless" and "wifi"? It does not look consistent.

imobachgs (changes requested):
Now that we have plenty of time, I would take the opportunity to improve a few things. For instance, if the network is not initialized, we display [display nothing](https://github.com/yast/d-installer/blob/master/web/src/components/network/Network.test.jsx#L52), and it says nothing to the user. Additionally, if you do not have any network device (e.g., the network cable is disconnected and you do not have any wireless device), you basically see an empty section, right?

web/src/client/network/network_manager.js
teclator: Originally it represented the state of rfill enablement but that is not the case anymore, so, the method and variable should be renamed to WiFi as we are not interested in other kind of devices by now.

teclator: Adapted, I basically renamed the settings property and the method to hopefully a more meaningful name.

teclator:
@imobachgs I addressed the suggestions from our previous comment, so, only deciding what to show in case that the section is completely empty.

Screenshot from 2022-12-16 11-29-26

dgdavid:
> @imobachgs I addressed the suggestions from our previous comment, so, only deciding what to show in case that the section is completely empty.
> 

I know we (the YaST team) haven't talk about that yet, but in my opinion the summary mustn't contain an empty section. The section can be there or not, depending on the - not yet ready - configuration (depending on the product? on the architecture? both?), but no empty.

So, I'd add a _summary_ telling the user that D-Installer was not able to find neither, a wired network connection nor a network hardware for connection to a WiFi network.

teclator:
> > @imobachgs I addressed the suggestions from our previous comment, so, only deciding what to show in case that the section is completely empty.
> 
> I know we (the YaST team) haven't talk about that yet, but in my opinion the summary mustn't contain an empty section. The section can be there or not, depending on the - not yet ready - [product|architecture|workflow] configuration, but no empty.
> 
> So, I'd add a _summary_ telling the user that D-Installer was not able to find neither, a wired network connection nor a network hardware for connection to a WiFi network.

@dgdavid @imobachgs something like this?

web/src/client/network/network_manager.js
```diff
@@ -500,19 +503,40 @@ class NetworkManagerAdapter {
...
+  availableWifiDevices() {
+    return Object.values(this.proxies.devices).filter(d => d.DeviceType === NM_DEVICE_TYPE_WIFI);
+  }
+
+  /*
+  * Returns whether the system is able to scan wifi networks based on rfkill and the presence of
+  * some wifi device
+  *
+  * @return {boolean}
+  */
```
dgdavid, line 520: NP: looks to me that those comments are miss-aligned.

```js
/*
*
```

should actually be

```js
/* 
 *
```

web/src/client/network/network_manager.js
```diff
@@ -500,19 +503,40 @@ class NetworkManagerAdapter {
...
+
+    if (!manager) return false;
+    if (!(manager.WirelessEnabled && manager.WirelessHardwareEnabled)) return false;
+
+    return this.availableWifiDevices().length > 0;
   }
 
   /*
   * Returns NetworkManager general settings
+  *
```
dgdavid, line 532: NP: Ah, I see we have a mix on comments style :) So keep them as they are :man_shrugging:

teclator:
Screenshot from 2023-01-02 11-26-53

dgdavid (approved):
LGTM, thank you!

dgdavid (commented):
Do not forget to update the .changes file.

web/src/client/network/network_manager.js
imobachgs: I do not think these are mixed styles, just miss-aligned comments :) And sure we have more of them. 😞

web/package/cockpit-d-installer.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Jan  2 11:46:37 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not show the link to configure wifi networks when there are no
+  wifi devices and also inform the user when no network connection
+  was detected. (gh#yast/d-installer#323).
```
imobachgs, line 6: np: I would remove the first dot

imobachgs (approved):
Just a minor comment in the changes log. Otherwise, LGTM. Thanks!

web/package/cockpit-d-installer.changes
teclator: Ups, fixed :P

PR #375, opened 2022-12-16
Require NPM 18

mvidner:
## Problem

We spent some hours debugging a problem where the web frontend was not updated because the machine had NPM 17 which is no longer enough

## Solution

Tell zypper, and npm, to install v 18 at least

## Testing

- Tested manually: ran `./setup.sh` with 
```diff
-  "engines": { "node": ">=18" },
+  "engines": { "node": ">=1899" },
 ```

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: undefined,
npm WARN EBADENGINE   required: { node: '>=1899' },
npm WARN EBADENGINE   current: { node: 'v19.1.0', npm: '8.19.3' }
npm WARN EBADENGINE }
```

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,7 +4,7 @@
 # package. This script is supposed to run within a repository clone.
 
 sudo zypper --non-interactive install gcc gcc-c++ make openssl-devel ruby-devel \
-  npm git augeas-devel cockpit jemalloc-devel || exit 1
+  'npm>=18' git augeas-devel cockpit jemalloc-devel || exit 1
 
 sudo systemctl start cockpit
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: ["18.x"]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by mvidner** [comment]
@dgdavid says "it was a consequence of https://github.com/yast/d-installer/pull/331 I believe"

OK, anyway I hope `package.json` has a way to say "complain loud and clear unless npm >= 18"

**Comment by mvidner** [comment]
Ha!!
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines

> 
> You can specify the version of node that your stuff works on:
> 
>  
> { "engines" : { "node" : ">=0.10.3 <0.12" } }
> 
> And, like with dependencies, if you don't specify the version (or if you specify "*" as the version), then any version of node will do.

File: setup.sh [hunk 1]
```diff
@@ -4,7 +4,7 @@
 # package. This script is supposed to run within a repository clone.
 
 sudo zypper --non-interactive install gcc gcc-c++ make openssl-devel ruby-devel \
-  npm git augeas-devel cockpit jemalloc-devel || exit 1
+  'npm>=18' git augeas-devel cockpit jemalloc-devel || exit 1
```
mvidner, line 7:
```console
$ rpm -q --provides node19
...
nodejs-npm = 19.1.0
npm = 19.1.0   # match here
npm(npm) = 8.19.3
npm19 = 19.1.0-1.1
npm19(x86-64) = 19.1.0-1.1
```

dgdavid (approved):

PR #379, opened 2022-12-20
Provide logs

jreidinger:
## Problem

There is missing easy way to get logs from d-installer.

## Solution

Provide on dbus method that returns path with logs. For start it will just provide yast logs, but it can be extended in future with whatever is needed.

## Testing

- *Tested manually

## Future Steps

- CLI support
- web support to download them
- web log viewer


service/lib/dinstaller/dbus/manager.rb
```diff
@@ -60,6 +60,7 @@ def initialize(backend, logger)
         dbus_method(:Probe, "") { config_phase }
         dbus_method(:Commit, "") { install_phase }
         dbus_method(:CanInstall, "out result:b") { can_install? }
+        dbus_method(:ProvideLogs, "out tarball_filesystem_path:s, in user:s") { |u| provide_logs(u) }
```
lslezak, line 63: To make it extensible in the future I'd rather return a hash or array of strings here.

lslezak: But let's start with a simple API and extend it if needed later.

jreidinger: As said on IRC I find it hard to use when it is array or hash as usual action for logs will be display or download->upload to bzilla or gh issues, and in such case having more than one file is badly usable. What can be used in future is hash for another options if we want e.g. filter out debug output or logs just for specific parts like web logs, but I think we can extend it in future if the need grows. So far simple file that download/upload user should be enough

dgdavid:
Among other stuff, I have moved the button to a sidebar navigatio hidden by the _hamburguer menu pattern_.

| Sidebar hidden | Sidebar  visible |
|:-:|:-:|
| Screen Shot 2023-01-16 at 16 35 34 | Screen Shot 2023-01-16 at 16 35 26 |


I do not considered it finished since still pending some tasks apart from sidebar look&feel, which is out of the scope of this PR). Namely,

* Add tests for the `LogsButton` component
* Make the action of collecting and downloading links a single click shot.

dgdavid:
@jreidinger friendly ping. Just to let you know the above comment.

web/src/components/layout/Layout.jsx
```diff
@@ -31,10 +32,20 @@ const FooterActions = createTeleporter();
...
+ *
+ * To achieve an {@link https://gregberge.com/blog/react-scalable-layout scalable layout},
+ * it uses {@link https://reactjs.org/docs/portals.html React Portals} through
+ * {@link https://github.com/gregberge/react-teleporter react-teleporter}. In other words,
+ * it is mounted only once and gets influenced by other components by using the created
+ * slots (Title, PageIcon, MainActions, etc).
+ *
+ * So, please ensure that {@link test-utils!mockLayout } gets updated when adding or deleting
+ * slots here. It's needed in order to allow testing the output of components that interact
+ * with the layout using that mechanism.
```
dgdavid, line 48: Updated here due to changes done in the testing side. It adds comments missing at https://github.com/yast/d-installer/pull/216

imobachgs (commented):
In general, it looks good, although I do not feel comfortable with the naming at all.

web/src/components/core/Sidebar.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+// Test that opening the sidebar does not navigate to the initial route
+// This is achive by preventing the default link click behavior
+// Read https://testing-library.com/docs/dom-testing-library/api-events#fireevent and
+// https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented
+it("prevents the default event when the user click on the open link", async () => {
+  plainRender(<Sidebar />);
+  const link = await screen.findByLabelText(/Open/i);
+  const clickEvent = createEvent.click(link);
+  fireEvent(link, clickEvent);
+  expect(clickEvent.defaultPrevented).toBe(true);
```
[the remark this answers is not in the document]
dgdavid: Sure, this is a quite particular test that must not be used as an example for other components. But I want to keep it because the intention is to avoid a _regression_ when opening the sidebar: I myself faced the problem of accidentally navigating from _Product Selection_ page to the _Overview_ when opening the sidebar.

Hopefully, this will change in the near future, after sanitizing the correct use of  `button` or `hiperlink`.

> In no particular order, below are the next UI changes I'd like to make
> 
>    * To fix the misuse of buttons/links
> 
>       https://ashleemboyer.com/blog/should-i-use-a-button-or-a-link

<sup>^^^ An excerpt from https://github.com/yast/d-installer/pull/391 description. ^^^</sup>

web/src/components/layout/Layout.jsx
```diff
@@ -31,10 +32,20 @@ const FooterActions = createTeleporter();
...
 /**
- * D-Installer main layout component.
  *
- * It displays the content in a single vertical responsive column with fixed
- * header and footer.
+ * The D-Installer main layout component.
+ *
+ * It displays the content in a single column with a fixed header and footer.
+ *
+ * To achieve an {@link https://gregberge.com/blog/react-scalable-layout scalable layout},
```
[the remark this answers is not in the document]
dgdavid: You're right. Although I bet I'll make the same mistake in the future.  Thanks :)

imobachgs (approved):

PR #380, opened 2022-12-20
Refresh image configuration

imobachgs:
* Include `root.tar` and `etc.tar` as OBS sources.
* Use `default.tar.xz` and `alp.tar.xz` for flavors configuration.
* Make D-Installer the default Cockpit module.
* Add missing packages.
* Bump to version 0.6.2.

The reason for many of these changes is to make it easier to introduce temporary changes (like switching to the PUBLISH URL) without having to change the repository.


ancorgs (approved):
LGTM

dgdavid (approved):
LGTM, BTW, I guess this closes https://github.com/yast/d-installer/issues/365, right?

PR #381, opened 2022-12-21
initial software validation code

jreidinger:
## Problem

When software proposal contains any issues, user does not know it in advance and it can cause that he things that installation succeed, even when there are serious issues.

## Solution

Add validation that checks if software proposal and solver run is without critical issues.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

| Without errors | Not expanded errors | Expanded errors |
|-|-|-|
|size_sum|size_sum_errors |Screen Shot 2022-12-30 at 14 51 39-fullpage |

## TODO

- [x] test changing product if it changes also product
- [ ] check repository erros if it is properly reported
- [x] add to software section info about size of installed packages
- [x] remove that Please manually select


service/lib/dinstaller/software/manager.rb
```diff
@@ -112,10 +113,14 @@ def propose
...
-        solve_dependencies
+        deps_result = solve_dependencies
 
-        # do not return proposal hash, so intentional nil here
-        nil
+        proposal_result(proposal, deps_result)
+      end
+
+      def validate
+        msgs = propose
```
joseivanlopez, line 122: I am not sure if calling to `#propose` unconditionally is a good idea. If a D-Bus client calls to Manager#Propose and then Manager#Errors in a row, then `#propose` would be called twice without any real need.

Just as idea: maybe the callbacks could indicate to Manager that the proposal has to be recalculated. So, `#validate` would only call to `#propose` if needed.

jreidinger: OK, I changed code and now it is called only when needed with probed check. IT also helps with not showing intermediate errors that will alter gone.

service/lib/dinstaller/dbus/software/proposal.rb
```diff
@@ -54,6 +55,7 @@ def initialize(logger)
           dbus_method :AddResolvables,
             "in Id:s, in Type:y, in Resolvables:as, in Optional:b" do |id, type, resolvables, opt|
             Yast::PackagesProposal.AddResolvables(id, TYPES[type], resolvables, optional: opt)
+            changed
```
dgdavid, line 58: NP: This looks to me more like these bang methods. I mean, `change!` instead of `changed`.

dgdavid: Or maybe a `call_on_change_callbacks`, which for sure is too long but more self-explanatory.

jreidinger:
@imobachgs @joseivanlopez this PR is now ready for review

@joseivanlopez I am not sure if some adaptation is needed in CLI as I failed to find where validation is done in CLI

dgdavid (commented):
Just a little neat picking comment from my side :sweat_smile: 

For the UI, we should find a better pattern/strategy/approach to display these errors rather than just having an empty section. Another story would be to have an overview/SoftwareSection displaying more information about the software installed and, if necessary, any errors. In any case, please don't block the PR for that reason. Let's move on and try to come up with a proposal that could be implemented in a separate PR.

As for the backend, I'd prefer a review from another team member more involved in the backend right now. So, let's wait a few days :wink:

service/lib/dinstaller/dbus/software/proposal.rb
jreidinger: well, I would definitively like `change!` as it emphasize to me that it change something which is not true. It is more about notification. the other one is a bit long for me. What about `notify_change` or with bang like `notify_change!`

dgdavid: >I would definitively like `change!`

By the context, I guess you meant you don't like.

>  What about notify_change or with bang like notify_change!

Both looks better to me than `changed`. But maybe you can wait for another review to change (or not) it. More brains, more ideas :)

joseivanlopez:
> @joseivanlopez I am not sure if some adaptation is needed in CLI as I failed to find where validation is done in CLI

I wouldn't invest time on adapting the CLI. We are redefining it completely.

ancorgs:
Shouldn't the summary contain something? A simple sentence like this would make it, in my opinion:

```368 packages will be installed```

**Comment by imobachgs** [comment]
> Shouldn't the summary contain something? A simple sentence like this would make it, in my opinion:
> 
> `368 packages will be installed`

Definitely, the summary should contain something. But I do not think the amount of packages is useful at all (perhaps the size might be better). Not to mention that we already have a button to change the product to install that *might* be in this section.

**Review by imobachgs — CHANGES_REQUESTED** [review]
Code-wise, it looks good, thanks! However, I am not that sure about the approach. We can merge this PR as a first step (once the changelogs are updated), but we should refine this solution as soon as possible. IMHO, we should not ship the UI as it is right now.

As a proposal, the software section might contain the product you are about to install and the link to change it. Additionally, displaying the installation size or other stuff might be useful. And, of course, I would move this section to a more prominent place (e.g., after the language section).

Last but not least, I would revisit the idea of using a skeleton in the software section. Reading the repositories might take some time, and the skeleton is not that useful. But I do not consider this issue as urgnet (yet).

**Comment by dgdavid** [comment]
> As a proposal, the software section might contain the product you are about to install and the link to change it. Additionally, displaying the installation size or other stuff might be useful. And, of course, I would move this section to a more prominent place (e.g., after the language section).

As a starting point, I like the idea. Thanks

**Comment by dgdavid** [comment]
> Code-wise, it looks good, thanks! However, I am not that sure about the approach. We can merge this PR as a first step (once the changelogs are updated), but we should refine this solution as soon as possible. IMHO, we should not ship the UI as it is right now.
> 
> As a proposal, the software section might contain the product you are about to install and the link to change it. Additionally, displaying the installation size or other stuff might be useful. And, of course, I would move this section to a more prominent place (e.g., after the language section).
> 
> Last but not least, I would revisit the idea of using a skeleton in the software section. Reading the repositories might take some time, and the skeleton is not that useful. But I do not consider this issue as urgnet (yet).

Somehow, I agree. In fact, it's in my _personal roadmap_ to improve a bit how sections and loaders are rendered... because we have too many _skeletons blinks_ IMHO.

File: web/package/cockpit-d-installer.changes [hunk 3]
```diff
@@ -16,6 +22,7 @@ Mon Jan  2 10:03:18 UTC 2023 - David Diaz <dgonzalez@suse.com>
 
 - Ensure custom fonts are including in the build
   (gh#yast/d-installer#385).
+>>>>>>> origin/master
```
lslezak, line 25: Some reminder after a merge, please remove it.

lslezak (changes requested):
Conflict mark in the code

web/package/cockpit-d-installer.changes
jreidinger: @lslezak thanks, fixed

service/lib/dinstaller/dbus/software/proposal.rb
```diff
@@ -64,18 +66,28 @@ def initialize(logger)
...
+        def on_change(&block)
+          @on_change_callbacks << block
+        end
+
       private
 
         # @return [Logger]
         attr_reader :logger
+
+        def notify_change!
```
joseivanlopez, line 88: NP: Why a bump method?

service/lib/dinstaller/manager.rb
```diff
@@ -89,7 +90,6 @@ def install_phase
       start_progress(8)
 
       progress.step("Reading software repositories") do
```
joseivanlopez, line 92: I woud say this step is not true anymore.

service/lib/dinstaller/software/manager.rb
```diff
@@ -112,10 +116,17 @@ def propose
...
-        # do not return proposal hash, so intentional nil here
-        nil
+        proposal_result(proposal, deps_result)
+      end
+
+      def validate
+        # validation without probing does not make sense and produces false errors
+        return [] unless @probed
+
+        msgs = propose
```
joseivanlopez, line 128: I would say `#propose` is still performed with each call to `#validate`.

service/lib/dinstaller/dbus/software/manager.rb
```diff
@@ -78,6 +81,8 @@ def initialize(backend, logger)
             backend.package_installed?(name)
           end
 
+          dbus_method(:UsedDiskSpace, "out SpaceSize:s") { backend.used_disk_space }
```
joseivanlopez, line 84: Shouldn't this belong to the proposal iface?

web/src/components/overview/SoftwareSection.jsx
```diff
@@ -0,0 +1,114 @@
...
+    };
+
+    updateProposal();
+  }, [client.software, cancellablePromise, state.busy]);
+
+  const errors = showErrors ? state.errors : [];
+
+  const UsedSize = () => {
+    if (state.size === "") {
+      return <InstallerSkeleton lines={1} />;
```
joseivanlopez, line 88: Skeleton?

I would do this:

~~~
if proposal -> "Install 456 packages (876 MiB)"
else --> "No packages to install"
~~~

Of course, a validation error will be present in case of 0 packages:

~~~
Software
3 errors found
No packages to install
~~~

service/lib/dinstaller/dbus/software/manager.rb
joseivanlopez: Moreover, I would offer more info, for example the number of packages to install.

web/src/components/overview/SoftwareSection.jsx
joseivanlopez: BTW, in case there is no network, should the software be proposed again after a connection is set?

jreidinger: @joseivanlopez see https://github.com/yast/d-installer/issues/388 I already mention it there.

jreidinger: We agreed that we will handle it in separate PR.

service/lib/dinstaller/dbus/software/proposal.rb
jreidinger: see discussion at https://github.com/yast/d-installer/pull/381#discussion_r1059424403

service/lib/dinstaller/manager.rb
jreidinger: right, it should be already read.

service/lib/dinstaller/dbus/software/manager.rb
jreidinger: well, so far we agreed on that used space when we discuss it. I am sure in future we will need to have there more. About manager versus proposal, I am not sure where it belongs as I see reasons for both and having it in manager is easier for me for implementation in frontend as frontend so far does not use proposal.

web/src/components/overview/SoftwareSection.jsx
jreidinger: State.size == "" does not mean that there is no package to install, but it means some issue when couting it like if probe/proposal is not yet done, but it is also not busy as storage proposal is not yet finish, so software proposal does not even start.

service/lib/dinstaller/software/manager.rb
jreidinger: well, it is only for validation called after probing. In that case something should trigger it, as validation has its properties and only if something invalidate it, it is recomputed. And it is invalidated from software proposal which change required packages or after change of product. So when I test it, I do not see unnecessary calls to validate as it is already cached and I need to reset it appropriate places with `update_validation`. So in fact my issue was to reset cache and not to cache more.

service/lib/dinstaller/dbus/software/proposal.rb
joseivanlopez: I like what is suggested here https://thecodest.co/blog/ruby-software-development-what-are-the-bang-methods-and-when-to-create-them/: "There is no point in creating a bang method if its counterpart without an exclamation mark does not exist. There are methods that are dangerous because they change the object or raise an exception, and yet they don't have an exclamation mark at the end of the name. Creating a bang method with no equivalent would confuse a programmer using your code."

jreidinger: It is completely fine for me if we use it consistently :)

service/lib/dinstaller/software/manager.rb
joseivanlopez: Ok, I see. The only problem I see is when calling to `software.manager.propose` D-Bus method. In that case, the proposal is done twice, right?

joseivanlopez (commented):
For the records:

* If software probing/proposal failed, it  should be calculated after activating a network connection.
* Storage adds required packages during the installation phase. Therefore, the information reported by software manager is missing the storage packages. Storage could add its resolvables just after calculating a proposal.

jreidinger:
@joseivanlopez btw we need also manual way to retrigger proposal as e.g. what happened to me was wrong firewall configuration cause blocking of traffic to download.opensuse.org, so after fix I think it will be useful to have simple way to retry.

service/lib/dinstaller/software/manager.rb
jreidinger: yeap, I do not notice it much, but should check logs for timestamp, so we have idea how long it take.

joseivanlopez:
* https://github.com/yast/d-installer/issues/388
* https://github.com/yast/d-installer/issues/396

imobachgs (approved):

lslezak (approved):

PR #390, opened 2023-01-09
CLI guidelines

joseivanlopez:
## Problem

D-Installer already offers an initial CLI prototype for managing and driving the installation process. That CLI was created as a proof of concept, and its current API needs some refactoring:

* It sometimes uses verbs for the sub-command action (e.g., `dinstallerctl user clear`), and for other sub-commands adjectives or noums are used (e.g., `dinstallerctl language selected <id>`).
* There is a sub-command per each area, for example `dinstallerctl language`, `dinstallerctl software`, `dinstallerctl storage`, etc. Having a sub-command for each area is not bad per se, but for some areas like storage the sub-command could grow with too many actions and options.

## Solution

A CLI guideline document is provided, which defines the new syntax. That new API was proposed as result of the discussion here: https://gist.github.com/joseivanlopez/808c2be0cf668b4b457fc5d9ec20dc73.

There are some details to decide later, for example, how to provide a non-interative mode.

### DIFF
--- a/doc/cli_guidelines.md
+++ b/doc/cli_guidelines.md
@@ -1,0 +1,196 @@
+# D-Installer CLI Guidelines
+
+This document defines the syntax for the D-Installer CLI. For more CLI specific aspects like return values, flags, etc, please refer to [clig.dev](https://clig.dev/). Guidelines from clig.dev are agnostic about programming languages and tooling in general, and it can be perfectly used as reference for the D-Installer CLI.
+
+## CLI Syntax
+
+Note: The syntax presented here was previously discussed in this [document](https://gist.github.com/joseivanlopez/808c2be0cf668b4b457fc5d9ec20dc73).
+
+The installation settings are represented by a YAML structure, and the CLI is defined as a set of generic sub-commands and verbs that allow to edit any YAML value in a standard way.
+
+The CLI offers a `config` sub-command for editing the YAML config. The `config` sub-command has verbs for the following actions:
+
+* To load a YAML config file with the values for the installation.
+* To edit any value of the config without loading a new complete file again.
+* To show the current config for the installation.
+* To check the current config.
+
+Moreover, the CLI also offers sub-commands for these actions:
+
+* To ask for the possible values that can be used for some setting (e.g., list of available products).
+* To start and abort the installation.
+* To see the installation status.
+* To answers questions.
+
+### Sub-commands and Verbs
+
+This is the list of all sub-commands and verbs:
+
+~~~
+$ dinstaller install
+Starts the installation.
+
+$ dinstaller abort
+Aborts the installation.
+
+$ dinstaller status
+Prints the current status of the installation process and informs about pending actions (e.g., if there are questions waiting to be answered, if a product is not selected yet, etc).
+
+$ dinstaller watch
+Prints messages from the installation process (e.g., progress, questions, etc).
+
+$ dinstaller config load <file>
+Loads installation config from a YAML file, keeping the rest of the config as it is.
+
+$ dinstaller config show [<key>]
+Prints the current installation config in YAML format. If a <key> is given, then it only prints the content for the given key.
+
+$ dinstaller config set <key>=<value> ...
+Sets a config value for the given key.
+
+$ dinstaller config unset <key>
+Removes the current value for the given key.
+
+$ dinstaller config reset [<key>]
+Sets the default value for the given <key>. If no key is given, then the whole config is reset.
+
+$ dinstaller config add <list-key> [<key>=]<value> ...
+Adds a new entry with all the given key-value pairs to a config list. The key is omitted for a list of scalar values (e.g., languages).
+
+$ dinstaller config delete <list-key> [<key>=]<value> ...
+Deletes any entry matching all the given key-value pairs from a config list. The key is omitted for a list of scalar values.
+
+$ dinstaller config check
+Validates the config and prints errors
+
+$ dinstaller config info <key> [<value>]
+Prints info about the given key. If no value is given, then it prints what values are admitted by the given key. If a value is given, then it shows extra info about such a value.
+
+$ dinstaller summary [<section>]
+Prints a summary with the actions to perform in the system. If a section is given (e.g., storage, software, ...), then it only shows the section summary.
+
+$ dinstaller questions
+Prints questions and allows to answer them.
+
+~~~
+
+In those commands `<key>` represents a YAML key from the config structure and `<value>` is the value associated to the given key.
+
+### YAML Config
+
+The config settings are defined by this YAML structure:
+
+~~~
+---
+product: "Tumbleweed"
+
+languages:
+  - "es_ES"
+  - "en_US"
+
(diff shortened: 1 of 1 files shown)


doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+* To edit any value of the config without loading a new complete file again.
+* To show the current config for the installation.
+* To check the current config.
+
+Moreover, the CLI also offers sub-commands for these actions:
+
+* To ask for the possible values that can be used for some setting (e.g., list of available products).
+* To start and abort the installation.
+* To see the installation status.
+* To answers questions.
```
jreidinger, line 23: I would probably add here also subcommand for validation. So you can ask if current config is valid and if not, to show details like wrong software selection or users issue.

doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+Prints messages from the installation process (e.g., progress, questions, etc).
+
+$ dinstaller config load <file>
+Loads installation config from a YAML file, keeping the rest of the config as it is.
+
+$ dinstaller config show [<key>]
+Prints the current installation config in YAML format. If a <key> is given, then it only prints the content for the given key.
+
+$ dinstaller config set <key>=<value> ...
+Sets a config value for the given key.
```
jreidinger, line 49: in general I like this structure as it looks like man page or help. But for both case I think we should mention nested values separated dot here and not few lines below under yaml config section.

doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+Sets the default value for the given <key>. If no key is given, then the whole config is reset.
+
+$ dinstaller config add <list-key> [<key>=]<value> ...
+Adds a new entry with all the given key-value pairs to a config list. The key is omitted for a list of scalar values (e.g., languages).
+
+$ dinstaller config delete <list-key> [<key>=]<value> ...
+Deletes any entry matching all the given key-value pairs from a config list. The key is omitted for a list of scalar values.
+
+$ dinstaller config check
+Validates the config and prints errors
```
jreidinger, line 64: is this same as validation interface? or does it check only yaml file?

doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+$ dinstaller config set storage.lvm=false
+$ dinstaller config set storage.encryption_password=12345
+
+# Add and delete candidate devices
+$ dinstaller config add storage.candidate_devices /dev/sda
+$ dinstaller config delete storage.candidate_devices /dev/sdb
+
+# Add and delete storage volumes
+$ dinstaller config add storage.volumes mountpoint=/ minsize=10GiB
+$ dinstaller config delete storage.volumes mountpoint=/home
```
jreidinger, line 147: to be honest I find this at least a bit confusing. As it can be read as "delete volume where mountpoint is /home" or "delete from volumes key mountpoint with value /home". I know it is described that it should be the first case, but if I see command without reading documentation I will be confused.

doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+Sometimes answering pending questions is required before performing the requested command. For example, for single product live images the storage proposal is automatically done (the target product is already known). If some questions were emitted during the process, then they have to be answered before continuing using the CLI. Commands would show a warning to inform about the situation and how to proceed:
+
+~~~
+$ dinstaller config show
+There are pending questions. Please, answer questions first: dinstaller questions.
+~~~
+
+## Non Interactive Mode
+
+Commands should offer a `--non-interactive` option to make scripting possible. The non interactive mode should allow answering questions automatically.
```
jreidinger, line 194: :+1:

doc/cli_guidelines.md
joseivanlopez: There is already a command for that in the list of commands: `dinstaller config check`.

doc/cli_guidelines.md
joseivanlopez: Yes, the idea is to show the current validations errors reported by the backend. For YAML validation we can report errors when the file is loaded with `dinstaller config load <file>`.

doc/cli_guidelines.md
joseivanlopez: Umm, `#add` and `#delete` verbs are intended to add/delete an element to/from a list. I see your point but I am not able to propose something more intuitive. Any idea? Anyway, elements in a list are "atomic", you add or delete them, there is no way to modify a single attribute.

doc/cli_guidelines.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Dump config into a file
+$ dinstaller config show > ~/config.yaml
+
+# Show info of a key
+$ dinstaller config info storage.candidate_devices
+$ dinstaller config info languages
+
+# Show info of a specific value
+$ dinstaller config info storage.candidate_devices /dev/sda
```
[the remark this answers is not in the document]
joseivanlopez: The idea is to show extra info. For example, for a disk we can print its size, operating systems it contains, UUID, free size, etc. For a product, it could show the product description, etc.

doc/cli_guidelines.md
```diff
@@ -24,6 +24,8 @@ Moreover, the CLI also offers sub-commands for these actions:
 
 ### Sub-commands and Verbs
 
+In the following commands `<key>` represents a YAML key from the config structure and `<value>` is the value associated to the given key. Nested keys are dot-separated (e.g., `user.name`).
```
jreidinger, line 27: :+1:

doc/cli_guidelines.md
jreidinger: well, to be honest I do not like much idea of having it hidden behind config check, as it can be independent of config. I probably prefer to have `dinstaller check` which prints details about errors as I expect usual workflow is
```
dinstaller config load abc.yaml
dinstaller install
dinstaller check # when install is refused or failed
```

doc/cli_guidelines.md
```diff
@@ -0,0 +1,197 @@
...
+* To answers questions.
+
+### Sub-commands and Verbs
+
+In the following commands `<key>` represents a YAML key from the config structure and `<value>` is the value associated to the given key. Nested keys are dot-separated (e.g., `user.name`).
+
+This is the list of all sub-commands and verbs:
+
+~~~
+$ dinstaller install
```
jreidinger, line 32: One idea that I have during our planning meeting. Would it be possible to allow on CLI separation of steps? I mean something like `dinstaller install --partitioning` or `dinstaller install --software`?
With this ability it will be easy to have scripts/hooks with CLI based API. as you will have list of commands including whatever you need between steps.

jreidinger: yeap, that check makes sense and we probably will need to adapt also backend, but it looks flexible solution for me. @imobachgs what do you think?

jreidinger: and btw with that split we can allow to use d-installer as e.g. partioning only tool

jreidinger (approved):

PR #391, opened 2023-01-10
[web] Rework UI internals

dgdavid:
Over Christmas, I started to rework the web UI internals to simplify how we build it.

Among other topics, I had concerns about the amount of DOM nodes used for laying out the UI[1]. Of course, browsers are so performant these days, and the D-Installer's web UI is far from troublesome in that regard. In the end, it is rather simple compared to the UI of medium/large web applications. However, I think it's a worthwhile idea to stop using layout components (PF4/Split, PF4/Stack, and friends) and rely more on plain CSS.

I also took the opportunity to start applying the [CUBE CSS](https://cube.fyi/) methodology after watching the [Be the browser’s mentor, not its micromanager](https://www.youtube.com/watch?v=5uhIiI9Ld5M) talk from its author, Andy Bell.

It is true that PatternFly, the component library we're using, uses BEM instead. Still, CUBE looks simpler to me (yet somehow BEM compatible). So why not give it a try?

As a whole, there is still work to be done. However, this PR can be seen as a first step[2] for rebuilding the UI internals by introducing CUBE CSS, leaving behind the layout components and using PatternFly strictly as a component library.

## Screenshots

| Before | After |
|-|-|
|Screen Shot 2023-01-10 at 19 39 54-fullpage |Screen Shot 2023-01-11 at 09 41 28-fullpage |

## Some _metrics_

<details>
<summary>Click to show/hide some not so relevant numbers</summary>

---

| Test | Before | After |
|-|:-:|:-:|
|  document.body<br />  &nbsp;&nbsp;.getElementsByTagName("*")<br /> &nbsp;&nbsp;.length | 118 | 74 |
| domVisualization<br />  (thanks to [domVisualizer](http://bioub.github.io/dom-visualizer/))| domVisualization_before | domVisualization_after|

</details>

## Next steps

In no particular order, below are the next UI changes I'd like to make

* To fix the misuse of buttons/links 

  https://ashleemboyer.com/blog/should-i-use-a-button-or-a-link

* Use the section headers as links to the corresponding page

  After playing with it, I found that using the underlined link[3] is the best, more intuitive way for the user to know there is a page. Even more considering that all sections in the summary will potentially have their Page counterpart.

* Better CUBE  CSS implementation (along with better CSS variables naming and all related stuff for having a themeable UI)

* Start using a fluid type scale

  https://www.fluid-type-scale.com

* Introduce Design Tokens to simplify the process of creating themes or just applying changes (not sure, need to read more about it)

  https://tr.designtokens.org/format/

  https://tokencss.com/

  https://equinusocio.dev/blog/design-tokens-misconceptions-and-theme-tokens/

  ...

---

[1] Premature optimization, I know. But I felt that we are/were in a phase in which we can explore and address these topics more easily.

[2] Or even an attempt, since it might not be the best/final solution. After all, I'm (re)learning by doing :wink:

[3] it can be a highlighted link instead, but only if all section headers are links…. Anyway, something to discuss later in a different PR.

### DIFF
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 11 11:54:29 UTC 2023 - David Diaz <dgonzalez@suse.com>
+
+- Rework UI internals by using plain CSS as much as possible for
+  building the layout instead of relying on wrapper components
+  (gd#yast/d-installer#391).
+
+-------------------------------------------------------------------
 Tue Jan 10 10:30:05 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Connect to the dedicated D-Bus server (gh#yast/d-installer#384).


web/src/assets/styles/blocks.scss
```diff
@@ -0,0 +1,52 @@
+// Standar section
```
imobachgs, line 1: typo: standard

web/src/assets/styles/global.scss
```diff
@@ -0,0 +1,49 @@
...
+
+table {
+  border-collapse: collapse;
+}
+
+th {
+  text-align: start;
+}
+
+// :focus {
```
imobachgs, line 32: Do we need to keep these lines?

web/src/assets/styles/patternfly-overrides.scss
```diff
@@ -0,0 +1,114 @@
...
+
+  // Font size
+  --pf-global--FontSize--4xl: 2em;
+  --pf-global--FontSize--3xl: 1.75em;
+  --pf-global--FontSize--2xl: 1.5em;
+  --pf-global--FontSize--xl: 1.3em;
+  --pf-global--FontSize--lg: 1.15em;
+  --pf-global--FontSize--md: var(--fs-base);
+  // TODO: find better equivalences for sm and xs
+  // --pf-global--FontSize--xs: #{$pf-global--FontSize--xs};
```
imobachgs, line 26: Do we need to keep these lines?

web/src/components/core/InstallationFinished.jsx
```diff
@@ -46,24 +46,22 @@ function InstallationFinished() {
...
+        <Title headingLevel="h2" size="4xl">
+          Congratulations!
+        </Title>
+        <EmptyStateBody className="pf-c-content">
+          <div>
+            <Text>The installation on your machine is complete.</Text>
+            <Text>
+              At this point you can 'Reboot' the machine to log in to the new system.
+            </Text>
+            <Text>Have a lot of fun! Your openSUSE Development Team.</Text>
```
imobachgs, line 60: Just a thought: we might consider replacing the "openSUSE Development Team" (it is not only about installing openSUSE).

web/src/components/layout/Center.jsx
```diff
@@ -20,10 +20,36 @@
...
+ *   .parent:has(.vertically-centered) {
+ *     display: grid;
+ *     place-items: center;
+ *     block-size: 100%;
+ *   }
+ *
+ *   .vertically-centered { inline-size: 100%; }
+ *
+ *   We can use \@support CSS at rule and use a workaround when :has not available, but somehow
+ *   prefer waiting until Firefox get support.
```
imobachgs, line 39: typo: gets

web/src/components/network/AddressesDataList.jsx
```diff
@@ -118,24 +114,16 @@ export default function AddressesDataList({
...
-      <StackItem>
-        <DataList isCompact gridBreakpoint="none" title="Addresses data list">
-          {addresses.map(address => renderAddress(address))}
-        </DataList>
-      </StackItem>
-    </Stack>
+    <>
+      <div className="split justify-between">
+        <FormLabel isRequired={!allowEmpty}>Addresses</FormLabel>
+        <Button isSmall variant="secondary" className="btn-sm" onClick={() => addAddress()}>
```
imobachgs, line 120: If you prefer, I think you can write: `onClick={addAddress}`. However, it is out of the scope of this PR.

imobachgs (commented):
Great work! Just a few typos. Otherwise, it looks good to me.

web/src/assets/styles/patternfly-overrides.scss
dgdavid: Actually not, good point. Thanks.

web/src/components/core/InstallationFinished.jsx
dgdavid: I agree. Let's think about it and sent a PR with the new text.

web/src/components/network/AddressesDataList.jsx
dgdavid: Yep, you're right. To be honest, I didn't realize about it because I was focus in the UI topic. Let's improve it in another PR.

Anyway, it's interesting to see that there is a _className="btn-sm"_ leftover there. And it was added by me in https://github.com/yast/d-installer/pull/391, probably because it's a component borrowed from cockpit-wicked where we [introduced that CSS class](https://github.com/openSUSE/cockpit-wicked/blob/f26f4de86b258ccabc77857a179a93c37746695f/src/app.scss#L46-L49) :man_shrugging: :man_shrugging: :man_shrugging:

dgdavid: Finally, I have changed those onClick function to their shorter versions since I had to edit the file for removing the leftover CSS.

imobachgs (approved):

PR #392, opened 2023-01-11
[web] Add function for mocking components

dgdavid:
## Problem

After cutting down wrappers for laying out the UI in #391, some unit tests using mocked components started to complain about _`Unable to find an element with the text: ...`_.

This happens because we're using [getByText](https://testing-library.com/docs/queries/bytext) Testing Libarry's query with the exact text we're looking for. The text is still there, but it is no longer "alone". I.e., it's not explicitly wrapped by a layout node and thus, the query throws an error.

An example can be found at this moment after rebasing #381 on top of master for solving some conflicts.

```
FAIL src/components/overview/Overview.test.jsx
  ● renders the Overview and the Install button

    Unable to find an element with the text: Users Configuration. [...]

    ...

              <div
                class="stack content"
              >
                Network Configuration
              </div>
            </section>
            <div>
              Storage Section
            </div>
            Software Section
            ,
            Users Configuration
          </main>
``` 

<sup> See https://github.com/yast/d-installer/actions/runs/3893089545/jobs/6645272239#step:7:312 (as long as the link remains available)</sup>

## Solution

The simplest and direct solution is to manually fix failed tests by wrapping mocked content in a `div`, `span`, or whatever other HTML element of our taste. At the end, it's a mock. But it's a suboptimal _fix_.

Instead, the solution proposed here is to use a _util function_ for mocking components. This give us a _unified_ way of producing mocked components and can ensure that their content is somehow isolated from other nodes, making _getByText_ queries happy again.

## Testing

- `npm run test` ends successfully.

## Notes

* We cannot use the function directly in the [`jest.mock`](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options)  call because its second argument must be an inline function.
  
  > FAIL  src/App.test.jsx
  > ● Test suite failed to run
  > 
  >  &nbsp;&nbsp;&nbsp;&nbsp;TypeError: src/App.test.jsx: The second argument of `jest.mock`must be an inline function.
  

* A better name for the function is, as always, more than welcome. But it must be start with `mock`.
  
  
  > FAIL  src/App.test.jsx
  > ● Test suite failed to run
  > 
  >  &nbsp;&nbsp;&nbsp;&nbsp;ReferenceError: src/App.test.jsx: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
  >  &nbsp;&nbsp;&nbsp;&nbsp;Invalid variable access: fakeComponent
  > &nbsp;&nbsp;&nbsp;&nbsp;Allowed objects: [...]
  >  &nbsp;&nbsp;&nbsp;&nbsp;Note: This is a precaution to guard against uninitialized mock variables. If it is ensured that the mock is required lazily, variable names prefixed with `mock` (case insensitive) are permitted.


web/src/App.test.jsx
```diff
@@ -21,26 +21,26 @@
...
+  Outlet: mockComponent("Content"),
+}));
+
 // Mock some components,
 // See https://www.chakshunyu.com/blog/how-to-mock-a-react-component-in-jest/#default-export
-
-jest.mock("@components/questions/Questions", () => () => <div>Questions Mock</div>);
-jest.mock("@components/layout/DBusError", () => () => <div>D-BusError Mock</div>);
+jest.mock("@components/layout/DBusError", () => () => "D-BusError Mock");
 jest.mock("@components/layout/LoadingEnvironment", () => () => "LoadingEnvironment Mock");
```
dgdavid, line 39: I fail to understand why I couldn't use the `mockComponent` function for components living in the @components/layout _scope_. When try to do so, Jest complains with

>  FAIL  src/App.test.jsx
>  &nbsp;&nbsp;● Test suite failed to run
> 
>  &nbsp;&nbsp;&nbsp;TypeError: Cannot read properties of undefined (reading 'mockComponent')
>
>  &nbsp;&nbsp;&nbsp;&nbsp;  > 38 | jest.mock("@components/layout/DBusError", () => mockComponent("D-BusError Mock"));

If those components are moved to, let's say, @components/core and imported and mocked from there, it works.

If, instead, we keep the in @components/layout BUT we remove the `import { Layout } from "@components/layout";` from _src/test-utils.js_ and stop using it there, it works too.

Obviously, we cannot stop wrapping content into `<Layout>` for all tests just because of this. Just trying to find the explanation for that complaint.

@imobachgs, Do you have any hint about what is going on here?

imobachgs: I guess it is related to the fact that Jest executes the `test-utils.js` first, so you are accidentally already loading the `DBusError` (and any other layout-related component). I am not sure about the internals of Jest mocking, but the propose patch looks good.

imobachgs: I guess that, as a rule of thumb, we should avoid loading any application component from the `test-utils.js` (we can make an exception just with the layout).

dgdavid: > so you are accidentally already loading the DBusError

Which makes no sense to me. It's importing only (apparently) the named Layout export... maybe something is happening when re-exporting components from layout/index.js... Don't know

> but the propose patch looks good.

Ok, I'll send it

dgdavid: Just a reminder, `jest.mock` calls are hoisted by [babel-plugin-jest-hoist](https://github.com/facebook/jest/blob/main/packages/babel-plugin-jest-hoist/README.md)

From Jest [manual-mocks](https://jestjs.io/docs/manual-mocks#using-with-es-module-imports) documentation

> [...] For this reason, Jest will automatically hoist jest.mock calls to the top of the module (before any imports)

But it's pretty clear that something was happening in the shadow with that ` import { Layout } from "@components/layout";` :exploding_head:

joseivanlopez (approved):
LGTM

web/src/App.test.jsx
imobachgs: Yes, you are right. There are many moving pieces I am not aware of. But removing [this line](https://github.com/yast/d-installer/blob/e222ec35ebfd40d9e4d4ffb002bb3221dea21fce/web/src/components/layout/index.js#L26) fixes the problem (well, and adjusting the places where you need `DBusError`).

Actually, you can ruin any of the other usages of `mockComponent` in this file by just importing any other of the modules you want to mock.

imobachgs: I guess the root of the issue is that you are [importing something](https://github.com/yast/d-installer/blob/e222ec35ebfd40d9e4d4ffb002bb3221dea21fce/web/src/test-utils.js#L28) that it is already mocked, and that mock calls to a function which is not defined yet.

dgdavid: Yay! You're completely right! I understand now what is going wrong there: when doing the `import { Layout } from "@components/layout";` all the _imports_ in the `src/components/layout/index.js` file are being processed (?). But, because of the Jest hoisting, at this point the `DBusError` component is already _mocked_ and tries to execute the `jest.mock` inline function. Unfortunately, the `mockComponent` function is not available: it has not been imported yet.

You're absolutely right! In my understanding, all imports in src/components/layout/index.js get processed when importing `{ Layout } from "@components/layout"` (??). However, due to the Jest hoisting, the `DBusError` component at this point is already mocked and tries to execute the `jest.mock` inline function. Unfortunately, the `mockComponent` function is not available: it has not been imported yet.

So, something like...

* $ npm run tests src/App.test.jsx
  - Jest found these `jest.mock` calls and put/move them ant the very top
    - Component `DBusError` is now `() => mockComponent("DBusError Mock")`
  - A `test-utils` import is found and processed
  - The `import { Layout } from "@components/layout";` is found and processed
  - The src/components/layout/index.js contains an implicit import to `DBusError`
  - Jest executes the `DBusError` mock.
  - But `mockComponent` has not been defined nor imported yet.

     > TypeError: Cannot read properties of undefined (reading 'mockComponent')
    
<p align="center">
:exploding_head: :exploding_head: :exploding_head: 
</p>

Another solution proposed by @joseivanlopez is to move the `mockComponent` function to another file and import the function BEFORE importing the current test-utils. Fun fact: I made _that_ test yesterday, but forgot about the order and imported it AFTER. :man_facepalming: 

It worked as expected. I'm not sure when to do it: as part of this PR (i.e., create a test-util directory and split current utils into at least two files) or later when we have more utils. In any case, I think we should stick to the default Layout import.

dgdavid: FYI,

Searching more information about the topic, I have found that interesting article https://www.coolcomputerclub.com/posts/jest-hoist-await/ It's almost two years old. Maybe mentioned [top-levels await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await)s are nearest to be a reality in the Jest world nowadays.

dgdavid: > Another solution proposed by @joseivanlopez is to move the mockComponent function to another file and import the function BEFORE importing the current test-utils

Done at https://github.com/yast/d-installer/pull/392/commits/5ef9c478ac4f07e88ec3297cfab1da1617e1b252. We can revert the commit if it looks too much.

dgdavid: Finally, after discussing it _online_ we agreed on reverting 5ef9c47.

> 17:05:03 \<imobach> jilopez, dgdavid: basically, as a user, I find rather tricky that you have to import mocks/ and renderers/ in that specific order
> 
> 17:06:21 \<imobach> jilopez, dgdavid: I consider not loading stuff that we do not need in the test-utils.js as a better alternative

imobachgs (approved):

PR #393, opened 2023-01-12
Check for installed packages in the target system

imobachgs:
Properly asks for installed packages *in the target system* (not in the instsys). We detected this problem while running D-Installer on top of Iguana.


joseivanlopez (approved):
LGTM

PR #397, opened 2023-01-16
Simplify network configuration

imobachgs:
## Problem

The `save_network` client does not work correctly when D-Installer runs on Iguana. 

## Solution

Given that we are relying on *NetworkManager* and supporting just a minimal set of use cases, we decided to simplify the network configuration as much as possible. We will pull the needed bits from `save_network` as they are needed.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/service/lib/dinstaller/network.rb
+++ b/service/lib/dinstaller/network.rb
@@ -24,7 +24,6 @@ require "yast"
 require "yast2/systemd/service"
 require "y2network/proposal_settings"
 
-Yast.import "Lan"
 Yast.import "Installation"
 
 module DInstaller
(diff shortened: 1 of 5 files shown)


teclator:
Code looks good just the probe is a leftover as it is not needed anymore:

https://github.com/yast/d-installer/blob/eb30c4bd80089ef40c99c0c8f1a4a5900014a82d/service/lib/dinstaller/network.rb#L38

service/lib/dinstaller/network.rb
```diff
@@ -31,24 +33,56 @@ def initialize(logger)
...
-    end
-
     # Writes the network configuration to the installed system
+    #
+    # * Copies the connections configuration for NetworkManager, as D-Installer is not
+    #   performing further configuration of the network.
+    # * Enables the NetworkManager service.
     def install
-      Yast::WFM.CallFunction("save_network", [])
+      copy_files
```
teclator, line 42: We could use the new config copier class instead of duplicating the logic, but in case of NM it is bare minimum..

https://github.com/yast/yast-network/pull/1318/files#diff-f65870082b39875af4467bb05dd5ae84681d5c2688c5e213257dc311b64eb4b2

And the new code needs to be reviewed yet.

teclator: `
      Y2Network::ConfigCopier.for(:network_manager)&.copy
`

teclator (approved):
LGTM

PR #398, opened 2023-01-17
Document libvirt + PXE + Iguana setup

imobachgs:
This document explains how to run D-Installer on PXE with the help of Iguana. The described setup uses libvirt, but you can adapt the overall approach to other scenarios (like running your own TFTP server).

### DIFF
--- a/doc/pxe.md
+++ b/doc/pxe.md
@@ -45,7 +45,7 @@ timeout		50
 Do not worry about the kernel, the initrd or the `d-installer.yaml` file, we will jump into it
 later.
 
-### Configure libvirt to server TFTP files
+### Configure libvirt to serve TFTP files
 
 To instruct libvirt to serve the TFTP files, you must add the `tftp` and `bootp` elements to the
 network configuration. Use the `virsh net-edit default` command to edit the configuration and adapt
--- a/doc/pxe.md
+++ b/doc/pxe.md
@@ -1,0 +1,111 @@
+# Using PXE and Iguana
+
+This document explains how to run D-Installer on PXE with the help of Iguana. The described setup
+uses libvirt, but you can adapt the overall approach to other scenarios (like running your own
+TFTP server).
+
+The process can be summarized in these steps:
+
+1. Set up the TFTP tree, defining a boot option for D-Installer + Iguana.
+2. Configure libvirt network to serve the tree.
+3. Prepare the initial ramdisk image (initrd), based on Iguana.
+4. Boot from PXE.
+
+## Set up the TFTP tree
+
+The TFTP files are available in the `tftpboot-installation-openSUSE-Tumbleweed-ARCH` packages. They
+are placed in `/usr/share/tftpboot-installation`, but you should copy them (or at least the `net`
+subdirectory) to customize them. In the example below we are just copying all of them.
+
+    zypper in tftpboot-installation-openSUSE-Tumbleweed-x86_64
+    mkdir /srv/tftpboot
+    cp -a /usr/share/tftpboot-installation/openSUSE-Tumbleweed-x86_64 /srv/tftpboot
+
+To define a boot option to run D-Installer, edit the `net/pxelinux.cfg/default` to include:
+
+```
+label iguana
+  ipappend 2
+  kernel boot/x86_64/loader/iguana-new/vmlinuz-iguana
+  append initrd=boot/x86_64/loader/initrd-iguana rd.iguana.control_url=tftp://192.168.122.1/net/d-installer.yaml rd.iguana.debug=1 rd.neednet=1
+```
+
+Do not worry about the kernel, the initrd or the `d-installer.yaml` file, we will jump into it
+later.
+
+## Configure libvirt to server TFTP files
+
+To instruct libvirt to serve the TFTP files, you must add the `tftp` and `bootp` elements to the
+network configuration. Use the `virsh net-edit default` command to edit the configuration and adapt
+it accordingly. Here is an example:
+
+```xml
+<network connections='1'>
+  <name>default</name>
+  <uuid>639e02a7-fcbd-4cf3-a563-6db083aef051</uuid>
+  <forward mode='nat'>
+    <nat>
+      <port start='1024' end='65535'/>
+    </nat>
+  </forward>
+  <bridge name='virbr0' stp='on' delay='0'/>
+  <mac address='52:54:00:fb:7c:8e'/>
+  <ip address='192.168.122.1' netmask='255.255.255.0'>
+    <tftp root='/srv/tftpboot/openSUSE-Tumbleweed-x86_64'/>
+    <dhcp>
+      <range start='192.168.122.2' end='192.168.122.254'/>
+      <bootp file='net/pxelinux.0'/>
+    </dhcp>
+  </ip>
+</network>
+```
+
+## initrd preparation
+
+**Please, build the initrd on a virtual machine to avoid messing up your system.**
+
+Iguana provides a universal initrd in which actual functionality is implemented in containers. Which
+containers to use and how to set them up is defined in a *workflow definition*. The [Iguana
+repository](https://github.com/openSUSE/iguana) includes a [definition for
+D-Installer](https://github.com/openSUSE/iguana/blob/main/iguana-workflow/examples/d-installer.yaml).
+
+In the future, you should be able to get the kernel and the initrd files from the [Iguana
+package](https://build.opensuse.org/package/show/home:oholecek:iguana/iguana). However, for
+D-Installer to work, we need to do include an `/etc/NetworkManager` directory, so you need to
+rebuild the image by now.
+
+
+1. Install [dracut-iguana](https://github.com/openSUSE/iguana/tree/main/dracut-iguana) from
+[OBS](https://build.opensuse.org/package/show/home:oholecek:iguana/dracut-iguana).
+
+2. (Optional) For debugging purposes, you might want to add SSH support so you can connect to the
+   system. If that's the case, please install the `dracut-ssh` package and place your public SSH key
+   on `/root/.ssh/authorized_keys`.
+
+3. Create a `/etc/NetworkManager` directory to be included in the initrd:
+
+        mkdir -p rd.live.networkmanager/etc/NetworkManager
+
+4. Rebuild the image:
+
+       dracut --verbose --force --no-hostonly --no-hostonly-cmdline --no-hostonly-default-device --no-hostonly-i18n --reproducible --include rd.live.networkmanager / iguana-initrd
+
+5. Copy the kernel (`/boot/vmlinuz-VERSION`), the initrd and the workflow definition to the TFTP
+   tree. You must use the same paths specified in the `ìguana` boot option (see [Set up the TFTP
+   tree](#set-up-the-tftp-tree) section).
+
+## Booting from PXE
+
+To boot from PXE, you need to enable the boot menu for your VM. You can do it easily by using
+`virt-manager` and marking the `Enable boot menu` option in the `Boot options` section of your
+virtual machine. Alternatively, you can edit the XML definition (`virsh edit NAME`) and add the `bootmenu`
+element to `<os>` section:
+
+```xml
+ <os>
+   <bootmenu enable='yes'/>
+ </os>
+```
+
+After starting the machine, you should choose the option to boot from PXE and introduce `iguana` in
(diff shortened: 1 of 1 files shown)


doc/pxe.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Set up the TFTP tree
+
+The TFTP files are available in the `tftpboot-installation-openSUSE-Tumbleweed-ARCH` packages. They
+are placed in `/usr/share/tftpboot-installation`, but you should copy them (or at least the `net`
+subdirectory) to customize them. In the example below we are just copying all of them.
+
+    zypper in tftpboot-installation-openSUSE-Tumbleweed-x86_64
+    mkdir /srv/tftpboot
+    cp -a /usr/share/tftpboot-installation/openSUSE-Tumbleweed-x86_64 /srv/tftpboot
```
aaannz, line 22: This works, but note that this is not needed.
Minimal setup is:
- installing `syslinux`
- copy `/usr/share/syslinux/pxelinux.0` from installed package to `/srv/tftpboot`
- create `/srv/tftpboot/pxelinux.cfg` directory
- create `/srv/tftpboot/pxelinux.cfg/default` as described below

syslinux is 2MB, tftpboot-installation is 500MB+

of course using syslinux will change paths in configs below, like not having `/srv/tftpboot/net` and `/srv/tftpboot/net/boot` and all files are related just to `/srv/tftpboot`

doc/pxe.md
```diff
@@ -0,0 +1,111 @@
...
+1. Install [dracut-iguana](https://github.com/openSUSE/iguana/tree/main/dracut-iguana) from
+[OBS](https://build.opensuse.org/package/show/home:oholecek:iguana/dracut-iguana).
+
+2. (Optional) For debugging purposes, you might want to add SSH support so you can connect to the
+   system. If that's the case, please install the `dracut-ssh` package and place your public SSH key
+   on `/root/.ssh/authorized_keys`.
+
+3. Create a `/etc/NetworkManager` directory to be included in the initrd:
+
+        mkdir -p rd.live.networkmanager/etc/NetworkManager
```
aaannz, line 87: Why is this needed?

aaannz (commented):
This is for custom modification of iguana, right?

Because for testing d-installer (without doing changes in iguana) it is enough to install `iguana` package which contains both kernel and initrd.

Another option for tests is to use direct kernel boot instead of pxe booting. In virt-manager you can in boot loader section you can directly provide paths to kernel, initrd and set kernel options.

doc/pxe.md
imobachgs: Because we need to bind-mount the `/etc/NetworkManager` directory in the backend container to copy the configuration files at the end of the installation.

doc/pxe.md
imobachgs: > This is for custom modification of iguana, right?
> 
> Because for testing d-installer (without doing changes in iguana) it is enough to install `iguana` package which contains both kernel and initrd.
> 

We need to modify the initrd by now (to add the `/etc/NetworkManager` directory).

> Another option for tests is to use direct kernel boot instead of pxe booting. In virt-manager you can in boot loader section you can directly provide paths to kernel, initrd and set kernel options.

Yes, that's right. I personally have been using that approach until now, but it was a requirement to document a PXE setup for our QA team.

imobachgs:
@aaannz Thanks a lot for your comments! I will update the document.

doc/pxe.md
aaannz: Ok, this leaves more to explain.
This will create directory in the initrd fs. Once initrd switches root or reboots, then this dir is lost.

Who is the supposed consumer of this configuration? Initrd or installed system?

imobachgs: NetworkManager runs in the host system, although you can change the configuration using its D-Bus interface from D-Installer web UI. The changes are written by NetworkManager to that directory.

At the end of the installation, `dinstaller-backend` needs to copy those files to the installed system. If you reboot into the installed system, those files should be there.

Thinking about it twice, I guess we could define a "step" to create the directory in the workflow itself. Is it possible to run an arbitrary command in the host (without using a container)? I could use a container to create the directory, but it is probably too much :)

imobachgs: I guess support for [steps](https://github.com/openSUSE/iguana/blob/ba332ca6aa38d607e816e26b9b4a526afbeb21b3/iguana-workflow/examples/control.yaml#L23-L24) is not implemented yet. I am just guessing because the [steps field](https://github.com/openSUSE/iguana/blob/1fbfbe7ebf5d1993e37945bccd0ba1b7f5c46ca3/iguana-workflow/src/workflow.rs#L34) is not referenced.

aaannz: > NetworkManager runs in the host system, although you can change the configuration using its D-Bus interface from D-Installer web UI. The changes are written by NetworkManager to that directory.
> 
> At the end of the installation, `dinstaller-backend` needs to copy those files to the installed system. If you reboot into the installed system, those files should be there.

However iguana is running in initrd. Anything you create at initrd stage in root (like /etc/NM) will not be visible in the booted system and is also not preserved for next boot. Usually system to boot is mounted under `/sysroot` so you would want `/sysroot/etc/NetworkManager`. Iguana mounts it after all containers run based on `/iguana/mountlist` file. I still think this is not doing what you expect it.

Maybe we are talking about different things here so lets go to basics. D-Installer partition disk, format and install root fs. So D-installer container have next root mounted at some point. And you want to copy generated network manager config to this mounted root. Right?

If you just need to have shared storage between backend and web container then you can already used any volume (like `dbus` volume is already shared)

> 
> Thinking about it twice, I guess we could define a "step" to create the directory in the workflow itself. Is it possible to run an arbitrary command in the host (without using a container)? I could use a container to create the directory, but it is probably too much :)

Yes, steps are not implemented (yet, I am not sure if we need/want them). Using different container will not help as container do not have access to initrd fs.

imobachgs: > Maybe we are talking about different things here so lets go to basics. D-Installer partition disk, format and install root fs. So D-installer container have next root mounted at some point. And you want to copy generated network manager config to this mounted root. Right?

Yes, I tend to think we are talking about different things 😅. I will try to summarize our POV (that might be wrong):

* NetworkManager runs on initrd (`nm-initrd.service`).
* The user can alter the configuration through the D-Installer UI (`dinstaller-web` service), which talks to NetworkManager via D-Bus. That's why we added `/var/run/dbus/system_bus_socket` to the list of volumes.
* Before finishing the installation, `dinstaller-backend` needs to copy the NetworkManager configuration files (from the initrd), to the installed system. So it needs to access the `/etc/NetworkManager` directory in initrd.

Mounting `/etc/NetworkManager` as a volume would make the trick. The problem is that the /etc/NetworkManager directory does not exist from the very beginning (only after you make a change in the configuration). So we cannot bind-mount the directory in the `dinstaller-backend`.

Does it explain the problem better?

> 
> If you just need to have shared storage between backend and web container then you can already used any volume (like `dbus` volume is already shared)

Not between the backend and the web container, but between the initrd and the backend.

> Yes, steps are not implemented (yet, I am not sure if we need/want them). Using different container will not help as container do not have access to initrd fs.

Unless you make some directory accessible as a volume, right?

aaannz: > * Before finishing the installation, `dinstaller-backend` needs to copy the NetworkManager configuration files (from the initrd), to the installed system. So it needs to access the `/etc/NetworkManager` directory in initrd.

But this will not copy it to the installed system. Unless NetworkManager does some tricks like copying initrd `/etc/NetworkManager` to installed system `/sysroot/etc/NetworkManager` (which is only available after all containers are finished). I checked NM dracut module and I did not find anything doing such thing.

> Not between the backend and the web container, but between the initrd and the backend.

So I guess my question is what then copies config from initrd to installed system.

imobachgs: > > * Before finishing the installation, `dinstaller-backend` needs to copy the NetworkManager configuration files (from the initrd), to the installed system. So it needs to access the `/etc/NetworkManager` directory in initrd.
> 
> But this will not copy it to the installed system. Unless NetworkManager does some tricks like copying initrd `/etc/NetworkManager` to installed system `/sysroot/etc/NetworkManager` (which is only available after all containers are finished). I checked NM dracut module and I did not find anything doing such thing.

Perhaps the "installed system" means different things: I am referring to the system that `dinstaller-backend` is installing, which is mounted under `/mnt` in the `dinstaller-backend` container.

> 
> > Not between the backend and the web container, but between the initrd and the backend.
> 
> So I guess my question is what then copies config from initrd to installed system.

The `dinstaller-backend` does it (getting the files from the bind-mounted `/etc/NetworkManager` directory).

I guess moving this discussion offline might help to clarify the situation.

imobachgs: After discussing this issue, we decided to implement https://github.com/openSUSE/iguana/issues/18 so rebuilding the initrd is not needed anymore.

doc/pxe.md
```diff
@@ -0,0 +1,132 @@
...
+
+Now your virtual machine should be ready to boot from PXE and start Iguana/D-Installer. Once the
+system boots and the services are started, you should be able to access D-Installer with a browser
+on port 9090.
+
+## Tips
+
+### Adding support for SSH
+
+**Please, build the initrd on a virtual machine to avoid messing up your system.**
```
dgdavid, line 111: NP: Perhaps adding more emphasis here worth it. Just adding an emoji can help, something like

```suggestion
:warning: **Please, build the initrd on a virtual machine to avoid messing up your system.**
```

doc/pxe.md
```diff
@@ -0,0 +1,132 @@
...
+After installing the `iguana` package, copy the kernel (`/usr/share/iguana/vmlinuz-VERSION`), the
+initrd (`/usr/share/iguana/iguana-initrd`) and the workflow definition to the TFTP tree. You must
+use the same paths specified in the `ìguana` boot option (see [Set up the TFTP
+tree](#set-up-the-tftp-tree) section).
+
+### Booting from PXE
+
+To boot from PXE, you need to enable the boot menu for your VM. You can do it easily by using
+`virt-manager` and marking the `Enable boot menu` option in the `Boot options` section of your
+virtual machine. Alternatively, you can edit the XML definition (`virsh edit NAME`) and add the `bootmenu`
```
jreidinger, line 94: Is it really needed? I use prague PXE without boot menu, just add to first place network card as boot option.

doc/pxe.md
```diff
@@ -0,0 +1,132 @@
...
+display		message
+implicit	1
+prompt		1
+timeout		50
+```
+
+Do not worry about the kernel, the initrd or the `d-installer.yaml` file, we will jump into it
+later.
+
+### Configure libvirt to server TFTP files
```
aaannz, line 48:
```suggestion
### Configure libvirt to serve TFTP files
```

doc/pxe.md
```diff
@@ -0,0 +1,132 @@
...
+
+Now your virtual machine should be ready to boot from PXE and start Iguana/D-Installer. Once the
+system boots and the services are started, you should be able to access D-Installer with a browser
+on port 9090.
+
+## Tips
+
+### Adding support for SSH
+
+:warning: **Please, build the initrd on a virtual machine to avoid messing up your system.**
```
jreidinger, line 111: btw when we promo container. Is it possible to do it in container as it should be easier and faster then VM.

doc/pxe.md
```diff
@@ -0,0 +1,132 @@
...
+
+### initrd preparation
+
+Iguana provides a universal initrd in which actual functionality is implemented in containers. This
+ramdisk and its corresponding kernel are included in the [`iguana`
+package](https://build.opensuse.org/package/show/home:oholecek:iguana/iguana).
+
+Which containers to use and how to set them up is defined in a *workflow definition*. The [Iguana
+repository](https://github.com/openSUSE/iguana) includes a [definition for
+D-Installer](https://github.com/openSUSE/iguana/blob/main/iguana-workflow/examples/d-installer.yaml).
```
aaannz, line 83: Just hint, you don't have to download definition file, using raw github link as control url works too. e.g.: `rd.iguana.control_url=https://raw.githubusercontent.com/openSUSE/iguana/main/iguana-workflow/examples/d-installer.yaml`

aaannz (approved):
LGTM, just two nitpicks to consider from my side.

doc/pxe.md
imobachgs: No, you are right, it is not needed.

jreidinger (approved):

dgdavid (approved):

PR #399, opened 2023-01-17
Adding first POMs for testing with Playwright

jknphy:
Adding first [POMs](https://playwright.dev/docs/pom) for testing with Playwright.
I tried to use the [recommended locators](https://playwright.dev/docs/api/class-framelocator#frame-locator-get-by-role) while we still don't have customized locators in the application (i.e.: data-testid).


.github/workflows/ci.yml
```diff
@@ -236,8 +237,8 @@ jobs:
 
     - name: Install packages
       # TODO: check which fonts are really needed
-      run:  zypper --non-interactive install --no-recommends gcc gcc-c++ make openssl-devel ruby-devel
-            npm augeas-devel cockpit dbus-1 xorg-x11-server libgtk-3-0
+      run:  zypper --non-interactive install --no-recommends gcc-c++ ruby-devel
+            'npm>=18' augeas-devel cockpit xorg-x11-server libgtk-3-0
```
lslezak, line 241: Ideally we should somehow automate this or share the list of packages. For the regular YaST packages you can run `rake build_dependencies:list` to get the build dependencies.

Unfortunately that does not work for the D-Installer correctly because it has multiple `*.spec` files in different subdirectories. I'll check if could improve that...

jknphy:
hi @lslezak I tried to increase the test timeout and the navigation timeout but no luck in the CI.
In my local setup works without issues, but my setup is remote, so I have playwright in my machine and the application installed in a VM. For the traces I can see that the application is not ready, so somehow we need to wait more before starting to interact. Let me know if you have any suggestion.

jknphy:
> hi @lslezak I tried to increase the test timeout, the navigation timeout and even the expect tiemout but no luck in the CI. In my local setup works without issues, but my setup is remote, so I have playwright in my machine and the application installed in a VM. For the traces I can see that the application is not ready, so somehow we need to wait more before starting to interact. Let me know if you have any suggestion.

seems that those links for language are tricky to find a common locator depending on the current value that they have. Things will be easier with test-id locators in future. I added another todo. The problem is that when the app is slower there is an initial value which is not English (US) but some other initialization value and a generic locator was not good enough not having  a reference to the current value. But all good now, I just duplicated the links in the POM for now. CI passing :)

PR #400, opened 2023-01-19
[web] Update aliases for using ~/ instead of @ 

dgdavid:
Because "@" is used by npm for packages published under a [named scope](https://docs.npmjs.com/about-scopes). From their documentation:

  > The scope name is everything between the @ and the slash

Additionally, "~" use to kind of means "home" or "root", which suits well for an alias pointing to the "root" directory.

Consequently, it is easier to understand now whether an import comes from a dependency or the source code of the project.

---

Related to https://github.com/yast/d-installer/pull/325 and https://github.com/yast/d-installer/pull/325/commits/f33e3273833bd7b3dae17d4861126befb0dd5c35

### DIFF
--- a/web/tsconfig.json
+++ b/web/tsconfig.json
@@ -9,14 +9,14 @@
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
     "paths": {
-      "@/*": ["src/*"],
-      "@assets/*": ["src/assets/*"],
-      "@client": ["src/client/index.js"],
-      "@client/*": ["src/client/*"],
-      "@components/*": ["src/components/*"],
-      "@context/*": ["src/context/*"],
-      "@lib/*": ["src/lib/*"],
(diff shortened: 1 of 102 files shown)


web/tsconfig.json
```diff
@@ -9,14 +9,9 @@
...
-      "@assets/*": ["src/assets/*"],
-      "@client": ["src/client/index.js"],
-      "@client/*": ["src/client/*"],
-      "@components/*": ["src/components/*"],
-      "@context/*": ["src/context/*"],
-      "@lib/*": ["src/lib/*"],
-      "~icons/*": ["node_modules/@material-symbols/svg-400/outlined/*"]
+      "~/*": ["src/*"],
+      "~/client": ["src/client/index.js"],
+      "@icons/*": ["node_modules/@material-symbols/svg-400/outlined/*"]
```
dgdavid, line 14: Use the @ here because this alias actually is a shortcut for an import from a node package.

imobachgs (approved):

PR #402, opened 2023-01-20
D-Bus API for iSCSI

joseivanlopez:
## Problem

D-Installer does not provide a way of connecting to iSCSI targets yet.

https://trello.com/c/iQ7BCexU/3261-5-d-installer-d-bus-interface-configure-the-iscsi-initiator-and-the-iscsi-targets

## Solution

Extend current D-Bus API of the storage service in order to support iSCSI.

~~~
/DInstaller/Storage1
  .ObjectManager
  .DInstaller.ServiceStatus1
  .DInstaller.Progress1
  .DInstaller.Validation1
  .DInstaller.Storage1
  .DInstaller.Storage1.Proposal.Calculator
  .DInstaller.Storage1.ISCSI.Initiator
/DInstaller/Storage1/Proposal
  .DInstaller.Storage1.Proposal
/DInstaller/Storage1/iscsi_nodes/[0-9]+
  .DInstaller.Storage1.ISCSI.Node
~~~

Now the main object `/DInstaller/Storage1` implements a new interface `.DInstaller.Storage1.ISCSI.Initiator
`. That interface offers methods for configuring the iSCSI initiator, performing discovey, and deleting node.

Discovered nodes are dynamically exported in the path `/DInstaller/Storage1/iscsi_nodes/[0-9]+`, and such nodes implements the interface `.DInstaller.Storage1.ISCSI.Node`. A node allows to login and logout.

Follow-up:

* Configure offload card [create issue].
* Validate proposal after changing iSCSI nodes [create issue].

## Examples

Configure iSCSI initiator name:

~~~
$ busctl --address unix:path=/run/d-installer/bus set-property org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage1 org.opensuse.DInstaller.Storage1.ISCSI.Initiator InitiatorName s "iqn.1996-04.de.suse:01:351e6d6249"
~~~

Perform discovery

~~~
$ busctl --address unix:path=/run/d-installer/bus call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage1 org.opensuse.DInstaller.Storage1.ISCSI.Initiator Discover sua{sv} 192.168.100.215 3260 4 Username s testi Password s testi ReverseUsername s testt ReversePassword s testt
~~~

Login

~~~
$ busctl --address unix:path=/run/d-installer/bus call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage1/iscsi_nodes/2 org.opensuse.DInstaller.Storage1.ISCSI.Node Login a{sv} 5 Username s testi Password s testi ReverseUsername s testt ReversePassword s testt Startup s onboot
~~~

Logut

~~~
$ busctl --address unix:path=/run/d-installer/bus call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage1/iscsi_nodes/1 org.opensuse.DInstaller.Storage1.ISCSI.Node Logout
~~~

Delete

~~~
$ busctl --address unix:path=/run/d-installer/bus call org.opensuse.DInstaller.Storage /org/opensuse/DInstaller/Storage1 org.opensuse.DInstaller.Storage1.ISCSI.Initiator Delete o /org/opensuse/DInstaller/Storage1/iscsi_nodes/1
~~~

## Testing

- Added unit tests
- Tested manually

## Screenshots

Screenshot from 2023-02-10 16-56-59


mvidner:
Not directly applicable right now, but eventually I hope:
"Document D-Bus API inside introspection XML" https://github.com/yast/d-installer/pull/415

mvidner:
In the review somebody mentioned that service/lib/dinstaller/dbus/storage/manager.rb was getting quite big, with all the interfaces it needs to have.

One way to address it would be to put each of the interfaces in its own Module... and we're already doing it with `DInstaller::DBus::Interfaces::Progress`, `ServiceStatus` an `Validation`.

Perhaps doing the same for all the remaining interfaces would solve the problem? Or split the manager object up according to the responsibilities?

If ruby-dbus can get an improvement to help with it, I'd happily add it, but so far I don't see one.

joseivanlopez:
> In the review somebody mentioned that service/lib/dinstaller/dbus/storage/manager.rb was getting quite big, with all the interfaces it needs to have.
> 
> One way to address it would be to put each of the interfaces in its own Module... and we're already doing it with `DInstaller::DBus::Interfaces::Progress`, `ServiceStatus` an `Validation`.
> 
> Perhaps doing the same for all the remaining interfaces would solve the problem? Or split the manager object up according to the responsibilities?
> 
> If ruby-dbus can get an improvement to help with it, I'd happily add it, but so far I don't see one.

The problem I see is that the interface properties and methods are defined by the dbus object. If a dbus object implements quite some ifaces, then it has to define all the methods for each intereface, which implies to have a dbus object with a lot of possibly unrelated logic. Moreover, we need to pay special attention to property/method name collition, specially when importing interfaces from a module (e.g., `Progress` iface).

I would like to have some kind of composition or similar approach for aisolating the interface logic and definition. An idea could be to do something similar to what a service does with exporting D-Bus objects. For example, interfces are defined in its own class, and then the D-Bus object add them:

~~~
class MyObject < DBus::Object
  def initialize
    foo_iface = DBus::Interface.new
    bar_iface = DBus::Interface.new

    add_interfaces foo_iface, bar_iface
  end
end
~~~

We could workaround it by implementing some kind of `Interface` and `Object` wrappers around `DBus::Object` in D-Installer, but maybe it would be better to do it directly in *ruby-dbus*.

mvidner:
> If a dbus object implements quite some ifaces, then it has to define all the methods for each intereface, which implies to have a dbus object with a lot of possibly unrelated logic.

But that problem is also true if we leave out D-Bus out of the picture, and then it looks to me like an ordinary OOP design problem: mixing too many responsibilities in a single class.

I would solve that by using different classes/objects. Here, the ISCSI.Initiator interface is IMHO a natural case for one.

Even if the objects lifetimes are tightly coupled (so you could take that a reason to combine them in a single class) I'd split them up for clarity and ease of maintenance.

service/lib/dinstaller/dbus/storage/manager.rb
```diff
@@ -131,6 +123,75 @@ def calculate_proposal(dbus_settings)
...
+        def iscsi_discover(address, port, options)
+          success = backend.iscsi.discover_send_targets(address, port, iscsi_auth(options))
+          success ? 0 : 1
+        end
+
+        # Deletes an iSCSI node from the database
+        #
+        # @param path [::DBus::ObjectPath]
+        # @return [Integer] 0 on success, 1 on failure
+        def iscsi_delete(path)
```
ancorgs, line 176: This returns 1 both for a not found node and for a failure in deleting. And at first sight I don't see any trace in the log making any difference. I said "at first sight". Maybe the log traces left by `iscsi_nodes_tree` and `backend.iscsi.delete` are enough to spot the difference, but...

doc/dbus_api.md
```diff
@@ -111,66 +111,327 @@ Iface: o.o.YaST.Installer1.Software
...
+
+Arguments:
+
+* `in s address`: IP address of the iSCSI server.
+* `in u port`: Port of the iSCSI server.
+* `in a{sv} options`:
+  * `InitiatorUsername s`: Username for incoming authentication.
+  * `InitiatorPassword s`: Password for incoming authentication.
+  * `TargetUsername s`: Username for outgoing authentication.
+  * `TargetPassword s`: Password for outgoing authentication.
```
ancorgs, line 364: I fear the terms "incoming" and "outgoing" regarding iSCSI authentication. They are not so well established and I find they can easily lead to confusion.

I guess with "incoming authentication" you mean the Target authenticates the initiator and with "outgoing" you mean the initiator authenticates the target.

But then in the code below you do this:

```ruby
Y2IscsiClient::Authentication.new.tap do |auth|
  auth.username = target_username
  auth.password = target_password
  auth.username_in = initiator_username
  auth.password_in = initiator_password
end
```

Which is the opposite. `Y2IscsiClient::Authentication#username_in` is documented as "_Username for the CHAP challenge during authentication by the initiator_". That would be "outgoing authentication" in your jargon, so you should have assigned `TargetUsername` to `username_in`.

service/lib/dinstaller/storage/iscsi/initiator.rb
```diff
@@ -0,0 +1,72 @@
...
+        end
+
+        # Sets the iSCSI offload card
+        #
+        # @param value [String]
+        def offload_card=(value)
+          Yast::IscsiClientLib.SetOffloadCard(value)
+        end
+
+        # Wheter iBTF is used
```
ancorgs, line 63: Whet**h**er (miissing "h")

I would be a more specific: "_Whether the initiator name was set via iBFT_"

service/lib/dinstaller/storage/iscsi/manager.rb
```diff
@@ -0,0 +1,220 @@
...
+        # @param node [Node]
+        # @param authentication [Y2IscsiClient::Authentication]
+        # @param startup [String, nil] Startup status
+        #
+        # @return [Boolean] Whether the action successes
+        def login(node, authentication, startup: nil)
+          probe_after do
+            startup ||= Yast::IscsiClientLib.default_startup_status
+
+            return false unless ["onboot", "manual", "automatic"].include?(startup)
```
ancorgs, line 120: I guess we should log the reason here.

ancorgs (commented):
Just releasing some comments (I did hit the wrong button when adding them).

service/lib/dinstaller/dbus/storage/iscsi_node.rb
```diff
@@ -0,0 +1,148 @@
...
+          dbus_properties_changed(ISCSI_NODE_INTERFACE, properties, [])
+        end
+
+        # Creates an iSCSI session
+        #
+        # @param options [Hash<String, String>] Options from a D-Bus call:
+        #   @option Username [String] Username for authentication by target
+        #   @option Password [String] Password for authentication by target
+        #   @option ReverseUsername [String] Username for authentication by initiator
+        #   @option ReversePassword [String] Username for authentication by inititator
```
ancorgs, line 126: ~~Username~~ Password

service/lib/dinstaller/dbus/storage/manager.rb
```diff
@@ -131,6 +123,77 @@ def calculate_proposal(dbus_settings)
...
+          dbus_node = iscsi_nodes_tree.find(path)
+          logger.info("iSCSI node #{path} is not exported")
```
ancorgs, line 178: This "not exported" is always printed to the log even if it's not true. Isn't it?

service/lib/dinstaller/dbus/storage/manager.rb
```diff
@@ -131,6 +123,77 @@ def calculate_proposal(dbus_settings)
...
+          return 1 unless dbus_node
+
+          success = backend.iscsi.delete(dbus_node.iscsi_node)
+          logger.info("Fail to delete iSCSI node #{path}")
```
ancorgs, line 182: Same than above. It always reports failure in the logs even if it's not true.

Am I overlooking something?

service/lib/dinstaller/dbus/storage/manager.rb
```diff
@@ -131,6 +123,77 @@ def calculate_proposal(dbus_settings)
...
+        # @return [Integer] 0 on success, 1 on failure
+        def iscsi_discover(address, port, options = {})
+          success = backend.iscsi.discover_send_targets(address, port, iscsi_auth(options))
+          success ? 0 : 1
+        end
+
+        # Deletes an iSCSI node from the database
+        #
+        # @param path [::DBus::ObjectPath]
+        # @return [Integer] 0 on success, 1 on failure
```
ancorgs, line 175: Ignorant question: is that the typical behavior on D-Bus APIs? I mean, doesn't it make sense to return different exit codes for the different problems instead of "1" in all cases?

service/lib/dinstaller/storage/iscsi/manager.rb
```diff
@@ -0,0 +1,225 @@
...
+        def initialize(logger: nil)
+          @logger = logger || ::Logger.new($stdout)
+          @initiator = ISCSI::Initiator.new
+
+          @on_probe_callbacks = []
+        end
+
+        # Performs actions for activating iSCSI
+        def activate
+          return if @activated
```
ancorgs, line 55: I'm fine with this. But actually I think running activation twice would be safe.

All current operations are safe.

The original goal of having an `@activated` flag (well, it was called `@probed` back then) was to ensure other operations were never performed before a probing/activation. Not preventing the activation/probing to be run twice.

For example, you should never try to discover is you have not activated before (because maybe there is not a valid initiator name).

ancorgs (commented):
Generally it looks pretty nice. Great work. But I added a couple of comments.

service/lib/dinstaller/dbus/storage/manager.rb
joseivanlopez: Ups.

service/lib/dinstaller/dbus/storage/manager.rb
joseivanlopez: My fault. Thanks.

service/lib/dinstaller/dbus/storage/manager.rb
joseivanlopez: Yes, it makes sense to return different error code. Fixing it.

ancorgs (approved):
LGTM. Great work!

PR #404, opened 2023-01-20
yupdate - allow building the web frontend in the development mode

lslezak:
## Problem

The `yupdate` script builds the web frontend in the production mode. That has some disadvantages:
- The debugging files (`*.map`) are missing, you cannot get a reference to the original sources, this makes debugging more difficult
- The build takes longer time

## Solution

- Allow reading the `NODE_ENV` value from the current environment

## Testing

- Tested manually with `NODE_ENV=development yupdate ...`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -115,7 +115,18 @@ if File.exist?("/.packages.initrd") || `mount`.match?(/^[\w]+ on \/ type overlay
 
     puts "Installing the Web frontend..."
     Dir.chdir("web") do
-      sh "NODE_ENV=production make install"
+      node_env = ENV["NODE_ENV"] || "production"
+      sh "NODE_ENV=#{node_env.shellescape} make install"
+
+      # clean up the extra files when switching the development/production mode
+      if node_env == "production"
+        # remove the uncompressed and development files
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -115,7 +115,18 @@ if File.exist?("/.packages.initrd") || `mount`.match?(/^[\w]+ on \/ type overlay
...
+
+      # clean up the extra files when switching the development/production mode
+      if node_env == "production"
+        # remove the uncompressed and development files
+        FileUtils.rm_f(Dir.glob("/usr/share/cockpit/d-installer/index.{css,html,js}"))
+        FileUtils.rm_f(Dir.glob("/usr/share/cockpit/d-installer/*.map"))
+      else
+        # remove the compressed files
+        FileUtils.rm_f(Dir.glob("/usr/share/cockpit/d-installer/*.gz"))
+      end
```
dgdavid, line 129: NP: Why not just removing the full content before running the `make install`?

Anyway, it looks good, thanks.

dgdavid (approved):
Looks good, just an idea commented there. Thank you!

Rakefile
lslezak: If `make install` failed you would have empty directory, i.e. no d-installer at all, this is safer.

PR #407, opened 2023-01-24
Added a button for displaying the YaST logs

lslezak:
## Problem

- It should be possible to easily see the YaST logs from the UI to make debugging easier
- Related to gh#yast/d-installer#379

## Solution

- Added a new button which opens a popup with the `y2log` file
- So far a simple solution, no message colorizing or filtering. We can improve it later.

## Testing

- Tested manually

## TODO

- [ ] Unit tests

## Screenshots

| Options Menu  | Log Popup |
| ------------- | ------------- |
| log_view_button  | log_view_popup  |

### DIFF
--- a/web/src/components/core/LogPopup.jsx
+++ b/web/src/components/core/LogPopup.jsx
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React, { useState } from "react";
+import { Popup } from "~/components/core";
+import { classNames } from "~/utils";
+
+import "./logpopup.scss";
+
+export default function LogPopup({ className, log, onCloseCallback, title }) {
+  const [isOpen, setIsOpen] = useState(true);
+
+  const close = () => {
+    setIsOpen(false);
+    if (onCloseCallback) onCloseCallback();
+  };
+
+  const lines = log.split("\n").map((line, index) => {
+    return <div className="d-installer-logline" key={`log-line-${index}`}>{line}</div>;
+  });
+
+  return (
+    <>
+      <Popup
+        isOpen={isOpen}
+        title={title}
+        variant="large"
+      >
+        <div className={classNames("d-installer-logpopup", className)}>
+          {lines}
+        </div>
+
+        <Popup.Actions>
+          <Popup.Confirm onClick={close} autoFocus>Close</Popup.Confirm>
+        </Popup.Actions>
+      </Popup>
+    </>
+  );
+}
--- a/web/src/components/core/logpopup.scss
+++ b/web/src/components/core/logpopup.scss
@@ -1,0 +1,12 @@
+div.d-installer-logpopup {
+  word-break: break-all;
+  white-space: pre-wrap;
+  font-family: var(--pf-global--FontFamily--monospace);
+  font-size: 80%;
+}
+
+div.d-installer-logline {
+  &:hover {
+    background-color: var(--color-gray-dark);
+  }
+}
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jan 24 09:32:13 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Added a button for displaying the YaST logs
+  (related to gh#yast/d-installer#379)
+
+-------------------------------------------------------------------
 Fri Jan 20 09:03:22 UTC 2023 - David Diaz <dgonzalez@suse.com>
 
 - UI fixes (gh#yast/d-installer#401):
--- a/web/src/components/core/ShowLogButton.jsx
+++ b/web/src/components/core/ShowLogButton.jsx
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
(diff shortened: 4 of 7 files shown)


web/src/components/core/logpopup.scss
```diff
@@ -0,0 +1,12 @@
...
+  white-space: pre-wrap;
+  font-family: var(--pf-global--FontFamily--monospace);
+  font-size: 80%;
+}
+
+div.d-installer-logline {
+  &:hover {
+    background-color: var(--color-gray-dark);
+  }
+}
```
dgdavid, line 12: As said in another comment, this could be moved to either, utilities or block CSS files. Let's start with a block by now, something like

```diff
+
+
+.filecontent {
+  font-family: var(--ff-code);
+  font-size: 80%;
+  word-break: break-all;
+  white-space: pre-wrap;
+
+  div:hover {
+    background-color: var(--color-gray-dark);
+  }
+}
```

With a better name for the class, but avoiding the d-installer prefix.

dgdavid (commented):
Thanks @lslezak!

The approach looks good and simple as a first step for having an easy way to check the logs. However, I have left a few suggestions, please check them before going for the unit testing.

Thanks!

web/src/components/core/LogPopup.jsx
```diff
@@ -0,0 +1,57 @@
...
+export default function LogPopup({ className, log, onCloseCallback, title }) {
+  const [isOpen, setIsOpen] = useState(true);
+
+  const close = () => {
+    setIsOpen(false);
+    if (onCloseCallback) onCloseCallback();
+  };
+
+  const lines = log.split("\n").map((line, index) => {
+    return <div className="d-installer-logline" key={`log-line-${index}`}>{line}</div>;
```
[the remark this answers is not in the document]
dgdavid: Maybe not, don't waste too much time with that comment.

web/src/components/core/FileViewer.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+import React, { useState, useEffect } from "react";
+import { Popup } from "~/components/core";
+import { Alert, Spinner } from "@patternfly/react-core";
+import cockpit from "../../lib/cockpit";
+
+import "./fileviewer.scss";
+
+// file loading indicator
+const spinner = () => {
```
dgdavid, line 30: NP: we already have a `Loading` component at https://github.com/yast/d-installer/blob/master/web/src/components/layout/LoadingEnvironment.jsx

Of course, it needs some love starting by the name (we should use something more generic than `LoadingEnvironment`). It could be moved to _components/core_ too.

So, you could use `<LoadingEnvironment title="Reading ${file} file" />`  (although admittedly would be better `<Loader text=...>`)

Anyway, I'm not against for changing the loader icon, but I think we could use only one.

web/src/components/core/FileViewer.jsx
```diff
@@ -0,0 +1,97 @@
...
+  const [isOpen, setIsOpen] = useState(true);
+  // error message for failed load
+  const [error, setError] = useState(null);
+  // the file content
+  const [content, setContent] = useState("");
+  // current state
+  const [state, setState] = useState("init");
+
+  useEffect(() => {
+    if (state !== "init") return;
```
dgdavid, line 49: I have to test it, but I think you do not need that _"init" state_. If you do not use dependencies in the `useEffect` it will be executed only once. So, you drop the _state_ thingy and directly relay in _error_ and _content_. That is, displaying the spinner until either, error or content is not null.

dgdavid (commented):
Looks great, just a few comments

web/src/components/core/FileViewer.jsx
```diff
@@ -0,0 +1,97 @@
...
+    setIsOpen(false);
+    if (onCloseCallback) onCloseCallback();
+  };
+
+  return (
+    <>
+      <Popup
+        isOpen={isOpen}
+        title={title || file}
+        variant="large"
```
dgdavid, line 75: The fragment (aka `<>`) is not longer needed here.

dgdavid, line 72: Not related with this PR, but leaving here as a note for myself. We have a `height` Popup prop that somehow collides with the variant. The intention of this prop is to limit the height and to avoid _"blinking"_ effect we have here when changing from loading to the file content. But the actual size for _height="large"_ looks not enough. We can address and improve the Popup behavior later.

web/src/components/core/FileViewer.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+import cockpit from "../../lib/cockpit";
+
+import "./fileviewer.scss";
+
+// file loading indicator
+const spinner = () => {
+  return (
+    // set max. height to avoid resizing after the file is loaded (no "blinking")
+    <div style={{ height: "100vh" }}>
```
dgdavid, line 35: Sorry for being nitpicker here, but this will not work as expected since the _spinner_ is a child of the _Popup_. Thus, the blinking will happen when instead of the file content you end up with an error (or simply when displaying a file with just a bunch of lines).

What is more, since you are using viewport units in a child… scroll will appear even when it's not needed.

Take a look at the screenshots below

| Spinner with scrollbars | The popup with an alert | The popup displaying small file |
|-|-|-|
|Screenshot from 2023-02-06 11-52-48 | Screenshot from 2023-02-06 11-54-15 | Screenshot from 2023-02-06 11-56-17 |

Last but not least, I'd not use styles directly in the components unless strictly needed.

So, my suggestion if you want to fix this `Popup` _misbehavior_ here without touching other files, is to add a utility class "max-height", "full-height", "tallest", or so and use it with the popup. I.e., something like


```diff
diff --git a/web/src/assets/styles/utilities.scss b/web/src/assets/styles/utilities.scss
index bdbcd0c52..110b8901b 100644
--- a/web/src/assets/styles/utilities.scss
+++ b/web/src/assets/styles/utilities.scss
@@ -98,3 +98,10 @@
 .bottom-shadow {
   box-shadow: 0px -3px 10px 0px var(--color-gray-darker);
 }
+
+.tallest {
+  /** block-size fallbacks **/
+  height: 95dvh;
+  /** END block-size fallbacks **/
+  block-size: 95dvh;
+}
diff --git a/web/src/components/core/FileViewer.jsx b/web/src/components/core/FileViewer.jsx
index 7fe523429..837aea0cc 100644
--- a/web/src/components/core/FileViewer.jsx
+++ b/web/src/components/core/FileViewer.jsx
@@ -28,16 +28,6 @@ import cockpit from "../../lib/cockpit";
 
 import "./fileviewer.scss";
 
-// file loading indicator
-const spinner = () => {
-  return (
-    // set max. height to avoid resizing after the file is loaded (no "blinking")
-    <div style={{ height: "100vh" }}>
-      <LoadingEnvironment text="Reading file..." />
-    </div>
-  );
-};
-
 export default function FileViewer({ file, title, onCloseCallback }) {
   // the popup is visible
   const [isOpen, setIsOpen] = useState(true);
@@ -72,8 +62,9 @@ export default function FileViewer({ file, title, onCloseCallback }) {
       isOpen={isOpen}
       title={title || file}
       variant="large"
+      className="tallest"
     >
-      { state === "loading" && spinner() }
+      { state === "loading" && <LoadingEnvironment text="Reading file..." /> }
       { (content === null || error) &&
         <Alert
           isInline
```

Also remember that, whenever possible, we're trying to use the [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) with minimal fallbacks that, hopefully, we could automate in a near future :)

web/src/components/core/fileviewer.scss
```diff
@@ -0,0 +1,7 @@
+
+.filecontent {
```
dgdavid, line 2: Please, remember to move this to the _src/assets/styles/blocks.css_ to avoid using ad-hoc CSS files by now.

dgdavid (approved):
LGTM :100:

PR #417, opened 2023-02-07
Y2log filtering

lslezak:
## Problem

Improve the y2log viewer, related to #407 

- Error level highlighting (warnings in orange, errors in red) to quickly spot the problems
- Allow filtering by log level, e.g. display only errors and warnings
- Allow filtering by component, e.g. display only libzypp or libstorage messages
- By default hide the not important data in the view (like date, host name, PID)
- Also automatically scroll to end of the y2log, the most interesting part is usually at the very end

## Questions and TODOs

- [ ] Split the code to a separate component (move from `components/core` to `components/logviewer` or so?)
- [ ] Fix the SCSS (get the values from PatternFly SCSS)
- [ ] Fix the layout, replace the `<div>` workaround with `<Flex>` PatternFly components (maybe still needs a custom CSS)
- [ ] Add more comments
- [ ] Add unit tests, fix broken ones

## Testing

- Tested manually

## Screenshots

An example how to display only the libstorage errors and warnings

y2log_filter

### DIFF
--- a/web/src/assets/styles/index.scss
+++ b/web/src/assets/styles/index.scss
@@ -7,6 +7,7 @@
 @use "~/assets/styles/utilities.scss";
 @use "~/assets/styles/composition.scss";
 @use "~/assets/styles/blocks.scss";
+@use "~/assets/styles/y2logviewer.scss";
 
 // PatternFly overrides
 @use "~/assets/styles/patternfly-overrides.scss";
--- a/web/src/assets/styles/y2logviewer.scss
+++ b/web/src/assets/styles/y2logviewer.scss
@@ -1,0 +1,87 @@
+@use "sass:color";
+
+// TODO: get the base text color from the PatternFly SCSS
+$base-color: #151515;
+
+// lighter color for less important log values
+$faded-color: color.scale($base-color, $lightness: +40%);
+
+// colors for each log level
+$loglevel-0-color: color.scale($faded-color, $lightness: +20%);
+$loglevel-2-color: #fdd3a2;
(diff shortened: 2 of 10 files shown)


dgdavid:
> * Split the code to a separate component (move from `components/core` to `components/logviewer` or so?)

Good question. Until now, we're not splitting so granular but maybe this case deserves its own namespace. Not sure, we should evaluate how reusable are the blocks used to build it.

> 
> * Fix the SCSS (get the values from PatternFly SCSS)

I prefer to do it in the other way around: to use our own CSS Custom Properties and overrides PatternFly is needed.

> 
> * Fix the layout, replace the `<div>` workaround with `<Flex>` PatternFly components (maybe still needs a custom CSS)
> 

No please, don't do that ;) Some PR's ago I proposed to stop using PatternFly for laying out the UI and use it only as a component library... and it got merged https://github.com/yast/d-installer/pull/391. I'm not against of rolling back that decision if needed, but let's first explore how far we can walk alone without adding more DOM nodes than needed.

> * Add more comments
> 
> * Add unit tests, fix broken ones

dgdavid:
> 
> If we can make it without the PatternFly layouts and have few our own helpers I'm fine with that. We just should avoid re-implementing what PatternFly already has, no need to reinvent the wheel...

I know what you mean and, for sure, we're in the same page to that regard. But let's see it in another way: at this point we would like to use a bike wheel, not a truck one :)

lslezak:
Closing, this is low priority and makes sense only for YaST experts anyway...

kobliha:
But it's lovely :) I like that.

PR #414, opened 2023-01-26
Properly inform the user about software proposal problems

imobachgs:
## Problem

Recently, we added a *Software* section that informs the user about missing packages and the installation size. However, it has a few problems:

* When the network is not available (at this time, D-Installer uses online repos), it just reports that there are missing packages.
* While reading the repositories, the user has no feedback about the progress.

## Solution

* Display progress information while reading the repositories.
* Inform the user about problems reading the repositories.
* Add a mechanism to force refreshing the repositories.
* Reread the repositories if the network becomes available again.

## Screenshots

<details>
<summary>Progress information</summary>

Captura desde 2023-02-09 09-47-36
</details>

<details>
<summary>Report errors about unreachable repositories</summary>

Captura desde 2023-02-09 09-50-26
</details>

<details>
<summary>Do not inform about missing packages/patterns where there are no reachable repositories</summary>

Captura desde 2023-02-09 09-59-48
</details>

## To do

- [x] Move most of the proposal logic to a separate class ([DInstaller::Software::Manager](https://github.com/yast/d-installer/blob/master/service/lib/dinstaller/software/manager.rb#L41) contains a lot of logic).
- [x] Inform the user when a repository is not accessible.
- [x] Do not try to make a proposal if there are no repositories.
- [x] Display some progress in the corresponding section or, at least, explain what is happening.
- [x] Add a mechanism to allow refreshing the repositories (#388)
- [x] Refresh the repositories metadata when the network is available again

## Testing

- *Added a new unit test*
- *Tested manually*

PR #412, opened 2023-01-25
Storage D-Bus reorganization

joseivanlopez:
## Problem

The storage D-Bus API is going to be extended in order to offer iSCSI iniciator support for discovering targets and logging. While researching about how to define the iSCSI API, we found some flaws in the current organization of the storage D-Bus API:

* The proposal object (`/DInstaller/Storage/Proposal1`) has too many responsabilites: a) it provides neccesary info for calculating a proposal (e.g., available devices), b) calculates a new proposal, and c) provides info about the result.
* The storage service status is managed by two different objects: `/DInstaller/Storage1` and `/DInstaller/Storage/Proposal1`.
* D-Bus API versioning is wrong.

Those issues should be fixed before extending the API with iSCSI support.

Part of https://trello.com/c/iQ7BCexU/3261-5-d-installer-d-bus-interface-configure-the-iscsi-initiator-and-the-iscsi-targets.

## Solution

The storage D-Bus API was re-organized to fix some of the dectected issues.

The API was structured in the following way:

~~~
/DInstaller/Storage1
  .DInstaller.Storage1
  .DInstaller1.ServiceStatus
  .DInstaller1.Progress
  .DInstaller1.Validation
/DInstaller/Storage1/Proposal
  .DInstaller.Storage1.Proposal
~~~

And now the new API organization is:

~~~
/DInstaller/Storage1
  .ObjectManager
  .DInstaller.ServiceStatus1
  .DInstaller.Progress1
  .DInstaller.Validation1
  .DInstaller.Storage1
  .DInstaller.Storage1.Proposal.Calculator
    #Calculate
    #Result -> o
    #VolumeTemplates 
    #AvailableDevices
/DInstaller/Storage1/Proposal
  .DInstaller.Storage1.Proposal
    #LVM
    #EncryptionPassword
    #CandidateDevices
    #Volumes 
    #Actions
~~~

The main changes are:

* Object `/DInstaller/Storage1` implements a new iface `.DInstaller.Storage1.Proposal.Calculator`. That iface offers an API with all the methods/properties for calculating a proposal (available devices, volume templates, etc).
* Object `/DInstaller/Storage1/Proposal` will be dynamic, and it will be present in the object tree only if a proposal was calculated (successful or not). For example, during the *startup phase* there is no proposal yet.
* Objects and interfaces versioning is fixed.

Bonus: some refactoring was done for `questions` service, removing unnecessary code.

More improvements will come in the future to address these issues:

* https://github.com/yast/d-installer/issues/408
[...]

PR #418, opened 2023-02-08
Do not crash when there is no storage proposal yet

dgdavid:
## Problem

After commit https://github.com/yast/d-installer/commit/f0db9e4420056fa72d4ce3e7e2acc53841a7e7c9, the UI crahs "silently" the first time a product is selected after running the service. Hard to explain why (maybe @imobachgs or @joseivanlopez can edit this description for adding more info)

## Solution

To protect the code by using the [optional chaining (?) operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) **AND** intercepting the broken promise for returning `undefined`. I.e., adapting the code for returning an empty object when the proposal does not exist yet or is not `.valid`.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

| Before | After |
|:-:|:-:|
| Screenshot from 2023-02-08 16-16-15| Screenshot from 2023-02-08 17-27-41 |


web/src/client/storage.js
```diff
@@ -50,10 +50,10 @@ class StorageBaseClient {
    */
   async getProposal() {
     const storageProxy = await this.client.proxy(PROPOSAL_CALCULATOR_IFACE, STORAGE_PATH);
-    const proposalProxy = await this.client.proxy(PROPOSAL_IFACE);
+    const proposalProxy = await this.client.proxy(PROPOSAL_IFACE).catch(() => undefined);
 
     // Check whether proposal object is already exported
-    if (!proposalProxy.valid) return {};
+    if (!proposalProxy?.valid) return {};
```
dgdavid, line 56: Looks like we cannot comment in [untouched lines yet](https://github.com/github/roadmap/issues/347), so let's use this one for commenting another approach we could follow in this method.

The idea is simple but can be split in two

* To use a try/catch block for surrounding the full method. I.e., returning an _empty proposal_ whenever an error happens. Note that here we are getting a broken promise because `cockpit.dbus.proxy` is crashing due to something (?) missing in the Dbus side.

  <details>
  <summary>Click to show/hide a diff</summary>

  ---

  ```diff
  diff --git a/web/src/client/storage.js b/web/src/client/storage.js
  index 086e68a38..dccf0e08c 100644
  --- a/web/src/client/storage.js
  +++ b/web/src/client/storage.js
  @@ -49,51 +49,57 @@ class StorageBaseClient {
      * @return {Promise<object>}
      */
     async getProposal() {
  -    const storageProxy = await this.client.proxy(PROPOSAL_CALCULATOR_IFACE, STORAGE_PATH);
  -    const proposalProxy = await this.client.proxy(PROPOSAL_IFACE).catch(() => undefined);
  -
  -    // Check whether proposal object is already exported
  -    if (!proposalProxy?.valid) return {};
  -
  -    const volume = dbusVolume => {
  -      const valueFrom = dbusValue => dbusValue?.v;
  +    const emptyProposal = {};
  +
  +    try {
  +      const storageProxy = await this.client.proxy(PROPOSAL_CALCULATOR_IFACE, STORAGE_PATH);
  +      const proposalProxy = await this.client.proxy(PROPOSAL_IFACE);
  +
  +      // Check whether proposal object is already exported
  +      if (!proposalProxy.valid) return emptyProposal;
  +
  +      const volume = dbusVolume => {
  +        const valueFrom = dbusValue => dbusValue?.v;
  +
  +        const valuesFrom = (dbusValues) => {
  +          if (dbusValues === undefined) return [];
  +          return dbusValues.v.map(valueFrom);
  +        };
  +
  +        return {
  +          mountPoint: valueFrom(dbusVolume.MountPoint),
  +          optional: valueFrom(dbusVolume.Optional),
  +          deviceType: valueFrom(dbusVolume.DeviceType),
  +          encrypted: valueFrom(dbusVolume.Encrypted),
  +          fsTypes: valuesFrom(dbusVolume.FsTypes),
  +          fsType: valueFrom(dbusVolume.FsType),
  +          minSize: valueFrom(dbusVolume.MinSize),
  +          maxSize: valueFrom(dbusVolume.MaxSize),
  +          fixedSizeLimits: valueFrom(dbusVolume.FixedSizeLimits),
  +          adaptiveSizes: valueFrom(dbusVolume.AdaptiveSizes),
  +          snapshots: valueFrom(dbusVolume.Snapshots),
  +          snapshotsConfigurable: valueFrom(dbusVolume.SnapshotsConfigurable),
  +          snapshotsAffectSizes: valueFrom(dbusVolume.SnapshotsAffectSizes),
  +          sizeRelevantVolumes: valueFrom(dbusVolume.SizeRelevantVolumes)
  +        };
  +      };
   
  -      const valuesFrom = (dbusValues) => {
  -        if (dbusValues === undefined) return [];
  -        return dbusValues.v.map(valueFrom);
  +      const action = dbusAction => {
  +        const { Text: { v: textVar }, Subvol: { v: subvolVar }, Delete: { v: deleteVar } } = dbusAction;
  +        return { text: textVar, subvol: subvolVar, delete: deleteVar };
         };
   
         return {
  -        mountPoint: valueFrom(dbusVolume.MountPoint),
  -        optional: valueFrom(dbusVolume.Optional),
  -        deviceType: valueFrom(dbusVolume.DeviceType),
  -        encrypted: valueFrom(dbusVolume.Encrypted),
  -        fsTypes: valuesFrom(dbusVolume.FsTypes),
  -        fsType: valueFrom(dbusVolume.FsType),
  -        minSize: valueFrom(dbusVolume.MinSize),
  -        maxSize: valueFrom(dbusVolume.MaxSize),
  -        fixedSizeLimits: valueFrom(dbusVolume.FixedSizeLimits),
  -        adaptiveSizes: valueFrom(dbusVolume.AdaptiveSizes),
  -        snapshots: valueFrom(dbusVolume.Snapshots),
  -        snapshotsConfigurable: valueFrom(dbusVolume.SnapshotsConfigurable),
  -        snapshotsAffectSizes: valueFrom(dbusVolume.SnapshotsAffectSizes),
  -        sizeRelevantVolumes: valueFrom(dbusVolume.SizeRelevantVolumes)
  +        availableDevices: storageProxy.AvailableDevices.map(([id, label]) => ({ id, label })),
  +        candidateDevices: proposalProxy.CandidateDevices,
  +        lvm: proposalProxy.LVM,
  +        encryptionPassword: proposalProxy.EncryptionPassword,
  +        volumes: proposalProxy.Volumes.map(volume),
  +        actions: proposalProxy.Actions.map(action)
         };
  -    };
  -
  -    const action = dbusAction => {
  -      const { Text: { v: textVar }, Subvol: { v: subvolVar }, Delete: { v: deleteVar } } = dbusAction;
  -      return { text: textVar, subvol: subvolVar, delete: deleteVar };
  -    };
  -
  -    return {
  -      availableDevices: storageProxy.AvailableDevices.map(([id, label]) => ({ id, label })),
  -      candidateDevices: proposalProxy.CandidateDevices,
  -      lvm: proposalProxy.LVM,
  -      encryptionPassword: proposalProxy.EncryptionPassword,
  -      volumes: proposalProxy.Volumes.map(volume),
  -      actions: proposalProxy.Actions.map(action)
  -    };
  +    } catch {
  +      return emptyProposal;
  +    }
     }
   
     /**
   ```
  </details>  

* To move these _auxiliary_ method (`volume`, `actions`) outside. This could apply for the _dbusVolume_ and _dbusVolumes_ in the method below.


  <details>
  <summary>Click to show/hide a diff</summary>

  ---

  ```diff

    diff --git a/web/src/client/storage.js b/web/src/client/storage.js
  index 086e68a38..6086f5d46 100644
  --- a/web/src/client/storage.js
  +++ b/web/src/client/storage.js
  @@ -43,57 +43,63 @@ class StorageBaseClient {
       this.client = new DBusClient(STORAGE_SERVICE, address);
     }
   
  +  volume(dbusVolume) {
  +    const valueFrom = dbusValue => dbusValue?.v;
  +
  +    const valuesFrom = (dbusValues) => {
  +      if (dbusValues === undefined) return [];
  +      return dbusValues.v.map(valueFrom);
  +    };
  +
  +    return {
  +      mountPoint: valueFrom(dbusVolume.MountPoint),
  +      optional: valueFrom(dbusVolume.Optional),
  +      deviceType: valueFrom(dbusVolume.DeviceType),
  +      encrypted: valueFrom(dbusVolume.Encrypted),
  +      fsTypes: valuesFrom(dbusVolume.FsTypes),
  +      fsType: valueFrom(dbusVolume.FsType),
  +      minSize: valueFrom(dbusVolume.MinSize),
  +      maxSize: valueFrom(dbusVolume.MaxSize),
  +      fixedSizeLimits: valueFrom(dbusVolume.FixedSizeLimits),
  +      adaptiveSizes: valueFrom(dbusVolume.AdaptiveSizes),
  +      snapshots: valueFrom(dbusVolume.Snapshots),
  +      snapshotsConfigurable: valueFrom(dbusVolume.SnapshotsConfigurable),
  +      snapshotsAffectSizes: valueFrom(dbusVolume.SnapshotsAffectSizes),
  +      sizeRelevantVolumes: valueFrom(dbusVolume.SizeRelevantVolumes)
  +    };
  +  }
  +
  +  action(dbusAction) {
  +    const { Text: { v: textVar }, Subvol: { v: subvolVar }, Delete: { v: deleteVar } } = dbusAction;
  +    return { text: textVar, subvol: subvolVar, delete: deleteVar };
  +  }
  +
     /**
      * Returns storage proposal values
      *
      * @return {Promise<object>}
      */
     async getProposal() {
  -    const storageProxy = await this.client.proxy(PROPOSAL_CALCULATOR_IFACE, STORAGE_PATH);
  -    const proposalProxy = await this.client.proxy(PROPOSAL_IFACE).catch(() => undefined);
  +    const emptyProposal = {};
   
  -    // Check whether proposal object is already exported
  -    if (!proposalProxy?.valid) return {};
  +    try {
  +      const storageProxy = await this.client.proxy(PROPOSAL_CALCULATOR_IFACE, STORAGE_PATH);
  +      const proposalProxy = await this.client.proxy(PROPOSAL_IFACE);
   
  -    const volume = dbusVolume => {
  -      const valueFrom = dbusValue => dbusValue?.v;
  -
  -      const valuesFrom = (dbusValues) => {
  -        if (dbusValues === undefined) return [];
  -        return dbusValues.v.map(valueFrom);
  -      };
  +      // Check whether proposal object is already exported
  +      if (!proposalProxy.valid) return emptyProposal;
   
         return {
  -        mountPoint: valueFrom(dbusVolume.MountPoint),
  -        optional: valueFrom(dbusVolume.Optional),
  -        deviceType: valueFrom(dbusVolume.DeviceType),
  -        encrypted: valueFrom(dbusVolume.Encrypted),
  -        fsTypes: valuesFrom(dbusVolume.FsTypes),
  -        fsType: valueFrom(dbusVolume.FsType),
  -        minSize: valueFrom(dbusVolume.MinSize),
  -        maxSize: valueFrom(dbusVolume.MaxSize),
  -        fixedSizeLimits: valueFrom(dbusVolume.FixedSizeLimits),
  -        adaptiveSizes: valueFrom(dbusVolume.AdaptiveSizes),
  -        snapshots: valueFrom(dbusVolume.Snapshots),
  -        snapshotsConfigurable: valueFrom(dbusVolume.SnapshotsConfigurable),
  -        snapshotsAffectSizes: valueFrom(dbusVolume.SnapshotsAffectSizes),
  -        sizeRelevantVolumes: valueFrom(dbusVolume.SizeRelevantVolumes)
  +        availableDevices: storageProxy.AvailableDevices.map(([id, label]) => ({ id, label })),
  +        candidateDevices: proposalProxy.CandidateDevices,
  +        lvm: proposalProxy.LVM,
  +        encryptionPassword: proposalProxy.EncryptionPassword,
  +        volumes: proposalProxy.Volumes.map(this.volume),
  +        actions: proposalProxy.Actions.map(this.action)
         };
  -    };
  -
  -    const action = dbusAction => {
  -      const { Text: { v: textVar }, Subvol: { v: subvolVar }, Delete: { v: deleteVar } } = dbusAction;
  -      return { text: textVar, subvol: subvolVar, delete: deleteVar };
  -    };
  -
  -    return {
  -      availableDevices: storageProxy.AvailableDevices.map(([id, label]) => ({ id, label })),
  -      candidateDevices: proposalProxy.CandidateDevices,
  -      lvm: proposalProxy.LVM,
  -      encryptionPassword: proposalProxy.EncryptionPassword,
  -      volumes: proposalProxy.Volumes.map(volume),
  -      actions: proposalProxy.Actions.map(action)
  -    };
  +    } catch {
  +      return emptyProposal;
  +    }
     }
   
     /**
  ```
  </details>

I'm not saying that the method extraction should be done in that way. At the time of writing I have not clear how to do it (as it is in the diff but with proper documentation? using [private methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields)? moving them even further to an storage utils module?) I'm just comenting few changes I believe could simplify both, the code and the test.

But sadly I have no enough understanding about what is going wrong in the backend and that's why I'd like your thoughs here :)

dgdavid (commented):
Please guys, take your time to review this issue. 

Would be nice if any of you can clarify better than me the reason of this crash. As far as I know, it's related with changes introduced in #412 and the fact that 

> Object `/DInstaller/Storage1/Proposal` will be dynamic, and **it will be present in the object tree only if a proposal was calculated** (successful or not). For example, during the startup phase there is no proposal yet.

which makes `cockpit.dbus.proxy` not happy and the promise ends up broken. That's why the `catch` is needed for returning an _empty proposal_.

web/src/components/core/FileViewer.test.jsx
```diff
@@ -28,19 +28,19 @@ import cockpit from "../../lib/cockpit";
...
     read: readFn
-  }
+  };
 });
 
 cockpit.file.mockImplementation(fileFn);
 
 // testing data
-const file_name = "/testfile"
+const file_name = "/testfile";
```
dgdavid, line 43: Actually not related with the issue discussed here, but _needed_ to make ESLint happy.

imobachgs: I had the same problem in #414. After all pending PRs are merged, we should do something about that.

web/src/client/storage.js
imobachgs: I would limit to fix the original problem. About moving those functions, I do not have a strong opinion. Both of them look good to me.

imobachgs (approved):

imobachgs (approved):
Much better now 😄

PR #420, opened 2023-02-10
Reduce blinking

imobachgs:
## Problem

The `<App>` component listens for the manager status indicator, which changes at the end of the startup phase. In that case, the whole UI gets refreshed, although it is not needed. It is annoying because if you have opened a popup, it is automatically closed once all the modules are ready (and the status changes to 'idle').

## Solution

Do not make `<App>` listen for the status, just listen where needed (`LoadingEnvironment` and a new `Installation` one).

## Testing

- *Update the new unit test*
- *Tested manually*


dgdavid (approved):
LGTM. I have tested it manually too and the annoying blinking has gone. Nevertheless, we could keep improving this area in the future by _decoupling_ a bit the routing from the `<App />` component.

Thanks Imo!

PR #424, opened 2023-02-14
[service] Start the daemon in the systemd service

imobachgs:
## Problem

* D-Installer components crash in the installation media because the services are not started properly.

## Solution

On the one hand, we are *spawning* a process with the `dbus-daemon`. The rest of the services are activated properly.

After implementing the previous fix, I detected that the UI cannot get the validation errors for the storage layer. It cannot create the proxy and the component is not updated anymore. To fix this problem, I changed the implementation of the `getValidationErrors()` method to handle potential errors and not rely on proxies (better performance).

## Testing

- *Tested manually*


joseivanlopez (approved):
LGTM

joseivanlopez (approved):

PR #426, opened 2023-02-15
Added terminal

lslezak:
## Problem

To make debugging easier it should be possible to open a terminal directly from the D-Installer.

## Solution

Embed the cockpit terminal into an `iframe`. See the cockpit documentation for more details:
  - https://cockpit-project.org/guide/latest/embedding.html#embedding-components
  - https://cockpit-project.org/guide/latest/api-terminal-html.html

## Testing

- Tested manually
- Added unit tests

## Screenshots

terminal

### DIFF
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Feb 16 12:56:24 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Integrate cockpit terminal application
+
+-------------------------------------------------------------------
 Mon Feb 13 14:30:56 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Version 0.7
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -4,9 +4,11 @@ Thu Feb 16 12:56:24 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
 - Integrate cockpit terminal application
 
 -------------------------------------------------------------------
-Mon Feb 13 14:30:56 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+Wed Feb 15 16:35:54 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Version 0.7
+- Do not use a proxy to get the errors lists
+  (gh#yast/d-installer#424).
 
 -------------------------------------------------------------------
 Thu Feb  9 15:29:58 UTC 2023 - David Diaz <dgonzalez@suse.com>
--- a/web/package/cockpit-d-installer.changes
+++ b/web/package/cockpit-d-installer.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Feb 16 12:56:24 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
 
-- Integrate cockpit terminal application
+- Integrate cockpit terminal application (gh#yast/d-installer#426)
 
 -------------------------------------------------------------------
 Wed Feb 15 16:35:54 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -94,3 +94,8 @@ section > .content {
   word-break: break-all;
   white-space: pre-wrap;
 }
+
+iframe.terminal {
+  height: 100%;
+  width: 100%;
+}
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -94,8 +94,3 @@ section > .content {
   word-break: break-all;
   white-space: pre-wrap;
 }
-
-iframe.terminal {
-  height: 100%;
-  width: 100%;
-}
--- a/web/src/assets/styles/utilities.scss
+++ b/web/src/assets/styles/utilities.scss
@@ -105,3 +105,9 @@
   /** END block-size fallbacks **/
   block-size: 95dvh;
 }
+
+// fill all available space
+.fill {
+  block-size: 100%;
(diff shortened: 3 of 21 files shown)


web/src/assets/styles/blocks.scss
```diff
@@ -94,3 +94,8 @@ section > .content {
   word-break: break-all;
   white-space: pre-wrap;
 }
+
+iframe.terminal {
+  height: 100%;
```
dgdavid, line 99: And please, remember to use block-size and inline-size instead :)

web/src/assets/styles/blocks.scss
```diff
@@ -94,3 +94,8 @@ section > .content {
   word-break: break-all;
   white-space: pre-wrap;
 }
+
+iframe.terminal {
```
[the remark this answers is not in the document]
dgdavid: > I think this should not be a "block" but an "utility". I.e., adding the ".full-height" and ".full-width" utilities CSS classes into the "utilities.scss" and using them as classNames for the iframe.

Or using even a

```css
.fill {
  block-size: 100%;
  inline-size: 100%
}
```

in the utilities.scss, but not coupling this "fill all the available area" CSS class to the "terminal". We can use it later for telling other elements to fill the available area too.

web/src/components/core/ShowTerminalButton.jsx
```diff
@@ -0,0 +1,63 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React, { useState } from "react";
+import { Terminal } from "~/components/core";
+import { Icon } from "~/components/layout";
+import { Button } from "@patternfly/react-core";
```
dgdavid, line 25: NP: I used to put the imports from library first, before ours  :)

web/src/components/core/ShowTerminalButton.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  describe("when user clicks on it", () => {
+    it("displays the Terminal component", async () => {
+      const { user } = plainRender(<ShowTerminalButton />);
+      const button = screen.getByRole("button", "Terminal");
+      await user.click(button);
+      screen.getByText(/Terminal Mock/);
+    });
+  });
```
dgdavid, line 43: NP: for me, this test could be simplified by just using the second `it` and renaming it to something like `"renders a button that allows displaying the Terminal component"` or so. 

It's already testing that the button is rendered (it was found), that it is not disabled (click event didn't failed), and that it behaves as expected (the terminal mock was rendered) in just one shot.

I know that we used to write more granular test when using RSpec, but I feel that testing library is more about testing the use case.

Of course, I can be wrong. Either way, it is just a suggestion, your unit test is fine.

web/src/components/core/Terminal.jsx
```diff
@@ -0,0 +1,52 @@
...
+  // https://cockpit-project.org/guide/latest/api-terminal-html.html
+  return (
+    <Popup
+      isOpen={isOpen}
+      variant="large"
+      className="tallest"
+      aria-label="terminal popup"
+    >
+
+      <iframe data-testid="iframe-terminal" className="fill" src="/cockpit/@localhost/system/terminal.html" />
```
dgdavid, line 45: I'd avoid the use of data-testid as much as possible. What is more, I'd use the `title` here for both, accessibility (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#accessibility_concerns) and testing (getByTitle).

Regarding a11n, I know that this particular iframe is alone and inside a popup that already describes what the user will find in it... but in any case...

web/src/components/core/Terminal.test.jsx
```diff
@@ -0,0 +1,56 @@
...
+import React from "react";
+
+import { screen, waitFor, within } from "@testing-library/react";
+import { plainRender } from "~/test-utils";
+import { Terminal } from "~/components/core";
+
+describe("Terminal", () => {
+  it("displays the cockpit terminal in an iframe", async () => {
+    plainRender(<Terminal />);
+    const iframe = screen.getByTestId("iframe-terminal");
```
dgdavid, line 31: As said above, I'd use title/getByTitle intead of the data-testid

dgdavid (commented):
Thanks @lslezak. I left a few comments there and I'll check how we can get ride of the popup scrollbars.

Other than that, remember to update the changelog!

web/src/assets/styles/utilities.scss
```diff
@@ -105,3 +105,9 @@
   /** END block-size fallbacks **/
   block-size: 95dvh;
 }
+
+// fill all available space
+.fill {
+  block-size: 100%;
+  inline-size: 100%;
```
dgdavid, line 112: I have been playing a bit with the Terminal and its Popup wrapper. To avoid the Popup scrollbars, you can

* To use a block size slightly smaller than 100%. `99%` or `calc(100% - 6px)` (not sure why 6px, do not have time now for further investigation), **or**
* To use the `.vertically-centered` CSS class instead. Looks like the CSS Grid manage the space better.

I'd use the second option by now.

web/src/components/core/Terminal.jsx
lslezak: That `aria-label` is actually needed by an unit test, the `Modal` component complains that there is not title set. But the "Terminal" title is already included inside the `iframe` so it does not make sense to have it twice.

web/package/cockpit-d-installer.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Feb 16 12:56:24 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Integrate cockpit terminal application
```
imobachgs, line 4: Please, add a reference to the PR (gh#yast/d-installer#426). It makes it easier to share the changes later in the blog.

imobachgs (commented):
It looks good to me. Just, please, add the PR number.

imobachgs (approved):
Given that the issues pointed out by @dgdavid are fixed and that it contains the PR number, I would say we can merge. Thanks a lot!

PR #439, opened 2023-02-24
Improved development server

lslezak:
Allow running the D-Installer module from a local machine, forward the connection to the target Cockpit instance.

## Problem

- The current solution contains several hacks
- It is difficult to start developing the web frontend
- Hot Module Replacement does not work as expected

## Solution

- Instead of putting the development server under the Cockpit server do it the other way round. Make the development server the main entry point and proxy the Cockpit requests to the real server

## Details 

- `COCKPIT_TARGET` environment  variable sets the Cockpit target server, if not set it uses the local running Cockpit (`https://localhost:9090`)

## Testing

- Tested manually


web/README.md
```diff
@@ -14,33 +26,58 @@ link your build folder (`dist`) or just rely on the `devel-install` task:
...
-the command below. But please, do not forget that you have to reload the code
-in your browser manually.
+### Using a development server
 
+To start the [webpack-dev-server](https://github.com/webpack/webpack-dev-server)
+use this command:
 
 ```
-    npm run watch
+    npm run server -- --open
```
imobachgs, line 50: The `--open` parameter does not work if the regular case. It tries to open `http://localhost:8080` and it does not work. It should open `https://localhost:9090/cockpit/@localhost/d-installer/index.html` (or, at least, `http://localhost:9090`).

imobachgs (changes requested):
It looks like a nice addition but, at least for me, it breaks the simple case of having Cockpit and the front-end on the same machine. Even if I do not use the `--open` argument, it does not work as it keeps reloading the page time after time.

imobachgs:
In general, I find using a virtual machine (with a secondary disk, so the storage proposal succeeds) easy enough. Then, if you have a good connection with the machine, using X-forwarding just works: `ssh -X yourmachine firefox`. HMR will work with no problems in that scenario.

imobachgs:
Things are starting to get complex, so I would like to add more information about the problem and why we are going through this solution.

> ## Problem
>
>     * The current solution contains several hacks

Which hacks? It would be nice to document them.
 
>     * It is difficult to start developing the web frontend

Actually, the real blocker is that `cockpit` is not available in openSUSE Leap. If you are using Tumbleweed, it is pretty straightforward. However, when you try to use the ISO (or a separate VM) for development, you get into the HMR trap (see below).

>     * Hot Module Replacement does not work as expected

The problem is that if you try to use it on a remote system (not using localhost), it will not properly connect to the HMR. Perhaps it is easy to solve (should we try?) but, as a workaround, going through X forwarding solves the issue (`ssh -X mymachine firefox 'http://localhost:9090/cockpit/@localhost/d-installer/index.html`). If you do not have a good connection to the system you are trying, then you are in trouble 😄

imobachgs:
About the current solution, it works just fine :-)

I only have two minor objections, but none of them are important (although I would like to debug the 2nd one):

* Now we use a different mechanism for loading the code in production and development. It is not a problem but something to consider if we detect some issues in the future.
* For some reason, I do not get administrative access unless I log in as "root". As I said, I would like to find out why it behaves differently, but it is not a problem.

So once you consider it done, we can start with the code review process.

Thanks!

imobachgs:
Oh, I forgot to comment on something. If at some point we decide to get rid of Cockpit's login form (and serve the UI with a different process), this code could actually be useful even in production.

lslezak:
> Which hacks? It would be nice to document them.

I meant mainly the `manifest.dev.json` file which duplicates lot of things from `manifest.json` and needs to setup special CSP rules. That's ugly and fragile.

And that file is removed by this PR so there is nothing to document. :wink: 

> The problem is that if you try to use it on a remote system (not using localhost), it will not properly connect to the HMR. Perhaps it is easy to solve (should we try?) but, as a workaround, going through X forwarding solves the issue (`ssh -X mymachine firefox 'http://localhost:9090/cockpit/@localhost/d-installer/index.html`). If you do not have a good connection to the system you are trying, then you are in trouble smile

Yes, I tried that workaround too. But the problem is that then you are limited by the installed browsers (FireFox only), you cannot easily test it in Chrome/Chromium or other browsers and you do not have your plugins available. I use the [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) plugin in Chrome to inspect the properties of the components.

lslezak:
> * Now we use a different mechanism for loading the code in production and development. It is not a problem but something to consider if we detect some issues in the future.

Yes, the loading mechanism is different, but the difference is very tiny. For plugins Cockpit behaves like a static web server serving static files (*.html, *.css, *.js) and that's it. The only difference is that when it detects you are not logged in it will return the login page instead of the files belonging to the requested plugin. In theory you could even load the plain files without any web server (via `file://`).

> * For some reason, I do not get administrative access unless I log in as "root". As I said, I would like to find out why it behaves differently, but it is not a problem.

Strange, no idea why it does not work for you. I can log into Cockpit without any problems.

web/src/DevServerWrapper.jsx
```diff
@@ -0,0 +1,112 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React, { useEffect, useRef, useState } from "react";
+import { Button, Title, Text, EmptyState, EmptyStateIcon, EmptyStateBody } from "@patternfly/react-core";
+import { Center, Icon, Loading } from "~/components/layout";
+
+// path to any internal Cockpit component to force displaying the login dialog
+const loginPath = "/cockpit/@localhost/system/terminal.html";
```
dgdavid, line 27: It looks like the path for the terminal component. What am I missing?

web/src/DevServerWrapper.jsx
```diff
@@ -0,0 +1,112 @@
...
+      // 401 = Authentication failed
+      else if (xhr.status === 401)
+        setIsAuthenticated(false);
+      else
+        setIsError(true);
+
+      setIsLoading(false);
+    };
+    xhr.timeout = 5000;
+    xhr.open("GET", "/cockpit/login");
```
dgdavid, line 67: This looks like the right path to login.

web/src/DevServerWrapper.jsx
lslezak: Yes, it is the path to the terminal component. But if you are not logged in you will see the login screen instead. We could use path to any Cockpit component here, I just selected the terminal as we already use it so it should work.

web/src/DevServerWrapper.jsx
lslezak: Um, not really. It just reports the current login status as JSON or returns Error 401 (Unauthorized) if not logged in. It does not display the login form.

web/src/DevServerWrapper.jsx
dgdavid: Ah, sorry. I understand it now. I think I didn't read the word "any" in the comment before. Please, excuse me.

imobachgs (approved):

imobachgs:
Great job! I think it is ready to merge. Thank you! 🎆

PR #441, opened 2023-02-28
Adapt storage web client

joseivanlopez:
## Problem

The D-Bus API of the storage service was modified and extended to support iSCSI. The storage D-Bus client used by the web client needs to be adapted in order to provide all the required features to build an iSCSI UI. 

* Part of https://trello.com/c/3gLX6N94/3285-5-d-installer-ui-to-configure-iscsi
* See https://github.com/yast/d-installer/pull/402

## Solution

Adapt the storage D-Bus client to all the changes in the D-Bus API. Moreover, this PR extends the storage D-Bus API to provide some missing features (e.g., iBFT property). 

## Testing

- Added unit tests
- Tested manually


web/src/client/storage.js
```diff
@@ -25,145 +25,687 @@ import DBusClient from "./dbus";
 import { WithStatus, WithValidation } from "./mixins";
 import cockpit from "../lib/cockpit";
 
-const STORAGE_SERVICE = "org.opensuse.DInstaller.Storage";
-const STORAGE_PATH = "/org/opensuse/DInstaller/Storage1";
-const PROPOSAL_CALCULATOR_IFACE = "org.opensuse.DInstaller.Storage1.Proposal.Calculator";
-const PROPOSAL_IFACE = "org.opensuse.DInstaller.Storage1.Proposal";
+/**
```
joseivanlopez, line 28: Disclaimer: I know this file is growing too much. Probably some of this classes can be extracted to a separate file. For example, I expect `ObjectManagerIface` will be required by others clients, too. For now, I would keep everything in this file. We can extract classes later when it is really needed.

web/src/client/storage.js
```diff
@@ -23,147 +23,453 @@
...
+ *  property2: { t: b, v: false },
+ *  property3: { t: "s", v: undefined }
+ * });
+ * //returns { property1: { t: "s", v: "foo" }, property2: { t: "b", v: false } }
+ *
+ * @param {object} cockpitObject
+ * @returns {object}
  */
-class StorageBaseClient {
+const removeUndefinedCockpitProperties = (cockpitObject) => {
```
imobachgs, line 48: It looks like a generic function that might be useful in other parts. Of course, it can live here by now, but it will be reusable in the future (when we add support for more iSCSI-like APIs). And when we move that function, we could extend it to work on nested objects.

web/src/client/storage.js
```diff
@@ -23,147 +23,453 @@
...
+  const result = {};
+
+  Object.keys(cockpitObject).forEach(key => {
```
imobachgs, line 51: What about `filter`?

```javascript
filtered = Object.entries(cockpitObject).filter(([_, v]) => v !== undefined)
return Object.fromEntries(filtered);
```

imobachgs (approved):

dgdavid (changes requested):
What about changes files?

imobachgs:
> What about changes files?

It is part of the iSCSI support, so I do not think it is needed yet.

dgdavid (approved):
Thanks Iván!

PR #445, opened 2023-03-01
Write the Iguana mountlist

ancorgs:
## Problem

If D-Installer is executed from Iguana, it must create a file for Iguana to know what the resulting root file system is.

The file must be called at `/iguana/mountlist` and contain three columns:

* the device (`/dev/sdX`, `/dev/disk/by-uuid/...`, `uuid=`...)
* the mount point (`/sysroot`)
* the mount options

## Solution

Implement the behavior described above, only if the directory `/iguana` is already there.

This pull request also mitigates a related problem: D-Installer already does many things related to storage at the end of the installation. Likely, all that logic should be moved to YaST (for sure, the FDE part should be there and not in D-Installer itself[1]). But for the time being this pull request extracts that logic to a separate and properly object-oriented `DInstaller::Storage::Finisher`.

[1] https://trello.com/c/5x8G7xj8/89-d-installer-better-support-for-tpm-unlocking-of-fde

## Testing

- Added a new unit test
- Extensively tested manually, including FDE, to ensure no regressions were introduced with `DInstaller::Storage::Finisher`.

### DIFF
--- a/service/lib/dinstaller/storage/finisher.rb
+++ b/service/lib/dinstaller/storage/finisher.rb
@@ -1,0 +1,295 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2023] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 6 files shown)


imobachgs (changes requested):
It looks great but, please, add an entry to the changes file.

ancorgs:
Changelog entry added.

imobachgs (approved):

joseivanlopez (approved):
LGTM

PR #455, opened 2023-03-06
[web] Storage fixes

dgdavid:
## Problem

The web UI crashes when no available devices were found because it ends up trying to access to properties on not yet defined object.

<p align="center">

An screenshot with the 'Uncaught TypeError: device is undefined' error message

</p>

## Solution

To fix the `ProposalSummary` component for not trying to render the device label when there are no available devices ~~plus stop using it for the `ProposalTargetSection`. Instead, the section has been adapted to render a _"Not available devices"_ message or the devices selector directly.~~ an adapt the `ProposalTargetSection` to show a _Not available devices_ message when proceed.

## Testing

- Added a new unit test
- Tested manually

## Notes

* This section will change soon in the context of adding support for iSCSI and other storage technologies.
* Because of the above,  IMHO this changes does not deserve an entry in the changelog.

## Screenshots

| Overview page with no available devices | Storage proposal with no available devices |
|-|-|
| Screen Shot 2023-03-06 at 10 43 16-fullpage | Screen Shot 2023-03-06 at 10 43 08-fullpage |

### DIFF
--- a/web/src/components/storage/ProposalSummary.jsx
+++ b/web/src/components/storage/ProposalSummary.jsx
@@ -24,12 +24,6 @@ import { Text } from "@patternfly/react-core";
 import { Em } from "~/components/core";
 
 export default function ProposalSummary({ proposal }) {
-  const DeviceLabel = ({ device }) => {
-    return (
-      <Em>{device.label}</Em>
-    );
-  };
-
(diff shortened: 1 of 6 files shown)


joseivanlopez (approved):
LGTM

PR #458, opened 2023-03-06
[web] Section component improvements

dgdavid:
## Problem

Prior to using the section header for navigating to a page, we introduced a temporary mechanism for placing an action close to the section title when adding the first storage features. The action was intended to open a dialog with section-related settings/options/forms.

This approach was not suited to our needs because we didn't want multiple "settings" icons on one page. Personally, I thought the user interface was overcrowded and confused without a clear justification.

## Solution

Now that we took a step ~~backward~~ forward to use just a plain old HTML link (can I coin the POHL acronym for that?) we realized that no matter if the section allows the user to navigate deeper or simply displays a dialog: following the same pattern of the heading being a link looks discoverable enough in both cases.

<sub>   ^^^ (thanks @joseivanlopez for being so nit-picking :stuck_out_tongue_winking_eye:)</sub>

Thus, this PR adapts our `Section` component for,

* Navigate to a _path_ or trigger an _openDialog_ function. Naturally, only one action can be done so _openDialog_ is ignored when both params are given.

  <sub>^^^ maybe we could use a single _action_ param and delegates the responsibility to call the _navigate_ function to the component _rendering_ the section. But let's play with the component for a while as it is now and postpone that decision a bit. In any case, we start needing our own _navigate_ helper function.</sub>

* Remove the complicated and no longer needed logic for displaying the settings icon and its tooltip.

* Drop the ability to render a separator. Originally, it served as a visual alignment aid for the mentioned icon and section title, but it has no longer been required. Furthermore, the separator is less helpful now that a section title can be underlined if it is actionable. However, the CSS class has been kept just in case a gradient separator is needed to divide a lengthy section. We can remove it in the future if we no longer need it.

* Remove the _description_ param because is not needed at this moment.

## Testing

- Unit tests adapted accordingly
- Tested manually

## Screenshots

|Before|After|
|:-:|:-:|
|Screen Shot 2023-03-07 at 09 02 17-fullpage |Screen Shot 2023-03-07 at 09 00 05-fullpage|

<sub>Please, ignore the _Result_ section in the screenshots above. They are out of sync because a bit of hacking for taking the screenshots :stuck_out_tongue_closed_eyes: </sub>

### DIFF
--- a/web/src/components/core/Section.jsx
+++ b/web/src/components/core/Section.jsx
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2022] SUSE LLC
+ * Copyright (c) [2022-2023] SUSE LLC
  *
  * All Rights Reserved.
  *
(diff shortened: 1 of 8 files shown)


web/src/components/core/Section.jsx
```diff
@@ -23,148 +23,130 @@
...
+const SectionIcon = ({ name, size = 32 }) => {
   if (!name) return null;
 
   return <Icon name={name} size={size} aria-hidden />;
 };
 
+/**
+ * Internal component for rendering the section title
+ *
+ *  NOTE: a section can do either, navigate to a page following given path or
```
imobachgs, line 48: np: I would drop the "NOTE: " part. It is a comment that makes perfect sense in the description of the component.

web/src/components/core/Section.jsx
```diff
@@ -23,148 +23,130 @@
...
 /**
  *
  * Displays an installation section
  * @component
  *
- * @example <caption>Simple usage</caption>
- *   <Section title="Users" icon={UsersIcon}>
- *     <UserSectionContent />
- *   </Section>
+ *  NOTE: a section can do either, navigate to a page following given path or
```
imobachgs, line 97: DRY :-) If the behavior changes, you need to remember to change both comments.

imobachgs (approved):
Much better now 😍 Thanks!

web/src/components/core/Section.jsx
dgdavid: Sure, and it is there already :)

web/src/components/core/Section.jsx
dgdavid: You're right. And I knew it... but  I was looking for more chances to have everyone warned. Anyway, I'm going to follow your advice and remove the comment from the internal component. Thanks!

dgdavid: >  I'm going to follow your advice

Which was helpful because both comments had leftovers :man_facepalming:

PR #461, opened 2023-03-10
Copy udev rules before calling bootloader finish client

teclator:
## Problem

At the end of the installation there are some important files that must be copied to the target system.

In the traditional YaST installation which relies on the inst_finish client this copy is performed by the copy_files_finish_client

- https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_finish.rb#L320
- https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/copy_files_finish.rb

We might identify which files are still important to be copied but by now at least the udev rules should be transferred to the target system.

## Solution

Copy the udev rules to the target system at the end of the installation and also call the cio_ignore_finish client which could be important in case of LPAR installations.

## Testing

- *Added a new unit test*
- *Tested manually*


service/lib/dinstaller/storage/finisher.rb
```diff
@@ -121,6 +127,45 @@ def staging_graph
...
+          { dir: "/etc/udev/rules.d", file: "41-*" },
+          { dir: "/etc/udev/rules.d", file: "70-persistent-net.rules" }
+        ].freeze
+
+        def label
+          "Copying important installation files to the target system"
+        end
+
+        def run?
+          !glob_files.empty?
```
ancorgs, line 145: why not `glob_files.any?`. This double negation makes my brain spin.

service/lib/dinstaller/storage/finisher.rb
```diff
@@ -36,6 +38,9 @@ class Finisher
       include Helpers
 
       # Constructor
+      # @param logger { Logger }
+      # @param config { Config }
+      # @param security { Security }
```
ancorgs, line 43: We use [] to document the types normally.

ancorgs (approved):
LGTM for the time being since this allow us to move the s390x case forward.

But in the mid-term we need to re-evaluate the actions in the storage finisher.

PR #462, opened 2023-03-11
[web] Sidebar improvements

dgdavid:
Originally created for the iSCSI UI https://github.com/yast/d-installer/pull/435, this set of changes had the sole purpose of allowing us to teleport links from a page to a layout slot created for this use case. However, a few other changes were either necessary for the implementation of the feature or deemed important enough for the next release. Namely

* Taken advantage of the event bubbling for auto-closing the sidebar when an action is clicked instead of passing the "closeCallback" around.

* Dropped the `TargetIpsPopup` action and component since they became a bit useless now we have the addresses summary in the network section at Overview page.

* Adapted the action for navigating to the product selection page:

  - Use a link instead of a button, since it's an element that lets the user navigate

  - It's displayed only on the Overview page

  - Has been moved to the software namespace

* Started using the dark green (aka Pine Green) as the primary color due to the low contrast of the lighter one (aka Jungle Green), which has been relegated to the :hover states (still having the same low contrast, though)

* Kept betting on plain old HTML links. I.e., use the "default" underlined text for actions in the sidebar too (even though some of the actions there are actually buttons, but visually it is accepted to present them as links)

* Use a custom disclosure widget for grouping some actions, like the "Diagnostic tools" ones.

  - The PatternFly ExpandableSection was discarded by now because it's not possible to add the custom `data-keep-sidebar-open` attribute to the toggler.

  - The native details/summary element was discarded because of https://www.scottohara.me/blog/2022/09/12/details-summary.html

* Changed the test-suite a bit to use a MemoryRouter and improve how navigation is mocked. That was a change due to https://github.com/yast/d-installer/pull/462/commits/05604980662e2dac9d76726bd4318ce945ebf0c3, which was later "reverted". Still, changes in test remain because it's cleaner now to deal with routes (see lines dropped at https://github.com/yast/d-installer/pull/462/commits/0413912c7be0a224929f86de6039642d56457b9a)

And maybe some other small stuff not summarized above.

| Before | After (with diagnostic tools collapsed) | After (with diagnostic tools expanded)|
|:-:|:-:|:-:|
| Screen Shot 2023-03-13 at 15 11 42-fullpage | Screen Shot 2023-03-13 at 15 33 33-fullpage |Screen Shot 2023-03-13 at 15 33 29-fullpage |

<sub>Note: after latest changes, _after_ screenshots are a bit out-dated. Now the link is "Change product" instead of "Change selected product"</sub>


web/src/components/software/ChangeProductLink.jsx
```diff
@@ -20,30 +20,19 @@
...
-      variant="link"
-      icon={<Icon name="edit_square" size="24" />}
-      onClick={() => {
-        navigate("/products");
-        onClickCallback();
-      }}
-    >
+    <Link to="/products">
+      <Icon name="edit_square" size="24" />
       Change selected product
```
dgdavid, line 35: I think we could simply use `Change product` instead.

web/package/cockpit-d-installer.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Mon Mar 13 15:51:08 UTC 2023 - David Diaz <dgonzalez@suse.com>
+
+- Sidebar improvements (gh#yast/d-installer#462)
+  * Allow adding actions from a page
+  * Remove network information
+  * Use underlined links and darker green color for improving contrast
+  * Start using a disclosure widget for grouping relate actions
```
imobachgs, line 8: typo: related

Additionally, I would add a full stop after each line.

web/src/components/core/PageOptions.jsx
```diff
@@ -0,0 +1,82 @@
...
+ * React tree only.
+ *
+ * @example <caption>Simple usage</caption>
+ *   <PageOptions title="Storage options">
+ *     <Link to="/storage/iscsi">Configure iSCSI devices</Link>
+ *     <Button
+ *       onClick={showStorageHwInfo}
+ *       data-keep-sidebar-open
+ *     >
+ *       Show Storage Hardaware info
```
imobachgs, line 43: typo: Hardware

web/src/components/software/ChangeProductLink.jsx
imobachgs: Yes, please, change it.

imobachgs (approved):
Awesome work! Just some tiny comments. Once they are addressed, you can merge.

PR #464, opened 2023-03-14
D-Bus API for managing DASD devices - first version

ancorgs:
D-Bus API that allows to:

- Query all the available DASD devices with their corresponding information
- Set the DIAG value for a set of DASD devices
- Activate a set of DASD devices
- Deactivate a set of DASD devices
- Format a set of DASD devices (see below)

Only if D-Installer is running in an s390x system, the Storage1 object will implement the interface `.DInstaller.Storage1.DASD.manager` that is the entry point for all the mentioned operations. Documentation about that new interface and the corresponding ``.DInstaller.Storage1.DASD.device` is included in the pull request.

Although the format operation works, the current implementation is temporary and will be refined in a subsequent pull request. It's a slow operation (potentially several minutes), so [we decided](https://gist.github.com/ancorgs/390b064373995595a1b1dfb08d00507a?permalink_comment_id=4502266#gistcomment-4502266) to create the concept of "processes" or "jobs" and expose them in the D-Bus tree. But it makes no sense to block this pull request meanwhile.

### Testing

- All operations manually tested in a real s390 system.
- Unit tests included.

### DIFF
--- a/doc/dbus_api.md
+++ b/doc/dbus_api.md
@@ -126,10 +126,13 @@ Service for managing storage devices.
   .DInstaller.Storage1
   .DInstaller.Storage1.Proposal.Calculator
   .DInstaller.Storage1.ISCSI.Initiator
+  .DInstaller.Storage1.DASD.Manager (Only available on s390 systems)
 /DInstaller/Storage1/Proposal
   .DInstaller.Storage1.Proposal
 /DInstaller/Storage1/iscsi_nodes/[0-9]+
   .DInstaller.Storage1.ISCSI.Node
+/DInstaller/Storage1/dasds/[0-9]+ (Only available on s390 systems)
+  .DInstaller.Storage1.DASD.device
 ~~~
 
 ### D-Bus Objects
@@ -145,6 +148,7 @@ Service for managing storage devices.
   .DInstaller.Storage1
   .DInstaller.Storage1.Proposal.Calculator
   .DInstaller.Storage1.ISCSI.Initiator
+  .DInstaller.Storage1.DASD.manager
 ~~~
 
 Main object exported by the service `org.opensuse.DInstaller.Service`. This object implements the `org.freedesktop.DBus.ObjectManager` interface and should be used by clients to discover other objects.
@@ -170,6 +174,15 @@ This object is exported only if a proposal was already calculated (successful or
 ~~~
 
 Objects representing iSCSI nodes are dynamically exported when a successful iSCSI discovery is performed, see `.org.opensuse.DInstaller.Storage1.ISCSI.Initiator` interface.
+
+#### `/org/opensuse/DInstaller/Storage1/dasds/[0-9]+` Objects
+
+~~~
+/DInstaller/Storage1/dasds/[0-9]+
+  .DInstaller.Storage1.DASD.Device
+~~~
+
+Objects representing DASDs are dynamically exported when a successful probing is performed by the `DASD.manager` interface of the main storage object, see `.org.opensuse.DInstaller.Storage1.DASD.manager`.
 
 ### D-Bus Interfaces
 
@@ -337,7 +350,7 @@ Delete(in o  iscsi_node_path,
 ##### Properties
 
 ~~~
-IniciatorName readable,writable   s
+InitiatorName readable,writable   s
 ~~~
 
 ##### Details
@@ -351,7 +364,7 @@ Discover(in  s       address,
          out u       result)
 ~~~
 
-Performs nodes discovery. Discovered nodes are exported with the path `/org/opensuse/DInstaller/iscsi/node[0-9]+`.
+Performs nodes discovery. Discovered nodes are exported with the path `/org/opensuse/DInstaller/iscsi_nodes/[0-9]+`.
 
 Arguments:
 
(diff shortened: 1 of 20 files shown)


doc/dbus_api.md
```diff
@@ -435,6 +449,127 @@ Arguments:
 
 * `out u result`: `0` on success and `1` on failure.
 
+#### `org.opensuse.DInstaller.Storage1.DASD.Manager` Interface
+
+Provides methods for configuring DASDs. It's only available if the D-Bus service is running on an
```
jreidinger, line 454: I think here it should be "on **a**"

doc/dbus_api.md
```diff
@@ -435,6 +449,127 @@ Arguments:
...
+~~~
+
+* `Id`: The device channel id (eg. "0.0.0150")
+* `Enabled`: Whether the device is enabled.
+* `DeviceName`: Device name of the DASD in the linux system (eg. "/dev/dasda"). Empty string if the
+  device is not enabled.
+* `Formatted`: whether the device is formatted.
+* `Diag`: Whether the DIAG access method is used (or will be used when the device is enabled).
+* `Type`: The DASD type (eg. EKCD or FBA).
+* `AccessType`: Empty string if unknown. Either "rw" or "ro" otherwise.
```
jreidinger, line 571: for Type and AccessType I am not sure if unsigned flags won't be better as it can provide better mapping and also easier check for known and unknown methods

service/lib/dinstaller/storage/dasd/diag_operation.rb
```diff
@@ -0,0 +1,63 @@
...
+      private
+
+        # @return [Boolean] new wanted value for the use_diag flag
+        attr_reader :value
+
+        def process_dasd(dasd)
+          dasd.diag_wanted = value
+          return true if dasd.offline?
+
+          dasd_configure.disable && dasd_configure.enable
```
ancorgs, line 49: I just found a bug here. I will submit the fix together with the corresponding unit test.

doc/dbus_api.md
```diff
@@ -435,6 +449,127 @@ Arguments:
...
+* `PartitionInfo`: Description of the partitions. Empty if the information is unknown.
```
jreidinger, line 572: Well, if this parameter should be useful it needs more information like it is human readable description useless for machine processing? so something just to display? or can it be parsed? if so, then it needs document its format.

doc/dbus_api.md
```diff
@@ -435,6 +449,127 @@ Arguments:
...
 
 ## Users
```
jreidinger, line 574: :+1: in general good documentation

service/lib/dinstaller/dbus/interfaces/dasd.rb
```diff
@@ -0,0 +1,178 @@
...
+        # returns 1 without trying to process any of the DASDs.
+        #
+        # If all the paths are on the D-Bus tree, it runs the block and returns 0 if it succeeds
+        # or 2 otherwise.
+        #
+        # @return [Integer]
+        def dasds_dbus_method(paths)
+          dbus_dasds = dasds_tree.find_paths(paths)
+          if dbus_dasds.size != paths.size
+            logger.error "Unknown path(s) at #{paths}"
```
jreidinger, line 144: Does it log unknown path somewhere in y2log or not?

service/lib/dinstaller/dbus/interfaces/dasd.rb
```diff
@@ -0,0 +1,178 @@
...
+          end
+
+          dasd_backend.on_refresh do |dasds|
+            dasds_tree.update(dasds)
+          end
+        end
+
+        # Tree with the devices representing the DASDs in the system
+        def dasds_tree
+          @dasds_tree ||= Storage::DasdsTree.new(@service, logger: logger)
```
jreidinger, line 166: Can we have caching issue here when you call new probe and it will find another dasds?

doc/dbus_api.md
ancorgs: Well, I can add "as coming from YaST" :-)

doc/dbus_api.md
ancorgs: I though the same, but for the time being I'm only exposing the values coming from YaST. I don't want to operate too much on them.

service/lib/dinstaller/storage/dasd/diag_operation.rb
```diff
@@ -0,0 +1,58 @@
...
+      class DiagOperation < SequentialOperation
+        # Constructor
+        def initialize(dasds, logger, value)
+          super(dasds, logger)
+          @value = value
+        end
+
+      private
+
+        # @return [Boolean] new wanted value for the use_diag flag
```
jreidinger, line 42: maybe "newly"

doc/dbus_api.md
ancorgs: In fact, if the possible values for "AccessType" are really "rw" or "ro" it could be better turned into a "ReadOnly" boolean attribute.

But, as said, for this first version I wanted to stay close to the YaST values since I don't know all the particularities of every attribute and why they are modeled that way in YaST.

service/lib/dinstaller/dbus/interfaces/dasd.rb
ancorgs: I don't think so. The @dasd_tree object is always the very same. Its content gets updated on every probe and/or refresh.

doc/dbus_api.md
ancorgs: I added a note about the topic.

doc/dbus_api.md
ancorgs: Added an example.

service/lib/dinstaller/dbus/interfaces/dasd.rb
ancorgs: Now (in the newly added commit) it does ;-)

jreidinger (approved):

PR #466, opened 2023-03-15
[web] Change cspell dependency source

dgdavid:
## Problem

@teclator told me that there were problems at the time to build the package in OBS because of adding the `cspell` dependency from Github. This made me realize that we can use it from npm instead.

## Solution

Change the dependency source.


teclator (approved):
LGTM

PR #469, opened 2023-03-16
Service configuration changes for the S390 arch

teclator:
## Problem

Do not restrict ALP product to x86_64 and aarch64 but do with Leap 16


imobachgs (changes requested):
As the image is being developed in OBS, I will remove the `image/` directory. So, please, remove the changes to the kiwi definition and add an entry to the changes file.

imobachgs (approved):

PR #475, opened 2023-03-20
Fix iSCSI discover

joseivanlopez:
## Problem

iSCSI discover form uses local storage, but it fails when threre is nothing stored yet.

## Solution

Initialized local storage with default data.

## Testing

- Tested manually


imobachgs (approved):

imobachgs:
The problem in the integration tests does not seem to be related.

PR #476, opened 2023-03-20
Report progress of DASD formatting to the D-Bus interface

ancorgs:
## Problem

#464 implemented a D-Bus API for managing DASDs, but a way to track the progress of the formatting progress was still missing.

## Solution

This adds the concept of storage jobs, that are exported at `/org/opensuse/DInstaller/Storage1/jobs/[0-9]+`

Every time a formatting operation starts properly, the path of a new job is returned. That job can be used to track the formatting status of all the DASDs involved in the operation.

## Testing

- Added some minimal unit tests
- Manually tested in a real s390 with real formatting (thanks @teclator)

## Documentation

- Expanded `doc/dbus_api.md`

### DIFF
--- a/doc/dbus_api.md
+++ b/doc/dbus_api.md
@@ -133,6 +133,9 @@ Service for managing storage devices.
   .DInstaller.Storage1.ISCSI.Node
 /DInstaller/Storage1/dasds/[0-9]+ (Only available on s390 systems)
   .DInstaller.Storage1.DASD.device
+/DInstaller/Storage1/jobs/[0-9]+
+  .DInstaller.Storage1.Job
+  .DInstaller.Storage1.DASD.Format
 ~~~
 
 ### D-Bus Objects
@@ -183,6 +186,16 @@ Objects representing iSCSI nodes are dynamically exported when a successful iSCS
 ~~~
 
 Objects representing DASDs are dynamically exported when a successful probing is performed by the `DASD.manager` interface of the main storage object, see `.org.opensuse.DInstaller.Storage1.DASD.manager`.
+
+#### `/org/opensuse/DInstaller/Storage1/jobs/[0-9]+` Objects
+
+~~~
+/DInstaller/Storage1/jobs/[0-9]+
+  .DInstaller.Storage1.Job
+  .DInstaller.Storage1.DASD.Format
+~~~
+
+Objects representing long-running processes, like formatting of DASDs.
 
 ### D-Bus Interfaces
 
@@ -484,10 +497,10 @@ Disable(in  ao devices,
 SetDiag(in  ao devices,
         in  b  diag,
         out u  result)
+Format(in  ao devices,
+       out u  result,
+       out o  job)
 ~~~
-
-In addition, a `Format()` method is provided, but it's not documented here because it's going to
-change heavily in the short term.
 
 ##### Details
 
@@ -543,6 +556,23 @@ Arguments:
 * `in b diag`: new value for the flag.
 * `out u result`: `0` if `use_diag` is correctly set for all the requested DASDs. `1` if any of the given paths is invalid (ie. it does not correspond to a known DASD), `2` in case of any other error.
 
+###### `Format` Method
+
+~~~
+Format(in  ao devices,
+       out u  result,
+       out o  job)
+~~~
+
+Starts a format process for the DASDs in the given list. It creates a job to represent such a
+process.
+
+Arguments:
+
+* `in ao devices`: paths of the D-Bus objects representing the DASDs to format.
+* `out u result`: `0` if the format operation starts correctly and the job to track it is created. `1` if any of the given paths is invalid (ie. it does not correspond to a known DASD), `2` in case of any other error.
+* `out o job`: if the result is 0, path of the new job that can be used to track the formatting. Empty string in any other case.
+
 #### `org.opensuse.DInstaller.Storage1.DASD.Device` Interface
 
 This interface is implemented by objects exported at the `/org/opensuse/DInstaller/Storage1/dasds/[0-9]+`
@@ -575,6 +605,43 @@ string `AccessType` could be replaced by a boolean `ReadOnly`).
 * `AccessType`: Empty string if unknown. Either "rw" or "ro" otherwise.
 * `PartitionInfo`: Partition names (and sometimes their type) separated by commas.
   Eg. "_/dev/dasda1 (Linux native), /dev/dasda2 (Linux native)_". Empty if the information is unknown.
+
+#### `org.opensuse.DInstaller.Storage1.Job` Interface
+
+This interface is implemented by objects exported at the `/org/opensuse/DInstaller/Storage1/jobs/[0-9]+`
+paths. It provides information about a long-running process.
+
+##### Properties
+
+~~~
+Running  readable b
+ExitCode readable u
+~~~
+
+* `Running`: Whether the Job is still being executed or it has already finished.
+* `ExitCode`: Final result. Zero for running processes.
+
+##### Signals
+
+* `Finished(u exit_code)`: the Job is not longer running and the exit code has been set to its final value.
+
+#### `org.opensuse.DInstaller.Storage1.DASD.Format` Interface
+
+This interface is implemented by those Job objects used to represent the formatting of a set of DASDs.
+
+##### Properties
+
+~~~
+Summary readable a{s(uub)}
+~~~
+
+* `Summary`: A hash where each key is the path of one of the DASDs being formatted and the value is
+  the status represented by a triplet of total cylinders, cylinders already processed and a boolean
+  indicating whether the process for that particular disk is completed.
+
+##### Signals
+
+* `SummaryUpdated(a{s(uub)} progress)`: the summary has been updated with new progress information for some of the DASDs.
 
 ## Users
 
--- a/doc/dbus_api.md
+++ b/doc/dbus_api.md
@@ -623,7 +623,11 @@ ExitCode readable u
 
 ##### Signals
 
-* `Finished(u exit_code)`: the Job is not longer running and the exit code has been set to its final value.
+* `Finished(u exit_code)`: the Job is not longer running and the exit code has been set to its final
+  value.
+* `PropertiesChanged()`: in parallel to the mentioned `Finished` signal, the standard
+  `PropertiesChanged` signal from `org.freedesktop.DBus.Properties` is also triggered at the end of
+  the job execution to reflect the corresponding changes in the properties.
 
 #### `org.opensuse.DInstaller.Storage1.DASD.Format` Interface
 
--- a/doc/dbus_api.md
+++ b/doc/dbus_api.md
@@ -645,7 +645,7 @@ Summary readable a{s(uub)}
 
 ##### Signals
 
-* `SummaryUpdated(a{s(uub)} progress)`: the summary has been updated with new progress information for some of the DASDs.
+* `PropertiesChanged`, as standard from `org.freedesktop.DBus.Properties`.
 
 ## Users
 
--- a/doc/dbus_api.md
+++ b/doc/dbus_api.md
@@ -571,7 +571,7 @@ Arguments:
 
 * `in ao devices`: paths of the D-Bus objects representing the DASDs to format.
 * `out u result`: `0` if the format operation starts correctly and the job to track it is created. `1` if any of the given paths is invalid (ie. it does not correspond to a known DASD), `2` in case of any other error.
-* `out o job`: if the result is 0, path of the new job that can be used to track the formatting. Empty string in any other case.
+* `out o job`: if the result is 0, path of the new job that can be used to track the formatting. Contains the string `/` (no job) if the result is not zero.
 
 #### `org.opensuse.DInstaller.Storage1.DASD.Device` Interface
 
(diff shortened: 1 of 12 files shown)


mvidner:
>  as far as I know, that integration test is failing for all pull requests
> and I would say it's being ignored (ie. we are merging PRs despite the test)
> it was already broken for all these already merged PRs: 475, 478, 482 and 483

doc/dbus_api.md
```diff
@@ -543,6 +556,23 @@ Arguments:
...
+~~~
+
+Starts a format process for the DASDs in the given list. It creates a job to represent such a
+process.
+
+Arguments:
+
+* `in ao devices`: paths of the D-Bus objects representing the DASDs to format.
+* `out u result`: `0` if the format operation starts correctly and the job to track it is created. `1` if any of the given paths is invalid (ie. it does not correspond to a known DASD), `2` in case of any other error.
+* `out o job`: if the result is 0, path of the new job that can be used to track the formatting. Empty string in any other case.
```
mvidner, line 574: An empty object path is not allowed, and I believe the bus would drop the connection.

The best practice for a `nil` object path is `/` (which assumes the service uses the API guidelines and does not use `/`  for its main object)

doc/dbus_api.md
```diff
@@ -576,6 +606,43 @@ string `AccessType` could be replaced by a boolean `ReadOnly`).
...
+Summary readable a{s(uub)}
+~~~
+
+* `Summary`: A hash where each key is the path of one of the DASDs being formatted and the value is
+  the status represented by a triplet of total cylinders, cylinders already processed and a boolean
+  indicating whether the process for that particular disk is completed.
+
+##### Signals
+
+* `SummaryUpdated(a{s(uub)} progress)`: the summary has been updated with new progress information for some of the DASDs.
```
mvidner, line 644: This should not be necessary, the [standard signal `PropertiesChanged`](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-properties) is designed for it and implemented by ruby-dbus

mvidner (commented):
First review: the API

doc/dbus_api.md
mvidner: Yes, the connection drops: https://github.com/mvidner/ruby-dbus/issues/119

doc/dbus_api.md
ancorgs: Well, it's a bit different here. The property "Summary" is a hash like:

```
{
  /path/to/dasds/1 => [2000, 100, false],
  /path/to/dasds/2 => [2500, 150, false],
  /path/to/dasds/3 => [2500, 200, false],
  /path/to/dasds/4 => [2000, 100, false],
  /path/to/dasds/5 => [3000, 120, false],
}
```

Typically there are more entries (maybe 20?).

Now assume that only the formatting of dasds 3 and 4 have progressed. As far as I understand `PropertiesChanged` the whole hash would be in that signal.

But with the current definition of `SummaryUpdated` only the two entries for the DASDs 3 and 4 are part of the signal. I don't even know if that's important/relevant or whether that small saving pays off. But that's the reason why we decided to have a specific signal instead of using `PropertiesChanged`.

mvidner: See the [detailed docs for `dbus_reader`](https://www.rubydoc.info/gems/ruby-dbus/0.19.0/DBus/Object#dbus_reader-class_method)

doc/dbus_api.md
ancorgs: Cool. I will change that.

doc/dbus_api.md
mvidner: Aah, I missed that subtle detail

mvidner: So
```suggestion
* `SummaryUpdated(a{s(uub)} progress)`: the summary has been updated with new progress information for **some** of the DASDs.
```

doc/dbus_api.md
```diff
@@ -576,6 +606,43 @@ string `AccessType` could be replaced by a boolean `ReadOnly`).
...
+
+##### Properties
+
+~~~
+Summary readable a{s(uub)}
+~~~
+
+* `Summary`: A hash where each key is the path of one of the DASDs being formatted and the value is
+  the status represented by a triplet of total cylinders, cylinders already processed and a boolean
+  indicating whether the process for that particular disk is completed.
```
mvidner, line 640:
```suggestion
  indicating whether the process for that particular disk is completed.
  
  Its PropertiesChanged signal does not include the value, see SummaryUpdated for a partial hash.
```

doc/dbus_api.md
mvidner: Let me recap our options:

1. standard `PropertiesChanged` only: least surprising, but it may repeat too much data
2. `SummaryUpdated`: to be useful we should omit `PropertiesChanged`, or more precisely, omit its *value*, declaring it as `invalidates`.
3. @joseivanlopez proposed a `LastFormattedDevices` property using `PropertiesChanged`. IMHO that is more confusing than (2)

joseivanlopez: Ok, sorry for the confusion on IRC. Now I got the problem after reading the code. So, we have:

* A `Summary` property which does not emit a `PropertiesChanged` signal when it value changes.
* A `SummaryUpdated` signal that is emitted when the `Summary` changes, but it only contains the modified parts of the summary.

Having properties without emitting a `PropertiesChanged` signal sounds a bit weird to me. Some clients like cockpitjs expect it, otherwise the proxy never udpates the value of the property. I would try to follow the same approach as the rest of D-Bus APIs in D-Installer, so I would emit a `PropertiesChanged` signal when Summary changes. And, if we do that, maybe we don't need the `SummaryUpdated` signal anymore. We could include another value for each entry in the summary to indicate if the value is new or not.

service/lib/dinstaller/dbus/storage/dasds_format_job.rb
```diff
@@ -0,0 +1,166 @@
...
+          dbus_reader(:running, "b")
+          dbus_reader(:exit_code, "u")
+          dbus_signal(:Finished, "exit_code:u")
+        end
+
+        DASD_FORMAT_INTERFACE = "org.opensuse.DInstaller.Storage1.DASD.Format"
+        private_constant :DASD_FORMAT_INTERFACE
+
+        dbus_interface DASD_FORMAT_INTERFACE do
+          dbus_reader(:summary, "a{s(uub)}")
```
mvidner, line 91:
```suggestion
          dbus_reader(:summary, "a{s(uub)}", emits_changed_signal: :invalidates)
```

mvidner (commented):
Omitting the PropertiesChanged signal would indeed confuse caching clients like cockpit.dbus, but we'll fix that with [EmitsChangedSignal](https://dbus.freedesktop.org/doc/dbus-specification.html#introspection-format)

doc/dbus_api.md
mvidner: > we should check that cockpit.dbus supports it.

Apparently it does: when it gets a signal that Summary was invalidated, it asks for it with Get.
https://github.com/cockpit-project/cockpit/blob/2a8a630ed0a56dc1c9c4946eae997d902b16383a/src/bridge/cockpitdbuscache.c#L969

service/lib/dinstaller/dbus/interfaces/dasd.rb
```diff
@@ -124,7 +123,17 @@ def dasd_set_diag(paths, diag)
         # @param paths [Array<String>]
         # @return [Integer]
```
mvidner, line 124:
```suggestion
        # @return [Array(Integer,::DBus::ObjectPath)]
```

service/lib/dinstaller/dbus/storage/dasds_format_job.rb
```diff
@@ -0,0 +1,166 @@
...
+
+        # Marks the job as finished
+        #
+        # @note A Finished signal is always emitted.
+        #
+        # @param exit_code [Integer]
+        def finish_format(exit_code)
+          @running = false
+          @exit_code = exit_code
+          Finished(exit_code)
```
mvidner, line 135: D_Bus pedantry not to confuse clients that cache properties:

```suggestion
          Finished(exit_code)
          dbus_properties_changed(JOB_INTERFACE, interfaces_and_properties.slice(JOB_INTERFACE), [])          
```

Here we use [DBus::Object#interfaces_and_properties](https://www.rubydoc.info/gems/ruby-dbus/0.19.0/DBus/Object#interfaces_and_properties-instance_method) combined with [Hash#slice](https://docs.ruby-lang.org/en/2.7.0/Hash.html#method-i-slice), neatly shortening this more obvious alternative:

```rb
dbus_properties_changed(JOB_INTERFACE, {"Running" => @running, "ExitCode" => @exit_code}, [])
```

service/lib/dinstaller/storage/dasd/format_operation.rb
```diff
@@ -27,37 +27,33 @@ module DInstaller
...
       # Operation to format the given set of DASDs
-      #
-      # TODO: we plan to change the approach to DASD formatting, so this class will change
-      # very soon.
       class FormatOperation
         # Constructor
         #
         # @param dasds [Array<Y2S390:Dasd>] devices to format
-        def initialize(dasds)
+        def initialize(dasds, on_progress = [], on_finish = [])
```
mvidner, line 34: YARD docs missing for new params

service/lib/dinstaller/storage/dasd/format_operation.rb
```diff
@@ -27,37 +27,33 @@ module DInstaller
...
-          #       But the code at DasdActions::Format contains a comment stating otherwise
-          return false unless start?
+          return nil unless start?
 
           process.initialize_summary
-          while process.running?
-            process.update_summary
-            # TODO: trigger a callback to notify the process status information
-            sleep(0.2)
+          Thread.new do
```
mvidner, line 52: Oh, a thread.

ruby-dbus is not thread safe if the threads are sharing a connection.

We need to prove that threads will not collide on the bus, maybe add synchronization *somewhere*

service/lib/dinstaller/dbus/interfaces/dasd.rb
```diff
@@ -124,7 +123,17 @@ def dasd_set_diag(paths, diag)
...
         def dasd_format(paths)
-          dasds_dbus_method(paths) { |dasds| dasd_backend.format(dasds) }
+          dasds = find_dasds(paths)
+          return [1, ""] if dasds.nil?
+
+          job = nil
+          progress = proc { |statuses| job.update_format(statuses) }
```
mvidner, line 130: How does this work, are we calling `update_format` on `nil`?

service/lib/dinstaller/dbus/storage/jobs_tree.rb
```diff
@@ -0,0 +1,68 @@
...
+          @service = service
+          @logger = logger
+        end
+
+        # Registers a new job to format a set of DASDs
+        #
+        # @param initial [Array<Y2S390::FormatStatus] initial status information for all the
+        #   involved DASDs
+        # @param dasds_tree [DasdsTree] up-to-date representations of the DASDs in the D-Bus tree
+        def add_dasds_format(initial, dasds_tree)
```
mvidner, line 50:
```suggestion
        # @return [DasdsFormatJob]
        def add_dasds_format(initial, dasds_tree)
```

mvidner (changes requested):
Bad news, this design fails because ruby-dbus is not thread safe

service/lib/dinstaller/dbus/interfaces/dasd.rb
ancorgs: Not really. The callback is only called by the child thread after that thread has slept, so the initialization below has happened "for sure".

service/lib/dinstaller/dbus/interfaces/dasd.rb
```diff
@@ -124,7 +123,17 @@ def dasd_set_diag(paths, diag)
...
+          finish = proc { |result| job.finish_format(result) }
+          initial_statuses = dasd_backend.format(dasds, on_progress: progress, on_finish: finish)
```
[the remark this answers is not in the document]
mvidner: Hmm, the methods in the thread only send signals, meaning that they only write to the D-Bus socket. Maybe that is less prone to races than reading.

Anyway, here is an old branch from 2010 when I got a multithreading patch and failed to make it work https://github.com/mvidner/ruby-dbus/compare/master...multithreading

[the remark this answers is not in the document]
mvidner: https://github.com/mvidner/ruby-dbus/issues/73 is a basic demonstration of the way it fails

service/lib/dinstaller/dbus/interfaces/dasd.rb
ancorgs: Clarified with comments.

service/lib/dinstaller/storage/dasd/format_operation.rb
ancorgs: Added.

doc/dbus_api.md
ancorgs: Changed at b3bab6998866c15dfa5c45, where `SummaryUpdated` is dropped in favor of the more standard `PropertiesChanged`.

mvidner (approved):
LGTM, thanks!

imobachgs (approved):
Assuming that the rebase did not introduce any problem, I am checking the changes entry and approving the PR because it was already reviewed.

PR #477, opened 2023-03-20
Wifi page options

teclator:
In order to unify the interface of the current Network Page with others like Users or Storage we must move the option to connect a WiFi network to the sidebar at least when there is already an active WiFi connection.

There are also some small issues like the alignment and the look of the button which should be fixed to be consistent with the rest of interfaces.

## Testing

- *Adapted unit test*
- *Tested manually*

## Screenshots

Screenshot from 2023-03-20 10-25-58
Screenshot from 2023-03-20 10-26-05
Screenshot from 2023-03-20 10-26-12
Screenshot from 2023-03-20 10-26-37


dgdavid (approved):

imobachgs:
Please, do no forget to update the changes file, even if It is a small changes.

PR #478, opened 2023-03-20
Fix iSCSI client initialization

imobachgs:
## Problem

After building a new ISO, we discovered that the storage section does not work correctly in the `Overview` section. If you reload, everything works. That's what we have found so far:

* The idea to move the setUp of the storage client does not work: many operations fails because the service is busy at the beginning. So you end up with some invalid proxies. That takes us to the next point.
* The [DBusClient#proxy](https://github.com/yast/d-installer/blob/master/web/src/client/dbus.js#L98) method does not handle the failure when trying to get the proxy information.

## (Temporary) Solution

This PR reverts the change to move the initialization to an early stage, postponing it as much as possible (memoizing the D-Bus client in the ISCSIManager seems to work).

## Debugging

Checking the logs, the difference we see is that, in the working case (reloading the page), Cockpit uses introspection to get the information; however, in the failing one, it relies on GetManagedObject. That behavior is documented in [Cockpit's documentation](https://cockpit-project.org/guide/latest/cockpit-dbus).

We need to find out why it behaves differently, but at least we have a workaround for our next prototype.


joseivanlopez (approved):
LGTM

joseivanlopez (approved):

PR #486, opened 2023-03-21
Modify the packages and patterns to support arch conditional

teclator:
## Problem

The current configuration does not have arch conditional for the patterns or packages needed by each product, as well as we do not have a general support for it.

## Solution

Extend the current software installation repositories approach supporting the arch conditional also in the software selection

## Testing

- *Added a new unit test*
- *Tested manually*


service/lib/dinstaller/software/manager.rb
```diff
@@ -212,25 +212,27 @@ def import_gpg_keys
...
+      def arch_collection_for(section, key)
+        arch_select(section).map { |r| r.is_a?(Hash) ? r[key] : r }
       end
 
       def selected_base_product
         @config.data["software"]["base_product"]
       end
 
+      def arch_match?(archs)
+        return true if archs.nil?
```
imobachgs, line 229: `nil?` or `empty?`?

imobachgs: Just forget about it. It is fine as it is.

imobachgs (approved):

service/package/rubygem-d-installer.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Mar 21 21:51:48 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Allow to define architecture specific patterns and packages
+  (gh#yast/d-installer#86).
```
dgdavid, line 5: #86 or #486 ?

dgdavid (changes requested):

service/package/rubygem-d-installer.changes
imobachgs: Oh, you are right, the number is wrong.

imobachgs (approved):

PR #491, opened 2023-03-22
Bus startup race

mvidner:
## Problem

In an autoinstallation script (TODO link), you start the service using `systemctl`, then try to use it but it fails even though systemd says it went fine. And it eventually (within 1 second) works but we were too fast.

## Solution

Fix the systemd service declaration so that systemd knows when we are ready and waits before proceeding:

[`Type=forking`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=)

## Testing

- Tested manually with
```sh
systemctl stop d-installer.service
time systemctl start d-installer.service; dinstaller config show
```

## Screenshots

no

### DIFF
--- a/service/Gemfile.lock
+++ b/service/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    d-installer (0.8)
+    d-installer (0.8.1)
       cfa (~> 1.0.2)
       cfa_grub2 (~> 2.0.0)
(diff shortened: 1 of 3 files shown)


mvidner:
The CI failure is a known failure (TODO link?) in the integration tests: 
> ✘  3 [chromium] › root_password.spec.ts:8:3 › The user section › can set the root password (30s)

jreidinger (approved):

PR #495, opened 2023-03-23
Autoinstallation

jreidinger:
## Problem

DInstaller missing auto installation.

## Solution

Implement it. ( now it is just dump of used files, will be proper package in future ).

## Testing

- *Tested manually*


imobachgs (approved):
LGTM

PR #504, opened 2023-03-27
Updated screenshots, automate the process using Playwright

lslezak:
## Problem

- The screenshots displayed in the main `README.md` are outdated
- You have to take the screenshots manually :worried: 

## Solution

- Let's write a Playwright test which can automate the process :+1: 

## Testing

- Added a new integration test
- Tested manually, works fine with the D-Installer running from the Live ISO, it also works with the webpack development server configured as a proxy to the same machine

## Notes

- Initially I placed the test into a separate subdirectory so it is not started with the other tests. That's because it really starts the installation (that's a pretty dangerous operation) and also it cannot be used with other tests (once you start the installation you cannot change anything).
- But not starting this test regularly would mean it could easily stop working properly and it could be difficult to make it working again. So I made the test more flexible, it can skip some optional steps and it starts the real installation only when `RUN_INSTALLATION=1` is set. So we can at least partly run the test regularly, even in GitHub Actions.
- Because there is some conflict with the root password test (`playwright/tests/root_password.spec.ts` file) I changed the Playwright configuration to run the tests sequentially using only one worker. Currently there are just few tests so there is almost no difference. Later, when we have much more tests, we could switch it to some sequential-parallel mix using the [test lists files](https://playwright.dev/docs/test-parallel#use-a-test-list-file).
- I changed the screenshots resolution to 768x1024 to have some commonly used resolution
- It takes several screenshots during the package installation, pick the one you like most :smiley:


playwright/README.md
```diff
@@ -39,6 +39,22 @@ To run just a specific test:
 npx playwright test tests/root_password.spec.ts
 ```
 
+## Updating the Screenshots
+
+There is one test specially designed for refreshing the screenshots displayed
+in the main `README.md` file.
+
+To fully run the installation run this:
```
imobachgs, line 47: np: I would replace "run" with "type" (or something similar) to not repeat the term "run".

imobachgs (approved):
LGTM!

imobachgs (approved):

PR #505, opened 2023-03-28
Improved Playwright test

lslezak:
## Problem

The Playwright screenshoting test is generic and handles some optional steps. For example the [product selection dialog](https://github.com/yast/d-installer/blob/master/doc/images/screenshots/product-selection.png) is skipped when there is only one product defined.

But the current implementation relies on timeout when the dialog is not displayed, that makes the test run unnecessary longer.

## Solution

- Avoid timeouts by running waits in parallel using [Promise.any](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any) - it returns the first `Promise` which succeeds, we can check for multiple objects on the page in parallel
- Also avoids handling the timeout errors, makes the code cleaner

## Testing

- Tested manually, taking screenshots still works :smile:

### DIFF
--- a/playwright/tests/take_screenshots.spec.ts
+++ b/playwright/tests/take_screenshots.spec.ts
@@ -25,26 +25,32 @@ test.describe("The Installer", () => {
     // set screenshot size to 768x1024
     page.setViewportSize({ width: 768, height: 1024 });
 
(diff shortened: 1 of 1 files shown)


joseivanlopez (approved):
LGTM

PR #507, opened 2023-03-28
Rename D-Installer to Agama

imobachgs:
D-Installer is now Agama. So let's try to rename everything to the new name.


service/lib/agama/dbus/clients/manager.rb
```diff
@@ -54,7 +54,7 @@ def commit
         end
 
         def current_installation_phase
-          dbus_phase = dbus_object["org.opensuse.DInstaller.Manager1"]["CurrentInstallationPhase"]
+          dbus_phase = dbus_object["org.opensuse.Agama1.Manager"]["CurrentInstallationPhase"]
```
mvidner, line 57: watch out, the autorename did not work right here

it should be `Agama.Manager1` for consistency with the other interfaces

mvidner (changes requested):

doc/dbus_api.md
```diff
@@ -111,95 +111,95 @@ Iface: o.o.YaST.Installer1.Software
...
-/DInstaller/Storage1/iscsi_nodes/[0-9]+
-  .DInstaller.Storage1.ISCSI.Node
-/DInstaller/Storage1/dasds/[0-9]+ (Only available on s390 systems)
-  .DInstaller.Storage1.DASD.device
-/DInstaller/Storage1/jobs/[0-9]+
-  .DInstaller.Storage1.Job
-  .DInstaller.Storage1.DASD.Format
+  .Agama1.ServiceStatus
+  .Agama1.Progress
+  .Agama1.Validation
```
mvidner, line 125: also these 3, I believe we want `.Agama.Foo1`

service/lib/agama/dbus/manager.rb
```diff
@@ -138,7 +138,7 @@ def safe_run(&block)
       end
 
       def busy_error
-        ::DBus.error("org.opensuse.DInstaller.Error.Busy")
+        ::DBus.error("org.opensuse.Agama1.Error.Busy")
```
mvidner, line 141:
```suggestion
        ::DBus.error("org.opensuse.Agama.Error1.Busy")
```

doc/dbus_api.md
imobachgs: Not really, the idea is to have the version number in the service name.

service/lib/agama/dbus/manager.rb
imobachgs: The same that the rest.

Issue #517, opened 2023-03-31
NetworkManager is returning a wrong nameserver (wrong byte ordering or integer) on s390x


teclator:
When trying to solve the previous issue #516 I saved the network configuration as it was and realized the nameserver was wrong. 

Currently we are using the ipv4.dns value which is deprecated instead of the ipv4.dns-data one.

See https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/63eaf168d1b8a8cbac4fbeb1d6806cebccbabdb2

and also related https://bugzilla.redhat.com/show_bug.cgi?id=1153559

I also checked that the same happen with Cockpit.

![Screenshot from 2023-03-31 10-44-54](https://user-images.githubusercontent.com/7056681/229168552-f9a5cd1b-a0fb-48a0-9108-4d6cea141cad.png)
![Screenshot from 2023-03-31 10-44-27](https://user-images.githubusercontent.com/7056681/229168561-3b8ef1e2-19e2-41d8-983f-348dc5913890.png)

teclator:
Already fixed by #518

PR #518, opened 2023-04-03
Fix dns handling as the deprecated ipv4.dns options has precedence over ipv4["dns-data"] when given

teclator:
It fixes issue #517 

As described in the issue, the **GetSettings** method is returning a wrong **ipv4.dns** value caused by an endianess problem therefore we should rely on the **ipv4["dns-data"]** as we do with the addresses and gateway. It is not enough to just get and set the **ipv4["dns-data"]** because when both options are given it looks like ipv4.dns still has precedence which looks like a bug in NetworkManager DBUS API.

```bash
busctl call org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/Settings/1 org.freedesktop.NetworkManager.Settings.Connection GetSettings

#returns something like
... "dns" au 1 3300888568 "dns-data" as 1 "10.160.0.1" ...
```

From https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/63eaf168d1b8a8cbac4fbeb1d6806cebccbabdb2

> Btw, in particular "au" is bad, because we put there a big-endian
> number. There is no D-Bus type to represent big endian numbers, so "u"
> is bad because it can cause endianess problem when trying to remote
> the D-Bus communication to another host (without explicitly
> understanding which "u" properties need to swap for endinness).
> 

- https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html

### DIFF
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Mon Apr  3 08:13:55 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
 
-- Fix dns handling (gh#openSUSE/agama#518).
+- Fix DNS handling by using ipv4.dns-data instead of ipv4.dns (gh#openSUSE/agama#518).
 
 -------------------------------------------------------------------
 Wed Mar 29 11:32:13 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
LGTM

mvidner (changes requested):
making a little more obvious that we're not using the bad data

mvidner (approved):

PR #521, opened 2023-04-03
Storage proposal volumes

joseivanlopez:
## Problem

Currently, the proposal UI only offers a few options. That UI has been [recently revisited](https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md) in order to offer a more prowerful proposal. We need to move forward to the direction pointed in that document.    

## Solution

This is the first iteration to implement the new proposal UI. The main changes introduced with this PR are:

* The proposal UI contains only two sections (Settings and Planned actions).
* The settings section provides the widgets directly, without opening a popup.
* The settings section shows a table with the file systems to create, and allows to add a new file system according to the volumes defined in the product settings.

## Testing

* Added new unit tests.
* Tested manually.

## Screenshots

<details>
  <summary>Show/hide</summary>

Screenshot from 2023-04-13 15-52-59

Screenshot from 2023-04-13 15-53-06

Screenshot from 2023-04-13 15-53-17

</details>


web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -19,58 +19,289 @@
...
+          id="encryptionPasswordForm"
+          password={password}
+          onSubmit={acceptForm}
+          onValidate={validateForm}
         />
         <Popup.Actions>
-          <Popup.Confirm form="settings-form" type="submit" isDisabled={!isValid}>Accept</Popup.Confirm>
-          <Popup.Cancel onClick={() => setIsOpen(false)} autoFocus />
+          <Popup.Confirm form="encryptionPasswordForm" type="submit" isDisabled={!isFormValid}>Accept</Popup.Confirm>
+          <Popup.Cancel onClick={cancelForm} autoFocus />
```
dgdavid, line 237: Here is the problem with the autoFocus. We are forcing the dialog to put the focus in the Cancel action. Probably a copy/paste issue :)

dgdavid: So, we should have a look to all Popup.Actions

web/src/components/core/RowActions.jsx
```diff
@@ -0,0 +1,77 @@
...
+    >
+      <Icon name="more_vert" size="24" />
+    </DropdownToggle>
+  );
+
+  return (
+    <ActionsColumn
+      items={actions}
+      actionsToggle={actionsToggle}
+      {...rest}
```
dgdavid, line 74: NP: Despite what we talked this afternoon, I see now that we should extract the `aria-label` attribute from the `rest`  props to avoid having such attribute in both, the `DropdownToggle`and the `ActionsColumn`. Note that the latest one is "inheriting" all the `rest` props.

So, I'd change this by

```suggestion
export default function RowActions({ id, actions, "aria-label": toggleAriaLabel, ...rest }) {
  const actionsToggle = (props) => (
    <DropdownToggle
      id={id}
      aria-label={toggleAriaLabel || "Actions"}
      toggleIndicator={null}
      isDisabled={props.isDisabled}
      onToggle={props.onToggle}
    >
      <Icon name="more_vert" size="24" />
    </DropdownToggle>
  );

  return (
    <ActionsColumn
      items={actions}
      actionsToggle={actionsToggle}
      {...rest}
```

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -0,0 +1,332 @@
...
+
+const VolumesTable = ({ volumes, onVolumesChange, isLoading }) => {
+  const columns = {
+    mountPoint: "At",
+    details: "Details",
+    size: "Size limits",
+    actions: "Actions"
+  };
+
+  const VolumeRow = ({ id, volume, isLoading }) => {
```
imobachgs, line 191: If those components are getting all that they need via args, I would move them out of the `VolumesTable`. For me, it is fine for pretty small components or helper functions. But for this kind of stuff, I am not that sure that they should be here.

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -0,0 +1,332 @@
...
+          value={sizeText(volume.maxSize)}
+          label="Maximum Size"
+          isDisabled
+        />
+      </FormGroup>
+    </Form>
+  );
+};
+
+const GeneralActions = ({ volumes, templates, onVolumesChange }) => {
```
imobachgs, line 117: np: I am not sure whether it is a good idea to get the list of volumes. After all, you are only interested in the list to append a new element (or to clear the list). Should be this component responsible for that?

imobachgs (changes requested):
Great job! In general, the PR looks good. However, is a little short on documentation. Remember that we want to be able to check the types (in the future) using jsdoc and TypeScript, but if we do not annotate the types, that's impossible.

web/src/components/storage/ProposalActionsSection.jsx
```diff
@@ -19,15 +19,93 @@
...
+
+  const generalActions = actions.filter(a => !a.subvol);
+  const subvolActions = actions.filter(a => a.subvol);
+  const userAction = isExpanded ? "Hide" : "Show";
+  const toggleText = `${userAction} ${subvolActions.length} subvolumes actions`;
+
+  return (
+    <>
+      <Text>
+        Actions to perform for creating the file systems and for ensuring the system boots.
```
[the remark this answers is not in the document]
joseivanlopez: Note that this sentence was already there. It was agreed sometime ago, but your proposal is shorter and more direct, I would say.

web/src/components/storage/ProposalPage.jsx
```diff
@@ -88,16 +105,24 @@ export default function ProposalPage() {
...
+    const newSettings = { ...state.settings, ...settings };
+
+    dispatch({ type: "UPDATE_SETTINGS", payload: { settings: newSettings } });
+
+    loadProposal({ before: () => calculateProposal(newSettings) }).catch(console.error);
   };
 
   const PageContent = () => {
-    if (state.loading || state.proposal?.result === undefined) return <SectionSkeleton lines={3} />;
+    const validVolumeTemplates = () => {
```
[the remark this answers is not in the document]
joseivanlopez: Maybe the "valid" adjective is confusing here. Basically, that method filters out the templates for the file systems already proposed.

web/src/components/storage/ProposalVolumes.jsx
joseivanlopez: Good point.

web/src/components/storage/ProposalVolumes.jsx
joseivanlopez: Yes, I am not sure neither. In general, I have only extracted the subcomponents of the main component. But, for a sucomponent itself, I have defined all its subcomponents inside. We need to agree the pattern to follow. So, let's continue with the approach suggested by you: "only use inner components for very small things".

web/src/components/storage/ProposalPage.jsx
```diff
@@ -88,16 +105,24 @@ export default function ProposalPage() {
...
+    const calculateProposal = async (settings) => {
+      await client.proposal.calculate(settings);
     };
 
-    loadProposal({ before: calculate }).catch(console.error);
+    const newSettings = { ...state.settings, ...settings };
+
+    dispatch({ type: "UPDATE_SETTINGS", payload: { settings: newSettings } });
+
+    loadProposal({ before: () => calculateProposal(newSettings) }).catch(console.error);
```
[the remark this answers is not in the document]
joseivanlopez: Yes, I was not happy with that implementation. I am going to change it a bit.

web/src/components/overview/StorageSection.jsx
```diff
@@ -20,11 +20,33 @@
...
 import { BUSY } from "~/client/status";
-import { ProgressText, Section } from "~/components/core";
-import { ProposalSummary } from "~/components/storage";
+import { Em, ProgressText, Section } from "~/components/core";
+
+const ProposalSummary = ({ proposal }) => {
+  const { availableDevices = [], result } = proposal;
+
+  // When there are no availableDevices the proposal does not make sense.
+  // Returning nothing because a parent component should be displaying the proper error message to the user.
```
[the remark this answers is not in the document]
joseivanlopez: After revisiting it (note that this code was simply move from a deleted file) it turns out that the check is not needed at all. Goot catch!

web/src/client/storage.js
```diff
@@ -72,30 +72,62 @@ class ProposalManager {
...
    *
    * @typedef {object} ProposalData
    * @property {AvailableDevice[]} availableDevices
+   * @property {Volume[]} volumeTemplates
    * @property {Result} result
    */
   async getData() {
     const availableDevices = await this.getAvailableDevices();
+    const volumeTemplates = await this.getVolumeTemplates();
     const result = await this.getResult();
```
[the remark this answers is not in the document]
joseivanlopez: At least they are under the proposal attribute: `client.proposal.getData()`.

imobachgs (approved):

PR #529, opened 2023-04-11
Legacy-free Locale Service, D-Bus API proposal

mvidner:
## Problem

> D-Installer currently have a separate Language service, although it's rather
> simplistic. It just allows to set the language of the installed system using
> Yast::Language.Set. And it's quite memory demanding for such an unimpressive
> task.

> That service would be a nice candidate to be rewritten from scratch with no
> dependencies on YaST or Ruby. It's small enough and could give us a good
> overview on how much can we save.

- https://trello.com/c/MCAMvPMZ/3317-3-agama-design-the-service-to-setup-locale-language-keyboard-timezone-etc

## Solution

Design document for the new service API

## Testing

- ?

## Screenshots

- ?


doc/locale_api.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+#### Design of Elementary Layer (lower)
+
+- Just use the systemd API, don't add any API of our own
+
+#### Design of Proposal Layer (uppper)
+
+- when setting the locale, adjust the proposed package selection and keyboard
+  accordingly. And timezone.
```
jreidinger, line 48: important information here is what happens if user already changed keyboard. And another question is if we allow installation of multi localized environment. As current yast adds single locale, but I think in software selection you can add more. So if user of agama can say I want locale cs_CZ, but install also de_DE localization so user can select itself which to use. And if we decide to not support it, then we should document it with reason.

doc/locale_api.md
```diff
@@ -0,0 +1,97 @@
...
+Just use the systemd API, don't add any API of our own. We will use
+- .ListTimezones
+- .SetLocalRTC
+- .SetTimezone
+- .LocalRTC
+- .Timezone
+
+#### Design of Proposal Layer (uppper)
+
+None needed?
```
jreidinger, line 97: well, if we allow change of RTC that we should get here proposal, not?

doc/locale_api.md
```diff
@@ -0,0 +1,97 @@
...
+.TimeUSec                 property  t    1681214874046139 -
+.Timezone                 property  s    "Europe/Prague"  emits-change
+```
+
+"LocalRTC" means "is the local time zone used for the real time clock",
+so it's !hwclock_in_UTC
+
+#### Design of Elementary Layer (lower)
+
+Just use the systemd API, don't add any API of our own. We will use
```
jreidinger, line 88: just to make it clear, by using systemd API you mean ahving own service, but use org.freedesktop.timedate1 interface, right?

mvidner: Well, I wanted to omit it completely and use systemd directly, but maybe it is worth it to add a forwarder because it uses a different bus (Agama vs system)

Agama.Locale1 service -> Agama/Locale1 object -> freedesktop.timedate1 interface, forwards to freedesktop/timedate1

jreidinger: Well, systemd cannot be used directly as we do not want to modify insts-sys ( or live system ) and only target ( aka installed ) one.

doc/locale_api.md
mvidner: I think you are confusing the caller and callee of the proposal?

Agama.locale.SetLocale takes a locale and proposes software, keyboard, timezone

Agama.locale.SetLocalRTC does nothing on top of systemd. Something else calls it depending on what disks are there. It is not needed.

doc/locale_api.md
mvidner: **1.** you mean, first I change the keyboard from US to German, and then the language from English to Czech?

Should we override the user's choice of keyboard? Let's see the aspects of the choice:
1. does it waste a lot of user's work (for example, the whole partitioning layout should not be reproposed automatically)
2. is the change done secretly (even a trivial change should not be done if the GUI has no way of telling the user; for CLI it's fine as long as clearly documented)
3. is it reasonably easy in code and API?

I'm designing a better-detailed proposal layer with these considerations

mvidner: OK, addressed all of this in a recent commit

doc/locale_api.md
```diff
@@ -42,20 +42,123 @@ org.freedesktop.locale1 interface -     -
...
+(The other systemd keyboard settings are X11Model and X11Options, we don't
+have UI or data for that)
+
+```
+# this is gdbus syntax BTW
+node ...Agama/Locale1 {
+  interface ...Agama.Locale1 {
+    methods:
+      # FIXME: the method part is unclear to me. anyone better at proposals?
+      Commit();
```
jreidinger, line 82: I think we also need proposal method to 1. do initial proposal for given product 2. allow to reset values.
Commit method is fine for me.

doc/locale_api.md
```diff
@@ -42,20 +42,123 @@ org.freedesktop.locale1 interface -     -
...
+>  Locale: German, Keyboard: Czech (qwerty)
+</details>
+
+##### Simple Design: Always Repropose
+
+We can easily afford throwing away the user's choice of keyboard layout and
+simply set what we consider a good default for a newly set locale, because:
+
+1. it is just one setting (as opposed to whole partitioning layout)
+2. the change will be visible in the UI, I assume
```
jreidinger, line 125: not so true for CLI

doc/locale_api.md
```diff
@@ -42,20 +42,123 @@ org.freedesktop.locale1 interface -     -
...
+      readwrite (yas)   Locale = (23, ['cs_CZ.UTF-8', 'de_DE.UTF-8']);
+      readonly  (y(ss)) X11Keyboard = (42, ('cz','qwerty));
+  };
+};
+```
+
+You may know a [similar settings in libzypp][resstatus] where it has 4 levels.
+
+*: maybe this is a crazy optimization? I am not too opposed to use strings for
+this on the bus.
```
jreidinger, line 158: or maybe other way. Just use priority like it is on libzypp repos? Another thing to consider here is that beside user/machine it can be also controlled by product, so e.g. some locale is not available on ALP Micro ( now only en is there ). So it can remove invalid value with high priority.

doc/locale_api.md
mvidner: For CLI, I see two UI modes: 
a. if you have a "dumb" CLI, it does not need to give feedback, it does exactly as it is told, except it may throw errors
b. for a "smart" CLI like our proposal logic, it needs to output the resulting state/changes, so that the user does not confuse it with a dumb one

doc/locale_api.md
mvidner: re (2) higher priority than Human?? IMHO a bad idea

doc/locale_api.md
mvidner: `local_rtc: Priority<bool>` should solve this, right?

doc/locale_api.md
mvidner: Ah, good point :thinking:

doc/locale_api.md
mvidner: re (1): **priority** is a confusing concept, but I have found nothing better :frowning_face: 

The basic problem is:
1. in **words**, everyone agrees that when lower priority and higher priority meet, the higher priority wins
2. if you express them in **numbers**
  a. _bigger_ number means bigger=higher priority
  b. lower number means higher priority, because a lower number is higher on the page in an _ordered list_.

Anyway, both my numeric values (23 and 42) and libzypp use the "ordered list" semantic.
So, I don't understand what you mean, actually?

mvidner: man zypper:

>  -p, --priority positive-integer

>    Set priority of the repository. Priority of 1 is the highest, the higher the number the lower the priority. Default priority is 99. Packages from repositories with higher priority will be preferred even in case there is a higher installable version available in the repository with a lower priority.

mvidner: Human always wins. (Haven't you watched Terminator? Haven't you known Asimov's laws since you've been ~~manufactured~~ born??)

Product specific? Then specify how it should work and I will name it.

doc/locale_api_WIP.md
```diff
@@ -0,0 +1,48 @@
+
+Note: ListFoo APIs combine the IDs and the translations. 
+If it's easier to return just the IDs and ask for the translations separately,
+let's use such different API.
+
+### Locales
+
+```
+ListLocales(
+  in  s     display_locale # "en_US.UTF-8"
```
jreidinger, line 10: Do we need this param or should we use what @shundhammer mention on call, so always return label in form english - native e.g. `Czech - Čeština`

doc/locale_api_WIP.md
```diff
@@ -0,0 +1,48 @@
...
+  out a(ss) id_label_pairs
+)
+
+SetTimezone(in s tz_id)
+
+TimezoneForLocale(
+  in  s locale_id
+  out s tz_id
+)
+```
```
jreidinger, line 48: There will be probably one `Commit` method to write all of them down. I also miss here some method to get locale/keyboard/timezone as UI should display what user already selected.

doc/locale_api_WIP.md
mvidner: OK, YaST does it the english+native way already

doc/locale_api_WIP.md
```diff
@@ -0,0 +1,46 @@
+
+Note: ListFoo APIs combine the IDs and the translations. 
+If it's easier to return just the IDs and ask for the translations separately,
+let's use such different API.
+
+### Locales
+
+```
+ListLocales(
+  out a(sss) id_english_native
```
jreidinger, line 10: what langtable has is not locales specific, but language oriented. So here it makes more sense to have something like `out a(sssas) id_english_native_variants` where values will be like `(en,English,English,[en_US.UTF-8,en_GB.UTF-8,en_IN.UTF-8,...]). So in UI user can select language and then variant which is not translated. Similar data we can get for territories where is territory like Zambia and locales and languages speaked there.

doc/locale_api_WIP.md
```diff
@@ -0,0 +1,46 @@
...
+### Keyboards
+
+```
+ListX11Keyboards(
+  out as    ids  # id like "layout(variant)"
+)
+
+SetX11Keyboard(in s kb_id)
+
+X11KeyboardForLocale(
```
jreidinger, line 25: and same here. Keyboard is linked to language or territory, but not to locale. So we do not have such data.

doc/locale_api_WIP.md
```diff
@@ -0,0 +1,46 @@
...
+```
+ListTimezones(
+  in  s     display_locale # "en_US.UTF-8"
+  out a(ss) id_label_pairs
+)
+
+SetTimezone(in s tz_id)
+
+TimezoneForLocale(
+  in  s locale_id
```
jreidinger, line 43: and same here. There are languages or territories for timezone suggestion, but not for locale.

doc/locale_api_WIP.md
jreidinger: and when I try to implement it I also found that for some less common languages there are no translations in langtable itself like `Central Aymara`. I expect it will solve when we also get data for filtering to only supported languages

PR #530, opened 2023-04-11
Fix netmask handling (bsc#1210104)

teclator:
## Problem

Agama does not handle correctly the IP address **netmask** but works fine when a **prefix length** is used.

- https://bugzilla.suse.com/show_bug.cgi?id=1210104

## Solution

When a connection is updated it always expect a prefix length but in the IP Input form it is allowed to use a netmask or a prefix length.

Screenshot from 2023-04-11 14-51-38

Therefore, the code has been adapted converting the **netmask** to a **prefix length** when needed.

netmask

## Testing

- *Added a new unit test*
- *Tested manually*


dgdavid (approved):
LGTM, thanks.

PR #531, opened 2023-04-12
Kill cockpit when we're done

mchf:
## Problem

Iguana wants cockpit-ws killed as an indication that installation is over.

- Everything what was needed on iguana side is here https://github.com/openSUSE/iguana/pull/32 Mostly a changes in workflow.

## Solution

Reboot button becomes Finish button. The button tells agama to commit suicide by ```/usr/bin/agamactl -k```
An update on the iguana side was also needed - a change in workflow file.

## Testing

- Tested manually using some mocked setup
- Tested with iguana containers from BS by the iguana  maintainer

### DIFF
--- a/service/lib/agama/dbus/manager.rb
+++ b/service/lib/agama/dbus/manager.rb
@@ -56,6 +56,9 @@ module Agama
       CONFIG_PHASE = 1
       INSTALL_PHASE = 2
 
+      IGUANA_NOT_USED = 0
+      IGUANA_USED = 1
+
       dbus_interface MANAGER_INTERFACE do
         dbus_method(:Probe, "") { config_phase }
         dbus_method(:Commit, "") { install_phase }
@@ -63,6 +66,7 @@ module Agama
         dbus_method(:CollectLogs, "out tarball_filesystem_path:s, in user:s") { |u| collect_logs(u) }
         dbus_reader :installation_phases, "aa{sv}"
         dbus_reader :current_installation_phase, "u"
+        dbus_reader :iguana_backend, "b"
         dbus_reader :busy_services, "as"
       end
 
@@ -112,6 +116,11 @@ module Agama
         return STARTUP_PHASE if backend.installation_phase.startup?
         return CONFIG_PHASE if backend.installation_phase.config?
         return INSTALL_PHASE if backend.installation_phase.install?
+      end
+
+      # States whether installation runs on iguana
+      def iguana_backend
+        return backend.iguana? ? IGUANA_USED : IGUANA_NOT_USED
       end
 
       # Name of the services that are currently busy
--- a/service/lib/agama/dbus/manager.rb
+++ b/service/lib/agama/dbus/manager.rb
@@ -56,14 +56,15 @@ module Agama
       CONFIG_PHASE = 1
       INSTALL_PHASE = 2
 
-      IGUANA_NOT_USED = 0
-      IGUANA_USED = 1
+      IGUANA_NOT_USED = false
+      IGUANA_USED = true
 
       dbus_interface MANAGER_INTERFACE do
         dbus_method(:Probe, "") { config_phase }
         dbus_method(:Commit, "") { install_phase }
         dbus_method(:CanInstall, "out result:b") { can_install? }
         dbus_method(:CollectLogs, "out tarball_filesystem_path:s, in user:s") { |u| collect_logs(u) }
+        dbus_method(:Finish, "") { finish_phase }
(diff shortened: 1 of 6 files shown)


imobachgs (commented):
What do you think about moving most of this logic to the backend? Instead of running `shutdown -r now` (as we do [in line 124](https://github.com/openSUSE/agama/blob/ccd7142e3e963416c26e6de6aa5c79c5628afef0/web/src/client/manager.js#L124) we could have a `Reboot` D-Bus method (in the Manager D-Bus object) that takes care of doing whatever is needed. So the client does not have to worry about whether we are running on Iguana or not.

Otherwise, we might need to update the CLI to behave in the same way too.

web/src/client/manager.js
```diff
@@ -123,6 +123,20 @@ class ManagerBaseClient {
   rebootSystem() {
     return cockpit.spawn(["/usr/sbin/shutdown", "-r", "now"]);
   }
+
+  /**
+   * Returns whether finishing cockpit succeeded
+   */
+  finishCockpit() {
+    return cockpit.spawn(["systemctl", "stop", "cockpit-wsinstance-http.service"]);
```
aaannz, line 131: Does this really work? Systemd is probably included as container is based on tumbleweed one, but it's not explicitly included. What happens when base container switches to BCI one or ALP based one? Is systemd required in agama containers?

Also cockpit-ws is started directly, not as a service. At least that's how I see it in [here](https://build.opensuse.org/package/view_file/YaST:Head:Containers/agama-web/run_background_agama.sh?expand=1)

imobachgs: Good point. Actually, neither my proposal to move that logic to the service would work (because it runs on a different container).

mchf: discussed off the github. We went with moving to backend, changing command for restarting and some more changes on iguana side (workflow file update)

web/src/components/core/InstallationFinished.jsx
```diff
@@ -34,15 +34,26 @@ import { useInstallerClient } from "~/context/installer";
 
 function InstallationFinished() {
   const client = useInstallerClient();
-  const onRebootAction = () => client.manager.rebootSystem();
+  const [iguana, setIguana] = useState(false);
+  const closingAction = client.manager.finishInstallation();
```
aaannz, line 38: This is throwing error for me:

```Expected `onClick` listener to be a function, instead got a value of `object` type.```

I have none expertise with react, but just by comparing what was before and we have now, should this be:

```suggestion
  const closingAction = () => client.manager.finishInstallation();
```

aaannz (approved):
Confirm that it works with Iguana now.

One thing to address is UX. Now when the installation finish, you click 'Finish' and then in the web you see no change, nothing.
This will work in case of local install, but in remote install there is no change. Not even when iguana stops and system is booted.

service/lib/agama/dbus/manager.rb
```diff
@@ -114,6 +124,11 @@ def current_installation_phase
         return INSTALL_PHASE if backend.installation_phase.install?
       end
 
+      # States whether installation runs on iguana
+      def iguana_backend
+        backend.iguana? ? IGUANA_USED : IGUANA_NOT_USED
```
dgdavid, line 129: NP: I think these constant are not needed

```suggestion
        backend.iguana?
```

mchf: you're right .. it's a leftover from previous implementation

service/lib/agama/dbus/manager.rb
```diff
@@ -56,13 +56,18 @@ def initialize(backend, logger)
       CONFIG_PHASE = 1
       INSTALL_PHASE = 2
 
+      IGUANA_NOT_USED = false
```
imobachgs, line 59: np: Perhaps this is too much... what about using true/false?

web/src/client/manager.test.js
```diff
@@ -36,6 +36,7 @@ const managerProxy = {
   wait: jest.fn(),
   Commit: jest.fn(),
   Probe: jest.fn(),
+  Finish: jest.fn().mockReturnValue(true),
```
imobachgs, line 39: Good catch!

imobachgs (approved):
Just a minor comment. Otherwsie, LGTM.

service/lib/agama/dbus/manager.rb
mchf: removed as underlying ```backend.iguana?``` returns true/false anyway ... this was just leftover of good old mocking times ;-)

imobachgs (approved):

PR #533, opened 2023-04-14
Rust locale

jreidinger:
## Problem

Language dbus backend is quite limited in functionality and still eating and lot of memory.

## Solution

try to write language component in rust to compare it.

1. The D-Bus interface `org.opensuse.Agama.Language1` is replaced by `org.opensuse.Agama.Locale1`, described in [doc/locale_api.md](/openSUSE/agama/blob/1cf6b64e4d4498567905f0b2d202ed3bd4b8c49b/doc/locale_api.md)

2. Two new crates in the `rust/` directory:
- **agama-dbus-server** the common server for Rust-hosted services
- **agama-locale-data** library interfacing with [langtable](https://github.com/mike-fabian/langtable), in particular `python-langtable-data.rpm` which is just the XML part without the Python code

3.
```console
$ cd service; bundle exec bin/agamactl 
ERROR: Using 'agamactl' to start all services no longer works.
NOTE: It had race conditions all along and now there's a Rust service it can't reach too.
NOTE: Use `systemctl start agama.service` instead
NOTE: which is setup by a) RPMs b) ./setup-service.sh
```


rust/agama-locale-data/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub description: String,
+    pub ascii: bool,
+    pub comment: Option<String>,
+    pub languages: RankedLanguages,
+    pub territories: RankedTerritories
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Keyboards {
+    pub keyboard: Vec<Keyboard>
```
imobachgs, line 48: np: If you do not plan to add anything else, you could use the [newtype pattern idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html).

rust/agama-locale-data/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+use serde::Deserialize;
+use std::fs::File;
+use std::io::BufReader;
+use quick_xml::de::Deserializer;
+use flate2::bufread::GzDecoder;
+
+pub fn add(left: usize, right: usize) -> usize {
```
imobachgs, line 7: leftover :-)

imobachgs (approved):

rust/agama-locale-data/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+use flate2::bufread::GzDecoder;
+
+pub mod keyboard;
+pub mod language;
+pub mod territory;
+pub mod timezone_part;
+pub mod ranked;
+
+pub fn get_keyboards() -> keyboard::Keyboards {
+    const FILE_PATH: &str = "/usr/share/langtable/data/keyboards.xml.gz";
```
imobachgs, line 14: I would put all the constants outside of the functions so it is easier to adjust the names. Although I would expect Rust not define the const on every single call.

rust/agama-locale-data/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    let file = File::open(FILE_PATH).expect("Failed to read langtable-data.");
+    let reader = BufReader::new(GzDecoder::new(BufReader::new(&file)));
+    let mut deserializer = Deserializer::from_reader(reader);
+    keyboard::Keyboards::deserialize(&mut deserializer).expect("Failed to deserialize keyboard entry")
+}
+
+pub fn get_languages() -> language::Languages {
+    const FILE_PATH: &str = "/usr/share/langtable/data/languages.xml.gz";
+    let file = File::open(FILE_PATH).expect("Failed to read langtable-data.");
```
imobachgs, line 23: All the functions look similar. I guess we could reduce the repetition by extracting some logic to a separate function.

imobachgs (commented):
I don't know whether we need to split the code into so many modules, but it looks fine, as I do not have enough experience. I would do reduce the amount of repeated code too.

And thanks for writing those unit tests 💯

rust/agama-dbus-server/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn list_x11_keyboards(&self) -> Vec<(String, String)> {
+        let keyboards = agama_locale_data::get_keyboards();
+        return keyboards.keyboard.iter().map(|k| (k.id.clone(), k.description.clone())).collect()
+    }
+
+    fn set_x11_keyboard(&mut self, keyboard: &str) {
+        self.keyboard_id = keyboard.to_string();
+    }
+
+    fn list_timezones(&self, locale: &str) -> Vec<(String, String)> {
```
mvidner, line 32: A bad input makes a thread panic and then the service is running in an unusable state. We should send error replies instead

```console
$ ./target/debug/agama-dbus-server &
$ busctl --user call org.opensuse.Agama.Locale1 /org/opensuse/Agama/Locale1 org.opensuse.Agama.Locale1 ListTimezones s idontknow
thread 'zbus::Connection executor' panicked at 'Wrong language {language}', agama-locale-data/src/localization.rs:12:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Call failed: Connection timed out
```

mvidner:
@jreidinger I have pushed code that enables error reporting via the D-Bus methods

Funnily, it does not help with the error that I originally encountered: ListTimezones("nosuchlanguage") because that one is pretty deep down the call stack and I want to think about the general API there first.

rust/agama-locale-data/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        let result = get_keyboards().unwrap();
+        assert_eq!(result.keyboard.len(), 247);
+        let first = result.keyboard.first().expect("no keyboards");
+        assert_eq!(first.id, "ad")
+    }
+
+    #[test]
+    fn test_get_languages() {
+        let result = get_languages().unwrap();
+        assert_eq!(result.language.len(), 356);
```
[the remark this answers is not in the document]
jreidinger: interesting why you have different data from langtable.

rust/agama-locale-data/src/localization.rs
```diff
@@ -6,11 +6,13 @@ pub struct Localization {
...
 impl Localization {
-    pub fn name_for(&self, language: &str) -> String {
-        self.name.iter()
-            .find(|n| n.language == language)
-            .expect("Wrong language {language}")
-            .value.clone()
+    pub fn name_for(&self, language: &str) -> Option<String> {
+        let entry = self.name.iter()
+            .find(|n| n.language == language);
+        match entry {
```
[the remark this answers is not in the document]
mvidner: Or even better way with [Option's `?`](https://doc.rust-lang.org/nightly/core/option/index.html#the-question-mark-operator-) :smile: 

```rs
self.name.iter()
  .find(|n| n.language == language)?
  .value.clone())
```

(but I haven't tried either, actually)

rust/agama-dbus-server/src/main.rs
```diff
@@ -2,36 +2,46 @@ pub mod error;
...
-            .map(|l| (l.id.clone(), l.names.name_for("en").unwrap_or_default(),
-                l.names.name_for(l.id.as_str()).unwrap_or_default(),
-                l.locales.locale.iter().map(|l| l.id.to_owned()).collect()))
-             // filter out missing translations as language that does not translate itself is very minor
-            .filter(|r| r.1 != "" && r.2 != "")
-            .collect()
+    /// get labels for given locale. The first pair is english language and territory
+    /// and second one is localized one to target language from locale.
+    /// 
+    fn labels_for_locale(&self, locale: &str) -> Result<((String, String), (String, String)), Error> {
```
jreidinger, line 21: Here I actually think if we should not accept list of locales and return list of pair of pairs as method is kind of costy as it needs double XML parsing and I think in UI we want to display all supported locales.

mvidner: From the API alone I don't understand what "double XML parsing" you mean

jreidinger: "double XML parsing" means that I need to parses languages.xml and territories.xml, so quite costy operation.

jreidinger: Regarding your API it has one important issue. rust service does not know list of supported locales and plan is to get it from repository or product..so something for software component and here it will just return labels for it...so example for modified one is:

```
LabelsForLocales("en_US.UTF-8", "cs_CZ.UTF-8", "de_DE.UTF-8")
-> [(("English", "United States"), ("English", "United States")), (("Cestina", "Cesko"), ("Czech", "Czechia")), ...]
```

rust/agama-dbus-server/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+    timezone_id: String
+}
+
+#[dbus_interface(name = "org.opensuse.Agama.Locale1")]
+impl Locale {
+    // Can be `async` as well.
+    /// get labels for given locale. The first pair is english language and territory
+    /// and second one is localized one to target language from locale.
+    /// 
+    fn labels_for_locale(&self, locale: &str) -> Result<((String, String), (String, String)), Error> {
```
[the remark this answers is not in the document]
jreidinger: so far `"en"` only is not working, but I thinking it should work and return empty territory string. I am not sure about specific locale type and also how you would pass it on other side?

rust/agama-dbus-server/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+            .find(|t| t.id == loc_territory).context("territory for passed locale not found")?;
+
+        let default_ret = (language.names.name_for(DEFAULT_LANG).context("missing default translation for language")?,
+            territory.names.name_for(DEFAULT_LANG).context("missing default translation for territory")?);
+        let localized_ret = (language.names.name_for(language.id.as_str()).context("missing default translation for language")?,
+            territory.names.name_for(language.id.as_str()).context("missing default translation for territory")?);
+        Ok((default_ret, localized_ret))        
+    }
+
+    fn set_locale(&mut self, locale: &str) {
```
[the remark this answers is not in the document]
jreidinger: question is where to check for invalid values. Here or during commit? Also I am not sure if we even can check properly all locales, but we can somehow construct list of it from langtable I expect. or from `localectl list-locales` but it can be limited....or third option which is probably best is to have setter for supported locales like `set_locales` that software uses and then we get return labels without input parameter. What do you think?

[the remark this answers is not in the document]
imobachgs: As a D-Bus user, I would like to check the value when it is assigned (the sooner the better). Otherwise, the service might contain some invalid state.

doc/locale_api.md
```diff
@@ -0,0 +1,283 @@
...
+      )
+      # langtable has X11 keyboards and no localization of their labels
+      ListX11Keyboards(
+        out as    ids  # id like "layout(variant)"
+      )
+
+      # ProposeKeyboard(); # not needed? adjusted automatically, same object
+      # Sets Agama/TimeDate1's Timezone (but not LocalRTC, that's for Storage to say?)
+      ProposeTimeDate(); # different object but same service
+      ProposeSoftware(); # different service
```
jreidinger, line 169: if we use properties, it should not need method and instead software should observe property for change, not?

doc/locale_api.md
```diff
@@ -0,0 +1,283 @@
...
+        out a(sss) id_english_native # [('cs_CZ.UTF-8', 'Czech', 'Čeština'),…]
+      )
+      # langtable has X11 keyboards and no localization of their labels
+      ListX11Keyboards(
+        out as    ids  # id like "layout(variant)"
+      )
+
+      # ProposeKeyboard(); # not needed? adjusted automatically, same object
+      # Sets Agama/TimeDate1's Timezone (but not LocalRTC, that's for Storage to say?)
+      ProposeTimeDate(); # different object but same service
```
jreidinger, line 168: I think timezone should be in this object as we will also set it with systemd-firstboot, not?

rust/agama-locale-data/src/localization.rs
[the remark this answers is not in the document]
jreidinger: I like the second, just it does not add Some. See my final code.

rust/agama-dbus-server/src/locale.rs
```diff
@@ -0,0 +1,166 @@
...
+        Ok(ret)
+    }
+
+    #[dbus_interface(property)]
+    fn timezone(&self) -> &str {
+        return &self.timezone_id.as_str();
+    }
+
+    #[dbus_interface(property)]
+    fn set_timezone(&mut self, timezone: &str) -> Result<(), zbus::fdo::Error> { // NOTE: cannot use crate::Error as property expect this one
```
mvidner, line 143: I see, that's fine, it does have a generic [`zbus::fdo::Error::Failed(String)` constructor](https://docs.rs/zbus/3.12.0/zbus/fdo/enum.Error.html#variant.Failed)

jreidinger: OK, so I will add some validation code with it, so we can test it

imobachgs: You can convert from one type into the other if needed. For instance,

```rust
impl From<MyErrorType> for zbus::fdo::Error {
    fn from(value: MyErrorType) -> zbus::fdo::Error {
        zbus::fdo::Error::Failed(format!("Some error: {}", value.to_string()))
    }
}
```

So when using a `?` it gets automatically converted.

mvidner:
@jreidinger I plan to add the `Priority<T>` wrappers to the API

Also, what about the RTC clock setting? If we add it in this PR, I would also split off the `.../TimeDate1` object

rust/agama-dbus-server/src/locale.rs
```diff
@@ -176,15 +176,26 @@ impl Locale {
...
         const SERVICE_NAME: &str = "org.opensuse.Agama.Locale1";
+        const SERVICE_PATH: &str = "/org/opensuse/Agama/Locale1";
 
-        let locale = Locale::new();
-        const ADDRESS : &str = "unix:path=/run/agama/bus";
-        let conn = ConnectionBuilder::address(ADDRESS)? //TODO: use agama bus instead of session one
-            .name(SERVICE_NAME)?
-            .serve_at("/org/opensuse/Agama/Locale1", locale)?
+        // First connect to the Agama bus, then serve our API,
+        // for better error reporting.
```
jreidinger, line 184: :+1: nice change

mvidner:
Testing this in a container, the web UI does not display contents for **Localization** because the service returns a D-Bus error: 
> Object { problem: null, name: "org.opensuse.Agama.Locale1.Anyhow", message: "Failed to read langtable-data (/usr/share/langtable/data/languages.xml.gz): No such file or directory (os error 2)", toString: toString() }

so I need to make sure python-langtable-data.rpm gets installed (currently in devel:languages:python only)

mvidner:
CI Failures:
- integration-tests: Playwright... ?? Do I need to run it locally to figure it out
- cli_build: (that's Rust build) python-langtable-data is missing; the job runs cargo on ububtu-latest

mvidner:
@jreidinger I made CI green again, yay :)

rust/README.md
```diff
@@ -1,8 +1,9 @@
-# Agama Command Line Interface
+# Agama Command Line and DBus Interface
 
 This project aims to build a command-line interface for
 [Agama](https://github.com/yast/agama), a service-based Linux installer featuring a nice
-web interface.
+web interface. The second aim is dbus service that does not depend heavily on YaST to
```
imobachgs, line 5:
```suggestion
web interface. The second aim is D-Bus service that does not depend heavily on YaST to
```

rust/README.md
```diff
@@ -102,3 +111,15 @@ frontend](./agama-cli/doc/backend-for-testing.md)*.
...
+
+## Packaging
+
+A packaging files lives in `package` directory. Agama follows
+[rust packaging guidelines](https://en.opensuse.org/openSUSE:Packaging_Rust_Software).
+For testing changes to spec file use simple `osc branch YaST:Head:Agama agama-cli`
+and copy modified spec file to that branch.
+If it needs also specific code from git branch, then modify `_service` file and
+change git branch name in `<revision>` tag. Then run `osc service runall`.
+Note: for leap `cargo_audit` does not work with older python, so comment out that
```
imobachgs, line 123:
```suggestion
Note: for openSUSE Leap `cargo_audit` does not work with older Python, so comment out that
```

imobachgs (commented):
The approach looks good. I have some comments, especially about the internal APIs. Nothing to block this PR, though.

rust/README.md
mvidner: Oh, _this_ is the part that I hit yesterday :facepalm: . I should read READMEs more often.

rust/agama-dbus-server/src/locale.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// Note: check how often it is used and if often, it can be easily cached
+    fn labels_for_locales(&self) -> Result<Vec<((String, String), (String, String))>, Error> {
+        const DEFAULT_LANG: &str = "en";
+        let mut res = Vec::with_capacity(self.supported_locales.len());
+        for locale in self.supported_locales.as_slice() {
+            let (loc_language, loc_territory) = agama_locale_data::parse_locale(locale.as_str())?;
+            let languages = agama_locale_data::get_languages()?;
+            let territories = agama_locale_data::get_territories()?;
+            let language = languages
```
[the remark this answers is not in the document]
jreidinger: Agreed, I think it makes sense. I just want to have something working quickly and then abstract usable parts to those structs. Parts that are really used.

rust/agama-dbus-server/src/main.rs
```diff
@@ -0,0 +1,15 @@
+pub mod error;
+pub mod locale;
+
+use std::future::pending;
+
+// Although we use `async-std` here, you can use any async runtime of choice.
```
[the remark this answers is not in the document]
jreidinger: yeap, I use exactly this example as starting point :)

rust/agama-locale-data/src/language.rs
```diff
@@ -0,0 +1,17 @@
...
+pub struct Language {
+    #[serde(rename(deserialize = "languageId"))]
+    pub id: String,
+    pub territories: RankedTerritories,
+    pub locales: RankedLocales,
+    pub names: crate::localization::Localization
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Languages {
```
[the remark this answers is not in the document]
jreidinger: will newtype pattern work with Deserialize Trait?

rust/agama-dbus-server/src/locale.rs
```diff
@@ -0,0 +1,199 @@
...
+        Ok(res)
+    }
+
+    #[dbus_interface(property)]
+    fn locales(&self) -> Vec<String> {
+        return self.locales.to_owned();
+    }
+
+    #[dbus_interface(property)]
+    fn set_locales(&mut self, locales: Vec<String>) -> Result<(), zbus::fdo::Error> {
```
imobachgs, line 65: Oh, I overlooked this one during the review. If you prefer, this is equivalent to `zbus::fdo::Result<()>`.

imobachgs (approved):

mvidner (approved):
Yay, thanks, LGTM now :smile:

PR #540, opened 2023-04-19
Add storage issues

joseivanlopez:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/service/lib/agama/dbus/storage/dasd_manager_interface.rb
+++ b/service/lib/agama/dbus/storage/dasd_manager_interface.rb
@@ -23,7 +23,7 @@ require "dbus"
 
 module Agama
   module DBus
-    module Interfaces
+    module Storage
       # Mixin to define the D-Bus interface to manage DASD devices
       #
       # @note This mixin is expected to be included in a class inherited from
@@ -31,7 +31,7 @@ module Agama
       #
       # @note This mixin is expected to be included only if the namespace Y2S390 (which
       # traditionally lives in the yast2-s390 package) is available.
-      module Dasd
+      module DasdManagerInterface
         DASD_MANAGER_INTERFACE = "org.opensuse.Agama.Storage1.DASD.Manager"
         private_constant :DASD_MANAGER_INTERFACE
 
--- a/service/lib/agama/dbus/storage/manager.rb
+++ b/service/lib/agama/dbus/storage/manager.rb
@@ -26,7 +26,7 @@ require "agama/dbus/with_service_status"
 require "agama/dbus/interfaces/progress"
 require "agama/dbus/interfaces/service_status"
 require "agama/dbus/interfaces/validation"
-require "agama/dbus/interfaces/dasd"
+require "agama/dbus/storage/dasd_manager_interface"
 require "agama/dbus/storage/proposal"
 require "agama/dbus/storage/proposal_settings_converter"
 require "agama/dbus/storage/volume_converter"
@@ -64,7 +64,7 @@ module Agama
           register_software_callbacks
           return unless Yast::Arch.s390
 
-          singleton_class.include DBus::Interfaces::Dasd
+          singleton_class.include DasdManagerInterface
           register_and_extend_dasd_callbacks
         end
 


web/src/components/core/NotificationMark.jsx
```diff
@@ -38,5 +40,23 @@ import React from "react";
...
+  }, [client, cancellablePromise, setHasIssues]);
+
+  useEffect(() => {
+    checkIssues();
+    return client.onIssuesChange(checkIssues);
+  }, [client, checkIssues]);
+
+  return (
+    <If
+      condition={hasIssues}
```
dgdavid, line 58: Thinking aloud:

Maybe in the future we will display the mark because whatever other reason, not only for issues. If so, a way to know the reasons why there are notifications will be needed too.

But for me this is a very good start because it's the only use case we have by now. We can improve it as soon as we have another reason for displaying such a UI element.

Thus, I guess we can use the component in the "Show Issues" link for displaying the mark close to it too. CSS will need adjustments, though.

web/src/components/core/IssuesPage.jsx
```diff
@@ -0,0 +1,170 @@
...
+
+  return (
+    <div>
+      <HelperText className="issue">
+        <HelperTextItem variant={variant} hasIcon icon={<Icon name={icon} />}>
+          {issue.description}
+        </HelperTextItem>
+        <If
+          condition={hasDetails}
+          then={<HelperTextItem>{issue.details}</HelperTextItem>}
```
dgdavid, line 52: NP: let's use [`pre`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element) by now for honoring as mush as possible the details format (break lines and spaces)

```suggestion
          then={<HelperTextItem><pre>{issue.details}</pre></HelperTextItem>}
```

web/src/client/issues.test.js
```diff
@@ -0,0 +1,94 @@
...
+  it("returns all the storage issues", async () => {
+    const client = new IssuesClient(clientsWithIssues);
+
+    const { storage } = await client.getAll();
+    expect(storage).toEqual(expect.arrayContaining(storageIssues));
+  });
+});
+
+describe("#any", () => {
+  describe("if there are storage issues", () => {
```
dgdavid, line 58: NP: now we only have issues from storage, but I guess it does not matter the source of issues for #any. I mean, I guess we are not going to test each case individually.

web/src/components/core/InstallButton.jsx
```diff
@@ -22,50 +22,75 @@
...
+      );
+    };
+
+    return (
+      <p className="bold">
+        There are some reported issues. Please, check <IssuesLink text="the list of issues" /> before
+        proceeding with the installation.
+      </p>
+    );
+  };
```
dgdavid, line 48: NP: would be nice to add an screenshot of this in the PR description

web/src/components/core/InstallButton.jsx
```diff
@@ -22,50 +22,75 @@
...
+  );
+};
 
 const CannotInstallPopup = ({ onClose }) => (
   <Popup
     title="Problems Found"
     isOpen
   >
-    <Text>
+    <p>
```
dgdavid, line 78: :+1:

web/src/components/core/IssuesPage.jsx
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Renders an issue
+ * @component
+ *
+ * @param {object} props
+ * @param {import ("~/client/mixins").Issue} props.issue
+ */
+const IssueItem = ({ issue }) => {
+  const variant = issue.severity === "warn" ? "warning" : "error";
+  const icon = issue.severity === "warn" ? "warning" : "error";
```
dgdavid, line 41: NP: they look like a duplicated line. Maybe we can go for an `severity`, `type`, `issueType`, or similar and use it in both, _variant_ and _name_ props.

web/src/components/software/ChangeProductLink.test.jsx
```diff
@@ -70,7 +70,7 @@ describe("ChangeProductLink", () => {
     });
 
     it("renders a link for navigating to the selection product page", async () => {
-      installerRender(<ChangeProductLink />, { usingProvider: true });
+      installerRender(<ChangeProductLink />);
```
dgdavid, line 73: :roll_eyes:

dgdavid (commented):
Looks good, just a few minor comments. Once again, I'd ask for another review from someone less involved than me in the PR.

dgdavid (approved):

web/src/client/issues.test.js
joseivanlopez: Yes, it does not matter the source, but I would like to check it returns true for each of them.

web/src/components/core/IssuesPage.jsx
joseivanlopez: Well, I was also thinking about it when I wrote that lines. But having both variant and icon name with exactly the very same value ("warning" or "error") was only a great coincidence. For example, if we want to use another icon for errors, then we need to split it in two lines again.

PR #543, opened 2023-04-20
Tooltip for the autocalculated partition size limits

lslezak:
## Problem

- There is no hint for the users which settings can affect the auto calculated limit sizes
- https://trello.com/c/3L9aF34g/3346-2-agama-explanation-for-auto-calculated-volumes
- https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md

## Solution

- Add a tooltip with the items which affect the calculation

## TODO

- [x] Create a separate component for handling tooltips
- [x] Add documentation comments
- [x] Add unit tests

## Screenshots

Icon inside the label, pointer cursor, Popover for the whole label

tooltip5

## Notes

The `Tooltip` component can be configured to trigger on click instead of mouse over. Then it behaves exactly the same as the `Popover` component with the basic layout. So mainly the difference is in the appearance (dark vs. light background).

## Old Screenshots

<details>
<summary>Previous versions</summary>

calculated_tooltip

### Tooltip

This uses the [Tooltip](https://www.patternfly.org/v4/components/tooltip) component.

tooltip2

### Popover

This uses the [Popover](https://www.patternfly.org/v4/components/popover) component.

The Popover displays a closing symbol, by default it has focus which looks a bit ugly:

tooltip3

With disabled closing symbol:

tooltip4

</details>

### DIFF
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Apr 27 08:24:27 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
 
-- Display details for the autocalculated label in the storage
+- Display details for the "autocalculated" label in the storage
   settings
 
 -------------------------------------------------------------------
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Apr 27 08:24:27 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Display details for the autocalculated label in the storage
+  settings
+
+-------------------------------------------------------------------
 Tue Apr 25 13:42:22 UTC 2023 - David Diaz <dgonzalez@suse.com>
 
 - UI: Fix dropdown content alignment at storage proposal page
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -196,3 +196,8 @@ section > .content {
     --pf-c-progress--GridGap: var(--spacer-small);
   }
 }
+
+// compact lists in tooltips
+[role="tooltip"] li + li {
+  margin: 0;
+}
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -207,6 +207,3 @@ section > .content {
   margin-left: 4px;
 }
 
-.with_description {
-  cursor: pointer;
-}
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -197,8 +197,8 @@ section > .content {
   }
 }
 
-// compact lists in tooltips
-[role="tooltip"] li + li {
+// compact lists in popover
+.pf-c-popover li + li {
   margin: 0;
 }
 
@@ -207,6 +207,6 @@ section > .content {
   margin-left: 4px;
 }
 
-.has_help {
+.with_description {
   cursor: pointer;
 }
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -201,9 +201,3 @@ section > .content {
 .pf-c-popover li + li {
   margin: 0;
 }
-
-// extra margin for "i" icons in labels
-.pf-c-label__content > svg {
-  margin-left: 4px;
-}
-
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -201,3 +201,12 @@ section > .content {
 [role="tooltip"] li + li {
   margin: 0;
 }
+
+// extra margin for "i" icons in labels
+.pf-c-label__content > svg {
+  margin-left: 4px;
+}
+
+.has_help {
+  cursor: pointer;
+}
--- a/web/src/components/core/Attribute.jsx
+++ b/web/src/components/core/Attribute.jsx
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Label } from "@patternfly/react-core";
+
+import { Description } from "~/components/core";
+import { Icon } from "~/components/layout";
+
+/**
+ * @component
+ *
+ * @param {object} hint - content displayed in a bubble after clicking the content
+ */
+export default function Attribute ({ hint = null, children }) {
+  if (hint) {
+    return (
+      <Description description={hint}>
+        <Label isCompact>{children}<Icon name="info" size="16" /></Label>
+      </Description>
+    );
+  }
+
+  return <Label isCompact>{children}</Label>;
+}
--- a/web/src/components/core/Attribute.jsx
+++ b/web/src/components/core/Attribute.jsx
@@ -38,7 +38,7 @@ export default function Attribute ({ description, children }) {
   if (description) {
     return (
       <Description description={description}>
-        <Label isCompact>{children}<Icon name="info" size="16" /></Label>
+        <Label isCompact>{children}&nbsp;<Icon name="info" size="16" /></Label>
       </Description>
     );
   }
--- a/web/src/components/core/Attribute.jsx
+++ b/web/src/components/core/Attribute.jsx
@@ -28,7 +28,7 @@ import { Icon } from "~/components/layout";
 /**
  * Display a label with additional details. The details are displayed after
  * clicking the label and the "i" icon indicates available details.
- * If the label is not defined or is empty it behaves  like a plain label.
+ * If the label is not defined or is empty it behaves like a plain label.
  * @component
  *
  * @param {object} description details displayed after clicking the label
--- a/web/src/components/core/Attribute.jsx
+++ b/web/src/components/core/Attribute.jsx
@@ -26,14 +26,18 @@ import { Description } from "~/components/core";
 import { Icon } from "~/components/layout";
 
 /**
+ * Display a label with additional details. The details are displayed after
+ * clicking the label and the "i" icon indicates available details.
+ * If the label is not defined or is empty it behaves  like a plain label.
  * @component
  *
- * @param {object} hint - content displayed in a bubble after clicking the content
+ * @param {object} description details displayed after clicking the label
+ * @param {object} children the content of the label
  */
-export default function Attribute ({ hint = null, children }) {
-  if (hint) {
+export default function Attribute ({ description, children }) {
+  if (description) {
     return (
-      <Description description={hint}>
+      <Description description={description}>
         <Label isCompact>{children}<Icon name="info" size="16" /></Label>
       </Description>
     );
--- a/web/src/components/core/Description.jsx
+++ b/web/src/components/core/Description.jsx
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Popover } from "@patternfly/react-core";
+
+/**
+ * @component
+ *
+ * @param {object} description - content displayed in a bubble after clicking the content
+ */
+export default function Description ({ description, children, ...otherProps }) {
+  return (
+    <Popover showClose={false} bodyContent={description} {...otherProps}>
+      <span className="with_description">{children}</span>
+    </Popover>
+  );
+}
--- a/web/src/components/core/Description.jsx
+++ b/web/src/components/core/Description.jsx
@@ -23,14 +23,21 @@ import React from "react";
 import { Popover } from "@patternfly/react-core";
 
 /**
+ * Displays details popup after clicking the children elements
  * @component
  *
- * @param {object} description - content displayed in a bubble after clicking the content
+ * @param {object} description content displayed in a popup
+ * @param {object} children the wrapped content
  */
 export default function Description ({ description, children, ...otherProps }) {
-  return (
-    <Popover showClose={false} bodyContent={description} {...otherProps}>
-      <span className="with_description">{children}</span>
-    </Popover>
-  );
+  if (description) {
+    return (
+      <Popover showClose={false} bodyContent={description} {...otherProps}>
+        <span className="with_description">{children}</span>
+      </Popover>
+    );
+  }
+
+  // none or empty description, just return the children
+  return children;
 }
--- a/web/src/components/core/Description.jsx
+++ b/web/src/components/core/Description.jsx
@@ -33,7 +33,7 @@ export default function Description ({ description, children, ...otherProps }) {
   if (description) {
     return (
       <Popover showClose={false} bodyContent={description} {...otherProps}>
-        <span className="with_description">{children}</span>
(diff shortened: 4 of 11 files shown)


web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -56,6 +57,29 @@ const sizeText = (size) => {
   return filesize(size, { base: 2 });
 };
 
+const AutoCalculatedTooltip = ({ volume, children }) => {
+  // the size is not affected by snapshots or other volumes
+  if (!volume.snapshotsAffectSizes && volume.sizeRelevantVolumes.length === 0) {
+    const content = <Text>These limits are not affected by any other settings</Text>;
+    return <Tooltip content={content}>{children}</Tooltip>;
+  }
```
ancorgs, line 65: If it's not affected then there is no tooltip because there is no "auto-calculated" label. Isn't it?

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -56,6 +57,29 @@ const sizeText = (size) => {
...
+
+  const content = (
+    <>
+      <Text>These limits are affected by:</Text><br />
+      <List>
+        {volume.snapshotsAffectSizes &&
+          <ListItem>The configuration of snapshots</ListItem>}
+        {volume.sizeRelevantVolumes.length > 0 &&
+          <ListItem>By presence of other volumes ({volume.sizeRelevantVolumes.join(", ")})</ListItem>}
```
ancorgs, line 74: We are using "file system" instead of volume in the UI.

ancorgs: Duplicated "by", BTW

web/src/components/storage/ProposalVolumes.jsx
lslezak: Yes, that's it.

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -56,6 +57,29 @@ const sizeText = (size) => {
...
+      </List>
+    </>
```
dgdavid, line 76: NP: Please, do not use a hard-coded line break `<br />`. Instead, wrap these elementes with a `div` using the `stack` CSS class.

```suggestion
    <div className="stack">
      <Text>These limits are affected by:</Text>
      <List>
        {volume.snapshotsAffectSizes &&
          <ListItem>The configuration of snapshots</ListItem>}
        {volume.sizeRelevantVolumes.length > 0 &&
          <ListItem>By presence of other volumes ({volume.sizeRelevantVolumes.join(", ")})</ListItem>}
      </List>
    </div>
```

dgdavid: Also, I'd use a simple `<p>` instead of `<Text>` because I feel the latter does not provide too much value at this time and I tend to use plain HTML when possible.

dgdavid:
BTW, sorry for not writing it in the Trello card (I did in the chat during the planning, my fault) but I wonder if a [Popover](https://www.patternfly.org/v4/components/popover) would be a better component for this use case. Not sure, but maybe worth it exploring the idea.

dgdavid:
> That said, both looks good to me for this use case, the `Tooltip` and the `Popover`

Thinking twice, I would add an "i" icon or similar to activate whatever component we use. Why? To prevent users from associating that labels might have more information hidden, and "force" them to hover over all the labels to see if there are any more things to see.

joseivanlopez:
IMHO, I would prefer to have the "i" icon more integrated with the label (e.g., placing it inside, similar to the examples here https://www.patternfly.org/v4/components/label#examples). I am not sure whether having the "i" outside is a good idea when there are more than one labels (e.g., auto and snapshots). I expect the "auto" label would provide the explanation about auto-calculating, and what conditions affects the calc. And the "snapshots" label would provide would explain that snapshots is active and the effects it has, for example.

Some ideas that come to my mind:

* We could "sacrifice" the custom icon (e.g, auto icon, snapshots icon) in favor of the "i" icon when the label needs to show info.
* Add some kind of indicator (e.g., underlining the label text) to indicate that the label is clickable.
* Use another kind of element instead of labels.

dgdavid:
My thoughts related to the PR description:

> The Popover displays a closing symbol, by default it has focus which looks a bit ugly:

I've been working in how the focused/active Element is displayed by the browser. It's a bit tricky and I already did some changes at https://github.com/openSUSE/agama/pull/544 

There is pending work related with :foucs and :focus-visible CSS pseudo selectors.

Anyway, we always can improve the uglyness :) :)

> The Tooltip component can be configured to trigger on click instead of mouse over. Then it behaves exactly the same as the Popover component with the basic layout. So mainly the difference is in the appearance (dark vs. light background).

If we're going to use always the basic layout we can go ahead with the tooltip. The background is not a problem, we probably can change it by redefining a PF CSS Custom Property. Don't worry because of that. 

Maybe the way could be to the same we did with the `Em` component: to make our own component, although initially it will be just a wrapper of a PF component. Thus, if we change our mind and want to switch from Tooltip to a Popover we can do it easily, apart from other stuff. But I fully understand that it is out of the scope of your current PBI :)

dgdavid:
> On the other hand, I fear that labels in tables should always show some info, and having the "i" for all them would be a bit ... repetitive. I guess we need to find a better element for that labels in tables.

I do no agree. Or, in other words, I'm not able to foresee that behaviour. In any case, for me it's not the same an icon that an icon inside a label :man_shrugging:

dgdavid:
> I do not agree. Or, in other words, I'm not able to foresee that behaviour. In any case, for me it's not the same an icon that an icon inside a label man_shrugging

I meant, I do not feel that

> labels in tables should always show some info

If that's the case, I fear we're misusing labels. A label should be auto explanatory. This storage case is a very special case in having to explain why something happened. Or maybe a fault in our side trying to compact a lot of information in just one word.

Not sure. I fully trust in the fact that we're going to improve and polish the UI in each iteration, but assuming now that labels show always be linked to a tooltip or a "show more" behavior is something I can't do at this time :)

joseivanlopez:
> Not sure. I fully trust in the fact that we're going to improve and polish the UI in each iteration, but assuming now that labels show always be linked to a tooltip or a "show more" behavior is something I can't do at this time :)

I don't know, but now we have two possible labels for storage ("auto" and "snapshots"). For both I expect the "i" icon. Anyway, we could start with that approach (i.e., using "i" icon or nothing) and reevalute later if needed.

dgdavid:
> > Not sure. I fully trust in the fact that we're going to improve and polish the UI in each iteration, but assuming now that labels show always be linked to a tooltip or a "show more" behavior is something I can't do at this time :)
> 
> I don't know, but now we have two possible labels for storage ("auto" and "snapshots"). For both I expect the "i" icon. Anyway, we could start with that approach (i.e., using "i" icon or nothing) and reevalute later if needed.

But still being the very same special case in my eyes. Although I don't know now why you expect more explanations about snapshots in a Tooltip/Popover/i form, it just a storage section. Not enough for thinking that all labels on all tables will have an associated tooltip IMHO.

> Anyway, we could start with that approach (i.e., using "i" icon or nothing) and reevalute later if needed.

That's what I think it's the best we can do right now... and even later. Start simple, improve it later if needed.

joseivanlopez:
> But still being the very same special case in my eyes. Although I don't know now why you expect more explanations about snapshots in a Tooltip/Popover/i form, it just a storage section. Not enough for thinking that all labels on all tables will have an associated tooltip IMHO.

"Snapshots" have implications, I expect to explain something to the user. And somehow, I feel that labels in tables would be used to indicate very specefic active options, but as said, let's see later how it evolves.

web/src/assets/styles/blocks.scss
```diff
@@ -207,6 +207,6 @@ section > .content {
   margin-left: 4px;
 }
 
-.has_help {
+.with_description {
   cursor: pointer;
 }
```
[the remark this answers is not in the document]
dgdavid: Ok, let me clarify the situation for future readers:

My previous comment was motivated by a miss understanding from my side. After talking about it in IRC I realized that you are using this CSS *utility* class for the element that triggers the `Popover`, not for it contents itself. My fault, please excuse me. That's one thing I sometimes _hate_ from PatternFly: sometimes the content is the children, sometimes is a prop. I'm pretty sure they have strong reasons for that... but I'd use the children as content and the prop for creating the trigger :)

So far, so good. It makes sense that show the right cursor to the users in order to make them easy to detect that there is something clickable. Thus, I think we should use a native clickable element. As we discussed in IRC too, GitHub uses a very nice and well styled `<details>` for building their own popovers. Very clever, IMHO. But please, go ahead with your approach for this PBI. We can re-think that thing later... and I personally think you deserve both, to get the PR merged and to work in an improvement, if needed, later.

Regarding the CSS utility class, I'd do two things:

* Name it `cursor-pointer`, which makes it more explicit and generic.
* Place it in the `utilities.scss` file instead. 

I really hope to not see an intensive use of this CSS class across the project... because it does not make sense at all. 

Thanks for the feedback gave in the IRC Libera #yast channel and, again, excuse me for the noise :) In any case, I feel it has been constructive in the end.

web/src/components/core/Attribute.jsx
```diff
@@ -0,0 +1,47 @@
...
+ * @component
+ *
+ * @param {object} description details displayed after clicking the label
+ * @param {object} children the content of the label
+ */
+export default function Attribute ({ description, children }) {
+  if (description) {
+    return (
+      <Description description={description}>
+        <Label isCompact>{children}<Icon name="info" size="16" /></Label>
```
dgdavid, line 41: I'm not sure now, but I think that you can leave a space there instead of adding an extra margin via CSS.

```suggestion
        <Label isCompact>{children} <Icon name="info" size="16" /></Label>
```

web/src/components/core/Attribute.jsx
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Display a label with additional details. The details are displayed after
+ * clicking the label and the "i" icon indicates available details.
+ * If the label is not defined or is empty it behaves  like a plain label.
+ * @component
+ *
+ * @param {object} description details displayed after clicking the label
+ * @param {object} children the content of the label
+ */
+export default function Attribute ({ description, children }) {
```
dgdavid, line 37: NP: I do not like the `Attribute` name. Sounds weird. That said, I have no better suggestion, although the word "badge" comes quickly to my mind.

dgdavid: >  Display a label with additional **details**. The **details** are displayed after
>  clicking the label and the "i" icon indicates available **details**.

Maybe the component description is telling us something... `Deatils` could be the name for this component. However, yesterday we somehow agree on avoiding to use names that _collides_ with native HTML elements. And, as you already know because we use it a lot in the PRs descriptions, there is a [`details`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) HTML element

web/src/components/core/Attribute.jsx
```diff
@@ -0,0 +1,47 @@
...
+  if (description) {
```
dgdavid, line 38: NP: We also have an [`If` component](https://github.com/openSUSE/agama/blob/master/web/src/components/core/If.jsx) for this kind of scenarios where we are just displaying a component or another based on a condition.

web/src/components/core/Attribute.jsx
```diff
@@ -0,0 +1,47 @@
...
+import React from "react";
+import { Label } from "@patternfly/react-core";
+
+import { Description } from "~/components/core";
+import { Icon } from "~/components/layout";
+
+/**
+ * Display a label with additional details. The details are displayed after
+ * clicking the label and the "i" icon indicates available details.
+ * If the label is not defined or is empty it behaves  like a plain label.
```
dgdavid, line 31:
```suggestion
 * If the label is not defined or is empty it behaves like a plain label.
```

web/src/components/core/Description.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { Popover } from "@patternfly/react-core";
+
+/**
+ * Displays details popup after clicking the children elements
+ * @component
+ *
+ * @param {object} description content displayed in a popup
+ * @param {object} children the wrapped content
+ */
+export default function Description ({ description, children, ...otherProps }) {
```
dgdavid, line 32: I'm a bit lost here. I think this component is redundant with the `Attribute` component, which returns a plain label when no description is given. And it's the only place where `Description` is used.

web/src/components/core/Description.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+    // click it
+    const item_node = screen.getByText(item);
+    await user.click(item_node);
+
+    // then the description is visible in a dialog
+    screen.getByRole("dialog");
+    screen.getByText(description);
+  });
+
+  const empty_tester = async (subject) => {
```
dgdavid, line 47: NP: until now we have been writing more explicit tests, not using this kind of helper functions because usually they saves keystrokes at the cost of doing them hard to read/follow. Anyway, I'm not against them. Actually, I have no strong opinion here. Just naming suggestions

* empty_tester -> expectNoDescription or something similar. Note the camelCase instead the snake_case
* subject -> content, which is what you are "plain rendering"
*

dgdavid (commented):
Thanks for a very great job, @lslezak :+1: I know you have put a lot of time and love in this PBI :heart: So, despite I have left some nitpicking comments, I think the PR deserves a review from another team member since maybe I'm a bit biased due to the fact that the approach we ended up following here do not look fully right to me (and, to be clear, I can be completely wrong).

web/src/components/core/Description.jsx
lslezak: The reason is that I'd like to share the code which uses `Popover` for displaying details or help so it can be used elsewhere. When we later decide to use `Tooltip` instead of `Popover` (or change the look) then we need to update just one place.

web/package/cockpit-agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Apr 27 08:24:27 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Display details for the autocalculated label in the storage
```
mvidner, line 4:
```suggestion
- Display details for the "autocalculated" label in the storage
```

quotes matter :-)

web/src/components/core/Description.test.jsx
lslezak: The tests we like 95% the same code, repeating the same three times was ugly to me. So I created this shared function.

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -56,6 +57,33 @@ const sizeText = (size) => {
   return filesize(size, { base: 2 });
 };
 
+/**
+ * Generates an hint describing which attributes affect the auto-calculated limits.
+ * If the limits are not affected then it returns `null`.
+ * @function
+ *
+ * @param {object} volume - storage volume object
+ * @returns {object} component to display (can be `null`)
```
mvidner, line 66: can we be more specific with the documented types? `@returns {ReactComponent} or null` ?
same for _volume_

web/src/components/core/Attribute.jsx
lslezak: Yeah, the naming is usually the biggest problem... It's not the best name but I could not find any better. 

Suggesstions? Ideas? :thinking:  Anyone...

mvidner: I am aware that I am bikeshedding... but "attribute" sounds like a non-UI thing to me, colliding with an HTML/XML concept.

What about Hint, Tooltip, Tip?

web/src/components/storage/ProposalVolumes.jsx
mvidner: And same for the other `@param {object}` added in this PR

dgdavid (approved):

web/src/assets/styles/blocks.scss
```diff
@@ -196,3 +196,14 @@ section > .content {
     --pf-c-progress--GridGap: var(--spacer-small);
   }
 }
+
+// compact lists in popover
+.pf-c-popover li + li {
```
joseivanlopez, line 201: Shouldn't  these stiles go to "patternfly-overrides.scss"?

web/src/components/core/Attribute.jsx
lslezak: Unfortunately "Hint" and "Tooltip" are already used by PatternFly, I'd avoid using the same names.

"Tip" sounds OK.

web/src/components/core/Description.jsx
joseivanlopez: Umm, but we can do the same having only the `Attribute` component, right?. I mean, we can modify the internal implementation of `Attribute` to use a tooltip or to provide other look. And, I would use `Attribute` (or `Tip`) only for the cases in which we want a label with more info. Otherwise, I would use the `Label` directly.

web/src/assets/styles/blocks.scss
lslezak: PF defines ".pf-c-list li + li" this uses ".pf-c-popover li + li" so technically it is not an override but rather an extension.

joseivanlopez (approved):
LGTM

web/src/components/core/Attribute.jsx
lslezak: I tried that it does not look nice with multiple nested components spread over several lines:

```js
  return (
    <If
      condition={description}
      then={
        <Popover showClose={false} bodyContent={description} {...otherProps}>
          <span className="cursor-pointer">{children}</span>
        </Popover>
      } else={
        // none or empty description, just return the children
        children
      }
    />
  );
```

IMHO this looks a bit better:

```js
  if (description) {
    return (
      <Popover showClose={false} bodyContent={description} {...otherProps}>
        <span className="cursor-pointer">{children}</span>
      </Popover>
    );
  }

  // none or empty description, just return the children
  return children;

```

web/src/components/core/Description.jsx
lslezak: The point is that you can use a tooltip also for something else, not only `Label`.

joseivanlopez: Ah, now I got it. Thanks!

PR #545, opened 2023-04-21
[web] Change the location of page options

dgdavid:
## Problem

Since _page actions_ (also known as "Contextual actions") were introduced, there have been complaints regarding their "discoverability" from within and outside the team.

It seems that there is too much subtlety in the navigation (and, admittedly, I like that fact) to make it easy for the users to discern when they are in a new page and, thus, the _Hamburguer Menu_ icon (aka _Sidebar_) may contain different options.

The situation can be further complicated now that we plan to add a _global notification mark_ alongside the _Hamburguer Menu_ icon for trying to catch users' attention in a no disruptive way when there is something that might be interesting for them but not critical or urgent enough for blocking the UI with a Popup.

<details>
<summary>Click to show/hide few screenshots</summary>

---

| Overview page (sidebar closed) | Overview page (sidebar open) | Storage page (sidebar closed) | Storage page (sidebar open)
|-|-|-|-|
|Overview page (sidebar closed) |Overview page (sidebar open) | Storage page (sidebar closed)| Storage page (sidebar open) |
</details>

## Solution

To extract the page related actions from the Sidebar and add a more evident element in the header to reveal them.

| Menu closed | Menu open |
|-|-|
| Storage page (menu closed) | Storage page (menu open) |

<details>
<summary>Click to show/hide more screenshots</summary>

---

| Overview page (sidebar closed) | Overview page (sidebar open) | Storage page (sidebar closed) | Storage page (sidebar open)
|-|-|-|-|
|Overview page (sidebar closed) | Overview page (sidebar closed) |Storage page (sidebar closed) | Storage page (sidebar open) |

</details>

## Side effects

These changes entail the removal of the `PageOptions` component introduced at https://github.com/openSUSE/agama/pull/462 as well as the `Sidebar.OpenButton` introduced at https://github.com/openSUSE/agama/pull/500

## Testing

- Added/Adapted unit tests
- Tested manually

---

:warning: screenshots above can be slightly out dated after code review if only texts or descriptions were changed.


web/src/components/storage/ProposalPageOptions.jsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const ISCSILink = () => {
+  const href = useHref("/storage/iscsi");
+
+  return (
+    <ContextualActions.Item
+      key="iscsi-link"
+      href={href}
+      description="Connect to iSCSI targets"
```
dgdavid, line 48: With this new approach, it's a lot of easy to add a description for each menu entry and help the users to quickly identify that it's really the action they want to choose. That's why I added this _description_ prop here after talking about that with @ancorgs. Please, check it and propose a better description if you have one in mind. The idea is to avoid using only the acronyms as much as possible too.

web/src/components/storage/ProposalPageOptions.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const { storage: client } = useInstallerClient();
+
+  useEffect(() => {
+    client.dasd.isSupported().then(setShowDasdLink);
+  }, [client.dasd]);
+
+  return (
+    <ContextualActions>
+      <ContextualActions.Group
+        label="Configure"
```
dgdavid, line 66: Instead of having a lot of "similar actions" like "Configure DASD", "Configure iSCSI", and "Configure FCoE" for example, we decided to group them under the "Configure" group. Now can be difficult to see the benefit of grouping page options, but think that most probably it will grow sooner than later. Anyway we can re-evalute it if needed an introduce groups later, but please keep in mind that its a bit confusing having a bunch of "Configure ACRONYM" options in a row.

web/src/components/storage/ProposalPageOptions.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { If, ContextualActions } from "~/components/core";
+
+const DASDLink = () => {
+  const href = useHref("/storage/dasd");
+
+  return (
+    <ContextualActions.Item
+      key="dasd-link"
+      href={href}
+      description="Manage and format"
```
dgdavid, line 34: As in the iSCSI link, 

> With this new approach, it's a lot of easy to add a description for each menu entry and help the users to quickly identify that it's really the action they want to choose. That's why I added this description prop here after talking about that with @ancorgs. Please, check it and propose a better description if you have one in mind. The idea is to avoid using only the acronyms as much as possible too.

web/src/components/core/ContextualActions.jsx
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+import React, { useState } from 'react';
+import { Button, Dropdown, DropdownItem, DropdownGroup } from '@patternfly/react-core';
+import { Icon, ContextualActions as ContextualActionsSlot } from "~/components/layout";
+
+const Toggler = ({ onClick }) => {
+  return (
+    <Button onClick={onClick} variant="plain">
+      <Icon name="expand_more" />
```
dgdavid, line 29: I actually like the "+" icon, which for me has a lot of sense (_more page actions_). But since I thought it might be a bit controversial... I opted for using a "dropdown arrow" like one. Anyway, see how they look side by side

| "Expand more" icon | "Add" icon |
|-|-|
| Storage page using "expand more" icon | Storage page using "add" icon |

Either way, there are plenty of them at https://fonts.google.com/icons. Have a look to see if you find one that fits better.

web/src/components/network/NetworkPage.jsx
```diff
@@ -24,35 +24,8 @@ import { Button, Skeleton } from "@patternfly/react-core";
...
-        onClick={() => setWifiSelectorOpen(true)}
-        icon={<Icon name="wifi_find" size="24" />}
-      >
-        Connect to a Wi-Fi network
-      </Button>
-      <WifiSelector isOpen={wifiSelectorOpen} onClose={() => setWifiSelectorOpen(false)} />
-    </>
-  );
-};
+import { If, Page, Section } from "~/components/core";
```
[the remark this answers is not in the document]
teclator: it should not collide by now as we are working in the design of the DBUS API service

web/src/components/network/NetworkPageOptions.jsx
```diff
@@ -0,0 +1,49 @@
...
+  wifiScanSupported = false,
+  openWifiSelectorCallback
+}) {
+  return (
+    <ContextualActions>
+      <ContextualActions.Item
+        key="open-wifi-selector"
+        onClick={openWifiSelectorCallback}
+        isDisabled={!wifiScanSupported}
+        description={!wifiScanSupported ? <WifiScanNotSupported /> : null}
```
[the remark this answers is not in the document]
teclator: If there is no wireless device then it does not make sense to show any link at all...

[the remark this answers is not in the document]
dgdavid: > If there is no wireless device then it does not make sense to show any link at all...

Ok, I'll change the behavior then. Thanks for the input.

web/src/components/core/ContextualActions.jsx
joseivanlopez: I like both, and I see the down arrow more generic. I would it for now.

joseivanlopez (approved):
LGTM

joseivanlopez (changes requested):
Sorry, I overlooked one detail. We need to fix the `ProposalSettingsSection` which stills uses the sidebar: https://github.com/openSUSE/agama/blob/master/web/src/components/storage/ProposalSettingsSection.jsx#L170

dgdavid:
> Sorry, I overlooked one detail. We need to fix the `ProposalSettingsSection` which still links to the sidebar: https://github.com/openSUSE/agama/blob/master/web/src/components/storage/ProposalSettingsSection.jsx#L170

Right! I though I had adapted it already. Which remains me that we could drop the [Sidebar.OpenButton](https://github.com/openSUSE/agama/blob/306aa118f70b3823d9a607267af5da8ed678265e/web/src/components/core/Sidebar.jsx#L139-L154) by now too.

web/src/components/storage/ProposalPageOptions.jsx
jreidinger: Just to be sure this leads only to new connections or does it allow also to modify existing/delete connections?

dgdavid: Actually, it leads to the iSCSI management page. The idea to use the "Create" word it's based in the fact that probably it's the "most wanted" action

jreidinger: In that case I think "Manager connections to iSCSI targets" sounds better to me.

web/src/components/core/ContextualActions.jsx
joseivanlopez: As said, I also like the arrow. But I am asking myself if the icon is dicoverable enough. At first sight it looks to me like a button for minimizing or something similar (maybe it is my fault as user). Moreover, how can we refer to it? I mean, in the devices selector popup we have a sentence for suggesting the user to configure more devices if needed. Do we need to keep that tip? If so, how?

joseivanlopez: BTW, how do you see to add a button? E.g., "More options V", with good styles to integrate it nicely with the sidebar.

dgdavid: I do not think it is needed such verbosity. But if you want, go ahead.

dgdavid: Maybe I'd use the tooltip here.

jreidinger (approved):

joseivanlopez (approved):

PR #546, opened 2023-04-25
[web] Add a the initial version of a notification mark

dgdavid:
A simple component for rendering a notification/status mark, consisting of an absolute right positioned circle on top of a relative positioned element.

The idea is to catch the users' attention for clicking on the element presenting the mark, but having in mind that they can continue with the installation without doing so.

| Overview w/o notifications | Overview with notifications |
|-|-|
| Overview w/o notifications | Overview with notifications |

## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/web/src/components/core/NotificationMark.jsx
+++ b/web/src/components/core/NotificationMark.jsx
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+
+/**
+ * A notification mark that  icon for catching the users attention when there is
+ * something that can be interesting for them but not urgent enough to use a
+ * (blocking) Popup.
+ *
+ * @component
+ *
+ * Initially though to be displayed on top of the Sidebar icon.
+ *
+ * Use only when the information to show might be overlooked without risk and/or
+ * when the information will be displayed sooner or later in other way (in a
+ * confirmation dialog, for example).
+ *
+ * @param {object} props
+ * @param {string} props.label - the label to be announced by screen readers
+ */
+export default function NotificationMark ({ label }) {
+  return <span className="notification-mark" role="status" aria-label={label} />;
+}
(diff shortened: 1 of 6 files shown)


web/src/components/core/Sidebar.jsx
```diff
@@ -93,6 +94,7 @@ export default function Sidebar ({ children }) {
           aria-controls="navigation-and-options"
           aria-expanded={isOpen}
         >
+          <NotificationMark />
```
dgdavid, line 97: It should be rendered conditionally. We can do it either, by adding the logic for rendering here or, ideally, in the own component.

web/src/components/core/NotificationMark.jsx
```diff
@@ -0,0 +1,42 @@
...
+ * Initially though to be displayed on top of the Sidebar icon.
+ *
+ * Use only when the information to show might be overlooked without risk and/or
+ * when the information will be displayed sooner or later in other way (in a
+ * confirmation dialog, for example).
+ *
+ * @param {object} props
+ * @param {string} props.label - the label to be announced by screen readers
+ */
+export default function NotificationMark ({ label }) {
```
dgdavid, line 40: Label could be an internal status instead of a prop. It depends on the approach we're going to follow. Although it could end up being both, the label and the internal status, one as a fallback or even complement of the other.

web/src/components/core/NotificationMark.jsx
```diff
@@ -0,0 +1,42 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+
+/**
+ * A notification mark that  icon for catching the users attention when there is
```
joseivanlopez, line 25: NP: "mark that  icon"

web/src/components/core/NotificationMark.jsx
```diff
@@ -0,0 +1,42 @@
...
+ * something that can be interesting for them but not urgent enough to use a
+ * (blocking) Popup.
+ *
+ * @component
+ *
+ * Initially though to be displayed on top of the Sidebar icon.
```
joseivanlopez, line 31: What this means? I did not get this explanation.

web/src/components/core/NotificationMark.jsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Use only when the information to show might be overlooked without risk and/or
+ * when the information will be displayed sooner or later in other way (in a
+ * confirmation dialog, for example).
+ *
+ * @param {object} props
+ * @param {string} props.label - the label to be announced by screen readers
+ */
+export default function NotificationMark ({ label }) {
+  return <span className="notification-mark" role="status" aria-label={label} />;
```
joseivanlopez, line 41: If I got it, the logic to render the mark should be added here, right?

web/src/components/core/NotificationMark.jsx
dgdavid: Yeah, too many leftovers there. I meant

> A notification mark for catching the users attention....

Good catch.

web/src/components/core/NotificationMark.jsx
dgdavid: Initially yes, but I really don't know if it will enough or the right approach. I, for sure, would start adding it here.

PR #547, opened 2023-04-25
[web] Fix a dropdown position

dgdavid:
## Problem

The right positioned dropdown for file system actions in the Storage page might cause horizontal scroll because it's positioning its content to the left.

## Solution

To use its `position` prop with the `right` value.

## Testing

- *Tested manually*

## Screenshots

| Before | After |
|-|-|
| Dropdown content causing horizontal scroll | Dropdown content right aligned with the toggle |


dgdavid (commented):
Excuse if it's already answered, but do we need an entry in the changes files for these kinds of small fixes?

joseivanlopez (approved):
LGTM

imobachgs:
I do not agree at all: it is a user-face change, so it should be included in the changes files. We can skip this one but, please, take that into account in the future.

imobachgs (approved):

PR #549, opened 2023-04-25
[web] Do not reserve space for section icons

dgdavid:
## Problem

The initial sketch of Agama UI included icons close to the section titles, not only for aesthetic reasons but as a visual aid for people with no sight problems. They can help to quickly identify a section and also to create an indentation that helps to identify at first sight the section related content.

But, sadly, we do not have icons for all sections. What is more, we can have full pages without a section icon, which is the case of the Storage Proposal Page right now.

For these pages, the indentation looks not justified or a bit wrong, specially when adding elements that do not obey the section layout.

| Page mounting sections with icons | Page mounting sections w/o icons |
|-|-|
| Screen Shot 2023-04-25 at 11 46 12 | Screen Shot 2023-04-25 at 11 45 56 |

## Solution

Since at this time we're using a CSS grid for laying out the section elements, using a not fixed value for the "icon" column looks like a good compromise fix.

| Before | After |
|-|-|
|Screen Shot 2023-04-25 at 11 45 56 | Screen Shot 2023-04-25 at 11 45 47 |

The only caveats I'm able to see at this time are,

 * Missing both in a same page, sections with and without icons, will create a lot of miss alignments
 * Using the `loading` prop in a section not using a representative icon (and, thus, reserving the space) will cause an ugly space distribution.

|Mixed kind of sections | A fake example of section using loading prop but not icon |
|-|-|
| Screen Shot 2023-04-25 at 11 47 18 | Screen Shot 2023-04-25 at 12 11 50 |

But, as mentioned, this is a compromise fix, and we can see these caveats as a _guideline_ of things we must avoid doing. What is more, for a section without icon the loading prop would be ignored and it must use a Skeleton instead. But that it's for another PR, if this got merged.

## Testing

- Tested manually


joseivanlopez (approved):
If I am not wrong, we don't mix sections with and without icon yet. So I think the fix is a good compromise for now.

PR #551, opened 2023-04-26
Add a network D-Bus service

imobachgs:
According to our [network support design](https://github.com/openSUSE/agama/blob/master/doc/network_support_design.md) document, we decided to implement our network D-Bus API. The idea is to use the same API from the web UI and the CLI.

Of course, we are not implementing a full solution for network handling but relying on existing stuff, like [nmstate](https://nmstate.io/).

⚠️ **The PR is not ready for review yet, but comments are welcome. However, please, read the full description before reviewing the code.**

## Implementation

This PR adds a new `network` module to `agama-lib`. It contains:

* [A data model](https://github.com/openSUSE/agama/blob/add-network-service/rust/agama-lib/src/network/state.rs#L30), which is basically a wrapper around [nmstate::NetworkState](https://docs.rs/nmstate/2.2.8/nmstate/struct.NetworkState.html).
* [A D-Bus service](https://github.com/openSUSE/agama/tree/add-network-service/rust/agama-lib/src/network/dbus).
* A new (and temporary) package to start the service. This logic should be moved to `agama-dbus-server` (what about `agama-service`?) when it is ready (see #533).

## The D-Bus API

We tried to keep the API as simple as possible. At this point, it publishes an object for each known network device, and another for the DNS settings.

```
Service org.opensuse.Agama.Network1:
└─/org
  └─/org/opensuse
    └─/org/opensuse/Agama
      └─/org/opensuse/Agama/Network1
        ├─/org/opensuse/Agama/Network1/Device
        │ ├─/org/opensuse/Agama/Network1/Device/enp2s0f0
        │ ├─/org/opensuse/Agama/Network1/Device/enp5s0f3u1u4
        │ └─/org/opensuse/Agama/Network1/Device/wlp3s0
        └─/org/opensuse/Agama/Network1/Dns
```

Depending on the type of each device, it would expose a different set of interfaces:

* `org.opensuse.Agama.Network1.Device` for all of them containing general information.
* `org.opensuse.Agama.Network1.IPv4` for devices that may have an IP configuration.
* `org.opensuse.Agama.Network1.Wireless` for wireless devices (not implemented yet).

## The D-Bus architecture

When it comes to the D-Bus service, we are unsure whether the approach is right. Currently, the `NetworkService` holds the network configuration behind a `Mutex`. Given that the entry point to modify the settings is the `nmstate::NetworkState` itself, all D-Bus interfaces have access to that object through an [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) (and use the mutex to avoid any collision).

## To do

- [ ] Add support for wireless devices
- [x] Allow setting the gateway
- [x] Add support to write the configuration
- [ ] "Hide" nmstate API behind our structs.
- [ ] Move the errors' definition to the right place.
- [ ] Extend the test coverage
- [ ] Emit events when a property change (most probably on a separate PR)


imobachgs:
Replaced by #272.

PR #553, opened 2023-04-27
[web] Fix CI and update README screenshots

dgdavid:
Now that the storage settings popup is no longer available, let's replace its screenshot by the one for the users page in the README.

**Note** that this PR also introduce two small fixes or improvements for the CI task:

* Add a higher timeout for the storage page.
* Mount `/dev` in the container to allow libstorage find devices


dgdavid:
@joseivanlopez @lslezak 

I don't understand why CI is failing. It works on my machine :tm: 

As far as I can see in the _trace_ downloaded from Summary -> Artifacts, it is not rendering the last UI version. So weird since following the CI output it's clearly checking out the right commit.

| The trace from CI | The trace from my machine |
|-|-|
|Screenshot from 2023-04-28 12-50-17 | Screenshot from 2023-04-28 12-51-03 |

dgdavid:
> @joseivanlopez @lslezak
> 
> I don't understand why CI is failing. It works on my machine tm
> 
> As far as I can see in the _trace_ downloaded from Summary -> Artifacts, it is not rendering the last UI version. So weird since following the CI output it's clearly checking out the right commit.

What is more, I think the test passed wrongly at https://github.com/openSUSE/agama/pull/521, where the Storage UI was changed and the "button" Settings was "dropped" (actually, the header stopped being a link)

I have downloaded the trace from there too and, as you can see below, the test was done against the previous Storage UI.

Screenshot from 2023-04-28 14-48-30


So, definitely there is something wrong with the CI task... or I'm completely lost.

lslezak (approved):

PR #555, opened 2023-05-02
[web] Do not show group options name until needed

dgdavid:
## Problem

During a review meeting, we've spot that displaying the _group options name_ when there is only a single group of options can be more confusing than helpful. 

## Solution

Display the group name only when needed. I.e., when really having more than one group of actions for a set of page options.

## Testing

- Added a new unit test
- Tested manually

## Notes

Since it's related with https://github.com/openSUSE/agama/pull/545 and it has not been merged to master yet, this PR belongs to the same changelog entry. No extra one needed, IMHO.

## Screenshots

| Before | After |
|-|-|
| Screenshot from 2023-05-02 10-10-53 |Screenshot from 2023-05-02 10-10-45 |

### DIFF
--- a/web/src/components/storage/ProposalPageOptions.jsx
+++ b/web/src/components/storage/ProposalPageOptions.jsx
@@ -74,10 +74,7 @@ export default function ProposalPageOptions () {
 
   return (
     <PageOptions>
-      <PageOptions.Group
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #556, opened 2023-05-03
[web] Use a button for displaying the description popover

dgdavid:
## Problem

We're using a [`span`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element) element, which _doesn't mean anything on its own_, for displaying a popover when the user clicks on it. It works and even hints visually to sighted users thanks to a CSS rule, but it isn't intended for user interaction.

## Solution

To use a [`button`](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) instead, which belongs to [`interactive content`](https://html.spec.whatwg.org/multipage/dom.html#interactive-content) category and might help assistive tech to announce the element.

## Testing

- Tested manually

## Screenshots

There are no visual changes, check it below

|Before (using `span`) | After (using `button`) |
|-|-|
|Screenshot from 2023-05-03 11-33-13 | Screenshot from 2023-05-03 11-32-46 |

## Notes

Since it somehow belongs to #543 a changelog entry is not needed here.

### DIFF
--- a/web/src/assets/styles/utilities.scss
+++ b/web/src/assets/styles/utilities.scss
@@ -136,7 +136,3 @@
   /** END block-size fallbacks **/
   block-size: 95dvh;
 }
-
-.cursor-pointer {
-  cursor: pointer;
-}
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #559, opened 2023-05-05
Load proposal if needed

joseivanlopez:
## Problem

If the proposal page is visited before having a proposal, then the page is always in loading state. 

## Solution

Loads the proposal if there is no proposal yet and the service finishes to calculate a proposal.

## Testing

* Added new unit tests
* Tested manually


dgdavid (approved):
LGTM

PR #566, opened 2023-05-09
Deploy also the JSdoc documentation to the GitHub pages

lslezak:
## Problem

So far we use surge.sh for publishing the JSdoc documentation. That has several drawbacks:

- Multiple places with documentation
- The `surge` NPM package has old dependencies with security problems, the GitHub security scanner complains about this
- We need to manage additional external account and need to use some secrets in the GitHub actions

## Solution

- Use GitHub pages, we already use it for publishing the DBus service documentation

## Testing

- For testing the new rendered JSdoc pages see https://lslezak.github.io/agama/jsdoc/ or the new index at https://lslezak.github.io/agama/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,13 +54,6 @@ jobs:
         flag-name: frontend
         parallel: true
 
-    - name: Build JSDoc Documentation
-      run: npm run jsdoc
-
-    - name: Publish JSDoc Documentation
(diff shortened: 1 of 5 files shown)


dgdavid (approved):
Apart from a missing entry in the changelog, It looks good to me.

mvidner (approved):
LGTM, thank you!

PR #568, opened 2023-05-09
Fix publishing GH pages

lslezak:
## Problem

- The GitHub Pages are not published automatically, you have to run the workflow manually

## Solution

- According to [this StackOverflow comment](https://stackoverflow.com/a/65723433) the `$default-branch` can be used only in workflow templates for generating the workflows, but not in the final workflows
- Let's try it and see if that helps...


mvidner (approved):
Makes sense, thanks for the catch!

PR #571, opened 2023-05-10
Add a Network service

imobachgs:
## Problem

Agama does not offer a service to set up the network. The web UI talks directly to NetworkManager and the CLI/autoinstallation does not support configuring the network.

## Solution

Trello card: https://trello.com/c/1kiDu6yN/ and https://trello.com/c/TCJqEQ2z/

This PR introduces a new service that allows configuring the network. The code is organized into these modules:

* `model` contains the types that represent the network concepts. They are supposed to be agnostic from the actual network service (e.g., NetworkManager).
* `adapter` offers an abstraction to support other backends in the future (if needed).
* `dbus` includes the D-Bus service implementation.
* `nm` implements support for interacting with NetworkManager.

<details>
<summary>Example of a network D-Bus tree</summary>

```
$ sudo busctl --address unix:path=/run/agama/bus tree org.opensuse.Agama.Network1
└─/org
  └─/org/opensuse
    └─/org/opensuse/Agama
      └─/org/opensuse/Agama/Network1
        ├─/org/opensuse/Agama/Network1/connections
        │ ├─/org/opensuse/Agama/Network1/connections/0
        │ ├─/org/opensuse/Agama/Network1/connections/1
        │ ├─/org/opensuse/Agama/Network1/connections/10
        │ ├─/org/opensuse/Agama/Network1/connections/2
        │ ├─/org/opensuse/Agama/Network1/connections/3
        │ ├─/org/opensuse/Agama/Network1/connections/4
        │ ├─/org/opensuse/Agama/Network1/connections/5
        │ ├─/org/opensuse/Agama/Network1/connections/6
        │ ├─/org/opensuse/Agama/Network1/connections/7
        │ ├─/org/opensuse/Agama/Network1/connections/8
        │ └─/org/opensuse/Agama/Network1/connections/9
        └─/org/opensuse/Agama/Network1/devices
          ├─/org/opensuse/Agama/Network1/devices/0
          ├─/org/opensuse/Agama/Network1/devices/1
          ├─/org/opensuse/Agama/Network1/devices/2
          ├─/org/opensuse/Agama/Network1/devices/3
          ├─/org/opensuse/Agama/Network1/devices/4
          └─/org/opensuse/Agama/Network1/devices/5
```
</summary>
</details>

<details>
<summary>Exposing a wireless network</summary>

```
$ sudo busctl --address unix:path=/run/agama/bus introspect org.opensuse.Agama.Network1 /org/opensuse/Agama/Network1/connections/4
NAME                                            TYPE      SIGNATURE RESULT/VALUE                                                   FLAGS
org.freedesktop.DBus.Introspectable             interface -         -                                                              -
.Introspect                                     method    -         s                                                              -
org.freedesktop.DBus.Peer                       interface -         -                                                              -
.GetMachineId                                   method    -         s                                                              -
.Ping                                           method    -         -                                                              -
org.freedesktop.DBus.Properties                 interface -         -                                                              -
.Get                                            method    ss        v                                                              -
.GetAll                                         method    s         a{sv}                                                          -
.Set                                            method    ssv       -                                                              -
.PropertiesChanged                              signal    sa{sv}as  -                                                              -
org.opensuse.Agama.Network1.Connection          interface -         -                                                              -
.Id                                             property  s         "Vodafone_4960_7DRS"                                           emits-change
.UUID                                           property  s         "889e7b38-82ed-42d4-bee2-5d10ba9c706b"                         emits-change
org.opensuse.Agama.Network1.Connection.IPv4     interface -         -                                                              -
.Addresses                                      property  a(su)     1 "192.168.0.99" 24                                            emits-change writable
.Gateway                                        property  s         "192.168.0.1"                                                  emits-change writable
.Method                                         property  y         1                                                              emits-change writable
.Nameservers                                    property  as        2 "6.6.6.6" "8.8.8.8"                                          emits-change writable
org.opensuse.Agama.Network1.Connection.Wireless interface -         -                                                              -
.Mode                                           property  y         0                                                              emits-change writable
.Password                                       property  s         ""                                                             emits-change writable
.SSID                                           property  ay        18 86 111 100 97 102 111 110 101 95 52 57 54 48 95 55 68 82 83 emits-change writable
.Security                                       property  y         3                                                              emits-change

```
</details>

## Documentation

This PR updates the [network_support_design.md document](https://github.com/openSUSE/agama/blob/network-service-2/doc/network_support_design.md#the-proposal), as we decided not to rely on third-party tools (that's the reason the branch is called `network-service-2` 😅).

In addition to that, we tried to document the code to make it easier to contribute.

<details>
<summary>Library documentation</summary>

Captura desde 2023-05-24 10-27-19
</details>

## Testing

As with most of our Rust code, it contains a few unit tests. I will try to add more before merging.

## Why is it still in draft mode?

- [x] Reload after saving (to get the correct UUIDs and clean-up removed devices)
- [x] Extend the documentation
- [x] Adapt to `agama-dbus-server` package (see #533).
- [ ] (optional) Report which connections failed to add/update/remove (some `Report` struct).
- [ ] (optional) Better error handling (using anyhow).

## What's next?

* Add support to the CLI to configure the network: https://trello.com/c/MowTUSWi/


rust/agama-dbus-server/src/locale.rs
```diff
@@ -170,30 +170,22 @@ impl Locale {
...
         const SERVICE_NAME: &str = "org.opensuse.Agama.Locale1";
         const SERVICE_PATH: &str = "/org/opensuse/Agama/Locale1";
 
         // First connect to the Agama bus, then serve our API,
         // for better error reporting.
-        let conn = ConnectionBuilder::address(ADDRESS)?
-            .build()
-            .await
-            .context(format!("Connecting to the Agama bus at {ADDRESS}"))?;
+        let connection = connection_to(address).await?;
```
jreidinger, line 179: :+1:

rust/agama-dbus-server/src/locale.rs
```diff
@@ -170,30 +170,22 @@ impl Locale {
...
 
         // When serving, request the service name _after_ exposing the main object
         let locale = Locale::new();
-        conn
-            .object_server()
-            .at(SERVICE_PATH, locale)
-            .await?;
-        conn
+        connection.object_server().at(SERVICE_PATH, locale).await?;
```
jreidinger, line 183: I would probably mention there that so far network and locale is independent, but if some service that will depend on other one like e.g. manager is added, the ordering is important.

rust/agama-network/src/adapter.rs
```diff
@@ -0,0 +1,8 @@
+use crate::NetworkState;
+use std::error::Error;
+
+/// A trait for the ability to read/write from/to a network service
+pub trait Adapter {
+    fn read(&self) -> Result<NetworkState, Box<dyn Error>>;
+    fn write(&self, network: &NetworkState) -> Result<(), Box<dyn Error>>;
```
jreidinger, line 7: to be honest I am not big fan of that box dyn errors...lets hope that with anyerror it can be just anyError

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+
+#[dbus_interface(name = "org.opensuse.Agama.Network1.Devices")]
+impl Devices {
+    /// Returns the D-Bus paths of the network devices.
+    pub fn get_devices(&self) -> Vec<ObjectPath> {
+        let objects = self.objects.lock();
+        objects
+            .devices_paths()
+            .iter()
+            .filter_map(|c| ObjectPath::try_from(c.clone()).ok())
```
jreidinger, line 43: so if I get it, it silently ignore if  there is some wrong device paths? it should contain at least explanation why object can fail to be converted to object path

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+
+#[dbus_interface(name = "org.opensuse.Agama.Network1.Device")]
+impl Device {
+    #[dbus_interface(property)]
+    pub fn name(&self) -> &str {
+        &self.device.name
+    }
+
+    #[dbus_interface(property)]
+    pub fn device_type(&self) -> u8 {
```
jreidinger, line 72: Why not simple type? Device is already in interface name

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+        self.device.ty as u8
```
jreidinger, line 73: hmm, that `ty` looks like typo for me. Lets see when I reach NetworkDevice what I will think about it

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+            .connections_paths()
+            .iter()
+            .filter_map(|c| ObjectPath::try_from(c.clone()).ok())
+            .collect()
+    }
+
+    /// Adds a new network connection.
+    ///
+    /// * `name`: connection name.
+    /// * `ty`: connection type (see [crate::model::DeviceType]).
```
jreidinger, line 112: to be honest I do not like much that `ty`. Why not `type` or `ctype`? so it is more readable

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+///
+/// It offers an API to query a connection.
+pub struct Connection {
+    connection: Arc<Mutex<NetworkConnection>>,
+}
+
+impl Connection {
+    /// Creates a Connection interface object.
+    ///
+    /// * `UUID`: Connection UUID.
```
jreidinger, line 152: ugh, what?

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+/// D-Bus interface for IPv4 settings
+pub struct Ipv4 {
+    actions: Arc<Mutex<Sender<Action>>>,
+    connection: Arc<Mutex<NetworkConnection>>,
+}
+
+impl Ipv4 {
+    /// Creates an IPv4 interface object.
+    ///
+    /// * `actions`: sending-half of a channel to send actions.
```
jreidinger, line 185: I do not get what means `sending-half of a channel`

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+#[dbus_interface(name = "org.opensuse.Agama.Network1.Connection.IPv4")]
+impl Ipv4 {
+    #[dbus_interface(property)]
+    pub fn addresses(&self) -> Vec<(String, u32)> {
+        let connection = self.get_connection();
+        connection
+            .ipv4()
+            .addresses
+            .iter()
+            .map(|(addr, prefix)| (addr.to_string(), *prefix))
```
jreidinger, line 223: so no netmask support, right?

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    #[dbus_interface(property)]
+    pub fn set_nameservers(&mut self, addresses: Vec<String>) -> zbus::fdo::Result<()> {
+        let mut connection = self.get_connection();
+        let ipv4 = connection.ipv4_mut();
+        addresses
+            .iter()
+            .map(|addr| addr.parse::<Ipv4Addr>())
+            .collect::<Result<Vec<Ipv4Addr>, AddrParseError>>()
```
jreidinger, line 270: Yeah, tricky collect, but interesting code.

rust/agama-network/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,402 @@
...
+            None => "".to_string(),
+        }
+    }
+
+    #[dbus_interface(property)]
+    pub fn set_gateway(&mut self, gateway: String) -> zbus::fdo::Result<()> {
+        let mut connection = self.get_connection();
+        let ipv4 = connection.ipv4_mut();
+        gateway
+            .parse::<Ipv4Addr>()
```
jreidinger, line 290: if I get code, it means you cannot set `""` to clear gateway. Am I right? If so, how to unset it?

rust/agama-network/src/dbus/service.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Represents the Agama networking D-Bus service.
+///
+/// It is responsible for starting the [NetworkSystem] on a different thread.
+pub struct NetworkService;
+
+impl NetworkService {
+    /// Starts listening and dispatching events on the D-Bus connection.
+    pub async fn start_service(address: &str) -> Result<(), Box<dyn Error>> {
+        let connection = connection_to(address).await?;
+        let mut network = NetworkSystem::from_network_manager(connection.clone())
```
jreidinger, line 17: so this means that network manager has to use same DBus socket as agama one. Is the restriction needed? I though that for containerized one we have NM on different bus.

rust/agama-network/src/dbus/service.rs
```diff
@@ -0,0 +1,37 @@
...
+        thread::spawn(move || {
+            async_std::task::block_on(async {
+                network
+                    .setup()
+                    .await
+                    .expect("Could not set up the D-Bus tree");
+                connection
+                    .request_name("org.opensuse.Agama.Network1")
+                    .await
+                    .expect("Could not get the 'org.opensuse.Agama.Network1' service");
```
jreidinger, line 30: this is kind of misleading. IT is more like that cannot get name for service and not service itself.

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+            actions,
+            objects: Default::default(),
+        }
+    }
+
+    /// Adds devices to the D-Bus tree.
+    ///
+    /// * `devices`: list of devices.
+    pub async fn add_devices(&mut self, devices: &Vec<Device>) -> Result<(), ServiceError> {
+        for (i, dev) in devices.iter().enumerate() {
```
jreidinger, line 37: this is kind of strange or I do not get logic, but if I call this method twice, then result is it that later ones rewrite ones from first call?

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(())
+    }
+
+    /// Adds a connection to the D-Bus tree.
+    ///
+    /// * `connection`: connection to add.
+    pub async fn add_connection(&self, conn: &Connection) -> Result<(), ServiceError> {
+        let mut objects = self.objects.lock();
+
+        let path = format!("{}/{}", CONNECTIONS_PATH, objects.connections.len());
```
jreidinger, line 77: this is interesting. So I failed to see renumbering  in remove_connection ( it can be of course my fault :wink: ), so what happen if there is add + add + removal first + add? Does the last add has same connection number as add2?

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+        let cloned = Arc::new(Mutex::new(conn.clone()));
+        self.add_interface(&path, interfaces::Connection::new(Arc::clone(&cloned)))
+            .await?;
+
+        self.add_interface(
+            &path,
+            interfaces::Ipv4::new(self.actions.clone(), Arc::clone(&cloned)),
```
jreidinger, line 84: so no IPv6 support yet?

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /// Removes a connection from the tree
+    ///
+    /// * `uuid`: UUID of the connection to remove.
+    pub async fn remove_connection(&mut self, uuid: Uuid) -> Result<(), ServiceError> {
+        // TODO: better error
+        let mut objects = self.objects.lock();
+        let Some(path) = objects.connection_path(uuid) else {
+            return Ok(())
```
jreidinger, line 107: I expect that that TODO should be here?

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Deregisters a network connection.
+    ///
+    /// * `uuid`: connection UUID.
+    pub fn deregister_connection(&mut self, uuid: Uuid) -> Option<String> {
+        self.connections.remove(&uuid)
+    }
+
+    /// Returns all devices paths.
+    pub fn devices_paths(&self) -> Vec<String> {
+        self.devices.values().map(|p| p.to_string()).collect()
```
jreidinger, line 199: just matter of taste, but as values is already String, I usually prefer `to_owned` over `to_string` as it is not type conversion.

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /// Returns all connection paths.
+    pub fn connections_paths(&self) -> Vec<String> {
+        self.connections.values().map(|p| p.to_string()).collect()
+    }
+}
```
jreidinger, line 206: note for me. End of review for today.

rust/agama-dbus-server/src/main.rs
```diff
@@ -1,11 +1,15 @@
...
+use agama_network::NetworkService;
 use std::future::pending;
 
+const ADDRESS: &str = "unix:path=/run/agama/bus";
+
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let _con = crate::locale::Locale::start_service().await?;
+    NetworkService::start_service(ADDRESS).await?;
+    let _con = crate::locale::Locale::start_service(ADDRESS).await?;
```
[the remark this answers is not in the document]
imobachgs: Well, the difference is that in the case of `NetworkService`, we keep a copy of the connection, so we do not need to return it back not to be destroyed. I am not sure whether I prefer, but the `let _con` sounds artificial to me.

rust/agama-network/src/action.rs
```diff
@@ -0,0 +1,18 @@
...
+/// These actions are meant to be processed by [crate::system::NetworkSystem], updating the model
+/// and the D-Bus tree as needed.
+#[derive(Debug)]
+pub enum Action {
+    /// Add a new connection with the given name and type.
+    AddConnection(String, DeviceType),
+    /// Update a connection (replacing the old one).
+    UpdateConnection(Connection),
+    /// Remove the connection with the given Uuid.
+    RemoveConnection(Uuid),
```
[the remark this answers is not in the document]
imobachgs: These actions follow an approach similar to Redux actions. They come from the D-Bus API and they are processed by the [NetworkSystem](https://github.com/openSUSE/agama/blob/0d1defaaa29429dff9896d3c803bdcedbd7c0ff8/rust/agama-network/src/system.rs#L75) that acts as the source of truth. They basically include just the relevant data:

* AddConnection: the NetworkSystem is responsible for creating the connection and it only needs a name and a type.
* UpdateConnection: the D-Bus API already holds a copy of the connection, so it just send the updated version.
* RemoveConnection: NetworkSystem just needs the uuid to remove the connection.

As the D-Bus API already holds a copy of the connection (except in the first case), it would not be a problem to send the whole object if you prefer.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: Yes, good point, we should report that somehow.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: Because `type` is a reserved word in Rust. I would rename the D-Bus name but I would keep the if the function (unless we find a better name for all *types*).

rust/agama-network/src/dbus/interfaces.rs
imobachgs: `type` is a reserved word, but I am open to find a better name 😅. Why `ctype`?

rust/agama-network/src/dbus/interfaces.rs
imobachgs: I got used to that name when working with [syn](https://docs.rs/syn/latest/syn/) (see [ty](https://docs.rs/syn/latest/syn/struct.Field.html#structfield.ty).

rust/agama-network/src/dbus/interfaces.rs
imobachgs: Oh, a leftover.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: You use an address and a prefix ("192.168.1.0/32"), so no need for a network mask.

rust/agama-network/src/dbus/service.rs
imobachgs: Actually no, it uses a different connection (if you [check the internals](https://github.com/openSUSE/agama/blob/0d1defaaa29429dff9896d3c803bdcedbd7c0ff8/rust/agama-network/src/system.rs#L33-L35)). But it is true that it is not clear from this call. I could try to inject the adapter or to improve the docs to make it clear.

rust/agama-network/src/dbus/tree_manager.rs
```diff
@@ -0,0 +1,164 @@
...
+impl Tree {
+    pub fn new(connection: zbus::Connection, network: Arc<Mutex<NetworkSystem>>) -> Self {
+        Self {
+            connection,
+            network,
+            objects: Default::default(),
+        }
+    }
+
+    pub async fn populate(&mut self) -> Result<(), ServiceError> {
```
jreidinger, line 25: I think here it would be useful to have documentation for this public method  as I kind of feel there will be some limitation

rust/agama-dbus-server/src/main.rs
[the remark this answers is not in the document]
jreidinger: for me it sounds a bit artificial, so we can keep it also in Locale. Feel free to change it to make it consistent.

rust/agama-network/src/action.rs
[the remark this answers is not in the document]
jreidinger: @imobachgs well, I just think about possible different backend that do not use uuid for connection identification, so seding whole connection is more generic from my POV. ( I expect that for different backend we will probably also need to change params for Add but it can be solvable later with some factory, but we should not add more limitation than necessary ).

rust/agama-network/src/dbus/interfaces.rs
jreidinger: fine for me. I am talking here mainly about dbus API

rust/agama-network/src/dbus/interfaces.rs
jreidinger: well, if it is established rust acronym, it is OK for me as I do not know it before.

rust/agama-network/src/dbus/interfaces.rs
jreidinger: ctype as connection type acronym...btw my suggestion is then to use `type_` everywhere. It is clear that it is type and another underscore does not hurt readability..

rust/agama-network/src/dbus/tree_manager.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn unregister_device(&mut self, name: &str) -> Option<String> {
+        self.devices.remove(name)
+    }
+
+    pub fn unregister_connection(&mut self, uuid: Uuid) -> Option<String> {
+        self.connections.remove(&uuid)
+    }
+
+    pub fn devices_paths(&self) -> Vec<String> {
+        self.devices.values().map(|p| p.to_string()).collect()
```
jreidinger, line 158: to be honest this looks quite inconsistent for me. connection path return `&str` and connection_paths `String`. I think it should ideally unify on one.

rust/agama-network/src/dbus/tree_manager.rs
```diff
@@ -0,0 +1,164 @@
...
+use uuid::Uuid;
+
+use crate::{dbus::interfaces, model::*, NetworkSystem};
+use std::collections::HashMap;
+use std::sync::Arc;
+
+/// Handle the objects in the D-Bus tree for the network state
+pub struct Tree {
+    connection: zbus::Connection,
+    network: Arc<Mutex<NetworkSystem>>,
```
mvidner, line 12: Ugh, I will need some intro to synchronization techniques, https://docs.rs/parking_lot/0.12.1/parking_lot/type.Mutex.html does not help much

imobachgs: I would recommend having a look to:

* [Shared-state concurrency](https://doc.rust-lang.org/book/ch16-03-shared-state.html) as it is the main reason to use mutexes/channels.
* The documentation for the [Mutex from the std](https://doc.rust-lang.org/std/sync/struct.Mutex.html).
* The documentation about [channels](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html).

rust/agama-network/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+//! `Apply` method to write the changes to the NetworkManager service.
+//!
+//! ## Limitations
+//!
+//! We expect to address the following problems as we evolve the API, but it is noteworthy to have
+//! them in mind:
+//!
+//! * The devices list does not reflect the changes in the system. For instance, it is not updated
+//! when a device is connected to the system.
+//! * Many configuration types are still missing (bridges, bonding, etc.).
```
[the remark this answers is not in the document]
imobachgs: Well, it explains the overall approach, but we should have some proper D-Bus documentation including methods, properties and so on.

rust/agama-network/src/dbus/tree_manager.rs
imobachgs: Actually, both of them should return `OwnedObjectPath` objects. I will adapt it.

rust/agama-network/src/action.rs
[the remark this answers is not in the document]
imobachgs: Actually, the point is that `Connection` represents kind of a generic struct, no matter which backend you use. Each adapter should return a `Connection`.

In an initial version I had a `NmConnection`, but it was simply too complex for our current scenario.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: Well, at least it is used in `syn`. I can do some research if needed.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: [sending half](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html) and [receiving half](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html) are the the two parts of a channel.

rust/agama-network/src/dbus/interfaces.rs
imobachgs: Good point. I have fixed it.

rust/agama-network/src/dbus/tree.rs
imobachgs: Yes, you are fully right. I have changed the API to prevent such situations. Actually, as I mentioned in a TODO, we should improve the mechanism to "refresh" the list of connections and devices (avoiding the brute-force approach).

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,206 @@
...
+    async fn remove_connections(&self) -> Result<(), ServiceError> {
+        let mut objects = self.objects.lock();
+        for path in objects.connections.values() {
+            self.remove_connection_on(path.as_str()).await?;
+        }
+        objects.connections.clear();
+        Ok(())
+    }
+
+    /// Refreshes the list of connections.
```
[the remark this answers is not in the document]
imobachgs: OK, let's change it to set, it makes sense.

rust/agama-network/src/dbus/tree_manager.rs
imobachgs: Ouch, that was an old version of the `tree`. I have removed the file.

rust/agama-network/src/dbus/tree.rs
imobachgs: GitHub has a nice feature that allows you to mark a file as already "viewed".

gh-marked-as-viewed

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    /// Get connection by UUID
+    ///
+    /// * `uuid`: connection UUID
+    pub fn get_connection(&self, uuid: Uuid) -> Option<&Connection> {
+        self.connections.iter().find(|c| c.uuid() == uuid)
+    }
+
+    /// Get connection by UUID as mutable
```
jreidinger, line 42: this is kind of inconsistent as get_device_mut missing

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+    pub name: String,
+    pub ty: DeviceType,
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum DeviceType {
+    Loopback = 0,
+    Ethernet = 1,
+    Wireless = 2,
+    Unknown = 3,
```
jreidinger, line 192: I do not like this ordering as it means that if we add new type, it result in change of unknown value or having unknown as 3 forever with other having higher number. So I check https://dbus.freedesktop.org/doc/dbus-api-design.html and it also shows having unknown as 0 and not high number. So I suggest to have Unknown 0 and all known ones as +1

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+        let base = BaseConnection {
+            id: id.to_string(),
+            ..Default::default()
+        };
+        match device_type {
+            DeviceType::Wireless => Connection::Wireless(WirelessConnection {
+                base,
+                ..Default::default()
+            }),
+            _ => Connection::Ethernet(EthernetConnection { base }),
```
jreidinger, line 228: I have to be blind, but if you create new device as loopback it will create ethernet, right? I think it should at least report error.

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+
+    pub fn base(&self) -> &BaseConnection {
+        match &self {
```
jreidinger, line 233: is it needed? arg there easier way to have method in this kind of inheritance? Or it needs macros or traits to do it?

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+    pub fn ipv4_mut(&mut self) -> &mut Ipv4Config {
+        &mut self.base_mut().ipv4
+    }
+
+    pub fn remove(&mut self) {
+        self.base_mut().status = Status::Missing;
+    }
+
+    pub fn is_missing(&self) -> bool {
+        self.base().status == Status::Missing
```
jreidinger, line 269: this is kind of inconsistent with action and is_action being different. What about `dismiss` and `is_missing` or `remove` and `is_removed`?

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+    pub nameservers: Vec<Ipv4Addr>,
+    pub gateway: Option<Ipv4Addr>,
+}
+
+#[derive(Debug, Default, Copy, Clone, PartialEq)]
+pub enum IpMethod {
+    #[default]
+    Auto = 0,
+    Manual = 1,
+    Unknown = 2,
```
jreidinger, line 307: Same here. I would use Unknown as 0, so if anything new appears it does not need renumber.

rust/agama-network/src/model.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+impl fmt::Display for IpMethod {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name = match &self {
+            IpMethod::Auto => "auto",
+            IpMethod::Manual => "manual",
+            IpMethod::Unknown => "auto",
+        };
+        write!(f, "{}", name)
```
jreidinger, line 316: why not `f.write_str(name)` ?

rust/agama-network/src/model.rs
imobachgs: It is missing because it is not needed. We do not allow changing anything for the devices yet. Once we implement support for that, we will add that method.

rust/agama-network/src/model.rs
imobachgs: I introduce the loopback devices in the very last minute and I forgot to fix this function.

rust/agama-network/src/model.rs
imobachgs: I got that from [the official docs](https://doc.rust-lang.org/std/fmt/trait.Display.html#examples).

rust/agama-network/src/model.rs
imobachgs: Finally, I have dropped all the "Unknown" variants and moved that check to the NetworkManager handling code. If agama-network does not understand something, it is not exposed.

In the future, we could expose it as a kind of read-only generic connections/devices. WDYT?

rust/agama-network/src/model.rs
imobachgs: Having said that, I do not have a strong opinion.

rust/agama-network/src/model.rs
imobachgs: Sure, I already thought about it. But I prefer not to add more complexity to this PR (adding a macro and so on). I will add a TODO so we can implement that in the future. It might be a nice exercise 😃

rust/agama-network/src/model.rs
jreidinger: with is_action I mean action like action => add so method `add` and `is_added`. So inconsistency for me is different name of actions. is_removed sounds good to me.

rust/agama-network/src/model.rs
jreidinger: yeap. I think when we start adding more connection types, it will be needed to make it as easy as possible without touching too man placing. TODO is fine for me.

rust/agama-network/src/dbus/tree.rs
jreidinger: thanks, will try. I usually manage to do review in one day, but this one is quite big and also needs a bit of rust study from me.

rust/agama-network/src/dbus/interfaces.rs
jreidinger: yeah, after reading about channels it makes perfect sense...I just was not familiar (or have association ) with term channel

rust/agama-network/src/model.rs
jreidinger: well, I expect that rust will optimize it out, just I see no reason for formatting, but also not have strong opinion and if it is common approach in implementing Format trait, it is ok for me.

rust/agama-lib/src/proxies.rs
```diff
@@ -1,18 +1,6 @@
...
-//!
-//! These DBus objects implements
-//! [standard DBus interfaces](https://dbus.freedesktop.org/doc/dbus-specification.html),
-//! (`org.freedesktop.DBus.*`) for which the following zbus proxies can be used:
-//!
-//! * [`zbus::fdo::PropertiesProxy`]
-//! * [`zbus::fdo::IntrospectableProxy`]
-//!
-//! …consequently `zbus-xmlgen` did not generate code for the above interfaces.
-//! Also some proxies can be used against multiple services when they share interface.
```
jreidinger, line 14: what is reason to remove these notes? I actually find it quite useful when I look for some proxies.

rust/agama-network/src/action.rs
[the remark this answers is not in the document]
jreidinger: ok, so lets keep it as it is. I think PR is already big enough :)

rust/agama-network/src/nm/adapter.rs
```diff
@@ -0,0 +1,62 @@
...
+                    if let Err(e) = self.client.remove_connection(conn.uuid()).await {
+                        eprintln!("Could not remove the connection {}: {}", conn.uuid(), e);
+                    }
+                } else {
+                    if let Err(e) = self.client.add_or_update_connection(conn).await {
+                        eprintln!("Could not add/update the connection {}: {}", conn.uuid(), e);
+                    }
+                }
+            }
+        });
```
jreidinger, line 59: Do I get it correctly that for now all write succeed? I think this deserve at least fixme.

rust/agama-network/src/nm/client.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        Ok(devs)
+    }
+
+    /// Returns the list of network connections.
+    pub async fn connections(&self) -> Result<Vec<Connection>, ServiceError> {
+        let proxy = SettingsProxy::new(&self.connection).await?;
+        let paths = proxy.list_connections().await?;
+        let mut connections: Vec<Connection> = vec![];
```
jreidinger, line 66: pro efficiency hint 
```suggestion
        let mut connections: Vec<Connection> = Vec::with_capacity(paths.len());
```

rust/agama-network/src/nm/client.rs
```diff
@@ -0,0 +1,114 @@
...
+        for path in paths {
+            let proxy = ConnectionProxy::builder(&self.connection)
```
jreidinger, line 68: NP: I just noticed it here, but having connection in NM and dbus connection using same name can cause confusion.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,443 @@
...
+
+const ETHERNET_KEY: &str = "802-3-ethernet";
+const WIRELESS_KEY: &str = "802-11-wireless";
+const WIRELESS_SECURITY_KEY: &str = "802-11-wireless-security";
+const LOOPBACK_KEY: &str = "loopback";
+
+/// Nested hash to send to D-Bus.
+type NestedHash<'a> = HashMap<&'a str, HashMap<&'a str, zvariant::Value<'a>>>;
+/// Nested hash as it comes from D-Bus.
+type OwnedNestedHash = HashMap<String, HashMap<String, zvariant::OwnedValue>>;
```
jreidinger, line 20: btw I think this types can be useful even in agama lib as we use hashes also CLI and I fight with it a bit.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,443 @@
...
+/// Converts a connection struct into a HashMap that can be sent over D-Bus.
+///
+/// * `conn`: Connection to cnvert.
+pub fn connection_to_dbus(conn: &Connection) -> NestedHash {
+    let mut result = NestedHash::new();
+    let mut connection_dbus =
+        HashMap::from([("id", conn.id().into()), ("type", ETHERNET_KEY.into())]);
+    result.insert("ipv4", ipv4_to_dbus(conn.ipv4()));
+
+    if let Connection::Wireless(wireless) = conn {
```
jreidinger, line 31: something for future but as more types will be added, I think we will need to somehow have type specific conversion as I expect each new type will need something specific.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,443 @@
...
+        return Some(Connection::Ethernet(EthernetConnection { base }));
+    };
+
+    None
+}
+
+/// Merges a NestedHash and an OwnedNestedHash connections.
+///
+/// * `original`: original hash coming from D-Bus.
+/// * `updated`: updatde hash to write to D-Bus.
```
jreidinger, line 69: typo `updatde`

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,443 @@
...
+    original: &'a OwnedNestedHash,
+    updated: &'a NestedHash,
+) -> NestedHash<'a> {
+    let mut merged = HashMap::new();
+    for (key, orig_section) in original {
+        let mut inner: HashMap<&str, zbus::zvariant::Value> = HashMap::new();
+        for (inner_key, value) in orig_section {
+            inner.insert(inner_key.as_str(), value.into());
+        }
+        if let Some(upd_section) = updated.get(key.as_str()) {
```
jreidinger, line 80: so this means that key that is not in original cannot be in updated. Can it happen? it should be at least documented, that it cannot happen.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,443 @@
...
+///
+/// * `original`: original hash coming from D-Bus.
+/// * `updated`: updatde hash to write to D-Bus.
+pub fn merge_dbus_connections<'a>(
+    original: &'a OwnedNestedHash,
+    updated: &'a NestedHash,
+) -> NestedHash<'a> {
+    let mut merged = HashMap::new();
+    for (key, orig_section) in original {
+        let mut inner: HashMap<&str, zbus::zvariant::Value> = HashMap::new();
```
jreidinger, line 76: as it is O^2 complexity, I would at least try to reduce reallocation with usng for both HashMap `::with_capacity` set to original, respective original_section which should eliminaty expensive reallacation.

rust/agama-lib/src/proxies.rs
imobachgs: Well, it is exactly the same text copied over and over. I can bring it back anyway.

rust/agama-network/src/nm/client.rs
imobachgs: > 

Yes, you are right, and I did not want to use `ConnectionIface` or something like that (it is redundant). I would keep it as it is and we can improve it later.

imobachgs: Yes, you are right, and I did not want to use ConnectionIface or something like that (it is redundant). I would keep it as it is and we can improve it later.

rust/agama-network/src/nm/dbus.rs
imobachgs: AFAIK, it cannot happen. I will document it just in case.

rust/agama-network/src/model.rs
imobachgs: I do not know how common it is, just what I saw in the docs. Perhaps we will learn something different tomorrow 😄

rust/agama-lib/src/dbus.rs
```diff
@@ -0,0 +1,7 @@
+use std::collections::HashMap;
+use zbus::zvariant;
+
+/// Nested hash to send to D-Bus.
+pub type NestedHash<'a> = HashMap<&'a str, HashMap<&'a str, zvariant::Value<'a>>>;
+/// Nested hash as it comes from D-Bus.
+pub type OwnedNestedHash = HashMap<String, HashMap<String, zvariant::OwnedValue>>;
```
jreidinger, line 7: :+1: thanks

rust/agama-network/src/nm/dbus.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    fn check_dbus_base_connection(conn_dbus: &NestedHash) {
+        let connection_dbus = conn_dbus.get("connection").unwrap();
+        let id: &str = connection_dbus.get("id").unwrap().downcast_ref().unwrap();
+        assert_eq!(id, "agama");
+
+        let ipv4_dbus = conn_dbus.get("ipv4").unwrap();
+        let gateway: &str = ipv4_dbus.get("gateway").unwrap().downcast_ref().unwrap();
+        assert_eq!(gateway, "192.168.0.1");
+    }
```
jreidinger, line 441: :+1: for having tests

rust/agama-network/src/nm/model.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+impl From<NmWirelessMode> for WirelessMode {
+    fn from(value: NmWirelessMode) -> Self {
+        match value.as_str() {
+            "infrastructure" => WirelessMode::Infra,
+            "adhoc" => WirelessMode::AdHoc,
+            "mesh" => WirelessMode::Mesh,
+            "ap" => WirelessMode::AP,
+            _ => WirelessMode::Other,
```
jreidinger, line 43: hmm, so other acts like `unknown` here? I see same issue here. If we add another mode, we will need to renumber other or other won't be the last in enum. So again suggest to place it at first or somehow skip it.

rust/agama-network/src/nm/model.rs
```diff
@@ -0,0 +1,161 @@
...
+impl From<NmKeyManagement> for SecurityProtocol {
+    fn from(value: NmKeyManagement) -> Self {
+        match value.as_str() {
+            "owe" => SecurityProtocol::OWE,
+            "ieee8021x" => SecurityProtocol::DynamicWEP,
+            "wpa-psk" => SecurityProtocol::WPA2,
+            "wpa-eap" => SecurityProtocol::WPA3Personal,
+            "sae" => SecurityProtocol::WPA2Enterprise,
+            "wpa-eap-suite-b192" => SecurityProtocol::WPA2Enterprise,
+            _ => SecurityProtocol::WEP,
```
jreidinger, line 109: so wep does not have own string? It kind of smells for me. I expect that "wep" has something and report error for something we do not know.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -71,7 +71,8 @@ pub fn merge_dbus_connections<'a>(
 ) -> NestedHash<'a> {
     let mut merged = HashMap::new();
```
jreidinger, line 72: also here it can use capacity
```suggestion
    let mut merged = HashMap::with_capacity(original.len());
```

rust/agama-network/src/nm/model.rs
imobachgs: Yes, you are right. I will proceed in the same way that with the rest.

rust/agama-network/src/nm/model.rs
```diff
@@ -33,18 +33,26 @@ impl NmWirelessMode {
...
         match value.as_str() {
-            "infrastructure" => WirelessMode::Infra,
-            "adhoc" => WirelessMode::AdHoc,
-            "mesh" => WirelessMode::Mesh,
-            "ap" => WirelessMode::AP,
-            _ => WirelessMode::Other,
+            "infrastructure" => Ok(WirelessMode::Infra),
+            "adhoc" => Ok(WirelessMode::AdHoc),
+            "mesh" => Ok(WirelessMode::Mesh),
+            "ap" => Ok(WirelessMode::AP),
```
jreidinger, line 44: I miss in this commit also removal of `Other` from `WirelessMode` enum. ( or is there reason to keep it? )

imobachgs: Right, I forgot to submit such a change. Fixed now.

mvidner (commented):
I'm being extra pedantic about API naming consistency here...

The aspects to observe are
- pluralization
- capitalization

For **pluralization**, it makes sense to use the plural name whenever there is a collection: `/.../things` and `/.../things/1`, `/.../things/33` (as opposed to the singular `/.../thing/33`)

For **capitalization**,
- the [Naming Conventions](https://dbus.freedesktop.org/doc/dbus-api-design.html#naming-conventions) want to use CamelCase even for objects
- here we are following the systemd convention of naming the objects in lowercase
- other parts of Agama use CamelCase for object names

IMHO we should use CamelCase
(but before reaching this conclusion I have adjusted the docs to document the code reality)

rust/agama-network/src/nm/client.rs
```diff
@@ -0,0 +1,114 @@
...
+
+            let device_type = NmDeviceType(proxy.device_type().await?);
+            if let Ok(device_type) = device_type.try_into() {
+                devs.push(Device {
+                    name: proxy.interface().await?,
+                    ty: device_type,
+                });
+            } else {
+                // TODO: use a logger
+                eprintln!("Unknown device type {:?}", &device_type);
```
mvidner, line 55: On my workstation, I get

> Unknown device type NmDeviceType(14)
> Unknown device type NmDeviceType(13)
> Unknown device type NmDeviceType(13)
> Unknown device type NmDeviceType(13)
> Unknown device type NmDeviceType(13)

so... we may also want to say _which_ device it is?

rust/agama-network/src/dbus/interfaces.rs
mvidner: In a strongly typed language I would expect the object paths to be always valid, unless coming as user input or as arguments to a constructor, which is not the case here.

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    /// Deregisters a network connection.
+    ///
+    /// * `uuid`: connection UUID.
+    pub fn deregister_connection(&mut self, uuid: Uuid) -> Option<String> {
+        self.connections.remove(&uuid)
+    }
+
+    /// Returns all devices paths.
+    pub fn devices_paths(&self) -> Vec<String> {
```
mvidner, line 221: We should be using specialized strings that are guaranteed to be valid object paths.
The [zbus_names](https://docs.rs/zbus_names/2.5.0/zbus_names/index.html) crate exists exactly for this purpose and... it does not include object paths... because they appear on the wire so [the type lives in `zvariant`](https://docs.rs/zvariant/3.12.0/zvariant/struct.ObjectPath.html).

So,
```rs
use zvariant::ObjectPath;
    pub fn devices_paths(&self) -> Vec<OwnedObjectPath> {
or
    pub fn devices_paths(&self) -> Vec<ObjectPath> {
```

here and in many other places.
With respect to the D-Bus API it is an implementation detail, so we can leave it for later.

mvidner (commented):
Strings vs ObjectPaths:

mvidner:
> For **pluralization**, it makes sense to use the plural name whenever there is a collection: `/.../things` and `/.../things/1`, `/.../things/33` (as opposed to the singular `/.../thing/33`)

BTW I don't have a source for this, it is "original research" :smile: and the reasoning is that the tree is more compact when we use the same word forms for both the "manager" and the "items"

jreidinger (changes requested):
OK, so now after updates only important stuff for me is DBus naming ( consistency and that ty as property ). And in general I would like to see `type_` instead of `ty` as mvidner agrees with me :) Others can wait for next PR

imobachgs:
> IMHO we should use CamelCase (but before reaching this conclusion I have adjusted the docs to document the code reality)

I would say we are using systemd capitalization style in new code (basically in storage). We could re-evaluate that decision but, at least, we should document why we went that road (and plan for adopting it in the rest of the APIs).

imobachgs:
> OK, so now after updates only important stuff for me is DBus naming ( consistency and that ty as property ). And in general I would like to see `type_` instead of `ty` as mvidner agrees with me :) Others can wait for next PR

OK, I still do not like `type_`. To be clear, the property is [exposed as `Type`](https://github.com/openSUSE/agama/blob/e23a865016ddc92e82a3586fd0ff6279e3309a50/rust/agama-network/src/dbus/interfaces.rs#L71) over D-Bus, so it remains as an internal detail.

I have found more usages of `ty` in the Rust world:

* [The `ty` module: representing types](https://rustc-dev-guide.rust-lang.org/ty.html) in the Compiler Development Guide.
* Many fields in the [syn crate](https://docs.rs/crate/syn/latest), which is widely used to develop derive macros (e.g., https://docs.rs/syn/latest/syn/struct.ItemType.html#structfield.ty).
* Other crates (you can find a few [on this list](https://docs.rs/releases/search?query=ty)) although I do not know how relevant they are.

However, I can understand that it looks weird, so I have renamed it by now to unblock this PR.

rust/agama-network/src/dbus/tree.rs
imobachgs: I agree.

jreidinger (approved):

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,229 @@
...
+    {
+        let object_server = self.connection.object_server();
+        Ok(object_server.at(path.clone(), iface).await?)
+    }
+}
+
+/// Objects paths for known devices and connections
+#[derive(Debug, Default)]
+pub struct ObjectsRegistry {
+    pub devices: HashMap<String, String>,
```
mvidner, line 185:
```suggestion
    /// device_name (eth0) -> object_path
    pub devices: HashMap<String, String>,
```

rust/agama-network/src/dbus/tree.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub connections: HashMap<Uuid, String>,
```
mvidner, line 186:
```suggestion
    /// uuid -> object_path
    pub connections: HashMap<Uuid, String>,
```

rust/agama-network/src/dbus/tree.rs
mvidner, line 185: also, wouldn't zvariant::OwnedObjectPath be a more appropriate type?

rust/agama-network/src/action.rs
```diff
@@ -0,0 +1,18 @@
+use crate::model::{Connection, DeviceType};
+use uuid::Uuid;
+
+/// Networking actions, like adding, updating or removing connections.
+///
+/// These actions are meant to be processed by [crate::system::NetworkSystem], updating the model
+/// and the D-Bus tree as needed.
+#[derive(Debug)]
+pub enum Action {
```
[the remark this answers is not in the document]
imobachgs: This enum models the actions to update the model. By now, they come from the D-Bus interface, but in the future, they could be originated in NetworkManager itself (e.g., changing the state).

If you are curious, it is inspired by [Redux](https://redux.js.org/tutorials/essentials/part-1-overview-concepts).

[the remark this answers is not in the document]
imobachgs: I have added a TODO to not forget about it.

jreidinger (approved):

rust/agama-network/src/dbus/tree.rs
imobachgs: I have added a TODO to not forget about it.

mvidner: > you must also deal with invalid paths using that type.

Yes, but only once, which is the point. But it's fine to postpone it

mvidner (changes requested):
Thank you!

Just a .changes entry is missing I think

imobachgs:
> Thank you!
> 
> Just a .changes entry is missing I think

Oops, good point. I will add it.

mvidner (approved):
Yay!

Issue #565, opened 2023-05-08
Create a hook for working with siblings elements


dgdavid:
In #563 we introduced an internal function to work with siblings in the core/Sidebar component. It works with the DOM API directly and it works. However, would be nice to extract it to a hook, which is a bit more _React alike_ and will allow (probably among other benefits):

* Better reusability
* Better testing
* A central place for implementing a logic to be polite and restore the previous attributes values if they were present instead of simply removing them.

balsa-asanovic:
Hey there, can I try to work on this?

imobachgs:
Hi, @balsa-asanovic. Sure, you can work on this one if you wish. Thanks in advance!

balsa-asanovic:
I am trying to run the project using ./setup.sh script, but I am running into problems.

Could you help me? @imobachgs 

Here's the screenshot:

![image](https://github.com/openSUSE/agama/assets/112288843/f298abfe-e9df-44e0-8617-9e307aa27f68)

imobachgs:
Hi @balsa-asanovic,

Let's try. It looks like the D-Bus service is not running. Did you get any errors when running the `setup.sh` script? You could try to run the service from sources by typing:

```sh
cd service
sudo bundle exec bin/agama
```

It will take a few seconds to boot.

balsa-asanovic:
Hey @imobachgs,

Thanks for the help.

I don't actually get any errors when running `setup.sh` script.

However, when running `sudo bundle exec bin/agamactl`, some issues do pop-up:

![image](https://github.com/openSUSE/agama/assets/112288843/c311c1b2-23df-42d6-b189-76c6b3a0200e)
![image](https://github.com/openSUSE/agama/assets/112288843/8f675181-ee60-4c65-ae45-a8da4bbf8e77)

balsa-asanovic:
Okay, I've just did that, so now when I run the service it seems to be issues free, but the initial problem remains:

![image](https://github.com/openSUSE/agama/assets/112288843/16086a63-620d-4f12-adc2-690a5a6c7615)

balsa-asanovic:
Yeah, that's right. After I run cockpit.logout() I get logged out. When I log back in again, I get the same message

![image](https://github.com/openSUSE/agama/assets/112288843/06586285-2f11-48eb-809d-e5e91b5d834c)
![image](https://github.com/openSUSE/agama/assets/112288843/5629d31e-f575-4c0a-bb2a-4bfed1e4ad5b)

[...]

PR #564, opened 2023-05-08
[web] Blur and grayscale `aria-hidden` and `inert` div nodes

dgdavid:
## Problem

While the sidebar is open, the rest of the content is _inert_ and _aria hidden_. This on purpose, done in https://github.com/openSUSE/agama/pull/563.

But there is no is not enough visual hints about this neither, when the Agama sidebar is open nor a PatternFly Modal is displayed.

## Solution

To use some [CSS filters](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) to make more evident what is going on. Grayscale and blur, to be more precise.

## Testing

- Tested manually

## Notes

There were some issues with modal dialogs. See comments marked as outdated and/or read https://github.com/openSUSE/agama/pull/580 to know more.

## Screenshots

<details>
<summary>Click to show/hide screenshots</summary>

---

|Sidebar closed | Sidebar open |
|-|-|
|Screen Shot 2023-05-08 at 16 59 06 | Screen Shot 2023-05-08 at 16 59 01 |
| Screen Shot 2023-05-08 at 17 17 13 | Screen Shot 2023-05-08 at 17 17 00 |

|Modal closed | Modal open |
|-|-|
|Screen Shot 2023-05-08 at 17 01 20 | Screen Shot 2023-05-08 at 17 01 35 |
| Screen Shot 2023-05-08 at 17 17 13 | Screen Shot 2023-05-08 at 17 17 09 |
</details>

PR #572, opened 2023-05-10
[web] Ensure Sidebar siblings do not remain inert forever 

dgdavid:
## Problem

As mentioned in https://github.com/patternfly/patternfly-react/pull/9096, #564 revealed that unmounting a PF4/Modal component results in its siblings remaining hidden from the accessibility API.

Since #563 followed the same pattern of hiding the Sidebar siblings, it may suffer the same problem if it is unmounted (although it is not supposed to happen). Therefore, it would be better to minimize the risk of keeping siblings inert if the Sidebar is unmounted in the future.

## Solution

To remove the `inert` and `aria-hidden` attributes from its siblings **BEFORE** the Sidebar is unmounted. Note the use of [`useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect) for doing so while the component still having access to the parent and its children.

## Testing

- Added another unit test

- Tested manually

## Notes

* Consider this PR as an extension or improvement of #563. I.e., no entry in the changelog needed.
* I was tempted to drop the `aria-hidden` attribute and use only `inert`, which [should do both](https://html.spec.whatwg.org/multipage/interaction.html#inert-subtrees), make the element not interactive and remove it from the accessibility tree

  > Inert nodes generally cannot be focused, and user agents do not expose the inert nodes to accessibility APIs or assistive technologies.

  But I have left there meanwhile to be _in sync_ with PF4/Modal. We can drop it in the future, once `inert` is more adopted.

  To know more, read https://developer.chrome.com/articles/inert/

* We started doing this in #563 for _mimicking_ PF4/Modal **and** to be ready for #564. Now I have realized that all of this is well described in the [Dialog (Modal) pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) available at [ARIA Authoring Practices Guide (APG) ](https://www.w3.org/WAI/ARIA/apg/). So, looks like we're in the right path :wink: (and doing _almost the same_ that the modal version of the native [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element, which would be nice to start using at some point in the future).
* Worth mentioning/reminding that for now the component is just adding/removing these attributes, but not tracking if they were already there (which is unlikely) for restoring them instead of removing. Related to #565.

## Screenshots

N/A

### DIFF
--- a/web/src/components/core/Sidebar.jsx
+++ b/web/src/components/core/Sidebar.jsx
@@ -101,7 +101,7 @@ export default function Sidebar ({ children }) {
   }, [isOpen]);
 
   useLayoutEffect(() => {
-    // Ensure siblings do not remain inert when the component is unmount.
+    // Ensure siblings do not remain inert when the component is unmounted.
     // Using useLayoutEffect over useEffect for allowing the cleanup function to
     // be executed immediately BEFORE unmounting the component and still having
     // access to siblings.
--- a/web/src/components/core/Sidebar.jsx
+++ b/web/src/components/core/Sidebar.jsx
@@ -32,9 +32,9 @@ import { useNotification } from "~/context/notification";
  * @returns {HTMLElement[]}
  */
 const siblingsFor = (node) => {
-  if (!node?.parentNode) return [];
+  const parent = node?.parentNode;
 
-  return [...node.parentNode.children].filter(n => n !== node);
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #579, opened 2023-05-17
Adapt storage client

joseivanlopez:
## Problem

The UI for the device selector should show more information, for example, the systems installed into each device, its model, etc. The information exposed on D-Bus was extended to provide all the required info, see https://github.com/openSUSE/agama/pull/574. But the UI is not adapted yet. 

Part of https://trello.com/c/zpweLQMI/3347-3-agama-selection-of-boot-disk.

## Solution

Adapt the storage client to read the new information from D-Bus. Components are also adapted to work with the new device objects.

## Testing

* Added new tests
* Tested manually

### DIFF
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -20,6 +20,7 @@
  */
 
 // @ts-check
+// cspell:ignore ptable
 
 import DBusClient from "./dbus";
 import { WithIssues, WithStatus, WithProgress } from "./mixins";
@@ -38,6 +39,7 @@ const PROPOSAL_IFACE = "org.opensuse.Agama.Storage1.Proposal";
 const STORAGE_OBJECT = "/org/opensuse/Agama/Storage1";
 const STORAGE_JOB_IFACE = "org.opensuse.Agama.Storage1.Job";
 const STORAGE_JOBS_NAMESPACE = "/org/opensuse/Agama/Storage1/jobs";
+const STORAGE_SYSTEM_NAMESPACE = "/org/opensuse/Agama/Storage1/system";
 
 /**
  * Removes properties with undefined value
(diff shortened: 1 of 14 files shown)


web/src/client/storage.js
```diff
@@ -58,25 +60,140 @@ const removeUndefinedCockpitProperties = (cockpitObject) => {
...
+        device.udevIds = blockProperties.UdevIds.v;
+        device.udevPaths = blockProperties.UdevPaths.v;
+      };
+
+      const addPtableProperties = (device, ptableProperties) => {
+        device.partitionTable = { type: ptableProperties.Type.v };
+      };
+
+      const device = {
+        sid: path.split("/").pop(),
```
dgdavid, line 144: NP: [`pop()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop) is a mutating method, which make it less performant than simply using `[0]`

web/src/components/storage/utils.js
```diff
@@ -0,0 +1,59 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
```
dgdavid, line 21: NP: great having storage utils together in a separate files. However, I have no clear if it is better to place the file here, under `~/components/storage` or we should go for `~/utils` and move it to `~/utils/storage` and the current utils to `~/utils/common` or so. At the end, these utils can be used in no-component files too.

Anyway, just thinking aloud. I have no strong opinion in that regard at this moment, and what is better, we can do proposed change later.

dgdavid (approved):
Great, thank you. 

Just minor comments from my side, but it looks good to me :+1

web/src/client/storage.js
joseivanlopez: Yes, but we need the last element. I guess that pop operation is cheap enough anyway.

PR #580, opened 2023-05-17
[web] Workaround for removing `aria-hidden` leftovers

dgdavid:
## Problem

When a PF4/Modal is unmounted instead of set as `isOpen={false}`, its siblings remain as `aria-hidden`. To know more,  read following links

* https://github.com/patternfly/patternfly-react/pull/9096, the fix sent to upstream
* https://github.com/openSUSE/agama/pull/564, the RFC where we were aware about the issue
* https://github.com/openSUSE/agama/pull/576, an attempt to see how difficult would be to stop unmounting these dialogs and switch them to their _close_ state instead.

  - There is a _collision issue_ when more than one modal is mounted in the same parent component: in each parent state change, the re-rendering process will trigger their `componentDidUpdate` method which is directly manipulating the DOM. Thus, the first dialog rendered as open adds the `aria-hidden` attribute to its siblings but subsequents rendered as not open dialogs will remove the same attribute from the same siblings.

We have been told by the Patternfly team that the patternfly-react repo is [code frozen and the fix will be review for the next release, in Q3 at the earliest](https://github.com/patternfly/patternfly-react/pull/9096#issuecomment-1550184679).

## Solution

To use a workaorund meanwhile: removing the `aria-hidden` attributes from the `document.body` children when the Agama/Popup is about to being unmounted. 

Why not using a [`ref`](https://react.dev/learn/manipulating-the-dom-with-refs) to ask for its siblings instead as we did in https://github.com/openSUSE/agama/pull/572 for the similar behavior of the `Sidebar`? Because PF4/Modal does not forward the ref. Read https://react.dev/learn/manipulating-the-dom-with-refs#accessing-another-components-dom-nodes

There are other tricky ways to workaround the issue (such as looking for the element with [backdrop open CSS class](https://github.com/patternfly/patternfly-react/blob/a0f857c4de39dd415792a8701e7c6ac7fd853024/packages/react-core/src/components/Modal/Modal.tsx#L177) :dizzy_face: :no_mouth:), but since Agama is not using the `appendTo` PF4/Modal prop it is enough removing the attribute from the [default target](https://github.com/patternfly/patternfly-react/blob/a0f857c4de39dd415792a8701e7c6ac7fd853024/packages/react-core/src/components/Modal/Modal.tsx#L115) children.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

N/A


ancorgs (approved):
LGTM

PR #581, opened 2023-05-18
Update ruby-dbus dependency to 0.22.x

mvidner:
## Problem

With a new ruby-dbus in OBS, the old version on which agama depends is gone. Agama containers in OBS no longer build

## Solution

Update the dependency, until the next change.

Next time I will consider making a looser dependency, like ">= 0.23", "< 1.0", which could be less work fixing the breaking changes than bumping the harmless changes more often.

## Testing

OBS will start building, for sure

## Screenshots

no


aaannz (approved):
Thanks!

joseivanlopez (approved):
LGTM

PR #583, opened 2023-05-22
[web] Fix header and footer block-size

dgdavid:
## Problem

The Web UI header and footer were using [viewport relative units](https://www.w3.org/TR/css-values-3/#viewport-relative-lengths) for calculating their block-size (aka height), which make them looks broken and ugly in small screen.

## Solution

Use [`auto`](https://www.w3.org/TR/css-sizing-3/#automatic-size) instead.

## Testing

- Tested manually

## Screenshots

| Before | After |
|-|-|
| agama_header_and_footer_before | agama_header_and_footer_after |


ancorgs (approved):
LGTM

PR #590, opened 2023-05-25
Edit volume sizes

joseivanlopez:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


dgdavid:
@joseivanlopez 

Apart from the PR description and the changes file, this is hopefully ready for a first review round. Please, jump into the branch and give it a shot locally. Thanks.

web/src/components/storage/VolumeForm.jsx
```diff
@@ -0,0 +1,479 @@
...
+ * @return {object} an object ready to be used as a "form state"
+ */
+const prepareFormData = (volume) => {
+  const { size: minSize, unit: minSizeUnit } = splitSize(volume.minSize);
+  const { size: maxSize, unit: maxSizeUnit } = splitSize(volume.maxSize);
+
+  return {
+    size: minSize,
+    sizeUnit: minSizeUnit,
+    minSize, // : minSize.replace(/\.0+$/, ""),
```
joseivanlopez, line 336: ?

dgdavid: Leftover, good catch :)

web/src/components/core/NumericTextInput.jsx
```diff
@@ -0,0 +1,63 @@
...
+ * @note It allows empty value too.
+ *
+ * @param {object} props
+ * @param {string|number} props.value - the input value
+ * @param {onChangeFn} props.onChange - the callback to be called when the entered value match the input pattern
+ * @param {object} props.textInputProps - @see {@link https://www.patternfly.org/v4/components/text-input/#textinput}
+ *
+ * @returns {ReactComponent}
+ */
+export default function NumericTextInput({ value = "", onChange = noop, ...textInputProps }) {
```
dgdavid, line 50: We could go further and do this helper more generic by receiving the pattern via prop. But let's wait until we really need it. Do you agree?

BTW, we are not using directly the HTML [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) attribute because it's not intended for limiting the input from the user but for performing validations before send the form. But since we're preventing the form submission, it is not as direct as the proposed solution. To learn more, read https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation#constraint_validation_process

Of course, we can evaluate the use of `pattern` attribute in the future.

web/src/components/storage/utils.test.js
```diff
@@ -44,3 +44,44 @@ describe("deviceLabel", () => {
...
+    expect(splitSize(null)).toEqual({ size: 0, unit: "B" });
+  });
+
+  it("returns an 'empty' size object when -1 is given", () => {
+    expect(splitSize(-1)).toEqual({ size: "", unit: "" });
+  });
+
+  it("returns an 'empty' size object when empty string is given", () => {
+    expect(splitSize("")).toEqual({ size: "", unit: "" });
+  });
```
imobachgs, line 86: What about `splitSize("30")`?

web/src/components/storage/VolumeForm.jsx
```diff
@@ -0,0 +1,478 @@
...
+ * using the form id and the callback function used to perform the submission
+ * once the form has been validated.
+ *
+ * @param {object} props
+ * @param {string} props.id - Form ID
+ * @param {object[]} props.templates - Volume templates
+ * @param {onSubmitFn} props.onSubmit - Function to use for submitting a new volume
+ *
+ * @callback onSubmitFn
+ * @param {object} volume
```
imobachgs, line 404: np: Although we have defined a [Volume type](https://github.com/openSUSE/agama/blob/58ca603d7e533e46e72d3161c8d722efafca20a9/web/src/client/storage.js#L210), I guess we lost track in some of the previous calls. For someone unfamiliar with the code, it might help to put the proper types. 😅

web/src/components/storage/VolumeForm.jsx
```diff
@@ -0,0 +1,478 @@
...
+        <MountPointFormSelect
+          id="mountPoint"
+          value={state.formData.mountPoint}
+          onChange={changeVolume}
+          volumes={currentVolume ? [currentVolume] : templates}
+          isDisabled={currentVolume !== undefined}
+        />
+      </FormGroup>
+      <FormGroup isRequired label="File system type" fieldId="fsType">
+        <NumericTextInput
```
imobachgs, line 466: Why a `NumericTextInput`?

web/src/components/storage/VolumeForm.jsx
```diff
@@ -0,0 +1,478 @@
...
+              validated={errors.maxSize && 'error'}
+              aria-label="Maximum desired size"
+              value={formData.maxSize}
+              onChange={(maxSize) => onChange({ maxSize })}
+
+            />
+            <SizeUnitFormSelect
+              id="maxSizeUnit"
+              aria-label="Max size unit"
+              units={Object.values(SIZE_UNITS)}
```
imobachgs, line 215: Perhaps it can be part of the `SizeUnitFormSelect`. After all, we are always giving the same value.

web/src/components/storage/utils.js
```diff
@@ -19,28 +19,101 @@
...
+ * parseSize("1 KiB")
+ * // returns 1024
+ *
+ * parseSize("")
+ * // returns 0
+ *
+ * @param {string|number} size
+ * @returns {number}
+ */
+const parseSize = (size) => {
```
imobachgs, line 110: Parse to what? I would improve the naming here.

imobachgs (commented):
In general, it looks good. Great job! Otherwise, I have a few comments.

web/src/components/storage/utils.js
```diff
@@ -19,28 +19,101 @@
...
+ * @returns {SizeObject}
+ */
+const splitSize = (size) => {
+  // From D-Bus, maxSize comes as -1 when set as "unlimited", but for Agama UI
+  // it means "leave it empty"
+  const sanitizedSize = size !== -1 ? size : "";
+  const parsedSize = typeof sanitizedSize === "string" ? sanitizedSize : xbytes(sanitizedSize, { iec: true });
+  const [qty = "", unit = ""] = parsedSize.split(" ");
+
+  return { unit, size: qty === "" ? qty : Number(qty) };
```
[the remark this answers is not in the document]
dgdavid: Not anymore, I've changed this while you were doing the review. See 8d19eece78bf66d24be7fe3c51dff21a24097772

web/src/components/storage/utils.js
```diff
@@ -19,28 +19,100 @@
...
 
-// cspell:ignore filesize
+// cspell:ignore xbytes
 
-import { filesize } from "filesize";
+import xbytes from "xbytes";
+
+/**
+ * @typedef {Object} SizeObject
+ * @property {number} size - The "amount" of size
```
imobachgs, line 28: np: So now they can be undefined, right? I would extend the docs.

imobachgs (approved):

web/src/components/storage/utils.test.js
dgdavid: I have _improved_ the `splitSize` code. Admittedly, it's a bit ugly function. I'm not happy with it, but I hope we can improve the situation in next iterations. Sorry, I have no better idea by now for dealing with sizes in this form.

web/src/components/storage/VolumeForm.jsx
dgdavid: You're completely right. I'll try to fix it.

web/src/components/storage/VolumeForm.jsx
dgdavid: For limiting the user input to numbers (including floats) only. See 417f8bf0d10d13b82e47faea960ddfe01c74b108 and https://github.com/openSUSE/agama/pull/590#discussion_r1223162713

web/src/components/storage/VolumeForm.jsx
dgdavid: And initially it was part of it. But then I extracted it because _a premature improvement_, thinking in a context where the selector cuold be used with a narrowed set of units. Although it's true that

- the SizeUnitFormSelect is not exposed yet
- is is using always the same units in this form

If you don't mind, I'm going to keep it as it is now.

web/src/components/storage/utils.js
dgdavid: To bytes. Let's rename that function

web/src/components/storage/utils.js
dgdavid: Yes. The doc update is on the way, along other changes done after reading all your comments :) Thanks!

imobachgs (approved):

imobachgs (approved):

PR #591, opened 2023-05-25
Improve the yupdate scripts

lslezak:
## Problem

- The `yupdate` script might fail when installing the missing packages because the Agama service can hold the libzypp lock

## Solution

- Stop the Agama service to possibly release the libzypp lock, start it back after installing the packages

## Additional Fixes

- Use `set -e` to fail on error
- Use `set -x` when $DEBUG is set to `1` for debugging

## Testing

- Tested manually

### DIFF
--- a/.yupdate.post
+++ b/.yupdate.post
@@ -10,6 +10,12 @@
 #   next yupdate run
 #
 
+set -e
+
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #593, opened 2023-05-26
Install cargo as part of setup-service.sh

lkocman:
* cargo is not availble on fresh [distrobox](https://en.opensuse.org/Distrobox) installation but the setup-service.sh is expecting it to be present


mvidner (approved):
Thanks!

And this made me finally install distrobox :smiley:

PR #595, opened 2023-05-29
[RFC] add initial proposal for unattended questions

jreidinger:
## Problem

With introduced questions mechanism it is impossible to have real unattended installation.

trello: https://trello.com/c/eMVlSQnU/3364-2-agama-research-design-a-mechanism-to-ask-questions-during-auto-installation

## Solution

Create proposal how to modify questions and add support for unattanded installation to them.

### DIFF
--- a/doc/questions.md
+++ b/doc/questions.md
@@ -1,0 +1,48 @@
+## Questions
+
+Questions are mechanism that allows user to react on certain situations that happen in installer.
+
+### Questions and Unattended Installation
+
+Questions works well, but in some cases like automatic installation or installation driven by CLI
+blocking questions can be problematic. So we introduce mechanism how to make even questions
+non-blocking and non-interactive. The main goal is to allow user flexible way to answer questions that they
+know in advance and for rest that is not specified or do not see yet, simply use the default answer.
+
+#### Requirements for Questions
+
+To make it works there are several requirements for questions:
+
+1. each question has ID that is hierarchical so partial match can be used. Example `storage.luks.password` or
+`software.repository.checksum`
+2. each question has defined default safe answer.
+3. questions can be up to 4 additional params that helps user to answer them. Example repository url and its checksum.
+   Partition identification for luks password.
+4. Question can work in attended and unattended mode.
+
+#### Features of Unattended Installation
+
+1. Questions beside ids contain also values ( up to 4 with keys value1..4 ) that can be matched. For matching
+id and values it can be used string, regexp or array. Array list all possible values and regexp is used for matching.
+2. if value is not specified in answer that it matches all of them ( so partial answer is possible )
+3. For questions without defined answer default answer will be used.
+4. All questions and answers ( together with source from where it comes )
+   will be logged for later audit ( maybe own file? or maybe directly write
+   it as answers.yml? )
+5. asnwers can be provided also in interactive mode. In that case it auto answer known answers and
+   do not use defaults for others.
+
+
+#### Use Cases and Its Solutions
+
+1. I run unattended installation for first time and want to see what questions appear to be able to
+modify it to my needs for mass deployment. -> There is questions.log which writes down exact questions
+with all its params, ids, etc and used answer with note if it is automatic answer or one comming from
+file.
+2. I run modified ISO which points to my own software repository. I want to auto agree with my own
+  GPG key that I know in advance. -> use answers.yml file
+3. I do mass deployment in environment where previous requirement was to have luks encryption with some
+   random password and this partition should not be reused -> do nothing as default answer is to skip such partition
+4. I have my own vendor iso and I want to preconfigure installer using CLI before showing web UI. And some actions can/will
+   provide questions that I want to answer before user see UI -> Providing to CLI answers.yml but
+   do not set unattended installation, so user actions that result in questions are answered by user.


doc/questions.md
```diff
@@ -0,0 +1,48 @@
+## Questions
+
+Questions are mechanism that allows user to react on certain situations that happen in installer.
```
mvidner, line 3: Please use https://www.deepl.com/write for grammar improvement. My work taken over by the robots, YAYYY!

mvidner: Let's make it clear that an Agama Question is not just any question.

```suggestion
[Questions][] are mechanism that allows user to react on certain situations that happen in installer.

[Questions]: https://opensuse.github.io/agama/dbus/ref-org.opensuse.Agama.Questions1.html
```

mvidner (commented):
I'd like to know how this design affects the D-Bus API

doc/questions.md
```diff
@@ -0,0 +1,48 @@
...
+know in advance and for rest that is not specified or do not see yet, simply use the default answer.
+
+#### Requirements for Questions
+
+To make it works there are several requirements for questions:
+
+1. each question has ID that is hierarchical so partial match can be used. Example `storage.luks.password` or
+`software.repository.checksum`
+2. each question has defined default safe answer.
+3. questions can be up to 4 additional params that helps user to answer them. Example repository url and its checksum.
```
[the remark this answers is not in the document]
jreidinger: well, I consider 4 as reasonable number..but it is true, that hash can be also helpful and more flexible.

doc/questions.md
```diff
@@ -0,0 +1,48 @@
...
+
+1. I run unattended installation for first time and want to see what questions appear to be able to
+modify it to my needs for mass deployment. -> There is questions.log which writes down exact questions
+with all its params, ids, etc and used answer with note if it is automatic answer or one comming from
+file.
+2. I run modified ISO which points to my own software repository. I want to auto agree with my own
+  GPG key that I know in advance. -> use answers.yml file
+3. I do mass deployment in environment where previous requirement was to have luks encryption with some
+   random password and this partition should not be reused -> do nothing as default answer is to skip such partition
+4. I have my own vendor iso and I want to preconfigure installer using CLI before showing web UI. And some actions can/will
```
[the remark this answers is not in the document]
jreidinger: well, basically question is if there is reason to have answers and interactive mode or not ( so basically answer only known questions and for rest ask user instead of using auto reply ).

mvidner:
> never trust robots that you do not program yourself :)

@jreidinger very true, you still have to proofread what DeepL gave you to check that it didn't change the meaning.

mvidner:
So again, I'd like to know how this design affects the D-Bus API.

I can imagine a design, where all of the unattended handling just prevents some/all of the Questions from appearing on the bus so the API would stay the same.

If the API does change, I think it is worthwhile specifying now in this PR

doc/questions.md
```diff
@@ -0,0 +1,48 @@
...
+
+#### Features of Unattended Installation
+
+1. Questions beside ids contain also values ( up to 4 with keys value1..4 ) that can be matched. For matching
+id and values it can be used string, regexp or array. Array list all possible values and regexp is used for matching.
+2. if value is not specified in answer that it matches all of them ( so partial answer is possible )
+3. For questions without defined answer default answer will be used.
+4. All questions and answers ( together with source from where it comes )
+   will be logged for later audit ( maybe own file? or maybe directly write
+   it as answers.yml? )
```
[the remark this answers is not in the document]
lslezak: Yes, some filtering should be done.
RoR uses `[FILTERED]` IIRC...

doc/questions.md
```diff
@@ -0,0 +1,53 @@
...
+3. questions can define additional params in map that help user to answer them. Example repository url and its checksum.
+   Partition identification for luks password. Why: Getting data only from question text is hard and very unreadable with regexp.
+   So questions need to specify some easy to match data in addition to the pure text.
+   E.g. checksum and repository url for automatic repository approval.
+4. Question can work in attended and unattended mode. Why: To know when to apply default answer. See 2.
+
+#### Unattended Installation Features
+
+1. Questions contain values in addition to ids that can be matched. To match
+id and values, string, regexp or array can be used. Array lists all possible values and regexp is used for matching.
```
mvidner, line 30: IMHO regexp (regular expressions) are overkill (incompatible dialects...) and I hope substrings or shell patterns will be enough

mvidner:
To avoid Mode in the service...
Another (crazy?) design: the service just asks a question on the bus, and there is an "auto answerer client" that answers all questions

doc/questions.md
jreidinger: that is good suggestion. Maybe for the first phase we can omit regexp completely and if needed, we can extend it ( so some string that start with some fancy character to support advanced matching ).

jreidinger:
> My thoughts:
> 
>     * I agree with:
>       
>       * Having a file with the provided answers.
>       * With a matching mechanism (e.g., what type of question, what device, what repo, etc).
>       * With a default answer (but optional).
> 
>     * I disagree with:
>       
>       * Having modes (auto and manual).
>       * Having extra data (for matching) in the question itself.
> 
> 
> How do I envision it?
> 
>     * Passing an optional (and encrypted?) file with the answers (similar to autoinstall profile).
> 
>     * Having a matching mechanism as proposed.
> 
>     * Having an optional default answer for each question.
> 
>     * A question is not asked if its answer is already known (why to ask for luks password if it was already provided by an answers file?).
> 
>     * No difference between auto and manual install. If the answer is known, then take it. Otherwise, ask a question.

Thanks for notes. I think there is a lot of agreement and just a few collisions.
I think the main one is how unattended installation should look like. For me, if we have unattended installation, it should really be unattended unless major and critical issue appear ( like not able to install anything, not able to partition, etc. ). And for that we have issues and not questions. Question is for me when installer missing some info, but should be able to continue without that info. And in unattended mode questions should never be asked from my POV.

Just to clarify few notes:

1. modes will be only in question object. So no global mode like in yast, just set questions object to unattended mode.
2. we can maybe use that extra data from questions in UI itself, as question can be more limited then data, so we can show it in some details for question or in CLI. E.g. consider some additional details about disk that is luks encrypted.
3. if we do not have extra data in questions for matching, we still need to somehow match answer and question, so if it is only in answer, it means that each object needs to do matching on its own, which can be quite messy.
4. in my proposal is also passing answers file, but if it is missing, just defaults are used. Note about encrypted is good, I will add TODO as we probably do not need it for first iteration
5. What you describe in the last point is basically how can happen my proposal if answers are provided and questions are not set to unattended mode. Question is if it is useful for users.

@imobachgs how do you see it? Especially about requirements for unattended mode and possible blocked by question that is not specified in advance.

imobachgs:
About the "unattended mode", I see how it is useful but I also understand @joseivanlopez's concern.

I am a little bit late to the discussion but, IMHO, the important thing is to set the *default behavior* when an answer is not defined (using an `answers.yml` file or whatever mechanism). For instance, when performing an autoinstallation, you could set the behavior to *use default*. On a regular installation, the behavior could be *ask the user*.

So you do not need to model the concept of *mode*, but just to set the fallback behavior. Does it make any sense?

doc/questions.md
```diff
@@ -1,53 +1,55 @@
 ## Questions
 
-[Questions][] are a mechanism that allows the user to react to certain situations that occur in the installer.
+[Questions][] are a mechanism that allows the installer to ask for additional data which are needed without knowing it in advance.
```
joseivanlopez, line 3: which **is** needed

joseivanlopez (approved):
LGTM

PR #597, opened 2023-06-01
add initial dummy questions service

jreidinger:
## Problem & Solution

Move questions service from ruby to rust to reduce memory consumption as there is no dependency on yast.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


imobachgs (commented):
I don't know why you are getting the error about not finding the interface, but I have a few comments about the current implementation. Of course, take them with a grain of salt. I hope it helps.

## Services and interfaces

I would not use `QuestionsService` as a D-Bus interface itself. In my opinion, we are mixing two concepts: the service itself and the `org.opensuse.Agama.Questions1` D-Bus interface.

* The service object would be responsible for initializing and publishing the `/org/opensuse/Agama/Questions1` object and, optionally, keeping the list of questions. It might also be responsible for holding the information about unattended questions (in the future).
* The interface would be responsible for handling the D-Bus calls.

The problem when using that approach is how (and where) to keep the list of questions (having concurrency into account). I can think of two options:

* Not having a list of questions at all 😅 When `New` is called, the `org.opensuse.Agama.Questions1` interface builds the question and publishes the object on the tree. After all, the interface has access to the object server. See the [zbus::dbus_interface example](https://docs.rs/zbus/latest/zbus/attr.dbus_interface.html#example).
* Keeping the state in the service (or another struct) and using some synchronization mechanism (a mutex, channels, etc.). Sure, I think this is the way to go but...

We could implement the first (ugly) solution and move to the second one later.

## The `Question` trait

To be honest, I do not think that the questions need to be aware that they are being added to a D-Bus tree. I think a D-Bus interface (or the service or a special struct) should be responsible for handling the tree.


## The data model

Instead of using trait objects, I would use an enum to model the questions. Perhaps something like:

```rust
struct BaseQuestion {
  id: u32,
  text: String,
  // whatever it is shared  
}

struct LuksQuestion {
  base: Question,
  password: String,
}

enum Question {
  Generic(BaseQuestion),
  Luks(LuksQuestion)
}

impl Question {
  // you can implement methods like text()
}
```

However, we can discuss about this later in time, as using a `dyn` is correct too, just that I am unsure about its benefits in this scenario (well, not having to put questions into boxes is one of them). But the enum approach might be quite more verbose.

mvidner:
> I would not use `QuestionsService` as a D-Bus interface itself. In my opinion, we are mixing two concepts: the service itself and the `org.opensuse.Agama.Questions1` D-Bus interface.

You mean like `NetworkService` is separated from `Tree` in agama-network?

imobachgs:
It looks much better after the refactoring. 😃

rust/agama-dbus-server/src/main.rs
```diff
@@ -9,6 +10,7 @@ const ADDRESS: &str = "unix:path=/run/agama/bus";
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // When adding more services here, the order might be important.
+    crate::questions::start_service(ADDRESS).await?;
```
imobachgs, line 13: Three lines and all of them are different :)

* The first line uses the full path to call a function (`start_service`).
* The second line calls an associated function (`start`) of a struct (`LocaleService`) using its full path. However, it uses the `let _con` trick (because it does not own the connection), which looks obscure to me.
* The third one is a combination: it calls an associated function (`start`) of a struct (`NetworkService`) that is previously imported (no full path).

Let's think about having a style guide. Otherwise, the code will get out of control pretty quickly.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub fn set_answer(&mut self, value: &str) -> Result<(), zbus::fdo::Error> {
+        // TODO verify if answer exists in options or if it is valid in other way
+        self.answer = value.to_string();
+
+        Ok(())
+    }
+}
+
+#[derive(Clone,Debug)]
+pub struct LuksQuestion {
```
imobachgs, line 70: Please, document the struct.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -0,0 +1,215 @@
...
+    password: String,
+    attempt: u8,
+    generic_question: GenericQuestion,
+}
+
+impl LuksQuestion {
+    fn device_info(device: &str, label: &str, size: &str) -> String{
```
imobachgs, line 77:
```suggestion
    fn device_info(device: &str, label: &str, size: &str) -> String {
```

We should adopt rustfmt and clippy.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut result = device.to_string();
+        if !label.is_empty() {
+            result = format!("{} {}", result, label);
+        }
+
+        if !size.is_empty() {
+            result = format!("{} ({})", result, size);
+        }
+
+        return result.to_string();
```
imobachgs, line 87: In Ruby we omit the `return` word whenever we can. Why not doing the same here?

```suggestion
        result.to_string()
```

imobachgs (commented):
About the approach, I find that `GenericQuestion` and `LuksQuestion` have "too many" responsibilities: they represent a question and its D-Bus interface. I usually prefer to have separate concepts: the question itself (kind of the model) and its D-Bus representation.

Having said that, I guess it is a matter of taste and the code looks good enough as a first version. But please, document it. :wink:

rust/agama-dbus-server/src/main.rs
jreidinger: agreed. To be honest here when I need to start service I do not justify to me why it needs anything to own so that is why I do not use struct...at least I will try to unify locale and and here questions to looks similar.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -0,0 +1,215 @@
...
+
+#[dbus_interface(name = "org.opensuse.Agama.Questions1")]
+impl Questions {
+    #[dbus_interface(name = "New")]
+    async fn new_question(&mut self, text: &str, options: Vec<&str>, default_option: Vec<&str>) -> Result<ObjectPath, zbus::fdo::Error> {
+        let id = self.last_id;
+        self.last_id += 1; // TODO use some thread safety
+        let options = options.iter().map(|o| o.to_string()).collect();
+        // TODO: enforce default option and do not use array for it to avoid that unwrap
+        let question = GenericQuestion::new(id, text.to_string(), options, default_option.first().unwrap().to_string());
```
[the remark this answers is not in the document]
jreidinger: it is mentioned in TODO. When we in next step touch API we should really kill that array for default option as it makes no sense at all. It should be just single value which is default value ( so unwrap is not needed here ).
Another unwraps I use for building ObjectPath as I know it is correct one in advance.

imobachgs (approved):
To be honest, I did not have a problem with having the locale service as a struct. Actually, I prefer that approach. It would even has an API to stop, clean-up the service, and so on.

But let's discuss this outside of this PR and let's move the code to the staging area.

PR #606, opened 2023-06-06
GitHub Actions: update systemsmanagement:Agama:Staging in OBS

lslezak:
## Problem

- Automatically synchronize the packages in the [systemsmanagement:Agama:Staging](https://build.opensuse.org/project/show/systemsmanagement:Agama:Staging) OBS project

## Solution

- Run the OBS services manually in a GitHub workflow
- [Example run](https://github.com/openSUSE/agama/actions/runs/5186203885/jobs/9347009788)  for the [agama-cli](https://build.opensuse.org/package/show/systemsmanagement:Agama:Staging/agama-cli) package

## Notes

As the GitHub workflow is executed only for the master branch there is no run for the last commits. If you want to see an example run check the [actions history](https://github.com/openSUSE/agama/actions) where I enabled the run also for the branch for this pull request.

The `rake osc:commit` by default calls `osc:build` target as a dependency. But building the Ruby gem package does not provide much value (it basically just unpack the files from `*.gem` to `/usr/lib64/ruby...` and compress that to an RPM). It just makes the submission much longer (1min -> 3mins).

## Additional Changes

- Use `--local` option when building the Ruby gem, it is much faster (1sec vs. 15sec) and the generated file is the same
- Build the gem version from git tag + offset, do not hardcode it in a file

## Tasks

- [x] Synchronize the `agama-cli` package
- [x] Synchronize the `cockpit-agama` package
- [x] Synchronize the `cockpit-agama-playwright` package
- [x] Synchronize the `rubygem-agama` package
- [x] Share some parts of the workflow ([GitHub documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows))
- [x] Add documentation


jreidinger (approved):
overall LGTM

PR #612, opened 2023-06-08
Fix the initialization of the Locale1 service

imobachgs:
Until we find a better way, we need to keep the connection around. Otherwise, it will get dropped.

A better solution might be to pass a copy of the connection instead of the address to each service. But, today, let's keep the API as it is.

### DIFF
--- a/rust/agama-dbus-server/src/main.rs
+++ b/rust/agama-dbus-server/src/main.rs
@@ -11,7 +11,7 @@ const ADDRESS: &str = "unix:path=/run/agama/bus";
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // When adding more services here, the order might be important.
     crate::questions::start_service(ADDRESS).await?;
-    crate::locale::start_service(ADDRESS).await?;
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

rust/agama-dbus-server/src/main.rs
```diff
@@ -11,7 +11,7 @@ const ADDRESS: &str = "unix:path=/run/agama/bus";
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // When adding more services here, the order might be important.
     crate::questions::start_service(ADDRESS).await?;
-    crate::locale::start_service(ADDRESS).await?;
+    let _conn = crate::locale::start_service(ADDRESS).await?;
```
jreidinger, line 14: btw we should document that we need to store connection, otherwise it will drop whole dbus object. So next time we do not remove it :)

imobachgs: Yes, we should document it. Additionally, trying the approach I proposed in the description might help.

PR #615, opened 2023-06-12
[web] Users: improve the root auth section

dgdavid:
## Problem

The current way Agama is rendering the _Root Authentication_ section in the _Users_ page do not make evident that there is none auth method defined, and it's easy to overlook the button for displaying the actions available for each method.

## Solution

Provide the user more comprehensive information if there is no authentication method for the root user, which may result in a system without administrative access.

For that, the _empty state_ pattern we already have in the user section is used when there is none auth method defined. Once the user define one, the table listing the auth methods, their state and the actions is show _again_. The idea is to improve this too, but in a subsequent PR once we have more clear how to really improve it.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

_Note: screenshots below might have change because minor changes asked during the review, specially in regard to the wording_

### Before

| No auth method defined | At least one auth method defined |
|-|-|
|Screen Shot 2023-06-12 at 12 04 57 |Screen Shot 2023-06-12 at 12 04 31 |

### After

| No auth method defined | At least one auth method defined |
|-|-|
|Screen Shot 2023-06-12 at 12 04 19 |Screen Shot 2023-06-12 at 12 04 31 |

### DIFF
--- a/web/src/components/users/RootAuthMethods.jsx
+++ b/web/src/components/users/RootAuthMethods.jsx
@@ -20,7 +20,7 @@
  */
 
 import React, { useState, useEffect } from "react";
-import { Skeleton, Truncate } from "@patternfly/react-core";
+import { Button, Skeleton, Truncate } from "@patternfly/react-core";
 import { TableComposable, Thead, Tr, Th, Tbody, Td } from '@patternfly/react-table';
 import { Em, RowActions } from '~/components/core';
 import { RootPasswordPopup, RootSSHKeyPopup } from '~/components/users';
@@ -28,6 +28,18 @@ import { RootPasswordPopup, RootSSHKeyPopup } from '~/components/users';
 import { useCancellablePromise } from "~/utils";
 import { useInstallerClient } from "~/context/installer";
 
+const MethodsNotDefined = ({ setPassword, setSSHKey }) => {
+  return (
+    <div className="stack">
+      <div className="bold">No root auth method defined yet</div>
+      <div>Please, define at least one root authentication method for being able to log into the system as administrative user.</div>
+      <div className="split">
+        <Button variant="primary" onClick={setPassword}>Set a password</Button>
+        <Button variant="secondary" onClick={setSSHKey}>Upload a SSH Public Key</Button>
+      </div>
+    </div>
+  );
+};
 export default function RootAuthMethods() {
   const { users: client } = useInstallerClient();
   const { cancellablePromise } = useCancellablePromise();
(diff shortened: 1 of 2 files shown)


web/src/components/users/RootAuthMethods.jsx
```diff
@@ -20,14 +20,26 @@
...
 import { useInstallerClient } from "~/context/installer";
 
+const MethodsNotDefined = ({ setPassword, setSSHKey }) => {
+  return (
+    <div className="stack">
+      <div className="bold">No root auth method defined yet</div>
+      <div>Please, define at least one root authentication method for being able to log into the system as administrative user.</div>
+      <div className="split">
+        <Button variant="primary" onClick={setPassword}>Set a password</Button>
+        <Button variant="secondary" onClick={setSSHKey}>Upload a SSH Public Key</Button>
```
dgdavid, line 38: No strong opinion at this point from my side here. Just setting one action as primary and the other as secondary for not crowding the UI of primary actions.

| Both as primary action | Only "Set a password" as primary action |
|-|-|
| Screen Shot 2023-06-12 at 12 10 49 | Screen Shot 2023-06-12 at 12 04 19 |

```suggestion
        <Button variant="primary" onClick={setPassword}>Set a password</Button>
        <Button variant="primary" onClick={setSSHKey}>Upload a SSH Public Key</Button>
```

web/src/components/users/RootAuthMethods.jsx
```diff
@@ -20,14 +20,26 @@
...
 import { Em, RowActions } from '~/components/core';
 import { RootPasswordPopup, RootSSHKeyPopup } from '~/components/users';
 
 import { useCancellablePromise } from "~/utils";
 import { useInstallerClient } from "~/context/installer";
 
+const MethodsNotDefined = ({ setPassword, setSSHKey }) => {
+  return (
+    <div className="stack">
+      <div className="bold">No root auth method defined yet</div>
```
dgdavid, line 34: What about 

```suggestion
      <div className="bold">None root auth method defined</div>
```

web/src/components/users/RootAuthMethods.jsx
```diff
@@ -20,14 +20,26 @@
...
+      <div>Please, define at least one root authentication method for being able to log into the system as administrative user.</div>
```
joseivanlopez, line 35: Maybe this shorter version? "Please, define an authentication method for loggin into the system as root."

web/src/components/users/RootAuthMethods.jsx
joseivanlopez: I visually prefer the version with primary and secondary buttons. And I would say it makes sense to offer the password as primary (it is the most common).

web/src/components/users/RootAuthMethods.test.jsx
```diff
@@ -70,12 +70,50 @@ describe("when loading initial data", () => {
 });
 
 describe("when ready", () => {
-  it("renders a table holding available methods", async () => {
-    installerRender(<RootAuthMethods />);
+  describe("and none method is defined", () => {
```
joseivanlopez, line 73: NP: "no method"

jreidinger (approved):
for me much better now

PR #617, opened 2023-06-12
workaround not complete objects exposed on dbus

jreidinger:
## Problem

Zbus implementation signal any new interface appearing and we failed to find way how to attach more interfaces together. This cause strange behavior on UI side when signal arrived. Refresh fixed it.

## Solution

Lets use a workaround that basically attach generic interface as the last one and on UI side object is considered complete only when generic interface appears.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/questions.rs
+++ b/rust/agama-dbus-server/src/questions.rs
@@ -200,14 +200,16 @@ impl Questions {
         );
         let object_path = ObjectPath::try_from(question.base().object_path()).unwrap();
 
-        let base = question.base();
+        let base = question.base().clone();
+        self.connection
+        .object_server()
+        .at(base.object_path(), question)
+        .await?;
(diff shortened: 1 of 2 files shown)


rust/agama-dbus-server/src/questions.rs
```diff
@@ -200,14 +200,16 @@ impl Questions {
         );
         let object_path = ObjectPath::try_from(question.base().object_path()).unwrap();
 
-        let base = question.base();
+        let base = question.base().clone();
         self.connection
-            .object_server()
-            .at(base.object_path(), base.clone())
-            .await?;
+        .object_server()
```
imobachgs, line 205: np: The indentation looks weird

imobachgs (approved):
Please, just fix the indentation issue. Otherwise, LGTM. Thanks!

PR #618, opened 2023-06-13
[web] Adjust volume size options wording and order

dgdavid:
## Problem

During the review meeting we had yesterday morning, we got feedback about the recently added options for adjusting a volume size

* To use "Fixed" instead of "Manual" for the option to tell the installer the exact desired size.
* To rearrange the options for ordering them from fully-automatic to fixed, keeping the semi-auto option (aka "Range") in the middle.

## Solution

To apply these suggestions.

## Testing

- Unit tests adapted
- Tested manually

## Screenshots

| Before | After |
|-|-|
|Screen Shot 2023-06-13 at 07 48 00 | Screen Shot 2023-06-13 at 07 47 30 |

### DIFF
--- a/web/src/components/storage/VolumeForm.jsx
+++ b/web/src/components/storage/VolumeForm.jsx
@@ -130,7 +130,7 @@ const SizeManual = ({ errors, formData, onChange }) => {
           <NumericTextInput
             id="size"
             name="size"
-            aria-label="Desired size"
+            aria-label="Exact size"
             value={formData.size}
             onChange={(size) => onChange({ size })}
             validated={errors.size && 'error'}
--- a/web/src/components/storage/VolumeForm.jsx
+++ b/web/src/components/storage/VolumeForm.jsx
@@ -240,7 +240,7 @@ const SizeOptions = ({ errors, formData, volume, onChange }) => {
   const { sizeMethod } = formData;
   const sizeWidgetProps = { errors, formData, volume, onChange };
 
-  const sizeOptions = [SIZE_METHODS.MANUAL, SIZE_METHODS.RANGE];
+  const sizeOptions = [SIZE_METHODS.RANGE, SIZE_METHODS.MANUAL];
 
   if (volume.adaptiveSizes) sizeOptions.unshift(SIZE_METHODS.AUTO);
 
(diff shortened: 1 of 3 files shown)


web/src/components/storage/VolumeForm.jsx
```diff
@@ -240,7 +240,7 @@ const SizeOptions = ({ errors, formData, volume, onChange }) => {
   const { sizeMethod } = formData;
   const sizeWidgetProps = { errors, formData, volume, onChange };
 
-  const sizeOptions = [SIZE_METHODS.MANUAL, SIZE_METHODS.RANGE];
+  const sizeOptions = [SIZE_METHODS.RANGE, SIZE_METHODS.MANUAL];
 
   if (volume.adaptiveSizes) sizeOptions.unshift(SIZE_METHODS.AUTO);
```
dgdavid, line 245: Another option could be going from "Fixed" to "Auto". I.e., 


```suggestion
  const sizeOptions = [SIZE_METHODS.MANUAL, SIZE_METHODS.RANGE];

  if (volume.adaptiveSizes) sizeOptions.push(SIZE_METHODS.AUTO);
```

| Proposed now | Another proposal |
|-|-|
| Screen Shot 2023-06-13 at 07 47 30 | Screen Shot 2023-06-13 at 07 56 52 |

shundhammer:
BTW I am also not sure about our consistency with "file system" vs. "filesystem". IIRC we use both. Sadly, there does not seem to be a clear standard anywhere on the web for it; both are in widespread use.

See https://www.google.com/search?client=opera&q="file+system"+vs.+"filesystem"&sourceid=opera&ie=UTF-8&oe=UTF-8

shundhammer:
BTW always keep in mind that a large part of our user base uses English despite not being native speakers, so more often than not it's better to keep wordings simple so those users have no doubt what our messages mean.

web/src/components/storage/VolumeForm.jsx
joseivanlopez: Now that you mention it, I would vote for "Fixed", "Range", "Auto". Note that "Auto" is only shown for some volumes (typically root) so the other two options will always be at the same position independently on the presence of "Auto".

joseivanlopez (approved):
LGTM

I left a comment about the order of the options but I would not block the PR. Feel free to adapt it if you agree with my comment.

PR #620, opened 2023-06-13
[web] Drop heigh prop from Popup component

dgdavid:
## Problem

Sometimes a change in its content causes a modal dialog to have scrollable content even when the viewport is big enough to hold all the content.

This happens because Agama limits dialog height via the Popup/height prop since 8b8de4c. In order to keep the dialog visually stable, we intentionally prevent the dialog from changing size as content changes.

But turned out that actually it is worse as it is now than just giving it the freedom to adapt to the content as long as the viewport height permits it.

## Solution

Drop and stop using the `Popup/height` prop and use only the `fullsize` CSS class when is need to force it to occupy as much size as possible.

## Testing

- Tested manually

## Screenshots

N/A. 

*Reviewers*: Please, give it a shot by your self if you want to check the behavior.


joseivanlopez (approved):
LGTM

PR #622, opened 2023-06-13
Added support for showing the network configuration through the CLI

teclator:
## Problem

The CLI should support reading/writing the network configuration. As it is not trivial, let’s start by just exposing the current configuration through the agamactl config show command.

As a first step, we should define how the configuration should look like in Jsonnet as it would affect the internal representation. Once that’s clear, it is a matter of implementing support for reading such settings over D-Bus.

- [*PBI*](https://trello.com/c/MowTUSWi/3355-5-agama-export-the-network-configuration-in-the-cli)


rust/agama-lib/src/network/proxies.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn remove_connection(&self, uuid: &str) -> zbus::Result<()>;
+}
+
+#[dbus_proxy(
+    interface = "org.opensuse.Agama.Network1.Connection.Wireless",
+    default_service = "org.opensuse.Agama.Network1",
+    default_path = "/org/opensuse/Agama/Network1"
+)]
+trait Wireless {
+    /// Mode property
```
imobachgs, line 31: What the "Mode" means? "Mode property" is not informative at all 😉

imobachgs (commented):
In general, it looks good. But, please, improve the documentation.

Anyway, great job! :-)

PS: I fixed the `nm::dbus` tests, btw.

rust/agama-lib/share/examples/profile.json
```diff
@@ -21,5 +21,20 @@
   "root": {
     "password": "nots3cr3t",
     "sshKey": "..."
+  },
+  "network": {
+    "connections": [
+      {
+        "name": "Ethernet network device 1",
+        "method": "manual",
```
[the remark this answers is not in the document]
teclator: So far it is based on what we have implemented and there is no need to have a different configuration in the profile. So by now it is planned to use the same settings for both.

rust/agama-lib/share/examples/profile.jsonnet
```diff
@@ -29,4 +29,27 @@ local findBiggestDisk(disks) =
...
+      {
+        name: 'AgamaNetwork',
+        wireless: {
+          password: 'agama.test',
+          security: 'wpa-psk',
+          ssid: 'AgamaNetwork'
+        }
+      },
+      {
+        name: "Etherned device 1",
```
imobachgs, line 43: It would be nice to be consistent and use single quotes always.

rust/agama-lib/src/network/client.rs
```diff
@@ -38,6 +39,10 @@ impl<'a> NetworkClient<'a> {
 
         Ok(connections)
     }
+
+    // It returns the NetworkConnection for the given connection path
+    //
+    //  `path`: the connections path to get the config from
```
imobachgs, line 45:
```suggestion
    //  * `path`: the connections path to get the config from
```

rust/agama-lib/src/network/client.rs
```diff
@@ -79,6 +84,8 @@ impl<'a> NetworkClient<'a> {
         })
     }
 
+    // It resturs the [wireless settings][WirelessSettings] for the given connection path
+    //  `path`: the connections path to get the wireless config from
```
imobachgs, line 88:
```suggestion
    //
    //  * `path`: the connections path to get the wireless config from
```

rust/agama-lib/src/network/client.rs
```diff
@@ -22,6 +22,7 @@ impl<'a> NetworkClient<'a> {
         })
     }
 
+    // It returns an array of NetworkConnection using the connections_proxy
```
imobachgs, line 25:
```suggestion
    // Returns an array of network connections
```

Note: Whether it uses a proxy, is an implementation detail.

rust/agama-lib/src/network/client.rs
```diff
@@ -79,6 +84,8 @@ impl<'a> NetworkClient<'a> {
         })
     }
 
+    // It resturs the [wireless settings][WirelessSettings] for the given connection path
```
imobachgs, line 87:
```suggestion
    // Returns the [wireless settings][WirelessSettings] for the given connection
```

imobachgs (approved):
I have added a few suggestions to make the documentation consistent with the existing one. Otherwise, LGTM.

PR #626, opened 2023-06-14
Extend zFCP D-Bus API

joseivanlopez:
## Problem

Agama already provides a D-Bus API for managing zFCP devices. But the current API offers nothing to check whether a controller is automatically scanning LUNs. That info is important for clients, because the actions over a LUN depend on that config.

For example, if a controller is not doing auto LUN scan, then LUNs have to be manually activated one by one. But, if auto LUN scan is active, then all LUNs are activated as side effect of activating a controller. Moreover, automatically activated LUNs cannot be manually deactivated.

## Solution

Extend D-Bus API. Now, *Agama.Storage1.ZFCP.Manager* interface has the *AllowLUNScan* attribute, which indicates whether *allow_lun_scan* is active in the system. And *Agama.Storage1.ZFCP.Controller* interface has the *LUNScan* attribrute, which indicates whether the controller is automatically scanning LUNs. Note that the controller must be running in NPIV mode.

Requires https://github.com/yast/yast-s390/pull/106.

Note: *allow_lun_scan* is active by default in the zfcp module since SLE 12. A kernel parameter can be used to disable it (i.e., *zfcp.allow_lun_scan=0*). If it is used, then it should be persisted in the target system, see https://github.com/yast/yast-bootloader/pull/687. 

Note: some CI tasks are failing because a coveralls error, but all checks actually pass. 

## Testing

* Added new unit tests
* Tested manually

### DIFF
--- a/doc/dbus/bus/org.opensuse.Agama.Storage1.ZFCP.Controller.bus.xml
+++ b/doc/dbus/bus/org.opensuse.Agama.Storage1.ZFCP.Controller.bus.xml
@@ -49,6 +49,7 @@
       <arg name="result" direction="out" type="u"/>
     </method>
     <property type="b" name="Active" access="read"/>
+    <property type="b" name="LUNScan" access="read"/>
     <property type="s" name="Channel" access="read"/>
   </interface>
 </node>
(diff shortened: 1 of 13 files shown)


teclator (approved):
LGTM

PR #594, opened 2023-05-29
zFCP D-Bus API

joseivanlopez:
## Problem

The storage service already provides a D-Bus API for managing iSCSI and DASD devices, but zFCP is not supported yet.

## Solution

Add D-Bus API for managing zFCP devices (requires https://github.com/yast/yast-s390/pull/105).

<details>
<summary>Show API</summary>

~~~
s390vsl111:~ # busctl --address unix:path=/run/agama/bus tree org.opensuse.Agama.Storage1
└─/org
  └─/org/opensuse
    └─/org/opensuse/Agama
      └─/org/opensuse/Agama/Storage1
        ├─/org/opensuse/Agama/Storage1/zfcp_controllers
        │ ├─/org/opensuse/Agama/Storage1/zfcp_controllers/1
        │ └─/org/opensuse/Agama/Storage1/zfcp_controllers/2
        └─/org/opensuse/Agama/Storage1/zfcp_disks
          ├─/org/opensuse/Agama/Storage1/zfcp_disks/1
          └─/org/opensuse/Agama/Storage1/zfcp_disks/2

s390vsl111:~ # busctl --address unix:path=/run/agama/bus introspect org.opensuse.Agama.Storage1 /org/opensuse/Agama/Storage1 org.opensuse.Agama.Storage1.ZFCP.Manager
NAME                                     TYPE      SIGNATURE RESULT/VALUE FLAGS
.Probe                                   method    -         -            -

s390vsl111:~ # busctl --address unix:path=/run/agama/bus introspect org.opensuse.Agama.Storage1 /org/opensuse/Agama/Storage1/zfcp_controllers/1 org.opensuse.Agama.Storage1.ZFCP.Controller
NAME                                        TYPE      SIGNATURE RESULT/VALUE FLAGS
.Activate                                   method    -         u            -
.ActivateDisk                               method    ss        u            -
.DeactivateDisk                             method    ss        u            -
.GetLUNs                                    method    s         as           -
.GetWWPNs                                   method    -         as           -
.Active                                     property  b         true         emits-change
.Channel                                    property  s         "0.0.fa00"   emits-change

s390vsl111:~ # busctl --address unix:path=/run/agama/bus introspect org.opensuse.Agama.Storage1 /org/opensuse/Agama/Storage1/zfcp_disks/1 org.opensuse.Agama.Storage1.ZFCP.Disk
NAME                                  TYPE      SIGNATURE RESULT/VALUE         FLAGS
[...]

PR #628, opened 2023-06-14
Fix D-Bus docs after ZFCP and Questions changes

mvidner:
## Problem

After merging the ZFCP changes (#594) to master, CI fails because it thinks the D-Bus API and its docs have diverged. Well it has, but...

## Solution

- [ ] seed the new API. TODO add the seeding to CI
- [x] document new API... kindof
- [x] run that check in the branches already, not only as part of publishing the docs in master

The diff is messy because Ruby and Rust have random differences in equivalent introspection XML :-/
Reviewing by commits helps that a bit.

## Testing

- existing CI tests


doc/dbus/bus/org.opensuse.Agama.Questions1.LuksActivation.bus.xml
```diff
@@ -1,41 +1,53 @@
...
+    </method>
+  </interface>
+  <interface name="org.freedesktop.DBus.Peer">
+    <method name="Ping">
+    </method>
+    <method name="GetMachineId">
+      <arg type="s" direction="out"/>
+    </method>
+  </interface>
+  <interface name="org.opensuse.Agama.Questions1.Generic">
```
joseivanlopez, line 17: *Questions1.Generic* is defined three times (*Questions1.bus.xml*, *Questions1.Generic.bus.xml* and *Questions1.LuksActivation.bus.xml*). Is that expected? I am not sure if that is the best way for documenting objects with dynamic interfaces.

joseivanlopez: In fact, I think the file names could be misleading. For example, I would expect that *doc/dbus/bus/org.opensuse.Agama.Questions1.Generic.bus.xml* defines the *Questions1.Generic* interface, but it defines all the interfaces of an object that only includes *Questions1.Generic* interface.

I would expect to have both, object docs and interface docs. And an object doc only includes the name of the interfaces, for example:

~~~
doc/dbus/objects/question/generic.xml
<node>
  <interface name="org.freedesktop.DBus.Introspectable" />
  <interface name="org.freedesktop.DBus.Peer" />
  <interface name="org.opensuse.Agama.Questions1.Generic" />
</node

doc/dbus/objects/question/luks_activation.xml
<node>
  <interface name="org.freedesktop.DBus.Introspectable" />
  <interface name="org.freedesktop.DBus.Peer" />
  <interface name="org.opensuse.Agama.Questions1.Generic" />
  <interface name="org.opensuse.Agama.Questions1.LuksActivation" />
</node

doc/dbus/interfaces/org.opensuse.Agama.Questions1.Generic.xml
<interface name="org.opensuse.Agama.Questions1.Generic">
  <property name="Answer" type="s" access="readwrite"/>
  <property name="DefaultOption" type="s" access="read"/>
  <property name="Id" type="u" access="read"/>
  <property name="Options" type="as" access="read"/>
  <property name="Text" type="s" access="read"/>
</interface>
~~~

In general, something similar to what [UDisks2](http://storaged.org/doc/udisks2-api/latest/ref-dbus.html) does. What do you think?

mvidner: > I would expect to have both, object docs and interface docs.

Very good point, I have so far concentrated on the automated part and limited the output to the interfaces.
We also need documenting the objects (object trees) and which interfaces they implement. Probably some manually written pieces.

PR #629, opened 2023-06-14
Add initial logging capability to agama-dbus-server

jreidinger:
## Problem

Currently agama-dbus-service does not provide any logging. So troubleshooting is hard.

trello: https://trello.com/c/qjrhMrmi/3378-2-agama-enable-a-logging-mechanism-in-agama-dbus-server

## Solution

Add initial logging capability with systemd logger to be consistent with ruby counterpart that also logs to journal.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/main.rs
+++ b/rust/agama-dbus-server/src/main.rs
@@ -4,15 +4,22 @@ pub mod questions;
 
 use agama_network::NetworkService;
 use std::future::pending;
+use systemd_journal_logger::JournalLog;
+use log::LevelFilter;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    JournalLog::default().install().unwrap();
+    log::set_max_level(LevelFilter::Info);
     // When adding more services here, the order might be important.
     crate::questions::start_service(ADDRESS).await?;
+    log::info!("Started questions interface");
     let _conn = crate::locale::start_service(ADDRESS).await?;
+    log::info!("Started locale interface");
     NetworkService::start(ADDRESS).await?;
+    log::info!("Started network interface");
 
     // Do other things or go to wait forever
     pending::<()>().await;
--- a/rust/agama-dbus-server/src/main.rs
+++ b/rust/agama-dbus-server/src/main.rs
@@ -4,15 +4,25 @@ pub mod questions;
 
 use agama_network::NetworkService;
 use std::future::pending;
-use systemd_journal_logger::JournalLog;
 use log::LevelFilter;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    JournalLog::default().install().unwrap();
-    log::set_max_level(LevelFilter::Info);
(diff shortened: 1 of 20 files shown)


mvidner:
> Add initial logging capability with systemd logger to be consistent with ruby counterpart that also logs to journal.

No. AFAIK the Ruby part simply logs to *stderr* and it is the controlling daemons, `systemd` and `dbus-daemon`, that forward the output to the journal

rust/agama-dbus-server/src/main.rs
```diff
@@ -4,15 +4,22 @@ pub mod questions;
...
 use agama_network::NetworkService;
 use std::future::pending;
+use systemd_journal_logger::JournalLog;
+use log::LevelFilter;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    JournalLog::default().install().unwrap();
```
mvidner, line 14: `unwrap`...
should be used when you *know* better than the compiler that it cannot fail

The function is already returning a `Result`

```suggestion
    JournalLog::default().install()?;
```

mvidner: ... and it turns out that you are right, it fails only if the logger has been set already
https://docs.rs/log/0.4.17/log/fn.set_boxed_logger.html

In that case I'd add a comment *explaining* why it can't fail. Might as well keep the `?`.

jreidinger: ok, will document it. As I am not big fan of Box<Error> due to its wide scope and with unwrap it will make it change to more precise error definition easier.

mvidner (approved):
Thanks!

rust/agama-dbus-server/src/main.rs
```diff
@@ -4,15 +4,32 @@ pub mod questions;
...
 use agama_network::NetworkService;
 use std::future::pending;
+use log::LevelFilter;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // be smart with logging and log directly to journal if connected to it
+    if systemd_journal_logger::connected_to_journal(){
```
mvidner, line 14:
```suggestion
    if systemd_journal_logger::connected_to_journal() {
```

That reminds me, I don't know how to check for formatting problems. How? Should we make it mandatory?

jreidinger: well, I again forget to run `rustfmt` before submit. I will try to configure my IDE to do it

jreidinger: btw I can add it to CI with something like `find rust/agama-*/src/ | grep '.rs$' | xargs rustfmt --check --edition 2021`

rust/agama-dbus-server/src/main.rs
mvidner: > As I am not big fan of Box due to its wide scope and with unwrap it will make it change to more precise error definition easier.

Just some nitpicking while I wait for the `rustfmt` CI run :smile: 

I think it is the `dyn` part, not the `Box`, that you object to :)

In general yes, it makes the error API too catch-all... but then we are in `fn main` and we can only crash... which also refutes my initial objection :rofl:

PR #633, opened 2023-06-16
Rust formatting in CI

jreidinger:
## Problem

Often we have formatting issue with rust code

## Solution

Try to add to ci running of rustfmt.

## Testing

CI should pass ( does not work yet )

@imobachgs maybe you have idea how to handle it as [stable rustfmt is ancient 0.10.0](https://crates.io/crates/rustfmt) that cannot do 'check' and [nightly](https://crates.io/crates/rustfmt-nightly) uses features from non stable channel.


imobachgs:
I use `rustup` to set up my Rust environment (actually, `rustup` is available as an openSUSE package). Running `rustup component add rustfmt` should do the trick.

```
$ rustup component add rustfmt
info: downloading component 'rustfmt'
info: installing component 'rustfmt'
$ rustfmt --version
rustfmt 1.5.2-stable (90c5418 2023-05-31)
```

.github/workflows/ci.yml
```diff
@@ -202,6 +202,9 @@ jobs:
     - name: Rust unit tests
       run:  podman exec agama bash -c "cd /checkout/rust; cargo test --verbose"
 
+    - name: Check rust formatting
+      run:  podman exec agama bash -c "cd /checkout/rust; rustup componet add rustfmt; cargo fmt --all -- --check"
```
imobachgs, line 206: It looks like there is a typo: It should be "component" (instead of "componet").

jreidinger: yeah, I tried it in small spare time at the end of night shift and make a typo. Will try fixed version now.

.github/workflows/ci.yml
```diff
@@ -161,6 +161,44 @@ jobs:
...
+      image: registry.opensuse.org/opensuse/tumbleweed:latest
+
+    steps:
+
+    - name: Git Checkout
+      uses: actions/checkout@v3
+
+    - name: Configure and refresh repositories
+      # disable unused repositories to have faster refresh
+      run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && ( zypper ref || zypper ref || zypper ref )
```
mvidner, line 186: Please comment which monster from another dimension is invoked by speaking `zypper ref` three times ! ! !

mvidner: OK, I see it is copied from other jobs, but still it would be nice to know. I prefer Cth*lhu to Candym*an. @lslezak ?

mvidner:
Finally the CI is failing in a good way, complaining about code formatting, yay!! :sob: :rofl:

.github/workflows/ci.yml
jreidinger: I would say that @lslezak do it to prevent ref failures

jreidinger:
I push formatting fix to this branch.
@imobachgs  I think it is from your changes. I think your editor does not catch it somehow.

imobachgs (approved):

PR #635, opened 2023-06-20
Draft for FCoE D-Bus interface

ancorgs:
This pull request is the skeleton for a D-Bus interface that would allow the same FCoE operations currently supported by the YaST installer. In a nutshell, this offers a list of the network interfaces of the system, allowing to create an FCoE VLAN interface for any of them. It also makes it possible to modify the configuration of such a FCoE VLAN and to delete it.

The line of work was abandoned because it was kind of agreed that all those explicit operations should not be needed anymore in any system that is currently in production. All FCoE systems currently in existence work out-of-the-box without intervention from the installer.

This needs the following patches to yast2-fcoe-client: https://github.com/yast/yast-fcoe-client/pull/70

The current implementation for modifying the properties of an FCoE VLAN (`fcoe_service`, `dcb_service` and `auto_vlan`) is quite inefficient. Modifying them should not immediately introduce any change in the system or any restart of the services. Instead, we should add an explicit D-Bus call to write the changes for a given interface (so first the properties are set and then the changes are "applied" by calling that method). But doing so would implies more changes in yast2-fcoe-client to offer that per-interface `write` (something that is perfectly doable).

### DIFF
--- a/service/lib/agama/dbus/storage/fcoe_interface.rb
+++ b/service/lib/agama/dbus/storage/fcoe_interface.rb
@@ -1,0 +1,144 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2023] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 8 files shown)


mvidner:
@ancorgs Do we have links to Bugzilla or Jira for this? Or Trello but its card could be archived (=lost) by the time we eventually resurrect this.

imobachgs:
We are not implementing FCoE support so I am closing this PR.

PR #636, opened 2023-06-21
Reorganize agama-lib

imobachgs:
The `agama-lib` package is growing with support for more settings and services. This PR proposes to change the approach a bit and organize the code using a different module for each topic (users, storage, etc.). So now we have `network`, `users`, `software` and `storage` modules which contain the logic specific to those topics.

We introduce this approach in #622 and it seems to work better than the previous one.


rust/agama-lib/src/software/settings.rs
```diff
@@ -0,0 +1,13 @@
+//! Representation of the software settings
+
+use crate::settings::{Settings, SettingValue};
```
jreidinger, line 3: I am not sure why this use is needed. I do not see its usage. I only see `Settings` from agama_derive

jreidinger (approved):

rust/agama-lib/src/network/settings.rs
```diff
@@ -1,6 +1,5 @@
-//! Configuration network settings handling
-//!
-//! This module implements the mechanisms to load and store the installation settings.
+//! Representation of the software settings
```
mvidner, line 1:
```suggestion
//! Representation of the network settings
```

mvidner (commented):
I want to understand the genaral meaning of Store, Settings, Client better.
I'll render the docs and look there.

Only Proxy is clear: the API  exactly corresponding to D-Bus

rust/agama-lib/src/software/store.rs
```diff
@@ -1,6 +1,7 @@
+//! Implements the store for the storage settings.
```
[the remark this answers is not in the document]
imobachgs: I preferred to keep three different concepts, but sure we can merge Client and Store (and would keep Settings alone).

* Settings: represents the settings themselves, which can be (de)serialized.
* Clients: basic and kind of "agnostic". They know nothing about the settings.
* Stores: as you said, the connector that can read/write the settings relying on clients.

I have more ideas in mind to improve the code. For instance, I think we could create a trait (e.g., `Store<T>`) for stores and so on. But it is out of scope for this refactoring.

rust/agama-lib/src/software/settings.rs
imobachgs: Because it is used in the derive macro. I have fixed it, so it is not needed anymore.

rust/agama-lib/src/software/store.rs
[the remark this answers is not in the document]
imobachgs: I will write this information as part of the crate's documentation.

mvidner (approved):
Thanks for the description!

PR #639, opened 2023-06-23
Adjust to ruby-dbus-0.23.0.beta2

mvidner:
## Problem

`agama config show` would fail

DBus::BusConnection#initialize now calls send_hello already and the bus will Error when it sees a second one

## Solution

remove `send_hello` from the constructor of our derived class

## Testing

- Tested manually with `agama config show`


joseivanlopez (approved):
LGTM

PR #641, opened 2023-06-29
CLI: update the network configuration

imobachgs:
⚠️ For some reason, we cannot comment on this PR. It is replaced with #647.


imobachgs:
For some reason, we cannot comment on this PR. We get an error: "Line must be part of the diff and Diff hunk can't be blank". We have replaced this one with #647.

PR #642, opened 2023-06-30
Code coverage for the Rust code

imobachgs:
## Problem

We do not have any report about the test coverage for the Rust code.

## Solution

This PR enables code coverage reporting for the Rust code using [Tarpaulin](https://github.com/xd009642/tarpaulin).

* All jobs are using `coverallsapp/github-action@v2` now.
* Use [cargo-binstall](https://github.com/cargo-bins/cargo-binstall) to install Tarpaulin, saving the build time. An alternative approach might be to package Tarpaulin ourselves.
* `rust_ci` uses the yast-ruby container because the Coveralls GitHub Action was not working with the Tumbleweed image (IDK the reason).
* `rust_ci` uses the *cobertura* format instead of *lcov* because the file names in the latter are absolute (which looks wrong). See the [list of supported formats](https://github.com/coverallsapp/coverage-reporter#supported-coverage-report-formats) if you are interested.
* Code coverage is ~15% 😞 

## Screenshots

Screenshot from 2023-06-30 11-42-29

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,9 +171,12 @@ jobs:
 
     strategy:
       fail-fast: false
+      matrix:
+        distro: [ "tumbleweed" ]
 
     container:
-      image: registry.opensuse.org/opensuse/tumbleweed:latest
+      image: registry.opensuse.org/yast/head/containers_${{matrix.distro}}/yast-ruby
+      options: --security-opt seccomp=unconfined
 
     steps:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -93,12 +92,12 @@ jobs:
...
 
     - name: Coveralls GitHub Action
-      uses: coverallsapp/github-action@v1
+      uses: coverallsapp/github-action@v2
       with:
         base-path: ./service
-        path-to-lcov: ./service/coverage/lcov.info
         flag-name: backend
         parallel: true
+        debug: true
```
jreidinger, line 100: do we need it in final configuration?

.github/workflows/ci.yml
```diff
@@ -189,17 +191,34 @@ jobs:
...
     - name: Install Rust toolchains
       run: rustup toolchain install stable
 
+    - name: Install cargo-binstall
+      uses: taiki-e/install-action@v2
+      with:
+        tool: cargo-binstall
+
+    - name: Install Tarpaulin (for code coverage)
+      run: cargo-binstall --no-confirm cargo-tarpaulin
```
jreidinger, line 205: so we have third way to install rust stuff? rustup component, cargo install and now cargo-binstall? It start to be quite messy

jreidinger (approved):

.github/workflows/ci.yml
imobachgs: Just to clarify: rustup is for managing the rust toolchain. `cargo install` is for binary crates (building for sources) and cargo-binstall to avoid building for sources.

PR #643, opened 2023-07-03
Bring back the rb_build_versions definition

imobachgs:
By mistake, we removed the `rb_build_versions` macro. We should bring it back.


joseivanlopez (approved):
LGTM

PR #647, opened 2023-07-04
CLI: write network configuration

imobachgs:
## Problem

#622 added support to display the network configuration through the CLI: `agama config show`. However, writing the configuration was still missing.

Trello card: https://trello.com/c/vVe8CboU/3354-8-agama-set-the-network-configuration-using-a-jsonnet-file

## Solution

This PR implements support to set the network configuration using the CLI. But, as usual, we took the chance to do some refactoring and clean-up.

## Note: naming network connections

Agama's network service API works with connection names (Id) instead of UUIDs. In NetworkManager, you can use any of them. However, we prefer our users to work with Ids, not UUIDs, because they are more human-readable. The problem is that, in NetworkManager, the Id is not unique.

This PR modifies Agama to rename duplicated network connections. For instance, if we have multiple connections with the same name (e.g., `virbr0`), Agama will add a suffix to the duplicates (e.g., `virbr0-1`, `virbr0-2`, etc.).

## Testing

- [x] Added (a few) unit tests
- [x] Tested manually

## Future

We would like to implement a few changes, but we should address them separately.

- Bound a connection to a specific network interface. By now, NetworkManager chooses the device automatically.
- 
- The software service might need to wait until the network is ready.
- Do not update those connections that did not change.

### DIFF
--- a/rust/agama-dbus-server/src/network/action.rs
+++ b/rust/agama-dbus-server/src/network/action.rs
@@ -1,4 +1,5 @@
-use crate::network::model::{Connection, DeviceType};
+use crate::network::model::Connection;
+use agama_lib::network::types::DeviceType;
 
 /// Networking actions, like adding, updating or removing connections.
 ///
--- a/rust/agama-dbus-server/src/network/action.rs
+++ b/rust/agama-dbus-server/src/network/action.rs
@@ -1,5 +1,4 @@
 use crate::network::model::{Connection, DeviceType};
-use uuid::Uuid;
 
 /// Networking actions, like adding, updating or removing connections.
 ///
@@ -12,7 +11,7 @@ pub enum Action {
     /// Update a connection (replacing the old one).
     UpdateConnection(Connection),
     /// Remove the connection with the given Uuid.
-    RemoveConnection(Uuid),
+    RemoveConnection(String),
     /// Apply the current configuration.
     Apply,
 }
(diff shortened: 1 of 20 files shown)


rust/agama-dbus-server/src/network/action.rs
```diff
@@ -1,5 +1,5 @@
-use crate::network::model::{Connection, DeviceType};
-use uuid::Uuid;
+use crate::network::model::Connection;
+use agama_lib::network::types::DeviceType;
```
imobachgs, line 2: x

imobachgs, line 2: test

imobachgs, line 2: dsadsa

doc/dbus/org.opensuse.Agama.Network1.Connection.IPv4.doc.xml
```diff
@@ -3,9 +3,30 @@
  "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
 <node>
   <interface name="org.opensuse.Agama.Network1.Connection.IPv4">
-    <property name="Addresses" type="a(su)" access="readwrite"/>
+    <!--
+     List of IP addresses.
```
imobachgs, line 7: x

rust/agama-dbus-server/src/network/action.rs
```diff
@@ -1,5 +1,5 @@
-use crate::network::model::{Connection, DeviceType};
-use uuid::Uuid;
+use crate::network::model::Connection;
```
imobachgs, line 1: xx

joseivanlopez (approved):
LGTM

PR #648, opened 2023-07-04
Update the org.opensuse.Agama.Network1 documentation

imobachgs:
Update the org.opensuse.Agama.Network1 D-Bus service documentation.

### DIFF
--- a/doc/dbus/bus/org.opensuse.Agama.Network1.Device.bus.xml
+++ b/doc/dbus/bus/org.opensuse.Agama.Network1.Device.bus.xml
@@ -39,7 +39,19 @@
     </signal>
   </interface>
   <interface name="org.opensuse.Agama.Network1.Device">
+    <!--
+     Device name.
+
+     Kernel device name, e.g., eth0, enp1s0, etc.
+     -->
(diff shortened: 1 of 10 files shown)


doc/dbus/bus/org.opensuse.Agama.Network1.Device.bus.xml
```diff
@@ -39,7 +39,19 @@
...
+     Kernel device name, e.g., eth0, enp1s0, etc.
+     -->
     <property name="Name" type="s" access="read"/>
+    <!--
+     Device type.
+
+     Possible values: 0 = loopback, 1 = ethernet, 2 = wireless.
+
+     See [crate::model::DeviceType].
+     -->
```
imobachgs, line 54: x

joseivanlopez (approved):
LGTM

I think we have to improve our doc mechanism to avoid repetition. For example, the *org.opensuse.Agama.Network1.Connection.Wireless* interface is defined 4 times. BTW, maybe we only need the *doc/dbus/bus* files if no extra info is going to be added.

PR #649, opened 2023-07-05
Add support for questions with no default options

imobachgs:
The Questions service crashes if no default options are given. It was supported in the Ruby-based version, so we need to fix the new service to support it.

According to a comment on the code, we should stop using an array for the default options, forcing a default option anyway. But let's discuss that later.


teclator (approved):
LGTM. I tested the patch and it fixes the issue, thanks!! :partying_face:

imobachgs:
Thanks for testing it!

PR #650, opened 2023-07-05
zFCP improvements

joseivanlopez:
## Problem

Two things to fix/improve:

* Automatically activated zFCP disks are not probed after activating a controller.
* There is no info about allow_lun_scan when the option is disabled.

## Solution

* Delay zFCP probing in order to give some time to disk activation.
* The system is marked as deprecated after probing, if needed.
* allow_lun_scan status is always informed.
* The allow_lun_scan explanation is improved.

Bonus: members of a device are sorted in the disk selector. 

## Testing

* Adapted unit tests
* Tested manually

## Screenshots

<details>
<summary>Show/hide</summary>

localhost_8080_ (13)

localhost_8080_ (14)

</details>

### DIFF
--- a/service/lib/agama/storage/zfcp/manager.rb
+++ b/service/lib/agama/storage/zfcp/manager.rb
@@ -91,13 +91,22 @@ module Agama
 
         # Activates the controller with the given channel id
         #
-        # @note: If "allow_lun_scan" is active, then all LUNs are automatically activated.
+        # @note: If "allow_lun_scan" is active, then all its LUNs are automatically activated.
         #
         # @param channel [String]
         # @return [Integer] Exit code of the chzdev command (0 on success)
         def activate_controller(channel)
           output = yast_zfcp.activate_controller(channel)
-          update_disks if output["exit"] == 0
+          if output["exit"] == 0
+            # LUNs activation could delay after activating the controller. This usually happens when
+            # activating a controller for first time because some SCSI initialization. Probing the
+            # disks should be done after all disks are activated.
+            #
+            # FIXME: waiting 2 seconds should be enough, but there is no guarantee that the disks
+            # are actually exported.
+            sleep(2)
+            update_disks
+          end
           output["exit"]
         end
 
--- a/service/lib/agama/storage/zfcp/manager.rb
+++ b/service/lib/agama/storage/zfcp/manager.rb
@@ -21,6 +21,7 @@
 
 require "yast"
 require "y2s390/zfcp"
+require "y2storage/storage_manager"
 require "agama/storage/zfcp/controller"
 require "agama/storage/zfcp/disk"
 
@@ -54,14 +55,20 @@ module Agama
 
         # Probes zFCP
         #
-        # Callbacks are called at the end, see {#on_probe}.
+        # Callbacks {#on_probe} are called after probing, and callbacks {#on_disks_change} are
+        # called if there is any change in the disks.
         def probe
           logger.info "Probing zFCP"
 
+          previous_disks = disks
           yast_zfcp.probe_controllers
           yast_zfcp.probe_disks
 
           @on_probe_callbacks&.each { |c| c.call(controllers, disks) }
+
+          return unless disks_changed?(previous_disks)
+
+          @on_disks_change_callbacks&.each { |c| c.call(disks) }
         end
 
         # zFCP controllers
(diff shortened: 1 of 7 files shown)


service/package/rubygem-agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jul  5 14:02:23 UTC 2023 - José Iván López González <jlopez@suse.com>
+
+- Delay zFCP probing after activating a controller.
```
imobachgs, line 4:
```suggestion
- Delay zFCP probing after activating a controller
```

service/package/rubygem-agama.changes
```diff
@@ -1,3 +1,9 @@
...
+- gh#openSUSE/agama#650
```
imobachgs, line 5:
```suggestion
  (gh#openSUSE/agama#650).
```

web/package/cockpit-agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jul  5 13:59:58 UTC 2023 - José Iván López González <jlopez@suse.com>
+
+- Add info about deactivated zFCP auto_lun_scan.
+- Sort members in disk selector.
```
imobachgs, line 5:
```suggestion
- Sort members in disk selector (gh#openSUSE/agama#650).
```

web/package/cockpit-agama.changes
```diff
@@ -1,3 +1,10 @@
...
+- gh#openSUSE/agama#650
```
imobachgs, line 6: ... and remove this line.

imobachgs: Or if you prefer, something like this:

```
- Improvements to the zFCP handling (gh#openSUSE/agama#650):
  * Add info about deactivated zFCP auto_lun_scan.
  * Sort members in disk selector.
```

web/src/components/storage/ZFCPPage.jsx
```diff
@@ -422,26 +422,31 @@ const ControllersSection = ({ client, manager, load = noop, isLoading = false })
...
+          condition={allowLUNScan}
+          then={
+            <p>
+              Automatic LUN scan is <b>enabled</b>. Activating a controller which is running in NPIV
+              mode will automatically configures all its LUNs.
+            </p>
+          }
+          else={
+            <p>
+              Automatic LUN scan is <b>disabled</b>. LUNs have to be manually configured after
```
teclator, line 437: :+1:

joseivanlopez:
> What does it happen if it takes more than 2 seconds? Would the user eventually see the list of disks? And, in that case, can the user select one of them for installation?

According to our tests, 2 seconds should be more than enough. Anyway, if exporting disks takes longer and the disks are not read, then zFCP has to be probed again. Right now there nothing in the UI for doing the reprobing. To force a zFCP probing you have to leave the zFCP page (going back) and visit it again. Once the list of zFCP disk is read, the system is properly marked as deprecated.

As said, this should not happen with the sleep delay. And I would avoid adding a button for manually re-read zFCP only just in case. What do you think?

service/lib/agama/storage/zfcp/manager.rb
```diff
@@ -54,14 +55,20 @@ def on_disks_change(&block)
...
 
+          previous_disks = disks
           yast_zfcp.probe_controllers
           yast_zfcp.probe_disks
 
           @on_probe_callbacks&.each { |c| c.call(controllers, disks) }
+
+          return unless disks_changed?(previous_disks)
+
+          @on_disks_change_callbacks&.each { |c| c.call(disks) }
```
imobachgs, line 71: np: Is it possible that `@on_disks_change_callbacks` is nil? Not something to change in this PR, but I rather prefer to have an empty array.

imobachgs (approved):

service/lib/agama/storage/zfcp/manager.rb
joseivanlopez: Yes, because is initialize on assignement, see https://github.com/openSUSE/agama/blob/master/service/lib/agama/storage/zfcp/manager.rb#L51. We could avoid it by initializing the list in the constructor or by adding a `#on_disk_change_callbacks` method.

PR #655, opened 2023-07-10
Document the Jsonnet example

imobachgs:
Add some comments to the Jsonnet-based example.


rust/agama-lib/share/examples/profile.jsonnet
```diff
@@ -1,4 +1,10 @@
+// This is a Jsonnet file. Please, check https://jsonnet.org/ for more
+// information about the language.
+
+// The hw.libsonnet file contains hardware information from the "lshw" tool.
```
mvidner, line 4: On  Leap 15.4 at least, lshw.rpm or jsonnet.rpm do not contain a hw.libsonnet file. Where do I find it or how do I know what API it provides?

rust/agama-lib/share/examples/profile.jsonnet
```diff
@@ -1,4 +1,10 @@
+// This is a Jsonnet file. Please, check https://jsonnet.org/ for more
+// information about the language.
```
mvidner, line 2: OK, that's about the syntax.
As a user I'd also want to know the schema, so 

```suggestion
// information about the language.
// For the schema, see
// https://github.com/openSUSE/agama/blob/master/rust/agama-lib/share/profile.schema.json
```

mvidner (approved):
Thanks!

`lshw` makes sense now, it contains the *data* about the current system. Somehow I tricked myself into thinking that it contained jsonnet *code*.

rust/agama-lib/share/examples/profile.jsonnet
```diff
@@ -1,4 +1,18 @@
...
+// For the schema, see
+// https://github.com/openSUSE/agama/blob/master/rust/agama-lib/share/profile.schema.json
+
+// The "hw.libsonnet" file contains hardware information of the storage devices
+// from the "lshw" tool. It is generated by running as root:
```
lslezak, line 7: I think the most important information is that it is explicitly run by Agama itself, there is no magic for that.
https://github.com/openSUSE/agama/blob/b4bf25fdffd46735800a8396f1715fffe181e1db/rust/agama-lib/src/profile.rs#L128

imobachgs: I merged too soon. I opened https://github.com/openSUSE/agama/pull/655.

PR #656, opened 2023-07-10
New storage proposal settings

ancorgs:
## Problem

The information about the products offered for installation is provided to Agama by means of a config file (e.g., */etc/agama.yaml*). That config file contains an *storage* section which indicates the volumes (file systems) to create in the target system. Right now, this storage section is basically the same as the section for the YaST config file. But the storage use cases for Agama are only a subset of all the possibities YaST offers. Because of that, some YaST storage options simply do not make sense for Agama, but the config file still has to indicate them in order to make YaST to work properly. 

We need an easier way to specify the storage options for the Agama use cases. The very YaST specific options should not be exposed in the Agama config file.

* https://trello.com/c/drW2FI0q/3388-8-agama-storage-proposal-settings
*  https://gist.github.com/ancorgs/afaf1e8b9279b05f25ffe6a7bb1efefc
* https://gist.github.com/joseivanlopez/ce50025c2703c7b5ce04b8e0853c0c18

## Solution

The storage settings for Agama config are redefined.

<details>
<summary>
See example
</summary>

~~~
  storage:
    lvm: true
    encryption:
      method: luks2
      pbkd_function: pbkdf2
    space_policy: delete
    volumes:
      - "/"
    volume_templates:
      - mount_path: "/"
        mount_options:
          - default
          - data=ordered
        filesystem: btrfs
        btrfs:
          snapshots: true
          read_only: true
          default_subvolume: "@"
          subvolumes:
            - path: root
            - path: home
            - path: opt
            - path: srv
            - path: var
              copy_on_write: false
        size:
          auto: false
          min: 5 GiB
          max: unlimited
        outline:
          required: true
          filesystems:
            - btrfs
            - ext4
            - xfs
          auto_size:
            base_min: 10 GiB
            base_max: 100 GiB
            snapshots_increment: 200%
            adjust_by_ram: false
            min_fallback_for:
              - "/home"
            max_fallback_for:
              - "/home"
          snapshots_configurable: false
~~~

</details>

The main changes are in the *volumes* section. Now, *volumes* only contains the mount points of the volumes to create:

~~~
volumes:
  - "/"
  - "swap"
~~~

And there is a new *volume_templates* section. This new section contains a template for each volume. A template consists on a set of default values and a volume outline. The outline represents some specific information about the volume which cannot be changed, for example, the possible file system types for the volume. 

~~~
    volume_templates:
      - mount_path: "/"
        filesystem: btrfs
        size:
          auto: false
          min: 5 GiB
          max: unlimited
        outline:
          required: true
          filesystems:
            - btrfs
            - ext4
            - xfs
~~~

There is also a generic volume template to be used for a new arbitrary volume. The generic volume template does not indicate a mount path.

~~~
    volume_templates:
      - filesystem: xfs
        size:
          auto: false
          min: 1 GiB
          max: unlimited
        outline:
          required: false
          filesystems:
            - btrfs
            - ext4
            - xfs
~~~

### Volumes and storage proposal

Agama initially calculates the storage proposal trying to create all the volumes indicated in the *volumes* section of the control file. A new proposal can be triggered from D-Bus calling to `#Calculate` method from `org.opensuse.Agama.Storage1.Proposal.Calculator` interface. `#Calculate` receives the settings to use for the new proposal.

### Proposal settings conversions

In Agama a proposal is calculated by calling to `Agama::Storage::Proposal#calculate`, which receives a `Agama::Storage::ProposalSettings` object. The proposal settings object needs to be generated from two different sources: the agama config file for the initial proposal and from the information provided from D-Bus for calculating a new proposal:

* The class `Agama::Storage::ProposalSettingsReader` generates the proposal settings from the config file.
* There is a conversion class `Agama::DBus::Storage::ProposalSettingsConversion::FromDBus` to generate the proposal settings from the D-Bus information.

And then, Agama delegates the real computation of the proposal to Y2Storage. For that, the proposal settings are converted to the Y2Storage format thanks to the class `Agama::Storage::ProposalSettingsConversion::ToY2Storage`.
 
There are two more conversions for the proposal settings:

* `Agama::DBus::Storage::ProposalSettingsConversion::ToDBus`: generates a D-Bus representation of the proposal settings, useful for the D-Bus interfaces providing information about the proposal.
* `Agama::Storage::ProposalSettingsConversion::FromY2Storage`: generates proposal settings from the Y2Storage format. This is useful to rebuild the proposal settings once Y2Storage has finished to calculate the proposal.

## Testing

* New unit tests.
* Tested manually.


service/etc/agama.yaml
```diff
@@ -20,6 +20,88 @@ web:
...
+      method: luks2
+      pbkd_function: pbkdf2
+    volumes:
+      - "/"
+    volume_templates:
+      - mount_path: "/"
+        filesystem: btrfs
+        btrfs:
+          snapshots: true
+          read_only_root: true
```
ancorgs, line 75: I think the code expects this to be `read_only` instead of `read_only_root`. I'm not sure which one is better, but it should be consistent. Either fix it here, either adapt VolumeTemplatesBuilder.

service/etc/agama.yaml
```diff
@@ -20,6 +20,88 @@ web:
...
+              copy_on_write: false
+        size:
+          auto: false
+          min: 5 GiB
+        outline:
+          required: true
+          filesystems:
+            - btrfs
+          snapshots_configurable: false
+
```
ancorgs, line 104: Currently, it's mandatory to have a default template for arbitrary volumes.

service/etc/agama.yaml
```diff
@@ -20,6 +20,88 @@ web:
...
+    mandatory_patterns:
+      - alp_base_zypper
+      - alp_cockpit
+      - alp_hardware
+    optional_patterns: null # no optional pattern shared
+    mandatory_packages:
+      - package: device-mapper # Apparently needed if devices at /dev/mapper are used at boot (eg. FDE)
+      - package: fde-tools # Needed for FDE with TPM, hardcoded here temporarily
+        archs: aarch64, x86_64
+      - package: libtss2-tcti-device0 # Same than fde-tools
```
jreidinger, line 44: I do not get comment, what means "Same than"? I would use "Same case as fde-tools". But then why fde tools are arch based and this not?

service/etc/agama.yaml
```diff
@@ -20,6 +20,88 @@ web:
   ssl_cert: null
   ssl_key: null
 
+ALP-Dolomite:
```
jreidinger, line 23: why it needs to be moved? ( sorry for question if it is in commit message, but I failed to find way to see commit belonging to final change in pull request and there are too many commits )

service/lib/agama/storage/volume_outline.rb
```diff
@@ -0,0 +1,118 @@
...
+      attr_accessor :base_max_size
+
+      # @return [Boolean]
+      attr_accessor :adjust_by_ram
+      alias_method :adjust_by_ram?, :adjust_by_ram
+
+      # @return [Array<String>] mount paths of other volumes
+      attr_accessor :min_size_fallback_for
+
+      # @return [Array<String>] mount paths of other volumes
```
jreidinger, line 62: copy and paste error. I found it mainly as I found the name `max_size_fallback_for` hard to understand what it means and comment here does not help...and same for attr above.

service/lib/agama/dbus/storage/manager.rb
```diff
@@ -22,14 +22,15 @@
 require "dbus"
 require "yast"
 require "y2storage/storage_manager"
+require "agama/dbus/storage/proposal"
+require "agama/dbus/storage/proposal_settings_conversion"
+require "agama/dbus/storage/volume_conversion"
+require "agama/storage/volume_templates_builder"
```
jreidinger, line 28: why you change ordering? alphabetical one is usually the best one

service/lib/agama/dbus/storage/manager.rb
```diff
@@ -132,26 +132,38 @@ def result
...
         # Calculates a new proposal
         #
         # @param dbus_settings [Hash]
         # @return [Integer] 0 success; 1 error
         def calculate_proposal(dbus_settings)
           logger.info("Calculating storage proposal from D-Bus settings: #{dbus_settings}")
 
-          converter = ProposalSettingsConverter.new
-          success = proposal.calculate(converter.to_agama(dbus_settings))
+          settings = ProposalSettingsConversion.from_dbus(dbus_settings, config: config)
```
jreidinger, line 150: I would probably also log here resulting settings so in logs we see dbus one and converted one for easier debugging.

service/lib/agama/dbus/storage/proposal.rb
```diff
@@ -44,50 +44,76 @@ def initialize(backend, logger)
...
 
-        # Devices used by the storage proposal
+        # Device used as boot device by the storage proposal
         #
-        # @return [Array<String>]
-        def candidate_devices
-          return [] unless backend.calculated_settings
-
-          backend.calculated_settings.candidate_devices
+        # @return [String]
```
jreidinger, line 68: empty string fallback should be documented here what it means.

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
```diff
@@ -0,0 +1,165 @@
...
+          # Performs the conversion from D-Bus format.
+          #
+          # @return [Agama::Storage::ProposalSettings]
+          def convert
+            settings = Agama::Storage::ProposalSettingsReader.new(config).read
+
+            settings.tap do |target|
+              dbus_settings.each do |dbus_property, dbus_value|
+                converter = CONVERSIONS[dbus_property]
+                # FIXME: likely ignoring the wrong attribute is not the best
```
jreidinger, line 51: at least log it, so we can catch typos in logs.

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
```diff
@@ -0,0 +1,165 @@
...
+          # @param value [String]
+          def encryption_password_conversion(target, value)
+            target.encryption.password = value.empty? ? nil : value
+          end
+
+          # @param target [Agama::Storage::ProposalSettings]
+          # @param value [String]
+          def encryption_method_conversion(target, value)
+            method = Y2Storage::EncryptionMethod.find(value.to_sym)
+            return unless method
```
jreidinger, line 109: again I would add here some logging to catch typos in encryption method names

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
```diff
@@ -0,0 +1,165 @@
...
+
+            target.encryption.method = method
+          end
+
+          # @param target [Agama::Storage::ProposalSettings]
+          # @param value [String]
+          def encryption_pbkd_function_conversion(target, value)
+            function = Y2Storage::PbkdFunction.find(value)
+            return unless function
```
jreidinger, line 118: and here

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
```diff
@@ -0,0 +1,165 @@
...
+          end
+
+          # Missing required volumes
+          #
+          # @param required_volumes [Array<Agama::Storage::Volume>]
+          # @param volumes [Array<Agama::Storage::Volume>]
+          #
+          # @return [Array<Agama::Storage::Volume>]
+          def missing_volumes(required_volumes, volumes)
+            mount_paths = volumes.map(&:mount_path)
```
jreidinger, line 157: just small efficiency hint. Set allows much faster searching so with `require "set" you can use  `mount_paths = volumes.map(&:mount_path).to_set` and then include is much faster

jreidinger: Damn, set checks for object equality and not for its value, so it will probably not work here :( https://apidock.com/ruby/v2_5_5/Set/include%3F So lets keep it as it is, I think we should not have too much volumes.

jreidinger: my quick test in irb shows that it should work, but who knows :)

service/lib/agama/dbus/storage/proposal_settings_conversion/to_dbus.rb
```diff
@@ -0,0 +1,60 @@
...
+          def convert # rubocop:disable Metrics/AbcSize
+            {
+              "BootDevice"             => settings.boot_device.to_s,
+              "LVM"                    => settings.lvm.enabled,
+              "SystemVGDevices"        => settings.lvm.system_vg_devices,
+              "EncryptionPassword"     => settings.encryption.password.to_s,
+              "EncryptionMethod"       => settings.encryption.method.id.to_s,
+              "EncryptionPBKDFunction" => settings.encryption.pbkd_function&.value || "",
+              "SpacePolicy"            => settings.space.policy.to_s,
+              "SpaceActions"           => settings.space.actions,
```
jreidinger, line 47: can `settings.space.actions` be nil or is it guarantee that it is "Hash" ?

jreidinger: and same question for system_vg_devices and also if lvm.enabled can be nil or it is always boolean.

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
```diff
@@ -0,0 +1,147 @@
...
+          #
+          # @return [Agama::Storage::Volume]
+          def convert
+            builder = Agama::Storage::VolumeTemplatesBuilder.new_from_config(config)
+            volume = builder.for(dbus_volume["MountPath"] || "")
+
+            volume.tap do |target|
+              dbus_volume.each do |dbus_property, dbus_value|
+                converter = CONVERSIONS[dbus_property]
+                # FIXME: likely ignoring the wrong attribute is not the best
```
jreidinger, line 50: same here. at least log it. Exception and proper error report would be even better to quickly catch it.

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
```diff
@@ -0,0 +1,147 @@
...
+          # @param value [String]
+          def target_vg_conversion(target, value)
+            target.separate_vg_name = value
+          end
+
+          # @param target [Agama::Storage::Volume]
+          # @param value [String]
+          def fs_type_conversion(target, value)
+            fs_type = target.outline.filesystems.find do |type|
+              type.to_human_string.downcase == value.downcase
```
jreidinger, line 108: small efficiency note: `value.downcase` can be computed outside of loop.
another option is `type.to_human_string.casecmp?(value)`

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
```diff
@@ -0,0 +1,147 @@
...
+
+            return unless fs_type
+
+            target.fs_type = fs_type
+          end
+
+          # @param target [Agama::Storage::Volume]
+          # @param value [Integer]
+          def min_size_conversion(target, value)
+            target.min_size = Y2Storage::DiskSize.new(value)
```
jreidinger, line 119: value is in bytes or kilobytes?

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
```diff
@@ -0,0 +1,147 @@
...
+          def auto_size_conversion(target, value)
+            return unless target.auto_size_supported?
+
+            target.auto_size = value
+          end
+
+          # @param target [Agama::Storage::Volume]
+          # @param value [Booelan]
+          def snapshots_conversion(target, value)
+            return unless target.outline.snapshots_configurable?
```
jreidinger, line 139: I would say that for both case where we skip conversion it should be logged what happen.

service/lib/agama/storage/encryption_settings.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      # @!attribute encryption_password
+      #   Password to use when creating new encryption devices
+      #   @return [String, nil] nil if undetermined
+      secret_attr :password
+
+      # @return [Y2Storage::EncryptionMethod::Base]
+      attr_accessor :method
+
+      # @return [Y2Storage::PbkdFunction, nil]
```
jreidinger, line 39: what means nil? use default function?

service/lib/agama/storage/lvm_settings.rb
```diff
@@ -0,0 +1,47 @@
...
+      attr_accessor :system_vg_devices
+
+      def initialize
+        @enabled = false
+        @system_vg_devices = []
+      end
+
+      # @return [Boolean]
+      def enabled?
+        !!@enabled
```
jreidinger, line 43: this looks a bit inconsistent as according to comments it should not be nil.

service/lib/agama/storage/proposal.rb
```diff
@@ -20,139 +20,103 @@
...
-      def calculate(settings = nil)
-        @settings = settings || ProposalSettings.new
-        @settings.freeze
-        y2storage_settings = to_y2storage_settings(@settings)
-
-        @proposal = new_proposal(y2storage_settings)
-        storage_manager.proposal = proposal
-
-        @on_calculate_callbacks.each(&:call)
+      # @return [ProposalSettings, nil]
```
jreidinger, line 104: I would document when it can return nil

service/lib/agama/storage/proposal.rb
```diff
@@ -258,67 +188,39 @@ def storage_manager
...
+      # Returns an issue if there is no boot device.
       #
       # @return [Issue, nil]
-      def empty_candidate_devices_issue
-        return if !proposal || candidate_devices.any?
+      def boot_device_issue
+        return if settings.boot_device
 
-        Issue.new("No devices are selected for installation",
+        Issue.new("No device selected for installation",
```
jreidinger, line 197: localization? As very soon (TM) dbus will be able to return localized string, I think that those errors will be useful.

service/lib/agama/storage/proposal_settings.rb
```diff
@@ -19,49 +19,46 @@
...
+      attr_reader :space
 
-      # Device names of the disks that can be used for the installation. If nil, the proposal will
-      # try find suitable devices
+      # Device name of the disk that will be used for booting the system and also to allocate all
+      # the partitions, except those that have been explicitly assigned to other disk(s).
       #
-      # @return [Array<String>]
-      attr_accessor :candidate_devices
+      # @return [String, nil]
```
jreidinger, line 48: what means nil? Same as before that it will auto find it?

service/lib/agama/storage/proposal_settings_conversion/from_y2storage.rb
```diff
@@ -0,0 +1,129 @@
...
+        end
+
+        # @param target [Agama::Storage::ProposalSettings]
+        def lvm_conversion(target)
+          target.lvm.enabled = settings.lvm
+
+          # Only assign system VG devices if candidate devices contains any device different to the
+          # root device.
+          candidate_devices = settings.candidate_devices || []
+          return unless candidate_devices.reject { |d| d == settings.root_device }.any?
```
jreidinger, line 70: I think this case also deserve own log line.

service/lib/agama/storage/proposal_settings_conversion/from_y2storage.rb
```diff
@@ -0,0 +1,129 @@
...
+        # @param target [Agama::Storage::ProposalSettings]
+        def encryption_conversion(target)
+          target.encryption.password = settings.encryption_password
+          target.encryption.method = settings.encryption_method
+          target.encryption.pbkd_function = settings.encryption_pbkdf
+        end
+
+        # @param target [Agama::Storage::ProposalSettings]
+        def space_settings_conversion(target)
+          # FIXME: No way to infer space settings from Y2Storage.
```
jreidinger, line 84: at least logging?

service/lib/agama/storage/proposal_settings_conversion/to_y2storage.rb
```diff
@@ -0,0 +1,189 @@
...
+        # @param target [Y2Storage::ProposalSettings]
+        def boot_device_conversion(target)
+          target.root_device = settings.boot_device
+        end
+
+        # @param target [Y2Storage::ProposalSettings]
+        def candidate_devices_conversion(target)
+          candidate_devices = []
+          candidate_devices = settings.lvm.system_vg_devices if settings.lvm.enabled?
+          candidate_devices = [settings.boot_device] if candidate_devices.none?
```
jreidinger, line 72: well, more common is to use `empty?` instead of `none?`.

jreidinger: ah, it is used here to filter if there are only nils in array....definitively tricky usage. Maybe more readable will be using compact already in a line above or here in test and use common `empty?` as I think it can confuse more people.

service/lib/agama/storage/proposal_settings_conversion/to_y2storage.rb
```diff
@@ -0,0 +1,189 @@
...
+
+        # @param target [Y2Storage::ProposalSettings]
+        def fallbacks_conversion(target)
+          target.volumes.each do |spec|
+            spec.fallback_for_min_size = find_min_size_fallback(spec.mount_point)
+            spec.fallback_for_max_size = find_max_size_fallback(spec.mount_point)
+          end
+        end
+
+        # @param mount_path [String]
```
jreidinger, line 139: return here can be String or nil.

service/lib/agama/storage/proposal_settings_conversion/to_y2storage.rb
```diff
@@ -0,0 +1,189 @@
...
+            settings.volumes.map(&:device)
+
+          devices.compact.uniq.map { |d| device_or_partitions(d) }.flatten
+        end
+
+        # @param device [String]
+        # @return [String, Array<String>]
+        def device_or_partitions(device)
+          partitions = partitions(device)
+          partitions.any? ? partitions : device
```
jreidinger, line 170: can partitions return array with some nils? it does not look for me. So I would maybe use more readable ` partitions.empty? ? device : partitions` to avoid confusion

service/lib/agama/storage/proposal_settings_reader.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+
+      # Reads the proposal settings from the control file.
+      #
+      # @return [ProposalSettings]
+      def read
+        ProposalSettings.new.tap do |settings|
+          config.data.fetch("storage", {}).each do |key, value|
+            reader = READERS[key]
+            send(reader, settings, value) if reader
```
jreidinger, line 43: what mean if there is no reader? Does it make sense to log that key does not have reader?

service/lib/agama/storage/space_settings.rb
```diff
@@ -0,0 +1,48 @@
...
+  module Storage
+    # Settings regarding the Agama storage proposal for making space
+    class SpaceSettings
+      POLICIES = [:delete, :keep, :resize, :custom].freeze
+      private_constant :POLICIES
+
+      # @return [Symbol] :delete, :keep, :resize or :custom
+      attr_accessor :policy
+
+      # @return [Hash] used only for :custom
```
jreidinger, line 32: btw I think there is even design pattern that suggest to use object and inheritance for different kind of policy. But I think here it is overkill. But if you plan to extend it in future, then it is something to consider.

service/lib/agama/storage/volume_conversion/from_y2storage.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        # @return [Y2Storage::VolumeSpecification]
+        attr_reader :spec
+
+        # @return [Agama::Config]
+        attr_reader :config
+
+        # @param target [Agama::Storage::Volume]
+        def sizes_conversion(target)
+          target.auto_size = !(spec.ignore_fallback_sizes? && spec.ignore_snapshots_sizes?)
```
jreidinger, line 64: to be honest I found this expression harder to read than more explicit `!spec.ignore_fallback_sizes? || !spec.ignore_snapshots_sizes?`

service/lib/agama/storage/volume_conversion/from_y2storage.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          planned = planned_device_for(spec.mount_point)
+          target.min_size = planned&.min || spec.min_size
+          target.max_size = planned&.max || spec.max_size
```
jreidinger, line 68: to be honest I do not get what this code means. Why planned device min size has precedence over min size from specification? I understand code, but do not understand reason. It would be nice to have rationale explained in comment.

service/lib/agama/storage/volume_conversion/to_y2storage.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        # @param target [Y2Storage::VolumeSpecification]
+        def sizes_conversion(target)
+          auto = volume.auto_size?
+
+          target.ignore_fallback_sizes = !auto
+          target.ignore_snapshots_sizes = !auto
+
+          target.min_size = auto ? volume.outline.base_min_size : volume.min_size
+          target.max_size = auto ? volume.outline.base_max_size : volume.max_size
```
jreidinger, line 69: and here I am also a bit confused as conversion from y2storage and to y2storage looks like have different logic. So I would like to know also here rationale, why for auto it sets outline and otherwise directly on volume.

service/lib/agama/storage/volume_conversion/to_y2storage.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        # @param target [Y2Storage::VolumeSpecification]
+        def btrfs_conversion(target)
+          target.snapshots = volume.btrfs.snapshots?
+          target.snapshots_configurable = volume.outline.snapshots_configurable?
+          target.snapshots_size = volume.outline.snapshots_size || Y2Storage::DiskSize.zero
+          target.snapshots_percentage = volume.outline.snapshots_percentage || 0
```
jreidinger, line 77: I am not sure if defaults is proper to have here. Does y2storage have elsewhere these defaults? or maybe volume can have at least constant with default values if not defined?

service/lib/agama/storage/volume_outline.rb
```diff
@@ -0,0 +1,118 @@
...
+      attr_accessor :filesystems
+
+      # Base value to calculate the min size for the volume (if #auto_size is set to true
+      # for that final volume) or to use as default value (if #auto_size is false)
+      #
+      # @return [Y2Storage::DiskSize]
+      attr_accessor :base_min_size
+
+      # Base value to calculate the max size for the volume (if #auto_size is set to true
+      # for that final volume) or to use as default value (if #auto_size is false)
```
jreidinger, line 50: I am a bit lost here with comments. Both base_min_size and base_max_size mentions that it will be used as default if auto_size is empty. So what happens when it has different value?

service/lib/agama/storage/volume_outline.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # Whether snapshots option can be configured
+      #
+      # @return [Boolean]
+      attr_accessor :snapshots_configurable
+      alias_method :snapshots_configurable?, :snapshots_configurable
+
+      # Size required for snapshots
+      #
+      # @return [Y2Storage::DiskSize, nil]
```
jreidinger, line 73: I miss here document what nil means. Does it mean that snapshots does not need additional size?

service/lib/agama/storage/volume_templates_builder.rb
```diff
@@ -0,0 +1,198 @@
...
+      end
+
+      def cleanpath(path)
+        Pathname.new(path).cleanpath.to_s
+      end
+
+      def fs_type(filesystem)
+        return filesystem if filesystem.is_a?(Y2Storage::Filesystems::Type)
+
+        Y2Storage::Filesystems::Type.find(filesystem.downcase.to_sym)
```
jreidinger, line 194: to be honest I expect such method in that Type class instead of exposing that it is only lowercase here.

service/test/agama/rspec/matchers/storage.rb
```diff
@@ -0,0 +1,48 @@
...
+      module Storage
+        ::RSpec::Matchers.define(:eq_outline) do |expected|
+          match do |received|
+            methods = [
+              :required?, :filesystems, :base_min_size, :base_max_size, :adjust_by_ram?,
+              :min_size_fallback_for, :max_size_fallback_for, :snapshots_configurable?,
+              :snapshots_size, :snapshots_percentage
+            ]
+
+            methods.all? { |m| received.public_send(m) == expected.public_send(m) }
```
jreidinger, line 36: to be honest, why not simply define method `eq` on given objects? so you can easily use `expect(a).to eq b`?

service/test/agama/storage/volume_templates_builder_test.rb
```diff
@@ -0,0 +1,277 @@
...
+          expect(subvols.size).to eq 5
+          expect(subvols).to all be_a(Y2Storage::SubvolSpecification)
+          expect(subvols.map(&:path)).to contain_exactly("home", "root", "srv", "boot", "var")
+          expect(subvols.map(&:archs)).to contain_exactly(nil, nil, nil, nil, "x86_64")
+          expect(subvols.map(&:copy_on_write)).to contain_exactly(true, true, true, true, false)
+        end
+      end
+    end
+  end
+end
```
jreidinger, line 277: :+1: for extensive test of that big class

service/etc/agama.yaml
joseivanlopez: These packages where orininally added here https://github.com/openSUSE/agama/pull/485 for ALP-Bedrock, and the architecture filter was added here https://github.com/openSUSE/agama/pull/486. Maybe @teclator and @imobachgs remember some details.

service/etc/agama.yaml
joseivanlopez: I simply reorganized the file a bit to define the products in the same order they are listed above. Sorry for the confusion.

service/etc/agama.yaml
joseivanlopez: For now, let's remove this misleading comment. All these about FDE, etc should be revisit, but it is out of scope of this PR.

service/lib/agama/storage/volume_outline.rb
joseivanlopez: Added explanation.

service/lib/agama/dbus/storage/manager.rb
joseivanlopez: To be honest, we have not followed any specific criteria yet. I take note, because it is something to consider in general.

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
joseivanlopez: We are going to create a follow-up issue to validate data from both D-Bus and config file.

service/lib/agama/dbus/storage/proposal_settings_conversion/from_dbus.rb
```diff
@@ -0,0 +1,165 @@
...
+            return unless function
+
+            target.encryption.pbkd_function = function
+          end
+
+          # @param target [Agama::Storage::ProposalSettings]
+          # @param value [String]
+          def space_policy_conversion(target, value)
+            policy = value.to_sym
+            return unless Agama::Storage::SpaceSettings.policies.include?(policy)
```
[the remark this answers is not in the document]
joseivanlopez: Yes, I agree, but there are many more things to check, like the existence of the given devices, etc. For now this is not worsing the previous status (validations have been always missing), so I would prefer to address the validations as a separate issue as soon as possible.

service/lib/agama/dbus/storage/proposal_settings_conversion/to_dbus.rb
joseivanlopez: They cannot be nil.

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
joseivanlopez: In bytes, let'd document it.

service/lib/agama/storage/encryption_settings.rb
joseivanlopez: Let's document it.

service/lib/agama/storage/lvm_settings.rb
joseivanlopez: Right!

service/lib/agama/storage/proposal.rb
joseivanlopez: Yes. I would prefer to adapt all text as part of a separate issue once the localization is ready.

service/lib/agama/storage/proposal_settings_conversion/from_y2storage.rb
joseivanlopez: Not sure. In this case logging does not give that much info here. There are a couple of FIXMEs in the code to address this in the future (if needed).

service/lib/agama/storage/proposal_settings_conversion/from_y2storage.rb
joseivanlopez: I am not sure if we need to log this kind of decisions. I would prefer to log the intial settings and the converted ones. And of course, the code should explain the details.

service/lib/agama/storage/proposal_settings_conversion/to_y2storage.rb
joseivanlopez: Well, I am very used to `#any?` and `#none?`. For me it is even easier to read than `#empty?`. But let's use `#empty?` if you think it is more common.

service/lib/agama/storage/proposal_settings_reader.rb
joseivanlopez: No reader means there is an unknown key in the config file. Again, all these cases fall down into the missing validation part that will be added later.

service/lib/agama/storage/space_settings.rb
joseivanlopez: Yes. In fact, @ancorgs and me have already discussed about using some objects for the size config too (`AutoSize` vs `ManualSize`). This here is a similar case. For now we decided to keep this simple approach and improve it later if needed.

service/lib/agama/storage/volume_conversion/from_y2storage.rb
joseivanlopez: Ok, added.

service/lib/agama/storage/volume_conversion/to_y2storage.rb
joseivanlopez: Ok.

service/lib/agama/storage/volume_conversion/to_y2storage.rb
joseivanlopez: Well, we don't want to rely on the Y2Storage defaults. The defaults for Agama could differ. And I am not sure about defining those values for Volume, because from Agama logic is perfectly fine to have undefined `#snapshots_size` and `snapshots_percentage`. It is only in the time of talking to Y2Storage when we need to provide some value if nothing was defined.

service/lib/agama/storage/volume_outline.rb
joseivanlopez: I would say the documentation is not true for the false case. If auto is false, base sizes are not used at all.

service/lib/agama/storage/volume_outline.rb
joseivanlopez: Yes, if not defined, no extra size will be added for snapshots.

service/lib/agama/storage/volume_templates_builder.rb
joseivanlopez: Again, I would say this is part of the validation.

service/test/agama/rspec/matchers/storage.rb
joseivanlopez: I thought about it, but finally I decided to avoid defining `#eq` because is not needed for the code, only for testing. So, I preferred to "manually" compare the values I was interested on.

service/lib/agama/storage/volume_outline.rb
joseivanlopez: Doc updated.

service/lib/agama/dbus/storage/manager.rb
jreidinger: well, I usually use alphabetic one as I can use simple stuff like in vim just select lines and use `:sort` and it is done.

service/lib/agama/storage/proposal_settings_conversion/to_y2storage.rb
jreidinger: well, I think that code you write and code above looks quite different. So let me write your code from comment in consistent way:
```
candidate_devices = if settings.lvm.enabled? && settings.lvm.system_vg_devices.any?
  settings.lvm.system_vg_devices
else
  [settings.boot_device]
end.compact
```

this way it will be at least visible that candidates are either system_vg_devices or boot_device and not some strange combination of both.

service/lib/agama/storage/volume_templates_builder.rb
jreidinger: Well, I am not sure if it is about validation that we allow to have also upper case fs names, but as change will need change in yast it is ok for me to have it as it is.

doc/dbus/org.opensuse.Agama.Storage1.Proposal.Calculator.doc.xml
```diff
@@ -21,8 +21,8 @@
         TargetDevice            s
         TargetVG                s
         FsType                  s
-        MinSize                 t
-        MaxSize                 t (optinal, max size is considered as unlimited if omitted)
+        MinSize                 t (bytes)
+        MaxSize                 t (bytes. Optinal, max size is considered as unlimited if omitted)
```
jreidinger, line 25: typo `Optinal`

doc/dbus/org.opensuse.Agama.Storage1.Proposal.doc.xml
```diff
@@ -19,8 +19,8 @@
         TargetDevice            s
         TargetVG                s
         FsType                  s
-        MinSize                 t
-        MaxSize                 t (optinal, max size is considered as unlimited if omitted)
+        MinSize                 t (bytes)
+        MaxSize                 t (bytes. Optinal, max size is considered as unlimited if omitted)
```
jreidinger, line 23: and same typo here.

service/lib/agama/storage/volume_conversion/to_y2storage.rb
```diff
@@ -65,6 +65,11 @@ def sizes_conversion(target)
           target.ignore_fallback_sizes = !auto
           target.ignore_snapshots_sizes = !auto
 
+          # The range of sizes is defined by the volume outline in case of auto size (mix and max
+          # sizes cannot be configured if auto size is set).
+          # And note that the inal range of sizes used by the Y2Storage proposal is calculated by
```
jreidinger, line 70: typo `inal`

service/lib/agama/storage/volume_outline.rb
```diff
@@ -56,10 +56,22 @@ class VolumeOutline
...
-      # @return [Array<String>] mount paths of other volumes
+      # Lists the mount paths of the volumes for which this volume is a min size fallback.
+      #
+      # Being a min size fallback means that the min size of the volume would be increased by the
+      # min size of other volumes, if any of that other volumes is not used for the proposal.
+      #
+      # For example, let's say the root volume is a fallback for the min size of /home and /var
+      # (root.min_size_fallback_for => ["/home", "/var"]). And a proposal is calculated with only
+      # root and /home. In that case, the min size of /var is added to the min size of root. The
+      # same would happen for /home if the proposal does not include it.
```
jreidinger, line 67: Thanks for great explanation. Now it makes much more sense.

jreidinger (approved):
nothing critical, just typos and one possible code improvement

service/lib/agama/dbus/storage/manager.rb
joseivanlopez: Let's sort them.

PR #659, opened 2023-07-12
CLI: improve error reporting

imobachgs:
## Problem

CLI error reporting is rather poor. In some cases, the errors are pretty hard to read (e.g., when it cannot connect to the D-Bus server) and, in other cases, the problem is not properly reported (e.g., when lshw or jsonnet are missing or when the jsonnet invocation fails).

The anyhow crate helps with reporting the problems, including a backtrace if supported. We should make use of its reporting capabilities.

Errors to focus on:

* ServiceError, which is anyone coming from the D-Bus layer.
* Missing binaries (`lshw`, jsonnet).
* Cannot download a profile (from libcurl).
* Failing commands (e.g., running jsonnet).

### Better CLI errors

`agama-cli` uses now [anyhow](https://crates.io/crates/anyhow) to report the errors. Let's see some examples:

```
# agama profile download http://missing-domain.net/profile.jsonnet
Could not read the profile

Caused by:
    [6] Couldn't resolve host name (Could not resolve host: missing-domain.net)

# agama profile evaluate bad.jsonnet
Could not evaluate the profile

Caused by:
    Jsonnet evaluation failed:
    profile.jsonnet:39:15-30 Unknown variable: findBiggestDisk
    
            name: findBiggestDisk(agama.disks),

# agama config show
Could not connect to Agama bus at 'unix:path=/run/agama/bus'

Caused by:
    0: I/O error: No such file or directory (os error 2)
    1: No such file or directory (os error 2)
```

### Exit code

`agama-cli` now returns an exit code which is especially important when using the CLI in scripts. As usual, it returns 0 for success and 1 for errors.

```
Unknown product 'Unknown'. Available products: '["ALP-Bedrock", "ALP-Micro", "Tumbleweed", "Leap16"]'
$ echo $?
1
```

### Stop using &str as error types

In Rust, anything can be used as an error. Nothing stops you from using, e.g., a string slice (`&str`) as your error type. We used this approach when handling settings from the command line. This PR introduces a new `SettingsError` enum instead.

### Partially get rid of `Box<dyn Error>`

We have replaced `Box<dyn Error>` with concrete error types.

## Testing

- *Tested manually*

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -13,6 +13,7 @@ name = "agama-cli"
 version = "1.0.0"
 dependencies = [
  "agama-lib",
+ "anyhow",
  "async-std",
  "clap",
  "console",
--- a/rust/agama-cli/Cargo.toml
+++ b/rust/agama-cli/Cargo.toml
@@ -16,6 +16,7 @@ async-std = { version ="1.12.0", features = ["attributes"] }
 thiserror = "1.0.39"
 convert_case = "0.6.0"
 console = "0.15.7"
+anyhow = "1.0.71"
 
 [[bin]]
 name = "agama"
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -31,7 +31,7 @@ pub enum ConfigAction {
     Load(String),
 }
 
-pub async fn run(subcommand: ConfigCommands, format: Format) -> Result<(), Box<dyn Error>> {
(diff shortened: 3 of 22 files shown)


mvidner:
Thanks for pointing out Luca Palmieri's [Error Handling In Rust - A Deep Dive](https://www.lpalmieri.com/posts/error-handling-rust/) as a detailed background piece

mvidner (approved):
Nice, thanks!

PR #660, opened 2023-07-13
Improve the reporting of D-Bus connection error

imobachgs:
Improve the error reporting when a D-Bus connection cannot be established. By the way, it updates the changes file (I forgot to do it on #659).

Before:

```
Error: Connecting to Agama bus at unix:path=/run/agama/bus

Caused by:
    0: Connecting to Agama bus at unix:path=/run/agama/bus
    1: I/O error: Broken pipe (os error 32)
    2: Broken pipe (os error 32)
```

Now:

```
Could not connect to Agama bus at 'unix:path=/run/agama/bus'

Caused by:
    0: I/O error: Broken pipe (os error 32)
    1: Broken pipe (os error 32)
```

### DIFF
--- a/rust/agama-lib/src/error.rs
+++ b/rust/agama-lib/src/error.rs
@@ -6,8 +6,10 @@ use zbus;
 
 #[derive(Error, Debug)]
 pub enum ServiceError {
-    #[error("D-Bus service error: {0}")]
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #663, opened 2023-07-13
[RFC] Avoid calling await while holding a lock

imobachgs:
Clippy reports that in our code we call `await` while holding a lock. There are two possible solutions:

* Make sure than the Mutex is drop before calling `await` (this).
* Use an async-aware Mutex implementation (#664).

This PR implements the first solution. I have decided to use an scope instead of calling `drop` because of https://github.com/rust-lang/rust-clippy/issues/6446.


imobachgs:
Finally went with #664.

PR #665, opened 2023-07-14
Fix Jsonnet example to select the biggest disk

imobachgs:
It wrongly selected the smallest disk instead of the biggest one 😅.


mvidner (approved):
oh. right :sweat_smile: .

PR #666, opened 2023-07-17
Better Settings derive macro

imobachgs:
## Problems

* Although our `Settings` derive macro generates some code, it cannot generate code for `InstallSettings` or `NetworkSettings`.
* Writing tests for the macro is hard because it cannot be reused from external crates. The generated code refers to `crate::settings::Settings`, which does not work from another package; changing it to `agama_lib::settings::Settings` would make it fail for agama-lib.

### Making the macro reusable

To make the derive macro reusable and easy to test, I moved it to a separate package, `agama-settings`. Now we can implement [an integration test](https://github.com/imobachgs/agama/blob/better-agama-derive/rust/agama-settings/tests/settings.rs).

In the future, we could refactor the code to make it easier to test the internals (checking the generated code, error handling, and so on). I got some ideas to improve the implementation from [Structuring, testing and debugging procedural macro crates](https://ferrous-systems.com/blog/testing-proc-macros/).

### Extending the macro

This PR extends the macro to handle "scalar" (e.g., a string, a boolean, a number), "nested" (e.g., InstallSettings#network) and "collection" attributes (e.g., NetworkSettings#connections).

By now, I have decided to rely on some attributes instead of inferring which kind of settings are (scalar, nested or collections). The reason is that you do not have access to the type system, just to the syntax tree. You can assume that `Vec` is a collection, but what if you are parsing `std::vec::Vec` instead? OK, I know it is not the best example, but I guess you get what I meant.

If we decide to try to infer the values, it is as easy as changing the [parse_setting_fields](https://github.com/openSUSE/agama/blob/19b6782064483b2edbe277861076d00c46f3ad21/rust/agama-derive/src/lib.rs#L115C2-L115C2) function with something else.

### Improving the error reporting

The `SettingsError` struct has been changed to make it easier to report errors like the example below:

```
# agama config set storage.lvm=fasle
Could not update 'storage.lvm': Invalid value 'fasle', expected a boolean
```

## Testing

If you are curious enough, you can generate the code produced by the macro:

1. Install [cargo expand](https://github.com/dtolnay/cargo-expand)
2. Run `cargo expand -p agama-lib > expanded.rs`

## Hints for reviewers

Look for the _deleted_ code implementing `impl Settings for FooSettings` because that shows you concrete examples of what the macro should expand to.


rust/agama-settings/src/error.rs
```diff
@@ -0,0 +1,13 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum SettingsError {
+    #[error("Unknown attribute '{0}'")]
+    UnknownAttribute(String),
+    #[error("Unknown collection '{0}'")]
+    UnknownCollection(String),
+    #[error("Invalid value '{0}'")]
```
mvidner, line 9: Invalid value 'flase' for setting storage.lossy_compression, expecting a Float

rust/agama-settings/src/error.rs
```diff
@@ -0,0 +1,13 @@
...
+    InvalidValue(String),
+    #[error("Missing key '{0}'")]
```
mvidner, line 11: Missing key 'id' for software.pattern

mvidner (commented):
Idea to improve the messages for `SettingsError`, if the context doesn't already come from elsewhere

rust/agama-derive/src/lib.rs
```diff
@@ -46,51 +70,155 @@ pub fn agama_attributes_derive(input: TokenStream) -> TokenStream {
...
+    quote! {
+        fn add(&mut self, attr: &str, value: agama_settings::SettingObject) -> Result<(), agama_settings::SettingsError> {
+            #nested_handling
+            #collection_handling
             Ok(())
         }
     }
 }
+
+// Extracts information about the settings fields
```
mvidner, line 195:
```suggestion
// Extracts information about the settings fields.
//
// syn::Field is "A field of a struct or enum variant.", 
// has .ident .ty(pe) .mutability .vis(ibility)...
```

rust/agama-lib/src/users/settings.rs
```diff
@@ -1,49 +1,19 @@
...
 
 /// User settings
 ///
 /// Holds the user settings for the installation.
-#[derive(Debug, Default, Serialize, Deserialize)]
+#[derive(Debug, Default, Settings, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct UserSettings {
     #[serde(rename = "user")]
+    #[settings(nested)]
```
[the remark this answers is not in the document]
imobachgs: I have checked how serde documents the `Serialize` and `Deserialize` macros and they have the same problem. I will try to have a closer look, but it is something we can postpone a bit.

mvidner (approved):
When I try to test this with `agama config` CLI, it fails (mostly by failing to report my mistakes, more on that later) but it also fails the same way in `master`, so probably not fault of this PR.

I still don't understand the metaprogramming.

On the plus side:
- the generated code compiles
- tests pass, there are new tests
- better error messages
- better rustdoc

imobachgs:
> When I try to test this with `agama config` CLI, it fails (mostly by failing to report my mistakes, more on that later) but it also fails the same way in `master`, so probably not fault of this PR.

OK, then it is something we need to improve. Do you have an invocation example? Thanks!

> I still don't understand the metaprogramming.

You will understand it much better when you have to write your first derive macro 😅

mvidner:
For integration tests of the settings code, I should have used `agama config load`, it avoids the problems of `config set` mentioned in #688

```console
# echo '{"user":{"fullName":"tux","userName":"Tux","password":"secret","autologin":true}}' > user.json
# agama config load user.json
I, [2023-08-03T12:11:02.347584 #16187]  INFO -- manager: Setting first user tux
# agama -f yaml config show
user:
  fullName: tux
  userName: Tux
  password: secret
  autologin: true
...
```

Well... it also hides the errors :-/

```console
# echo '{"user":{"shortname":"tux","userName":["Tux"],"password":"secret","autologin":42}}' > userbad.json
# agama config load userbad.json
```

PR #667, opened 2023-07-17
Update the changes file

imobachgs:
Fix the logic to decide which network connections to write due to a bug I introduced in gh#openSUSE/agama#662.


teclator (approved):
LGTM

PR #668, opened 2023-07-17
Cli automatic answers to questions

jreidinger:
## Problem

Add to CLI ability to use default answers for questions. As we do not have yet in place questions handling, it will be for near future only way to answer it on CLI.

## Solution

Add command `agama questions use-defaults`. It is not as described at https://github.com/openSUSE/agama/blob/master/doc/cli_guidelines.md because currently auto answering can be only set on dbus API. Also there is risk of collision as auto answering is global for all questions, not just ones caused by given CLI command, so it can confuse user. And last but not least it can lead to some confusion where to use --non-interactive as sometimes it is hard to know what settings can cause reprobe or some action that can lead to answer.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-cli/src/questions.rs
+++ b/rust/agama-cli/src/questions.rs
@@ -1,0 +1,28 @@
+use anyhow::{Ok,Context};
+use clap::Subcommand;
+use agama_lib::connection;
+use agama_lib::proxies::Questions1Proxy;
+
+#[derive(Subcommand, Debug)]
+pub enum QuestionsCommands {
+    /// To use default answers for questions.
+    UseDefaults,
+}
+
+// TODO when more commands is added, refactor and add it to agama-lib and share a bit of functionality
+async fn use_defaults() -> anyhow::Result<()> {
+    let connection = connection().await.context("Connection to DBus failed.")?;
+    let proxy = Questions1Proxy::new(&connection).await.context("Failed to connect to Questions service")?;
+
+    // TODO: how to print dbus error in that anyhow?
+    proxy.use_default_answer().await.context("Failed to set default answer")?;
+
+
+    Ok(())
+}
+
+pub async fn run(subcommand: QuestionsCommands) -> anyhow::Result<()> {
+    match subcommand {
+        QuestionsCommands::UseDefaults => use_defaults().await,
+    }
+}
--- a/rust/agama-cli/src/commands.rs
+++ b/rust/agama-cli/src/commands.rs
@@ -1,5 +1,6 @@
 use crate::config::ConfigCommands;
 use crate::profile::ProfileCommands;
+use crate::questions::QuestionsCommands;
 use clap::Subcommand;
 
 #[derive(Subcommand, Debug)]
@@ -20,4 +21,7 @@ pub enum Commands {
     /// Autoinstallation profile handling
     #[command(subcommand)]
     Profile(ProfileCommands),
+    /// Questions handling
+    #[command(subcommand)]
+    Questions(QuestionsCommands),
 }
--- a/rust/agama-cli/src/main.rs
+++ b/rust/agama-cli/src/main.rs
@@ -6,6 +6,7 @@ mod error;
 mod printers;
 mod profile;
 mod progress;
+mod questions;
 
 use crate::error::CliError;
 use agama_lib::error::ServiceError;
@@ -22,6 +23,7 @@ use std::{
     thread::sleep,
     time::Duration,
 };
+use questions::run as run_questions_cmd;
 
 #[derive(Parser)]
 #[command(name = "agama", version, about, long_about = None)]
@@ -128,6 +130,7 @@ async fn run_command(cli: Cli) -> anyhow::Result<()> {
             let manager = build_manager().await?;
             block_on(install(&manager, 3))
         }
+        Commands::Questions(subcommand) => block_on(run_questions_cmd(subcommand)),
         _ => unimplemented!(),
     }
 }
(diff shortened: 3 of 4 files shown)


rust/agama-cli/src/questions.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[derive(Subcommand, Debug)]
+pub enum QuestionsCommands {
+    /// To use default answers for questions.
+    UseDefaults,
+}
+
+// TODO when more commands is added, refactor and add it to agama-lib and share a bit of functionality
+async fn use_defaults() -> anyhow::Result<()> {
+    let connection = connection().await.context("Connection to DBus failed.")?;
```
imobachgs, line 14: I think there is no need to add the `context` because it does not add any information to the error coming from the `connection` function.

imobachgs (commented):
I am not 100% sure about the overall approach. What happens if I want to disable this behavior? Do we need a separate command? Perhaps we use a single command like `agama questions mode [interactive|defaults]` or something similar.

jreidinger:
@imobachgs issue with having `interactive` is that it does not need to be interactive. Basically idea is to have multiple strategies that you add to questions. So you e.g. specify few answers and then also non-interactive. This will lead to AnswerStrategy File + Default . BTW currently we do not have a way to disable using default answers. Is it needed?

imobachgs:
> @imobachgs issue with having `interactive` is that it does not need to be interactive. Basically idea is to have multiple strategies that you add to questions. So you e.g. specify few answers and then also non-interactive. This will lead to AnswerStrategy File + Default . BTW currently we do not have a way to disable using default answers. Is it needed?

Well, as said, perhaps "interactive" is not the right word. My point is that we might want to support different strategies, and I would use a single command to set them. But I do not have a strong opinion about that.

About the `--non-interactive` flag, we could implement that in the future by setting the strategy before running the command and rolling back to the original one after the end.

rust/agama-cli/src/questions.rs
```diff
@@ -0,0 +1,32 @@
...
+    let proxy = Questions1Proxy::new(&connection)
+        .await
+        .context("Failed to connect to Questions service")?;
+
+    // TODO: how to print dbus error in that anyhow?
+    proxy
+        .use_default_answer()
+        .await
+        .context("Failed to set default answer")?;
```
imobachgs, line 23:
```suggestion
        .context("Failed to set default answers strategy")?;
```

imobachgs (approved):
After discussing on IRC, we might need to think of a more flexible approach. But, by now, it is good enough. So let's unblock and continue from here.

joseivanlopez:
Sorry, I was a bit disconnected of this topic for some time, so I had to revisit it. If I got it correctly, we could have 3 different ways for answering questions: using default answers, using answers from a file, and asking to the user. And we don't know in adavance if a command could raise a question.

And then it is the problem with globally configuring the questions service, because the configuration affects to all questions, not only to the questions comming from a specific command.

Well, for the global mode issue, I think it would be difficult to avoid because a new question can be dispatched after setting the mode. I would not focus on this problem now, but maybe we could explore the possibility of temporary setting the questions mode as part of the question call, and then set it back to its previous mode.

And for settings the modes, I agree with a proposal similar to what @imobachgs exposed:

~~~
agama questions mode [interactive|non-interactive [path]]

agama questions mode interactive # default mode, ask questions
agama questions mode non-interactive # do not ask questions, use defaults
agama questions mode non-interactive /tmp/answers.json # do not ask questions, use answers from file or defaults
 ~~~

doc/cli_guidelines.md
```diff
@@ -192,6 +192,10 @@ There are pending questions. Please, answer questions first: agama questions.
 
 ## Non Interactive Mode
 
-Commands should offer a `--non-interactive` option to make scripting possible. The non interactive mode should allow answering questions automatically.
+There are questions subcommand that support `mode` subcommand to set non-interactive mode. Another option is to use `answers` subcommand to pass file with
```
imobachgs, line 195:
```suggestion
There is a `questions` subcommand that implements a `mode` subcommand to set the non-interactive mode. Another option is to use `answers` subcommand to pass a file with
```

rust/agama-cli/src/questions.rs
```diff
@@ -0,0 +1,49 @@
...
+    #[arg(value_enum)]
+    value: Modes,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+pub enum Modes {
+    Interactive,
+    NonInteractive,
+}
+// TODO when more commands is added, refactor and add it to agama-lib and share a bit of functionality
```
imobachgs, line 24: Beware that the code which is specific to the CLI should live in `agama-cli`. That's the main intention of this package.

imobachgs (approved):

doc/cli_guidelines.md
joseivanlopez: NP: I woud rephase it as: There is a `questions` subcommand that support `mode` and `answers` subcommands. The `mode` subcommand allows to set `interactive` and `no-interactive` modes. And `answers` allows to pass a file with prefefined answers.

rust/agama-cli/src/questions.rs
jreidinger: well, I mean here code that handles questions, do connection, use Question API and so on. But lets see how future will look like. For rust it is nice that strict separation of public and private, so it is easy to refactor internal stuff.

joseivanlopez (approved):

PR #669, opened 2023-07-19
adapt dbus API to prepare for answers file and change use_defaults to…

jreidinger:
… interactive property

## Problem

We want to have support for predefined answers and it is missing.

https://trello.com/c/9plYc08k/3373-8-agama-add-support-for-unattended-questions

## Solution

Adapt dbus API to allow to add such file.

## Testing

- *Tested manually*
- *Unit tested*

## Follow-ups

- document all types of questions we have
- some tracking/reporting feature so user can see questions that was asked/answered.

### DIFF
--- a/doc/dbus/org.opensuse.Agama.Questions1.doc.xml
+++ b/doc/dbus/org.opensuse.Agama.Questions1.doc.xml
@@ -79,15 +79,30 @@ when the question is answered and the answer is successfully read.
     <method name="Delete">
       <arg name="question" direction="in" type="o"/>
     </method>
-
     <!--
-      UseDefaultAnswer:
+      Delete:
+      @question: object path of question that should be deleted.
 
-      Switches questions to be automatically answered by default answer.
-      After this method call each follow up question is immediatelly answered.
-      Useful for doing non interactive installation.
+      Deletes question. Useful when question is answered and service that asks
+      already read answer. Usually called by the one that previously created it.
     -->
-    <method name="UseDefaultAnswer">
+    <method name="Delete">
+      <arg name="question" type="o" direction="in"/>
     </method>
+    <!--
+      AddAnswerFile:
+      @path: Local fs path to answers file in yaml format.
+
+      Adds file with list of predefined answers that will be used to
+      automatically answer matching questions.
+    -->
+    <method name="AddAnswerFile">
+      <arg name="path" type="s" direction="in"/>
+    </method>
+    <!--
+     property that defines if questions is interactive or automatically answered with
+     default answer
+     -->
+    <property name="Interactive" type="b" access="readwrite"/>
   </interface>
 </node>
--- a/doc/dbus/org.opensuse.Agama.Questions1.doc.xml
+++ b/doc/dbus/org.opensuse.Agama.Questions1.doc.xml
@@ -76,9 +76,6 @@ when the question is answered and the answer is successfully read.
       <arg name="data" direction="in" type="a{ss}"/>
       <arg direction="out" type="o"/>
     </method>
-    <method name="Delete">
-      <arg name="question" direction="in" type="o"/>
-    </method>
     <!--
       Delete:
       @question: object path of question that should be deleted.
--- a/rust/agama-cli/src/questions.rs
+++ b/rust/agama-cli/src/questions.rs
@@ -1,6 +1,6 @@
 use agama_lib::connection;
 use agama_lib::proxies::Questions1Proxy;
-use anyhow::{Context, Ok};
+use anyhow::Context;
 use clap::{Args, Subcommand, ValueEnum};
 
 #[derive(Subcommand, Debug)]
@@ -23,33 +23,31 @@ pub enum Modes {
     Interactive,
     NonInteractive,
 }
-// TODO when more commands is added, refactor and add it to agama-lib and share a bit of functionality
-async fn set_mode(value: Modes) -> anyhow::Result<()> {
-    match value {
-        Modes::NonInteractive => {
-            let connection = connection().await?;
-            let proxy = Questions1Proxy::new(&connection)
-                .await
-                .context("Failed to connect to Questions service")?;
 
-            // TODO: how to print dbus error in that anyhow?
-            proxy
-                .use_default_answer()
-                .await
-                .context("Failed to set default answer")?;
-        }
-        Modes::Interactive => log::info!("not implemented"), //TODO do it
-    }
+async fn set_mode(proxy: Questions1Proxy<'_>, value: Modes) -> anyhow::Result<()> {
+    // TODO: how to print dbus error in that anyhow?
+    proxy
+        .set_interactive(value == Modes::Interactive)
+        .await
+        .context("Failed to set default answer")
+}
 
-    Ok(())
+async fn set_answers(proxy: Questions1Proxy<'_>, path: String) -> anyhow::Result<()> {
+    // TODO: how to print dbus error in that anyhow?
+    proxy
+        .add_answer_file(path.as_str())
+        .await
+        .context("Failed to set default answer")
 }
 
 pub async fn run(subcommand: QuestionsCommands) -> anyhow::Result<()> {
+    let connection = connection().await?;
+    let proxy = Questions1Proxy::new(&connection)
+        .await
+        .context("Failed to connect to Questions service")?;
+
     match subcommand {
-        QuestionsCommands::Mode(value) => set_mode(value.value).await,
-        QuestionsCommands::Answers { path: _ } => {
-            log::info!("TODO: implement answers");
-            Ok(())
-        }
+        QuestionsCommands::Mode(value) => set_mode(proxy, value.value).await,
+        QuestionsCommands::Answers { path } => set_answers(proxy, path).await,
     }
 }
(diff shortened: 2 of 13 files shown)


rust/agama-dbus-server/src/questions.rs
```diff
@@ -227,10 +239,35 @@ impl Questions {
...
-    async fn use_default_answer(&mut self) -> Result<(), Error> {
-        log::info!("Answer questions with default option");
-        self.answer_strategies.push(Box::new(DefaultAnswers {}));
+    /// property that defines if questions is interactive or automatically answered with
+    /// default answer
+    #[dbus_interface(property)]
+    fn interactive(&self) -> bool {
+        let last = self.answer_strategies.last();
+        if let Some(real_strategy) = last {
+            real_strategy.id() == DefaultAnswers::id()
```
mvidner, line 248: What? Is this logic reversed?

As I understand it:
- *interactive* means that the UI pops up the question for the user to click on
- *non-interactive* means either
  - using the default answers for everything, or
  - reading the answers from a file (TODO)

Also an interesting point is whether the strategy is global (it is now, right?) or per question class (too complicated, not worth it)

rust/agama-dbus-server/src/questions.rs
```diff
@@ -227,10 +239,35 @@ impl Questions {
...
+            false
+        }
+    }
+
+    #[dbus_interface(property)]
+    fn set_interactive(&mut self, value: bool) {
+        log::info!("set interactive to {}", value);
+        if value {
+            if !self.interactive() {
+                self.answer_strategies.pop();
```
mvidner, line 259: This means that no strategies mean "no non-interactive" strategies, that is, the default strategy is interaction? Should be explicitly stated at Questions::answer_strategies declaration

rust/agama-dbus-server/src/questions.rs
```diff
@@ -227,10 +239,35 @@ impl Questions {
...
-        self.answer_strategies.push(Box::new(DefaultAnswers {}));
+    /// property that defines if questions is interactive or automatically answered with
+    /// default answer
+    #[dbus_interface(property)]
+    fn interactive(&self) -> bool {
+        let last = self.answer_strategies.last();
+        if let Some(real_strategy) = last {
+            real_strategy.id() == DefaultAnswers::id()
+        } else {
+            false
```
mvidner, line 250: As I understand it, this should be `true`

rust/agama-dbus-server/src/questions.rs
jreidinger: you are right. It should be `!=` ...and yes, it is global...if you want per question class you can achieve it by using answers file where you specify class and answer you want.

rust/agama-dbus-server/src/questions.rs
jreidinger: yes, no strategies means that user need to answer it itself. Will document it.

rust/agama-dbus-server/src/questions.rs
jreidinger: yes, right. My fault.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -227,10 +242,35 @@ impl Questions {
...
+            real_strategy.id() != DefaultAnswers::id()
+        } else {
+            true
+        }
+    }
+
+    #[dbus_interface(property)]
+    fn set_interactive(&mut self, value: bool) {
+        log::info!("set interactive to {}", value);
+        if value {
```
mvidner, line 260: Implementation nitpicking: I would find it easier to understand like this, and the pattern is better generalized to more-valued properties:

```rs
if value == self.interactive() {
  log::info!("Interactive property unchanged, {}", value);
  return;
}
log::info!("set interactive to {}", value);

if value {
  ...pop...
} else {
  ...push...
}
```

rust/agama-dbus-server/src/questions/answers.rs
```diff
@@ -0,0 +1,291 @@
...
+                ("data2".to_string(), "value2".to_string()),
+                ("data3".to_string(), "value3".to_string()),
+            ]),
+            answer: "".to_string(),
+        };
+        assert_eq!(Some("Maybe".to_string()), answers.answer(&question));
+    }
+
+    #[test]
+    fn test_full_data_match() {
```
mvidner, line 215: I think this test does the same thing as `partial_data_match`. What is the point?

jreidinger: well, difference here is that first test match only one answer. But here data2 match one answer and data1 another. So it test if wins correct answer.

mvidner: I see, maybe I can express it slightly better in code or comments, will try

rust/agama-cli/src/questions.rs
```diff
@@ -20,29 +23,31 @@ pub enum Modes {
...
+        .context("Failed to set default answer")
+}
 
-    Ok(())
+async fn set_answers(proxy: Questions1Proxy<'_>, path: String) -> anyhow::Result<()> {
+    // TODO: how to print dbus error in that anyhow?
+    proxy
+        .add_answer_file(path.as_str())
+        .await
+        .context("Failed to set default answer")
```
joseivanlopez, line 40: NP: The same here, maybe "Failed to set default answers from answers file".

rust/agama-dbus-server/src/questions.rs
```diff
@@ -103,7 +110,17 @@ trait AnswerStrategy {
 /// AnswerStrategy that provides as answer the default option.
 struct DefaultAnswers;
 
+impl DefaultAnswers {
+    pub fn id() -> u8 {
+        1
```
joseivanlopez, line 115: NP: I think this can be a field of the struct.

rust/agama-dbus-server/src/questions.rs
```diff
@@ -227,11 +244,43 @@ impl Questions {
...
-        log::info!("Answer questions with default option");
-        self.answer_strategies.push(Box::new(DefaultAnswers {}));
-        Ok(())
+    /// property that defines if questions is interactive or automatically answered with
+    /// default answer
+    #[dbus_interface(property)]
+    fn interactive(&self) -> bool {
+        let last = self.answer_strategies.last();
+        if let Some(real_strategy) = last {
+            real_strategy.id() != DefaultAnswers::id()
```
joseivanlopez, line 253: I am not sure if I understood the idea of having more than one strategy. Is it for priorities? First default, if no found file, if not found ask to user? And the mode is only considered interactive if the last assigned strategy is not DefaultAnswers. But what happens if the last strategy is answers?

rust/agama-dbus-server/src/questions.rs
```diff
@@ -81,7 +83,12 @@ enum QuestionType {
 }
 
 /// Trait for objects that can provide answers to all kind of Question.
+///
+/// If not strategy is used or all defined strategies does not know answer,
```
joseivanlopez, line 87: NP: a bit of rewording: "If no strategy is selected or the answer is unknown, then ask to the user".

rust/agama-dbus-server/src/questions/answers.rs
```diff
@@ -0,0 +1,291 @@
...
+/// The *answer* part is: `answer`, `password`.
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+struct Answer {
+    pub class: Option<String>,
+    pub text: Option<String>,
+    /// A matching GenericQuestion can have other data fields too
+    pub data: Option<HashMap<String, String>>,
+    /// The answer text is the only mandatory part of an Answer
+    pub answer: String,
+    /// All possible mixins have to be here, so they can be specified in an Answer
```
joseivanlopez, line 18: Mixins?

doc/dbus/org.opensuse.Agama.Questions1.doc.xml
```diff
@@ -76,18 +76,30 @@ when the question is answered and the answer is successfully read.
...
+    -->
     <method name="Delete">
-      <arg name="question" direction="in" type="o"/>
+      <arg name="question" type="o" direction="in"/>
     </method>
-
     <!--
-      UseDefaultAnswer:
+      AddAnswerFile:
+      @path: Local fs path to answers file in yaml format.
```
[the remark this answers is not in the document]
jreidinger: luckily it is easy with abstraction library provides. It is just about different crate. So when it is unify it is fine for me to convert it.

rust/agama-dbus-server/src/questions/answers.rs
jreidinger: Mixins to Base Question like `WithPassword`. So it means that all possible additional data entries from additional mixins to question have to be specified in answer.

rust/agama-dbus-server/src/questions.rs
jreidinger: reason why it is also class method is to allow detection of strategy like below that instance is DefaultAnswers

rust/agama-dbus-server/src/questions.rs
jreidinger: Reason for strategies is extensibility, which allows to add completely different source of answers from strategy like e.g. some remote server answering it.
So whole idea is to have list of strategies and the first one that can answer it, provide that answer.
Default strategy should be always able to answer, so it is reason why it is last as answer file has precedence. And also it means that it is non interactive.
If last strategy is answers and it does not provide answer to given question, it is asked user. It is general rule, if no strategy can answer, then ask user ( aka interactive mode ).

joseivanlopez (approved):

PR #670, opened 2023-07-19
Added weblate-update-pot.yml GitHub Action

lslezak:
## Problem

- Automatically extract and update the translations file (POT file)
- https://trello.com/c/RWh29AWA/3396-5-agama-set-up-the-translation-infrastructure-web-ui

## Solution

- Use a solution similar to the [Cockpit solution](https://github.com/cockpit-project/cockpit/blob/3d2d07cb751b141b6bd6ee9a3d4230815eb6dad9/.github/workflows/weblate-sync-pot.yml)
- Run a GitHub Action regularly which generates the POT file and pushes it to the `agama-weblate` repository
- To avoid unneeded commits the file is not updated when only the timestamp is changed and the rest of the file is the same

## Notes

- The `xgettext` call is just a temporary solution, it should be later replaced by some other call ([Trello card](https://trello.com/c/QPyKM2VR))
- Similarly the changes in the `FileViewer.jsx` file, these are there just to have some texts in the generated POT file and to have something to import to the Weblate

## Testing

- Tested manually
- See [this test run](https://github.com/lslezak/agama/actions/runs/5599328351/jobs/10240144236) and a [repeated test run](https://github.com/lslezak/agama/actions/runs/5599586720/jobs/10240745851) (skipping the commit step because of no changes)
- See the generated [agama.pot](https://github.com/openSUSE/agama-weblate/blob/main/web/agama.pot) file
- At runtime the original English texts are displayed as the translations are currently not supported in the Web UI

### DIFF
--- a/.github/workflows/weblate-update-pot.yml
+++ b/.github/workflows/weblate-update-pot.yml
@@ -1,0 +1,88 @@
+name: Weblate Update POT
+
+on:
+  schedule:
+    # run every working day (Monday-Friday) at 1:42AM UTC
+    - cron: "42 1 * * 0-4"
+
+  # allow running manually
+  workflow_dispatch:
+
+jobs:
+  update-pot:
+    # do not run in forks
+    if: github.repository == 'openSUSE/agama'
+
+    runs-on: ubuntu-latest
+
+    container:
+      image: registry.opensuse.org/opensuse/tumbleweed:latest
+
+    steps:
+      - name: Configure and refresh repositories
+        # disable unused repositories to have a faster refresh
+        run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && zypper ref
+
+      - name: Install tools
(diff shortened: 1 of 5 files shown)


.github/workflows/weblate-update-pot.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Checkout Agama sources
+        uses: actions/checkout@v3
+        with:
+          path: agama
+
+      - name: Generate POT file
+        # TODO: use a shared script for this
+        run: |
+          cd agama/web
```
jreidinger, line 37: so if I get it, it generate pot file only from web part and not from service ( aka ruby code ) and from rust code?

jreidinger (approved):

.github/workflows/weblate-update-pot.yml
lslezak: Yes, web frontend only. As noted in the TODO above, this should be replaced by some other script.

PR #671, opened 2023-07-19
Auto close Dropdown when user clicks outside

### LINKED ISSUE #552: Automatically close a Dropdown when user clicks outside
After #545 be merged in master branch, the page options will be shown by using a PatternFly [`Dropdown`](https://www.patternfly.org/v4/components/dropdown) placed close to the hamburger menu icon to open/close the sidebar.

The user can click on the _arrow down_ icon to check open the _menu_ holding the options related to the current _page_. But this dropdown will be closed only when 

* the user closes it by clicking in the arrow again
* the use clicks on any of the available options

Ideally, it should be closed automatically when clicking outside the dropdown itself too (right after losing the focus). At this time there is no prop (or I overlooked it) to tell the component to do so. But, looking at the source code it looks to me that current `PatternFly Dropdown` is/will be deprecated in favor of another that has a `onOpenChange` prop that _might_ help to achieve the mentioned behavior. Needs to check/test it better.

https://github.com/patternfly/patternfly-react/commit/d43cf2427dc9ea1ddb2381284f05276aa2c7cdb7

In any case, we should evaluate a way to automatically close that menu, although it is more a _nice to have_ than a requirement for me.

balsa-asanovic:
## Problem

As stated in #552 the Dropdown options menu of the page can currently be only closed by clicking the arrow again or by clicking any of the available options.

It would be nice for it to also close when the user clicks anywhere outside of the Dropdown.
This pull request introduces this functionality.

fixes #552 

## Solution

In the [PageOptions](https://github.com/openSUSE/agama/blob/master/web/src/components/core/PageOptions.jsx) component, I've added the useEffect hook which attaches event listener to the document on the component mount.
This event listener gets triggered on any click event through the whole document and if it is outside of the Dropdown it closes the options menu.

## Testing

Added unit test which checks if the drop down is closed on outside click.

## Screenshots

Dropdown opened
image

Dropdown closed on outside click
image


web/src/components/core/PageOptions.jsx
```diff
@@ -114,20 +114,40 @@ const Item = ({ children, ...props }) => {
...
-      <Dropdown
-        isOpen={isOpen}
-        toggle={<Toggler onClick={onToggle} />}
-        onSelect={onSelect}
-        dropdownItems={Array(children)}
-        position="right"
-        className="page-options"
-        isGrouped
-      />
+      <span ref={dropdownRef}>
```
imobachgs, line 140: Suggestion: I prefer a `<div>` instead of `<span>`. After all, the `Dropdown` base element is already a `<div>` so it looks more consistent to me.

Additionally, the `<div>` element supports [flow content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#flow_content) while the `<span>` is limited to [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content).

imobachgs (commented):
Hi @balsa-asanovic,

First of all, sorry for taking so much time to answer. The PR looks great, and it works as expected. I only have a suggestion: using a `<div>` instead of an `<span>`. What do you think?

web/src/components/core/PageOptions.jsx
balsa-asanovic: Yeah definitely, `<div>` is a better choice, sorry for missing this.

I've made the commit to PR addressing this.

imobachgs: No problem, thanks a lot for addressing it 😃

imobachgs (approved):
Now it looks even better 😃. We just released version 3 today, so your contribution will be part of the next release.

Just for the record, I fixed a minor problem in the changes file (not your fault, just a side effect of updating the file with the "Version 3" entry this morning).

I am approving and merging this fix. Once more time, thanks a lot!

PR #672, opened 2023-07-21
a basic .pot (template) file

lkocman:
Basic pot file template for weblate
I did use english strings from the .de file

I'm new to this, therefore not source of truth. This would be an actual .pot file template that I could use over the .de file. Seems like other projects use it too.

https://en.opensuse.org/openSUSE:Localization_guide#About_PO_and_POT_Files

Had to use git add -f as the file was matching the patterns in gitignore

### DIFF
--- a/web/po/agama.pot
+++ b/web/po/agama.pot
@@ -1,0 +1,35 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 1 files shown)


lkocman:
If this gets accepted we need to change template to point to this file

https://l10n.opensuse.org/admin/trans/component/903/change/

lslezak:
The POT file is now generated by a GitHub Action, see more details in https://github.com/openSUSE/agama/pull/670

PR #678, opened 2023-07-31
Handle translations in the webpack devel server

lslezak:
## Problem

When running the webpack development server locally the translations are not loaded, error 404 is reported in the browser log.

## Notes

- This is related to pull request #677
- The Cockpit server returns `po.<LANG>.js` content for the `po.js` request
- The requested language is obtained from the `CockpitLang` cookie sent in the HTTP header
- For missing translation files Cockpit returns an empty script

## Solution

- Reimplement the same logic in the webpack devel server
- Use redirection instead of reading another file (the built JS file only exists in the webpack memory, we cannot read it from disk)

## Testing

- Tested manually with Czech translations and fake Brazil Portuguese translations (to test the `pt_BR` locale behavior)
- The translation file is correctly read, the available translations are used in the web UI
- After updating the PO file the respective JS file is generated automatically by webpack, you only need to reload the page in the browser to load the updated translations

### DIFF
--- a/web/src/index.html
+++ b/web/src/index.html
@@ -13,7 +13,7 @@
     <!--
       The Cockpit server returns "po.<LANG>.js" content for the "po.js" request,
       the requested language is obtained from the "CockpitLang" cookie sent in
-      the HTTP header. For missing translations it returns an empty string.
+      the HTTP header. For missing translation files it returns an empty script.
     -->
     <script src="po.js" defer></script>
   </body>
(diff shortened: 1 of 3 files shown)


mvidner (approved):
Looking at the change it makes sense,
but I don't know how to try it out.

imobachgs (approved):

lslezak:
> Looking at the change it makes sense, but I don't know how to try it out.

```
COCKPIT_TARGET=<IP_ADDRESS> npm run server -- --open
```

where `IP_ADDRESS` points to a VM running the Agama installer

PR #674, opened 2023-07-25
ALP-Dolomite config

joseivanlopez:
## Problem

New ALP-Dolomite product (former ALP-Micro) is not offered for installation yet.

## Solution

Adapt config file to replace ALP-Micro by ALP-Dolomite. Moreover, ALP-Bedrock is removed.

The *agama-live* package was updated to generate the config for the new ALP flavor, see https://build.opensuse.org/package/rdiff/systemsmanagement:Agama:Staging/agama-live?linkrev=base&rev=4.

Note: some patterns are still missing in the new repositories: *patterns-alp-base*, *patterns-alp-cockpit* and *patterns-alp-container_runtime*. The product will not be installable until fixing this issue.

PR #679, opened 2023-07-31
Fix pattern names for Dolomite 1.0 Milestone2

ancorgs:
## Problem

#674 updated the configuration for ALP Dolomite. Unfortunately the names of the patterns are wrong:

- The real name schema for patterns is `alp_whatever` and not `patterns-alp-whatever`
- The package `alp_cockpit` is not in the repository

## Solution

Use the real names of the existing patterns from the repository

## Testing

Tested manually

### DIFF
--- a/service/etc/agama.yaml
+++ b/service/etc/agama.yaml
@@ -150,9 +150,8 @@ ALP-Dolomite:
         archs: ppc
 
     mandatory_patterns:
-      - patterns-alp-base
-      - patterns-alp-cockpit
-      - patterns-alp-hardware
+      - alp_base
+      - alp_hardware
     optional_patterns: null # no optional pattern shared
     mandatory_packages:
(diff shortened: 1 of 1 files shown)


service/etc/agama.yaml
```diff
@@ -150,9 +150,8 @@ ALP-Dolomite:
         archs: ppc
 
     mandatory_patterns:
-      - patterns-alp-base
-      - patterns-alp-cockpit
```
lslezak, line 154: Is removing the cockpit pattern OK?

ancorgs: Yes, @jsrain told me to do so

lslezak: ok, just wanted to be sure...

lslezak (approved):

PR #680, opened 2023-07-31
Use html-webpack-plugin for minifying HTML in production

lslezak:
## Problem

- The HTML file is not minified in production
- This is related to PR #677 which added a quite big comment into the `index.html` file

## Solution

- As the `index.html` is not that big it is not a major problem but IMHO it is a good idea to do it anyway
- Additionally it supports bundling the favicon so we can remove our hack, the favicon tag is injected by the plugin automatically

## Testing

- Tested manually, in the production mode the comments and the extra spacing is removed, in development mode the code is unchanged
- In both cases it injects the favicon tag properly

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -61,6 +61,7 @@
         "eslint-plugin-react": "^7.21.2",
         "eslint-plugin-react-hooks": "^4.1.2",
         "eslint-webpack-plugin": "^4.0.0",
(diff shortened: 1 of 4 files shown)


lslezak:
Note: Let's merge this one *after* releasing the Dolomite image.

imobachgs (approved):

PR #684, opened 2023-08-02
obs-release.yml - Fix the glob to also run for simple "v3" tag

lslezak:
## Problem

The release action was not called for simple "v3" tag

## Solution

Simplify the glob to match any tag starting `v` and a number.


imobachgs (approved):
Thanks!

PR #685, opened 2023-08-03
Webpack dev server: Fixed locale handling (null check)

lslezak:
Properly handle locales with just a language code without country (e.g. "cs").

In that case `country` is `null` and calling `toUpperCase()` fails... :open_mouth:


imobachgs (approved):
It looks good, but you might want to try [the optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator.

imobachgs (approved):
It still LGTM. Thanks!

Issue #688, opened 2023-08-03
Miscellaneous problems with `agama config set`


mvidner:
I found these while trying to test #666.

Good behavior for starters:
```console
# journalctl -f &
# agama config set software.product=foo
Unknown product 'foo'. Available products: '["ALP-Dolomite", "Tumbleweed", "Leap16"]'
# agama config set software.baseProduct=foo
Unknown attribute 'software.base_product'
# agama config set install.product=foo
Unknown attribute 'install.product'
```

CLI syntax errors are silently ignored: 

```console
# agama config set storage magic  # no equal sign
# agama config set storage=magic  # no section-dot-value
```

Now observe the `config show` output. I would expect `config set` and `config add` to use the same vocabulary:

```console
# agama -f yaml config show
user:
  fullName: ''
  userName: ''
  password: ''
  autologin: false
root:
  sshPublicKey: null
software:
  product: ''
storage:
  lvm: null
  encryptionPassword: null
  devices: []
network:
  connections:
  - id: foo
    method: disabled
  - id: lo
    method: manual
    addresses:
    - 127.0.0.1/8
```

Users are bad: 
```console
# agama config set users.blah=foo   # OK
Unknown attribute 'users.blah'
# agama config set user.blah=foo    # WTF, no error
# agama config set users.password=foo  # fair enough
Unknown attribute 'users.password'
# agama config set user.password=foo   # silently fails, bad
```

Storage... seems to work in the config layer but fails somewhere deeper. Might be some outdated dependencies in my container

```console
# agama config set storage.LVM=true
I, [2023-08-03T11:18:10.525118 #16216]  INFO -- storage: Calculating storage proposal from D-Bus settings: {"LVM"=>true, "CandidateDevices"=>[], "EncryptionPassword"=>""}
D-Bus service error

Caused by:
    org.freedesktop.DBus.Error.Failed: undefined method `software_raids' for nil:NilClass; caused by 1 sender=:1.28 -> dest=org.opensuse.Agama.Storage1 serial=7 reply_serial= path=/org/opensuse/Agama/Storage1; interface=org.opensuse.Agama.Storage1.Proposal.Calculator; member=Calculate error_name=
# agama config set storage.lvm=true
I, [2023-08-03T11:18:30.624296 #16216]  INFO -- storage: Calculating storage proposal from D-Bus settings: {"EncryptionPassword"=>"", "CandidateDevices"=>[], "LVM"=>true}
D-Bus service error

Caused by:
[...]

PR #689, opened 2023-08-03
Improve error handling

imobachgs:
* An attempt to address some of the issues reported in #688.
* Fix handling of users/root settings: the CLI ignored the settings but did not report any problem. This bug was introduced in #666. I needed to extend the derive macro to support `flatten` (in serde jargon) fields.
* Extend the `agama_settings` documentation to explain better how it works.

<details>
<summary>Click for a documentation screenshot :wink:</summary>

Screenshot from 2023-08-04 13-42-54
</details>

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -80,6 +80,18 @@ fn parse_config_command(subcommand: ConfigCommands) -> Result<ConfigAction, CliE
     }
 }
 
+/// Split the elements on '=' to make a hash of them.
+///
+/// ```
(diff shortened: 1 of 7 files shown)


mvidner:
OK, this now works:
```
agama config set user.user_name=tux user.password=foo
agama config set root.password=foo
agama config set root.sshPublicKey=Roooot
```

imobachgs:
I would recommend generating the documentation for `agama_settings` with `cargo doc --no-deps --package agama-settings --open`.

teclator (approved):
LGTM

PR #690, opened 2023-08-04
Run the integration tests only when needed

lslezak:
## Problem

- Running the integration tests in the CI always is quite expensive, usually they take 8-10 minutes
- There are many cases when running integration tests is not needed and it just wastes resources and you need to wait longer before merging a pull request

Some examples when the integration tests should be skipped (if only these components are updated):

  - Updating the command line interface
  - Updating a README file or any documentation files
  - Updating an unit test
  - Updating examples
  - Updating `*.spec` or `*.changes` files

### Note

We also run the CI after any commit to a branch. Which means it can also help in this scenario:

- You create a branch and commit a code fix there
- Later you find out that an unit test fails
- You fix the unit test in another commit

The integration tests will run for the first commit but not for the second. When you create a pull request from that branch both integration and unit tests will run.

## Solution

- Use path matching and run the integration tests only when a file relevant for the integration tests is updated
- The path list is not set in stone, we will need to adapt it if some files a removed or new ones are added.


.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,164 @@
...
+  push:
+    paths:
+      # NOTE: GitHub Actions do not allow using YAML references, the same path
+      # list is used below for the pull request event. Keep both lists in sync!!
+
+      # this file itself
+      - .github/workflows/ci-integration-tests.yml
+
+      # the web frontend
+      - web/**.json
```
imobachgs, line 13: I guess these globs are recursive, right?

.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,164 @@
...
+      # copy a simplified ALP config file, it skips the product selection at the beginning
+      run:  podman exec agama bash -c "cp /checkout/playwright/config/agama.yaml /checkout/service/etc/agama.yaml"
+
+    - name: Show NetworkManager log
+      run:  podman exec agama journalctl -u NetworkManager
+
+    - name: Show the D-Bus services log
+      run:  podman exec agama bash -c "journalctl | grep agama"
+
+    - name: Check DBus socket
```
imobachgs, line 134: np

```suggestion
    - name: Check D-Bus socket
```

imobachgs (approved):

.github/workflows/ci-integration-tests.yml
lslezak: "*" matches anything except "/", "**" matches really anything, including "/"
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

PR #691, opened 2023-08-07
Do not consider the arch when filtering the config

imobachgs:
## Problem

The `filter-config.rb` script uses `Agama::Config` to load the configuration before filtering for the wanted products. The problem is that such a class filters out the unsupported architectures, which causes the script to render different configuration files depending on the arch where it runs.

What is worse, when a product is unavailable for a given architecture, it causes the script to fail. For instance, as Leap16 is not supported in `ppc64le`, the following invocation fails:

`filter-config.rb /etc/agama.yaml Tumbleweed Leap16`

The desired result should be the YAML file containing TW and Leap16 products. Agama will filter `Leap16` when running on a `ppc64le` machine.

## Solution

Load and process the configuration file as plain YAML.

## Testing

- *Tested manually*

### DIFF
--- a/service/package/rubygem-agama.changes
+++ b/service/package/rubygem-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Aug  7 10:52:35 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM

PR #692, opened 2023-08-08
Revert "Use html-webpack-plugin for minifying HTML in production"

lslezak:
## Problem

The Agama installer displays an empty screen and the browser log contains errors like:
```
Uncaught TypeError: Cannot read properties of null (reading 'children')
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

Especially this error is strange:
```
Uncaught Error: React Refresh runtime should not be included in the production bundle.
```

According to out Webpack configuration that `react-refresh` plugin should be included only in the development mode, not in production. :thinking: 

## Solution

It turned out that the problem is caused by using the `html-webpack-plugin` which somehow conflicts with the `react-refresh` plugin. When it is enabled the production build contains this in the `index.html` file:

```html
<script defer="defer" src="reactRefreshSetup.js">
```

After reverting #680 it works fine.

That's quite strange, I did several tests with the `html-webpack-plugin` plugin and it worked fine, I'm not sure what I did wrong/differently... :flushed: 

## Replacement

I found the [html-minimizer-webpack-plugin](https://webpack.js.org/plugins/html-minimizer-webpack-plugin/), I'll try it later, maybe it works better...

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -61,7 +61,6 @@
         "eslint-plugin-react": "^7.21.2",
         "eslint-plugin-react-hooks": "^4.1.2",
         "eslint-webpack-plugin": "^4.0.0",
-        "html-webpack-plugin": "^5.5.3",
         "htmlparser": "^1.7.7",
         "identity-obj-proxy": "^3.0.0",
         "jed": "^1.1.1",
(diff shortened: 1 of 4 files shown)


ancorgs (approved):
LGTM. Let's go for the revert while we find something better.

imobachgs:
Thanks for finding the issue!

PR #694, opened 2023-08-09
Run the Rubocop CI check only when a Ruby file is changed

lslezak:
## Problem

- The Rubocop check is run always, even when no Ruby file is modified
- This makes CI slower and wastes some server resources

## Solution

- Put the Rubocop job into a separate file and use the `path` specifiers so it is run only when a matching file is updated

### DIFF
--- a/.github/workflows/ci-rubocop.yml
+++ b/.github/workflows/ci-rubocop.yml
@@ -12,7 +12,7 @@ on:
       - service/.rubocop.yml
       # all Ruby files
       - service/Gemfile
-      - service/bin/agamactl
+      - service/bin/**
       - service/**.rb
 
   pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-rubocop.yml
```diff
@@ -0,0 +1,54 @@
...
+      # NOTE: GitHub Actions do not allow using YAML references, the same path
+      # list is used below for the pull request event. Keep both lists in sync!!
+
+      # this file as well
+      - .github/workflows/ci-rubocop.yml
+      # Rubocop configuration
+      - service/.rubocop.yml
+      # all Ruby files
+      - service/Gemfile
+      - service/bin/agamactl
```
imobachgs, line 15: @teclator is about to add `service/bin/agama-setup-proxy` file, so perhaps we should use `bin/*`.

imobachgs (approved):

PR #696, opened 2023-08-09
Added proxy setup service

teclator:
## Problem

Currently Agama does not have support for setting up a proxy

In case of a network installation _**dracut url lib**_ uses the one provided by the kernel command line option:

https://github.com/openSUSE/dracut/blob/5e324584a14377f52425be589ab53299609f066b/modules.d/45url-lib/url-lib.sh#L62

We should write a service that continue writing the /etc/sysconfig/proxy and use the microos-tools proxy setup service to make it available to the systemd units if we wanted, or own systemd service to do that before running agama.

The service should read the **/proc/cmdline** in case it was given as a kernel commandline option or check the configuration written by dracut when asking for more kernel paramaters

- *https://trello.com/c/gmOcIc0b/3404-5-agama-proxy-configuration-first-simplistic-approach*

## Solution

An agama-proxy-setup service and executable will take care of parsing the /proc/cmdline writing the information to /etc/sysconfig/proxy

The setup-systemd-proxy-env path provided by microos-tools is enabled in order to write the information as environment variables to systemd units.

See https://build.opensuse.org/package/rdiff/home:teclator:branches:systemsmanagement:Agama:Devel/agama-live?opackage=agama-live&oproject=systemsmanagement%3AAgama%3ADevel&rev=5

## Testing

- *Tested manually*

ProxySetup


service/bin/agama-proxy-setup
```diff
@@ -0,0 +1,111 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+# Helper script to create a configuration file for a selected list of products.
+
+require "yast"
+require "uri"
+
+# This class is responsible of parsing the proxy url from the kernel cmdline or configured
+# during the boot proccess of the system
```
imobachgs, line 29: I would like to have an explanation of how it is expected to work (something similar to what you wrote on the solution section of the PR).

service/bin/agama-proxy-setup
```diff
@@ -0,0 +1,111 @@
...
+class SetupProxy
```
imobachgs, line 30: I would move this class to `lib` in order to write proper unit tests.

imobachgs (commented):
In general, it looks good. However, I miss better documentation and tests.

teclator:
An image can be tested from https://download.opensuse.org/repositories/home:/teclator:/branches:/systemsmanagement:/Agama:/Devel/images/iso/

doc/boot_arguments.md
```diff
@@ -26,3 +26,14 @@ agama.web.ssl=true agama.web.ssl_cert=http://192.168.122.1/mycert.pem agama.web.
 ```
 
 Changing complex options (e.g., collections) is not supported yet.
+
+## Proxy Setup
+
+Agama supports proxy setup using the `proxy=` kernel command line option like 
+`proxy=http:192.168.122.1:3128`.
```
mvidner, line 33: Is this supposed to be a URL?

```suggestion
`proxy=http://192.168.122.1:3128`.
```

If this setting is compatible with some standard or other tools, we should mention it here

service/lib/agama/proxy_setup.rb
```diff
@@ -0,0 +1,110 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+# Helper script to create a configuration file for a selected list of products.
```
mvidner, line 22: Leftover description from something else

service/lib/agama/proxy_setup.rb
```diff
@@ -0,0 +1,110 @@
...
+  # during the boot proccess of the system writing the configuration to /etc/sysconfig/proxy
+  class ProxySetup
+    include Singleton
+    include Yast
+    include Logger
+
+    CMDLINE_PATH = "/etc/cmdline"
+    CMDLINE_MENU_CONF = "/etc/cmdline-menu.conf"
+
+    attr_accessor :proxy
```
mvidner, line 38:
```suggestion
    # @return [URI::Generic]
    attr_accessor :proxy
```

service/lib/agama/proxy_setup.rb
```diff
@@ -0,0 +1,110 @@
...
+        ex["ftp_proxy"] = proxy.to_s
+      end
+      ex["enabled"] = true
+      ex
+    end
+
+    def write
+      return unless proxy
+
+      Proxy.Read
```
mvidner, line 100: Does this mean that the tool will *merge* with any existing proxy configuration?
We should document it if it's intentional.

doc/boot_arguments.md
```diff
@@ -26,3 +26,15 @@ agama.web.ssl=true agama.web.ssl_cert=http://192.168.122.1/mycert.pem agama.web.
 ```
 
 Changing complex options (e.g., collections) is not supported yet.
+
+## Proxy Setup
+
+Agama supports proxy setup using the `proxy=` kernel command line option like 
+`proxy=http://192.168.122.1:3128` when the installation requires to use an HTTP, HTTPS or FTP
+source . The supported proxy URL format is: protocol://[user[:password]@]host[:port]
```
imobachgs, line 34:
```suggestion
source. The supported proxy URL format is: protocol://[user[:password]@]host[:port]
```

service/lib/agama/proxy_setup.rb
```diff
@@ -0,0 +1,109 @@
...
+      write
+    end
+
+  private
+
+    def read
+      self.proxy = proxy_from_cmdline || proxy_from_dracut
+    end
+
+    # TODO
```
imobachgs, line 55: A left-over?

service/lib/agama/proxy_setup.rb
```diff
@@ -0,0 +1,109 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "uri"
+
+module Agama
+  # This class is responsible of parsing the proxy url from the kernel cmdline or configured
+  # during the boot proccess of the system writing the configuration to /etc/sysconfig/proxy
```
imobachgs, line 27: It would be nice to mention that the class tries to read the proxy settings from dracut and the kernel's command line.

setup-service.sh
```diff
@@ -64,38 +64,38 @@ $SUDO cp -v $MYDIR/service/share/dbus.conf /usr/share/dbus-1/agama.conf
...
+	# cleanup previous installation
+	[[ -d $DBUSDIR ]] && $SUDO rm -r $DBUSDIR
+
+	# create services
+	$SUDO mkdir -p $DBUSDIR
+	for SVC in org.opensuse.Agama*.service; do
+		sudosed "s@\(Exec\)=/usr/bin/@\1=$MYDIR/service/bin/@" $SVC $DBUSDIR/$SVC
+	done
+	sudosed "s@\(ExecStart\)=/usr/bin/@\1=$MYDIR/service/bin/@" \
+		agama.service /usr/lib/systemd/system/agama.service
```
imobachgs, line 79: You have changed the indentation of the whole file to change this line. I am not sure if that's correct at all.

imobachgs (commented):
It looks good. However I am not sure if we should keep the indentation in the `setup-service.sh` script.

setup-service.sh
mvidner: Ah, we have `.editorconfig` in `service` and `web` but not here at the top level :thinking:

service/lib/agama/proxy_setup.rb
teclator: That is what the 'or configured during the boot proccess..' referred, the thing is that it is still not implemented.

service/lib/agama/proxy_setup.rb
teclator: It is still not implemented, just a discussed POC with the agreement on the filename, I guess we could remove the TODO as it should work.

imobachgs (approved):

imobachgs (changes requested):
Oh, sorry, I forgot about the changes file. Please, add an entry and do not forget the bsc and/or github references.

imobachgs (approved):

PR #700, opened 2023-08-11
Mark translatable texts

lslezak:
## Problem

- The texts in the web UI are not marked for translation

## Notes 

- This is just the first round, it does not cover all files, let's do it in smaller steps
- Adjusted building the POT file using `xgettext`, originally copied from the Cockpit code but make it better:
  - Use the `JavaScript` format (it betters parses the code and does not add the `c-format` marks)
  - Define only the keywords used by Agama (e.g. we do not use contexts)


imobachgs (approved):
LGTM

PR #702, opened 2023-08-11
Support for special testing "xx" language

lslezak:
## Problem

- It is difficult to find out which texts are marked for translation
- Or to find out untranslated string

## Solution

- "Translate" all texts in the `_()` and `n_()` functions, just convert them to X sequences
  - Just be careful of special `%s` or `%d` sequences, those should be kept

## Notes

- This of course handles only the texts generated by the web UI itself, it cannot mark the texts coming from the backed

## Screenshots

- Use `?lang=xx` URL parameter, this can be entered only manually, we should never allow this in the UI

xx_lang

### DIFF
--- a/web/src/L10nWrapper.jsx
+++ b/web/src/L10nWrapper.jsx
@@ -19,7 +19,7 @@
  * find current contact information at www.suse.com.
  */
 
-// import React from "react";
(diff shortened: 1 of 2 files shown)


imobachgs (approved):
Interesting idea. LGTM. 😃

Issue #707, opened 2023-08-18
Startup and timing issues


imobachgs:
* Trello: https://trello.com/c/gk7xGLzB/
* Related to [bsc#1213194](https://bugzilla.suse.com/show_bug.cgi?id=1213194).

**The issue described in the bug report could result from several troubles. Let's try to collect them here. We are still investigating the problem, so our conclusions can be correct... or not.**

## Manager status on single-product scenarios

All Agama services implement the `ServiceStatus` D-Bus interface, which determines whether the service is busy. In a nutshell, when a service receives a request to perform a "blocking" operation (like probing, installing, etc.), it sets its status as busy. You can check the [storage D-Bus method as an example](https://github.com/imobachgs/agama/blob/032ab17145e29ef48bc712a32a5180e966e5349d/service/lib/agama/dbus/storage/manager.rb#L81-L83).

However, the current implementation has a small problem: it works at D-Bus level. So when the operation is not triggered through D-Bus, the status is not updated. And that's precisely the case in single-product scenarios (like `ALP`). The manager jumps into the next phase automatically (see [Agama::Manager#startup_phase](https://github.com/openSUSE/agama/blob/1f9547c59d96052af2609bb24a539b2d026893cb/service/lib/agama/manager.rb#L66)).

In that case, any client may think that the service is idle when it is not (it might reading the repositories). In case of the auto-installation, it may [try to start the installation](https://github.com/openSUSE/agama/blob/1f9547c59d96052af2609bb24a539b2d026893cb/rust/agama-cli/src/main.rs#L68) producing the `Could not start the installation...` message.

Fixed by #708.

## Manager timeout

In a single-product scenario, Agama analyzes the storage devices and reads the repositories from the beginning. In this situation, the manager service might be unresponsive while waiting for storage and software services. This problem is mitigated by dispatching D-Bus calls during progress reporting, but sometimes that's not enough.

It might be the reason behind `Failed to activate service 'org.opensuse.Agama1': timed out` messages.

imobachgs:
After a few successful tests we are considering the issue as fixed. We will open a new bug report (with fresh data) if tests fail again.

PR #708, opened 2023-08-21
Fix manager service status during startup

imobachgs:
## Problem

See #707, "Manager status on single-product scenarios".

## Solution

Let the manager set the service status outside of D-Bus. It does not solve all timing issues, but the `Could not start the installation...` should not happen.

## Testing

- Fixed unit tests
- Tested manually

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -25,6 +25,7 @@ require "agama/network"
 require "agama/with_progress"
 require "agama/installation_phase"
 require "agama/service_status_recorder"
+require "agama/dbus/service_status"
 require "agama/dbus/clients/locale"
 require "agama/dbus/clients/software"
 require "agama/dbus/clients/storage"
@@ -50,6 +51,9 @@ module Agama
     # @return [InstallationPhase]
     attr_reader :installation_phase
 
+    # @return [DBus::ServiceStatus]
+    attr_reader :service_status
+
     # Constructor
     #
     # @param logger [Logger]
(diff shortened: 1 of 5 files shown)


service/lib/agama/manager.rb
```diff
@@ -58,18 +62,22 @@ def initialize(config, logger)
       @logger = logger
       @installation_phase = InstallationPhase.new
       @service_status_recorder = ServiceStatusRecorder.new
+      @service_status = DBus::ServiceStatus.new.busy
```
mvidner, line 65: This looks like the manager will only rarely be _idle_ and [`safe_run`](https://github.com/openSUSE/agama/blob/40e46fa183077be4115de6491650c7957f92f121/service/lib/agama/dbus/manager.rb#L146) will fail more often?

jreidinger: I would not say rarely, but this ensures that it waits for startup phase, not?

service/package/rubygem-agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Aug 21 11:15:50 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Set the manager service as busy during the startup phase when
+  a single product is available (bsc#1213194).
```
jreidinger, line 5: I would say that this change is misleading..it sets busy always and that fixed issue with single product. But set it **always** during startup

service/lib/agama/manager.rb
imobachgs: Yes, that's the point. It starts as busy, but it is set as idle as soon as the startup_phase finishes.

mvidner: Ok then

jreidinger (approved):

PR #711, opened 2023-08-22
Proxy install

teclator:
## Problem

In #696 we added support for configuring the proxy when booting the installation system but the configuration is lost after the  installation.

- https://trello.com/c/gmOcIc0b/3404-5-agama-proxy-configuration-first-simplistic-approach

## Solution

The proxy configuration will be copied to the target system when a proxy is used as well as the **microos-tools** package will be installed.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/service/lib/agama/proxy_setup.rb
+++ b/service/lib/agama/proxy_setup.rb
@@ -22,6 +22,7 @@
 require "yast"
 require "uri"
 require "fileutils"
+require "agama/helpers"
 
 module Agama
   # This class is responsible of parsing the proxy url from the kernel cmdline or configured
@@ -31,6 +32,7 @@ module Agama
     include Singleton
     include Yast
     include Logger
+    include Helpers
 
     CMDLINE_PATH = "/proc/cmdline"
     CMDLINE_MENU_CONF = "/etc/cmdline-menu.conf"
@@ -54,11 +56,13 @@ module Agama
     end
 
     def install
-      Proxy.Read
-      return unless Proxy.enabled
+      on_local do
+        Proxy.Read
+        return unless Proxy.enabled
 
-      copy_files
-      add_packages
+        copy_files
+        add_packages
+      end
     end
 
   private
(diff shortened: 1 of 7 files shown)


service/lib/agama/proxy_setup.rb
```diff
@@ -30,23 +32,39 @@ class ProxySetup
...
       write
     end
 
+    def install
+      on_local do
+        Proxy.Read
+        return unless Proxy.enabled
+
+        copy_files
+        add_packages
```
teclator, line 64: We probably will need to do it separately as the packages should be added before the software is installed but then the microos-tools https://github.com/openSUSE/microos-tools/blob/master/systemd-proxy-env/setup-systemd-proxy-env.path need to be enabled later as well as the copy of the files.

service/lib/agama/proxy_setup.rb
```diff
@@ -48,20 +48,25 @@ def initialize
...
+      Proxy.Read
     end
 
     def run
       read
       write
     end
 
+    def propose
+      on_target { add_packages } if Proxy.enabled
```
imobachgs, line 61: No need to call `on_target`. If this code runs as part of the manager service, it uses D-Bus to talk to the software service, which takes care of installing that package later. Or am I missing something?

teclator: You are right, the problem is that I'm calling it after software proposal, but fixed in next commit.

imobachgs (commented):
I looks good to me, although I think that the logic to handle the proxy configuration should be in the Network service (instead of the Manager). However, let's consider the current implementation good enough.

Please add the entry to the changes file.

teclator:
> It looks good to me, although I think that the logic to handle the proxy configuration should be in the Network service (instead of the Manager). However, let's consider the current implementation good enough.
> 
> Please add the entry to the changes file.

I agree, lets do it separately by now and then integrate it in the network service implementing it completely in Rust.

imobachgs:
BTW, the integration tests are failing, but it is a known (and already) fixed issue. It would be fine after you merge.

imobachgs (approved):

PR #715, opened 2023-08-23
Install the ppc64-diag package

imobachgs:
* [bsc#1206898](https://bugzilla.suse.com/show_bug.cgi?id=1206898).

Consider this PR as a workaround while the patterns are improved.


jreidinger (approved):

hramrach:
The arch is ppc64le

imobachgs:
OK, I am filtering for `ppc64` only (in YaST, we do not distinguish between LE/BE).

PR #718, opened 2023-08-28
[RFC] Integration tests for `agama-dbus-server`

imobachgs:
## Problem

Test coverage of the Rust code is pretty low (around 16%). In Ruby we are used to inject and mock the dependencies. In Rust that's perfectly possible if you rely a lot on traits. But if that's not the case, it can be much more challenging.

See [mockall](https://crates.io/crates/mockall) documentation for examples.

## Solution

Using unit tests to cover code that does not rely on D-Bus or other external dependencies is pretty straightforward. We should keep doing that.

However, you are in trouble when you depend on a connection or a proxy. For those cases, I decided to go with [Rust integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html).

This PR includes some minimal infrastructure to test the `agama-dbus-daemon` package. It sets up a [D-Bus server](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) that you can use for testing.

## Results

* After writing two simple test cases, I **found a bug** and raised the coverage **from 16% to 36%**. Not bad.
* To make the code testable, I had to allow injecting a network adapter (fortunately, we already had the concept of `Adapter`) to the `NetworkSystem`. I wrote a `NetworkTestAdapter` that allows setting up a custom network state.

## Future

* Extend the tests to cover other services and try to extract common patterns. We can extend our test harness to support them.
* Improve error handling and get rid of `unwrap()` or `expect()` calls where it makes sense.
* As an experiment, I would like to split the `DBusServer` struct into two different ones: `RunningDBusServer` and `DBusServer`.

### DIFF
--- a/rust/agama-dbus-server/tests/common/mod.rs
+++ b/rust/agama-dbus-server/tests/common/mod.rs
@@ -1,6 +1,7 @@
 use agama_lib::{connection_to, error::ServiceError};
 use async_std::stream::StreamExt;
 use std::process::{Child, Command};
+use std::time::Duration;
 use uuid::Uuid;
 use zbus::{MatchRule, MessageStream, MessageType};
 
@@ -80,8 +81,8 @@ impl DBusServer {
     /// Waits until the D-Bus daemon is ready.
     // TODO: implement proper waiting instead of just using a sleep
     fn wait(&self) {
-        let wait_time = std::time::Duration::from_millis(500);
-        std::thread::sleep(wait_time);
+        const WAIT_TIME: Duration = Duration::from_millis(500);
+        std::thread::sleep(WAIT_TIME);
     }
 
     /// Builds a message stream.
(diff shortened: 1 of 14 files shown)


rust/agama-dbus-server/tests/common/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+            if &sname == name {
+                return;
+            }
+        }
+    }
+
+    /// Waits until the D-Bus daemon is ready.
+    // TODO: implement proper waiting instead of just using a sleep
+    fn wait(&self) {
+        let wait_time = std::time::Duration::from_millis(500);
```
jreidinger, line 83: why let and not `const`?

rust/agama-lib/src/network/client.rs
```diff
@@ -181,4 +193,20 @@ impl<'a> NetworkClient<'a> {
         proxy.set_password(&wireless.password).await?;
         Ok(())
     }
+
+    // Replace with a better implemenation based on signals.
```
jreidinger, line 197: it looks like dead code for me

PR #719, opened 2023-08-28
Mark more texts for translation

lslezak:
Just marking more texts for translation

### DIFF
--- a/web/src/components/software/ProductSelectionPage.jsx
+++ b/web/src/components/software/ProductSelectionPage.jsx
@@ -23,8 +23,7 @@ import React, { useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { useInstallerClient } from "~/context/installer";
 import { useSoftware } from "~/context/software";
-import format from "format-util";
-import { _, n_ } from "~/i18n";
+import { _ } from "~/i18n";
 
 import {
   Button,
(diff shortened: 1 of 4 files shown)


web/src/components/software/ProductSelectionPage.jsx
```diff
@@ -88,16 +90,27 @@ function ProductSelectionPage() {
...
         </Button>
       </MainActions>
 
       <Form id="product-selector" onSubmit={accept}>
-        <FormGroup isStack label={`Choose a product (${products.length} available)`} role="radiogroup">
+        <FormGroup
+          isStack
+          label={
+            format(
+              n_("Choose a product (%d available)", "Choose a product (%d available)", products.length),
```
lslezak, line 108: Here I'm not sure if we really need to display the number of available products.

- There are usually just few products
- The number does not provide any meaningful information for the user, you can always choose *only one* product, it does not matter how many products are available

So I'd rather make it easier and use just simple `_("Choose a product")` here.

Opinions?

imobachgs: Yes, I agree, I would get rid of the number of available products.

imobachgs (approved):

web/src/components/software/ProductSelectionPage.jsx
```diff
@@ -88,16 +89,18 @@ function ProductSelectionPage() {
 
   return (
     <>
-      <Title>Product selection</Title>
+      {/* TRANSLATORS: page header */}
+      <Title>{_("Product selection")}</Title>
```
jreidinger, line 93: not sure if in title it should not be all capitalized like `Product Selection`

imobachgs: Why? Do we have any guidelines about that?

PR #723, opened 2023-08-29
Add support to assign an interface or match settings to a connection

teclator:
## Problem

Agama allows to add/update network connections. However, it relies on **NetworkManager** to decide which interface to use. Alternatively, Agama should allow the user to assign a connection to an specific interface.

Instead of using just an interface name, **Agama** should offer some matching capabilities, just [like **NetworkManager**](https://manpages.debian.org/testing/network-manager/nm-settings.5.en.html#match_setting). How the API should look like is up to who implements the feature but the interface-name, the path and the driver should be supported.

- https://trello.com/c/H2ISHKvi/3395-2-agama-allow-assigning-a-network-configuration-to-an-specific-interface

## Solution

It has been added support to specify to which interfaces the connection should be bond to supporting it directly through the **interface name** attribute or through a set of match settings [as them are supported by **NetworkManager**](https://manpages.debian.org/testing/network-manager/nm-settings.5.en.html#match_setting).

```json
"network": {
    "connections": [
      {
        "id": "Ethernet network device 1",
        "method": "manual",
        "interface": "eth0",
        "addresses": [
          "192.168.122.100/24"
        ],
        "gateway": "192.168.122.1",
        "nameservers": [
          "192.168.122.1"
        ]
      }
    ]
  }
```

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/network/nm/dbus.rs
+++ b/rust/agama-dbus-server/src/network/nm/dbus.rs
@@ -177,42 +177,35 @@ fn wireless_config_to_dbus(conn: &WirelessConnection) -> NestedHash {
 ///
 /// * `match_config`: MatchConfig to convert.
 fn match_config_to_dbus(match_config: &MatchConfig) -> HashMap<&str, zvariant::Value> {
-    let mut match_config_dbus = HashMap::new();
-
     let drivers: Value = match_config
         .driver
         .iter()
-        .map(|dr| dr.to_string())
-        .collect::<Vec<String>>()
+        .cloned()
+        .collect::<Vec<_>>()
         .into();
-    match_config_dbus.insert("driver", drivers);
 
     let kernels: Value = match_config
         .kernel
         .iter()
-        .map(|dr| dr.to_string())
-        .collect::<Vec<String>>()
+        .cloned()
+        .collect::<Vec<_>>()
         .into();
-    match_config_dbus.insert("kernel-command-line", kernels);
 
-    let paths: Value = match_config
-        .path
-        .iter()
-        .map(|dr| dr.to_string())
-        .collect::<Vec<String>>()
-        .into();
-    match_config_dbus.insert("path", paths);
+    let paths: Value = match_config.path.iter().cloned().collect::<Vec<_>>().into();
 
     let interfaces: Value = match_config
         .interface
         .iter()
-        .map(|dr| dr.to_string())
-        .collect::<Vec<String>>()
+        .cloned()
+        .collect::<Vec<_>>()
         .into();
 
-    match_config_dbus.insert("interface-name", interfaces);
-
-    match_config_dbus
+    HashMap::from([
+        ("driver", drivers),
+        ("kernel-command-line", kernels),
+        ("path", paths),
+        ("interface-name", interfaces),
+    ])
 }
 
 fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
@@ -231,7 +224,7 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
 
     if let Some(interface) = connection.get("interface-name") {
         let interface: &str = interface.downcast_ref()?;
-        base_connection.interface = interface.parse().unwrap();
+        base_connection.interface = interface.to_string();
     }
 
     if let Some(match_config) = conn.get("match") {
@@ -254,7 +247,7 @@ fn match_config_from_dbus(
         let drivers = drivers.downcast_ref::<zbus::zvariant::Array>()?;
         for driver in drivers.get() {
             let driver: &str = driver.downcast_ref()?;
-            match_conf.driver.push(driver.parse().unwrap());
+            match_conf.driver.push(driver.to_string());
         }
     }
 
@@ -262,7 +255,7 @@ fn match_config_from_dbus(
         let interface_names = interface_names.downcast_ref::<zbus::zvariant::Array>()?;
         for name in interface_names.get() {
             let name: &str = name.downcast_ref()?;
-            match_conf.interface.push(name.parse().unwrap());
+            match_conf.interface.push(name.to_string());
         }
     }
 
@@ -270,7 +263,7 @@ fn match_config_from_dbus(
         let paths = paths.downcast_ref::<zbus::zvariant::Array>()?;
         for path in paths.get() {
             let path: &str = path.downcast_ref()?;
-            match_conf.path.push(path.parse().unwrap());
+            match_conf.path.push(path.to_string());
         }
     }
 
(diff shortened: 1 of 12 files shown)


rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +173,63 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
...
         base_connection.ipv4 = ipv4_config_from_dbus(ipv4)?;
     }
 
     Some(base_connection)
 }
 
+fn match_config_from_dbus(
+    match_config: &HashMap<String, zvariant::OwnedValue>,
+) -> Option<MatchConfig> {
+    let mut match_conf = MatchConfig {
```
imobachgs, line 194: Please, use `let mut match_conf = MatchConfig::default()`.

imobachgs (approved):

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +229,62 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
...
+fn match_config_from_dbus(
+    match_config: &HashMap<String, zvariant::OwnedValue>,
+) -> Option<MatchConfig> {
+    let mut match_conf = MatchConfig::default();
+
+    if let Some(drivers) = match_config.get("driver") {
+        let drivers = drivers.downcast_ref::<zbus::zvariant::Array>()?;
+        for driver in drivers.get() {
+            let driver: &str = driver.downcast_ref()?;
+            match_conf.driver.push(driver.parse().unwrap());
```
imobachgs, line 257: What are you "parsing"? Just converting to a String should be enough.

```suggestion
            match_conf.driver.push(driver.to_string());
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +229,62 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
...
+        }
+    }
+
+    if let Some(interface_names) = match_config.get("interface-name") {
+        let interface_names = interface_names.downcast_ref::<zbus::zvariant::Array>()?;
+        for name in interface_names.get() {
+            let name: &str = name.downcast_ref()?;
+            match_conf.interface.push(name.parse().unwrap());
```
imobachgs, line 265:
```suggestion
            match_conf.interface.push(name.to_string());
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +229,62 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
...
+        }
+    }
+
+    if let Some(paths) = match_config.get("path") {
+        let paths = paths.downcast_ref::<zbus::zvariant::Array>()?;
+        for path in paths.get() {
+            let path: &str = path.downcast_ref()?;
+            match_conf.path.push(path.parse().unwrap());
```
imobachgs, line 273:
```suggestion
            match_conf.path.push(path.to_string());
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +229,62 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
...
+        }
+    }
+
+    if let Some(kernel_options) = match_config.get("kernel-command-line") {
+        let options = kernel_options.downcast_ref::<zbus::zvariant::Array>()?;
+        for option in options.get() {
+            let option: &str = option.downcast_ref()?;
+            match_conf.kernel.push(option.parse().unwrap());
```
imobachgs, line 281:
```suggestion
            match_conf.kernel.push(option.to_string());
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -173,13 +229,62 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
         ..Default::default()
     };
 
+    if let Some(interface) = connection.get("interface-name") {
+        let interface: &str = interface.downcast_ref()?;
+        base_connection.interface = interface.parse().unwrap();
```
imobachgs, line 234:
```suggestion
        base_connection.interface = interface.to_string();
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -159,6 +173,48 @@ fn wireless_config_to_dbus(conn: &WirelessConnection) -> NestedHash {
...
+/// Converts a MatchConfig struct into a HashMap that can be sent over D-Bus.
+///
+/// * `match_config`: MatchConfig to convert.
+fn match_config_to_dbus(match_config: &MatchConfig) -> HashMap<&str, zvariant::Value> {
+    let mut match_config_dbus = HashMap::new();
+
+    let drivers: Value = match_config
+        .driver
+        .iter()
+        .map(|dr| dr.to_string())
```
imobachgs, line 185: You intention here is to `clone` the strings, right? You can use the [cloned](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned) function:

```suggestion
        .cloned()
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -159,6 +173,48 @@ fn wireless_config_to_dbus(conn: &WirelessConnection) -> NestedHash {
...
+        .collect::<Vec<String>>()
```
imobachgs, line 186: note: No need to specify the type of the element. However, I am fine with keeping it.

```suggestion
        .collect::<Vec<_>>()
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -159,6 +173,48 @@ fn wireless_config_to_dbus(conn: &WirelessConnection) -> NestedHash {
...
+    let interfaces: Value = match_config
+        .interface
+        .iter()
+        .map(|dr| dr.to_string())
+        .collect::<Vec<String>>()
+        .into();
+
+    match_config_dbus.insert("interface-name", interfaces);
+
+    match_config_dbus
```
imobachgs, line 215: You can do:

```rust
    HashMap::from([
        ("driver", drivers),
        ("kernel-command-line", kernels),
        ("path", paths),
        ("interface-name", interfaces),
    ])
```

It looks more readable to me and it is easy to find out what is included in the HashMap.

imobachgs (changes requested):

rust/agama-dbus-server/src/network/nm/dbus.rs
teclator: Yep, originally did so, then added some logic to inserting only the key if it was not empty but as we are inserting it always now it makes sense to do so. Thanks

imobachgs (approved):

PR #725, opened 2023-08-30
UI locale dbus

jreidinger:
## Problem

There are some string that come from backend and it is displayed to user. But we do not set localization for those strings.

trello: https://trello.com/c/Jp7S5kwh/3398-5-agama-add-support-for-translating-the-messages-from-the-d-bus-interface

## Solution

Implement in Locale method to enlist all possible backend localization ( so far only for ruby based ones ) and allow to set localization. In other services have helper that will set initial localization and also catch signal when it is changed.
As part of this PR also drop no longer used Language Service.

## Testing

- *Tested manually*

## Screenshots

Example localization in czech ( in czech is only data comming from dbus ):

localized_actions


lslezak (commented):
Looks OK to me, but someone with Rust knowledge should check it as well... :smiley:

rust/agama-dbus-server/src/locale.rs
```diff
@@ -89,6 +90,40 @@ impl Locale {
...
+    ///   assert!(locale.list_ui_locales().unwrap().len() > 0);
+    /// ```
+    #[dbus_interface(name = "ListUILocales")]
+    pub fn list_ui_locales(&self) -> Result<Vec<String>, Error> {
+        // english is always available ui localization
+        let mut result = vec!["en".to_string()];
+        const DIR : &str = "/usr/share/YaST2/locale/";
+        let entries = read_dir(DIR).context("Reading YaST2 locale")?;
+        for entry in entries {
+            let entry = entry.context("Failed to read entry in YaST2 locale dir")?;
```
imobachgs, line 120: So we are returning an error if we cannot read any of the dirs, right?

service/lib/agama/ui_locale.rb
```diff
@@ -0,0 +1,57 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+
+Yast.import "WFM"
+
+module Agama
+  # Object responsible for managing changes of localization produced by dbus backend.
```
imobachgs, line 27:
```suggestion
  # Object responsible for managing changes of localization produced by D-Bus backend.
```

imobachgs (approved):
It looks good to me. Just a few suggestions.

rust/agama-dbus-server/src/locale.rs
jreidinger: yeah, any read failure result in error. I would not expect it unless there is some issue with fs.

lslezak: The `/usr/share/YaST2/locale/` directory is owned by the base `yast2` package so the error should not happen normally.

lslezak (approved):

PR #726, opened 2023-08-30
Translate the storage texts

lslezak:
- Just marking more texts for translation, this time in the storage area
- Additionally I edited the `InstallButton.jsx` file to use square brackets - it looks better and it is more compatible with the Markdown format (it is cheap to do it now, we do not have translations yet)

### DIFF
--- a/web/src/components/storage/ProposalVolumes.jsx
+++ b/web/src/components/storage/ProposalVolumes.jsx
@@ -27,7 +27,9 @@ import {
   Toolbar, ToolbarContent, ToolbarItem
 } from "@patternfly/react-core";
 import { TableComposable, Thead, Tr, Th, Tbody, Td } from '@patternfly/react-table';
+import format from "format-util";
 
+import { _ } from "~/i18n";
 import { Em, If, Popup, RowActions, Tip } from '~/components/core';
 import { Icon } from '~/components/layout';
 import { VolumeForm } from '~/components/storage';
@@ -50,12 +52,16 @@ const AutoCalculatedHint = (volume) => {
 
   return (
     <>
-      These limits are affected by:
+      {/* TRANSLATORS: header for a list of items */}
+      {_("These limits are affected by:")}
       <List>
         {volume.snapshotsAffectSizes &&
-          <ListItem>The configuration of snapshots</ListItem>}
+        // TRANSLATORS: list item, this affects the computed partition size limits
+        <ListItem>{_("The configuration of snapshots")}</ListItem>}
         {volume.sizeRelevantVolumes && volume.sizeRelevantVolumes.length > 0 &&
-          <ListItem>Presence of other volumes ({volume.sizeRelevantVolumes.join(", ")})</ListItem>}
+        // TRANSLATORS: list item, this affects the computed partition size limits
+        // %s is replaced by a list of the volumes (like "/home, /boot")
+          <ListItem>{format(_("Presence of other volumes (%s)"), volume.sizeRelevantVolumes.join(", "))}</ListItem>}
       </List>
     </>
   );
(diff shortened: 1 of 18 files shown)


web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -50,12 +52,16 @@ const AutoCalculatedHint = (volume) => {
...
+      {/* TRANSLATORS: header for a list of items */}
+      {_("These limits are affected by:")}
       <List>
         {volume.snapshotsAffectSizes &&
-          <ListItem>The configuration of snapshots</ListItem>}
+        // TRANSLATORS: list item, this affects the computed partition size limits
+        <ListItem>{_("The configuration of snapshots")}</ListItem>}
         {volume.sizeRelevantVolumes && volume.sizeRelevantVolumes.length > 0 &&
-          <ListItem>Presence of other volumes ({volume.sizeRelevantVolumes.join(", ")})</ListItem>}
+        // TRANSLATORS: list item, this affects the computed partition size limits
```
joseivanlopez, line 62: NP: is correctly indented?

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -109,30 +115,33 @@ const GeneralActions = ({ templates, onAdd, onReset }) => {
             key="reset"
             onClick={onReset}
           >
-            Reset to defaults
+            {/* TRANSLATORS: dropdown menu label */}
```
joseivanlopez, line 118: Why do we need the comment between brackets? There are more cases below.

web/src/components/storage/VolumeForm.jsx
```diff
@@ -130,14 +142,21 @@ const SizeManual = ({ errors, formData, onChange }) => {
...
             id="size"
             name="size"
-            aria-label="Exact size"
+            // TRANSLATORS: requested partition size
+            aria-label={_("Exact size")}
+            // TODO: support also localization for numbers, e.g. decimal comma,
+            // either use toLocaleString()
+            //   (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)
+            // or use the "globalize" JS library which can also parse the localized string back
+            //   (https://github.com/globalizejs/globalize#number-module)
```
joseivanlopez, line 151: :+1:

web/src/components/storage/VolumeForm.jsx
```diff
@@ -224,6 +247,16 @@ const SizeRange = ({ errors, formData, onChange }) => {
   );
 };
 
+// constants need to be marked for translation with N_() and translated with _() later
+const SIZE_OPTION_LABELS = Object.freeze({
+  // TRANSLATORS: radio button label, fully automatically computed partition size, no user input
+  auto: N_("Auto"),
```
joseivanlopez, line 253: Question: why `N_` instead of `_` ?

web/src/components/storage/ProposalVolumes.jsx
lslezak: Because this is a JSX code, not regular JS.  The usual JS comments do not work here, you would get syntax error.
With curly braces you switch to JS code and there you can use the usual JS comments.

See more details e.g. in https://wesbos.com/react-jsx-comments

web/src/components/storage/ProposalVolumes.jsx
lslezak: Yes. Otherwise the `eslint` check would complain and fail. We have quite strict CI for that. :wink:

web/src/components/storage/VolumeForm.jsx
lslezak: That constant is evaluated too early, at that point the `cockpit.language` is not initialized yet and the text would not be translated.

web/src/components/storage/ProposalVolumes.jsx
joseivanlopez: Ah right! Sorry, I did not realize it was inside a JSX block.

web/src/components/storage/VolumeForm.jsx
joseivanlopez: Ahh, thanks for the explanation.

dgdavid: Hi guys!

Yesterday I found this looking more information about [N_](https://github.com/openSUSE/agama/blob/704b693ca563e48f6ead2481fa5e73224fa431c6/web/src/i18n.js#L94-L127) when reviewing [ancorgs's PR](https://github.com/openSUSE/agama/pull/1090/files#diff-0b738eb119a000a6ab33ab9e5234cc455461e03d12148e47a866140b2d3ec247R51). I know that `N_` is used to mark for translation, but I do not understand why it is needed either, here and in the @ancorgs's PR linked above.

Also, I wonder if, in case of really needed, this will change with the [new architecture](https://github.com/openSUSE/agama/discussions/1000).

lslezak: The `N_` macro/function used for translating constants, it does not depend on the architecture or programming language (you can use use it even in C or C++ code).

The problem is that 
```
Object.freeze({
  auto: _("Auto"),
   ...
```

would translate the value using the language configured at the initialization, it would not retranslate the text after changing the language later at runtime.

https://yastgithubio.readthedocs.io/en/latest/localization/#translating-constants

PR #735, opened 2023-09-01
Use the sprintf-js library instead of format-util

lslezak:
## Problem

- We use the `format-util` library for formatting strings, but that is too simple
- It only supports the basic `%s` or `%d` formats and does not support more complex cases like changing the order of arguments (`%2$s`) which might be needed in translations

## Solution

- Lets switch to the [sprintf-js](https://github.com/alexei/sprintf.js) library which can handle that

### DIFF
--- a/web/src/components/storage/VolumeForm.jsx
+++ b/web/src/components/storage/VolumeForm.jsx
@@ -27,7 +27,7 @@ import {
   Radio,
   TextInput
 } from "@patternfly/react-core";
-import format from "format-util";
+import { sprintf } from "sprintf-js";
 
 import { _, N_ } from "~/i18n";
 import { If, NumericTextInput } from '~/components/core';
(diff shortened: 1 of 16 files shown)


web/src/components/storage/VolumeForm.jsx
```diff
@@ -98,14 +98,14 @@ const SizeAuto = ({ volume }) => {
   if (volume.sizeRelevantVolumes && volume.sizeRelevantVolumes.length > 0)
     // TRANSLATORS: item which affects the final computed partition size
     // %s is replaced by a list of mount points like "/home, /boot"
-    conditions.push(format(_("the presence of the file system for %s"),
-                           // TRANSLATORS: conjunction for merging two list items
-                           volume.sizeRelevantVolumes.join(_(", "))));
+    conditions.push(sprintf(_("the presence of the file system for %s"),
+      // TRANSLATORS: conjunction for merging two list items
+                            volume.sizeRelevantVolumes.join(_(", "))));
```
lslezak, line 103: Note: This strange indentation is required by eslint, I guess there is some buggy behavior...

imobachgs (approved):

PR #742, opened 2023-09-07
Enhance the i18n documentation

lslezak:
Just updating the documentation.

### DIFF
--- a/doc/i18n.md
+++ b/doc/i18n.md
@@ -23,6 +23,10 @@
       - [Formatting Texts](#formatting-texts)
       - [TRANSLATORS Comments](#translators-comments)
       - [Missing Translations](#missing-translations)
+      - [Testing Language](#testing-language)
+      - [Building POT File](#building-pot-file)
+      - [Cockpit Details](#cockpit-details)
+      - [Development Server](#development-server)
 
 ---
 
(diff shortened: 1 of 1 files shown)


doc/i18n.md
```diff
@@ -391,3 +399,52 @@ even work, no crash. But there are still translation problems with them.
 - Use the translation functions only with string literals without any operators,
   texts like ~~`_("foo" + "bar")`~~ will not be extracted correctly to the POT
   file.
+
+In both cases the strings will not be extracted to the POT file.
+
+#### Testing Language
+
+The Agama web interface supports special testing `xx` language. That language
+needs to be configured manually by adding the `?lang=xx` query string to the
```
mvidner, line 408:
```suggestion
needs to be selected manually by adding the `?lang=xx` query string to the
```

doc/i18n.md
```diff
@@ -23,6 +23,10 @@
       - [Formatting Texts](#formatting-texts)
       - [TRANSLATORS Comments](#translators-comments)
       - [Missing Translations](#missing-translations)
+      - [Testing Language](#testing-language)
```
mvidner, line 26: Are you sure these are all 4th level under "Marking things for translation"? I would expect them one level up, under Technical details.

mvidner (approved):

doc/i18n.md
lslezak: Ah, right, I overlooked that...

PR #750, opened 2023-09-13
Created PasswordInput component with visibility button (eye icon)

### LINKED ISSUE #242: Consider iOS like entering of (wifi) password (last entered letter is always shown)
Hello team,

seems like users would like to see what key did just hit and simply don't hope for the best.
https://youtu.be/HejPzBYSExk?t=1523 

I personally have to admit that seeing the letter that I just typed makes me more confident and reduces amount of typos and reduces frustration while typing complex passwords and various devices with various keyboards that we nowadays use.

Unfortunatelly screenshots in iOS remove passwords / stars / dots entirely so here is my artistic take on it :-)
Typing a password "Password" would would look like this from user perspective 
(just imagine it's refreshing).

Password: P
Password: *a
Password: **s
Password: ***s
Password: ****w
Password: *****o
Password: ******r
Password: *******d

### LINKED ISSUE #760: `PasswordAndConfirmationInput` leftover
While reviewing #750, I realized that `PasswordAndConfirmationInput` has a leftover _split_ prop.

Initially it was there to allow choosing between a vertical or horizontal layout but actually it was never used. In any case, we should go for a _direction_ prop instead if we really need something similar the future.

For now, I'd just drop both, the prop and the wrapping div.

https://github.com/openSUSE/agama/blob/2d4284241bbd3b0960df06f2a442f9d22dc6e74a/web/src/components/core/PasswordAndConfirmationInput.jsx#L29-L57

Just a good-first-issue or something to commit in #750, #759, or later.

balsa-asanovic:
## Problem

When typing the Wi-Fi password user does not have a way to see what was actually typed in, since the password is hidden by default.

fixes #242 

## Solution

A new component, PasswordInput, has been created. It includes both a text input field and a visibility button with an eye icon.

When the visibility button is clicked, it toggles the text input's form type from 'password' to 'text,' making the characters visible to the user.

This component is now utilized wherever the input type 'password' was previously employed.

As of the time of writing, this applies to the following components:
> 
    src/components/storage/iscsi/AuthFields.jsx
    src/components/questions/LuksActivationQuestion.jsx
    src/components/network/WifiConnectionForm.jsx
    src/components/core/PasswordAndConfirmationInput.jsx

Also this PR now fixes #760 after the issue was raised while making reviews here.

## Testing

Added unit test for the new component in file PasswordInput.test.jsx.

Also tested manually.

## Screenshots

Password input in Users menu
image
image

Wi-Fi password input
image

iSCSI storage password input
image

Luks activation question password input
image


dgdavid (changes requested):
Hi Balsa!

To begin with, thanks a lot for this new contribution! Clearly, it resolves #242.

There are, however, a few changes I would like to suggest:

* Use the "eye icon pattern"

  I.e., to put an eye icon close to the password input for revealing or hiding its content. Similar to the Cockpit's login page implements it.

  An example of how to do it with PatternFly is [here](http://v4-archive.patternfly.org/v4/components/input-group#with-popover) and you can use the ["Visibility" and "Visibility off" icons](https://fonts.google.com/icons?icon.query=eye) provided by material symbols importing them in the [Icon component](https://github.com/openSUSE/agama/blob/5b3172e92c501654ca0476c3fc1f1e95bbf89faf/web/src/components/layout/Icon.jsx).

* Opting for a toggle action rather than a "momentary" one. 

  I.e., make the password input visible (or not) upon clicking.

Hope you don't mind, but I think it's a well-known, slightly better approach.

balsa-asanovic:
> Hope you don't mind, but I think it's a well-known, slightly better approach.

Hey David,

Don't mind at all, in fact thank you for taking the time to point me in the right direction.

I've made some commits in accordance with your comments, so check it out and let me know if there is anything more I should do regarding this.

I noticed something that bothered me a bit but wasn't sure how to fix it—the horizontal alignment of the eye icon within the button doesn't seem quite centered; it's slightly shifted towards the top. It might not be a big deal, but I thought I'd point it out.

Testing pics:

image

image

dgdavid (commented):
> Don't mind at all, in fact thank you for taking the time to point me in the right direction.

> I've made some commits in accordance with your comments, so check it out and let me know if there is anything more I should do regarding this.

Wow! That was really fast! Thank you!

I've added a few comments more. Once you address them, we can continue working to go a a bit further :innocent:  is you're willing to do so :wink: 
> I noticed something that bothered me a bit but wasn't sure how to fix it—the horizontal alignment of the eye icon within the button doesn't seem quite centered; it's slightly shifted towards the top. It might not be a big deal, but I thought I'd point it out.

The quick fix that comes to my mind now is to add a "display: flex" to the `span.pf-c-button__icon` wrapping the icon. Just add below lines to the end of _src/assets/styles/patternfly-overrides.scss_ file

```css

button#password_visibility span.pf-c-button__icon {
  display: flex
}
```

We can go for a better solution later, once the migration to PatternFly 5 is done.

balsa-asanovic:
Hey David,

I've went through your comments, I think I covered them all.

> Nitpicking: what about using an internal component instead?

Here, I chose the second option, without the internal component.
No particular reason, I've just went with the option which had a line of code less :smile: 
If you think that using internal component is a better choice, feel free to mention it and I'll update code accordingly.

> Once you address them, we can continue working to go a a bit further 😇 is you're willing to do so 😉

Certainly, keep 'em coming. :smile:

dgdavid:
> Hey David,
> 
> I've went through your comments, I think I covered them all.
> 
> > Nitpicking: what about using an internal component instead?
> 
> Here, I chose the second option, without the internal component. No particular reason, I've just went with the option which had a line of code less 😄 If you think that using internal component is a better choice, feel free to mention it and I'll update code accordingly.

No strong opinion from my side. Your choice is ok. Let's save lines of code until they are really needed :wink: 


> 
> > Once you address them, we can continue working to go a a bit further 😇 is you're willing to do so 😉
> 
> Certainly, keep 'em coming. 😄

Great!

So, to be clear: from my point of view, **the PR is ready to be merged as it is now**. However, with very little effort, we can improve it a lot for extending that behavior to all password inputs. The idea is

* Create a _src/components/core/PasswordInput.jsx_ component

  - And move the logic dealing with the password input and its _reveal_ button there.
 
* Add simple tests for that new component at _src/components/core/PasswordInput.test.jsx_ file

  - Note that, although the philosophy of React Testing Library is about [_testing the application components in the way the user would use it_](https://testing-library.com/docs/guiding-principles), here we do not have a way to satisfy it (or at least I do not know how to do it :stuck_out_tongue_winking_eye:). 
  
    Thus, instead of checking that the entered password **is not** in the _screen_ before clicking the "Show Password" we will have to directly check the `type` attribute of the input before and after such an action. 

    Definitely no in the way the user would use it, but looks like there are some caveats with the password input yet :)  See https://github.com/w3c/aria/issues/935 and https://github.com/testing-library/dom-testing-library/issues/567.

    Something like below should work  

    ```jsx

    const passwordInput = screen.getByLabelText("Password");
    const button = screen.getByText("Show Password");

    expect(passwordInput).toHaveAttribute("type", "password");
    await user.click(button);
    expect(passwordInput).toHaveAttribute("type", "text")
    ```
* Replace all password inputs by the new `PasswordInput` component to make the UI more consistent and allow the user to reveal desired password, not only these related to a WiFi network.

  ```bash
  ❯ grep -rl "type=\"password" src/
  src/components/storage/iscsi/AuthFields.jsx
  src/components/questions/LuksActivationQuestion.jsx
  src/components/network/WifiConnectionForm.jsx
  src/components/core/PasswordAndConfirmationInput.jsx
  ```

* Check everything works

Do you accept the challenge?

dgdavid:
> Do you accept the challenge?

Of course, you can ping us at any moment you have doubts or need help. We can even jump into the PR for writing the unit test if you get stuck there.

balsa-asanovic:
> Do you accept the challenge?

Accepted, I'll get on it. Also thanks for the pointers, they certainly help a lot.

balsa-asanovic:
> 

    Check everything works

    Change the PR title

    Update the PR description

    Adapt the changes file

    Do you accept the challenge?

I believe I covered everything :smile: 

Check it out and be sure to notify me if anything needs further polishing.

I didn't actually test the LuksActivationQuestion password input, since I wasn't sure how to reach that in-app.

Also, not sure if I handled component's props in the best way, maybe that'll require some tidying up.

web/src/assets/styles/patternfly-overrides.scss
```diff
@@ -165,3 +165,8 @@ table td > .pf-c-empty-state {
   outline: none;
   box-shadow: 0 0 0 1px var(--focus-color);
 }
+
+// Make icon centered in the visibility button in the Wi-Fi password entry from
```
nobkd, line 169:
```suggestion
// Center icon in the visibility button of password input form fields
```

nobkd:
I'm not sure if this matters, but I just thought of the case where there are multiple `PasswordInput`s on the same page.
Then the `id` (`#password_visibility`) has multiple elements that are referenced.
This probably isn't a problem for CSS, but AFAIK isnt a good practice, as an ID should be unique throughout the whole document.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
> The id global attribute defines an identifier (ID) which must be unique in the whole document. Its purpose is to identify the element when linking (using a fragment identifier), scripting, or styling (with CSS).

It might be better to use a class instead...

dgdavid:
> I'm not sure if this matters, but I just thought of the case where there are multiple `PasswordInput`s on the same page. Then the `id` (`#password_visibility`) has multiple elements that are referenced. This probably isn't a problem for CSS, but AFAIK isnt a good practice, as an ID should be unique throughout the whole document.
> 
> See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
> 
> > The id global attribute defines an identifier (ID) which must be unique in the whole document. Its purpose is to identify the element when linking (using a fragment identifier), scripting, or styling (with CSS).
> 
> It might be better to use a class instead...

Of course, it matters. I'm going to make a suggestion for fixing it but still using an id, which can be useful in some scenarios.

Thank you for the input! And welcome to the Agama project :)

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+describe("PasswordInput Component", () => {
+  it("has initial state for password visibility as false", () => {
+    plainRender(
+      <PasswordInput
+        id="password"
+        name="password"
+        label={_("Password")}
+        fieldId="password"
+        ariaLabel={_("User password")}
+      />
```
dgdavid, line 38: If finally `PF/FormGroups`s are moved back and props simplified as proposed, do not forget to clenaup the examples in the tests too. I.e., 

```suggestion
      <PasswordInput id="password" aria-abel={_("User password")} />
```

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+import React from "react";
+import { screen, fireEvent } from "@testing-library/react";
+import { plainRender } from "~/test-utils";
+import userEvent from "@testing-library/user-event";
+import PasswordInput from "./PasswordInput";
+import { _ } from "~/i18n";
+
+describe("PasswordInput Component", () => {
+  it("has initial state for password visibility as false", () => {
```
dgdavid, line 30: Nitpicking:

```suggestion
  it("renders a password input", () => {
```

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+        label={_("Password")}
+        fieldId="password"
+        ariaLabel={_("User password")}
+      />
+    );
+    const inputField = screen.getByLabelText("Password");
+    expect(inputField).toHaveAttribute("type", "password");
+  });
+
+  it("after user clicks the eye icon type switches from password to text", async () => {
```
dgdavid, line 44: Nitpicking:

```suggestion
  it("allows revealing the password", async () => {
```

Think that the _eye icon_ is not that relevant since... we can change our mind in the future and use another icon

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+        ariaLabel={_("User password")}
+        isDisabled={true}
+      />
+    );
+    const inputField = screen.getByLabelText("Password");
+    const visibilityButton = screen.getByRole("button");
+
+    expect(inputField).toBeDisabled();
+    expect(visibilityButton).toBeDisabled();
+  });
```
dgdavid, line 79: Not sure if worth it since both, `TextInput` and `Button` comes from an external library that it's well tested. @imobachgs, what do you think?

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+        fieldId="password"
+        ariaLabel={_("Password")}
+      />
+    );
+    const inputField = screen.getByLabelText("Password");
+    const visibilityButton = screen.getByLabelText("Password visibility button");
+
+    expect(inputField).toHaveAttribute("aria-label", "Password");
+    expect(visibilityButton).toHaveAttribute("aria-label", "Password visibility button");
+  });
```
dgdavid, line 147: Not relevant if we simplify the `Passwordinput` component

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+        ariaLabel={_("Password")}
+        value={inputValue}
+        onChange={handleChange}
+      />
+    );
+    const inputField = screen.getByLabelText("Password");
+    
+    await fireEvent.change(inputField, { target: { value: "newpassword" }});
+    expect(inputValue).toBe("newpassword");
+  });
```
dgdavid, line 170: Again, I think this is already tested by the components' library. But maybe I'm wrong, @imobachgs could you please double-check it?

dgdavid (changes requested):
@balsa-asanovic 

Once again, thanks a ton!

Believe me, I don't want to bother you with the code-review and suggestions. Far from it. However, I requested a new set of changes :sweat_smile: :sweat_smile: :sweat_smile: 


Please, do not hesitate to ask me for help in making these changes if you start feeling overwhelmed with my suggestions. I fully understand that it could be taking more time than you expected.

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,171 @@
...
+        label={_("Password")}
+        fieldId="password"
+        ariaLabel={_("Password")}
+        value={inputValue}
+        onChange={handleChange}
+      />
+    );
+    const inputField = screen.getByLabelText("Password");
+    
+    await fireEvent.change(inputField, { target: { value: "newpassword" }});
```
dgdavid, line 168: The spell checking is complaining because "newpassword" does not look as a valid word for it. You can use "new-password", "secret" or so, but if this test example is finally deleted the complaint will go as well.

dgdavid:
> I didn't actually test the LuksActivationQuestion password input, since I wasn't sure how to reach that in-app.

Don't worry, we're going to check it.

joseivanlopez:
> > I didn't actually test the LuksActivationQuestion password input, since I wasn't sure how to reach that in-app.

I think the easiest way is to encrypt any device (disk or partition). If you are using a virtual machine, you can simply add a new disk (e.g., /dev/sdb) and then run the following command:

~~~
# cryptsetup -y -v luksFormat /dev/sdb
~~~

After that, restart Agama (or change the product in the product selection page) and the encryption password popup should be shown.

ancorgs:
I have not read the pull request or any of the many comments. So sorry if I'm mentioning something that is already there. Even further, I'm just commenting based on the screenshots (most lazy comment ever).

I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.

dgdavid:
> I have not read the pull request or any of the many comments. So sorry if I'm mentioning something that is already there. Even further, I'm just commenting based on the screenshots (most lazy comment ever).
> 
> I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.

🤔 🤔 Hmmm, good point. Now I have the doubt. I understand your point but, at the same time, I'd like a consistent UI regarding the password input. Let's think a bit about it.

dgdavid:
> > I have not read the pull request or any of the many comments. So sorry if I'm mentioning something that is already there. Even further, I'm just commenting based on the screenshots (most lazy comment ever).
> > I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.
> 
> 🤔 🤔 Hmmm, good point. Now I do not have a clear opinion. I understand your point but, at the same time, I'd like a consistent UI regarding the password input. Let's think a bit about it.

imobachgs:
[..]

> I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.

I see your point, but I am not sure it is exactly the same use case. For instance, if my keyboard is set to a different language, the confirmation would not allow me to see that it is wrong. So I find revealing the password useful.

GNOME uses the same approach:

Screenshot from 2023-09-19 14-03-55

I do not have a strong opinion, but I would not block the PR because of this.

web/src/components/core/PasswordInput.jsx
```diff
@@ -0,0 +1,76 @@
...
+import {
+  Button,
+  FormGroup,
+  InputGroup,
+  TextInput
+} from "@patternfly/react-core";
+import { Icon } from "~/components/layout";
+import { _ } from "~/i18n";
+
+export default function PasswordInput({ id, name, ariaLabel, value, onChange, autoFocus, onBlur, fieldId, label, helperTextInvalid, validated = "default", isDisabled = false }) {
```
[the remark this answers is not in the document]
balsa-asanovic: > Thus, I'd move PF/FormGroups as they were and keep the PasswordInput as a simple as possible: basically a component holding a PF/TextInput and a PF/Button, both wrapped into a PF/InputGroup, with an internal state to know if it should use the type password or text.

Haha, yeah, actually this was how I initially made it and thought of a "better idea" :bulb:  to put FormGroup in as well.
I'll backtrack on this.

balsa-asanovic:
> I have not read the pull request or any of the many comments. So sorry if I'm mentioning something that is already there. Even further, I'm just commenting based on the screenshots (most lazy comment ever).

> I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.

Indeed, good point, although @imobachgs makes a good point here as well, that it is still useful:
>I see your point, but I am not sure it is exactly the same use case. For instance, if my keyboard is set to a different language, the confirmation would not allow me to see that it is wrong. So I find revealing the password useful.

>GNOME uses the same approach:

So, should I continue working in this direction or do we reconsider? :brain:

dgdavid:
> [..] For instance, if my keyboard is set to a different language, the confirmation would not allow me to see that it is wrong. So I find revealing the password useful.

Better explained than my previous comment :rofl: 

 
> I do not have a strong opinion, but I would not block the PR because of this.

Me neither, but my suggestion by now is to use the new `PasswordInput` component in all _password inputs_, regardless of whether there is confirmation or not. We can roll back the `PasswordAndConfirmationInut` later if needed; i.e., if we receive more feedback about it weirdness or inconvenience.

dgdavid:
> David made a same point here:
> 
> > Moreover, although the confirmation password let you know you have typed the same keystrokes in both fields... maybe you want to see that you typed what you wanted since perhaps you are not using the desired keyboard layout by mistake.
> 
> Sorry, for missing it 😄

Hehe, don't worry. Anyway, @imobachgs explained it better... and using less words :smiling_face_with_tear: 

BTW, I'm going to hide no longer relevant comments.

imobachgs:
[..]

> > GNOME uses the same approach:
> 
> So, should I continue working in this direction or do we reconsider? 🧠

Yes, please, let's keep it as it is by now. We can reconsider later, but not as part of this PBI.

Thanks!

joseivanlopez:
> I'm just not sure if using this widget makes sense in the first screenshot. The one with "password" and "password confirmation". It makes a lot of sense in all the others (iSCSI and WiFi), in which you are providing a previously set password. But I find combining a "password confirmation" with a password reveal to be weird.

Just my 2 cents about this topic. The "password confirmation" ensures you have typed the same in both inputs, and the "visibility icon" allows you to check you have typed what you really wanted. For me they are complementary. And IMHO, the "visibility icon" is even more important than the confirmation input. Reading the password is the safest way to check that I have really entered it correctly . In fact, having a confirmation is usually useless for me. Many times I write the password correctly but the confirmation wrongly. So having only the password input and the option to read it would be even faster and more secure for me.

It is only a reflection. I am not suggesting to do this ;-).

web/src/components/core/PasswordInput.test.jsx
dgdavid: Thinking it twice, as `PasswordInput` is like _an internal library component_ it's ok testing it behaves as we expect. But I'd do it following the React Testing Library principles. How? By using a fake component in the test as a consumer of `PasswordInput`.

We've few examples already, but now I only see the [TestingPopup](https://github.com/openSUSE/agama/blob/2d4284241bbd3b0960df06f2a442f9d22dc6e74a/web/src/components/core/Popup.test.jsx#L33) one.

I'm in rush now, I will update the comment later with more examples.
...

balsa-asanovic:
> Believe me, I don't want to bother you with the code-review and suggestions. Far from it. However, I requested a new set of changes 😅 😅 😅

Haha, why not, glad to clean up :smile:

I think I covered all suggested changes.
Regarding the unit tests where you believed they are probably tested as a part of components' library, I removed them.

Also, I addressed #760 while working on PasswordConfirmation component, so split prop is gone along with its wrapper.

Additionally, with the help of @joseivanlopez comment

> 
    I think the easiest way is to encrypt any device (disk or partition). If you are using a virtual machine, you can simply add a new disk (e.g., /dev/sdb) and then run the following command:

    cryptsetup -y -v luksFormat /dev/sdb

    After that, restart Agama (or change the product in the product selection page) and the encryption password popup should be shown.

    Note: if you either skip or introduce the correct password, then the dialog is not shown anymore (you have to restart Agama to see it again).

I succeeded in reaching Luks Activation Question component and testing the password input.
Attached screenshots for this in PR description.

web/src/assets/styles/patternfly-overrides.scss
```diff
@@ -165,3 +165,8 @@ table td > .pf-c-empty-state {
   outline: none;
   box-shadow: 0 0 0 1px var(--focus-color);
 }
+
+// Center icon in the visibility button of password input form fields
+button#password_visibility span.pf-c-button__icon {
```
nobkd, line 170: This selector won't work anymore, since the `id` is now changeable and not static `#password_visibility`. I still link, that a `class` on the component and in this selector is the easiest way to fix this.

dgdavid: You're right, the selector as it is won't work. Good catch 😉 Probably,  a class will be needed too,as you mention. A `password-toggler` or better suggestion if you have something in mind.

But keeping the `id` attribute in the button. I'm not a big fan of adding `id`s if they are not strictly needed, but somehow we've commited internally on having form elements reachable by id.

Thank you for taking care and reviewing the PR 💯

balsa-asanovic:
> This selector won't work anymore, since the id is now changeable and not static #password_visibility.
I still think, that the use of a class on the component and in this selector is the easiest way to fix this. :)

Indeed, you are right.
I've just updated the component with className and adjusted CSS override accordingly.

> Thank you for taking care and reviewing the PR 💯

A big thank you from me as well!

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) [2022] SUSE LLC
```
dgdavid, line 2: NP

```suggestion
 * Copyright (c) [2023] SUSE LLC
```

web/src/components/core/PasswordInput.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+    );
+
+    const passwordInput = screen.getByLabelText("User password");
+    const button = screen.getByRole("button");
+
+    expect(passwordInput).toHaveAttribute("type", "password");
+    await userEvent.click(button);
+    expect(passwordInput).toHaveAttribute("type", "text");
+  });
+
```
dgdavid, line 59: I've updated my comment at https://github.com/openSUSE/agama/pull/750#discussion_r1330293636, just in case you want to re-introduce the example testing that the PasswordInput honors given #onChange callback.

balsa-asanovic: Sure, why not, I'll re-introduce it according to your comment.

balsa-asanovic:
> I've updated my comment at https://github.com/openSUSE/agama/pull/750#discussion_r1330293636, just in case you want to re-introduce the example testing that the PasswordInput honors given #onChange callback.

Added the test back.

> Something like below (sorry, not much inspiration at this moment 😉)

Inspirational enough for me 😄

dgdavid (approved):
@balsa-asanovic 

From my point of view, it looks pretty good now. **Thanks a lot for your GREAT job here**

I've added a few minor suggestions, but the PR is ok as it is.

Also thanks to @nobkd for reviewing and contributing to make the PR even better.

balsa-asanovic:
> Thanks a lot for such a GREAT job

Don't mention it, anytime.

Added documentation comments and extracted the id prop.

PR #757, opened 2023-09-18
Tool for collecting agama logs

mchf:
## Problem

Something like save_y2logs was missed by bug reporters

## Solution

save_y2logs inspired script
Copies some well known log files and output of commands.

~~TODO: replace copy command by something what handles even directories (eg. fs_extra::copy_items)~~
~~TODO: check user's permissions on start~~

## Testing

- Tested manually

## Screenshots

command screen output in noisy mode 
agama-logs_noisy


imobachgs (changes requested):
Thanks for the PR! I have submitted a set of suggestions.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    // definition of destination as path to a file
+    fn to(&self) -> PathBuf;
+
+    // performs whatever is needed to store logs from "from" at "to" path
+    fn store(&self) -> bool;
+}
+
+impl LogItem for LogPath {
+    fn from(&self) -> &'static str {
+        return self.src_path.clone();
```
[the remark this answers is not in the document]
imobachgs: Or return a &str and let the caller to decide.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
+extern crate fs_extra;
+extern crate tempdir;
+
+use fs_extra::copy_items;
```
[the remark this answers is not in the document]
mchf: it doesn't copy directories

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+impl LogItem for LogPath {
+    fn from(&self) -> &'static str {
+        return self.src_path.clone();
+    }
+
+    fn to(&self) -> PathBuf {
+        // remove leading '/' if any from the path (reason see later)
+        let r_path = Path::new(self.src_path).strip_prefix("/").unwrap();
```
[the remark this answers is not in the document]
mchf: well I'm not. 

1) I plan to add an option for dynamically adding paths. Ok this is more less personal toy to improve my skills when practicing using rust script for implementing cmdline tools.
2)  Moreover if at some point of time someone adds some strange path this could solve potentially weird issue for free.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+// A wrapper around println which shows (or not) output depending on noisy boolean variable
+macro_rules! show
+{
+    ($n:expr, $($arg:tt)*) => { if($n) { print!($($arg)*) } }
+}
+
+// Struct for log represented by a file
+struct LogPath {
+    // log source
+    src_path: &'static str,
```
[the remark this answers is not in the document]
mchf: There was a problem with strings with "not living long enoug" (or something like that), so I switched to ```str```. I can give it another try since the code evolved a bit since that time.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn from(&self) -> &'static str {
+        return self.src_path.clone();
+    }
+
+    fn to(&self) -> PathBuf {
+        // remove leading '/' if any from the path (reason see later)
+        let r_path = Path::new(self.src_path).strip_prefix("/").unwrap();
+
+        // here is the reason, join overwrites the content if the joined path is absolute
+        return self.dst_path.join(r_path);
```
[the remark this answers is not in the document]
mchf: ah conventions ... ok, tried to replace all occurences.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    dst_path: PathBuf,
+}
+
+// Struct for log created on demmand by a command
+struct LogCmd {
+    // command which stdout / stderr is logged
+    cmd: &'static str,
+
+    // place where to collect logs
+    dst_path: PathBuf,
```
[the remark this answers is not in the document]
mchf: same as above ;-) will check again, as above

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+trait LogItem {
+    // definition of log source
+    fn from(&self) -> &'static str;
+
+    // definition of destination as path to a file
+    fn to(&self) -> PathBuf;
+
+    // performs whatever is needed to store logs from "from" at "to" path
+    fn store(&self) -> bool;
```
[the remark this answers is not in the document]
mchf: yeah ... that's chapter I didn't read good enough yet ... ;-) will try to fix it.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    "/linuxrc.config",
+];
+
+const DEFAULT_RESULT: &str = "/tmp/agama_logs";
+const DEFAULT_NOISY: bool = true;
+const DEFAULT_COMPRESSION: (&str, &str) = ("bzip2", "tar.bz2");
+const DEFAULT_TMP_DIR: &str = "agama-logs";
+
+// A wrapper around println which shows (or not) output depending on noisy boolean variable
+macro_rules! showln
```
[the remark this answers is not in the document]
mchf: eh ... why not? println which is wrapped inside is a macro too. So implementing a function which wraps a macro sounds ... weird to me.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn store(&self) -> bool {
+        let mut res = false;
+
+        // for now keep directory structure close to the original
+        // e.g. what was in /etc will be in /<tmp dir>/etc/
+        if fs::create_dir_all(self.to().parent().unwrap()).is_ok() {
+            let options = CopyOptions::new();
+            res = copy_items(
+                &[self.src_path],
+                self.to().parent().unwrap().as_os_str().to_str().unwrap(),
```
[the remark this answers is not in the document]
mchf: I've reduced that nice chain of calls

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
imobachgs: I do not think it is weird. The question is: why do you need a macro? Why a macro offers in this use case that a function does not?

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
imobachgs: Yes, you are right. I wanted to say that you are safe while you control the values :-)

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: well i don't know exactly how it is in rust but i guess it is similar to C at least a bit.

So macros are replaced during compilation by its body. Functions are runtime thing which uses all that handling of params and returns using the stack and so on.  So in my POV macros are something "simplier" than functions and functional envelope around a macro is just adding unnecessary (internal) complexity.

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: ok this one should be modified now.

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: ok, I didn't get it ... obviously ;-)

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: ok, even in case of copy_items I work with the Result, so behavior should be more pretty (fingers crossed ;-)

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: As it seems this one was turned into ```String``` successfully

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: After some reading, my understanding is:
* ```String``` is owned ```str``` pointer / borrowed and
* ```Path``` is pointer / borrowed and ```PathBuf``` owned

since I've already moved ```cmd``` (in case of this struct) into String, both ```cmd``` and ```dst_path``` are covered with owned types. So, should be fine now, shouldn't it?

rust/Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
   "agama-derive",
   "agama-lib",
   "agama-locale-data",
+  "agama-logs",
```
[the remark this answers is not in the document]
mchf: I'll deal with this one once everything else is accepted. It is way easier to deal with standalone script for me than fighting with agama-cli now.

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
imobachgs: I'm afraid I have to disagree. In Rust, you should use a macro when a function falls short. Otherwise, you might write most of your code using macros, which would be hard to read and maintain.

IMHO the only justification to write a macro here is that you want to use a variable number of arguments. That's the reason for `println!` to be a macro: Rust does not support variadic functions. However, I would most probably use an already existing macro for that:

```rust
showln(true, format!("Storing result {}", true))
```

About being `simpler`, I find:

```rust
macro_rules! showln
{
    ($n:expr, $($arg:tt)*) => { if($n) { println!($($arg)*) } }
}
```

harder to read than:

```rust
fn showln(show: boolean, text: &str) {
  if (show) { println!(text) }
}
```

However, I am not going to block the PR because of this. But, please, document the macro.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    ($n:expr, $($arg:tt)*) => { if($n) { print!($($arg)*) } }
+}
+
+// Struct for log represented by a file
+struct LogPath {
+    // log source
+    src_path: &'static str,
+
+    // directory where to collect logs
+    dst_path: PathBuf,
```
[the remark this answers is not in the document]
imobachgs: I forgot to drop the link to the [newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html). As said, just a matter of taste, but it is interesting to read about it.

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: honestly, I don't care. One of my motivation was to try implement macro(s) in rust  as they are way too more complex then in C. So, I did and my soul is calm now. So, I can re-write it into function ... that should be simple enough ;-)

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+        log_sources.push(Box::new(LogCmd {
+            cmd: cmd.to_string(),
+            dst_path: tmp_dir.path().to_path_buf(),
+        }));
+    }
+
+    // 4. store it
+    showln!(true, "Storing result in: \"{}\"", result);
+
+    for src in log_sources.iter() {
```
imobachgs, line 190: np: For consistency with the rest of `for` loops you could do `for src in log_sources` too (consuming the collection). If you want to keep it (not needed in this case), you could do `for src in &log_sources`.

imobachgs (changes requested):
[This comment](https://github.com/openSUSE/agama/pull/757/files#r1330157293) was not addressed (or discussed). Otherwise, I only have small things.

After moving this code to a proper `agama` comment, I would love to see the `DEFAULT_NOISY` gone.

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,221 @@
+extern crate fs_extra;
+extern crate tempdir;
```
imobachgs, line 2: Please, remove the `extern crate` lines. That's something from the past :-)

rust/agama-logs/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+    let tmp_dir = TempDir::new(DEFAULT_TMP_DIR)?;
+
+    // 2. collect existing / requested paths which should already exist
+    showln!(noisy, "\t- proceeding well known paths");
+    for path in paths {
+        // assumption: path is full path
+        if Path::new(path).try_exists().is_ok() {
+            log_sources.push(Box::new(LogPath {
+                src_path: path.to_string(),
+                dst_path: tmp_dir.path().to_path_buf(),
```
imobachgs, line 173: You can replace `to_path_buf()` with `to_owned()` which express your intent better. If you prefer, you can have a constructor that takes care of the conversion details:

```
impl LogPath {
  fn new(src: &str, dst: &Path) -> Self { 
    Self { src_path: src.to_string(), dst_path: dst.to_owned() }
  }
}
```

And you can use it as:

```rust
  log_sources.push(Box::new(LogPath::new(path, tmp_dir.path()))
```

rust/agama-logs/src/main.rs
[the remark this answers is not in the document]
mchf: Well, no strong reason. It was just a matter of first view when i started designing that (I originally didn't have ```store``` function, but ```from``` and ```to```) which somehow evolved during the time.

Thanks for the link. So far I've read 1/4 - 1/3 of the book and not necessarily in consistent pieces ;-) So, I've a lot reading ahead ... but i'm getting rust preferred over ruby - it is getting way closer to my low level heard ;-)  So I'll happily read it ;-)

rust/agama-logs/src/main.rs
mchf: done

rust/agama-logs/src/main.rs
mchf: I believe that I read somewhere that it still is good to use ... but i don't have the article at hand and it easily could be five years old one ;-) so removed.

rust/agama-logs/src/main.rs
mchf: nice ... some more reading to the list ;-)

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+impl LogPath {
+    fn new(src: &str, dst: &Path) -> Self {
+        Self {
+            src_path: src.to_string(),
+            dst_path: dst.to_owned(),
+        }
+    }
+}
+
+// Struct for log created on demmand by a command
```
imobachgs, line 98:
```suggestion
// Struct for log created on demand by a command
```

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
+use clap::Subcommand;
+use fs_extra::copy_items;
+use fs_extra::dir::CopyOptions;
+use nix::unistd::Uid;
+use std::fs;
+use std::fs::File;
+use std::io;
+use std::io::Error;
```
imobachgs, line 8: Personal taste, but I think it is better to use this error as `io::Error` instead (so you do not need to import it here). It makes possible to differentiate at first sight between `io::Error`, `format::Error` and [the Error trait](https://doc.rust-lang.org/std/error/trait.Error.html) (that I usually import as `Error`).

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+        self.dst_path.join(r_path)
+    }
+
+    fn store(&self) -> Result<(), Error> {
+        // for now keep directory structure close to the original
+        // e.g. what was in /etc will be in /<tmp dir>/etc/
+        fs::create_dir_all(self.to().parent().unwrap())?;
+
+        let options = CopyOptions::new();
+        // fs_extra's own Error doesn't implement From trait so ? operator is unusable
```
imobachgs, line 146: `fs_extra` error does not implement the conversion to `io::Error`. However, implements the [Error trait](https://doc.rust-lang.org/std/error/trait.Error.html), so you can do something like this:

```rust
    fn store(&self) -> Result<(), Box<dyn std::error::Error>> {
        // for now keep directory structure close to the original
        // e.g. what was in /etc will be in /<tmp dir>/etc/
        fs::create_dir_all(self.to().parent().unwrap())?;

        let options = CopyOptions::new();
        copy_items(
            &[self.src_path.as_str()],
            self.to().parent().unwrap(),
            &options,
        )?;
        Ok(())
    }
```
Of course, you can import the error trait as `use std::error::Error` if you wish and you need to adapt signature of the `store` function in the `LogItem` trait.

Another option might be to have your own error type and write the conversion from `fs_extra` error type. You can use [thiserror](https://crates.io/crates/thiserror) to simplify the process.

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+        match copy_items(
+            &[self.src_path.as_str()],
+            self.to().parent().unwrap(),
```
imobachgs, line 149: You are doing exactly the same in line 143. What about writing to a variable with a nice name?

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+        self.dst_path.as_path().join(format!("{}", self.cmd))
+    }
+
+    fn store(&self) -> Result<(), Error> {
+        let cmd_parts = self.cmd.split_whitespace().collect::<Vec<&str>>();
+        let file_path = self.to();
+        let output = Command::new(cmd_parts[0])
+            .args(cmd_parts[1..].iter())
+            .output()
+            .expect("Failed to run the command");
```
imobachgs, line 176: Which command? The user would get a panic but not a proper error. You might want to:

1. Return a proper error (instead of panicking). You coud use your own error type encoding more useful information
2. Or take advantage of `anyhow`'s `context` mechanism. We have other examples in the code.

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    // 0. preparation, e.g. in later features some log commands can be added / excluded per users request or
+    let commands = DEFAULT_COMMANDS;
+    let paths = DEFAULT_PATHS;
+    let result = format!("{}.{}", DEFAULT_RESULT, DEFAULT_COMPRESSION.1);
+    let noisy = verbose;
+    let compression = DEFAULT_COMPRESSION.0;
+    let mut log_sources: Vec<Box<dyn LogItem>> = Vec::new();
+
+    showln(noisy, "Collecting Agama logs:");
```
imobachgs, line 200: What about replacing `noisy` with `verbose`? Just for consistency reasons.

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    for log in log_sources.iter() {
+        show(
+            noisy,
+            format!("\t- storing: \"{}\" ... ", log.from()).as_str(),
+        );
+
+        // for now keep directory structure close to the original
+        // e.g. what was in /etc will be in /<tmp dir>/etc/
+        let res = match fs::create_dir_all(log.to().parent().unwrap()) {
```
imobachgs, line 232: What about the following?

```rust
        fs::create_dir_all(log.to().parent().unwrap())?;
        let res = match log.store() {
            Ok(_p) => "[Ok]",
            Err(_e) => "[Failed]",
        };
```

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+    let commands = DEFAULT_COMMANDS;
+    let paths = DEFAULT_PATHS;
+    let result = format!("{}.{}", DEFAULT_RESULT, DEFAULT_COMPRESSION.1);
+    let noisy = verbose;
+    let compression = DEFAULT_COMPRESSION.0;
+    let mut log_sources: Vec<Box<dyn LogItem>> = Vec::new();
+
+    showln(noisy, "Collecting Agama logs:");
+
+    // 1. create temporary directory where to collect all files (similar to what old save_y2logs
```
imobachgs, line 202: Perhaps splitting this function into smaller ones would make the code easier to follow. This comments would belong to those functions.

rust/agama-cli/src/logs.rs
```diff
@@ -0,0 +1,257 @@
...
+        result,
+        compression,
+        tmp_dir.path().display()
+    );
+    let cmd_parts = compress_cmd.split_whitespace().collect::<Vec<&str>>();
+
+    Command::new(cmd_parts[0])
+        .args(cmd_parts[1..].iter())
+        .status()
+        .expect("failed creating the archive");
```
imobachgs, line 254: Panicking :-(

imobachgs (commented):
Thanks! I tried it manually and it works quite well. However, we need to improve the error handling.

I could approve it by now so we get the command ready for the next testing release, but we should address the found issues as soon as possible.

What do you think?

rust/agama-cli/src/logs.rs
mchf: done

rust/agama-cli/src/logs.rs
mchf: done

rust/agama-cli/src/logs.rs
mchf: yeah that was a plan once i start working on the ```list``` subcommand.

rust/agama-cli/src/logs.rs
mchf: consequence of learning ;-) will modify, I see into it more now ;-)

rust/agama-cli/src/logs.rs
mchf: well if written this way it more less masks what is returned by copy_items, doesn't it?

So, I've tried to do the conversion simply by the match rule.

Anyway when ```?``` operator is used on ```copy_items``` it complains "trait From<fs_extra::error::Error>  is not implemented for std::io::Error" so this snippet doesn't work.

rust/agama-cli/src/logs.rs
mchf: it is still too much new for me to have strong opinion for such details ... however using io::Error sounds like it provides more info, so will change it.

rust/agama-cli/src/logs.rs
mchf: I'm still not sure if i get it correctly. Does it mean that if ```?``` operator's result is an Err then it returns immediately? (i don't want to call store when create_dir_all failes)

imobachgs: Yes, the `?` operator returns immediately. Additionally, it tries to convert the error type to the one returned by the function.

rust/agama-cli/src/logs.rs
imobachgs: Well, it does not work because you need to update the `store` signature in the `LogItem` trait. After that, it will work.

About masking the error by `copy_items`, you can have your own error type that keeps the original error if you wish. Actually, [anyhow](https://crates.io/crates/anyhow) allows you to keep the errors backtrace easily.

About error handling, you might be interested in [this article](https://www.lpalmieri.com/posts/error-handling-rust/) by [Luca Palmieri](https://www.lpalmieri.com/). Another link for your reading list 😄.

rust/agama-cli/src/logs.rs
mchf: first step done ;-)

rust/agama-cli/src/logs.rs
mchf: modified in a 1 way (as 2 is unknown to me yet ;-)

rust/agama-cli/src/logs.rs
mchf: but if it exists whole function, then it doesn't try to store other logs (the create dir is log specific in this case) ... it might but even might not fail for the others ....

imobachgs (commented):
Please, add an entry to the changes file (referencing this PR) and let's merge it :-) Thanks!

imobachgs (approved):

PR #758, opened 2023-09-18
WIP: Add Ipv6 support to BaseConnection

jcronenberg:
## Problem

Ipv6 is missing from Agama BaseConnection

## Solution

This PR aims to implement what is already there for ipv4 in BaseConnection also for ipv6.

## Testing

- *Extended existing unit test*
- *Tested manually*

## Notes

To not duplicate all the code I used the duplicate crate which makes the process of creating macros easier. The macros could also of course be written into the code directly but like I said this makes it more convenient and also has a cleaner syntax IMO. However it of course introduces a new dependency, let me know if this is a problem.
I also tried my hand at using generics to solve the code duplication problem but ran into some issues where honestly my rust knowledge wasn't enough, so if this path is desired I would need some help.
Let me know what you think :slightly_smiling_face:

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -35,6 +35,7 @@ dependencies = [
  "agama-locale-data",
  "anyhow",
  "async-std",
+ "duplicate",
  "futures",
  "log",
  "serde",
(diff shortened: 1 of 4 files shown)


jcronenberg:
I think the approach of #761 is better. So let's continue with that.

PR #763, opened 2023-09-22
Configure system volume group devices

joseivanlopez:
## Problem

Agama web UI has no option for selecting which disks to use for the system volume group. The volume group is always created over the installation device. 

## Solution

Add a config button along to the LVM switcher to configure the devices for the system volume group.

NOTE: probably the selection of the installation disk and the configuration of LVM will be merged in a single popup, see [create issue].

## Testing

* Added new unit tests
* Tested manually

## Screenshots

<details>
<summary>Show/hide</summary>

localhost_8080_ (21)

localhost_8080_ (22)

localhost_8080_ (23)

</details>


web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -40,19 +41,24 @@ import { noop } from "~/utils";
  */
 
 /**
- * Form for selecting the installation device
+ * Form for selecting the installation device.
```
dgdavid, line 44: NP: no strong opinion about it, but we can save one character by omitting the dot at the end of sentence in documentation. Especially when documenting paramos.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -167,55 +173,220 @@ file systems will be created on the selected device.");
...
+    const bootDevice = devices.find(d => d.name === settings.bootDevice);
+
+    return <DeviceList devices={[bootDevice]} />;
+  };
+
+  return (
+    <Form id={id} onSubmit={onSubmit}>
+      <div className="split">
+        <span>{_("Devices for creating the volume group")}</span>
+        <ToggleGroup aria-label="Volume group devices" isCompact>
```
dgdavid, line 238: NP: I'm not sure about the use of the aria-label here. It's not required for that component and we already have an text before it that I guess is enough

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -167,55 +173,220 @@ file systems will be created on the selected device.");
...
  * @callback onValidateFn
  * @param {boolean} valid
  */
-const EncryptionPasswordForm = ({ id, password: passwordProp, onSubmit, onValidate }) => {
+const EncryptionPasswordForm = ({
+  id,
+  password: passwordProp,
+  onSubmit = noop,
+  onValidate = noop
+}) => {
```
dgdavid, line 389: 👍

dgdavid (approved):
Despite the PR still in draft mode and I've somehow involved, I did a quick review and it LGTM. Ofncourde, there is a lot of room for improvements, but I think it's good enough for gathering feedback if we want to release it as it is and send the improvements later. 

I'm taking into account that it has been manually tested too.

web/src/components/storage/ProposalSettingsSection.jsx
dgdavid: Now you changed my mind too :P

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -167,55 +173,220 @@ file systems will be created on the selected device.");
...
+ * @param {string[]} devices - Name of the selected devices.
+ *
+ * @callback onValidateFn
+ * @param {boolean} valid
+ */
+const LVMSettingsForm = ({
+  id,
+  settings,
+  devices = [],
+  onSubmit: onSubmitProp = noop,
```
dgdavid, line 196: NP: I see that the rename is because there is an internal `onSubmit` function. I'd just use `onSubmission` or **`handleSubmit`** for the internal one just for avoiding the rename. A personal taste, of course.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -167,55 +173,220 @@ file systems will be created on the selected device.");
...
+  onValidate = noop
+}) => {
+  const [vgDevices, setVgDevices] = useState(settings.systemVGDevices);
+  const [isBootDeviceSelected, setIsBootDeviceSelected] = useState(settings.systemVGDevices.length === 0);
```
dgdavid, line 200: NP: sometimes I think about if `is` is really needed. For example, what about

```suggestion
  const [bootDeviceSelected, setBootDeviceSelected] = useState(settings.systemVGDevices.length === 0);
```

Of course, `is` hints you about there should be a boolean value instead of _the boot device selected_... but I wanted to share my mess :stuck_out_tongue_closed_eyes:

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -167,55 +173,220 @@ file systems will be created on the selected device.");
...
  * @param {boolean} lvm
  */
-const LVMField = ({ selected: selectedProp, isLoading, onChange }) => {
-  const [selected, setSelected] = useState(selectedProp);
+const LVMField = ({
+  settings,
+  devices = [],
+  isChecked: isCheckedProp = false,
+  isLoading = false,
+  onChange: onChangeProp = noop
```
dgdavid, line 288: NP: Same than above here. I'd use `handleOnChange` internally instead of renaming the prop here.

However I have no idea now to avoid the `isChecked` rename.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -246,16 +417,21 @@ const EncryptionPasswordForm = ({ id, password: passwordProp, onSubmit, onValida
...
+ * @param {boolean} [props.isLoading=false] - Whether to show the selector as loading
+ * @param {onChangeFn} [props.onChange=noop] - On change callback
  *
  * @callback onChangeFn
  * @param {object} settings
  */
-const EncryptionPasswordField = ({ selected: selectedProp, password: passwordProp, isLoading, onChange }) => {
-  const [selected, setSelected] = useState(selectedProp);
+const EncryptionPasswordField = ({
+  password: passwordProp = "",
```
dgdavid, line 429: NP: You could call it `value`, by mimicking the prop of other TextInputs. Of course, it is `password` internally.

web/src/components/storage/ProposalSettingsSection.test.jsx
```diff
@@ -189,7 +217,9 @@ describe("LVM field", () => {
 
   describe("if LVM is set to true", () => {
     beforeEach(() => {
-      props.settings = { lvm: true };
+      props.availableDevices = [vda, md0, md1];
+      // TODO: what happens if systemDevices is not set or an empty array?
```
dgdavid, line 221: NP: A leftover from me. I put it last night to not forget ask you about it.

web/src/components/storage/device-utils.jsx
```diff
@@ -224,38 +223,64 @@ const ItemContent = ({ device }) => {
...
+ * Component for listing storage devices.
  * @component
  *
  * @param {Object} props
- * @param {StorageDevice} [props.selected] - Default selected device, in any
- * @param {StorageDevice[]=[]} [props.devices] - Devices to show in the selector
- * @param {onSelectFn} [props.onSelect] - Function to be called when a device is selected
+ * @param {StorageDevice[]} props.devices - Devices to show.
+ */
+const DeviceList = ({ devices }) => {
```
dgdavid, line 232: Note: we've DeviceList and DeviceSelector into device-utils file by now. Most probably we will look for a better place once we manage to see how useful can the ListBox be exposed under core namesapce.

dgdavid (approved):
LGTM

I put few comments there, but are just for further discussion. I promise, I have no strong opinion about them at this moment.

joseivanlopez:
> LGTM
> 
> I put few comments there, but are just for further discussion. I promise, I have no strong opinion about them at this moment.

Thanks for your comments! I will keep these comments about renaming open. Let's manage them in a next iteration.

web/src/components/storage/ProposalSettingsSection.jsx
joseivanlopez: Yes, I think we should be consistent. IMHO, using `is` prefix for booleans is a good approach (similar to `?` sufix in ruby).

PR #768, opened 2023-09-25
Require tar and bzip2

imobachgs:
## Problem

`agama logs store` depends on `tar` and `bzip2`.

## Solution

Add both packages to the requires section.


ancorgs (approved):
LGTM

PR #770, opened 2023-09-25
Pattern modify dbus

jreidinger:
## Problem

DBus missing methods to modify patterns selection.

## Solution

Implement it.

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -215,6 +215,17 @@ module Agama
         selected_patterns_changed
       end
 
+      def user_patterns=(ids)
+        @user_patterns.each { |p| Yast::Pkg.ResolvableRemove(p, :pattern) }
+        @user_patterns = ids
+        @user_patterns.each { |p| Yast::Pkg.ResolvableInstall(p, :pattern) }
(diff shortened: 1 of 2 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -190,6 +193,52 @@ def patterns(filtered)
...
+        logger.info "Removing pattern #{res.inspect}"
+        @user_patterns << id
+
+        res = Yast::Pkg.PkgSolve(unused = true)
+        logger.info "Solver run #{res.inspect}"
+        selected_patterns_changed
+      end
+
+      def user_patterns=(ids)
+        @user_patterns.each { |p| Yast::Pkg.ResolvableRemove(p, :pattern) }
```
lslezak, line 219: Rather use the [Pkg.ResolvableNeutral()](https://github.com/yast/yast-pkg-bindings/blob/a790b6b072d6ead55dcda63dc9102ff0fa336d64/src/Resolvable_Install.cc#L404) call here.

`Pkg.ResolvableRemove()`  means "uninstall the resolvable", while `Pkg.ResolvableNeutral()` means "remove all scheduled transactions, do not touch it".

lslezak (approved):

PR #780, opened 2023-09-29
New subcommand for agama logs, refactoring

mchf:
## Problem

Trying to grow agama logs from baby age to teenage.

## Solution

- Implemented agama logs list subcommand
- Internal refactoring
- Changes in target tarball - related to: https://github.com/openSUSE/agama/issues/774
- Commands are not stored in files with names containing spaces anymore. 
- Command generated logs can be stored in files with user defined names

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -327,9 +327,11 @@ fn store(options: LogOptions) -> Result<(), io::Error> {
 
 // Handler for the "agama logs list" subcommand
 fn list(options: LogOptions) {
-    for list in [options.paths, options.commands] {
-        for item in list.iter() {
-            println!("{}", item);
+    for list in [ ("Log paths: ", options.paths), ("Log commands: ", options.commands)] {
+        println!("{}", list.0);
+
+        for item in list.1.iter() {
+            println!("\t{}", item);
         }
 
         println!();
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -29,16 +29,12 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
         LogsCommands::Store { verbose } => {
             // feed internal options structure by what was received from user
             // for now we always use / add defaults if any
-            let mut options = LogOptions::new();
-
-            options.verbose = verbose;
+            let options = LogOptions { verbose: verbose, ..Default::default() };
 
             Ok(store(options)?)
         }
         LogsCommands::List => {
-            let options = LogOptions::new();
-
-            list(options);
+            list(LogOptions::default());
 
             Ok(())
         }
@@ -100,8 +96,8 @@ struct LogOptions {
     verbose: bool,
 }
 
-impl LogOptions {
-    fn new() -> Self {
+impl Default for LogOptions {
+    fn default() -> Self {
         Self {
             paths: DEFAULT_PATHS.iter().map(|p| p.to_string()).collect(),
             commands: DEFAULT_COMMANDS.iter().map(|p| p.to_string()).collect(),
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -255,17 +255,19 @@ fn compress_logs(tmp_dir: &TempDir, result: &String) -> io::Result<()> {
         dir,
     );
     let cmd_parts = compress_cmd.split_whitespace().collect::<Vec<&str>>();
-
-    Command::new(cmd_parts[0])
+    let res = Command::new(cmd_parts[0])
         .args(cmd_parts[1..].iter())
-        .status()
-        .ok_or(
-            Ok(_o) => Ok(()),
-            Err(_e) => Err(io::Error::new(
-                io::ErrorKind::Other,
-                "Cannot create tar archive",
-            )
+        .status()?;
+
+    if res.success() {
+        Ok(())
+    }
+    else {
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            "Cannot create tar archive",
         ))
+    }
 }
 
 // Handler for the "agama logs store" subcommand
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -44,7 +44,7 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     }
 }
 
-const DEFAULT_COMMANDS: [(&str,&str); 3] = [
+const DEFAULT_COMMANDS: [(&str, &str); 3] = [
     ("journalctl -u agama", "agama"),
     ("journalctl -u agama-auto", "agama-auto"),
     ("journalctl --dmesg", "dmesg"),
@@ -103,7 +103,10 @@ impl Default for LogOptions {
     fn default() -> Self {
         Self {
             paths: DEFAULT_PATHS.iter().map(|p| p.to_string()).collect(),
-            commands: DEFAULT_COMMANDS.iter().map(|p| (p.0.to_string(), p.1.to_string())).collect(),
+            commands: DEFAULT_COMMANDS
+                .iter()
+                .map(|p| (p.0.to_string(), p.1.to_string()))
+                .collect(),
             verbose: false,
         }
     }
@@ -243,11 +246,18 @@ fn paths_to_log_sources(paths: &Vec<String>, tmp_dir: &TempDir) -> Vec<Box<dyn L
 }
 
 // Some info can be collected via particular commands only, turn it into log sources
-fn cmds_to_log_sources(commands: &Vec<(String, String)>, tmp_dir: &TempDir) -> Vec<Box<dyn LogItem>> {
+fn cmds_to_log_sources(
+    commands: &Vec<(String, String)>,
+    tmp_dir: &TempDir,
+) -> Vec<Box<dyn LogItem>> {
     let mut log_sources: Vec<Box<dyn LogItem>> = Vec::new();
 
     for cmd in commands.iter() {
-        log_sources.push(Box::new(LogCmd::new(cmd.0.as_str(), cmd.1.as_str(), tmp_dir.path())));
+        log_sources.push(Box::new(LogCmd::new(
+            cmd.0.as_str(),
+            cmd.1.as_str(),
+            tmp_dir.path(),
+        )));
     }
 
     log_sources
@@ -354,7 +364,10 @@ fn store(options: LogOptions) -> Result<(), io::Error> {
 fn list(options: LogOptions) {
     for list in [
         ("Log paths: ", options.paths),
-        ("Log commands: ", options.commands.iter().map(|c| c.0.clone()).collect()),
+        (
+            "Log commands: ",
+            options.commands.iter().map(|c| c.0.clone()).collect(),
+        ),
     ] {
         println!("{}", list.0);
 
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -34,14 +34,14 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
             options.verbose = verbose;
 
             Ok(store(options)?)
-        },
+        }
         LogsCommands::List => {
             let options = LogOptions::new();
 
             list(options);
 
             Ok(())
-        },
+        }
     }
 }
 
@@ -315,8 +315,7 @@ fn store(options: LogOptions) -> Result<(), io::Error> {
 }
 
 // Handler for the "agama logs list" subcommand
-fn list(options: LogOptions)
-{
+fn list(options: LogOptions) {
     for list in [options.paths, options.commands] {
         for item in list.iter() {
             println!("{}", item);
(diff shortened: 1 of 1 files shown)


rust/agama-cli/src/logs.rs
```diff
@@ -22,10 +23,25 @@ pub enum LogsCommands {
...
 
+// main entry point called from agama CLI main loop
 pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     match subcommand {
-        LogsCommands::Store { verbose } => Ok(store(verbose)?),
-        LogsCommands::List => Err(anyhow::anyhow!("Not implemented")),
+        LogsCommands::Store { verbose } => {
+            // feed internal options structure by what was received from user
+            // for now we always use / add defaults if any
+            let mut options = LogOptions::new();
```
imobachgs, line 32: I would suggest to implement [Default](https://doc.rust-lang.org/std/default/trait.Default.html) for `LogOptions` and change this code to:

```rust
let options = LogOptions { verbose, ..Default::default() }
```

And the same below (see the following comment).

rust/agama-cli/src/logs.rs
```diff
@@ -22,10 +23,25 @@ pub enum LogsCommands {
...
+
+            options.verbose = verbose;
+
+            Ok(store(options)?)
+        }
+        LogsCommands::List => {
+            let options = LogOptions::new();
```
imobachgs, line 39: You can change this to (if Default is implemented):

```rust
            list(LogOptions::default());
            Ok(())
```

rust/agama-cli/src/logs.rs
```diff
@@ -76,6 +92,24 @@ fn show(show: bool, text: &str) {
...
 
+// Configurable parameters of the "agama logs" which can be
+// set by user when calling a (sub)command
+struct LogOptions {
+    paths: Vec<String>,
+    commands: Vec<String>,
+    verbose: bool,
+}
+
+impl LogOptions {
```
imobachgs, line 103:
```rust
impl Default for LogOptions {
    fn default() -> Self {
        Self {
            paths: DEFAULT_PATHS.iter().map(ToString::to_string).collect(),
            commands: DEFAULT_COMMANDS.iter().map(ToString::to_string).collect(),
            verbose: false,
        }
    }
}
```

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +218,77 @@ impl LogItem for LogCmd {
...
 }
 
-// compress given directory into a tar archive
+// Compress given directory into a tar archive
 fn compress_logs(tmp_dir: &TempDir, result: &String) -> io::Result<()> {
     let compression = DEFAULT_COMPRESSION.0;
+    let tmp_path = tmp_dir
+        .path()
+        .parent()
+        .unwrap()
```
imobachgs, line 253: It would be nice to get rid of this `unwrap()` :wink:

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +218,77 @@ impl LogItem for LogCmd {
...
+        .as_os_str()
```
imobachgs, line 254: Why do you need to call `as_os_str()` here?

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +218,77 @@ impl LogItem for LogCmd {
...
+        .to_str()
+        .ok_or(io::Error::new(io::ErrorKind::InvalidInput, "Wrong path"))?;
```
imobachgs, line 256: Adding the path would be useful.

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +218,77 @@ impl LogItem for LogCmd {
...
-        .status()
-    {
-        Ok(_o) => Ok(()),
-        Err(_e) => Err(io::Error::new(
+        .status()?;
+
+    if res.success() {
+        Ok(())
+    } else {
+        Err(io::Error::new(
```
imobachgs, line 275: Perhaps you could define your own error type instead of reusing the `io::Error`.

rust/agama-cli/src/logs.rs
```diff
@@ -278,3 +328,14 @@ fn store(verbose: bool) -> Result<(), io::Error> {
 
     compress_logs(&tmp_dir, &result)
 }
+
+// Handler for the "agama logs list" subcommand
+fn list(options: LogOptions) {
+    for list in [options.paths, options.commands] {
```
imobachgs, line 334: It would indicate what is a file and what is a command in the UI.

imobachgs (commented):

rust/agama-cli/src/logs.rs
mchf: ah ... another chapter ... nice tip ;-)

rust/agama-cli/src/logs.rs
mchf: done

rust/agama-cli/src/logs.rs
mchf: already implemented as a consequence of previous sugestions ;-)

rust/agama-cli/src/logs.rs
mchf: I need to convert it somehow into a list of characters. As far as I understand the implementation, if you used e.g. display it may remove some weird (unprintable) characters from path. This way it should convert as much as possible. I consider such case a little bit obscure thought  but better safe than sorry ;-)

rust/agama-cli/src/logs.rs
mchf: Well ... it would ... But conversion of tmp dir's ```Path``` into a usable string is basically what fails here, so I cannot print it. May be I can rephrase it to something like "Temporary directory cannot be found"

rust/agama-cli/src/logs.rs
mchf: slightly modified

rust/agama-cli/src/logs.rs
mchf: And if you asked why I need ```as_os_str``` here and not in the subsequent case that's because here ```parent``` returns ```Path``` but ```file_name``` used later returns ```OsStr``` already.

rust/agama-cli/src/logs.rs
mchf: tried so

rust/agama-cli/src/logs.rs
mchf: Well ... perhaps ... but I somehow feel it as an io error here ;-)

rust/agama-cli/src/logs.rs
imobachgs: To be honest, I would not expect obscure characters here. But we can keep it as it is.

rust/agama-cli/src/logs.rs
imobachgs: OK, that's fine. I usually prefer to have my error types (keeping the source of the error). But given that `io::Error` belongs to the standard library, let's keep it as it is.

imobachgs (commented):
In general, it looks better. However, I still think that the content of the tarball looks kind of weird:

```
agama-logs.UrXFYtb6nDY7/journalctl--dmesg.err.log
agama-logs.UrXFYtb6nDY7/journalctl--dmesg.out.log
agama-logs.UrXFYtb6nDY7/journalctl-uagama-auto.err.log
agama-logs.UrXFYtb6nDY7/journalctl-uagama-auto.out.log
agama-logs.UrXFYtb6nDY7/journalctl-uagama.err.log
agama-logs.UrXFYtb6nDY7/journalctl-uagama.out.log
```


IMHO, it would be better to have something like `agama.log`, `agama.err`,  `dmesg.log`, `dmesg.err`, `agama-auto.log`, etc. Actually, I am not even sure whether I would keep the `.err` files, but that's another discussion.

Instead of inferring the name from the command, I think extending the `LogCmd` struct to include a basename (or some similar mechanism) is better.

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +230,96 @@ impl LogItem for LogCmd {
...
+        .ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "Malformed path to temporary directory",
+        ))?
+        .as_os_str()
+        .to_str()
+        .ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "Malformed path to temporary directory",
+        ))?;
```
imobachgs, line 281: Sorry, but I see a lot of repetition here with four errors that seem to be the same (but they are not). Anyway, if you are not interested in making any distinction, you could rewrite this code to:

```rust
    let path = tmp_dir.path();
    let tmp_path = path
        .parent()
        .map(|p| p.as_os_str().to_str())
        .flatten()
        .ok_or(io::Error::new(
            io::ErrorKind::InvalidInput,
            "Malformed path to temporary directory",
        ))?;
    let dir = path
        .file_name()
        .map(|p| p.to_str())
        .flatten()
        .ok_or(io::Error::new(
            io::ErrorKind::InvalidInput,
            "Malformed path to temporary directory",
        ))?;
```

And I still see some repetition... :-)

imobachgs (commented):
It looks better, but there is still some repetition.

rust/agama-cli/src/logs.rs
```diff
@@ -22,17 +23,31 @@ pub enum LogsCommands {
...
+        }
     }
 }
 
-const DEFAULT_COMMANDS: [&str; 3] = [
-    "journalctl -u agama",
-    "journalctl -u agama-auto",
-    "journalctl --dmesg",
+const DEFAULT_COMMANDS: [(&str, &str); 3] = [
+    ("journalctl -u agama", "agama"),
```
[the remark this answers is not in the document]
mchf: a few words written

rust/agama-cli/src/logs.rs
```diff
@@ -76,6 +91,27 @@ fn show(show: bool, text: &str) {
...
+    verbose: bool,
+}
+
+impl Default for LogOptions {
+    fn default() -> Self {
+        Self {
+            paths: DEFAULT_PATHS.iter().map(|p| p.to_string()).collect(),
+            commands: DEFAULT_COMMANDS
+                .iter()
+                .map(|p| (p.0.to_string(), p.1.to_string()))
```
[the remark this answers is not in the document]
mchf: why not. Done when touching the code anyway

rust/agama-cli/src/logs.rs
```diff
@@ -22,17 +23,31 @@ pub enum LogsCommands {
...
+// main entry point called from agama CLI main loop
 pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     match subcommand {
-        LogsCommands::Store { verbose } => Ok(store(verbose)?),
-        LogsCommands::List => Err(anyhow::anyhow!("Not implemented")),
+        LogsCommands::Store { verbose } => {
+            // feed internal options structure by what was received from user
+            // for now we always use / add defaults if any
+            let options = LogOptions {
+                verbose: verbose,
```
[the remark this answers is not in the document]
mchf: well i don't like this. I like to be explicit even if it costs some more letters. But if you wish so ...

rust/agama-cli/src/logs.rs
```diff
@@ -181,64 +230,96 @@ impl LogItem for LogCmd {
...
+        )));
     }
 
     log_sources
 }
 
-// compress given directory into a tar archive
+// Compress given directory into a tar archive
 fn compress_logs(tmp_dir: &TempDir, result: &String) -> io::Result<()> {
     let compression = DEFAULT_COMPRESSION.0;
```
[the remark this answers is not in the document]
mchf: placed a few words close to the DEFAULT_COMPRESSION definition

rust/agama-cli/src/logs.rs
mchf: well I can write slightly different message into each error and there won't be any repetition anymore ;-)

mchf: btw i slightly doubt that the snippet works ... bcs ```as_os_str().to_str()``` returns ```Option<&str>```, so you need to unpack it ... that's why I have another ```ok_or``` in my code.

imobachgs: I tried the code before suggesting the change. [parent](https://doc.rust-lang.org/std/path/struct.Path.html#method.parent) already returns an Option, so I call [map](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) to work with the `Some` case. As you said, `as_os_str().to_str()` returns another option, so you get an `Option<Option<&str>>`. And that's why I call [flatten](https://doc.rust-lang.org/std/option/enum.Option.html?search=flatten#method.flatten), which returns a single `Option<&str>` and allows me to call `ok_or`.

mchf: well to be honest, I see the suggested code a little bit less readable. However I haven't found anything better so far.

mchf: I mean ... it can be of course perfectly fine ... but i need to invest into it a bit to make it common to me ;-)

mchf: on the end after enough reading i found something similar, but one word shorter ;-) So, I'll test it and push if it works ;-)

imobachgs: Sure, and_then looks like a better option here. Actually, *Rust by example* [confirms It](https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html).

imobachgs (approved):

PR #781, opened 2023-09-29
Reconnect to the D-Bus client

imobachgs:
## Problem

When the system is under heavy load (e.g., you are running Agama on a VM and the host system is busy) it the web UI might fail to connect to the D-Bus service.

## Solution

* Implement a reconnection mechanism in the `InstallerClientProvider`. The provider keeps trying to reconnect to the server, but it displays an error message after a few attempts.
* We needed to move the `Layout` from the `App` to the `InstallerClientProvider`. The main reason is that there are several components between `InstallerClientProvider` and `App` which depend on the installer client.
* As a side-effect, we could eliminate several calls to the `mockLayout` function.
* Drop the `mockComponent` because 1) it was undefined most of the times and 2) it is simple enough.

## Testing

- Added a new unit test
- Tested manually

## TODO

- [x] Move the layout to the `InstallerClientProvider`
- [x] Implement a reconnection mechanism
- [x] Update the tests
- [x] Extend `InstallerClientProvider` tests


web/src/context/installer.jsx
```diff
@@ -52,18 +72,41 @@ const BUS_ADDRESS_FILE = "/run/agama/bus.address";
...
+  }, [setError, setRetry, retry]);
 
   useEffect(() => {
-    if (client !== undefined) {
-      const file = cockpit.file(BUS_ADDRESS_FILE);
-      file.read().then(address => {
-        setValue(createClient(address));
-      });
+    const connectClient = async () => {
+      const newClient = client || await createDefaultClient();
```
joseivanlopez, line 89: NP: `newClient` could be undefined.

web/src/context/installer.jsx
```diff
@@ -52,18 +72,41 @@ const BUS_ADDRESS_FILE = "/run/agama/bus.address";
...
+      } else {
+        resetClient();
+      }
+    };
+
+    if (value === undefined) {
+      connectClient();
     }
-  }, [client]);
+  }, [client, value, resetClient, retry]);
```
joseivanlopez, line 102: Perhaps defining `#resetClient` inside the `useEffect` saves some re-renderings, because the current callback is redefine every time it is called (it changes its `retry` dependency). So, calling to `resetClient` forces a render.

web/src/context/installer.jsx
```diff
@@ -41,33 +42,65 @@ function useInstallerClient() {
...
-const BUS_ADDRESS_FILE = "/run/agama/bus.address";
+const ATTEMPTS = 3;
+const INTERVAL = 2000;
 
 /**
   * @param {object} props
   * @param {import("~/client").InstallerClient|undefined} [props.client] client to connect to
   *   Agama service; if it is undefined, it instantiates a new one using the address
   *   registered in /run/agama/bus.address.
+  * @param {number} [props.interval] - Interval in milliseconds between connection attempts
```
dgdavid, line 53: NP:

```suggestion
  * @param {number} [props.interval=2000] - Interval in milliseconds between connection attempts
```

web/src/context/installer.jsx
```diff
@@ -41,33 +42,65 @@ function useInstallerClient() {
...
+  *   (2000 by default).
+  * @param {number} [props.max_attempts] - Connection attempts before displaying an
```
dgdavid, line 55: NP:

```suggestion
  * @param {number} [props.max_attempts=3] - Connection attempts before displaying an
```

web/src/test-utils.js
```diff
@@ -72,6 +74,12 @@ jest.mock('react-router-dom', () => ({
 const Providers = ({ children }) => {
   const client = createClient();
 
+  // FIXME: workaround to fix the tests. We should inject
+  // the client instead of mocking `createClient`.
+  if (!client.onDisconnect) {
+    client.onDisconnect = noop;
+  }
```
dgdavid, line 81: What about creating an issue from here.

dgdavid (approved):
Loos good, thanks

joseivanlopez (approved):

PR #782, opened 2023-09-29
Refine definition of storage volumes for ALP Dolomite

ancorgs:
## Problem

The storage configuration for ALP Dolomite was directly inferred from the corresponding image, just because it was the only source of information.

## Solution

Now we have a better knowledge of the storage needs of ALP Dolomite, so the configuration has been updated according to that.

- The root volume is still a mandatory Btrfs file system with read-only root subvolume and snapshots. That's a requirement for the immutable nature of Dolomite.
- Now it's possible to configure a separate `/var` file system. That's useful (even recommended) for hosting all the containers in a way in which they cannot lead to filling the main file-system[1].
- If a separate `/var` is used, the root volume will not grow beyond the 25 GiB, to leave the rest of the space to `/var`.
- No matter if `/var` is a subvolume or a separate file system, it's always configured to NOT do copy-on-write.
- Creating a separate swap is NOT offered as an option. Having a swap partition can lead to problems with Kubernetes.

[1] Having `/var` as a subvolume and using Btrfs quotas to limit the grow of containers is not an option because Btrfs quotas can be problematic when used alongside other Dolomite technologies.

## Testing

Manually tested. The file systems and subvolumes are created as expected and the UI behaves as seen in the following screenshots.

dolomite-var1

dolomite-var2

dolomite-var3

### DIFF
--- a/service/etc/agama.yaml
+++ b/service/etc/agama.yaml
@@ -82,6 +82,9 @@ ALP-Dolomite:
             - path: srv
             - path: boot/writable
             - path: usr/local
+            - path: var
+              copy_on_write: false
+            # Architecture specific subvolume
             - path: boot/grub2/arm64-efi
               archs: aarch64
             - path: boot/grub2/i386-pc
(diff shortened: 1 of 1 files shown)


dgdavid (approved):

joseivanlopez (approved):
LGTM

PR #783, opened 2023-09-29
Add bond model

cfconrad:
## Problem

* Bond wasn't implemented yet

## Solution

* Add bond Model in `network::model`
* Impl Bond interface and Parent device to NetworkManager DBus

## Testing

- *Tested manually* 
- See also https://github.com/jcronenberg/agama/pull/1 where we can run migration from wicked.xml files via 
```bash
cargo run --bin agama wicked migrate agama-migrate-wicked/tests/bond/
```


rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -412,6 +443,15 @@ fn wireless_config_from_dbus(conn: &OwnedNestedHash) -> Option<WirelessConfig> {
     Some(wireless_config)
 }
 
+fn bond_config_from_dbus(conn: &OwnedNestedHash) -> Option<BondConfig> {
+    let Some(_bond) = conn.get(BOND_KEY) else {
+        return None;
+    };
+
+    // TODO
```
cfconrad, line 451: I wasn't able to convert the given value to a HashMap<String, String>. Maybe someone can help?

jcronenberg: Something like this?
```
    let options: zvariant::Dict = bond
        .get("options")?
        .downcast_ref::<Value>()?
        .try_into()
        .unwrap();
    let miimon: &str = options.get("miimon").unwrap()?;
```

cfconrad: Almost, I don't need (and don't even know what could be there) an specific value but all.
More like this...
```
let h: HashMap<String,String> = HashMap::new();
options.iter().map(|e| h.insert(e.0 , e.1) );
```
But with correct syntax :)

jcronenberg: I guess something like this would be a way to generate a HashMap:
```
    let options_dict: zvariant::Dict = bond
        .get("options")?
        .downcast_ref::<Value>()?
        .try_into()
        .unwrap();
    let options: HashMap<String, String> = options_dict.try_into().unwrap();
```
although I'm not sure if there is maybe a simpler way.

PR #785, opened 2023-10-02
add warning if there are no repos for given product

jreidinger:
## Problem

Some product will no longer provide public repos and has mandatory registration.

## Solution

Inform user that for such products registration is mandatory.

## Screenshots

note: I manually modify agama.yml and remove all repos for ALP Dolomite product.
before:
agama-software-old-no-repos
after:
agama-software-new-no-repos


jreidinger:
replaced by different approach in registration branch

PR #788, opened 2023-10-03
[web] Start using `PF/Dropdown#onOpenChange` prop

dgdavid:
## Problem

No problem at all, actually :wink: Just a PR to start using the `#onOpenChange` prop of [PF5/Dropdown](https://www.patternfly.org/components/menus/dropdown#dropdown), which wasn't present in the former PF4/Dropdown as I commented in https://github.com/openSUSE/agama/issues/552.

## Solution

`PF/Dropdown#onOpenChange` prop should be used whenever needed. As a result, this undoes or replaces the [great contribution](https://github.com/openSUSE/agama/pull/671) made by @balsa-asanovic only 3 months ago. However, I feel relying on the library, when possible, is the right path to take. While click handling is nearly identical, they are taking care of the keyboard, and there is a `onOpenChangeKeys` beta prop that might be useful in the future. Have a look to the [source code](https://github.com/patternfly/patternfly-react/blob/7767e6a96195dffbf155c7608d9aad0c14303f03/packages/react-core/src/components/Dropdown/Dropdown.tsx#L96-L137) if you're curious.

## Tests

Adapted the original unit test just for using the user events API instead of the `fireClick` method, but the test passed as it was after made the change.

### DIFF
--- a/web/src/components/core/PageOptions.jsx
+++ b/web/src/components/core/PageOptions.jsx
@@ -19,7 +19,7 @@
  * find current contact information at www.suse.com.
  */
 
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState } from 'react';
 import {
   Dropdown, DropdownGroup, DropdownItem, DropdownList,
   MenuToggle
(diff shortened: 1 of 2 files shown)


dgdavid:
@balsa-asanovic, 

I'd like to have a review from you here :innocent:

balsa-asanovic:
> this undoes or replaces the https://github.com/openSUSE/agama/pull/671 made by @balsa-asanovic only 3 months ago

😢 

Kidding of course, I am glad for this change to happen as it clearly shows that the installer keeps getting better and better.

> I'd like to have a review from you here 😇

Haha, I think I prefer it the other way around, but here it goes 🤓 .

The change seems great to me, by using the new onOpenChange prop of Dropdown you get to remove the need to check if the user clicked outside of the component in the component itself, PatternFly now handles this.

I wasn't really a big fan of using useRef hook and then attaching event listener to the whole document to achieve this behavior, it seemed a bit crude, but at the time it looked like there wasn't a better option.

Now, this is all cleaned up (useRef, useEffect and attaching of listeners), with a simple: `onOpenChange={close}`.

So, by all means a great change, green light from me 🟢

balsa-asanovic (approved):

imobachgs (commented):
Thanks you @dgdavid and @balsa-asanovic. LGTM too.

dgdavid:
Thanks guys. I fear we need an approval from someone inside the team too for unblocking the _merge_ button.

imobachgs (approved):

PR #789, opened 2023-10-03
Implement read only flag

jreidinger:
## Problem

There is no indication if btrfs volume will be read only or not.

## Solution

Implement it.

## Testing

- *Tested manually*

## Screenshots
Note: No suitable icon in materials not found ( at least I do not like any of them ).

read_only_volume


web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -201,6 +201,10 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
 
   const Details = ({ volume, options }) => {
     const hasSnapshots = volume.fsType === "Btrfs" && volume.snapshots;
+    const readOnly = volume.fsType === "Btrfs" && volume.readOnly;
+    console.log("details:");
+    console.log(volume);
+    console.log(options);
```
dgdavid, line 207: Leftovers, right?

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -214,6 +218,8 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
         <If condition={options.encryption} then={<Em icon={lockIcon}>{_("encrypted")}</Em>} />
         {/* TRANSLATORS: filesystem flag, it allows creating snapshots */}
         <If condition={hasSnapshots} then={<Em icon={snapshotsIcon}>{_("with snapshots")}</Em>} />
+        {/* TRANSLATORS: filesystem flag, system is read only */}
```
dgdavid, line 221: Serious question. System or filesystem?

dgdavid (commented):
So far, LGTM. But have a look to these `console.log`. I think they are no really needed :)

dgdavid:
> Tested manually

@jreidinger I have added few unit tests for the UI part. Hope you don't mind. Please, have a look.

web/src/components/storage/ProposalVolumes.jsx
jreidinger: yeah, forget last force push

web/src/components/storage/ProposalVolumes.jsx
jreidinger: good question. As it is generic volume, it is filesystem as system can live elsewhere.

dgdavid (approved):
LGTM

dgdavid (changes requested):
Sorry, I think this one deserve an entry in the changelog files

ancorgs:
I have concerns regarding the wording. This is not exactly a "read-only" filesystem.

I will elaborate (entering a meeting now)

jreidinger:
@ancorgs OK, I will wait.

ancorgs:
`volume.btrfs.read_only?` does not really mean the file-system is read-only.

To be precise, it means the root subvolume of the Btrfs file-system is read-only. But that does not apply to the whole file-system. Other subvolumes on the same Btrfs file-system (like /home, /opt, /srv or /var) are still writable.

If I see a Volume with an attribute `ReadOnly` set to true or if a see a `read-only` label in the table of volumes in the UI, I would expect the whole volume/filesystem to be read-only. Maybe we even have an use-case for that in the future.

I don't know how to name the D-Bus property or the label in the UI, but I don't like the current ReadOnly and read-only. Maybe we should try with "transactional".

jreidinger:
@ancorgs  OK, makes sense. So lets use transactional name instead. I am just not sure if it btrfs exclusive

ancorgs (approved):
LGTM

jreidinger:
@dgdavid Can you please also recheck? as it looks like your requests to change blocks merge

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -214,6 +215,8 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
         <If condition={options.encryption} then={<Em icon={lockIcon}>{_("encrypted")}</Em>} />
         {/* TRANSLATORS: filesystem flag, it allows creating snapshots */}
         <If condition={hasSnapshots} then={<Em icon={snapshotsIcon}>{_("with snapshots")}</Em>} />
+        {/* TRANSLATORS: filesystem flag, filesystem is read only */}
```
dgdavid, line 218:
```suggestion
        {/* TRANSLATORS: flag for transactional file system  */}
```

dgdavid (approved):
LGTM, just check the comment for translators

PR #790, opened 2023-10-04
[web] Fix font families

dgdavid:
## Problem

Font families overwrites/definitions weren't properly updated in the migration to PatternFly 5.

## Solution

Update the _web/src/assets/styles/patternfly-overrides.scss_ file to be _in sync_ with https://github.com/patternfly/patternfly/blob/5233e05c41cae220cf062998839f8881dfd8a080/src/patternfly/base/_variables.scss#L246C1-L252

## Testing

- Tested manually

## Screenshots

| Before | After |
|-|-|
| Fonts before updating overrides file | Fonts after updating overrides file |

## Acknowledgements

Thanks @lslezak for catching and reporting it :smiley:

### DIFF
--- a/web/src/assets/styles/patternfly-overrides.scss
+++ b/web/src/assets/styles/patternfly-overrides.scss
@@ -3,17 +3,12 @@
   // from brand.suse.com
 
   // Font families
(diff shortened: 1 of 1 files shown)


lslezak (approved):
LGTM, thanks!

PR #791, opened 2023-10-04
D-Bus API for registration

joseivanlopez:
## Problem

Agama needs to know the repositories of the product to install. But for some products (e.g., SLE, ALP), the repositories are unknown until the product is registered.

## Solution

Export a new D-Bus object `/org/opensuse/Agama/Software1/Product` to manage the product selection and registration. Note that methods for selecting a product were moved from `/org/opensuse/Agama/Software1` object.

~~~
# busctl --address unix:path=/run/agama/bus introspect org.opensuse.Agama.Software1 /org/opensuse/Agama/Software1/Product
NAME                                 TYPE      SIGNATURE  RESULT/VALUE                             FLAGS
org.opensuse.Agama.Software1.Product interface -          -                                        -
.SelectProduct                       method    s          (us)                                     -
.AvailableProducts                   property  a(ssa{sv}) 3 "ALP-Dolomite" "SUSE ALP Dolomite" 1 … emits-change
.SelectedProduct                     property  s          "Tumbleweed"                             emits-change
org.opensuse.Agama1.Issues           interface -          -                                        -
.All                                 property  a(ssuu)    0                                        emits-change
org.opensuse.Agama1.Registration     interface -          -                                        -
.Deregister                          method    -          (us)                                     -
.Register                            method    sa{sv}     (us)                                     -
.Email                               property  s          ""                                       emits-change
.RegCode                             property  s          ""                                       emits-change
.Requirement                         property  u          0                                        emits-change
~~~

Note 1: the current code still uses the [libsuseconnect ruby bindings](https://github.com/SUSE/connect-ng/tree/main/yast) to communicate to SCC. That bindings only exist because of YaST, and it should not be used for new projects. The plan is to directly use SUSEConnect CLI in Agama instead. But for that, SUSEConnect needs some improvements in order to avoid applying changes on the running system. SCC team is already working on it, see https://github.com/SUSE/scc-docs/pull/287. 

Note 2: the current implementation works similar to YaST, that is, the service, repositories, SCC credentials, etc are directly added to the running system (e.g., TW running on agama-live) and then they are copied to the target system. Changes altering the running system should be avoided. This will be improved in a follow-up task, see https://trello.com/c/jqcFar3M/243-agama-registration-avoid-changes-in-running-system.

Note 3: RMT and SSL certificates are not covered yet. This will come in a next iteration.
 
Note 4: the target of this PR is the *registration* feature branch. It will be merged into *master* once the registration UI is implemented.

Note 5: changelog will be added later at the time of merging into master.

## Testing

* Added unit tests
* Tested manually


service/lib/agama/dbus/software/manager.rb
```diff
@@ -54,20 +54,20 @@ def initialize(backend, logger)
...
-          dbus_reader :selected_base_product, "s"
-
-          # documented way to be able to write to patterns and trigger signal
-          attr_writer :selected_patterns
+          dbus_reader :available_products, "a(ssa{sv})"
 
-          # selected patterns is hash with pattern name as id and 0 for user selected and
-          # 1 for auto selected. Can be extended in future e.g. for mandatory patterns
-          dbus_attr_reader :selected_patterns, "a{sy}"
+          dbus_reader :selected_product, "s"
```
jreidinger, line 70: when we touch it we probably also need to solve issue with version...as there can be two products with same name and different version like it happen with micro based on some SP. And registration need version....maybe product should be tuple of strings with name and version?

service/lib/agama/dbus/software/manager.rb
```diff
@@ -159,6 +162,45 @@ def finish
...
+        def email
+          backend.registration.email || ""
+        end
+
+        # Replace #State by #IsDisabled and #isOptional ?
+        def state
+          return 0 if backend.registration.disabled?
+          return 1 if backend.registration.optional?
+          return 2 unless backend.registration.optional?
+        end
```
jreidinger, line 202: when touching code I think we need also probe method for registration that compute state after product is selected.

service/lib/agama/registration.rb
```diff
@@ -26,9 +32,26 @@ class Registration
     # initializes registration with instance of software manager for query about products
     def initialize(software_manager)
       @software = software_manager
+      @on_state_change_callbacks = []
     end
 
-    def register(code, email: nil)
+    def register(code, email: "")
+      target_distro = "ALP-Dolomite-1-x86_64" # TODO read it
+      registration = Registration::Registration.new # intentional no url yet
```
joseivanlopez, line 40: I think this would not work if `#announce_system` fails because `#catch_registration_errors` could call to YaST UI.

I would directly call to `#announce_system` and `#create_credentials_file` instead of using `Registration`.

service/lib/agama/registration.rb
```diff
@@ -75,7 +77,7 @@ def deregister
       Y2Packager::NewRepositorySetup.instance.services.delete(@service.name)
 
       connect_params = {
-        token: @system_code,
+        token: reg_code,
```
jreidinger, line 80: thanks for fix. It was leftover when I testing what works with deregister

service/lib/agama/registration.rb
```diff
@@ -100,6 +104,18 @@ def on_change(&block)
...
 
+    attr_reader :software
+
+    def product
+      software.product
+    end
+
+    # E.g., "ALP-Dolomite-1-x86_64"
+    def target_distro
+      v = version.to_s.split(".").first || "1"
```
jreidinger, line 115: I think that we should at least log if version parameter is missing ( aka default 1 ) otherwise it can cause us some troubles later when we debug it

service/lib/agama/registration.rb
```diff
@@ -0,0 +1,123 @@
...
+
+      # TODO: fill it properly for scc
+      target_product = OpenStruct.new(
+        arch:       Yast::Arch.rpm_arch,
+        identifier: product.id,
+        version:    product.version
+      )
+      activate_params = {}
+      @service = SUSE::Connect::YaST.activate_product(target_product, activate_params, email)
+      Y2Packager::NewRepositorySetup.instance.add_service(@service.name)
```
jreidinger, line 69: @lslezak I am not sure if we need this call in Agama, at first I thought it adds to libzypp new service, but it looks more that it is just for migration. Am I right?

service/lib/agama/dbus/software/manager.rb
```diff
@@ -101,6 +102,13 @@ def initialize(backend, logger)
             ]
           end
 
+          # documented way to be able to write to patterns and trigger signal
```
mvidner, line 105: @jreidinger since ruby-dbus-0.23.1 (in Factory) you can use a single call,

```rb
dbus_reader_attr_accessor :selected_patterns, "a{sy}"
```

service/lib/agama/registration.rb
lslezak: Yes, that class is used to track the new added services or repositories so at the end of migration we can find out which are the new repositories (must be kept in the system) or the old ones (should be removed).

See https://github.com/yast/yast-yast2/blob/master/library/packages/src/lib/y2packager/new_repository_setup.rb#L19-L21

jreidinger: note for me: method we will be needed is this one https://github.com/yast/yast-registration/blob/master/src/lib/registration/sw_mgmt.rb#L365 I will try to reduce it as much as possible for currently used features in agama

service/lib/agama/registration.rb
```diff
@@ -0,0 +1,173 @@
...
+
+    attr_reader :software
+
+    def product
+      software.product
+    end
+
+    # E.g., "ALP-Dolomite-1-x86_64"
+    def target_distro
+      v = version.to_s.split(".").first || "1"
```
jreidinger, line 118: here it crashes with unknown method version. I will try to fix it

joseivanlopez: Thanks!

service/lib/agama/dbus/software/manager.rb
joseivanlopez: Property change signal for Requirement is emitted when a new product is selected.

PR #794, opened 2023-10-09
Unify L10n wrappers

imobachgs:
We use two different components to set the language: `L10nWrapper` and `L10nBackendWrapper`. After the refactoring introduced in #781, we can merge both of them into a single component. As an additional benefit, we save one reload when changing the language.

## Testing

- Manual tests
- Updated unit tests


web/src/L10nWrapper.jsx
```diff
@@ -19,68 +19,154 @@
...
+/**
+ * Returns the current locale according to Cockpit
+ *
+ * It takes the locale from the CockpitLang cookie.
+ *
+ * @return {string|undefined} language tag in xx_XX format or undefined if
+ *   it was not set.
+ */
+function cockpitLanguage() {
+  // language from cookie, empty string if not set (regexp taken from Cockpit)
```
jreidinger, line 37: if regexp is taken from cockpit I would appreciate if you can add also link to its source where it lives, so we can check if our copy is still identical.

imobachgs: Good point. Well, I just moved that code, but I guess that we took that expression from https://github.com/cockpit-project/cockpit/blob/98a2e093c42ea8cd2431cf15c7ca0e44bb4ce3f1/pkg/shell/shell-modals.jsx#L91.

jreidinger (approved):

jreidinger (approved):

PR #796, opened 2023-10-11
Allow changing the language for the web UI

imobachgs:
## Problem

The only way to change the language of the UI is through the `lang` argument in the URL (just for testing purposes).

## Solution

The idea was to implement a `LanguageSwitcher` to allow the user select a different language for the UI.

As a first step, I extracted the logic to get/set the UI language to a new `L10nContext`, getting rid of `L10nWrapper`. I also took the chance to refactor our providers, as we were unhappy with the current organization. That's the reason the PR is so "big".

Now, our providers do not display any UI and do not make decisions about whether to display the children components. They just *provide* data/functions to them.

About the language switcher, I follow @dgdavid's proposal and placed it in the sidebar.

## Testing

- *Added a new unit test* (pending)
- *Tested manually*

## To do

- [x] Add unit tests for the `LanguageSwitcher`.
- [x] Improve logic to fallback to a supported language.

## Screenshots

<details>
<summary>Changing the language from `en-US` to `es-ES`</summary>

[Grabación de pantalla desde 2023-10-11 09-34-39.webm](https://github.com/openSUSE/agama/assets/15836/ab4d0f10-e3cf-440f-8a6c-faa3580f78c4)
</details>


lslezak:
I have few comments to this topic:

### List of Supported Locales

Here is a summary how Cockpit handles that.

- When merging the translations from the [cockpit-weblate](https://github.com/cockpit-project/cockpit-weblate) to the main [cockpit](https://github.com/cockpit-project/cockpit/tree/main/po):
  -  it removes the files which have less than 50% translated texts ([code](https://github.com/cockpit-project/bots/blob/56a9418e07cf1156a50fe3e280fd39a2bc4c754b/po-refresh#L109))
  - it saves the language list into [LINGUAS](https://github.com/cockpit-project/cockpit/blob/953a6eeb9be2b6cd1c5c95fce7c0db19749bc72f/po/LINGUAS) and [manifest.json](https://github.com/cockpit-project/cockpit/blob/953a6eeb9be2b6cd1c5c95fce7c0db19749bc72f/pkg/shell/manifest.json#L6-L26) files
  - it uses the [language_map.txt](https://github.com/cockpit-project/cockpit/blob/953a6eeb9be2b6cd1c5c95fce7c0db19749bc72f/po/language_map.txt) with language names to generate the data, when a new language is added it's name is found automatically (unless it is some really rare language)
- The language selector loads the list of languages from the `manifest.json` file ([code](https://github.com/cockpit-project/cockpit/blob/953a6eeb9be2b6cd1c5c95fce7c0db19749bc72f/pkg/shell/shell-modals.jsx#L107))

### Agama

But the situation with Agama is more complex because some texts are obtained from 3rd party libraries (like libstorage-ng or libzypp). So we should offer only the languages which are covered by all components. On the other hand both libstorage-ng and libzypp support more than 50 languages so the vast majority is already covered so this should not be a big problem.

### Language Detection

We also need to improve the language detection from the browser. Currently it just uses the `navigator.language` value which returns a single language (the most preferred one). ([code](https://github.com/openSUSE/agama/blob/99ab3e45d960d166bb31ca103b79cd5fce6742dd/web/src/L10nWrapper.jsx#L154))

But maybe this most preferred language is not supported by Agama. It should rather use `navigator.languages` and find the first supported language in the list.

### Local Installation

When installing remotely we can detect the user preferred language (see the previous section). But when installing locally the preferred language is not set.

The language selector in the side bar is quite hidden. In this case we should probably display a popup or configuration page at the very beginning, something similar to the product selection dialog which is displayed before the main overview dialog.

(The local installation can be detected by the URL, it runs at `http://127.0.0.90`.)

lslezak:
### List of Supported Locales

Alternatively we can get the list of supported locales at the build time, using a webpack plugin in the `web/webpack.config.js` file:

```js
  new webpack.DefinePlugin({ LANGUAGES: JSON.stringify(
    glob.sync(path.resolve(__dirname, "po/*.po")).map(po => path.basename(po).slice(0, -3)))
  }),
```

then the `LANGUAGES` text in the code is replaced with the `['cs', 'es', 'ja', 'nl', 'ru', 'sv', 'uk']` array.  (This is just a PoC, we would need to enhance that to contain also the language names.)

#### Note

I'd avoid using DBus if possible, that just adds more complexity. And if accessing DBus fails you would get an untranslated error message. If you need to just read a file from the system you can use the [cockpit.file API](https://cockpit-project.org/guide/latest/cockpit-file.html) instead of writing a DBus service.

lslezak:
> Alternatively we can get the list of supported locales at the build time, using a webpack plugin in the `web/webpack.config.js` file:

Full example: https://github.com/openSUSE/agama/compare/poc_supported_languages?expand=1

imobachgs:
> I have few comments to this topic:
> ### List of Supported Locales
> 
> Here is a summary how Cockpit handles that.

[..]

Nice summary. We might use a similar approach.

> ### Agama
> 
> But the situation with Agama is more complex because some texts are obtained from 3rd party libraries (like libstorage-ng or libzypp). So we should offer only the languages which are covered by all components. On the other hand both libstorage-ng and libzypp support more than 50 languages so the vast majority is already covered so this should not be a big problem.

I would assume that the backend is always translated, at least as a first implementation. I would expect YaST, libstorage-ng, libzypp and friends to be already translated to *a lot* of languages.

> ### Language Detection
> 
> We also need to improve the language detection from the browser. Currently it just uses the `navigator.language` value which returns a single language (the most preferred one). ([code](https://github.com/openSUSE/agama/blob/99ab3e45d960d166bb31ca103b79cd5fce6742dd/web/src/L10nWrapper.jsx#L154))
> 
> But maybe this most preferred language is not supported by Agama. It should rather use `navigator.languages` and find the first supported language in the list.

👍🏻 

> ### Local Installation
> 
> When installing remotely we can detect the user preferred language (see the previous section). But when installing locally the preferred language is not set.
> 
> The language selector in the side bar is quite hidden. In this case we should probably display a popup or configuration page at the very beginning, something similar to the product selection dialog which is displayed before the main overview dialog.
> 
> (The local installation can be detected by the URL, it runs at `http://127.0.0.90`.)

Yes, the team already discussed this topic. Actually, @dgdavid's proposal contains [a welcome screen](https://github.com/openSUSE/agama/blob/draft-language-switcher/web/src/components/core/WelcomePage.jsx). But by now, it is out of scope of this PR.

imobachgs:
> ### List of Supported Locales

[..]

> #### Note
> 
> I'd avoid using DBus if possible, that just adds more complexity. And if accessing DBus fails you would get an untranslated error message. If you need to just read a file from the system you can use the [cockpit.file API](https://cockpit-project.org/guide/latest/cockpit-file.html) instead of writing a DBus service.

Good idea, it makes sense.

imobachgs:
A (partial) list of supported languages is now in the `manifest.json` file. We need to add the missing languages to such a list and to automate the process.

lslezak:
OK, then we need to update the [weblate-merge-po.yml](https://github.com/openSUSE/agama/blob/master/.github/workflows/weblate-merge-po.yml) action to modify the `manifest.json` file.

Maybe even a simple shell script can be sufficient, for modifying a JSON file you can use the [jq](https://jqlang.github.io/jq/manual/) tool. For the language names we could reuse the [language_map.txt](https://github.com/cockpit-project/cockpit/blob/main/po/language_map.txt) from Cockpit (and maybe convert to JSON to make the processing easier and unified).

See the [po-refresh](https://github.com/cockpit-project/bots/blob/main/po-refresh) Python script in Cockpit.

imobachgs:
Great. I will have a look into it.

About using Cockpit's map, I do not like that some languages are capitalized while others aren't ("English" vs "español"). Perhaps there is a reason for that, but I would prefer to have a more consistent list.

web/src/context/l10n.jsx
```diff
@@ -108,26 +128,52 @@ function languageToBackend(tag) {
...
 
 /**
- * This is a helper component to set the language. It uses the
+ * Returns the list of languages from the navigator in RFC 5646 (or BCP 78)
+ * format
+ *
+ * @return {Array<string>} List of languages from the navigator
+ */
+function navigatorLanguages() {
+  return navigator.languages.map(l => l.toLowerCase());
```
lslezak, line 137: Important note: the browser preferred languages might not contain the country code, only the language code. For example my personal setting is `['cs', 'en-US', 'en']`. That means Czech, English (US), English (generic, any variant available).

So if the manifest would contain `cs-cz` it would not match. The search for the supported language should be less strict and allow partial match.

web/src/index.js
```diff
@@ -72,31 +69,27 @@ const root = createRoot(container);
...
-                  <Route path="/issues" element={<IssuesPage />} />
-                </Route>
-                <Route path="products" element={<ProductSelectionPage />} />
-              </Route>
-            </Routes>
-          </HashRouter>
-        </NotificationProvider>
-      </SoftwareProvider>
-    </InstallerClientProvider>
+    <AgamaProviders>
```
dgdavid, line 72: :+1:

web/src/context/l10n.jsx
```diff
@@ -146,7 +146,27 @@ function navigatorLanguages() {
  */
 function findSupportedLanguage(languages) {
   const supported = Object.keys(cockpit.manifests.agama?.locales || {});
-  return languages.find(lang => supported.includes(lang));
+
+  for (const candidate of languages) {
+    const match = supported.find(s => s.startsWith(candidate));
```
lslezak, line 151: In theory there might be 3 letter languages, after selecting "Cherokee" in Chrome it returns `'chr'` in the `navigator.languages` list. I'd rather not rely on the fact that all language codes have the same length.

I'd rather split the locale with `-` and compare the parts. It is more complex, but more reliable.

lslezak (commented):
Language matching

lslezak (approved):

PR #797, opened 2023-10-13
Added ESlint plugin for checking translation issues

lslezak:
## Problem

- Recently I found in the code passing string templates to the translation function, something like ``_(`foo bar`)``
- Although technically it is OK, the code works, the problem is that the GNU `xgettext` cannot extract the string to the POT file, so in the end the text is not translated 
- Potentially there some more problematic cases like `_("foo" + "bar")`

## Solution

- Implement an ESLint rule which ensures that only string literals are passes to the translation functions

## Testing

- Added a new unit test

```
> npx jest --no-coverage string-literals
 PASS  eslint-plugin-agama-i18n/string-literals.test.js
  string-literals
    valid
      ✓ _("foo") (31 ms)
      ✓ _('foo') (3 ms)
      ✓ n_("one", "many", count) (3 ms)
      ✓ n_('one', 'many', count) (3 ms)
    invalid
      ✓ _(null) (4 ms)
      ✓ _(undefined) (3 ms)
      ✓ _(42) (2 ms)
      ✓ _(foo) (2 ms)
      ✓ _(foo()) (3 ms)
      ✓ _(`foo`) (3 ms)
      ✓ _("foo" + "bar") (3 ms)
      ✓ _('foo' + 'bar') (4 ms)
      ✓ _() (3 ms)
      ✓ n_('foo') (3 ms)
      ✓ n_("foo") (2 ms)
      ✓ n_(foo) (2 ms)
      ✓ n_(foo, bar) (2 ms)
      ✓ Nn_(foo, bar) (2 ms)

Test Suites: 1 passed, 1 total
Tests:       18 passed, 18 total
Snapshots:   0 total
Time:        1.731 s, estimated 2 s
Ran all test suites matching /string-literals/i.
```

### DIFF
--- a/doc/i18n.md
+++ b/doc/i18n.md
@@ -23,6 +23,7 @@
       - [Formatting Texts](#formatting-texts)
       - [TRANSLATORS Comments](#translators-comments)
     - [Missing Translations](#missing-translations)
+    - [ESLint Plugin](#eslint-plugin)
     - [Testing Language](#testing-language)
     - [Building POT File](#building-pot-file)
     - [Cockpit Details](#cockpit-details)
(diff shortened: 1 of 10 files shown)


dgdavid (approved):
LGTM, thanks a lot for such a great check.

PR #802, opened 2023-10-19
[web] Reintroduce blurring when sidebar is open

dgdavid:
## Problem

Few months ago, [a couple of CSS filters were added](https://github.com/openSUSE/agama/pull/564) to help sight users understanding quickly which parts of the UI are _inert_. Later, it was needed to [fine tune the rules](https://github.com/openSUSE/agama/pull/588) for applying these filters in order to not get blurred elements that shouldn't be. As a consequence, Agama stopped applying these filters to the main wrapper when the sidebar is open because the wrapper is not a direct child of `body`, but a direct child of `div#root`

## Solution

To use an `id` for identifying the _Agama main wrapper_ and use it for applying these effects too when it is _inert_ or _aria-hidden_.

## Testing

Tested manually

## Screenshots

| Before | After |
|-|-|
|Screen Shot 2023-10-19 at 15 02 17 |Screen Shot 2023-10-19 at 15 02 08 |

### DIFF
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Oct 19 22:07:11 UTC 2023 - David Diaz <dgonzalez@suse.com>
+
+- UI: fix CSS rule for applying grayscale and blur CSS filters to
+  main wrapper when the sidebar is open (gh#openSUSE/agama#802).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

imobachgs:
What about updating the changes file?

dgdavid:
> What about updating the changes file?

I didn't do it because I think this is a bug fix of what was introduced at https://github.com/openSUSE/agama/pull/564 but broken in #588

So, not sure if it deserves another entry in the changes file.

imobachgs (approved):

PR #812, opened 2023-10-23
Set more restrictive permissions to archive produced by agama logs store command

mchf:
## Problem

logs produced by ```agama logs store``` were publicly readable

## Solution

owner is set to root and permissions to r/w for the owner only

## Testing

- *Tested manually*

## Screenshots

agama-logs-permissions

### DIFF
--- a/rust/agama-cli/src/logs.rs
+++ b/rust/agama-cli/src/logs.rs
@@ -6,6 +6,7 @@ use std::fs;
 use std::fs::File;
 use std::io;
 use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use tempdir::TempDir;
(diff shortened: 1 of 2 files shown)


rust/package/agama-cli.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Oct 23 14:43:59 UTC 2023 - Michal Filka <mfilka@suse.com>
+
+- Improved "agama logs store"
```
imobachgs, line 4: Please, add a reference (`gh#openSUSE/agama#812` is OK).

rust/agama-cli/src/logs.rs
```diff
@@ -304,6 +305,25 @@ fn compress_logs(tmp_dir: &TempDir, result: &String) -> io::Result<()> {
...
+    let attr = fs::metadata(archive)?;
+    let mut permissions = attr.permissions();
+
+    // set the archive file permissions to -rw-------
+    permissions.set_mode(0o600);
+    fs::set_permissions(archive, permissions)?;
+
+    // set the archive owner to root:root
+    // note: std::os::unix::fs::chown is unstable for now
+    Command::new("chown")
```
imobachgs, line 320: Just a note: it was kind of unexpected (to me) that we have an API to change the permissions but not for changing the owner.

imobachgs (approved):

rust/agama-cli/src/logs.rs
mchf: well, whole ```std::os::*``` seems like a wild place ;-)

imobachgs (approved):
It still LGTM. Thank you!

PR #815, opened 2023-10-24
Fix installing the user selected patterns

lslezak:
## Problem

- The patterns selected by user are not installed because the package manager is reset before starting the installation and the selection is lot

## Solution

- Use the `PackagesProposal` module which remembers and restores the user selection
- Then we do not need the local `@user_patterns` attribute to hold the user selected patterns because it is now stored in the `PackagesProposal` module

## Testing

- Tested manually, the selected patterns were installed


imobachgs (approved):
Thanks!

PR #816, opened 2023-10-25
[web] Agama core/Section improvements

dgdavid:
## Problem

We've identified a use case where a section without title makes a lot of sense. The _introductory sections_ at the top of the page, which is the case of current Software page at the time of writing.

And probably more use cases will arise.

## Solution

To add support for sections without title and take the opportunity for improving the component a little bit and start adding support for a11n.

Additionally, the layout/Icon component has been improved a bit too, for outputting errors to the console instead of the UI and for including the icon name as data attribute of the rendered SVG for easily identify it.

## Side effects

As part of this PR `jsdom` dependency has been updated by using the [`overrides`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) _package.json_ key because of the use of  [`crypto.randomUUID`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) and a few tests using mocks for `windown.location` has to be rewritten since it's no longer possible to mock it anymore. See https://github.com/jsdom/jsdom/blob/master/Changelog.md#2100 and https://github.com/jsdom/jsdom/issues/3492 to know more.

In any case, `jest-environtment-jsdom` is about to update it in the next version, https://github.com/jestjs/jest/pull/13825#issuecomment-1564015010.  So, we're prepared already.

## Testing

- Adapted unit tests, adding new examples when needed.
- Tested manually too.

### DIFF
--- a/web/src/components/core/Section.jsx
+++ b/web/src/components/core/Section.jsx
@@ -142,7 +142,7 @@ export default function Section({
   children,
   "aria-label": ariaLabel
 }) {
-  const headerId = `${name ? name : Math.random()}-section-header`;
+  const headerId = `${name ? name : crypto.randomUUID()}-section-header`;
 
   if (!title && !ariaLabel) {
     console.error("The Section component must have either, a 'title' or an 'aria-label'");
--- a/web/src/components/core/Section.jsx
+++ b/web/src/components/core/Section.jsx
@@ -98,6 +98,11 @@ const SectionContent = ({ children }) => {
  *     <UsersSummary />
  *   </Section>
  *
+ * @example <caption>A section without title</caption>
+ *   <Section>
+ *     <UsersSummary />
+ *   </Section>
+ *
  * @example <caption>A section that allows navigating to a page</caption>
  *   <Section title="Users" icon="manage_accounts" path="/users">
  *     <UsersSummary />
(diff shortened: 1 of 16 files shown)


web/src/components/layout/Icon.jsx
```diff
@@ -149,10 +147,19 @@ const icons = {
  *
  */
 export default function Icon({ name, className = "", size = 32, ...otherProps }) {
+  if (!icons[name]) {
+    console.error(sprintf(_("Icon %s not found!"), name));
```
dgdavid, line 151: NOTE: I wonder if worth it adding a `process.env.NODE_ENV` check for not delivering these `console.error` in the production build.

```suggestion
    process.env.NODE_ENV !== "production" && console.error(sprintf(_("Icon %s not found!"), name));
```

or even to have a `logError` util for using it in every place we do a console.error

```js
const logError = (error) => {
  if (process.node.NODE_ENV !== "production") {
    console.error(error);
  }
};
```


```suggestion
    logError(sprintf(_("Icon %s not found!"), name));
```

Not sure if webpack automatically drop these blocks from the production build. Has to check.

https://webpack.js.org/guides/production/#specify-the-mode

web/src/components/core/Section.jsx
```diff
@@ -94,48 +94,81 @@ const SectionContent = ({ children }) => {
...
  *     <L10nSettings />
  *   </Section>
  *
- * @param {object} props
- * @param {string} [props.icon] - the name of the icon section, if any
- * @param {string} props.title - The title for the section
- * @param {string} props.path - The path where the section links to. If present, props.openDialog is ignored
- * @param {React.MouseEventHandler|undefined} [props.openDialog] - callback to be triggered
+ * @typedef { Object } SectionProps
+ * @property {string} [icon] - Name of the section icon. Not rendered if title not provided.
```
imobachgs, line 123: np: *if not title is provided* or something like that

imobachgs (approved):
It looks good. Thanks!

imobachgs:
Oh, it looks like you have a conflict to fix (originated after merging #831).

imobachgs (approved):

web/src/components/layout/Icon.jsx
dgdavid: I've checked by my own that blocks conditioned to `process.env.NODE_ENV !== 'production'` are not shipped in the production build. Which is great. So, creating an issue for wrap all `console.error`s

dgdavid: Discarded. See https://github.com/openSUSE/agama/issues/842

PR #818, opened 2023-10-25
Ensure that the software context is ready

imobachgs:
## Problem

The overview page is shown when the product is not selected. It is almost imperceptible for the human eye, but not for QA :smiley:

## Solution

* Make sure the software context is ready before loading the routes.

## Testing

- Tested manually

### DIFF
--- a/.github/workflows/ci-integration-tests.yml
+++ b/.github/workflows/ci-integration-tests.yml
@@ -146,7 +146,7 @@ jobs:
       run:  podman exec agama bash -c "cd /checkout/playwright && SKIP_LOGIN=true playwright test --trace on --project chromium"
 
     - name: Again show the D-Bus services log
-      run:  podman exec agama bash -c "journalctl | grep agama"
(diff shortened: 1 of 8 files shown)


joseivanlopez (approved):
LGTM

dgdavid (approved):
LGTM, thanks a lot for such a quick fix!

imobachgs:
Checking the problem in the CI, the problem turned out to be more complex. The comparison was correct (`null`: no product, `undefined`: not initialized). So, I decided to ensure the software context is ready before loading the rest of the application.

dgdavid:
> Checking the problem in the CI, the problem turned out to be more complex. The comparison was correct (`null`: no product, `undefined`: not initialized)

Ah, right. I forgot about that detail. We're using similar approach in other places. Good you realized about it. Thanks.

> So, I decided to ensure the software context is ready before loading the rest of the application.

Better, 👍

PR #819, opened 2023-10-26
Fix client initialization

imobachgs:
## Problem

The client is reconnected once even if it is not needed.

## Solution

Avoid the reconnection.

## Testing

- Update the unit tests
- Tested manual


imobachgs:
@joseivanlopez This stupid bug (authored by myself 🤦🏻‍♂️) might be affecting you.

dgdavid (approved):
LGTM

PR #821, opened 2023-10-26
Properly track the status

imobachgs:
## Problem

After the refactoring of the providers, it might happen that the `<LoadingEnvironment>` component does not detect the status changes.

## Solution

* Move the status changes detection to the `<App>` component, just like the phase tracking.
* Get rid of the `<LoadingEnvironment>` component. After all, it was just a `<Loading>` with status tracking.

## Testing

- Adapted the unit tests.
- Tested manually.


jreidinger (approved):
dropping code is always good

PR #823, opened 2023-10-27
Destination option for agama logs store

mchf:
## Problem

When storing logs during installation it would be nice to be able to define destination path

## Solution

added an option for ```agama logs store```

## Testing

- *Tested manually*

## Screenshots

TODO: add one


rust/agama-cli/src/logs.rs
```diff
@@ -19,6 +19,9 @@ pub enum LogsCommands {
         #[clap(long, short = 'v')]
         /// Verbose output
         verbose: bool,
+        #[clap(long, short = 'd')]
+        /// Destination path
```
mvidner, line 23: Is this a file name or a directory?
If a directory, what is the default file name?

rust/agama-cli/src/logs.rs
```diff
@@ -45,6 +50,32 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     }
 }
 
+// Whatewer passed in dest formed into an absolute path with archive name
+// if dest is none then a default is returned
+// if dest is directory then a default file name for the archive will be appended
+// TODO: if dest is path with a file name then it is used as is a name for resulting archive
```
jreidinger, line 56: well, I think this logs can be used in integration tests which now missing many logs ( it use only y2log and then traces from playwright. And for that I ideally need stable and unchanged path. So for me it would be ideal if I can pass complete path like `/tmp/agama.txz` and then point github to that file to upload it as artifact. Will it be possible?

mchf: it is plan ... see the todo comment

mchf: @jreidinger just added. Pls check if it fulfills your needs.

rust/agama-cli/src/logs.rs
```diff
@@ -45,6 +51,42 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
...
+// Whatewer passed in dest formed into an absolute path with archive name
+// if dest is none then a default is returned
+// if dest is directory then a default file name for the archive will be appended
+// if dest is path with a file name then it is used as is a name for resulting archive
+fn parse_dest(dest: Option<PathBuf>) -> Result<PathBuf, io::Error> {
+    let default = PathBuf::from(DEFAULT_RESULT);
+    let err = io::Error::new(io::ErrorKind::InvalidInput, "Invalid destination path");
+
+    match dest {
+        None => Ok(default),
```
imobachgs, line 63: What if `/tmp/agama_logs` is already a directory? What if tomorrow we decide to change the default value? I would run the same checks whether it is the default or a value from the user.

This way, we could drop the `match` and save some nesting:

```rust
let mut buff = dest.unwrap_or(PathBuf::from(DEFAULT_RESULT));
let path = buff.as_path();
```

rust/agama-cli/src/logs.rs
```diff
@@ -45,6 +51,42 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     }
 }
 
+// Whatewer passed in dest formed into an absolute path with archive name
+// if dest is none then a default is returned
```
imobachgs, line 55: For documenting function arguments, it is common [to use a list](https://doc.rust-lang.org/rust-by-example/meta/doc.html)

```
* `dest`: ...
```

rust/agama-cli/src/logs.rs
```diff
@@ -45,6 +51,42 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
...
+
+    match dest {
+        None => Ok(default),
+        Some(mut buff) => {
+            let path = buff.as_path();
+
+            // existing directory -> append an archive name
+            if path.is_dir() {
+                buff.push("agama-logs");
+                Ok(buff)
```
imobachgs, line 70: May I suggest returning early here and simplifying the rest?

```rust
   if path.is_dir() {
       buff.push("agama-logs");
       return Ok(buff)
    }

    if let Some(parent) = path.parent() {
        if parent.is_dir() {
            return Ok(buff)
        }
    }

    Err(err)
```

rust/agama-cli/src/logs.rs
imobachgs: BTW, why not `buff` instead of `buffer`? We are just saving two letters :-)

rust/agama-cli/src/logs.rs
```diff
@@ -45,6 +51,42 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
     }
 }
 
+// Whatewer passed in dest formed into an absolute path with archive name
+// if dest is none then a default is returned
+// if dest is directory then a default file name for the archive will be appended
+// if dest is path with a file name then it is used as is a name for resulting archive
+fn parse_dest(dest: Option<PathBuf>) -> Result<PathBuf, io::Error> {
```
imobachgs, line 58: You do not need to handle an `Option` if you set a default value for the `dest` field of `LogsCommands::Store`. But it is fine as it is.

rust/agama-cli/src/logs.rs
imobachgs: As you wrote, if `is_some_and` would be stable, it might be even simpler. But...

imobachgs (changes requested):
I would work a little on simplifying the `parse_dest` function. Beware, I have not tested my suggestions, so take them with a grain of salt 😉. Perhaps you should write a small unit test...

rust/agama-cli/src/logs.rs
mchf: if ```/tmp/agama_logs``` is already directory, then you will end up with ```/tmp/agama_logs.tar.bzip2``` archive.

mchf: regarding ```buff`` well ... just matter of habit as i have been using ```buff```` and ```dest``` since I was at school, so it is kind of convention for me ;-) It is even used in man pages of such tools like ```cp```,  ```rsync```, e.g. ```rsync``` defines its usage as ```rsync [OPTION...] SRC... [DEST]``` even in some options like ```--copy-dest=DIR```.

However, I don't have problem to extend it. It definitely is not stopper for me  ;-)

rust/agama-cli/src/logs.rs
mchf: tried to modify

rust/agama-cli/src/logs.rs
mchf: may be one day ;-)

rust/agama-cli/src/logs.rs
mchf: i did it this way because it was for free and kept the dispatch function a little bit simpler.

rust/agama-cli/src/logs.rs
```diff
@@ -19,6 +19,10 @@ pub enum LogsCommands {
         #[clap(long, short = 'v')]
         /// Verbose output
         verbose: bool,
+        #[clap(long, short = 'd')]
+        /// Path to destination directory. Optionally with the archive file name at the end.
+        /// An extension will be appended automatically depending on used compression.
+        dest: Option<PathBuf>,
```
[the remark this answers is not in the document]
mchf: commented later together with ```dest``` ;-)

rust/agama-cli/src/logs.rs
imobachgs: Not even for me, I was curious especially about the "buff" case :-)

imobachgs: About the `/tmp/agama_logs`, my point is that I would run *the same checks* in both cases. And in that case, you could save the `match` statement (too much nesting for my taste).

imobachgs (commented):
I have clarified what I meant about the checks of `/tmp/agama_logs`.

mchf:
> I have clarified what I meant about the checks of `/tmp/agama_logs`.

I got it ... I'm just testing it. Just used an opportunity and cleaned even another part.

imobachgs (approved):

PR #824, opened 2023-10-27
Add routes to network model

jcronenberg:
## Problem

No routes in network model

## Solution

Add a new `IpRoute` struct and add separate `routes4` and `routes6` to `IpConfig`

## Testing

- *Extended unit tests*
- *Tested manually*
- *Tested using the wicked migration*

### DIFF
--- a/rust/agama-dbus-server/src/network/nm/dbus.rs
+++ b/rust/agama-dbus-server/src/network/nm/dbus.rs
@@ -491,9 +491,10 @@ mod test {
     };
     use crate::network::{model::*, nm::dbus::ETHERNET_KEY};
     use agama_lib::network::types::SSID;
-    use std::{collections::HashMap, net::IpAddr};
+    use cidr::IpInet;
+    use std::{collections::HashMap, net::IpAddr, str::FromStr};
     use uuid::Uuid;
-    use zbus::zvariant::{self, OwnedValue, Value};
+    use zbus::zvariant::{self, Array, Dict, OwnedValue, Value};
 
     #[test]
     fn test_connection_from_dbus() {
@@ -508,6 +509,13 @@ mod test {
             ("prefix".to_string(), Value::new(24_u32)),
         ])];
 
+        let route_v4_data = vec![HashMap::from([
+            ("dest".to_string(), Value::new("192.168.0.0")),
+            ("prefix".to_string(), Value::new(24_u32)),
+            ("next-hop".to_string(), Value::new("192.168.0.1")),
+            ("metric".to_string(), Value::new(100_u32)),
+        ])];
+
         let ipv4_section = HashMap::from([
             ("method".to_string(), Value::new("auto").to_owned()),
             (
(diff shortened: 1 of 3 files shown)


rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -375,6 +405,34 @@ fn addresses_with_prefix_from_dbus(address_data: &OwnedValue) -> Option<Vec<IpIn
     Some(addresses)
 }
 
+fn routes_from_dbus(route_data: &OwnedValue) -> Option<Vec<IpRoute>> {
+    let route_data = route_data.downcast_ref::<zbus::zvariant::Array>()?;
+    let mut routes: Vec<IpRoute> = vec![];
+    for route in route_data.get() {
+        let route_dict = route.downcast_ref::<zvariant::Dict>()?;
+        let route_map =
+            <HashMap<String, zvariant::Value<'_>>>::try_from(route_dict.clone()).unwrap();
```
imobachgs, line 414: Instead of potentially panicking I would go for:

```suggestion
            <HashMap<String, zvariant::Value<'_>>>::try_from(route_dict.clone()).ok()?;
```

I know, we do not have a plan B if we cannot get the data, but it looks more polite to me :-)

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -520,6 +601,24 @@ mod test {
         );
         assert_eq!(ip_config.method4, Ipv4Method::Auto);
         assert_eq!(ip_config.method6, Ipv6Method::Auto);
+        assert!(ip_config.routes4.is_some());
```
imobachgs, line 604: You are already testing this in the following line, isn't it?

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -520,6 +601,24 @@ mod test {
...
+        assert_eq!(
+            ip_config.routes4,
+            Some(vec![IpRoute {
+                destination: IpInet::new("192.168.0.0".parse().unwrap(), 24_u8).unwrap(),
+                next_hop: Some(IpAddr::from_str("192.168.0.1").unwrap()),
+                metric: Some(100)
+            }])
+        );
+        assert!(ip_config.routes6.is_some());
```
imobachgs, line 613: Same than above.

imobachgs (changes requested):
In general, it looks good. I just added some minor comments. Please remember to update the changes file, if you don't mind.

jcronenberg:
> Please remember to update the changes file, if you don't mind.

Do you mean `agama-cli.changes`? Because this PR alone will have AFAICT no direct impact on `agama-cli`. This PR alone isn't touching the `settings` parts. I thought about implementing that as well, but thought this is something where you guys maybe want to decide e.g. the format of the settings etc. If you want we could discuss this as well and I'll implement something like that too, but that's up to you.

imobachgs:
Yes, no problem, we will take care of the `settings` part. However, as you can use `agama-lib` as a library, I would mention the changes in the `agama-cli.changes` file anyway. Something like:

```
* Add support for routing to the network model (gh#openSUSE/agama#824).
```

That would be enough. Otherwise, it LGTM (waiting for another tests run).

imobachgs (approved):

PR #825, opened 2023-10-27
Display the "InstallationFinished" again

imobachgs:
Similar to #821 we do not track the status properly. This is a quickfix for the current version, but a further investigation is due.

## Testing

- *Tested manually*

### DIFF
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -97,7 +97,7 @@ function App() {
     }
 
     if (phase === INSTALL) {
-      return <Installation />;
+      return <Installation status={status} />;
     }
 
     return <Outlet />;


web/src/components/core/Installation.jsx
```diff
@@ -19,19 +19,11 @@
...
 
-function Installation() {
-  const client = useInstallerClient();
-  const [status, setStatus] = useState(undefined);
-
-  useEffect(() =>
-    client.manager.onStatusChange(setStatus), [client.manager]
-  );
-
+function Installation({ status }) {
```
joseivanlopez, line 26: Who is passing the status to this component?

imobachgs: oh, I forgot to commit that change. Done.

joseivanlopez (approved):

PR #826, opened 2023-10-27
[Service] Delegate TPM-based encryption to YaST

ancorgs:
## Problem

Encryption with TPM-unlocking was implemented (in a pretty hacky way) directly at Agama.

## Solution

This PR goes together with https://github.com/yast/yast-storage-ng/pull/1363

Now `Y2Storage` takes care of most things.

As a bonus, the storage finisher is now called. We overlooked that in the past and it takes care of some relevant things like copying the multipath configuration to the target system.

### DIFF
--- a/service/lib/agama/dbus/y2dir/manager/modules/Package.rb
+++ b/service/lib/agama/dbus/y2dir/manager/modules/Package.rb
@@ -41,6 +41,13 @@ module Yast
 
     # Determines whether the package is available
     #
+    # @todo Perform a real D-Bus call.
+    def AvailableAll(_package_names)
+      true
+    end
+
+    # Determines whether the package is available
+    #
     # @see https://github.com/yast/yast-yast2/blob/b8cd178b7f341f6e3438782cb703f4a3ab0529ed/library/packages/src/modules/Package.rb#L121
     # @todo Perform a real D-Bus call.
     def Installed(package_name, target: nil)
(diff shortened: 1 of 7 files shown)


service/lib/agama/dbus/y2dir/manager/modules/Package.rb
```diff
@@ -39,6 +39,13 @@ def Available(_package_name)
       true
     end
 
+    # Determines whether the package is available
+    #
+    # @todo Perform a real D-Bus call.
+    def AvailableAll(_package_names)
+      true
```
joseivanlopez, line 46: Can we always assume true?

ancorgs: No. We can't. It's the same TODO we have for `Available` in the line above.

ancorgs: But not that you mention it, I will search for the corresponding Trello card/Github issue/Whatever and will reference it here.

ancorgs: Everything originated here (it says "_We need to add support for Package#Available and Package#Installed to D-Bus_):

ancorgs: Done. Now we have https://trello.com/c/oRcZ5P5k/251-implement-missing-software-methods

joseivanlopez (approved):

PR #836, opened 2023-11-01
fix spec for rubygem

jreidinger:
## Problem

The change to agama.yaml and move to web.yaml cause build failure of rubygem-agama

## Solution

Fix it

## Testing

- Tested osc build on modified source

### DIFF
--- a/service/package/gem2rpm.yml
+++ b/service/package/gem2rpm.yml
@@ -10,6 +10,7 @@
   install -m 0644 --target-directory=%{buildroot}%{_datadir}/dbus-1/agama-services %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/org.opensuse.Agama*.service
   install -D -m 0644 %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/agama.service %{buildroot}%{_unitdir}/agama.service
   install -D -m 0644 %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/agama-proxy-setup.service %{buildroot}%{_unitdir}/agama-proxy-setup.service
+  install --directory %{buildroot}/usr/lib/agama.d/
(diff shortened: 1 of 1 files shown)


service/package/gem2rpm.yml
```diff
@@ -10,6 +10,7 @@
   install -m 0644 --target-directory=%{buildroot}%{_datadir}/dbus-1/agama-services %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/org.opensuse.Agama*.service
   install -D -m 0644 %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/agama.service %{buildroot}%{_unitdir}/agama.service
   install -D -m 0644 %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/agama-proxy-setup.service %{buildroot}%{_unitdir}/agama-proxy-setup.service
+  install --directory %{buildroot}/usr/lib/agama.d/
```
imobachgs, line 13: Having a look to the [Filesystem Hierarchy Standard](https://es.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), I would expect configuration files to be under `/usr/share`. Something like `/usr/share/agama/conf.d`. WDYT?

imobachgs: However, I see `modprobe.d`, `pam.d`, etc. under `/usr/lib` too.

imobachgs (approved):

service/package/gem2rpm.yml
teclator: If I remember correctly we were told to put it under `/usr/share` already https://github.com/openSUSE/agama/issues/602#issuecomment-1582125164.

jreidinger: It is possible, but in that case we need to change path also in config file https://github.com/openSUSE/agama/blob/master/service/lib/agama/config_reader.rb#L46
@teclator as you are author of location how do you see it?

jreidinger: so lets merge this one now to fix  build failure and prepare new PR with location change.

PR #838, opened 2023-11-02
[web] Drop Section#openDialog prop

dgdavid:
## Problem

Some time ago, we introduced in the Agama core/Section component a feature to allow a modal dialog to be opened directly from the section title. However, as the installer has evolved, we have limited the title behavior to navigation only, which is more expected and less confusing. Thus, such a feature is actually not in use.

## Solution

Although additional features rarely cause problems, it's best to keep components as simple and clean as possible. Also, at this time, we believe it is more straightforward to use an explicit option inside the section to open a dialog than to hide it behind the header.

Therefore, this PR drops the mentioned feature. If we find a very compelling and valid use case, we can reintroduce it.

## Testing

- Unit tests adapted accordingly.

### DIFF
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Nov  2 07:38:22 UTC 2023 - David Diaz <dgonzalez@suse.com>
+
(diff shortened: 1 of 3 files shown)


imobachgs (approved):
Dropping code is usually good 😉

PR #840, opened 2023-11-02
Dev-server - display fullscreen Cockpit login form

lslezak:
## Problem

- When running the local development server the login form is not displayed in full screen, it is limited to the usual Agama main area.
- The problem is that in production the Cockpit server returns the login page instead of the Agama page when user is not authenticated.
- But in the development server we display the login page *inside* Agama (in an `iframe`), with the main page styling applied

devserver_login_orig

## Solution

- Remove the offending CSS property from the root node

devserver_login_fixed

## Notes

It is a bit hack, but as this is applied only in the development server (not in production) I think it is acceptable. Otherwise we would need to change the HTML structure and CSS quite a lot...

### DIFF
--- a/web/src/DevServerWrapper.jsx
+++ b/web/src/DevServerWrapper.jsx
@@ -115,6 +115,9 @@ export default function DevServerWrapper({ children }) {
   } else {
     // handle updating the iframe with the login form
     const onFrameLoad = () => {
+      // have a full screen login form
+      document.getElementById("root").style.maxInlineSize = "none";
+
       const passwordInput = iframeRef.current.contentWindow.document.getElementById(loginId);
       // reload the window so the manifests.js file referenced from the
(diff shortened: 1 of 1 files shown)


dgdavid (commented):
I wonder why do you want this change. I mean, I see no problem on having the Cockpit login form wrapped into the Agama layout at all. It's perfectly usable.

That said, I have no problem of going ahead with the proposed change as long as it only applies when working in the development mode, but I'd like to know the rationale of it.

dgdavid:
> but I'd like to know the rationale of it.

Ah, I guess if for having a similar experience than in production.

dgdavid (approved):

lslezak:
Yes, the reason is to have the very same look as in production. It's a bit confusing when you are presenting a demo using the development server...

PR #847, opened 2023-11-03
Merge master

joseivanlopez:
## Problem

There are too many new changes in *master* that are not present in the *registration* feature branch.

## Solution

Merge *master* branch into *registration*.

Note: only the last commit is the real meaninful change. The rest comes from *master*.

## Testing

- Adapted unit tests


dgdavid (approved):
Let's go!

PR #848, opened 2023-11-03
Replace `async_std` with `tokio`

imobachgs:
Regarding async/await, Rust does not endorse a runtime. There are different alternatives, but the most popular are [tokio](https://tokio.rs/), [async-std](https://github.com/async-rs/async-std) and [smol](https://github.com/smol-rs/smol). 

At the beginning of this project, we chose `async-std` because it was the default for [zbus](https://github.com/dbus2/zbus/). However, zbus offers [support for Tokio](https://docs.rs/zbus/latest/zbus/#special-tokio-support).

Taking into account that Tokio is the most used runtime in the Rust ecosystem, I am considering whether we should make the switch.

This pull request adapts the code to replace `async-std` with `tokio`. Sure, it deserves some discussion first, so I am opening it as a draft.

## Testing

In addition to the existing tests (including CI), I am building an  [agama-live](https://build.opensuse.org/package/show/home:IGonzalezSosa:branches:systemsmanagement:Agama:Staging/agama-cli) including the code in this `tokio` branch.

### DIFF
--- a/rust/agama-dbus-server/src/network/system.rs
+++ b/rust/agama-dbus-server/src/network/system.rs
@@ -1,6 +1,6 @@
 use crate::network::{dbus::Tree, model::Connection, Action, Adapter, NetworkState};
-use async_std::channel::{unbounded, Receiver, Sender};
 use std::error::Error;
+use tokio::sync::mpsc::{self, Receiver, Sender};
 
 /// Represents the network system using holding the state and setting up the D-Bus tree.
 pub struct NetworkSystem<T: Adapter> {
@@ -16,7 +16,7 @@ pub struct NetworkSystem<T: Adapter> {
 
 impl<T: Adapter> NetworkSystem<T> {
     pub fn new(conn: zbus::Connection, adapter: T) -> Self {
-        let (actions_tx, actions_rx) = unbounded();
+        let (actions_tx, actions_rx) = mpsc::channel(100);
         let tree = Tree::new(conn, actions_tx.clone());
         Self {
             state: NetworkState::default(),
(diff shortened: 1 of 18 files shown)


rust/agama-dbus-server/src/network/system.rs
```diff
@@ -16,7 +16,7 @@ pub struct NetworkSystem<T: Adapter> {
 
 impl<T: Adapter> NetworkSystem<T> {
     pub fn new(conn: zbus::Connection, adapter: T) -> Self {
-        let (actions_tx, actions_rx) = unbounded();
+        let (actions_tx, actions_rx) = mpsc::channel(100);
```
jreidinger, line 19: what is that `100`? Buffer size?

rust/agama-lib/src/progress.rs
```diff
@@ -63,14 +64,15 @@ pub struct Progress {
 
 impl Progress {
     pub async fn from_proxy(proxy: &crate::proxies::ProgressProxy<'_>) -> zbus::Result<Progress> {
-        let ((current_step, current_title), max_steps, finished) =
-            try_join3(proxy.current_step(), proxy.total_steps(), proxy.finished()).await?;
+        let (current_step, max_steps, finished) =
+            tokio::join!(proxy.current_step(), proxy.total_steps(), proxy.finished());
```
jreidinger, line 68: :+1: definitivelly nicer to have macro for it.

rust/agama-dbus-server/src/network/system.rs
imobachgs: Yes, sorry, I should have used an unbounded channel. I have adjusted the PR.

joseivanlopez (approved):
LGTM

jreidinger (approved):

PR #851, opened 2023-11-07
Generate and upload the product POT file to allow translating the product descriptions

lslezak:
## Problem

- The product descriptions in the `./products.d/*.yaml` files are not translated
- That looks bad as those texts are quite long and displayed in the very first Agama dialog

## Solution

- Extract the `description` texts from the YAML files
- Upload them to the [agama-weblate](https://github.com/openSUSE/agama-weblate) repository

## Testing

- Tested manually
- [Example GitHub Action run](https://github.com/lslezak/agama/actions/runs/6781746655/job/18432622534)
- [Uploaded POT file](https://github.com/openSUSE/agama-weblate/blob/ef597c21046e38b5a8d01ca46d98acc340b34cb9/products/products.pot)

## TODO

- [ ]  Set up a new [Weblate](https://l10n.opensuse.org/projects/agama/) translation project so the translators can actually translate the texts
- [ ] Implement merging the translations back to the original YAML files
- [ ] Use the translations in the backend/frontend
- [ ] Document the product translations (if somebody wants to define own product)

### DIFF
--- a/.github/workflows/product_translations/.gitignore
+++ b/.github/workflows/product_translations/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/.github/workflows/product_translations/package-lock.json
+++ b/.github/workflows/product_translations/package-lock.json
@@ -1,0 +1,218 @@
+{
+  "name": "product_translations",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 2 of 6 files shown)


imobachgs (approved):
LGTM. It might be interesting to document this further just in case someone is interested on creating its own product.

PR #853, opened 2023-11-08
Merge back the product translations

lslezak:
## Problem

- The product translations need to be merged back to the original YAML files
- We cannot store the translations in Agama because in theory there might be external (3rd party) product definitions so the translations needs to be included in the product YAML file
- Related to https://github.com/openSUSE/agama/pull/851

## Solution

- Read the translations from the PO files and and merge them back to the original YAML product files
- Implement a GitHub Action which regularly opens a pull request with updated translations (similar to the current solution with the web frontend translations)

## Notes

- Originally I wanted to use keys like `description-cs`, `description-de`,...  for the translations, but then the translations could be scattered over the whole file (the order is not guaranteed or forced)
- So use a new `translations` key with a reference to the original key:
  ```yaml
  description: "Original description ...."
  translations:
    description:
      cs: "Czech translation..."
      de: "German translation ..."
      es: "Spanish translation ..."
  ```
  This also allows translating some other keys if needed in the future.
- When changing the original description the translations need to be manually deleted because they would not match the original text anymore. If we forget that the next run of this GitHub Action will remove them anyway (or update with the new translations if they are already available) . So accidental invalid translations should not stay here for too long.

## Testing

- Tested manually with provided [Czech product translations](https://github.com/openSUSE/agama-weblate/blob/e8e766a746f8e16d30be97bd90dd4acce04e8a31/products/cs.po)
- See a testing [GitHub Action run](https://github.com/lslezak/agama/actions/runs/6795729562/job/18474362818)
- And the [PR created](https://github.com/lslezak/agama/pull/3/files) with the new translations

### DIFF
--- a/.github/workflows/product_translations/products_pot
+++ b/.github/workflows/product_translations/products_pot
@@ -51,7 +51,7 @@ class POFile {
    */
   timestamp() {
     const date = new Date();
-    return date.getUTCFullYear() + "-" + date.getUTCMonth() + "-" +
+    return date.getUTCFullYear() + "-" + (date.getUTCMonth() + 1) + "-" +
       date.getUTCDate() + " " + date.getUTCHours() + ":" + date.getUTCMinutes() +
       "+0000";
   }
--- a/products.d/ALP-Dolomite.yaml
+++ b/products.d/ALP-Dolomite.yaml
@@ -1,8 +1,14 @@
 id: ALP-Dolomite
 name: SUSE ALP Dolomite
+# ------------------------------------------------------------------------------
+# WARNING: When changing the product description delete the translations located
+# at the at translations/description key below to avoid using obsolete
+# translations!!
+# ------------------------------------------------------------------------------
 description: 'SUSE ALP Dolomite is a minimum immutable OS core, focused on
   security to provide the bare minimum to run workloads and services as
   containers or virtual machines.'
+translations:
 software:
   installation_repositories:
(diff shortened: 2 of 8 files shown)


.github/workflows/product_translations/products_pot
```diff
@@ -51,7 +51,7 @@ class POFile {
    */
   timestamp() {
     const date = new Date();
-    return date.getUTCFullYear() + "-" + date.getUTCMonth() + "-" +
+    return date.getUTCFullYear() + "-" + (date.getUTCMonth() + 1) + "-" +
```
lslezak, line 54: A fix for the previous PR, it turned out that the month (and only the month!!) starts from zero.

products.d/ALP-Dolomite.yaml
```diff
@@ -1,8 +1,14 @@
...
 name: SUSE ALP Dolomite
+# ------------------------------------------------------------------------------
+# WARNING: When changing the product description delete the translations located
+# at the at translations/description key below to avoid using obsolete
+# translations!!
+# ------------------------------------------------------------------------------
 description: 'SUSE ALP Dolomite is a minimum immutable OS core, focused on
   security to provide the bare minimum to run workloads and services as
   containers or virtual machines.'
+translations:
```
lslezak, line 11: Explicitly adding an empty `translations` key. Otherwise the script would add it at the end of the file and it is better to have the translations close to the original text.

imobachgs (approved):

PR #855, opened 2023-11-09
Use translated product descriptions in the backend

lslezak:
## Problem

- Return the translated product description from the backend
- Related to https://github.com/openSUSE/agama/pull/851

## Solution

- Find the translation matching the current locale
- When the translation is not available it returns the original (untranslated) description

## Notes

- For some reason when changing the language in the web UI it is not propagated to the Software service. But propagating to the Storage and Manager services works fine. :thinking: 
- This need some more debugging :lady_beetle: 

## Testing

- Added a new unit test
- Tested manually (with a manually patched Software service switched explicitly to the Czech language, see the note above)

### DIFF
--- a/service/lib/agama/dbus/software/manager.rb
+++ b/service/lib/agama/dbus/software/manager.rb
@@ -123,7 +123,7 @@ module Agama
 
         def available_base_products
           backend.products.map do |id, data|
-            [id, data["name"], { "description" => data["description"] }].freeze
+            [id, data["name"], { "description" => localized_description(data) }].freeze
           end
         end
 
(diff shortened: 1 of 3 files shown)


service/lib/agama/dbus/software/manager.rb
```diff
@@ -123,7 +123,7 @@ def initialize(backend, logger)
 
         def available_base_products
           backend.products.map do |id, data|
-            [id, data["name"], { "description" => data["description"] }].freeze
+            [id, data["name"], { "description" => localized_description(data) }].freeze
```
jreidinger, line 126: my wild guess here is that it won't raise signal when localization changes. So if UI cache list of base products, it will use it with old locale.

jreidinger (approved):

PR #866, opened 2023-11-14
[web] Update dependencies (2023-11-14)

dgdavid:
Another boring PR for keeping dependencies up-to-date via `npm update` and `npm install @package/name@latest`.  No more `--lockfile-version 2`  because [OBS already supports version 3](https://github.com/openSUSE/obs-service-node_modules/blob/2153192f0baf8453c12bacf7b872c0c786ed645f/node_modules.py#L342-L343) as we realized at https://github.com/openSUSE/agama/pull/843

Please, note that this time we manage to get rid of some dependencies which were there because cockpit-starter-kit (see https://github.com/openSUSE/agama/pull/127) but actually not needed anymore. In fact, they are no longer dependencies of starter-kit. Namely, 

* [core-js](https://github.com/zloirock/core-js)
  - As a weird side effect, few tests were _fixed_ to wait a bit for the expected content. Some black magic around _Promise.resolve_, I guess. Honestly, I didn't dig too further on it since waiting for the content is somehow _expected_.
*  [stdio](https://github.com/sgmonda/stdio)
* [sizzle](https://www.npmjs.com/package/sizzle)

Apart from these, [regenerator-runtime](https://github.com/facebook/regenerator) was dropped too because 

> It was added more than a year ago in the context of "Adapt React
> components to the Starter Kit". But probably is no longer needed since
> async functions, ES6 generators, and `yield` JavaScript operator have
> broad support among major browsers.

---

**Note for reviewers**: Even though it looks like everything works as expected, please give a try on your development environment and, if possible, different browsers before approving it.


imobachgs (approved):

dgdavid:
Now CI is failing, but I'm going ahead merging this because running

>  bundle exec yardoc --fail-on-warning

works locally. So, I'm assuming it is a temporary problem in CI.

Issue #859, opened 2023-11-13
Pattern descriptions are not translated


lslezak:
## Problem

After switching the display language to Spanish the pattern descriptions are still displayed in English.

![pattern-selector-spanish](https://github.com/openSUSE/agama/assets/907998/b57b9b6e-5169-4d83-ae38-6d5c558b3be1)

## Details

Libzypp loads the translations for the current language when the repository is refreshed. The translations are located in the `repo/*-susedata.<lang>.xml.gz` files. See examples in https://download.opensuse.org/tumbleweed/repo/oss/repodata/.

Libzypp loads only a single file with the current language, the rest is ignored to save time and the network bandwidth.

## Proposed Solution

Because the language might be switched after loading the initial repository we need to refresh the repositories again with the new language.

See related https://github.com/openSUSE/agama/pull/875.

*Note: For really using the new translations we probably also need to reload the libzypp pool from scratch. That means we need to restore also the selected products/patterns/packages back.*

ShatilKhan:
I can take this one @lslezak 
Please assign me

dgdavid:
Hi @ShatilKhan,

Sorry, but I think this one is rather difficult and should be fixed by the team. So, unless @lslezak has another view, I'd encourage you to pick another one if you want contribute.

Thanks anyway for your willing to help here.

lslezak:
I agree, this is not a trivial task, this needs a libzypp and yast-pkg-bindings expert. I'd recommend to start with something easier...

lslezak:
Implemented in PR #991, closing.

PR #875, opened 2023-11-16
Pass the changed language to the Software service and to libzypp

lslezak:
## Problem

When changing the display language in the web frontend the change is not propagated to the Software service. From the `y2log`:
```
2023-11-16 09:36:22 <1> localhost.localdomain(1708) [Ruby] agama/ui_locale.rb(change_locale):52 set yast language to es_ES
2023-11-16 09:36:22 <1> localhost.localdomain(1710) [Ruby] agama/ui_locale.rb(change_locale):52 set yast language to es_ES
```

So the [change_locale callback](https://github.com/openSUSE/agama/blob/e6e6e6bb6f270dfc1315700683df4a5ee00caa3c/service/lib/agama/ui_locale.rb#L52) was called only twice. According to the PID numbers only for the Manager and Storage service, not for the `Software` service.

## Solution

Just calling the `export` function before adding the language change callback in the Software service helped.

But actually I'm not sure whether this is the right fix. The same order is used in the Manager and Storage services where it works fine. Maybe there is some race condition and moving the code earlier/later just accidentally removed that collision and the problem is still there. :thinking:

(I updated the order also in the Storage and Manager service so it is consistent.)

With the "fix" the language change callback is correctly called triple times with all services (including the Software service).
```
2023-11-16 09:46:46 <1> localhost.localdomain(2765) [Ruby] agama/ui_locale.rb(change_locale):52 set yast language to es_ES
2023-11-16 09:46:46 <1> localhost.localdomain(2767) [Ruby] agama/ui_locale.rb(change_locale):52 set yast language to es_ES
...
2023-11-16 09:46:46 <1> localhost.localdomain(2771) [Ruby] agama/ui_locale.rb(change_locale):52 set yast language to es_ES
```

## Additional Fix

In the language change callback I added code which forwards the change also to libzypp. With this fix we can also display localized pattern names and descriptions. See the screenshot below.

But it still needs some improvement, libzypp uses the locale set at the repository refresh time, if the language is changed later then it is not reflected. It only works after manually changing the product to install which reloads the repository again.

But that's a different problem reported in #859. This fix at least proves that we can display also translated patterns, it's rather just a proof of concept.

## Testing

- Tested manually

## Screenshots

I tested the pattern selector with the Spanish language.

*Note: Some texts are not translated, but that is not Agama fault, the translations come from the package repository.*

patterns-spanish

### DIFF
--- a/service/lib/agama/dbus/manager_service.rb
+++ b/service/lib/agama/dbus/manager_service.rb
@@ -71,11 +71,11 @@ module Agama
       # @note The service runs its startup phase
       def start
         setup_cockpit
+        export
         # We need locale for data from users
         locale_client = Clients::Locale.new
         # TODO: test if we need to pass block with additional actions
         @ui_locale = UILocale.new(locale_client)
-        export
         manager.on_progress_change { dispatch } # make single thread more responsive
         manager.startup_phase
       end
--- a/service/package/rubygem-agama.changes
+++ b/service/package/rubygem-agama.changes
@@ -2,7 +2,7 @@
 Thu Nov 16 16:27:37 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
 
 - Software service - correctly change the locale, pass the changed
-  locale to libzypp
+  locale to libzypp (gh#openSUSE/agama#875).
 
 -------------------------------------------------------------------
 Wed Nov 15 12:31:10 UTC 2023 - José Iván López González <jlopez@suse.com>
--- a/service/package/rubygem-agama.changes
+++ b/service/package/rubygem-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Nov 16 16:27:37 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Software service - correctly change the locale, pass the changed
+  locale to libzypp
+
+-------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


service/lib/agama/dbus/manager_service.rb
```diff
@@ -71,11 +71,11 @@ def initialize(config, logger = nil)
       # @note The service runs its startup phase
       def start
         setup_cockpit
+        export
         # We need locale for data from users
         locale_client = Clients::Locale.new
         # TODO: test if we need to pass block with additional actions
         @ui_locale = UILocale.new(locale_client)
-        export
```
jreidinger, line 78: interesting that it helps. I actually would expect issues in other way. When you export service and then localize it, so I see small window when you can get unlocalized texts.

service/package/rubygem-agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Nov 16 16:27:37 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Software service - correctly change the locale, pass the changed
+  locale to libzypp
```
imobachgs, line 5:
```suggestion
  locale to libzypp (gh#openSUSE/agama#875).
```

imobachgs (approved):

PR #882, opened 2023-11-21
[DRAFT] [RFC] View Transitions API

dgdavid:
## :warning: This PR is only for illustration purposes. Merge not intended. :warning:

Rationale: to be written.

> Page transitions **not only look great, they also communicate direction of flow**, and make it clear which elements are related from page to page. — https://developer.chrome.com/docs/web-platform/view-transitions/#why-do-we-need-this-feature

https://github.com/openSUSE/agama/assets/1691872/b21dbcba-01f1-4836-8286-981c735ff010

Interesting links, 

* https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API
* https://developer.chrome.com/docs/web-platform/view-transitions/
* https://github.com/WICG/view-transitions/blob/main/explainer.md
* https://drafts.csswg.org/css-view-transitions-1/ 

Other interesting links,

* https://github.com/remix-run/react-router/discussions/10276
* https://bejamas.io/blog/what-is-view-transitions-api/#view-transition-and-accessibility
* https://developer.chrome.com/blog/view-transitions-case-studies/


dgdavid:
Although the idea of start using the View Transitions API still in the `to do` list, this PR can be closed by now.

PR #884, opened 2023-11-22
split a testing_using_container script out of backend-for-testing.md

mvidner:
## Problem

We have some documents

- `rust/agama-cli/doc/backend-for-testing.md`
- `doc/testing_using_container.md` (are they duplicate?)

and a script
- `service/run_tests_in_container.sh ` which is only enough for `rake test:unit`

but they either don't do what I want (below) or require cop-pasting

## Solution

Make a script that I can use for testing the entire Agama frontend and backend, in a container

## Testing

- Run the script
- xdg-open https://localhost:9090/agama
- click around

## Screenshots

N/A

### DIFF
--- a/testing_using_container.sh
+++ b/testing_using_container.sh
@@ -1,23 +1,33 @@
+#!/bin/bash
+set -x
+set -eu
+
 # https://build.opensuse.org/package/show/YaST:Head:Containers/agama-testing
 CIMAGE=registry.opensuse.org/yast/head/containers/containers_tumbleweed/opensuse/agama-testing:latest
 # rename this if you test multiple things
 CNAME=agama
-# the '?' here will report a shell error
-# if you accidentally paste a command without setting the variable first
-echo ${CNAME?}
 
 test -f service/agama.gemspec || echo "You should run this from a checkout of agama"
 
 # destroy the previous instance, can fail if there is no previous instance
-podman stop ${CNAME?}
-podman rm ${CNAME?}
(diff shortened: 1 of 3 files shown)


testing_using_container.sh
```diff
@@ -0,0 +1,46 @@
...
+podman rm ${CNAME?} || : no problem if there was nothing to remove
+
+# Update our image
+podman pull ${CIMAGE?}
+
+# Optionally use a git version of ruby-dbus
+# because Agama pushes its limits so it's better
+# to set up for easy testing
+MORE_VOLUMES=()
+if [ "${WITH_RUBY_DBUS-}" = 1 ]; then
```
jreidinger, line 23: Maybe document that env variable at the top of script together with some documentation what script does?

testing_using_container.sh
```diff
@@ -0,0 +1,46 @@
...
+
+${CEXEC?} "cd /checkout && ./setup.sh"
+
+# Now the CLI is in the same repo, just symlink it
+${CEXEC?} "ln -sfv /checkout/./rust/target/debug/agama /usr/bin/agama"
+
+# Manually start cockpit as socket activation does not work with port forwarding
+${CEXEC?} "systemctl start cockpit"
+
+# Optional: Interactive shell in the container
```
jreidinger, line 45: why Optional?

mvidner: good catch, that word is a leftover from the old file

PR #885, opened 2023-11-22
Add bonding support

teclator:
This PR adds support for configuring a Bond connection using Agama auto-installation.

## The model

Initially, we thought each connection struct would take ownership of its ports (basically, other connections). However, after some discussion with the networking team, we agreed to keep a "reference" in the opposite direction (from the port to the controller). At this point, each connection that belongs to a bond keeps its parent UUID.

## Configuring a port

When using the auto-installation you can define the port with something like this:

```rust
      {
        id: 'bond0',
        bond: {
            ports: ['eth0', 'eth1'],
            mode: 'active-backup',
            options: "primary=eth1"

        }
      },
      {
        id: 'Wired Adapter 1',
        interface: 'eth1'
      }
```

In NetworkManager, it is mandatory to specify an interface name. However, Agama will use the `id` if you do not define the `interface`. About the ports, it is not required to define them (in the example above, `eth0` is missing). Agama will create a connection automatically. And, again, you can use the `id` or the `interface` to link both connections.

## Improving the D-Bus interface

As the list of ports is not part of the `BondConnection` struct, it is not straightforward to expose that information in the D-Bus interface. After all, the code that exposes each connection over D-Bus only had access to the information of its specific connection (actually, it owns a clone of the connection).

For that reason, starting with the [Bond](https://github.com/openSUSE/agama/blob/add_bonding_support/rust/agama-dbus-server/src/network/dbus/interfaces.rs#L312) interface, we implemented a better mechanism (based on message passing) to interact with the data model. We will open a separate PR to adapt the rest of interfaces (with more information about the problems it solves).

## What's next?

* #918
* #924 
* Fix a ton of issues detected by Clippy.

## References

* @cfconrad started to work on bonding support on #783.

## Testing

* Tested manually (see https://github.com/teclator/agama/pull/1).
* Added some unit and integration tests.


rust/agama-dbus-server/src/network/model.rs
```diff
@@ -81,6 +90,52 @@ impl NetworkState {
...
+    ///
+    /// It uses the `id` to decide which connection to update.
+    ///
+    /// Additionally, it registers the connection to be removed when the changes are applied.
+    pub fn update_controller_connection(
+        &mut self,
+        mut conn: Connection,
+        settings: HashMap<String, ControllerSettings>,
+    ) -> Result<(), NetworkStateError> {
+        // let Some(old_conn) = self.get_connection_mut(conn.id()) else {
```
imobachgs, line 103: leftover

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -273,11 +335,18 @@ impl Connection {
         self.base_mut().interface = interface.to_string()
     }
 
+    pub fn controller(&self) -> &str {
+        self.base().controller.as_str()
+    }
+
+    pub fn set_controller(&mut self, controller: &str) {
```
imobachgs, line 342: What is a "controller"? It would be nice to put an example.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -55,6 +57,13 @@ impl NetworkState {
         self.connections.iter_mut().find(|c| c.id() == id)
     }
 
+    /// Get connection by name
+    ///
+    /// * `name`: connection name
```
imobachgs, line 62: Wrong name

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -81,6 +90,52 @@ impl NetworkState {
...
+        // let Some(old_conn) = self.get_connection_mut(conn.id()) else {
+        //     return Err(NetworkStateError::UnknownConnection(conn.id().to_string()));
+        // };
+        //
+        let mut new_ports = vec![];
+        let controller = conn.clone();
+
+        if let Connection::Bond(ref mut bond) = conn {
+            if let Some(ControllerSettings::Options(opts)) = settings.get("options") {
+                bond.bond.options = BondOptions::try_from(opts.clone()).unwrap()
```
imobachgs, line 112: I would try to avoid `bond.bond`. Something like `bond_conn` would be better (OK, I am pretty bad at naming things :wink:).

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -81,6 +90,52 @@ impl NetworkState {
...
+            }
+
+            if let Some(ControllerSettings::Ports(ports)) = settings.get("ports") {
+                for port in ports.iter() {
+                    new_ports.push(
+                        if let Some(new_port) =
+                            bond.bond.ports.iter().find(|c| c.interface() == port)
```
imobachgs, line 119: Perhaps we could move this stuff to a `BondConfig` function. With a good name, it would be more expressive.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -81,6 +90,52 @@ impl NetworkState {
...
+        let controller = conn.clone();
+
+        if let Connection::Bond(ref mut bond) = conn {
+            if let Some(ControllerSettings::Options(opts)) = settings.get("options") {
+                bond.bond.options = BondOptions::try_from(opts.clone()).unwrap()
+            }
+
+            if let Some(ControllerSettings::Ports(ports)) = settings.get("ports") {
+                for port in ports.iter() {
+                    new_ports.push(
```
imobachgs, line 117: I would reorganize this a bit to avoid a full `if-else` clause as a parameter for `new_ports`. It is hard to read (at least for me, I know, I am getting old).

rust/agama-dbus-server/src/network/model.rs
imobachgs: Actually, you could follow an approach like this (I know, it does not compile, it is just an idea):

```rust
new_port = bond
  .find_port(port)
  .cloned()
  .unwrap_or_else(|| {
    // ... here the code to build the port
  });
```

In my opinion, it expresses what you want better. But perhaps it is just me.

About building the port, your code is perfectly fine. But perhaps we could try something like [the builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html):

```rust
ConnectionBuilder::new(port.to_string(), DeviceType::Ethernet)
  .with_interface(port),
  .with_controller(controller.id())
  .build()
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -480,6 +550,47 @@ pub struct LoopbackConnection {
...
 #[derive(Debug, Default, PartialEq, Clone)]
+pub struct BondConnection {
+    pub base: BaseConnection,
+    pub bond: BondConfig,
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct BondOptions(pub HashMap<String, String>);
+
+impl TryFrom<String> for BondOptions {
```
imobachgs, line 561: Perhaps it would be better to implement [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) instead. Does it fit your use case? IMHO it is more versatile and you do not need to consume the `String`.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -480,6 +550,47 @@ pub struct LoopbackConnection {
...
+    type Error = NetworkStateError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let mut options = HashMap::new();
+
+        for opt in value.split_whitespace() {
+            let opt_word: Vec<&str> = opt.trim().split('=').collect();
```
imobachgs, line 568: You might be interested in [split_once](https://doc.rust-lang.org/std/primitive.str.html#method.split_once). Anyway, what does happen if there is no `=`?

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -480,6 +550,47 @@ pub struct LoopbackConnection {
...
+        }
+
+        Ok(BondOptions(options))
+    }
+}
+
+impl fmt::Display for BondOptions {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut opts = vec![];
+        for (key, value) in &self.0 {
```
imobachgs, line 579: As an exercise, could you try a `map`? IMHO, the best thing of using combinators is that they express an intention.

rust/agama-dbus-server/src/network/nm/client.rs
```diff
@@ -71,17 +75,27 @@ impl<'a> NetworkManagerClient<'a> {
...
             let proxy = ConnectionProxy::builder(&self.connection)
                 .path(path.as_str())?
                 .build()
                 .await?;
             let settings = proxy.get_settings().await?;
             // TODO: log an error if a connection is not found
-            if let Some(connection) = connection_from_dbus(settings) {
-                connections.push(connection);
+
+            if let Some(connection) = connection_from_dbus(settings.clone()) {
```
imobachgs, line 87: Note for myself: why does `connection_from_dbus` consume the argument?

rust/agama-dbus-server/src/network/system.rs
```diff
@@ -73,6 +73,16 @@ impl<T: Adapter> NetworkSystem<T> {
             Action::UpdateConnection(conn) => {
                 self.state.update_connection(conn)?;
             }
+            Action::UpdateControllerConnection(conn, settings, tx) => {
+                let id = &conn.clone();
```
imobachgs, line 77: Why do you need to clone the conn? I am not saying that it is wrong, but looks weird.

rust/agama-dbus-server/src/network/system.rs
```diff
@@ -73,6 +73,16 @@ impl<T: Adapter> NetworkSystem<T> {
...
+                let id = id.id();
+                self.state.update_controller_connection(conn, settings)?;
```
imobachgs, line 79: This method could return a reference to the updated connection.

imobachgs (commented):
Thanks! It is a lot of work.

I have not reviewed the code 100%, but it goes in the right direction. I have some suggestions (do not fear combinators :wink:) but we can discuss them in a pair programming session.

rust/agama-dbus-server/src/network/system.rs
imobachgs, line 77: You can avoid that `clone` (which looks weird) by doing:

```rust
let id = conn.id().to_owned();
// ...
if let Some(conn) = self.state.get_connection(&id) {
//                                            ^

File: rust/agama-dbus-server/src/network/model.rs [hunk 13]
```diff
@@ -81,6 +90,52 @@ impl NetworkState {
...
     }
 
+    /// Updates a controller connection with a new one.
+    ///
+    /// It uses the `id` to decide which connection to update.
+    ///
+    /// Additionally, it registers the connection to be removed when the changes are applied.
+    pub fn update_controller_connection(
+        &mut self,
+        mut conn: Connection,
```
imobachgs, line 100: missing docs

rust/agama-dbus-server/src/network/model.rs
teclator: It is not used at the end so, removing it.

rust/agama-dbus-server/src/network/model.rs
teclator: I'm assuming there is always a `=`, but we could be more robust.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -16,6 +16,8 @@ use thiserror::Error;
 use uuid::Uuid;
 use zbus::zvariant::Value;
 
+use super::dbus::ControllerSettings;
```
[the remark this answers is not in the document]
teclator: Yep, it is specific to DBUS but mainly because of the need to pass just one argument (the ports or the options) It probably could be better written and definitely it smells badly.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -480,6 +583,76 @@ pub struct LoopbackConnection {
...
+impl fmt::Display for BondOptions {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let opts = &self
+            .0
+            .iter()
+            .map(|(key, value)| format!("{key}={value}"))
+            .collect::<Vec<_>>();
+
+        write!(f, "{}", opts.join(" "))
+    }
```
cfconrad, line 639: Not enough rusty?

```suggestion
    let mut spacer = "";
    for (key, value) in &self.0 {
        write!(f, "{spacer}{key}={value}");
        spacer = " ";
    }
    Ok(())
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -81,6 +81,44 @@ impl NetworkState {
...
+            if let Some(ControllerConfig::Ports(ports)) = settings.get("ports") {
+                let new_ports: Vec<_> = ports
+                    .iter()
+                    .map(|port| {
+                        bond.find_port(port).cloned().unwrap_or_else(|| {
+                            ConnectionBuilder::new(port)
+                                .with_type(DeviceType::Ethernet)
+                                .with_interface(port)
+                                .with_controller(controller.id())
+                                .build()
```
cfconrad, line 110: I'm struggle with this, when I think of a scenario which have a other DeviceType then Ethernet, e.g.:
```
    eth1 <-l-  eth1.11 -m-+-> bond0
                          |
                  eth2 -m-+  
```
How do you think we are going to implement this?

rust/agama-dbus-server/src/network/nm/client.rs
```diff
@@ -71,34 +76,94 @@ impl<'a> NetworkManagerClient<'a> {
...
+
+        if let Connection::Bond(bond) = conn {
+            let _ = bond.bond.ports.iter().map(|port| {
+                self.add_or_update_port_connection(
+                    port,
+                    bond.base.interface.to_string(),
+                    "bond".to_string(),
+                )
+            });
+        }
```
cfconrad, line 166: Shouldn't this be the same as as simply calling `add_or_update_connection()` ?

rust/agama-dbus-server/src/network/nm/client.rs
```diff
@@ -71,34 +76,94 @@ impl<'a> NetworkManagerClient<'a> {
...
+                    controllers
+                        .entry(controller)
+                        .or_insert_with(Vec::new)
+                        .push(connection)
+                } else {
+                    connections.push(connection);
+                }
             }
         }
+
```
cfconrad, line 99: Maybe I miss something, but for me it looks like we are putting all connections which do not have a `controller` (none port connections) into a vector which is in that hashmap `controllers`... the other connections into `connections`. But we forget to link them.

rust/agama-lib/share/profile.schema.json
```diff
@@ -105,6 +105,24 @@
...
+                "description": "Bonding configuration",
+                "additionalProperties": false,
+                "properties": {
+                  "options": {
+                    "type": "string"
+                  },
+                  "ports": {
+                    "type": "array",
+                    "items": {
+                      "description": "A list of the ports to be bonded",
```
cfconrad, line 119: Maybe we need to be more precise here:
1) Is it a connection name or interface name
2) What happen if it is a connection name, but no connection with this name exists (defaults)

cfconrad (commented):
I'm not convinced to store the `ports` inside the controller device. 
Of course it is nice in the first place, when we have a simple configuration where we just wanna add some ports to a bond. But if theses `ports` it self are something different (wireless, vlan, bridge) they need a connection definition anyway.

If we do not have a connection of a `port` and we are going to auto create one, I think we should do it next to the user input. And internally just work with a list of `connections`, which have only **one** link to a lower/parent device (like VLAN has) and **one** link to a upper/controller (like a bridge/bond ports have).

rust/agama-dbus-server/src/network/model.rs
teclator: We should obtain the device type from the system.

rust/agama-lib/src/network/types.rs
```diff
@@ -24,11 +31,96 @@ impl From<SSID> for Vec<u8> {
...
 pub enum DeviceType {
     Loopback = 0,
     Ethernet = 1,
     Wireless = 2,
+    Bond = 3,
+}
+/// Bond mode
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum BondMode {
+    #[serde(rename = "balance-rr")]
```
cfconrad, line 44: Just a hint. As I'm not sure what makes the more sense/ is better to read!

In our xml parsing part we used [serde_with](https://crates.io/crates/serde_with) and [strum_macros](https://crates.io/crates/strum_macros) crates to achieve similar things.
https://github.com/jcronenberg/agama/blob/523ebe9abf4055c67952d2d50e6f1fbe734044b5/rust/agama-migrate-wicked/src/interface.rs#L167

rust/agama-dbus-server/src/network/model.rs
cfconrad: what is, if it's also a virtual interface like:
```
vlan9 -m-+
         |-> bond0
 eth2 -m-+  
```
And we just about to create them?

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -63,7 +65,21 @@ impl NetworkState {
...
+                let id = conn.id().to_string();
+                // FIXME: We should implement a trait Controller
+                // so those that acts as a controller offer the same API
+                // and we do not need to access "bond.bond".
+                let ports = bond.bond.ports.clone();
+                self.update_controller_ports(id, ports)?;
+            }
+            _ => {}
+        };
+
```
cfconrad, line 80: What if someone update a Connection which is a port. Do we need to update the Controller as well (if it is Bond). Something like: 
```
if let Some(ctrl_id) = old_conn.controller() {
    if let Some(conn) = self.get_connection_mut(ctrl_id) {
        if let Connection::Bond(bond) = conn {
            let id = old_conn.id().to_string();
            if ! bond.bond.ports.contains(&id) {
                bond.bond.ports.push(id);
            }
        }
    }
}
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -78,6 +94,37 @@ impl NetworkState {
...
+        match &old_conn {
+            Connection::Bond(bond) => {
+                let id = old_conn.id().to_string();
+                let ports = bond.bond.ports.clone();
+                self.update_controller_ports(id, ports)?;
+            }
+            _ => {}
+        };
+
```
cfconrad, line 106: same as here: https://github.com/openSUSE/agama/pull/885/files#r1417092019

rust/agama-dbus-server/src/network/model.rs
cfconrad: In general I think we should "declare", what need to be there first. The Bond connection or the Port connection. And depending on this, maybe we should throw an error if the corresponding reference doesn't exists.

I would prefere not having `bond.bond.ports` but replacing it with a method which creates a iterator on the fly with all corresponding ports.

imobachgs: Sure! Please have in mind that the PR is unfinished and not ready for review yet, although early comments are appreciated. There are many things, especially error handling, that we need to implement/improve. And, of course, we will replace the `bond.bond` thingie with something better as it hurts my eyes too :-)

imobachgs: > What if someone update a Connection which is a port. Do we need to update the Controller as well (if it is Bond). Something like:

Good point, we need to handle that case too. It is not possible with the D-Bus API (you always set the ports through the Bond), but it is something that we need to tackle.

rust/agama-dbus-server/src/network/model.rs
imobachgs: Well, I would say that it is more a Ruby-like approach :-) Both of them look expressive enough.

rust/agama-dbus-server/src/network/model.rs
imobachgs: We decided to simplify things and only keep the reference in the port. And, btw, we are using the UUID so if we rename any interface/connection we do not need to update the links.

rust/agama-dbus-server/src/network/action.rs
```diff
@@ -9,6 +16,16 @@ use agama_lib::network::types::DeviceType;
...
     /// Add a new connection with the given name and type.
     AddConnection(String, DeviceType),
+    /// Gets a connection
+    GetConnection(Uuid, Responder<Option<Connection>>),
+    /// Gets a controller connection
+    GetController(
+        Uuid,
+        Responder<Result<ControllerConnection, NetworkStateError>>,
+    ),
+    /// Sets a controller ports. It uses the Uuid of the controller and the IDs or interface names
```
jcronenberg, line 26:
```suggestion
    /// Sets a controller's ports. It uses the Uuid of the controller and the IDs or interface names
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -43,27 +45,50 @@ impl NetworkState {
 
     /// Get connection by UUID
     ///
-    /// * `uuid`: connection UUID
+    /// * `id`: connection UUID
```
jcronenberg, line 48:
```suggestion
    /// * `uuid`: connection UUID
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -43,27 +45,50 @@ impl NetworkState {
...
+    pub fn get_connection_by_uuid(&self, uuid: Uuid) -> Option<&Connection> {
+        self.connections.iter().find(|c| c.uuid() == uuid)
+    }
+
+    /// Get connection by interface
+    ///
+    /// * `id`: connection interface name
```
jcronenberg, line 55:
```suggestion
    /// * `name`: connection interface name
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -92,14 +117,50 @@ impl NetworkState {
         conn.remove();
         Ok(())
     }
+
+    /// Sets a controller ports.
```
jcronenberg, line 121:
```suggestion
    /// Sets a controller's ports.
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -269,19 +393,34 @@ impl Connection {
...
+    pub fn interface(&self) -> Option<&str> {
+        self.base().interface.as_deref()
     }
 
     pub fn set_interface(&mut self, interface: &str) {
-        self.base_mut().interface = interface.to_string()
+        self.base_mut().interface = Some(interface.to_string())
+    }
+
+    /// Ports controller name, e.g.: bond0, br0
```
jcronenberg, line 404:
```suggestion
    /// A port's controller name, e.g.: bond0, br0
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -269,19 +393,34 @@ impl Connection {
...
+    pub fn controller(&self) -> Option<Uuid> {
+        self.base().controller
+    }
+
+    /// Sets the ports controller.
```
jcronenberg, line 409:
```suggestion
    /// Sets the port's controller.
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -269,19 +393,34 @@ impl Connection {
...
+    ///
+    /// `controller`: Name of the controller (Bridge, Bond, Team), e.g.: bond0.
```
jcronenberg, line 411:
```suggestion
    /// `controller`: Uuid of the controller (Bridge, Bond, Team), e.g.: bond0.
```

rust/agama-lib/src/network/client.rs
```diff
@@ -100,6 +139,22 @@ impl<'a> NetworkClient<'a> {
         })
     }
 
+    /// Returns the [bond settings][BondSettings] for the given connection
+    ///
+    ///  * `path`: the connections path to get the wireless config from
```
jcronenberg, line 144:
```suggestion
    ///  * `path`: the connection's path to get the wireless config from
```

rust/agama-lib/src/network/client.rs
```diff
@@ -249,6 +309,29 @@ impl<'a> NetworkClient<'a> {
         Ok(())
     }
 
+    /// Updates the bond settings for network connection.
```
jcronenberg, line 312:
```suggestion
    /// Updates the bond settings for a network connection.
```

rust/agama-lib/src/network/proxies.rs
```diff
@@ -3,6 +3,30 @@
 //! This code was generated by `zbus-xmlgen` `3.1.0` from DBus introspection data.`.
 use zbus::dbus_proxy;
 
+#[dbus_proxy(
+    interface = "org.opensuse.Agama1.Network.Devices",
+    default_service = "org.opensuse.Agama1",
+    default_path = "/org/opensuse/Agama1/Network/devices"
+)]
+trait Devices {
+    /// GetConnections method
```
jcronenberg, line 12:
```suggestion
    /// GetDevices method
```

rust/agama-lib/src/network/proxies.rs
```diff
@@ -155,4 +181,30 @@ trait Match {
...
+    #[dbus_proxy(property)]
+    fn set_mode(&self, value: &str) -> zbus::Result<()>;
+
+    /// ports property
+    #[dbus_proxy(property)]
+    fn options(&self) -> zbus::Result<String>;
+    #[dbus_proxy(property)]
+    fn set_options(&self, value: &str) -> zbus::Result<()>;
+
+    /// ports property
```
jcronenberg, line 205:
```suggestion
    /// Ports property
    #[dbus_proxy(property)]
    fn options(&self) -> zbus::Result<String>;
    #[dbus_proxy(property)]
    fn set_options(&self, value: &str) -> zbus::Result<()>;

    /// Ports property
```
For consistency

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -92,14 +117,50 @@ impl NetworkState {
         conn.remove();
         Ok(())
     }
+
+    /// Sets a controller ports.
+    ///
+    /// If the connection is not a controller, returns an error.
+    ///
+    /// * `uuid`: controller UUID.
```
jcronenberg, line 125:
```suggestion
    /// * `controller`: controller to set ports on.
```

jcronenberg (commented):
Just had a quick look over mostly just the docs. Sorry for being nitpicky with the grammar and documentation :slightly_smiling_face:

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -22,31 +23,51 @@ const DUMMY_KEY: &str = "dummy";
...
-        let ethernet_config =
-            HashMap::from([("assigned-mac-address", Value::new(conn.mac_address()))]);
-        result.insert(ETHERNET_KEY, ethernet_config);
-    } else if let Connection::Wireless(wireless) = conn {
-        connection_dbus.insert("type", WIRELESS_KEY.into());
-        let wireless_dbus = wireless_config_to_dbus(wireless);
-        for (k, v) in wireless_dbus {
-            result.insert(k, v);
+    match &conn {
+        Connection::Ethernet(_) | Connection::Loopback(_) => {
```
cfconrad, line 49: I think `is_ethernet()` was the hook to identify connections which need to set the customized macaddr via `802-3-ethernet.assigned-mac-address`. AFAIK Dummy was one of them and Bond should be one.

If we do it this way. We would need to set mac in the corresponding match block of `Connection::Bond` and `Connection::Dummy`.
Or we extract the writing of `802-3-ethernet.assigned-mac-address` out of this match block.

rust/agama-dbus-server/src/network/model.rs
cfconrad: I thought the later would have less implicit memory effort, as it writes directly to the stream. But there is not critical path and I don't have hard feelings here :-)

cfconrad (approved):
Looks good to me, thx for the effort!

imobachgs (approved):
It works in our tests 🎆 (creating and updating the bond and its ports).

PR #887, opened 2023-11-23
Autosave logs at the end of installation

mchf:
## Problem

Agama should automatically store logs at the en of installation

## Solution

Call "agama logs store" at the end of installation

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -205,7 +205,7 @@ module Agama
 
     # Collects the logs and stores them into an archive
     #
-    # @path [String] directory where to store logs
+    # @param path [String] directory where to store logs
     #
     # @return [String] path to created archive
     def collect_logs(path: nil)
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -208,7 +208,7 @@ module Agama
     # @path [String] directory where to store logs
     #
     # @return [String] path to created archive
-    def collect_logs(path)
+    def collect_logs(path: nil)
       opt = "-d #{path}" if !path.nil? && !path.empty?
 
       %x(agama logs store #{opt}).strip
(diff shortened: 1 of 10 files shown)


service/lib/agama/manager.rb
```diff
@@ -205,18 +205,22 @@ def valid?
...
-    # @param user [String] local username who will own archive
+    # @param path [String] directory where to store logs
+    #
     # @return [String] path to created archive
-    def collect_logs(user)
-      output = Yast::Execute.locally!("save_y2logs", stderr: :capture)
-      path = output[/^.* (\/tmp\/y2log-\S*)/, 1]
-      Yast::Execute.locally!("chown", "#{user}:", path)
+    def collect_logs(path: nil)
+      opt = "-d #{path}" if !path.nil? && !path.empty?
```
imobachgs, line 212: np: would not be more readable `unless path.nil? || path.empty?`? Or even `unless path.to_s.empty?`.

service/lib/agama/manager.rb
```diff
@@ -205,18 +205,22 @@ def valid?
...
 
-      path
+      `agama logs store #{opt}`.strip
     end
 
     # Whatever has to be done at the end of installation
     def finish_installation
+      # TODO: make it optional
```
imobachgs, line 219: I do not think we are going to make it optional but, for sure, it will not happen just because of this comment :-) If it is important, we should create a Trello card or open an issue.

imobachgs (commented):
In general, it looks good. I am just wondering why the tests are failing...

service/lib/agama/manager.rb
mchf: it was just an idea which i didn't want to forget of. I'll remove it if you don't think that can be useful

PR #889, opened 2023-11-24
Install recommended packages

imobachgs:
## Problem

Agama does not install recommended packages. The problem is a lot of software that a user might want is not installed. For instance, you can select a desktop environment in the software section, but Agama will not install a proper display manager. So you are presented with the typical text interface.

Manpages and many other software is missing.

## Solution

Consider recommended packages for installation to have an usable system.

## Testing

- Manually tested.


ancorgs (approved):
It makes sense to me

lslezak:
LGTM, we should probably make it configurable, for ALP products it makes sense to have "more minimal" system

PR #893, opened 2023-11-29
Add mac assigning to network base connection

jcronenberg:
## Problem

It's not possible to assign a mac address to network connections

## Solution

Add a property `mac_address` to base connection and also to `NetworkConnection` inside settings to allow setting of custom mac addresses.

**Note**: Currently for simplicity sake it just stores it as `String` all around, but with this draft I also wanted to start a discussion on what agama should use to store the mac address inside `BaseConnection`. `String` is of course the simplest, but doesn't perform any validation before sending it to NetworkManager, so the result is a pretty non explanatory error that the connection couldn't be added, because NetworkManager rejects any invalid macs of course. Let me know what you would prefer :slightly_smiling_face: (adjustments to the type shouldn't be to much work so no worries if I should change it)

## Testing

- *Modified existing unit tests*
- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/network/dbus/interfaces.rs
+++ b/rust/agama-dbus-server/src/network/dbus/interfaces.rs
@@ -238,6 +238,19 @@ impl Connection {
         connection.set_interface(name);
         self.update_connection(connection).await
     }
+
+    /// Custom mac-address
+    #[dbus_interface(property)]
+    pub async fn mac_address(&self) -> String {
+        self.get_connection().await.mac_address()
+    }
+
+    #[dbus_interface(property)]
+    pub async fn set_mac_address(&mut self, mac_address: &str) -> zbus::fdo::Result<()> {
+        let mut connection = self.get_connection().await;
+        connection.set_mac_address(mac_address)?;
+        self.update_connection(connection).await
+    }
 }
 
 /// D-Bus interface for Match settings
--- a/rust/agama-dbus-server/src/network/dbus/interfaces.rs
+++ b/rust/agama-dbus-server/src/network/dbus/interfaces.rs
@@ -6,11 +6,13 @@ use super::ObjectsRegistry;
 use crate::network::{
     action::Action,
     error::NetworkStateError,
-    model::{Connection as NetworkConnection, Device as NetworkDevice, WirelessConnection},
+    model::{
+        Connection as NetworkConnection, Device as NetworkDevice, MacAddress, WirelessConnection,
+    },
 };
 
 use agama_lib::network::types::SSID;
-use std::sync::Arc;
+use std::{str::FromStr, sync::Arc};
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::{MappedMutexGuard, Mutex, MutexGuard};
 use zbus::{
@@ -248,7 +250,7 @@ impl Connection {
     #[dbus_interface(property)]
     pub async fn set_mac_address(&mut self, mac_address: &str) -> zbus::fdo::Result<()> {
         let mut connection = self.get_connection().await;
-        connection.set_mac_address(mac_address)?;
+        connection.set_mac_address(MacAddress::from_str(mac_address)?);
         self.update_connection(connection).await
     }
 }
--- a/rust/agama-dbus-server/src/network/model.rs
+++ b/rust/agama-dbus-server/src/network/model.rs
@@ -319,9 +319,8 @@ impl Connection {
         self.base().mac_address.to_string()
     }
 
-    pub fn set_mac_address(&mut self, mac_address: &str) -> Result<(), InvalidMacAddress> {
-        self.base_mut().mac_address = MacAddress::from_str(mac_address)?;
-        Ok(())
+    pub fn set_mac_address(&mut self, mac_address: MacAddress) {
+        self.base_mut().mac_address = mac_address;
     }
 }
 
--- a/rust/agama-dbus-server/src/network/model.rs
+++ b/rust/agama-dbus-server/src/network/model.rs
@@ -354,7 +354,7 @@ pub enum MacAddress {
     Random,
     Stable,
     #[default]
-    None,
+    Unset,
 }
 
 impl FromStr for MacAddress {
@@ -366,7 +366,7 @@ impl FromStr for MacAddress {
             "permanent" => Ok(Self::Permanent),
             "random" => Ok(Self::Random),
             "stable" => Ok(Self::Stable),
-            "" => Ok(Self::None),
+            "" => Ok(Self::Unset),
             _ => Ok(Self::MacAddress(match macaddr::MacAddr6::from_str(s) {
                 Ok(mac) => mac,
                 Err(e) => return Err(InvalidMacAddress(e.to_string())),
@@ -383,7 +383,7 @@ impl fmt::Display for MacAddress {
             Self::Permanent => "permanent".to_string(),
             Self::Random => "random".to_string(),
             Self::Stable => "stable".to_string(),
-            Self::None => "".to_string(),
+            Self::Unset => "".to_string(),
         };
         write!(f, "{}", output)
     }
--- a/rust/agama-dbus-server/src/network/model.rs
+++ b/rust/agama-dbus-server/src/network/model.rs
@@ -310,12 +310,26 @@ impl Connection {
     pub fn is_loopback(&self) -> bool {
         matches!(self, Connection::Loopback(_))
     }
+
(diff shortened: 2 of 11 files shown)


imobachgs:
Hi @jcronenberg. First of all, thanks! And now, let's review the approach.

I prefer using some MAC address type. @teclator suggested using [macaddr](https://crates.io/crates/macaddr), which looks simple enough, but we are open to other options. However, using a `String` could be an acceptable temporary solution.

However, if the value is not mandatory, I would go with an `Option` (`Option<MacAddr>` or `Option<String>`). I would avoid using an empty string to represent that there is no address.

jcronenberg:
> I prefer using some MAC address type. @teclator suggested using [macaddr](https://crates.io/crates/macaddr), which looks simple enough, but we are open to other options. However, using a `String` could be an acceptable temporary solution.

Alright I tend to agree, so I'll look into using that crate for the mac address type.

> However, if the value is not mandatory, I would go with an `Option` (`Option<MacAddr>` or `Option<String>`). I would avoid using an empty string to represent that there is no address.

When going the `MacAddr` type route I agree, wrapping it inside an `Option` probably makes the most sense. However for `String` I found it to be unnecessarily complex because in the end the `None` option would have to be mapped to an empty string anyway because NetworkManager requires this setting with an empty string to overwrite any already set mac address.

jcronenberg:
> > When going the `MacAddr` type route I agree, wrapping it inside an `Option` probably makes the most sense. However for `String` I found it to be unnecessarily complex because in the end the `None` option would have to be mapped to an empty string anyway because NetworkManager requires this setting with an empty string to overwrite any already set mac address.
> 
> IMHO the `Option` represents the reality better. I like to think about this model as backend-agnostic, so the fact that we need to map it to a string for NetworkManager should not drive our design.

I think for pretty much all backends this would be the case, but anyway it doesn't really matter.

jcronenberg:
I have updated the PR with how I would envision it to work with a dedicated MAC type. Though I am unsure if the model is now also too NM agnostic for your liking.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -324,6 +345,55 @@ impl PartialEq for BaseConnection {
...
+pub struct InvalidMacAddress(String);
+
+#[derive(Debug, Clone)]
+pub enum MacAddress {
+    MacAddress(macaddr::MacAddr6),
+    Preserve,
+    Permanent,
+    Random,
+    Stable,
+}
```
jcronenberg, line 359: This is what I mean. Since these values may be exclusive to NM this may be too NM specific.

imobachgs: That's fine with me. If we decide to add support for another backend in the future, we might want to change that. But by now, let's be pragmatic.

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -218,6 +224,13 @@ fn wireless_config_to_dbus(conn: &WirelessConnection) -> NestedHash {
     let wireless: HashMap<&str, zvariant::Value> = HashMap::from([
         ("mode", Value::new(config.mode.to_string())),
         ("ssid", Value::new(config.ssid.to_vec())),
+        (
+            "assigned-mac-address",
+            Value::new(match &conn.base.mac_address {
+                Some(mac) => mac.to_string(),
+                None => "".to_string(),
```
cfconrad, line 231: because of this, maybe we could place `mac_address_string()` into `BaseConnection`

cfconrad (approved):
LGTM

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -306,24 +306,96 @@ impl Connection {
...
 
+impl BaseConnection {
+    pub fn mac_address_string(&self) -> String {
+        match &self.mac_address {
+            Some(mac) => mac.to_string(),
+            None => "".to_string(),
+        }
+    }
+
+    pub fn set_mac_address(&mut self, mac_address: &str) -> Result<(), InvalidMacAddress> {
```
imobachgs, line 334: This one looks counter-intuitive to me. Why do I need to convert a MacAddress to a string? Not to mention the fact that an empty string means `None`. If you add the `None` variant to the `MacAddress` enum, well, problem solved.

If you continue with the `Option` approach, I would 1) pass an option or 2) have a `set_mac_address` and a `unset_mac_address`.

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -306,24 +306,96 @@ impl Connection {
...
+    Stable,
+}
+
+impl FromStr for MacAddress {
+    type Err = InvalidMacAddress;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "preserve" => Ok(Self::Preserve),
+            "permanent" => Ok(Self::Preserve),
```
imobachgs, line 369: Why is `"permanent"` converted to `Preserve`?

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -306,24 +306,96 @@ impl Connection {
...
+            "random" => Ok(Self::Permanent),
```
imobachgs, line 370: Why rare `"random"` and `"stable"` converted to `Permanent`?

imobachgs (changes requested):
When I proposed an `Option` I was not aware that it was not a simple address, but that you were defining your own enum. Sorry if I overlooked something.

In that case, the `None` variant could be part of your enum, so you would not need to cope with that case anymore and you could simplify the code.

And that `_string` method would not be needed at all.

rust/agama-dbus-server/src/network/model.rs
jcronenberg: Oops must have been some lazy copy pasting :smile:

rust/agama-dbus-server/src/network/dbus/interfaces.rs
```diff
@@ -238,6 +238,19 @@ impl Connection {
...
     }
+
+    /// Custom mac-address
+    #[dbus_interface(property)]
+    pub async fn mac_address(&self) -> String {
+        self.get_connection().await.mac_address()
+    }
+
+    #[dbus_interface(property)]
+    pub async fn set_mac_address(&mut self, mac_address: &str) -> zbus::fdo::Result<()> {
```
teclator, line 249:
```suggestion
    pub fn set_mac_address(&mut self, mac_address: MacAddress) {
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -306,12 +306,26 @@ impl Connection {
...
+
+    pub fn is_ethernet(&self) -> bool {
+        matches!(self, Connection::Loopback(_)) || matches!(self, Connection::Ethernet(_))
+    }
+
+    pub fn mac_address(&self) -> String {
+        self.base().mac_address.to_string()
+    }
+
+    pub fn set_mac_address(&mut self, mac_address: &str) -> Result<(), InvalidMacAddress> {
```
teclator, line 318:
```suggestion
    pub fn set_mac_address(&mut self, mac_address: MacAddress) {
```

rust/agama-dbus-server/src/network/model.rs
```diff
@@ -306,12 +306,26 @@ impl Connection {
...
+        self.base_mut().mac_address = MacAddress::from_str(mac_address)?;
```
teclator, line 319:
```suggestion
        self.base_mut().mac_address = mac_address;

File: rust/agama-dbus-server/src/network/model.rs [hunk 9]
```diff
@@ -306,12 +306,26 @@ impl Connection {
...
+        Ok(())
```
teclator, line 320:
```suggestion
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -291,6 +299,34 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Option<BaseConnection> {
         base_connection.match_config = match_config_from_dbus(match_config)?;
     }
 
+    if let Some(ethernet_config) = conn.get(ETHERNET_KEY) {
+        if let Some(mac_address) = ethernet_config.get("assigned-mac-address") {
```
imobachgs, line 303: Would be possible to move the processing of the assigned-mac-address to a separate function? Both branches look the same to me.

imobachgs (approved):

PR #899, opened 2023-12-01
Force writing the timezone

imobachgs:
## Problem

YaST is rewriting the timezone settings. See #903.

## Solution

* Write the locale settings on a single call.
* Mock the `InstFunctions` module to avoid loading some code when asking whether the second stage will be executed (no second stage in Agama). This code ended up writing the timezone with the wrong value.

## Testing

- Tested manually


dgdavid (approved):
Looks good, 👍🏼

imobachgs:
After some tests, the real problem seems to be that YaST is overwriting the localtime. I need to rethink the solution.

PR #901, opened 2023-12-01
Fix list search

joseivanlopez:
## Problem

New component `ListSearch` was not filtering correctly when the search term contains capital letters.

## Solution

Fix the matcher in `ListSearch` component.

## Testing

* Adapted unti tests.
* Tested manually

## Screenshots

localhost_8080_(iPad Mini) (19)


joseivanlopez:
Duplicated: https://github.com/openSUSE/agama/pull/900

PR #911, opened 2023-12-04
Add package dependencies

joseivanlopez:
## Problem

From time to time an issue is reported because some missing dependency in the Agama live image (many times related to missing storage packages). Moreover, Agama depencencies are directly added to the ISO kiwi file, forcing to  all our live images to include that list of packages. Agama should requires its own dependencies instead on relying on the live iso to provide them.  

## Solution

* Collect a list of Agama dependencies by inspecting the packages included by YaST installation images, the kiwi files and the storage features used by *yast2-storage-ng*:
  * https://github.com/openSUSE/installation-images/blob/master/obs/installation-images.spec
  * https://github.com/openSUSE/installation-images/blob/master/data/root/root.file_list
  * https://build.opensuse.org/package/show/systemsmanagement:Agama:Staging/agama-live 
  * https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/storage_feature.rb
* Add dependencies directly to the Agama spec files.
* Remove Agama dependencies from kiwi files: 
  * https://build.opensuse.org/package/rdiff/systemsmanagement:Agama:Staging/agama-live?linkrev=base&rev=26
  * https://build.opensuse.org/package/rdiff/systemsmanagement:Agama:Devel/agama-live?linkrev=base&rev=9
* Add dependencies to `setup` scripts.
* Install required depedencies for CI tasks.

Notes:

* Right now both the `setup` scripts and the spec files contain almost the same list of packages. They should be unified.
* For CI tasks not all dependencies are installed, but only the required dependencies for running the tasks (e.g., unit tests).
* [Agama live imagein IBS](https://build.suse.de/package/show/SUSE:ALP:Products:Installer:1.0/agama-live) was not updated (YaST team does not maintain that image). Agama packages with the new dependencies will be submitted to IBS with the next Agama release.

### DIFF
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -1,13 +1,11 @@
 #!/bin/sh -x
 
-# Using a git checkout in the current directory,
-# set up the service (backend) part of agama
-# so that it can be used by
-# - the web frontend (as set up by setup.sh)
+# Using a git checkout in the current directory and set up the services, so that it can be used by:
+# - the web frontend (as set up by setup-web.sh)
 # - the CLI
 # or both
 
-# Exit on error; unset variables are an error
+# Exit on error; unset variables are an error.
 set -eu
 
 MYDIR=$(realpath $(dirname $0))
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -104,11 +104,6 @@ $SUDO zypper --non-interactive --gpg-auto-import-keys install \
 # Only install cargo if it is not available (avoid conflicts with rustup)
 which cargo || $SUDO zypper --non-interactive install cargo
 
-# This repo can be removed once python-language-data reaches Factory.
-test -f /etc/zypp/repos.d/d_l_python.repo || \
-  $SUDO zypper --non-interactive \
-    addrepo https://download.opensuse.org/repositories/devel:/languages:/python/openSUSE_Tumbleweed/ d_l_python
-
 # Packages required by Rust code (see ./rust/package/agama-cli.spec)
 $SUDO zypper --non-interactive install \
   bzip2 \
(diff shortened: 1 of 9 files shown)


setup-services.sh
```diff
@@ -32,29 +30,94 @@ sudosed() {
...
-
 # if agama is already running -> stop it
 $SUDO systemctl list-unit-files agama.service &>/dev/null && $SUDO systemctl stop agama.service
 
-# - Install service rubygem dependencies
+# Ruby services
+
+# Packages required for Ruby development (i.e., bundle install).
+$SUDO zypper --non-interactive --gpg-auto-import-keys install \
+gcc \
```
imobachgs, line 40: I would indent the package names for better readability. It is what we usually do in (e.g., in our Dockerfiles).

setup-services.sh
```diff
@@ -32,29 +30,94 @@ sudosed() {
...
 
-# - build also rust service
+# Rust service, CLI and auto-installation.
+
+# Only install cargo if it is not available (avoid conflicts with rustup)
+which cargo || $SUDO zypper --non-interactive install cargo
+
+# This repo can be removed once python-language-data reaches Factory.
+test -f /etc/zypp/repos.d/d_l_python.repo || \
+  $SUDO zypper --non-interactive \
```
imobachgs, line 109: I would say this repository is not needed anymore once that the split of `python-langtable` (separate packages for the library and the data) landed in [openSUSE:Factory](https://build.opensuse.org/package/view_file/openSUSE:Factory/python-langtable/python-langtable.spec?expand=1).

imobachgs (commented):
Thanks! I only have two comments but, otherwise, it looks great.

imobachgs (approved):

imobachgs (approved):

PR #912, opened 2023-12-04
Only stop agama service in setup script if it exists

jcronenberg:
## Problem

Commit f65e6c9 broke my personal way of setting up a testing container, because the service isn't yet registered and the setup script exits with an error.

## Solution

Only stop service if it already exists.

## Testing

- *Tested manually*


imobachgs (approved):
Thanks!

PR #881, opened 2023-11-21
Localization

joseivanlopez:
## Problem

Agama already offers a D-Bus interface for localizing both the installer and the target system. But the web UI does not allow to select a keyboard or timezone yet.

## Solution

The localization page was redesigned in order to allow changing language, keyboard and timezone for the target system.

The D-Bus API also needed some rework. Now it offers an API adapted to what clients need: listing locales, keymaps and timezones, and also selecting locales, keymap and timezone. 

Note: Auto-installation is out of the scope of this PR.

## Testing

* Added new unit tests
* Tested manually

## Screenshots

<details>
<summary>System in Spanish with GNOME Desktop</summary>

gnome-locale-es
</details>

<details>
<summary>UI</summary>

localhost_8080_(iPad Mini) (15)

localhost_8080_(iPad Mini) (16)

localhost_8080_(iPad Mini) (17)

localhost_8080_(iPad Mini) (18)

</details>

PR #914, opened 2023-12-05
Autoinstallation for localization (l10n) settings

mvidner:
## Problem

In #881 localization was improved and works in the web UI.
This PR adds CLI and autoinstallation

- https://trello.com/c/AvrY5iBP

## Solution

Mimic the structure of the other configuration scopes.

TODO:

- [x] decide where to use `locale` and/or `localization` in the code: mostly "localization" except the D-Bus proxy which is "locale" 

## Testing

- Tested manually
- More unit tests for `KeymapId`

## Screenshots

`agama config show` now works for `localization`:

```console
7e24bb397c23:/checkout # (cd rust/; cargo run --bin agama -- -f yaml config show)
[... Compiling ...]
     Running `target/debug/agama -f yaml config show`
user:
  fullName: ''
  [... etc ...]
localization:
  language: en_US
  keyboard: us
  timezone: Europe/Berlin
```

### DIFF
--- a/rust/agama-lib/src/localization/client.rs
+++ b/rust/agama-lib/src/localization/client.rs
@@ -1,0 +1,37 @@
+use crate::error::ServiceError;
+use crate::proxies::LocaleProxy;
+use zbus::Connection;
+
+/// D-Bus client for the software service
+pub struct LocalizationClient<'a> {
+    localization_proxy: LocaleProxy<'a>,
+}
+
+impl<'a> LocalizationClient<'a> {
+    pub async fn new(connection: Connection) -> Result<LocalizationClient<'a>, ServiceError> {
+        Ok(Self {
+            localization_proxy: LocaleProxy::new(&connection).await?,
+        })
+    }
+
+    pub async fn language(&self) -> Result<String, ServiceError> {
+        let locales = self.localization_proxy.locales().await?;
+        if locales.is_empty() {
+            // FIXME is this right?
+            Ok("".to_owned())
+        } else {
+            // without clone:
+            // error[E0507]: cannot move out of index of `Vec<std::string::String>`
+            // but why, it's a Vec<String> I should be able to move it?!
+            Ok(locales[0].clone())
+        }
+    }
+
+    pub async fn keyboard(&self) -> Result<String, ServiceError> {
+        Ok(self.localization_proxy.keymap().await?)
+    }
+
+    pub async fn timezone(&self) -> Result<String, ServiceError> {
+        Ok(self.localization_proxy.timezone().await?)
+    }
+}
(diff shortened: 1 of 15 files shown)


rust/agama-lib/src/localization/client.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub async fn new(connection: Connection) -> Result<LocalizationClient<'a>, ServiceError> {
+        Ok(Self {
+            localization_proxy: LocaleProxy::new(&connection).await?,
+        })
+    }
+
+    pub async fn language(&self) -> Result<String, ServiceError> {
+        let locales = self.localization_proxy.locales().await?;
+        if locales.is_empty() {
+            // FIXME is this right?
```
jreidinger, line 20: well, maybe more correct way is to use `Result<Option<String>, ServiceError>` and return `Ok(None)` here.

rust/agama-lib/src/localization/client.rs
```diff
@@ -0,0 +1,50 @@
...
+            Ok("".to_owned())
+        } else {
+            // without clone:
+            // error[E0507]: cannot move out of index of `Vec<std::string::String>`
+            // but why, it's a Vec<String> I should be able to move it?!
```
jreidinger, line 25: I think reason is that you want to return just part of Vector that will be destructed. I think that if you return whole locales, then it should work as you move whole ownership.

rust/agama-lib/src/localization/store.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub async fn new(connection: Connection) -> Result<LocalizationStore<'a>, ServiceError> {
+        Ok(Self {
+            localization_client: LocalizationClient::new(connection).await?,
+        })
+    }
+
+    pub async fn load(&self) -> Result<LocalizationSettings, ServiceError> {
+        // TODO: we should use a single D-Bus call with Properties.GetAll
+        // but LocaleProxy does not have it, only get_property for individual methods
+        // and properties_proxy is private
```
jreidinger, line 23: Just use Property proxy instead of LocaleProxy for given object.
https://docs.rs/zbus/latest/zbus/fdo/struct.PropertiesProxy.html

jreidinger: so basically you can construct proxy yourself...but it is usually not necessary as it is cached and zbus checking PropertiesChange signal for you, so in the end it is very efficient if property does not change too much.

rust/agama-lib/src/proxies.rs
```diff
@@ -75,6 +75,8 @@ trait Manager {
     ) -> zbus::Result<Vec<std::collections::HashMap<String, zbus::zvariant::OwnedValue>>>;
 }
 
+// IT'S HERE (FIXME: remove this)
```
jreidinger, line 78: ?

rust/agama-lib/src/localization/client.rs
mvidner: `vec.into_iter().next()` does what I want, yay

PR #916, opened 2023-12-05
Make the user visible patterns configurable, added MicroOS products

lslezak:
## Problem

- https://trello.com/c/Fj1hN4To/3498-3-agama-patterns-and-products-according-to-recent-decision
- Agama displays too many patterns in the openSUSE Tumbleweed installation and also offers some incompatible patterns for the MicroOS product which are in the same repository

## Solution

- Do not display all available patterns in the pattern selector, display only the explicitly listed patterns configured in the product YAML file

## Additional Fixes

- The Leap 16.0 product has been removed (discussed with Luboš)
- Added new openSUSE MicroOS products
- Enhanced the yupdate code to also patch the product definitions

## Testing

- Added new unit tests
- Tested manually

## Screenshots

| openSUSE Tumbleweed | openSUSE MicroOS |
|---|---|
|tw-patterns | microos-patterns |
| openSUSE MicroOS Desktop | Product Selection |
| microdesktop-patterns | product-selection |


Rakefile
```diff
@@ -170,5 +170,21 @@ if ENV["YUPDATE_FORCE"] == "1" || File.exist?("/.packages.initrd") || live_iso?
...
+      files = Dir.glob("products.d/*.y{a}ml")
+      files.each do |f|
+        # the sources contain several products, update only the existing files
+        oldfile = File.join("/usr/share/agama/", f)
+        if File.exist?(oldfile)
+          target = File.join(destdir, "/usr/share/agama/", f)
+          FileUtils.mkdir_p(File.dirname(target))
+          FileUtils.cp(f, target)
+        else
+          # if there is a new product file it needs to be copied manually
```
ancorgs, line 184: Without putting too much thinking into it... not sure if this is the behavior I would expect.

service/lib/agama/software/product.rb
```diff
@@ -94,6 +99,8 @@ def initialize(id)
         @optional_packages = []
         @mandatory_patterns = []
         @optional_patterns = []
+        # nil = display all patterns, [] = display none patterns
```
ancorgs, line 102: grammar: "display no patterns"

Rakefile
jreidinger: well, problem is that it is hard to know which product belongs to which iso

products.d/tumbleweed.yaml
```diff
@@ -47,6 +47,16 @@ software:
...
     - enhanced_base # only pattern that is shared among all roles on TW
   optional_patterns: null # no optional pattern shared
+  user_patterns:
+    - basic_desktop
+    - xfce
+    - kde
+    - gnome
+    - yast2_basis
+    - yast2_desktop
+    - yast2_server
```
jreidinger, line 57: I am not sure if it makes sense to have all three. Maybe just desktop and server yast modules will be enough?

service/lib/agama/software/product.rb
```diff
@@ -94,6 +99,8 @@ def initialize(id)
         @optional_packages = []
         @mandatory_patterns = []
         @optional_patterns = []
+        # nil = display all patterns, [] = display no patterns
```
jreidinger, line 102: is it all patterns or just all user visible patterns?

jreidinger (approved):
LGTM, nothing critical

Rakefile
lslezak: The problem is that the Git repository contains YAML files for all product. If you are patching the openSUSE Agama installer then the Live medium contains only the openSUSE products.

If yupdate would install all files you would suddenly see the ALP Dolomite product in the list, that would be confusing. So I decided to patch only the existing files and to not blindly copy everything available. IMHO it is better to have a safer default behavior.

products.d/tumbleweed.yaml
lslezak: I'm not sure. If you want to have YaST just for the package management then the `yast2_basis` is enough. On the other hand the "YaST Desktop" adds just few megabytes more, not a big deal...

PR #917, opened 2023-12-05
Configure installer keymap

joseivanlopez:
## Problem

- https://trello.com/c/dlSUsO5r/3503-5-agama-allow-selecting-a-keyboard-in-the-local-installation
- It should be possible to change the keyboard layout in local installation (in remote installation you can do that in your desktop as usually)

## Solution

- Display a keyboard switcher in the side bar, displayed only in local installation
- Note: In the long term there should be some "welcome screen" displayed in the local installation which should contain settings for the language and the keyboard

## Testing

- Added a new unit test
- Tested manually

## Screenshots

| Local connection<sup>1)</sup> | Remote connection |
|----|----|
|keyboard_local2 | keyboard_remote10 |

<sup>1)</sup> Local connection using the loopback device or or forced via `LOCAL_CONNECTION=1` environment variable.

### DIFF
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -42,6 +42,7 @@ import {
 import { InstallerKeymapSwitcher, InstallerLocaleSwitcher } from "./components/l10n";
 import { Layout, Loading, Title } from "./components/layout";
 import { useInstallerL10n } from "./context/installerL10n";
+import { localConnection } from "~/utils";
 
 // D-Bus connection attempts before displaying an error.
 const ATTEMPTS = 3;
@@ -117,7 +118,7 @@ function App() {
         <div className="full-width highlighted">
           <div className="flex-stack">
             <InstallerLocaleSwitcher />
-            <InstallerKeymapSwitcher />
+            {localConnection() ? <InstallerKeymapSwitcher /> : null }
           </div>
         </div>
       </Sidebar>
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -39,7 +39,7 @@ import {
   ShowTerminalButton,
   Sidebar
 } from "~/components/core";
-import { InstallerLocaleSwitcher } from "./components/l10n";
+import { InstallerKeymapSwitcher, InstallerLocaleSwitcher } from "./components/l10n";
 import { Layout, Loading, Title } from "./components/layout";
 import { useInstallerL10n } from "./context/installerL10n";
 
@@ -117,6 +117,7 @@ function App() {
         <div className="full-width highlighted">
           <div className="flex-stack">
             <InstallerLocaleSwitcher />
+            <InstallerKeymapSwitcher />
           </div>
         </div>
       </Sidebar>
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -39,7 +39,7 @@ import {
   ShowTerminalButton,
   Sidebar
 } from "~/components/core";
-import { LanguageSwitcher } from "./components/l10n";
+import { InstallerLocaleSwitcher } from "./components/l10n";
 import { Layout, Loading, Title } from "./components/layout";
 import { useInstallerL10n } from "./context/installerL10n";
 
@@ -116,7 +116,7 @@ function App() {
         </div>
         <div className="full-width highlighted">
           <div className="flex-stack">
-            <LanguageSwitcher />
+            <InstallerLocaleSwitcher />
           </div>
         </div>
       </Sidebar>
--- a/web/src/App.jsx
+++ b/web/src/App.jsx
@@ -42,7 +42,6 @@ import {
 import { InstallerKeymapSwitcher, InstallerLocaleSwitcher } from "./components/l10n";
 import { Layout, Loading, Title } from "./components/layout";
 import { useInstallerL10n } from "./context/installerL10n";
-import { localConnection } from "~/utils";
 
 // D-Bus connection attempts before displaying an error.
 const ATTEMPTS = 3;
@@ -117,8 +116,10 @@ function App() {
         </div>
         <div className="full-width highlighted">
           <div className="flex-stack">
-            <InstallerLocaleSwitcher />
-            {localConnection() ? <InstallerKeymapSwitcher /> : null }
+            <div className="locale-container">
+              <div><InstallerLocaleSwitcher /></div>
+              <div><InstallerKeymapSwitcher /></div>
+            </div>
           </div>
         </div>
       </Sidebar>
--- a/web/src/components/l10n/InstallerKeymapSwitcher.jsx
+++ b/web/src/components/l10n/InstallerKeymapSwitcher.jsx
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) [2023] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Icon } from "../layout";
+import { FormSelect, FormSelectOption } from "@patternfly/react-core";
+import { _ } from "~/i18n";
+import { useInstallerL10n } from "~/context/installerL10n";
+import { useL10n } from "~/context/l10n";
+
+const sort = (keymaps) => keymaps.sort((k1, k2) => k1.name > k2.name ? 1 : -1);
+
+export default function InstallerKeymapSwitcher() {
+  const { keymap, changeKeymap } = useInstallerL10n();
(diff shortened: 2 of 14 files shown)


web/src/App.jsx
```diff
@@ -117,7 +118,7 @@ function App() {
         <div className="full-width highlighted">
           <div className="flex-stack">
             <InstallerLocaleSwitcher />
-            <InstallerKeymapSwitcher />
+            {localConnection() ? <InstallerKeymapSwitcher /> : null }
```
joseivanlopez: This is the typical case in which using `If` is very handly: `<If condition={localConnection()} then={<InstallerKeymapSwitcher />}`

web/src/utils.js
```diff
@@ -249,6 +249,39 @@ const setLocationSearch = (query) => {
...
+ * This function should be used only in special cases, the Agama behavior should
+ * be the same regardless of the user connection.
+ *
+ * The local connection can be forced by setting the `LOCAL_CONNECTION`
+ * environment variable to `1`. This can be useful for debugging or for
+ * development.
+ *
+ * @returns {boolean} `true` if the connection is local, `false` otherwise
+ */
+const localConnection = function (location = window.location) {
```
joseivanlopez, line 264: NP: we are using arrow functions to declare anonymous functions.

web/src/utils.js
```diff
@@ -249,6 +249,39 @@ const setLocationSearch = (query) => {
...
+};
+
+/**
+ * Is the Agama server running remotely?
+ *
+ * @see localConnection
+ *
+ * @returns {boolean} `true` if the connection is remote, `false` otherwise
+ */
+const remoteConnection = (location = window.location) => !localConnection(location);
```
joseivanlopez, line 283: NP: I would pass all arguments to `#localConnection`: const remoteConnection = (...args) => !localConnection(...args);

web/src/components/l10n/InstallerKeymapSwitcher.jsx
```diff
@@ -0,0 +1,68 @@
...
+import { Icon } from "../layout";
+import { _ } from "~/i18n";
+import { useInstallerL10n } from "~/context/installerL10n";
+import { useL10n } from "~/context/l10n";
+import { localConnection } from "~/utils";
+
+const sort = (keymaps) => {
+  // sort the keymap names using the current locale
+  const lang = cockpit.language || "en";
+  return keymaps.sort((k1, k2) => k1.name.localeCompare(k2.name, lang));
```
joseivanlopez, line 36: Keymaps should come already translated from D-Bus.

web/src/components/l10n/InstallerKeymapSwitcher.jsx
```diff
@@ -0,0 +1,68 @@
...
+    .map((keymap, index) => <FormSelectOption key={index} value={keymap.id} label={keymap.name} />);
+
+  return (
+    <>
+      <h3>
+        {/* TRANSLATORS: label for keyboard layout selection */}
+        <Icon name="keyboard" size="24" />{_("Keyboard")}
+      </h3>
+      {(localConnection() &&
+      <FormSelect
```
joseivanlopez, line 55: I would use If here:

~~~
<If
    condition={localConnection()}
    then={
      <FormSelect>
        ...
      </FormSelect>
    }
    else={_("...")}
/>
~~~

web/src/components/l10n/InstallerKeymapSwitcher.jsx
```diff
@@ -0,0 +1,68 @@
...
+        id="keyboard"
+        // TRANSLATORS: label for keyboard layout selection
+        aria-label={_("keyboard")}
+        value={keymap}
+        onChange={onChange}
+      >
+        {options}
+      </FormSelect>) ||
+      // TRANSLATORS:
+      _("Keyboard layout cannot be changed in remote installation") }
```
joseivanlopez, line 65: As suggested by @dgdavid during the review, I would show the current keymap as disabled for remote installation.

dgdavid: > As suggested by @dgdavid during the review, I would show the current keymap as disabled for remote installation.

Just displaying it in plain text is enough along with the proper explanation in the popover @lslezak is working on. I mean, not needed to put a disabled selector there IMHO.

That said, at this moment I do not have an strong opinion for choosing between these two options.

web/src/App.jsx
```diff
@@ -116,7 +116,10 @@ function App() {
         </div>
         <div className="full-width highlighted">
           <div className="flex-stack">
-            <LanguageSwitcher />
+            <div className="locale-container">
```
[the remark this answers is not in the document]
dgdavid: In fact, I'm reducing a lot the use of `className` prop in a PR I'm working on. Not saying that it's always possible, but encourage you guys to actively ask for a second view on CSS thingy.

web/src/components/l10n/InstallerLocaleSwitcher.jsx
```diff
@@ -40,10 +42,32 @@ export default function InstallerLocaleSwitcher() {
   const options = Object.keys(languages).sort()
     .map(id => <FormSelectOption key={id} value={id} label={languages[id]} />);
 
+  // TRANSLATORS: help text for the language selector in the sidebar,
+  // %s will be replaced by the "Localization" page link
+  const [msg1, msg2] = _("The language used by the installer. The language \
+for the installed system can be set in the %s page.").split("%s");
+
+  // "hide" is a function which closes the popover
+  const description = (hide) => (
```
dgdavid, line 51: Do you really need the hide callback here? I mean, when the user clicks on "Localization" I expect the installer navigating to Localization... and closing the Sidebar. Still the popover visible? If so, something to fix.

lslezak: Without the callback the link closes the sidebar and navigates to the localization page but the popover is still open.

web/src/App.jsx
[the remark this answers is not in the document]
lslezak: Unfortunately `split` does not work well, but feel free to improve it later.

keyboard_remote8

web/src/components/l10n/InstallerKeymapSwitcher.jsx
lslezak: I guess you meant "sorted" here...

Yes, they are sorted. But are they sorted using the current locale? I mean in some languages the sorting might be  a bit different than in English...

web/src/components/l10n/InstallerKeymapSwitcher.jsx
lslezak: I tried that and I didn't like it. There are several problems:

- We would need to describe why the keyboard layout selector is disabled. It is not obvious and that could be frustrating for the users.
- It does not display your current local keyboard layout, which will be very likely different than the "English US" default.
- You do not care what's the keyboard layout on the remote machine, the current value is irrelevant, it does not provide any information for the remote user

See the screenshot:
keyboard_remote9

dgdavid: IMHO, one of the problem you found it's because using a disable selector for something that will be not be a selector. That's why I proposed using plain text and moving the explanation to a popover as we're doing now with the Language explanation.

But let's move on, we can re-evaluate at any time.

web/src/components/l10n/InstallerKeymapSwitcher.jsx
joseivanlopez: I mean, why do we need `#localeCompare`? Isn't comparison of `keymap.name` values enough?

web/src/components/l10n/InstallerKeymapSwitcher.jsx
joseivanlopez: As I understood it, the idea was to show the local keymap instead of the keymap in the remote system running Agama. But I guess that is not possible, right?

joseivanlopez (commented):
LGTM, but I cannot approve it.

dgdavid:
> LGTM, but I cannot approve it.

So, I'm going to approve it for you.

dgdavid (approved):
Approving it based on https://github.com/openSUSE/agama/pull/917#pullrequestreview-1780622498

web/src/components/l10n/InstallerKeymapSwitcher.jsx
lslezak: Unfortunately that's not possible, you cannot get the local keyboard layout name in Javascript running in browser.

There is some experimental [Keyboard API](https://developer.mozilla.org/en-US/docs/Web/API/Keyboard_API), but it does not provide the functionality we need.

You can get the name of a physical key to get a specified character, e.g. you can find which button produces letter `Y`. On the usual keyboard it is `Y`, but if you use QWERTZ layout you should get `Z`. (BTW I tested that with Chrome in KDE and it does not work properly with CZ QWERTZ layout.)

With that we could possibly only distinguish between QWERTY/QWERTZ/AZERTY variants, but we cannot get the full layout name like "Czech QWERTY".

web/src/components/l10n/InstallerKeymapSwitcher.jsx
lslezak: Some languages have special rules for sorting, for example in Estonian the letter `Z` is not the last letter. It is between `S` and `T`. See https://en.wikipedia.org/wiki/Estonian_orthography

We even had some "fun" with that in `[a-z]` regular expressions, see bug [#177560](https://bugzilla.suse.com/show_bug.cgi?id=177560).

I'm not sure if `localectl` output is locale sorted or not, this is just to be really sure.

lslezak: You can even try that in browser console: 
`["t", "a", "z"].sort((k1, k2) => k1.localeCompare(k2, "et"))` returns ` ['a', 'z', 't']`

PR #918, opened 2023-12-06
Add vlan model

cfconrad:
## Problem

Agama does not support VLAN configuration.

## Solution

Add support for VLAN configuration

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/network/model.rs
+++ b/rust/agama-dbus-server/src/network/model.rs
@@ -396,6 +396,7 @@ impl Connection {
             DeviceType::Loopback => ConnectionConfig::Loopback,
             DeviceType::Dummy => ConnectionConfig::Dummy,
             DeviceType::Bond => ConnectionConfig::Bond(Default::default()),
+            DeviceType::Vlan => ConnectionConfig::Vlan(Default::default()),
         };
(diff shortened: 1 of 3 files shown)


rust/agama-dbus-server/src/network/model.rs
```diff
@@ -679,6 +682,20 @@ impl From<&IpRoute> for HashMap<&str, Value<'_>> {
     }
 }
 
+#[derive(Debug, Default, PartialEq, Clone)]
+pub enum VlanProtocol {
+    #[default]
+    Ieee802_1Q,
+    Ieee802Ad,
```
teclator, line 689: Why not 
```suggestion
    Ieee802_1Ad,
```
in this case?

teclator:
In general LGTM

rust/agama-dbus-server/src/network/model.rs
cfconrad: Indeed, that's more precise.

cfconrad: changed to `IEEE802_1Q` and `IEEE802_1ad`

teclator (approved):

PR #919, opened 2023-12-07
Auto register

jreidinger:
## Problem

It is not possible to register system using automatic installation.

trello: https://trello.com/c/K94qhHWT/3499-3-agama-add-support-to-register-the-system-using-auto-installation

## Solution

Implement it and when touching it also add support for selecting patterns.

## Testing

- *Tested manually*
- for unit testing I need to learn how to mock properly in rust


rust/agama-lib/src/product/settings.rs
```diff
@@ -0,0 +1,14 @@
+//! Representation of the software settings
```
mvidner, line 1:
```suggestion
//! Representation of the product settings
```

rust/agama-lib/src/product/store.rs
```diff
@@ -0,0 +1,59 @@
+//! Implements the store for the storage settings.
```
mvidner, line 1:
```suggestion
//! Implements the store for the product settings.
```

rust/agama-lib/src/product/store.rs
```diff
@@ -0,0 +1,59 @@
...
+                // avoid selecting same product and unnecessary probe
+                self.product_client.select_product(product).await?;
+                probe = true;
+            }
+        }
+        if let Some(reg_code) = &settings.registration_code {
+            if let Some(email) = &settings.registration_email {
+                self.product_client.register(reg_code, email).await?;
+            } else {
+                self.product_client.register(reg_code, "").await?;
```
mvidner, line 48: Note to myself: check that `email` data elsewhere is documented as (non)empty

rust/agama-lib/src/product/store.rs
```diff
@@ -0,0 +1,59 @@
...
+        let email = self.product_client.email().await?;
+
+        Ok(ProductSettings {
+            id: Some(product),
+            registration_code: Some(registration_code),
+            registration_email: Some(email),
+        })
+    }
+
+    pub async fn store(&self, settings: &ProductSettings) -> Result<(), ServiceError> {
```
mvidner, line 34: This method will happily register something even if no product is selected. Do we want that?

rust/agama-settings/src/settings.rs
```diff
@@ -80,6 +80,23 @@ impl From<HashMap<String, String>> for SettingObject {
     }
 }
 
+impl From<String> for SettingObject {
```
mvidner, line 83: This should get better documentation, because Settings is hard to understand now already, with generic words like Object and Value all around, plus the metaprogramming

mvidner (commented):
Reviewing the Product Client+Store
Still todo: Software Client+Store

rust/agama-lib/share/profile.schema.json
```diff
@@ -10,9 +10,27 @@
       "description": "Software settings (e.g., product to install)",
       "type": "object",
       "properties": {
-        "product": {
```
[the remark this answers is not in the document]
jreidinger: agreed

rust/agama-lib/src/product/store.rs
jreidinger: @imobachgs what is your opinion?

rust/agama-lib/src/product/client.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// register product
+    pub async fn register(&self, code: &str, email: &str) -> Result<(), ServiceError> {
+        let mut options = HashMap::new();
+        if !email.is_empty() {
+            options.insert("Email", email);
+        }
+        self.registration_proxy
+            .register(code, HashMap::new())
```
mvidner, line 95: Yaayyy, I've found a bug :smile:

rust/agama-lib/src/software/client.rs
```diff
@@ -1,71 +1,78 @@
...
         })
     }
 
-    /// Returns the available products
-    pub async fn products(&self) -> Result<Vec<Product>, ServiceError> {
-        let products: Vec<Product> = self
-            .product_proxy
-            .available_products()
+    /// Returns the available patterns
+    pub async fn patterns(&self, filtered: bool) -> Result<Vec<Pattern>, ServiceError> {
```
mvidner, line 36: `filtered` is not documented here nor in https://github.com/openSUSE/agama/blob/39534694aed2b3ce941e4c4b896051ab1cd0660b/doc/dbus/org.opensuse.Agama.Software1.doc.xml#L8

Filtered according to what?

Not your fault ;-) but let's fix it now

mvidner (commented):
In Patterns it turns out I am bothered by not propagating our API documentation through all the layers that present it, which becomes acute when the missing docs in question is the meaning of magic numbers.

It does not affect correctness of the code but it increases its poor maintainablilty

mvidner:
Please show me how you tested manually so that I can replicate (or run into errors, the error paths should be fun too)

mvidner (commented):
My manual test loses information:

```console
# grep -A3 product.*: rust/agama-lib/share/examples/profile_Dolomite.json
  "product": {
    "id": "ALP-Dolomite",
    "registrationCode": "FILL IT UP",
    "registrationEmail": "jreidinger@suse.com"
# (cd rust/; cargo run --bin agama -- -f yaml config load agama-lib/share/examples/profile_Dolomite.json)
# (cd rust/; cargo run --bin agama -- -f yaml config show)
[...]
product:
  id: ALP-Dolomite
  registrationCode: ''
  registrationEmail: ''
```

But Josef says it is actually working as designed: `config show` talks to our D-Bus layer which only changes `email` once `Register` is called. Hmm I think CLI users will find it confusing.

rust/agama-lib/src/software/client.rs
jreidinger: done in https://github.com/openSUSE/agama/pull/919/commits/b75a6e14a2fd1108aa6cf8bf0bba8c2fef8e25dc

rust/agama-lib/src/product/store.rs
```diff
@@ -0,0 +1,63 @@
...
+        }
+        if let Some(reg_code) = &settings.registration_code {
+            let (result, message);
+            if let Some(email) = &settings.registration_email {
+                (result, message) = self.product_client.register(reg_code, email).await?;
+            } else {
+                (result, message) = self.product_client.register(reg_code, "").await?;
+            }
+            if result != 0 {
+                return Err(ServiceError::FailedRegistration(message));
```
mvidner, line 52: Thanks!

... ... but ... my testing shows:
```console
# agama config load agama-lib/share/examples/profile_Dolomite.json
Registration failed: 'Connection to registration server failed'
```

So far so good, right?

```console
# tail -n2 daemon.log
Announcing system to https://scc.suse.com ...
[ERROR]: software: Error connecting to registration server: Unknown Registration Code.
```

So the CLI would have me debug network connections instead of checking for typos in the regcode. We still need to improve this.

mvidner: When I quickly insert the integer result code into `FailedRegistration` it is revealed to be `6` which means "API error" which is a disappointment, I hoped to see `8` "incorrect credentials".

mvidner: aaand I have fixed it in our Ruby backend :smile:

mvidner (commented):
tl;dr: This was not enough but I've fixed that myself :muscle:

mvidner (changes requested):
Adding the error reporting has revealed this, which makes the whole call fail:

```console
# rust/target/debug/agama config load agama-lib/share/examples/profile_tw.json
Registration failed: 'Product does not require registration'
```

`profile_tw.json` does not include `registrationCode`, the CLI is forcing it:
```json
  "product": {
    "id": "Tumbleweed"
  },
```

mvidner:
> Not a fault of this PR, but `agama config add` seems to have never worked:
> ```

On Imo's advice,
`agama config add software.patterns value=gnome`
has worked, but IMHO it's wrongly exposing an implementation detail.

Also, we should update the help text, `rust/agama-cli/doc/CLI_API.md` and `doc/cli_guidelines.md`
And fix the panics

mvidner:
Pattern selection errors are logged only the *second* time the profile is loaded, and hidden from the CLI.
(details to follow)

jreidinger:
Well, reason is that it is well hidden that pattern is missing. Code is from https://github.com/yast/yast-packager/blob/4e7687c61f165140248a8045149dc67cd547ebf1/src/modules/Packages.rb#L1907

jreidinger:
Ok, so code computes it is basically
https://github.com/yast/yast-packager/blob/4e7687c61f165140248a8045149dc67cd547ebf1/src/modules/Packages.rb#L1850
Question is what dbus API to adapt? Add/Remove/Set Patterns? Or just Set Patterns as it works on set of patterns? We need to modify its dbus API to report issue, so we will be able to display it. I do not expect that web UI is affected as it display only known and valid patterns.
Another option is to not modify DBus API and instead check list of all patterns ( so unfiltered ) in rust code and report it before calling those actions.
@imobachgs @mvidner opinions what will be better?

imobachgs:
> Ok, so code computes it is basically https://github.com/yast/yast-packager/blob/4e7687c61f165140248a8045149dc67cd547ebf1/src/modules/Packages.rb#L1850 Question is what dbus API to adapt? Add/Remove/Set Patterns? Or just Set Patterns as it works on set of patterns? We need to modify its dbus API to report issue, so we will be able to display it. I do not expect that web UI is affected as it display only known and valid patterns. Another option is to not modify DBus API and instead check list of all patterns ( so unfiltered ) in rust code and report it before calling those actions. @imobachgs @mvidner opinions what will be better?

As usual, I am not sure, but I think that the D-Bus API should be responsible for reporting that problem. Actually, I would expect all of them (Add/Remove/Set) to be able to inform us when a pattern does not exist. I do not know how hard it could be.

About error reporting, perhaps you could redefine the "Report" module for this case (as we did with other modules).

mvidner:
In general, D-Bus API can stay unchanged, as long as you are able to raise a D-Bus **error** with a single string description.
If the error needs more structured data, D-Bus protocol still allows that but tooling (D-Feet) gets confused, so an API change is better.

jreidinger:
@mvidner wrong pattern is now properly reported. See this screenshot from my VM:
agama_pattern_not_exist

rust/agama-lib/src/software/client.rs
```diff
@@ -73,7 +73,7 @@ impl<'a> SoftwareClient<'a> {
         let wrong_patterns = self.software_proxy
             .set_user_patterns(patterns.as_slice())
             .await?;
-        if wrong_patterns.is_empty() {
+        if !wrong_patterns.is_empty() {
```
mvidner, line 77: :hand_over_mouth:

doc/dbus/org.opensuse.Agama.Software1.doc.xml
```diff
@@ -4,18 +4,68 @@
...
+      -->
       <arg name="id" direction="in" type="s"/>
+      <!--
+        Returns false if pattern does not exist in current list of repositories.
+        Otherwise it returns true including case when pattern was not previously selected.
+      -->
+      <arg name="result" direction="out" type="b"/>
     </method>
+    <!--
+      Sets explicitelly selected patterns.
```
mvidner, line 57:
```suggestion
      Sets explicitly selected patterns.
```

rust/package/agama-cli.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Fri Dec  8 09:23:09 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
+
+- Change the config in a way that: (gh#openSUSE/agama#919)
+  1. product is moved to own section and is now under product.id
+  2. in product section is now also registrationCode and registrationEmail
+  3. in software section is now patterns to select patterns to install
+- adapt profile.schema according to above changes
```
mvidner, line 8:
```suggestion
- adapt profile.schema according to above changes
- org.opensuse.Agama.Software1 API changed to report missing patterns
```

rust/agama-lib/src/error.rs
```diff
@@ -16,6 +16,10 @@ pub enum ServiceError {
     Anyhow(#[from] anyhow::Error),
     #[error("Wrong user parameters: '{0:?}'")]
     WrongUser(Vec<String>),
+    #[error("Registration failed: '{0}'")]
+    FailedRegistration(String),
+    #[error("Failed to find these patterns: '{0:?}'")]
```
mvidner, line 21:
```suggestion
    #[error("Failed to find these patterns: {0:?}")]
```

The Debug format will use `["crocheting"]` already, no need for more quotes

mvidner (approved):
Thanks!

Errors get reported to the CLI user now

PR #920, opened 2023-12-08
Fixed broken integration test, the `setup-service.sh` script has been renamed to `setup-services.sh`

lslezak:
## Problem

- The integration test fails
- The problem is that the `setup-service.sh` script has been renamed to `setup-services.sh` in #911 but not all places which use it were fixed

## Solution

- Fix the script name everywhere


mvidner (approved):
Thanks!

PR #924, opened 2023-12-11
Changing the network model

imobachgs:
## Problem

Agama's network model is heavily inspired by [nmstate](https://github.com/nmstate/nmstate/blob/c9fd1e80d3c87e1f844edcd86e9d36ae499ce717/rust/src/lib/ifaces/ethernet.rs#L64). Connections are represented [by an enum](https://github.com/openSUSE/agama/blob/2bc452b39e4460122ab52c5a72ee0d1dfe154cec/rust/agama-dbus-server/src/network/model.rs#L220) and each variant contains a struct which depends on the type (e.g., the [WirelessConnection](https://github.com/openSUSE/agama/blob/2bc452b39e4460122ab52c5a72ee0d1dfe154cec/rust/agama-dbus-server/src/network/model.rs#L542) struct). Those structs have a `base` field that points to a [BaseConnection struct](https://github.com/openSUSE/agama/blob/master/rust/agama-dbus-server/src/network/model.rs#L328) which holds the common fields for all connection types.

And that's perfectly fine. However, we decided to introduce some methods to hide the fact that the common information lives in a separate `struct`. We could say it is for encapsulation reasons, but I don't know whether it is a good idea. Even we created some *getters* and *setters* for those fields, adding quite some [boilerplate](https://github.com/openSUSE/agama/blob/master/rust/agama-dbus-server/src/network/model.rs#L264-L299) [code](https://github.com/openSUSE/agama/blob/master/rust/agama-dbus-server/src/network/model.rs#L318-L324)

Additionally, until we introduced the `ConnectionBuilder` (in #885), creating a new connection with some custom data (like an id and a UUID) was not straightforward because you needed to make the `BaseConnection` first (see [this example](https://github.com/openSUSE/agama/blob/2bc452b39e4460122ab52c5a72ee0d1dfe154cec/rust/agama-dbus-server/src/network/model.rs#L136-L141).

## Proposal

Although I am not 100% sure whether it is a better idea, I am exploring the possibility of using a simpler layout to represent a connection. In this case, we would use a single `struct` containing the common fields plus a `config` field for specific stuff. That field contains an `enum` which represents each kind of connection.

```rust
/// Represents an availble network connection
#[derive(Debug, Default, Clone)]
pub struct Connection {
    pub id: String,
    pub uuid: Uuid,
    pub mac_address: MacAddress,
    pub ip_config: IpConfig,
    pub status: Status,
    pub interface: String,
    pub match_config: MatchConfig,
    pub config: ConnectionConfig,
}

#[derive(Default, Debug, PartialEq, Clone)]
pub enum ConnectionConfig {
    #[default]
    Ethernet,
    Wireless(WirelessConfig),
    Loopback,
    Dummy,
}
```

And not getters/setters as they are not needed at all.

## Testing

- Adapted the unit tested.

### DIFF
--- a/rust/agama-dbus-server/src/network/nm/dbus.rs
+++ b/rust/agama-dbus-server/src/network/nm/dbus.rs
@@ -28,29 +28,36 @@ pub fn connection_to_dbus<'a>(
     controller: Option<&'a Connection>,
 ) -> NestedHash<'a> {
     let mut result = NestedHash::new();
-    let mut connection_dbus =
-        HashMap::from([("id", conn.id().into()), ("type", ETHERNET_KEY.into())]);
+    let mut connection_dbus = HashMap::from([
+        ("id", conn.id.as_str().into()),
+        ("type", ETHERNET_KEY.into()),
+    ]);
 
-    if let Some(interface) = &conn.interface() {
+    if let Some(interface) = &conn.interface {
         connection_dbus.insert("interface-name", interface.to_owned().into());
     }
 
     if let Some(controller) = controller {
         connection_dbus.insert("slave-type", "bond".into()); // TODO: only 'bond' is supported
-        let master = controller.interface().unwrap_or(controller.id());
+        let master = controller
+            .interface
+            .as_deref()
+            .unwrap_or(controller.id.as_str());
         connection_dbus.insert("master", master.into());
     } else {
         connection_dbus.insert("slave-type", "".into()); // TODO: only 'bond' is supported
         connection_dbus.insert("master", "".into());
     }
 
-    result.insert("ipv4", ip_config_to_ipv4_dbus(conn.ip_config()));
-    result.insert("ipv6", ip_config_to_ipv6_dbus(conn.ip_config()));
-    result.insert("match", match_config_to_dbus(conn.match_config()));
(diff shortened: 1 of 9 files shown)


rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -57,25 +59,25 @@ pub fn connection_to_dbus(conn: &Connection) -> NestedHash {
 ///
 /// This functions tries to turn a OwnedHashMap coming from D-Bus into a Connection.
 pub fn connection_from_dbus(conn: OwnedNestedHash) -> Option<Connection> {
-    let base = base_connection_from_dbus(&conn)?;
+    let mut connection = base_connection_from_dbus(&conn)?;
```
cfconrad, line 62: Not sure if it is a problem at all. This function show the case that we now easily transform a Connection from one type into the other, by setting the `config` enum. Before it was like, collecting common settings + specific settings and then construct the Connection. 

But currently I do not see a problem with it. For me this approach has benefits over the current.

imobachgs: Yes, I have the same opinion and it does not look like a problem (at least by now). I think this approach is simpler. But I will work on this once the bonding support is finally merged.

teclator (approved):
LGTM

PR #926, opened 2023-12-13
File system and snapshots config

joseivanlopez:
## Problem

The file system type cannot be selected when adding or editing a file system. The file system is always created with its default file system type. 

* https://trello.com/c/JnSKpQnu/3418-3-storage-modify-basic-attributes-fstype-snapshots-etc-for-pre-defined-volumes

## Solution

Add a selector for the file system type.

* If the product allows configuring snapshots, then the selector includes both *Btrfs* and *Btrfs with snapshots* (and the rest of possible types). 
* If there is only an option for the file system type (e.g., *swap*), then the selector is not offered. A plain text with the unique value is shown.

Follow-up: https://github.com/openSUSE/agama/issues/932

## Testing

* Added new unit tests
* Tested manually

## Screenshots

[Screencast from 2023-12-15 14-52-33.webm](https://github.com/openSUSE/agama/assets/1112304/feac328d-8ffd-4485-be5b-997b2aad49b6)

### DIFF
--- a/web/src/assets/styles/patternfly-overrides.scss
+++ b/web/src/assets/styles/patternfly-overrides.scss
@@ -225,3 +225,9 @@ table td > .pf-v5-c-empty-state {
     --pf-v5-c-expandable-section__content--PaddingLeft: var(--spacer-normal);
   }
 }
+
+.pf-v5-c-form__group-label-help {
+  margin: 0;
+  padding: 0;
+  vertical-align: top;
+}
--- a/web/src/components/storage/VolumeForm.jsx
+++ b/web/src/components/storage/VolumeForm.jsx
@@ -19,13 +19,10 @@
  * find current contact information at www.suse.com.
  */
 
-import React, { useReducer } from "react";
-
+import React, { useReducer, useState } from "react";
 import {
-  InputGroup, InputGroupItem,
-  Form, FormGroup, FormSelect, FormSelectOption,
-  Radio,
-  TextInput,
+  InputGroup, InputGroupItem, Form, FormGroup, FormSelect, FormSelectOption, MenuToggle,
+  Radio, Select, SelectOption, SelectList
 } from "@patternfly/react-core";
 import { sprintf } from "sprintf-js";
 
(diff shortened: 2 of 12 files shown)


web/src/components/storage/VolumeForm.jsx
```diff
@@ -19,19 +19,17 @@
...
-  TextInput,
+  InputGroup, InputGroupItem, Form, FormGroup, FormSelect, FormSelectOption, MenuToggle,
+  Popover, Radio, Select, SelectOption, SelectList
 } from "@patternfly/react-core";
 import { sprintf } from "sprintf-js";
 
 import { _, N_ } from "~/i18n";
 import { FormValidationError, If, NumericTextInput } from '~/components/core';
 import { DEFAULT_SIZE_UNIT, SIZE_METHODS, SIZE_UNITS, parseToBytes, splitSize } from '~/components/storage/utils';
+import { Icon } from "../layout";
```
lslezak, line 32: It should be consistent with the other imports:

```suggestion
import { Icon } from "~/components/layout";
```

web/src/assets/styles/patternfly-overrides.scss
```diff
@@ -225,3 +225,9 @@ table td > .pf-v5-c-empty-state {
     --pf-v5-c-expandable-section__content--PaddingLeft: var(--spacer-normal);
   }
 }
+
+.pf-v5-c-form__group-label-help {
```
lslezak, line 229: Can it affect some other dialogs? I just afraid of some side effects...

web/src/components/storage/VolumeForm.jsx
```diff
@@ -81,6 +79,224 @@ const MountPointFormSelect = ({ volumes, ...formSelectProps }) => {
...
+/**
+ * Btrfs file system type can be offered with two flavors, with and without snapshots.
+ */
+const BTRFS_WITH_SNAPSHOTS = "BtrfsWithSnapshots";
+
+/**
+ * Possible file system type options for a volume.
+ * @function
+ *
+ * @param {import(~/clients/storage).Volume} volume
```
lslezak, line 91: This looks strange. Is it correct?

web/src/components/storage/VolumeForm.jsx
```diff
@@ -81,6 +79,224 @@ const MountPointFormSelect = ({ volumes, ...formSelectProps }) => {
...
+
+  const details = () => {
+    if (fsOption === BTRFS_WITH_SNAPSHOTS) return _("with snapshots");
+    return undefined;
+  };
+
+  const description = () => {
+    if (fsOption === BTRFS_WITH_SNAPSHOTS)
+      // TRANSLATORS: description of Btrfs snapshots feature.
+      return _("Allows rolling back any change done to the system and restoring its previous state");
```
lslezak, line 168: The description looks good. Could we have something similar for the other file systems?

At least for Ext2 we should mention that it does not support journaling so it can possibly get corrupted on crash or power outage...

web/src/components/storage/VolumeForm.jsx
```diff
@@ -81,6 +79,224 @@ const MountPointFormSelect = ({ volumes, ...formSelectProps }) => {
...
+          onClick={e => e.preventDefault()}
+          className="pf-v5-c-form__group-label-help"
+        >
+          <Icon name="info" size="16" />
+        </button>
+      </Popover>
+    );
+  };
+
+  const label = _("File system type");
```
lslezak, line 281: Please add some context comment for the translators (so they know it's a combo box label)

web/src/components/storage/VolumeForm.jsx
joseivanlopez: I think it makes sense to add some key info to other file system types too. That could help users to select the proper type for his/her use case. But, if you don't mind, I would address it in a seperate PR. We need to find good helps for each type.

web/src/assets/styles/patternfly-overrides.scss
joseivanlopez: As far as I was able to check, currently it is not affecting. Anyway, I would like to discuss it with @dgdavid. Maybe we need some custom components for unitying  that info tips. For example, the icon in the installer language selector is not wrapped by a clickable element (e.g., button).

lslezak (approved):

dgdavid:
Post merge comment:

Apart from having a look to what you mentioned in https://github.com/openSUSE/agama/pull/926#discussion_r1428123849, please remind me to talk about the "information" icon close to the label and the focus. By looking the screencast, I had the impression that it is "breaking" the form flow. Don't time to check it now, let's have a look on Monday.

Thanks guys.

PR #928, opened 2023-12-14
Add up or down interface state to network model

jcronenberg:
## Problem

Currently a connection always gets activated when it is added.
This is a problem for the wicked migration described here https://github.com/jcronenberg/agama/issues/20

## Solution

Add `Up` and `Down` state to `Status` to control this behavior.

## Testing

- *Tested manually*
- *Tested in the wicked migration https://github.com/jcronenberg/agama/tree/skip-activation*

### DIFF
--- a/rust/agama-dbus-server/src/network/nm/client.rs
+++ b/rust/agama-dbus-server/src/network/nm/client.rs
@@ -137,7 +137,11 @@ impl<'a> NetworkManagerClient<'a> {
             proxy.add_connection(new_conn).await?
         };
 
-        self.activate_connection(path).await?;
+        if conn.is_up() {
+            self.activate_connection(path).await?;
+        } else {
+            self.deactivate_connection(path).await?;
+        }
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


rust/agama-dbus-server/src/network/nm/client.rs
```diff
@@ -122,6 +126,24 @@ impl<'a> NetworkManagerClient<'a> {
...
+            Err(e) => {
+                // Ignore ConnectionNotActive error since this just means the state is already correct
+                if e.to_string().contains("ConnectionNotActive") {
+                    Ok(())
+                } else {
+                    Err(ServiceError::DBus(e))
+                }
+            }
+            _ => Ok(()),
+        }
```
jcronenberg, line 144: I am unsure about this part. This is needed for the migration otherwise it would show the `ConnectionNotActive` error when creating the connection with the state initially set to down. This was the best way I found but maybe there is a better way?

imobachgs: I do not think we get a specific error from the D-Bus service, so your approach looks good.

imobachgs (approved):

PR #929, opened 2023-12-14
Ruby service - mark translatable strings

lslezak:
# Service Translations

- This is just the first part of the Ruby support
- Mark the translatable texts for translation
- Added `Rakefile` to extract the texts into the POT file using `rake pot` call
- Updated the GitHub Action to create POT file also for the Ruby service

### DIFF
--- a/.github/workflows/weblate-update-pot.yml
+++ b/.github/workflows/weblate-update-pot.yml
@@ -24,7 +24,7 @@ jobs:
         run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && zypper ref
 
       - name: Install tools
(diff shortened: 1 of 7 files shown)


imobachgs (approved):
Thanks!

joseivanlopez (approved):

mvidner:
@lslezak I tried to test this locally with https://github.com/nektos/act ... it doesn't work since the openSUSE/agama-weblate is checked out with a GH_TOKEN secret.
Maybe that's correct and necessary, given the nature of the Action, but it's really hard to test+review this way

lslezak:
Unfortunately the problem is that if you need to push to a different repository then you have to explicitly specify the GitHub token. Without it you can pull, but pushing later fails. :worried:

PR #910, opened 2023-12-04
Drop Leap 16.0 for now

lkocman:
* We can probably expect openSUSE variants of Dolomite or Marble prior we have Leap 16.0
* Fixes Issue #907

PR #930, opened 2023-12-15
CI: do not refer to the leap16.yml file anymore

imobachgs:
## Problem

CI tests are failing because a reference to the leap16.yml file still remains. See #910.

## Solution

Remove that reference.


dgdavid (approved):
Thanks!

PR #931, opened 2023-12-15
Added GitHub Action for merging the Ruby translations

lslezak:
## Problem

- Add a GitHub Action to merge back the updated Ruby service translations
- The code is very similar to the existing action for updating the web frontend translations ([weblate-merge-po.yml](https://github.com/openSUSE/agama/blob/master/.github/workflows/weblate-merge-po.yml))
- Later I'll try doing some refactoring and sharing some parts, but for now I think it is good enough


joseivanlopez (approved):
LGTM

PR #940, opened 2023-12-19
Package the Ruby service translations

lslezak:
## Problem

- The Ruby service translations are not package to RPM

## Solution

- Enhance the `rake package` task to create the `package/po.tar.bz2` file with all `po/*.po` files
- Update the `gem2rpm.yml` to unpack the `po.tar.bz2` archive and build the binary `*.mo` files from the text `*.po` files (the MO files are smaller and easier to parse) during RPM build
- The MO files are installed into the usual YaST2 translation directory (`/usr/share/YaST2/locale`) so it can be read the same way as the standard YaST translations, no change in the code is required

## Testing

- Tested manually - I built the package locally using the `osc` command, the built RPM contains the `/usr/share/YaST2/locale/cs/LC_MESSAGES/agama.mo` file

### DIFF
--- a/service/package/install_translations.sh
+++ b/service/package/install_translations.sh
@@ -1,0 +1,10 @@
+#! /bin/sh
+
+# a helper script for compiling and installing the translations
+
+PODIR=$(mktemp --directory --suffix "-agama-po")
+
+tar xfjv "$1" -C "$PODIR"
+find "$PODIR" -name "*.po" -exec sh -c 'L=`basename "{}" .po` && mkdir -p "$RPM_BUILD_ROOT/usr/share/YaST2/locale/$L/LC_MESSAGES" && msgfmt -o "$RPM_BUILD_ROOT/usr/share/YaST2/locale/$L/LC_MESSAGES/agama.mo" "{}"' \;
+
+rm -rf "$PODIR"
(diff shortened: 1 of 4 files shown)


service/package/install_translations.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/sh
+
+# a helper script for compiling and installing the translations
+
+PODIR=$(mktemp --directory --suffix "-agama-po")
+
+tar xfjv "$1" -C "$PODIR"
+find "$PODIR" -name "*.po" -exec sh -c 'L=`basename "{}" .po` && mkdir -p "$RPM_BUILD_ROOT/usr/share/YaST2/locale/$L/LC_MESSAGES" && msgfmt -o "$RPM_BUILD_ROOT/usr/share/YaST2/locale/$L/LC_MESSAGES/agama.mo" "{}"' \;
```
lslezak, line 8: Note: Originally I wanted to put this line into the `gem2rpm.yml` file but I fell into the escaping hell (shell script in a `find` argument which is another shell script placed in an YAML file... :scream: ). So I put that into a separate script to make reading and writing easier.

jreidinger (approved):

PR #944, opened 2023-12-20
Software conflicts research

joseivanlopez:
## Problem

Software conflicts cannot be solved in Agama and we need to find out how to do it. This is specially evident when selecting incompatible patterns (e.g., KDE and Gnome in openSUSE MicroOS Desktop).

## Solution

Create a research document pointing to how to notify and solve conflicts by using the questions mechanism.

### DIFF
--- a/doc/research/software-conflicts.md
+++ b/doc/research/software-conflicts.md
@@ -27,7 +27,7 @@ zypp::ProblemSolution
     #details
 ~~~
 
-A list of solutions can be passed to `#applySolutions` method in order to fix the reported problems. The list of solutions is built by selecting a solution from each problem. Then the solver (`#resolvePool`) can be called again.
+A list of solutions can be passed to `#applySolutions` method in order to fix the reported problems. The list of solutions is built by selecting a solution from the problems. Then the solver (`#resolvePool`) can be called again. Note that `applySolutions` does not require a solution for all the problems. Sometimes a problem is gone as consequence of solving another problem.
 
 ~~~
 zypp::ProblemSolutionList solutions;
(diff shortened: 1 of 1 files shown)


doc/research/software-conflicts.md
```diff
@@ -0,0 +1,94 @@
...
+zypp::ResolverProblem
+    #details
+    #solutions (zypp::ProblemSolutionList)
+
+zypp::ProblemSolution
+    #description
+    #details
+~~~
+
+A list of solutions can be passed to `#applySolutions` method in order to fix the reported problems. The list of solutions is built by selecting a solution from each problem. Then the solver (`#resolvePool`) can be called again.
```
jreidinger, line 30: I think `The list of solutions is built by selecting a solution from each problem` is not correct. As it does not need to be from each, but from any number of problems.

doc/research/software-conflicts.md
```diff
@@ -0,0 +1,94 @@
...
+...
+zypp::getZYpp()->resolver()->applySolutions(solutions);
+bool success = zypp::getZYpp()->resolver()->resolvePool();
+~~~
+
+## Managing solver problems in Agama
+
+From Agama point of view, the software proposal could be considered as a not finished task meanwhile there are dependency conflicts. Following that approach, the questions mechanism fits well as a strategy for requesting user intervesion.
+
+Therefore, when a software proposal is calculated, Agama could raise a question if needed in order to make possible to finish the proposal without conflicts. The user should be able to answer the question by selecting a solution for each conflict. Then, Agama would try to calculate the proposal again. If the user decides to cancel the question and finish without solving the problems, then Agama would generate an issue indicating the dependency problems. Note that the installation will not start if there is any issue with level error. Consequently, such a dependency issue would prevent the installation to start.
```
jreidinger, line 43: Again here I would say `for any` and not `for each` as it is not needed.

doc/research/software-conflicts.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Notes
+
+Agama can be easily fixed to show an issue when selecting a pattern produces some conflicts. But the dependency issue reported by Agama is far from ideal. It only indicates the number of conflicts (e.g., *Found 2 dependency issues.*).
+
+The feature for reporting and fixing solver problems could be iteratively improved in several steps:
+
+1) Start by showing the dependency issue when selecting a pattern.
+2) Extend *yast-pkg-bindings* and improve the issue message, showing the details of the problems.
+3) Raise a question when the solver generates problems. This should be done for each call to the solver, not only for selecting patterns.
```
jreidinger, line 94: I still think that at least in some cases we should offer option to revert new selection to solve issue. Like you add qt..ok...you add gnome and get conflict with option revert that unselects gnome without too much thinking.

joseivanlopez: Yes, I think it is doable for some cases. We could emit a first question for undo the change. If the question is canceled, then we emit a question for solving issues.

I think it is something to evaluate when implementing it, but it would not require nothing new from D-Bus or yast-pkg-bindings.

PR #946, opened 2023-12-20
Display most representative country in timezone selector

ancorgs:
## Problems

Problem 1.

In the timezone selector the countries are only displayed if they are part of the name of the timezone (eg. America/Argentina/Buenos_Aires). In most cases, the name of the timezone includes only the global region and a concrete city (eg. Europe/Madrid). That makes hard to find the right timezone since the most intuitive mechanism (typing the name of the user country in the search box) is usually useless.

spain-old

Problem 2.

There are some duplicated or controversial timezones that are not associated to any country. For more details [this gist](https://gist.github.com/ancorgs/8bd5f42ea847e1ba7590bc7e3d54cdad).

buenos_aires-old

## Solution

Add the "most representative country" to each timezone and display that in the UI. That information is read from the file `/usr/share/zoneinfo/zone.tab`.

That makes it easier to find timezones.

spain-new

Timezones without a country (except those that really make sense) are filtered out. That removes controversy and duplicates.

buenos_aires-new

Works even for the two timezones that have a different identifier at langtable and at `zone.tab`. Again, see the [previously mentioned gist](https://gist.github.com/ancorgs/8bd5f42ea847e1ba7590bc7e3d54cdad).

rangoon-new

## Testing

- Added new unit tests
- Tested manually (as shown in the screenshots)

### DIFF
--- a/doc/dbus/bus/org.opensuse.Agama1.Locale.bus.xml
+++ b/doc/dbus/bus/org.opensuse.Agama1.Locale.bus.xml
@@ -70,9 +70,12 @@
      * The timezone identifier (e.g., "Europe/Berlin").
      * A list containing each part of the name in the language set by the
        UILocale property.
+     * The name, in the language set by UILocale, of the main country
+       associated to the timezone (typically, the name of the city that is
+       part of the identifier) or empty string if there is no country.
      -->
     <method name="ListTimezones">
-      <arg type="a(sas)" direction="out"/>
+      <arg type="a(sass)" direction="out"/>
     </method>
     <method name="Commit">
     </method>
--- a/doc/dbus/org.opensuse.Agama1.Locale.doc.xml
+++ b/doc/dbus/org.opensuse.Agama1.Locale.doc.xml
@@ -34,9 +34,12 @@
      * The timezone identifier (e.g., "Europe/Berlin").
      * A list containing each part of the name in the language set by the
        UILocale property.
+     * The name, in the language set by UILocale, of the main country
+       associated to the timezone (typically, the name of the city that is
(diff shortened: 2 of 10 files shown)


imobachgs (commented):
The code LGTM. But, please, update the changes files.

joseivanlopez:
Maybe it is a matter of taste, but using one more column makes the selector to look a bit crowded. What about showing all parts in the first column and country in the second?

~~~
Africa - Ceuta                          Spain              10:10
Atlantic Ocean - Canary        Spain              09:10  
~~~

Timezones without country would have an empty second column, but that cases are only some exceptions.

imobachgs (approved):

ancorgs:
> What about showing all parts in the first column and country in the second?

I have no strong preference. So let's see.

This is how it looks now in this PR (keeping what existed and just adding a column for the country):

four-column-a
four-column-b
four-column-c

And this is how it would like what you suggest (merge the parts in to the first column and use the second for the country):

three-columns-a
three-columns-b
three-columns-c

@dgdavid, @imobachgs, @joseivanlopez What do you prefer?

imobachgs:
> @dgdavid, @imobachgs, @joseivanlopez What do you prefer?

I would choose the second option as it is more clear to me: the localized name of the timezone and the country. If I see more columns, I need to somehow think what they are all about.

Having said that, I do not see the reason to use a hyphen to separate the components, but let's keep that for a different discussion.

ancorgs:
> If I see more columns, I need to somehow think what they are all about.

Good argument. Let's go for the second option then. I just added a commit to change that (although I don't think I will update the screenshots in the description).

imobachgs (approved):

PR #948, opened 2023-12-21
Add dependency on timezone

imobachgs:
Add timezone dependency in agama-cli as it needs access to `/usr/share/timezone/zone.tab`.


dgdavid (approved):
LGTM

PR #953, opened 2023-12-22
Limit the dependency on fde-tools to x86_64 and aarch64

imobachgs:
The `fde-tools` package is available only in `x86_64` and `aarch64` architectures. Thus, we need to adjust the dependency in the spec file.


teclator (approved):
LGTM

PR #954, opened 2023-12-22
Move the timezone dependency to the correct package

imobachgs:
It is a `agama-dbus-server` dependency (#948 is wrong, although it fixes the building problem by accident).

### DIFF
--- a/rust/package/agama-cli.spec
+++ b/rust/package/agama-cli.spec
@@ -32,7 +32,6 @@ BuildRequires:  cargo-packaging
 BuildRequires:  pkgconfig(openssl)
 # used in tests for dbus service
 BuildRequires:  python-langtable-data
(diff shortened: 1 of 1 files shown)


dgdavid (approved):
LGTM

rust/package/agama-cli.spec
```diff
@@ -43,8 +42,6 @@ Requires:       bzip2
 Requires:       tar
 # required for translating the keyboards descriptions
 Requires:       xkeyboard-config-lang
```
joseivanlopez, line 44: Should  xkeyboard-config-lang be moved?

imobachgs:
Accordgin to #1033 we plan to reorganize the packages a bit. So let's postpone this change.

PR #955, opened 2023-12-22
Add setup and CI dependencies

joseivanlopez:
## Problem

Some dependencies have been added, but the setup script and CI tasks do not include them. 

Note: we need some way of extracting deps automatically. Having spec, setup and CI manually in sync is error prone.

## Solution

Add dependencies.

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -68,7 +68,12 @@ jobs:
       run: zypper --non-interactive install rustup
 
     - name: Install required packages
-      run: zypper --non-interactive install python-langtable-data openssl-3 libopenssl-3-devel jq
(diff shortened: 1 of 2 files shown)


jreidinger (approved):

jreidinger:
@joseivanlopez btw in ruby we have `rake build_dependencies:install           # Install the packages required for building` maybe we can have something similar also for runtime dependencies

joseivanlopez:
> @joseivanlopez btw in ruby we have `rake build_dependencies:install # Install the packages required for building` maybe we can have something similar also for runtime dependencies

I think that is the idea of the setup script: to install and configure everything for running agama. Not sure if having a task only for installing deps has any value here.

PR #959, opened 2023-12-26
Mark some Agama service strings for translation

imobachgs:
## Problem

Although all the infrastructure to translate the Ruby-based Agama service is in-place, some strings are still not marked for translation.

## Solution

Mark the manager and software manager progress strings.


dgdavid (approved):
LGTM

PR #961, opened 2023-12-26
Improve D-Bus network interface

imobachgs:
## Problem

Until now, the communication between our network model and its D-Bus interface worked only in one direction (from D-Bus to the model). And that was pretty limiting.

## Solution

In #885 we introduce the usage of `oneshot::channel()` to enable bi-directional communication between the model and the D-Bus interface. It worked rather well, so it is time to extend this approach.

Some use cases that are covered:

* Perform a complex validation in the model (considering the whole model) and return the result to the D-Bus interface.
* Return the D-Bus path of a new connection (implemented in this PR).

But there is another interesting implication: the model is the single source of truth (no more cloned data in the interface). In the future, we are expected to listen to NetworkManager and update some parts of our model accordingly. Not having to replicate that information to D-Bus (except when new things add or disappear) makes things easier.

Note: `org.opensuse.Agama1.Network.Device` and `org.opensuse.Agama1.Network.Devices` are excluded by now.

## Testing

- *Added a new unit test*
- *Tested manually*

## To do

- [x] Reduce the boilerplate associated to the usage of channels.


rust/agama-dbus-server/src/network/dbus/interfaces/common.rs
```diff
@@ -0,0 +1,81 @@
...
+        rx.await
+            .unwrap()
+            .ok_or(NetworkStateError::UnknownConnection(
+                self.uuid().to_string(),
+            ))
+    }
+
+    /// Updates the connection data in the NetworkSystem.
+    ///
+    /// * `connection`: Updated connection.
```
teclator, line 39:
```suggestion
    /// * `func`: function to update the connection.
```

rust/agama-dbus-server/src/network/dbus/interfaces/connections.rs
```diff
@@ -0,0 +1,305 @@
...
+/// It offers an API to query the connections collection.
+pub struct Connections {
+    actions: Arc<Mutex<UnboundedSender<Action>>>,
+    objects: Arc<Mutex<ObjectsRegistry>>,
+}
+
+impl Connections {
+    /// Creates a Connections interface object.
+    ///
+    /// * `objects`: Objects paths registry.
```
teclator, line 25: actions is missing

rust/agama-dbus-server/src/network/dbus/interfaces/ip_config.rs
```diff
@@ -6,71 +6,51 @@
...
     actions: Arc<Mutex<UnboundedSender<Action>>>,
-    connection: Arc<Mutex<NetworkConnection>>,
+    uuid: Uuid,
 }
 
 impl Ip {
     /// Creates an IP interface object.
     ///
     /// * `actions`: sending-half of a channel to send actions.
     /// * `connection`: connection to expose over D-Bus.
```
teclator, line 32:
```suggestion
    /// * `uuid`: connection UUID..
```

rust/agama-dbus-server/src/network/dbus/tree.rs
```diff
@@ -76,62 +76,46 @@ impl Tree {
         Ok(())
     }
 
-    /// Adds a connection to the D-Bus tree.
+    /// Adds a connection to the D-Bus tree and returns the D-Bus path.
     ///
     /// * `connection`: connection to add.
```
teclator, line 81:
```suggestion
    /// * `conn`: connection to add.
```

teclator (approved):
LGTM, just added some suggestions but only about documentation

PR #976, opened 2024-01-02
[web][doc] Change the typedoc output

dgdavid:
In general, the Agama documentation start aging bad. It's either, disperse in a bunch of files, outdated, incomplete, or sometimes even wrong.

At the time of writing, using a static site generator like [Docusaurus](https://docusaurus.io/) or [Vitepress](https://vitepress.dev) look as good bet to start fixing these problems by putting all the Agama information accessible in a kind of website without too much effort.

In that context, with these tools in mind, this PR can be seen a starting point for having a better output of autogenerated documentation.

Next steps must include reviewing all the written documentation for fixing and improving at the same time that `@category`, `@internal`, and other [Typedoc tags](https://typedoc.org/guides/tags/) are used correctly for an even better output.

Follow the commits to dig into details ;)

---

<details>
<summary>Click to show/hide a not much helpful screenshots</summary>

---

The index, :see_no_evil: 

| Before | After |
|-|-|
|Screen Shot 2024-01-08 at 15 47 11 | Screen Shot 2024-01-08 at 15 47 04 |

<sub>^^^ Note the typo in Fronted</sub>

The generated doc

| Before | After |
|-|-|
| Screen Shot 2024-01-08 at 15 46 04 | Screen Shot 2024-01-08 at 15 44 16 |
| Note that only src/client doc was generated. Which actually could make sense, but let's generate the components docs too by now to help us to catch mistakes there. We can evaluate in the future if we want to link it or not | Screen Shot 2024-01-08 at 15 44 46 |


joseivanlopez (approved):
LGTM

PR #977, opened 2024-01-04
Fixed integration test

lslezak:
## Problem

- The integration test was failing
- https://github.com/openSUSE/agama/actions/runs/7395724982/job/20119447308

## Solution

During debugging I found several problems:

- The `agama.gemspec` file calls `git describe --tags`, that does not work when the Git checkout contains only the last commit and no history. Fixed by fetching the Git repository with full history. (Note: There is no significant slow down, even with the full history the Git repository is fetched in 2 seconds.)
- The Cockpit service is started manually but later systemd wants to start it as a service dependency and that fails because the port 9090 is already taken.
- The products directory now contains several YAML product files, delete all of them except the Tumbleweed configuration.
- Increased the timeout in the tests, refreshing the Tumbleweed repository sometimes might take more time and in that case the test fails because of time out.
- More precise "Back" button specification in `root_password.spec.ts` test, that text matched the "Czech (extra backslash)" text in the keyboard selection menu. (Playwright by default does a partial case-insensitive match, short texts might accidentally match a substring in longer texts.)
- Added some more debugging steps in the workflow to get more details when something fails

## Testing

- The integration test now passes without problems
- https://github.com/lslezak/agama/actions/runs/7407057992/job/20152559638

### DIFF
--- a/.github/workflows/ci-integration-tests.yml
+++ b/.github/workflows/ci-integration-tests.yml
@@ -22,6 +22,11 @@ jobs:
 
     - name: Git Checkout
       uses: actions/checkout@v3
+      with:
+        # fetch complete history with tags, agama.gemspec calls "git describe --tags"
+        # that would fail with just last commit checked out
+        fetch-depth: 0
+        fetch-tags: true
 
     - name: Created shared YaST log directory
       run: mkdir -p /tmp/log/YaST2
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
LGTM, thanks for taking care.

Worth to mention what we've talked in IRC, 

> **\<dgdavid>** lslezak: why only TW?
> **\<dgdavid>** lslezak: I mean, what if we want to set an integration test for another product?
> **\<lslezak>** dgdavid: so far the tests are simple and they check the same things regradless of the selected product
> **\<lslezak>** dgdavid: we can test more products after adding more tests

PR #886, opened 2023-11-23
Remove global issues page and start using popup instead

balsa-asanovic:
## Problem

"We have agreed to use the sidebar for actions related to the installer itself (e.g., diagnosis tools, display lang, etc). Everything related to the target system to install should not be placed in the siderbar.

The only thing related to the target system still present in the sidebar is the issues link. The issues should be globally accessible from another place (footer, header, ...)."

As described in #872 

This PR fixes #872 

## Solution

Issues page link is removed from the sidebar and all other places where it was present.

During the progress of this PR, decision was made to drop the showing of global (all) issues at once in the popup.

Because of this, the Issues popup (previously a page) can no longer be accessed from the sidebar (or from the icon in the header, which was an idea at some point).

The Issues Dialog (popup) now shows issues related only to a single section (software, product, storage, ...) and is opened by clicking on the warning link in the sections on the overview page (and a few other places).

For this reason, the prop id for the Section component was introduced, so the issues could be filtered to a specific type for the popup based on from where has the user clicked the warning.

## Testing

-  Tested manually
- Adjusted unit test files to some extent, possibly more work is needed here.

## Screenshots

294189450-e2b43bbc-65cb-42bb-aa64-092eb8e2b0f5

PR #979, opened 2024-01-08
[web] Fix Agama selectors look&feel

dgdavid:
## Problem

The `ul` styles were overwritten in #886 to make them looks like what they are: unordered lists. However, such a changed impacted in the look&feel of the current Agama selectors.

## Solution

Restore the Agama selectors by setting their list-type and margins explicetely.

## Testing

- Tested manually

## Screenshots

| Before | After |
|-|-|
| Screen Shot 2024-01-08 at 16 37 13 | Screen Shot 2024-01-08 at 16 36 58 |

## Additional notes

No changelog entry required.


joseivanlopez (approved):
LGTM

PR #980, opened 2024-01-08
[web] Drop core/NotificationMark component

dgdavid:
## Problem

The core/`NotificationMark` component and its styles are no longer used since https://github.com/openSUSE/agama/pull/886, where they were overlooked and not removed.

## Solution

Drop it to avoid having dead code around.

## Testing

- *Tested manually*

## Additional notes

This change does not require an entry in the changes file.


joseivanlopez (approved):
LGTM

PR #981, opened 2024-01-08
Avoid duplicated keymaps

joseivanlopez:
## Problem

The list of available keymaps contains some duplicated values.

## Solution

Fix the regexp to obtain the layout and variant names from `localectl`. Now it allows layouts with dots (e.g., *lt.std*) and variants with dashes (e.g., *pt-nativo-epo*).

Note for reviewers: regexp can be checked with https://rustexp.lpil.uk/.

## Testing

* Added new unit tests.
* Tested manually.

## Screenshots

<details>
<summary>Toggle</summary>

localhost_8080_(iPad Mini) (23)

localhost_8080_(iPad Mini) (22)

</details>


dgdavid (approved):
Although I do not know too much Rust, I could check the RegExp and it looks good :+1:

PR #985, opened 2024-01-10
[web] Update dependencies (2024-01-10)

dgdavid:
Update web dependencies to their latest versions via `npm update` and `npm install package-name@latest` when needed.

It does not update `stylelint` dev dependencies because `stylelint-webpack-plugin` [does not work with the latest stylelint v16 yet](https://github.com/webpack-contrib/stylelint-webpack-plugin/issues/342). We can wait for getting linters up-to-date.


dgdavid:
> Update web dependencies to their latest versions via `npm update` and `npm install package-name@latest` when needed.

Note for reviewers: take a look at the `package-lock.json` diff to see all the packages updated.

imobachgs (approved):

PR #988, opened 2024-01-10
Update README.md

nagu165:
- Added extra information regarding the download procedure for Agama ISO files.
- Gave an overview of how to run Agama using the virtual machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,9 @@ There are two flavors of live ISO images:
 
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
+* Make sure to download the correct ISO file according to your system architecture(Example: You would need to choose {https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build1.17.iso} if your processor is based on x64 architecture.
+* And do not forget to check out the ALP vs openSUSE difference mentioned above and download the correct file based on your requirements.
+* Inorder to launch Agama using virtual machine please refer https://github.com/openSUSE/agama/issues/932#issuecomment-1863007573.
 
 ### Manual Configuration
 
--- a/README.md
+++ b/README.md
@@ -125,10 +125,9 @@ There are two flavors of live ISO images:
 
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
-* Make sure to download the correct ISO file according to your system architecture(Example: You would need to choose {https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build1.17.iso} if your processor is based on x64 architecture.
-* And do not forget to check out the ALP vs openSUSE difference mentioned above and download the correct file based on your requirements.
-* Inorder to launch Agama using virtual machine please refer https://github.com/openSUSE/agama/issues/932#issuecomment-1863007573.
-
+* Make sure to download the correct ISO file according to your system architecture (eg.
+ you would need to choose an file including `x86_64` if you use an Intel or AMD 64-bit processor)
+ and according to the system you want to install (openSUSE vs ALP).
 ### Manual Configuration
 
 You can run Agama from its sources by cloning and configuring the project:
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ There are two flavors of live ISO images:
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
 * Make sure to download the correct ISO file according to your system architecture (eg.
- you would need to choose an file including `x86_64` if you use an Intel or AMD 64-bit processor)
+ you would need to choose a file including `x86_64` if you use an Intel or AMD 64-bit processor)
  and according to the system you want to install (openSUSE vs ALP).
 ### Manual Configuration
 


README.md
```diff
@@ -125,6 +125,9 @@ There are two flavors of live ISO images:
 
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
+* Make sure to download the correct ISO file according to your system architecture(Example: You would need to choose {https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build1.17.iso} if your processor is based on x64 architecture.
```
dgdavid, line 128: I got your point here, but I do not think it's a good idea linking to a specific version of Agama in the README. @imobachgs, @mvidner, do you have any idea how to improve this?

README.md
```diff
@@ -125,6 +125,9 @@ There are two flavors of live ISO images:
...
+* And do not forget to check out the ALP vs openSUSE difference mentioned above and download the correct file based on your requirements.
+* Inorder to launch Agama using virtual machine please refer https://github.com/openSUSE/agama/issues/932#issuecomment-1863007573.
```
dgdavid, line 130:
```suggestion
* In order to launch Agama using virtual machine please refer https://github.com/openSUSE/agama/issues/932#issuecomment-1863007573.
```

Probably is better to add a section about VirtualBox. Maybe @lslezak or @mvidner  can help with this.

README.md
lslezak: There is always an ISO build which does not contain the build number:, e.g.: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.x86_64-openSUSE.iso

However, I'd not recommend using it. It makes bug reporting more difficult as you do not know the exact build number you used. And you can also accidentally use a wrong file without noticing.

It is an example so I think it is fine.

README.md
lslezak: I'd rather extract the relevant text here. Linking to a comment in another long discussion is not nice.

dgdavid: > I'd rather extract the relevant text here. Linking to a comment in another long discussion is not nice.

Sure, that's what I meant with "a section about".

README.md
imobachgs: Well, it is actually a redirection. See:

```
$ curl -i https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.x86_64-openSUSE.iso
HTTP/2 302 
date: Thu, 11 Jan 2024 13:03:28 GMT
server: Mojolicious (Perl)
cache-control: public, max-age=300
content-length: 0
location: https://downloadcontent.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build2.62.iso
vary: Accept,COUNTRY,X-COUNTRY,Fastly-SSL
content-type: application/x-iso9660-image
```

Using a browser (tested with Firefox and Chromium) you get a file with the correct name. For cURL, you might need to do:

```
curl -L -O  https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.x86_64-openSUSE.iso
```

For wget you need to do:

```
wget --turst-server-names https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.x86_64-openSUSE.iso
```

Anyway, I agree with @lslezak, it is just an example.

imobachgs, line 128: Please, keep `x86_64` (actually, the important part IMHO is `x86`).

imobachgs, line 128:
```suggestion
* Make sure to download the correct ISO file according to your system architecture. For instance, you would need to choose {https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build1.17.iso} if your processor is based an x86_64.
```

README.md
imobachgs: Not only that: the comment talks about running the ISO *for development purposes*. IMHO it does not belong to the README.

README.md
```diff
@@ -125,6 +125,9 @@ There are two flavors of live ISO images:
 
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
+* Make sure to download the correct ISO file according to your system architecture(Example: You would need to choose {https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-live.x86_64-7.0.0-openSUSE-Build1.17.iso} if your processor is based on x64 architecture.
+* And do not forget to check out the ALP vs openSUSE difference mentioned above and download the correct file based on your requirements.
```
imobachgs, line 129: I would get rid of that part. After all, you already mentioned that there are two different flavors in one paragraph above.

imobachgs: Not related to this PR, but we should explain in some place what the `Playwright` flavors are about.

README.md
ancorgs: Exactly what @imobachgs said. That comment is not about using the downloaded ISO. Is about using that ISO as an starting point for Agama development. A completely different topic that does not belong here.

README.md
ancorgs: Agreed. This point just repeats the information given two lines above.

ancorgs:
First of all, thanks a lot for the contribution. It's really nice that you want to add some clarifications to avoid newcomers confusion. But...

In its current form, this pull request adds quite some text with three points in which (according to previous comments from me and others):

- The first point contains a direct link that can lead to errors
- The second one is redundant with the previous paragraph
- The third one corresponds to a different topic

I think the same goal could be achieved with a less intrusive change. Something like this:

```diff
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
+Make sure to download the correct ISO file according to your system system architecture (eg.
+you would need to choose an file including `x86_64` if you use an Intel or AMD 64-bit processor)
+and according to the system you want to install (openSUSE vs ALP).
```

Writing a good README is much harder than it looks (there is a hard balance about what to include and what not). And maintaining it up to date is also very important (says a lot about the health of a project). So thank you sincerely for taking care and sorry for being a bit nitpicking. ;-)

README.md
```diff
@@ -125,7 +125,9 @@ There are two flavors of live ISO images:
 
 You can download them from the [openSUSE Build
 Service](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/).
-
+* Make sure to download the correct ISO file according to your system architecture (eg.
+ you would need to choose an file including `x86_64` if you use an Intel or AMD 64-bit processor)
```
dgdavid, line 129:
```suggestion
 you would need to choose a file including `x86_64` if you use an Intel or AMD 64-bit processor)
```

dgdavid (approved):

dgdavid:
@nagu165 

Thank you a lot for this contribution, already merged in `master` branch. Hope it helps people to choose the right ISO.

Thanks also to my @openSUSE/yast team colleages for given their feedback here. Much appreciated.

Issue #987, opened 2024-01-10
wrong encoding of Swedish letters + gnome-terminal not starting


bittin:
Hey!

i have been testing Agama with openSUSE Tumbleweed during the first month of this year 
and the encoding of letters is wrong in the Wayland GNOME Environment for example when it creates the Downloads folder its called HÃ¤mtningar instead of Downloads

and also it has problems with launching gnome-terminal # Error constructing proxy for org.gnome.Terminal:/org/gnome/Terminal/Factory0: Fel vid anrop av StartServiceByName fÃ¶r org.gnome.Terminal: TidsgrÃ¤nsen uppnÃ¥ddes
tried to start it with english language but got the exact same problem: # Error constructing proxy for org.gnome.Terminal:/org/gnome/Terminal/Factory0: Error calling StartServiceByName for org.gnome.Terminal: Timeout was reached 

xfce4-terminal works however

imobachgs:
This problem does not happen on a system installed using YaST. That's why we think it is a bug.

mvidner:
This kind of garbled text means the text is encoded in UTF-8 (as expected) but the terminal interprets it as ISO-8859-x.

_Hypothesis_: we write the locale identifier as `sv_SE` (which is fine on the web) but that one uses ISO-8859-1. We should write `sv_SE.UTF-8` instead.

```console
$ locale -av
...
locale: sv_SE           directory: /usr/lib/locale/sv_SE
-------------------------------------------------------------------------------
    title | Swedish locale for Sweden
 language | Swedish
territory | Sweden
  codeset | ISO-8859-1

locale: sv_SE.utf8      directory: /usr/lib/locale/sv_SE.utf8
-------------------------------------------------------------------------------
    title | Swedish locale for Sweden
 language | Swedish
territory | Sweden
  codeset | UTF-8
...
```

imobachgs:
Hi all,

Thanks @mvidner, you are right! I confirmed that we are not including the encoding. We could 1) force to use UTF-8 always (which is almost a one-liner) or 2) take the encoding into account (which looks more correct to me). I am working on the solution 2).

imobachgs:
Once #992 has been merged, I will consider the issue solved. I have installed the system in Swedish and it seems to work:

![Captura desde 2024-01-16 13-38-32](https://github.com/openSUSE/agama/assets/15836/f6de4f45-ea73-44b3-8377-0a95ec9a24c5)

Thanks for the report!

bittin:
[...]

PR #992, opened 2024-01-11
Consider locales encoding

imobachgs:
## Problem

See #987.

## Solution

* The encoding should be included in the locale, although at this point we always use `UTF-8` (check `localectl list-locales`).
* I adapted the web UI to ignore the encoding (it is only interested in the language).

## Updating the spec file

I updated the `spec` file according to the latest changes of cargo_vendor:

```
* cargo_config is no longer created - it's part of the vendor.tar now
    * You can safely remove lines related to cargo_config from your spec file

* multiple cargotoml files can be specified and share a single vendor.tar
    * If multiple cargo.toml files are present update does not work. This is a known
      limitation of the process

* cargo_audit is now part of cargo_vendor, meaning you don't have to configure it separately
```

## Testing

* Manually tested using a branch in OBS.


dgdavid (commented):
Draft: So far, it looks good to me :+1:

dgdavid (approved):
LGTM 👍 However, you might want a review from another colleague for the Rust part. Up to you 😉

imobachgs:
> LGTM 👍 However, you might want a review from another colleague for the Rust part. Up to you 😉

Thanks a lot! I just wanted @mvidner to, please, have a look as our language expert :-)

rust/agama-dbus-server/src/l10n/helpers.rs
```diff
@@ -23,8 +23,7 @@ pub fn init_locale() -> Result<LocaleCode, Box<dyn std::error::Error>> {
 ///
 pub fn set_service_locale(locale: &LocaleCode) {
     // Let's force the encoding to be 'UTF-8'.
```
joseivanlopez, line 25: This comment does not apply anymore, right?

service/lib/agama/ui_locale.rb
```diff
@@ -46,10 +46,11 @@ def initialize(locale_client, &block)
 
     def change_locale(locale)
       # TODO: check if we can use UTF-8 everywhere including strange arch consoles
```
joseivanlopez, line 48: And this TODO?

joseivanlopez (approved):

PR #995, opened 2024-01-12
Make TPM-based encryption more explicit

ancorgs:
## Problem

Sometimes, Agama decides to use the encryption method `TPM_FDE` which results in the system being configured via `fde-tools` to open the encryption devices automatically during system boot without needing to enter the password.

That happens if the configuration parameter `encryption.tpm_luks_open` is set AND the system supports TPM unlocking. If that the case, the `TPM_FDE` encryption method is used without even asking the user. In any other case, the encryption method specified at the configuration parameter `encryption.method` is used.

That's all quite obscure, the users don't know whether TPM-based unlocking is going to be configured. Not even if it's possible to configure it or not.

## Solution

This pull request introduces some changes in how the whole thing is managed.

Now if the system or the distribution being installed don't support TPM-based decryption, the encryption method `LUKS2` is used and nothing is shown in the UI.

tpm_not_available

So no big change for the user except the fact that now LUKS2 with PBKDF2 as derivation function is the default for all distributions (it's a pretty sensible default for distributions based on Grub2 at 2024).

But if the system and the distribution both support to configure TPM-based opening of the LUKS devices, the user can choose between the `TPM_FDE` and the  `LUKS2` encryption methods via a checkbox shown in the UI.

attempt

The default encryption method (and thus, the default value of the checkbox) is configured per-product at `encryption.method`. If the value there is `"tpm_fde"` but the system does not support such a method (eg. there is no TPMv2 chip), Agama will use the default encryption method (`LUKS2`) as fallback.

Additionally, to make sure the user does not overlook the need to boot the machine directly to the new system in order to complete the setup, the following warning has been added to the page at the end of the installation process.

finish-tpm-b2

Expanded version:

finish-tpm-a2

Of course, if TPM encryption was not used, the hint is not there.

finish-no-tpm2

## Testing

- Tested manually
- New unit tests for the `InstallationFinished` page
- No tests added to the storage page, since it's going to [heavily change](https://github.com/openSUSE/agama/discussions/982) soon

### DIFF
--- a/service/lib/agama/storage/proposal_settings_reader.rb
+++ b/service/lib/agama/storage/proposal_settings_reader.rb
@@ -69,24 +69,19 @@ module Agama
       # @param settings [Agama::Storage::ProposalSettings]
       # @param encryption [Hash]
       def encryption_reader(settings, encryption)
-        method =
-          if try_tpm_fde?(encryption)
-            Y2Storage::EncryptionMethod::TPM_FDE
-          else
-            Y2Storage::EncryptionMethod.find(encryption.fetch("method", ""))
-          end
+        method = Y2Storage::EncryptionMethod.find(encryption.fetch("method", ""))
         pbkd_function = Y2Storage::PbkdFunction.find(encryption.fetch("pbkd_function", ""))
 
-        settings.encryption.method = method if method
+        settings.encryption.method = method if available_method?(method)
         settings.encryption.pbkd_function = pbkd_function if pbkd_function
       end
 
-      # @param encryption [Hash]
+      # @param method [Y2Storage::EncryptionMethod::Base, nil]
       # @return [Boolean]
-      def try_tpm_fde?(encryption)
-        return false unless encryption["tpm_luks_open"] == true
+      def available_method?(method)
+        return false unless method
 
-        Y2Storage::EncryptionMethod::TPM_FDE.possible?
+        EncryptionSettings.available_methods.include?(method)
       end
 
       # @param settings [Agama::Storage::ProposalSettings]
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2022-2023] SUSE LLC
+ * Copyright (c) [2022-2024] SUSE LLC
  *
  * All Rights Reserved.
  *
@@ -45,6 +45,17 @@ const ZFCP_CONTROLLERS_NAMESPACE = "/org/opensuse/Agama/Storage1/zfcp_controller
 const ZFCP_CONTROLLER_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Controller";
 const ZFCP_DISKS_NAMESPACE = "/org/opensuse/Agama/Storage1/zfcp_disks";
 const ZFCP_DISK_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Disk";
+
+/**
+ * Enum for the encryption method values
+ *
+ * @readonly
+ * @enum { string }
+ */
+const EncryptionMethods = Object.freeze({
+  LUKS2: "luks2",
+  TPM: "tpm_fde"
+});
 
 /**
  * Removes properties with undefined value
@@ -1433,4 +1444,4 @@ class StorageClient extends WithIssues(
   ), STORAGE_OBJECT
 ) { }
 
-export { StorageClient };
+export { StorageClient, EncryptionMethods };
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -227,6 +227,7 @@ class ProposalManager {
    * @typedef {object} ProposalSettings
    * @property {string} bootDevice
    * @property {string} encryptionPassword
+   * @property {string} encryptionMethod
    * @property {boolean} lvm
    * @property {string} spacePolicy
    * @property {string[]} systemVGDevices
@@ -281,6 +282,16 @@ class ProposalManager {
   async getProductMountPoints() {
     const proxy = await this.proxies.proposalCalculator;
     return proxy.ProductMountPoints;
+  }
+
+  /**
+   * Gets the list of encryption methods accepted by the proposal
+   *
+   * @returns {Promise<string[]>}
+   */
+  async getEncryptionMethods() {
+    const proxy = await this.proxies.proposalCalculator;
+    return proxy.EncryptionMethods;
   }
 
   /**
@@ -345,6 +356,7 @@ class ProposalManager {
           spacePolicy: proxy.SpacePolicy,
           systemVGDevices: proxy.SystemVGDevices,
           encryptionPassword: proxy.EncryptionPassword,
+          encryptionMethod: proxy.EncryptionMethod,
           volumes: proxy.Volumes.map(this.buildVolume),
           // NOTE: strictly speaking, installation devices does not belong to the settings. It
           // should be a separate method instead of an attribute in the settings object.
@@ -365,7 +377,7 @@ class ProposalManager {
    * @param {ProposalSettings} settings
    * @returns {Promise<number>} 0 on success, 1 on failure
    */
-  async calculate({ bootDevice, encryptionPassword, lvm, spacePolicy, systemVGDevices, volumes }) {
+  async calculate({ bootDevice, encryptionPassword, encryptionMethod, lvm, spacePolicy, systemVGDevices, volumes }) {
     const dbusVolume = (volume) => {
       return removeUndefinedCockpitProperties({
         MountPath: { t: "s", v: volume.mountPath },
@@ -381,6 +393,7 @@ class ProposalManager {
     const settings = removeUndefinedCockpitProperties({
       BootDevice: { t: "s", v: bootDevice },
       EncryptionPassword: { t: "s", v: encryptionPassword },
+      EncryptionMethod: { t: "s", v: encryptionMethod },
       LVM: { t: "b", v: lvm },
       SpacePolicy: { t: "s", v: spacePolicy },
       SystemVGDevices: { t: "as", v: systemVGDevices },
--- a/web/src/components/core/InstallationFinished.jsx
+++ b/web/src/components/core/InstallationFinished.jsx
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2022] SUSE LLC
+ * Copyright (c) [2022-2024] SUSE LLC
  *
  * All Rights Reserved.
  *
@@ -31,16 +31,17 @@ import {
   HintBody,
 } from "@patternfly/react-core";
 
-import { Page, If } from "~/components/core";
-import { Center, Icon } from "~/components/layout";
+import { If, Page } from "~/components/core";
+import { Icon } from "~/components/layout";
 import { useInstallerClient } from "~/context/installer";
+import { EncryptionMethods } from "~/client/storage";
 import { _ } from "~/i18n";
 
 const TpmHint = () => {
   const [isExpanded, setIsExpanded] = useState(false);
 
   return (
-    <Hint>
+    <Hint className="tpm-hint">
       <HintBody>
         <ExpandableSection
           isExpanded={isExpanded}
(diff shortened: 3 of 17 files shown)


web/src/client/storage.js
```diff
@@ -46,6 +46,17 @@ const ZFCP_CONTROLLER_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Controller";
...
+ * Enum for the encryption method values
+ *
+ * @readonly
+ * @enum { string }
+ */
+const EncryptionMethods = Object.freeze({
+  LUKS2: "luks2",
+  TPM: "tpm_fde"
+});
+
```
dgdavid, line 59: As I wrote in the commit message, following here the same approac as the current network code. We could, however, goes for what we did with _client/phase_ or _client/status_ instead and move it to a _client/storage-encryption-methods.js_ or _client/storage/encryption-methods.js_ 

```js
export const LUKS2 = "luks2";
export const TPM = "tpm_fde";

export default {
  LUKS2,
  TPM
};
```
and then import needed values 

```js
import { TPM } from "~/client/storage/encryption-methods"
``` 

```js
import * as EncriptionMethods from "~/client/storage/encryption-methods"
```

No strong/clear opinion. Something to think about for future enhancements.

web/src/components/core/InstallationFinished.jsx
```diff
@@ -26,32 +26,61 @@ import {
...
+          isExpanded={isExpanded}
+          onToggle={() => setIsExpanded(!isExpanded)}
+          toggleText={
+            <strong>
+              {_("Make sure to boot directly to the new system to finish the configuration of TPM-based decryption.")}
+            </strong>
+          }
+        >
+          <Text>
+            {_("The final step to configure the TPM to automatically open devices will take place during the first boot of the new system. For that to work, the machine needs to boot directly to the new boot loader. If a local media was used to run this installer, make sure is removed before next boot.")}
```
dgdavid, line 56: What a long line :stuck_out_tongue_winking_eye:

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -412,42 +421,63 @@ const EncryptionPasswordForm = ({
...
       />
+      <If
+        condition={methods.includes(tpmId)}
+        then={
+          <Checkbox
+            id="encryption_method"
+            label={_("Use the TPM to decrypt automatically on each boot")}
+            description={
+              // TRANSLATORS: The word 'directly' is key here. For example, booting to the installer media and then choosing
+              // 'Boot from Hard Disk' from there will not work. Keep it sort (this is a hint in a form) but keep it clear.
```
dgdavid, line 432: :+1: :ok_hand:

web/src/components/core/InstallationFinished.jsx
```diff
@@ -26,32 +26,61 @@ import {
...
 import { Center, Icon } from "~/components/layout";
 import { useInstallerClient } from "~/context/installer";
+import { EncryptionMethods } from "~/client/storage";
 import { _ } from "~/i18n";
 
+const TpmHint = () => {
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  return (
+    <Hint className="tpm-hint">
```
dgdavid, line 44: As you can see, I solved the `width` (a.k.a. `block-size`) issue with a temporary CSS class.

I have doubts (too long story for writing it here and now) that this combination of a Hint and ExpandableSection is the best approach for this kind of reminders. But I have neither, a clear idea how to make it better now nor the time for cooking some proposals. IMHO, it is good as it is for releasing it by now. Improvements can wait a bit.

web/src/components/core/InstallationFinished.jsx
```diff
@@ -67,19 +96,24 @@ function InstallationFinished() {
...
-                  ? _("At this point you can power off the machine.")
-                  : _("At this point you can reboot the machine to log in to the new system.")
-              }
+              <If
+                condition={usingIguana}
+                then={_("At this point you can power off the machine.")}
+                else={_("At this point you can reboot the machine to log in to the new system.")}
+              />
             </Text>
-            <Text>{_("Have a lot of fun! Your openSUSE Development Team.")}</Text>
```
dgdavid, line 76: :+1:

web/src/components/core/InstallationFinished.jsx
ancorgs: Do we have any rule about avoiding them?

For example, in the Ruby code we always break long translatable texts using `\`. Do we have any similar guideline for long texts in `jsx` that we want to keep as a single string for translations?

dgdavid: We use the `\` too, although it is ugly because we have to "break" the indentation. See an example at  https://github.com/openSUSE/agama/blob/df72d49fed7e05463d2e0666d4676144c45d8512/web/src/components/storage/VolumeForm.jsx#L410C14-L411

That said, I have no strong preference. Sorry for the NP comment.

service/lib/agama/storage/proposal_settings_reader.rb
```diff
@@ -69,24 +69,19 @@ def lvm_reader(settings, value)
...
         settings.encryption.pbkd_function = pbkd_function if pbkd_function
       end
 
-      # @param encryption [Hash]
+      # @param method [Y2Storage::EncryptionMethod::Base, nil]
       # @return [Boolean]
-      def try_tpm_fde?(encryption)
-        return false unless encryption["tpm_luks_open"] == true
+      def available_method?(method)
+        return false unless method
```
joseivanlopez, line 82: NP: this condition should be already covered by `#include?`.

web/src/components/core/InstallationFinished.jsx
```diff
@@ -26,60 +26,92 @@ import {
...
-    async function getIguana() {
-      const ret = await client.manager.useIguana();
-      setIguana(ret);
+    async function preparePage() {
+      const iguana = await client.manager.useIguana();
+      // FIXME: This logic should likely not be placed here, it's too coupled to storage internals.
+      // Something to fix when this whole page is refactored in a (hopefully near) future.
+      const { settings: { encryptionMethod } } = await client.storage.proposal.getResult();
+      setUsingIguana(iguana);
+      setUsingTpm(encryptionMethod === EncryptionMethods.TPM);
```
ancorgs, line 79: I just found a bug here. If you choose Dolomite, the default encryption method is TPM and, thus, the warning is displayed even if encryption is disabled. `usingTPM` should first check if encryption is used and only in that case check the encryption method.

web/src/components/core/InstallationFinished.jsx
```diff
@@ -26,60 +26,92 @@ import {
...
+          isExpanded={isExpanded}
+          onToggle={() => setIsExpanded(!isExpanded)}
+          toggleText={
+            <strong>
+              {_("Make sure to boot directly to the new system to finish the configuration of TPM-based decryption.")}
+            </strong>
+          }
+        >
+          <Text>
+            {_("The final step to configure the TPM to automatically open devices will take place during the first boot of the new system. For that to work, the machine needs to boot directly to the new boot loader. If a local media was used to run this installer, make sure is removed before next boot.")}
```
joseivanlopez, line 56: NP: I am not sure if the tip is clear enough. I mean "Make sure to boot directly to the new system ...". Me as user can ask "why am I not going to boot directly? I only have to reboot".

I would try to directly indicate an action, for example: "If an installation media was used for running this installer , then remove it before rebooting. Otherwise the TPM-based decryption will not be correctly configured".

And the details would explain that TPM requires to directly boot the new system to finish the configuration. And maybe it could indicate what happens if that is not done.

web/src/components/core/InstallationFinished.jsx
joseivanlopez: NP: I am not sure if the tip is clear enough. I mean "Make sure to boot directly to the new system ...". Me as user can ask "why am I not going to boot directly? I only have to reboot".

I would try to give a more explicit message, For example:

~~~
To finish the configuration of the TPM-based decryption, the new installed system must be directly run in the next boot:

* If an installation media was used for running this installer, then remove it before booting.
* If in your boot menu contains several options, then make sure you select the correct system.
~~~

And the details could explain what happens if the configuration is not correctly finished.

web/src/components/storage/ProposalPage.jsx
```diff
@@ -54,6 +54,11 @@ const reducer = (state, action) => {
       return { ...state, availableDevices };
     }
 
+    case "UPDATE_ENCRYPTION_METHODS": {
+      const { encryptionMethods } = action.payload;
+      return { ...state, encryptionMethods };
```
joseivanlopez, line 59: Please, add `encryptionMethods` to the `initialState` too.

web/src/components/core/InstallationFinished.jsx
dgdavid: Hopefully fixed at with changes made in e01933799dba67a22ac2db925dcbfbf02b00c056

web/src/components/core/InstallationFinished.jsx
dgdavid: Good point, but... 


> And the details could explain what happens if the configuration is not correctly finished.

Reinforce my thoughts that tat "expandable hint" component does not fit good here. 

If your suggestion (which I like) is meant to be also the "link" to see/hide these details... that's definetly too much.

dgdavid: Note that we already knew that we were somehow miss-using that component there but having in mind that next iterations should take care of that page. But if we're going to extend that message so long... it deserves a better design now.

web/src/components/core/InstallationFinished.jsx
dgdavid: Actually it was not fixed when I wrote the comment above. See https://github.com/openSUSE/agama/pull/995/commits/30fe3bcda8647d02502da1f3c3bc5cdf01cf74ba

joseivanlopez (commented):
Please, add changelog.

web/src/components/core/InstallationFinished.test.jsx
```diff
@@ -60,4 +72,47 @@ describe("InstallationFinished", () => {
...
+
+      it("does not show the TPM reminder", async () => {
+        const { user } = installerRender(<InstallationFinished />);
+        // Forcing the test to slow down a bit with a fake user interaction
+        // because actually the reminder will be not rendered immediately
+        // making the queryAllByText to produce a false positive if triggered
+        // too early here.
+        const congratsText = screen.getByText("Congratulations!");
+        await user.click(congratsText);
+        expect(screen.queryAllByText(/TPM/)).toHaveLength(0);
```
dgdavid, line 101: BTW, today I _learned_ that another option for doing this might be to add a helper function in test-uitls. Something like `slowDown`, `smallWait` or whatever other name suits better

```js

const shortBreak = async (ms = 500) => new Promise(resolve => setTimeout(resolve, ms));
```


and then in the test

```js
import { shortBreak } from "test-utils";

...

await shortBreak();
```

I haven't tested it, just got _inspiration_ from [react-spectrum](https://github.com/adobe/react-spectrum/blob/55aca809c4ce039efe672aaced8c77553d5c03f5/packages/%40react-spectrum/list/stories/ListView.stories.tsx#L327-L328)

dgdavid (approved):

web/src/components/core/InstallationFinished.test.jsx
```diff
@@ -60,4 +72,47 @@ describe("InstallationFinished", () => {
...
+
+  describe("when TPM is not set as encryption method", () => {
+    it("does not show the TPM reminder", async () => {
+      const { user } = installerRender(<InstallationFinished />);
+      // Forcing the test to slow down a bit with a fake user interaction
+      // because actually the reminder will be not rendered immediately
+      // making the queryAllByText to produce a false positive if triggered
+      // too early here.
+      const congratsText = screen.getByText("Congratulations!");
+      await user.click(congratsText);
```
dgdavid, line 114: BTW, today I _learned_ that another option for doing this might be to add a helper function in test-uitls. Something like `slowDown`, `smallWait` or whatever other name suits better

```js

const shortBreak = async (ms = 500) => new Promise(resolve => setTimeout(resolve, ms));
```


and then in the test

```js
import { shortBreak } from "test-utils";

...

await shortBreak();
```

I haven't tested it, just got _inspiration_ from [react-spectrum](https://github.com/adobe/react-spectrum/blob/55aca809c4ce039efe672aaced8c77553d5c03f5/packages/%40react-spectrum/list/stories/ListView.stories.tsx#L327-L328)

joseivanlopez (approved):
LGTM

PR #997, opened 2024-01-15
Preserve network configuration provided via kernel boot line

mchf:
## Problem

Network configuration provided by kernel boot line is lost

- [*bsc1210541*](https://bugzilla.suse.com/show_bug.cgi?id=1210541)

## Solution

Even NM's runtime configuration is copied to the target at the end of installation

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/network.rb
+++ b/service/lib/agama/network.rb
@@ -52,6 +52,7 @@ module Agama
     attr_reader :logger
 
     ETC_NM_DIR = "/etc/NetworkManager"
+    RUN_NM_DIR = "/run/NetworkManager"
     private_constant :ETC_NM_DIR
 
     def enable_service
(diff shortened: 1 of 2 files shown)


teclator (changes requested):
By now it is good enough... so, please add the changelog and mark as ready for review for merging it ASAP.

service/lib/agama/network.rb
```diff
@@ -68,6 +69,13 @@ def enable_service
     def copy_files
       return unless Dir.exist?(ETC_NM_DIR)
 
+      # runtime configuration is copied first, so in case of later modification
+      # on same interface it gets overwriten.
```
teclator, line 73:
```suggestion
      # on same interface it gets overwriten (bsc#1210541).
```

teclator (approved):
LGTM

PR #999, opened 2024-01-16
Update cockpit.js to version 307

imobachgs:
Update `cockpit.js` files to version 307, which we ship in the Live medium.


lslezak (approved):
LGTM

Issue #1001, opened 2024-01-16
The storage section is always like "loading"


imobachgs:
The storage section is blocked as it was loading. Reloading is the only way to get it working. Navigating to the Storage section does not help.

* [JavaScript console messages](https://github.com/openSUSE/agama/files/13950728/console-export-2024-1-16_14-43-21.txt)
* [agama.out.log](https://github.com/openSUSE/agama/files/13950752/agama.out.log)

Reproducible with agama-live.x86_64-7.0.0-openSUSE-Build2.84.iso.

imobachgs:
Fixed by #1003.

PR #1002, opened 2024-01-16
Do not crash when cannot get the list of issues

imobachgs:
## Problem

See #1001.

## Solution

Do not crash even if it is impossible to get the list of issues. It does not fix the root of the problem, but prevents the crashing.

## Testing

- Tested manually


imobachgs:
Closed in favor of #1003.

PR #1004, opened 2024-01-17
Package availability

joseivanlopez:
## Problem

The call to the sotware service to check for the availability of a package is mocked and always returns true, assuming the package is always available. Of course, the availability of a package depends of the currently selected product. 

## Solution

Perform a D-Bus call to the software service in order to know if a package is available.

Note: This change exposes a problem in our services. Asking for the product availability should be done once the software proposal is done, otherwise the result is not reliable at all, see https://github.com/openSUSE/agama/issues/1005. For example, the TPM option in the storage settings could not appear until the software service has finished, see https://github.com/openSUSE/agama/pull/995.

## Testing

* Added new unit tests
* Tested manually

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2021-2023] SUSE LLC
+# Copyright (c) [2021-2024] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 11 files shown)


dgdavid (approved):
LGTM

service/lib/agama/software/manager.rb
```diff
@@ -279,14 +279,23 @@ def on_selected_patterns_change(&block)
...
         on_target { Yast::Package.Installed(name, target: :system) }
       end
 
+      # Determines whether a package is available.
+      #
+      # @param name [String] Package name
+      # @return [Boolean]
+      def package_available?(name)
+        # Beware: apart from true and false, Available can return nil if things go wrong.
+        on_local { !!Yast::Package.Available(name) }
```
lslezak, line 296: Be careful that `Yast::Package.Available` checks for the RPM Provides. That means the package `name` might not be available, but if there is some other package providing it then it will return `true`.

So it depends what we actually expect from the API. Maybe have two separate calls? One for package name and the other for provides?

This could cause some unexpected problems later...

PR #1006, opened 2024-01-18
Adapt the web UI to use the Agama network service

imobachgs:
## Problem

The web UI still talks directly to NetworkManager to set up network connections.

## Solution

This is the first set of changes aimed to stop using the NetworkManager service from the web UI. Instead, it should talk to Agama's network service (which relies on NetworkManager itself).  

This pull request adapts the web UI to use Agama's network service to write the changes (including connecting to wireless devices). However, the status information is still retrieved from NetworkManager because those bits are missing from our API.

During the process, we needed to introduce a few changes to Agama's network service:

* Use UUIDs to identify connections instead of the Id (the name).
* Write only changed network connections.
* Rollback to the previous state if some connection cannot be updated.
* Better error handling at D-Bus level.

## Future

* Improve (almost non-existent) error handling.
* Retrieve the status information from Agama's network service.

## Testing

- *Adapted unit tests*
- *Tested manually*


web/src/client/network/index.js
```diff
@@ -156,48 +187,204 @@ o  *   NetworkManagerAdapter.
...
   /**
    * Adds a new connection
    *
+   * If a connection with the given ID already exists, it updates such a
+   * connection.
+   *
    * @param {Connection} connection - Connection to add
+   * @return {Promise<Connection>} the added connection
    */
   async addConnection(connection) {
```
joseivanlopez, line 214: NP: What do think about renaming this method to `AddOrUpdateConnection` ?

web/src/client/network/index.js
```diff
@@ -156,48 +187,204 @@ o  *   NetworkManagerAdapter.
...
 
   /**
    * Updates the connection
    *
-   * It uses the 'path' to match the connection in the backend.
+   * It uses the 'uuid' to match the connection in the backend.
    *
    * @param {Connection} connection - Connection to update
+   * @return {Promise<boolean>} - the promise resolves to true if the connection
+   *   was successfully updated and to false it it does not exist.
```
joseivanlopez, line 328: Typo: it it

web/src/client/network/model.js
```diff
@@ -60,6 +60,17 @@ const SecurityProtocols = Object.freeze({
   _8021X: "802.1X"
 });
 
+// security protocols
+// const AgamaSecurityProtocols = Object.freeze({
```
joseivanlopez, line 64: Leftover?

joseivanlopez (commented):
Looks great! Only minor comments.

web/src/client/network/index.js
imobachgs: I thought about that, but it is weird having an `addOrUpdateConnection` and `updateConnection.` Perhaps we might need to merge them and have something like `saveConnection` only. I will think about that.

web/src/client/network/model.js
imobachgs: Ouch, you are right.

joseivanlopez (approved):
LGTM

PR #1008, opened 2024-01-19
Add bridge model

cfconrad:
## Problem

Bridge interface wasn't implemented 

## Solution

 * Implement bridge interface type with the subset of configuration parameters needed for wicked migration (see https://github.com/jcronenberg/agama/pull/66). 

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-dbus-server/src/network/nm/dbus.rs
+++ b/rust/agama-dbus-server/src/network/nm/dbus.rs
@@ -20,6 +20,8 @@ const WIRELESS_SECURITY_KEY: &str = "802-11-wireless-security";
 const LOOPBACK_KEY: &str = "loopback";
 const DUMMY_KEY: &str = "dummy";
 const VLAN_KEY: &str = "vlan";
+const BRIDGE_KEY: &str = "bridge";
+const BRIDGE_PORT_KEY: &str = "bridge-port";
 
 /// Converts a connection struct into a HashMap that can be sent over D-Bus.
 ///
@@ -39,14 +41,22 @@ pub fn connection_to_dbus<'a>(
     }
 
     if let Some(controller) = controller {
-        connection_dbus.insert("slave-type", "bond".into()); // TODO: only 'bond' is supported
+        let slave_type = match controller.config {
+            ConnectionConfig::Bond(_) => BOND_KEY,
+            ConnectionConfig::Bridge(_) => BRIDGE_KEY,
+            _ => {
+                log::error!("Controller {} has unhandled config type", controller.id);
+                ""
+            }
+        };
+        connection_dbus.insert("slave-type", slave_type.into());
         let master = controller
             .interface
             .as_deref()
             .unwrap_or(controller.id.as_str());
         connection_dbus.insert("master", master.into());
     } else {
-        connection_dbus.insert("slave-type", "".into()); // TODO: only 'bond' is supported
+        connection_dbus.insert("slave-type", "".into());
         connection_dbus.insert("master", "".into());
     }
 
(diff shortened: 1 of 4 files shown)


imobachgs (approved):

imobachgs:
Oh, I forgot. Please, could you add an entry to the [agama-cli.changes](https://github.com/openSUSE/agama/blob/master/rust/package/agama-cli.changes) file?

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -38,7 +40,12 @@ pub fn connection_to_dbus<'a>(
     }
 
     if let Some(controller) = controller {
-        connection_dbus.insert("slave-type", "bond".into()); // TODO: only 'bond' is supported
+        let slave_type = match controller.config {
+            ConnectionConfig::Bond(_) => BOND_KEY,
+            ConnectionConfig::Bridge(_) => BRIDGE_KEY,
+            _ => "", //XXX maybe we should throw an error here?
```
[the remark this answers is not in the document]
cfconrad: Ok, I'm logging now this error:
```
log::error!("Controller {} has unhandled config type", controller.id);
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -313,6 +343,98 @@ fn bond_config_to_dbus(config: &BondConfig) -> HashMap<&str, zvariant::Value> {
...
+    }
+
+    hash
+}
+
+fn bridge_config_from_dbus(conn: &OwnedNestedHash) -> Option<BridgeConfig> {
+    let Some(bridge) = conn.get(BRIDGE_KEY) else {
+        return None;
+    };
+    let mut bc = BridgeConfig {
```
imobachgs, line 373: I would move this initialization below the next if:

```rust
let Some(stp) = bridge.get("stp") else {
  return None;
}

let mut bc = BridgeConfig {
   stp: *stp.downcast_ref::<bool>()?,
   ..Default::default()
}

File: rust/agama-dbus-server/src/network/nm/dbus.rs [hunk 3]
```diff
@@ -313,6 +343,98 @@ fn bond_config_to_dbus(config: &BondConfig) -> HashMap<&str, zvariant::Value> {
...
+
+    hash
+}
+
+fn bridge_port_config_from_dbus(conn: &OwnedNestedHash) -> Option<BridgePortConfig> {
+    let Some(bridge_port) = conn.get(BRIDGE_PORT_KEY) else {
+        return None;
+    };
+
+    let mut bpc = BridgePortConfig {
```
imobachgs, line 423:
```rust
let mut bpc = BridgePortConfig::default();
```

rust/agama-dbus-server/src/network/nm/dbus.rs
```diff
@@ -313,6 +343,98 @@ fn bond_config_to_dbus(config: &BondConfig) -> HashMap<&str, zvariant::Value> {
...
+        return None;
+    };
+    bc.stp = *stp.downcast_ref::<bool>()?;
+
+    if let Some(prio) = bridge.get("priority") {
+        bc.priority = Some(*prio.downcast_ref::<u32>()?);
+    }
+
+    if let Some(fwd_delay) = bridge.get("forward-delay") {
+        bc.forward_delay = Some(*fwd_delay.downcast_ref::<u32>()?);
```
imobachgs, line 387: I know we have followed the same approach everywhere, but I wonder whether we want to return `None` if we fail to downcast one of those tiny fields. I have no opinion about that. But, sure, nothing to change in this PR.

cfconrad: valid point, but we need to log the variable somehow, as it might be because of some API change?!
Another thing could be to use some macro in this regards. As it's also always the same but just exchanging the name in the hash and the name of the struct.

imobachgs: Good idea, we could explore that path.

imobachgs (approved):

PR #1012, opened 2024-01-24
[web] Adds a core/Selector component

dgdavid:
Adds a core/Selector component and uses it to render suitable selectors across Agama codebase instead of defining almost the same code repeatedly.

Note that, while it was originally intended to have a fully accessible selector, time and priorities constraints (among other reasons) forced to _just centralize_ the repetitive code for rendering selectors with an small approach change.

Once priorities permit, work on a better, fully accessible selector will be resumed.

---

Adapted selectors:

* L10n: Language, Keymap, Timezone
* Storage: Disk selectors, Space policy
* Product: product selection

<details>

----

| Before | After |
|-|-|
| Screen Shot 2024-01-25 at 11 04 50 | Screen Shot 2024-01-25 at 11 03 15 |
| Screen Shot 2024-01-25 at 11 06 13 | Screen Shot 2024-01-25 at 11 05 37 |
| Screen Shot 2024-01-25 at 11 06 46 | Screen Shot 2024-01-25 at 10 58 05 |
| Screen Shot 2024-02-10 at 01 01 38 | Screen Shot 2024-02-10 at 01 00 03 |
| Screen Shot 2024-02-10 at 01 01 32 | Screen Shot 2024-02-10 at 00 59 54 |
| Screen Shot 2024-02-10 at 01 01 48 | Screen Shot 2024-02-10 at 01 00 23 |
| Screen Shot 2024-02-14 at 12 22 44 | Screen Shot 2024-02-10 at 01 29 16 |

</details>

---

Related to https://github.com/openSUSE/agama/discussions/786#discussioncomment-8209053

### DIFF
--- a/web/src/components/core/Selector.jsx
+++ b/web/src/components/core/Selector.jsx
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) [2024] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+// @ts-check
+import React from 'react';
+
+/**
+ * Convenient component for letting the consumer build the selector options
+ */
+const Option = ({ children }) => children;
+
+/**
+ * Internal component for building the selector options
+ *
+ * FIXME: check if the used aria-labelledby is the way to go for making the
+ * <input /> component accessible.
+ */
+const Item = ({ id, type, isSelected = false, onClick, children }) => {
+  const whenClicked = () => onClick(id);
+  const rowId = `item-${id}`;
+
+  return (
+    <li
+      id={rowId}
+      role="row"
+      onClick={whenClicked}
+      aria-selected={isSelected || undefined}
+    >
+      <div role="gridcell">
+        <input
+          type={type}
+          checked={isSelected}
+          onChange={whenClicked}
+          aria-labelledby={rowId}
+        />
+        { children }
+      </div>
+    </li>
+  );
+};
+
+/**
+ * Agama component for building a selector
+ *
+ * @example <caption>Usage example</caption>
+ *   <Selector
+ *     isMultiple
+ *     aria-label="Available locales"
+ *     selectedIds={["es_ES", "en_GB"]}
+ *     onSelectionChange={(selection) => changePreferredLocales(selection)}
+ *   >
+ *     <Selector.Option id={"es_ES"}>Spanish - Spain</Selector.Option>
+ *     <Selector.Option id={"cs_CZ"}>Czech - Czechia</Selector.Option>
+ *     <Selector.Option id={"de_DE"}>German - Germany</Selector.Option>
+ *     <Selector.Option id={"en_GB"}>English - United Kingdom</Selector.Option>
+ *   </Selector>
+ *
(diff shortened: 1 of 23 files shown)


web/src/components/l10n/LocaleSelector.test.jsx
```diff
@@ -0,0 +1,77 @@
...
+    within(selector).getByRole("row", { name: "Spanish Spain es_ES" });
+    within(selector).getByRole("row", { name: "English United States en_US" });
+  });
+
+  it("renders an input for filtering locales", async () => {
+    const { user } = plainRender(
+      <LocaleSelector locales={locales} aria-label="Available locales" />
+    );
+
+    const filterInput = screen.getByRole("search");
```
dgdavid, line 54: :warning: **NOTE:** This made me realize that Agama is doing wrong in terms of accessibility and usability by not using proper `<label>`s for no other reason than _visual aesthetic_. This example is working just because there is only one input with the "search" role, but it would fail trying to get it by its name/label too:

```diff
-   const filterInput = screen.getByRole("search");
+   const filterInput = screen.getByRole("search", { name: /Filter by language/} );
```

Of course, taking a look at the UI **for us** and thanks to the placeholder it's clear the input functionality

Screenshot from 2024-01-24 19-41-52

but it might not be as obvious to other users.

Something to work on. But in a different PR.

---

From W3c (https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text)

> While placeholder text provides valuable guidance for many users, **placeholder text is not a replacement for labels**

From MDN (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-placeholder)

> **Warning:** Using a placeholder instead of a visible label harms accessibility and usability for many users including **older users and users with cognitive, mobility, fine motor skill and vision impairments**. Labels are better: they are always visible and they provide for a larger hit region to focus on the control.
> ...
> **Note:** Placeholders should only be used to show an example of the type of data that should be entered into a form; they don't replace a proper label. 


Other links, 

https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Placeholder_Research

---

/cc @joseivanlopez, @imobachgs

joseivanlopez: Good point! Definitely, "innovative" UI approaches are quite ... dangerous :)

joseivanlopez:
This PR looks quite good. Is there anything missing?

web/src/components/core/Selector.jsx
```diff
@@ -0,0 +1,144 @@
...
+ * Agama component for building a selector
+ *
+ * @example <caption>Usage example</caption>
+ *   <Selector
+ *     isMultiple
+ *     aria-label="Available locales"
+ *     selectedIds={["es_ES", "en_GB"]}
+ *     onSelectionChange={(selection) => changePreferredLocales(selection)}
+ *   >
+ *     <Selector.Option id={"es_ES"}>Spanish - Spain</Selector.Option>
```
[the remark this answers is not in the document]
dgdavid: Nope with this approach. I'd like to be ready for having a not controlled selector.

dgdavid:
> This PR looks quite good. Is there anything missing?

There is work ahead. Last thing I was doing was adapting the Storage selector. When finished, I would like to work the internal a bit for not using the React.Children API. There is some information in commit messages.

web/src/components/core/Selector.jsx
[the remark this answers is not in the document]
dgdavid: What is more, I believe I will drop the Selector.Option component in favor of a prop. I will explain why in the commit.

web/src/components/core/Selector.jsx
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @param {object} props - component props
+ * @param {string} [props.id] - Id attribute for selector.
+ * @param {boolean} [props.isMultiple=false] - Whether the selector should allow multiple selection.
+ * @param {Array<object>} props.options=[] - Item objects to build options.
+ * @param {function} props.renderOption=noop - Function used for rendering options.
+ * @param {string} [props.optionIdKey="id"] - Key used for retrieve options id.
+ * @param {Array<*>} [props.selectedIds=[]] - Identifiers for selected options.
+ * @param {onSelectionChangeCallback} [props.onSelectionChange=noop] - Callback to be called when the selection changes.
+ * @param {React.ReactNode} props.children - Selector options
```
dgdavid, line 65: Left over, I'll drop it in a while

joseivanlopez (approved):
LGTM

PR #1022, opened 2024-01-28
Auto suggest username

### LINKED ISSUE #112: auto-suggestion of username done differently
I love freedom (unlike in windows user generation) where I'm stuck initially with what they decide. However I was quite unhappy about the username generation in the old installer which usually generated luboskocman and I have to always do a correction otherwise I'd have to extensively use .ssh/config to override user etc.

This auto-suggestion feature is something I missed in the first-time user experience with the installer.

Fedora and other distros would suggest lkocman. Which would be my personal preference as well.

I think this should be our preferred generation as 99% of SUSE employees have such account anyway.
I'm not sure what the rules is for complex names like "Homer Jay Simpson" etc.

balsa-asanovic:
## Problem

As described in issue #112 this PR addresses the suggestion of username for user account creation based on full name input.

## Solution

Usernames are suggested based on the given full name in the following manner:

- Just the first word of the given name
- The first letter of the first word + all other full words
- The full first word + first letters of all other words
- First letters of all words except last + full last word
- All words of a name joined without white spaces

Also, all suggestions with less than 3 characters are filtered out.

This PR fixes #112

## Testing

- Manually tested
- Added unit tests in `web/src/components/users/utils.test.js`

## Screenshots

image


web/src/components/users/FirstUser.jsx
```diff
@@ -166,6 +172,26 @@ export default function FirstUser() {
     setFormValues({ ...formValues, [name]: value });
   };
 
+  const suggestUsernames = (fullName) => {
```
balsa-asanovic, line 175: I'll start immediately with some uncertainties :smile:

When the user types two words for the full name, these current suggestions work well.

However, if the user's full name has more than two words (names), the choice is not so clear.
People having longer full names is a common thing in some parts of the world (like Spain) and with the current approach to suggesting the username, Ronaldo Luís Nazário de Lima would be presented with the following choices:

image


The issues are apparent, the first suggestion is pretty long, the second might be as well and only the third one might be okay.

Should we maybe take only the first two words (names)?
Or maybe some other approach entirely, feel free to make a suggestion.

dgdavid: I see your point, but I fear that guessing which are the _right or expected parts_ is a bit difficult due to the fact on how different can be [personal names around the world](https://www.w3.org/International/questions/qa-personal-names) (long and old article).

Therefore, I wouldn't worry too much about them. They are just suggestions. Of course, the idea is to show up the more sensible ones, but we can't do so for every single user. I propose to below as first. Some of them are based on feedback received in the issue, some are not:


* Just the first word (the simplest approach)

  To make @kobliha [happy](https://github.com/openSUSE/agama/issues/112#issuecomment-1351026852)

  * Lubos Kocman: **lubos**
  * Imobach González Sosa: **imobach**
  * Ronaldo Luís Nazário de Lima: **ronaldo**

* First letter of first word plus second word

  It satisfies the [original request](https://github.com/openSUSE/agama/issues/112#issue-1189780790) from @lkocman 

  * Lubos Kocman: **lkocman**
  * Imobach González Sosa: **igonzalez**
  * Ronaldo Luís Nazário de Lima: **rluis**

* First letter of first word plus all other words

  It would suggest a username like the one @imobachgs [have at SUSE](https://github.com/openSUSE/agama/issues/112#issuecomment-1327196554)

  * Lubos Kocman: **lkocman**
  * Imobach González Sosa: **igonzalezsosa**
  * Ronaldo Luís Nazário de Lima: **rluisnazariodelima**

* First word plus first letter of the other words (maybe with a limit of just 2 more words?)

  * Lubos Kocman: **lubosk**
  * Imobach González Sosa: **imobachgs**
  * Ronaldo Luís Nazário de Lima: **ronaldolndl** (or **ronaldoln** if using a limit)

* First letter of all words except the last one, which would be complete

  * Lubos Kocman: **lkocman**
  * Imobach González Sosa: **igsosa**
  * Ronaldo Luís Nazário de Lima: **rlndlima**

* All words

  * Lubos Kocman: **luboskocman**
  * Imobach González Sosa: **imobachgonzalezsosa**
  * Ronaldo Luís Nazário de Lima: **ronaldoluiznaariodelima**

So, after removing duplicates, suggestions for these users will be, 

<table>
<thead>
<tr>
<th>Lubos Kocman</th>
<th>Imobach González Sosa</th>
<th>Ronaldo Luís Nazário de Lima</th>
</tr>
</thead>
<tbody>
<tr>
<td>

* lubos
* lkocman
* lubosk

</td>
<td>

* imobach
* igonzalez
* igonzalezsosa
* imobachgs
* igsosa
* imobachgonzalezsosa

</td>
<td>

* ronaldo
* rluis
* rluisnazariodelima
* ronaldolndl (or ronaldoln)
* rlndlima
* ronaldoluiznaariodelima

</td>
</tr>
</tdoby>
</table>

I do not have strong preference and ideas from others are welcome. I think that we can release the feature with these few (or even only the three or four first ones)  and let's see if we receive more suggestion requests.

BTW, as you can see in my examples above, I prefer not using the "special characters" in the suggested username (accents and friends). See https://stackoverflow.com/a/70288180 and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

web/src/components/users/FirstUser.jsx
```diff
@@ -220,6 +253,29 @@ export default function FirstUser() {
...
+                    <MenuList>
+                      { suggestUsernames(formValues.fullName).map((suggestion, index) => (
+                        <MenuItem
+                          key={index}
+                          itemId={index}
+                          onClick={() => { setInsideDropDown(false); setFormValues({ ...formValues, userName: suggestion }) }}
+                        >
+                          { /* TRANSLATORS: dropdown username suggestions */ }
+                          { _("Use suggested username ") }
+                          <strong>{suggestion}</strong>
```
dgdavid, line 273: I believe it's better to go for a `<b>` HTML element here instead. 

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#b_vs._strong
* https://html.spec.whatwg.org/#the-b-element
  > The **b** element represents a span of text to which attention is being drawn for utilitarian purposes **without conveying any extra importance** and with no implication of an alternate voice or mood

web/src/components/users/FirstUser.jsx
```diff
@@ -220,6 +253,29 @@ export default function FirstUser() {
...
+                  <MenuContent>
+                    <MenuList>
+                      { suggestUsernames(formValues.fullName).map((suggestion, index) => (
+                        <MenuItem
+                          key={index}
+                          itemId={index}
+                          onClick={() => { setInsideDropDown(false); setFormValues({ ...formValues, userName: suggestion }) }}
+                        >
+                          { /* TRANSLATORS: dropdown username suggestions */ }
+                          { _("Use suggested username ") }
```
dgdavid, line 272: I wouldn't add the space as part of a translation.

```
{_("Use suggested username")} <b>{suggestion}</b>
```

should work.

dgdavid (commented):
Hey @balsa-asanovic, 

Glad to see you around :wave: 

Sorry for the delay on reviewing this. Believe it or not, we're a little bit busier than usual. A lot of work ahead of us plus a bunch of long internal meetings lately :exploding_head: 

I've added a few comments about the feature, leaving a side the UI details since, judging by the screenshots, it looks reasonable as a starting point. We can polish some details before merging, though, once we have addressed the behavior.

Please, do not be discouraged if we take longer than usual to provide feedback. As said, we're a bit rushed off our feet these days.

busy

web/src/components/users/FirstUser.jsx
```diff
@@ -166,6 +172,26 @@ export default function FirstUser() {
     setFormValues({ ...formValues, [name]: value });
   };
 
+  const suggestUsernames = (fullName) => {
+    const cleanedName = fullName.replace(/[^\p{L}\p{N} ]/gu, "").toLowerCase();
```
dgdavid, line 176: At least for me, that line deserves an explanation / comment :exploding_head:

dgdavid: Also, I'm not sure if it pays off.

balsa-asanovic: > NP: At least for me, that line deserves an explanation / comment 🤯

Haha, indeed, this was weird for me as well.
This has almost identical effect as the regular expression `/[^a-zA-Z0-9 ]/g`, which cleans the string from all characters except alphanumeric and blank spaces.
However, when I just used this initially, non English characters (like š, ć, č, ...) would also get removed.

So, after googling for a bit, I found that since ECMAScript 2018 Unicode property escapes are available for regular expressions in JavaScript. So, the '\p{L}' matches any character from any language and '\p{N}' matches any numeric character.

> Also, I'm not sure if it pays off.

Of course, it is another thing should something like this be used at all. When I coded this, the idea was that users could enter some special characters in their full names, like ',', '.', ';', '-', '"', etc,... And I supposed that this isn't something that should be a part of the username.

dgdavid: Thanks a lot for the explanation!

If we keep it in the code, definitely it deserves a comment above it.

However, I think that "johnq.diaz" is a valid username for "John Q. Díaz" where "johnqdíaz" is not. If I'm right, we can sanitize the name in another way or with other regular expression. Either way, a comment in the code is more than welcome for future developers. 

Keep going!

balsa-asanovic:
Hey @dgdavid 

Sorry for a bit of a delay with these changes.

I've been on a vacation the week before last and then last week had some catching up to do.

I think I've addressed all your comments, so please check it out when you find the time.

I've modified the suggestUsername method in accordance with your suggestions and moved it to utils.js file

web/src/components/users/utils.js
```diff
@@ -0,0 +1,71 @@
...
+        suggestions[1] += namePart;
+      suggestions[2] += namePart;
+      suggestions[3] += namePart[0];
+      if (index === nameParts.length - 1)
+        suggestions[4] += namePart;
+      else
+        suggestions[4] += namePart[0];
+      suggestions[5] += namePart;
+    }
+  });
```
dgdavid, line 63: Most probably it's me, but I found this algorithm difficult to follow (even though it can be the most efficient one, I don't know)

What do you think about going for something more verbose like 

```js
//const name = "Maria del Carmen Fernandez Vega";
const suggestions = new Set();
const parts = cleanedName.split(/\s+/);
const firstLetters = parts.map(p => p[0]).join('');
const lastPosition = parts.length - 1;

const [firstPart, ...allExceptFirst] = parts;
const [firstLetter, ...allExceptFirstLetter] = firstLetters;
const lastPart = parts[lastPosition];

// Just the first part of the name
suggestions.add(firstPart);
// The first letter of the first part plus all other parts
suggestions.add(firstLetter + allExceptFirst.join(''));
// The first part plus the first letters of all other parts
suggestions.add(firstPart + allExceptFirstLetter.join(''));
// The first letters except the last one plus the last part
suggestions.add(firstLetters.substring(0, lastPosition) + lastPart);
// All parts without spaces
suggestions.add(parts.join(''));

// let's drop suggestions with less than 3 characters
suggestions.forEach(s => {
  if(s.length < 3) suggestions.delete(s)
});

return [...suggestions];
```

Just an idea. Any other suggestion or improvement about either, the one you proposed and this rough example are also welcome. Specially about naming :P

web/src/components/users/utils.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+/**
+ * Method which generates username suggestions based on first and last name.
```
dgdavid, line 23:
```suggestion
 * Method which generates username suggestions based on given full name.

File: web/src/components/users/utils.js (hunk 5 above)
Comment (dgdavid) [reply web/src/components/users/utils.js:63]: > (even though it can be the most efficient one, I don't know)

I was right and, although your approach contains more entries (e.g. _mdel_) than mine, it is more efficient in terms of performance, see https://jsben.ch/VuEsA.

So, we need another view here :)


@imobachgs, @jilopez. Friendly ping :)

File: web/src/components/users/utils.test.js [hunk 7]
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { suggestUsernames } from "./utils";
+
+describe('suggestUsernames', () => {
+  test('handles basic single name', () => {
+    expect(suggestUsernames('John')).toEqual(expect.arrayContaining(['john', 'j']));
```
dgdavid, line 26: I think `j` is a kind of useless suggestion. That's why I filtered out the suggestions with less than 3 characters in my above proposal.

web/src/components/users/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  test('handles basic two-part name', () => {
+    expect(suggestUsernames('John Doe')).toEqual(expect.arrayContaining(['john', 'jdoe', 'johnd', 'johndoe']));
+  });
+
+  test('handles name with middle initial', () => {
+    expect(suggestUsernames('John Q. Doe')).toEqual(expect.arrayContaining(['john', 'jq', 'jqdoe', 'johnqd', 'johnqdoe']));
```
dgdavid, line 34: Same here, `jq` does not look like a good suggestion to me.

web/src/components/users/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+  test('removes hyphens and apostrophes', () => {
+    expect(suggestUsernames("Jean-Luc O'Neill")).toEqual(expect.arrayContaining(['jeanluc', 'joneill', 'jeanluco', 'jeanluconeill']));
+  });
+
+  test('removes non-alphanumeric characters', () => {
+    expect(suggestUsernames("Anna*#& Maria$%^")).toEqual(expect.arrayContaining(['anna', 'amaria', 'annam', 'annamaria']));
+  });
+
+  test('handles long name with multiple parts', () => {
+    expect(suggestUsernames("Maria del Carmen Fernandez Vega")).toEqual(expect.arrayContaining(['maria', 'mdel', 'mdelcarmenfernandezvega', 'mariadcfv', 'mdcfvega', 'mariadelcarmenfernandezvega']));
```
dgdavid, line 50: Here we have an special case, "mdel". It has more than 3 characters, but still looking not good to me. By a coincidence, my proposal does not suggest it, probably because it is not iterating for specific name parts. Just playing with the first one, the last one and whatever it found in the middle.

dgdavid (commented):
@balsa-asanovic, 

Thanks for the adjustments. I hope you have enjoyed these vacations!

I left a few comments more, check them.

web/src/components/users/utils.js
joseivanlopez: If I have to choose, then I would vote for the second one. The difference in performance is insignificant and the code is much easier to read, understand and maintain. Moreover, this method will be executed only a couple of times. IMHO nothing to worry here about performance.

balsa-asanovic:
Hey @dgdavid,

I extracted menu to an internal component.

Also, I updated the suggestions algorithm to what you suggested, it is much more readable and easier to understand, while the performance difference is barely present and as José said, not really important here.

> Question: is this only ready for mouse users? I haven't checked it by myself, but ideally we should add keyboard support.

No, it is not 😄 . Sort of forgot about that.
How should the keyboard commands be handled here? When the user is focused on username input field, should the arrow down key press start going through the options of the dropdown list? Or maybe use Tab? Or something else?

> But honestly I'm not sure what is the best choice right now. Maybe screen readers user does not need the hint neither. Or maybe is useful for sight people too.
> 
> If you have an opinion / knowledge about the topic, please speak up 😉
> 
Sorry, not much knowledge about the accessibility topic 😅 
That said, I do like how it looks and it sort of clears up any dilemma about what is being suggested in the dropdown vs. only listing bare usernames. But again, no strong preference here.

web/src/components/users/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { suggestUsernames } from "./utils";
```
dgdavid, line 22: Ouch! `cspell` so picky here. Let's disable it for this file (I've to review the whole cspell configuration)

```suggestion
/* cspell:disable */

import { suggestUsernames } from "./utils";
```

dgdavid:
> How should the keyboard commands be handled here? When the user is focused on username input field, should the arrow down key press start going through the options of the dropdown list?

I think so. Plus `Enter` should select the suggestion, like the click does.

> Or maybe use Tab? 

No, tab should respect the normal follow of the form and just jump to the next field.

dgdavid (commented):
@balsa-asanovic,

I've had a short chat with @imobachgs about the state of this PR. We've conclude that it looks good to us as it is now :) :muscle: :tada: 

We think that the keyboard support can be done as a follow-up PR, following the same approach we are trying to apply in the team to avoid long PRs. You can do it if you want to keep working in this area, or just open an issue for it and pick another one to work on if you prefer something new instead.

So, just adding the changes file is enough for merging it into master. Also, you can update/reword a bit the PR description if you don't mind, since it is a bit outdated if I'm not wrong.

Thanks a lot!

dgdavid:
> Sorry, not much knowledge about the accessibility topic 😅 That said, I do like how it looks and it sort of clears up any dilemma about what is being suggested in the dropdown 

Ok, let's release it as it is and gather feedback from users.

balsa-asanovic:
Hay @dgdavid 

First, as always thank you a lot for continued support and mentorship. 👏 

> You can do it if you want to keep working in this area, or just open an issue for it and pick another one to work on if you prefer something new instead.

Well, since I started working on this, I guess I would like to finish that part as well.
Should I open this issue even in the case that I'll be working on it?

>So, just adding the changes file is enough for merging it into master. Also, you can update/reword a bit the PR description if you don't mind, since it is a bit outdated if I'm not wrong.

I've logged changes and also updated a bit the description of the PR

web/src/components/users/utils.test.js
dgdavid, line 22: Previously fixed, but a @balsa-asanovic `push --force` undid it

> Ouch! `cspell` so picky here. Let's disable it for this file (I've to review the whole cspell configuration)

```suggestion
/* cspell:disable */

import { suggestUsernames } from "./utils";
```

dgdavid (approved):
LGTM :+1: thanks a ton!

dgdavid:
> Hay @dgdavid
> 
> First, as always thank you a lot for continued support and mentorship. 👏

Anytime!

> 
> Should I open this issue even in the case that I'll be working on it?
> 

I don't think so, just reference this or #112 in the new PR description and/or the entry in the changes file.

> I've logged changes and also updated a bit the description of the PR

And now it's merged :)

PR #1028, opened 2024-02-06
Space policy UI

joseivanlopez:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,472 @@
...
+ * @type {object}
+ * @property {string} name
+ * @property {string} label
+ * @property {string} description
+ */
+
+/** @type {SpacePolicy[]} */
+const SPACE_POLICIES = [
+  {
+    name: "delete",
```
dgdavid, line 49: NP: I'd use `id` instead of `name`. What do you think?

joseivanlopez: Fine with me.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+};
+
+/**
+ * Indicates whether a device is a drive (disk, RAID).
+ * @function
+ *
+ * @param {StorageDevice} device
+ * @returns {boolean}
+ */
+const isDrive = (device) => Object.keys(device).includes("vendor");
```
dgdavid, line 85: NP: I think this should be something given by the backend. But, in any case, I'd move this helper funtion to storage-utils.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+ * @param {object} props
+ * @param {StorageDevice} props.device
+ */
+const DeviceDescriptionColumn = ({ device }) => {
+  return (
+    <>
+      <div>{device.name}</div>
+      <If
+        condition={isDrive(device)}
+        then={<div className="fs-small">{`${device.vendor} ${device.model}`}</div>}
```
dgdavid, line 100: Note for another iteration: use something different than a smaller size.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+        <If
+          condition={filesystem}
+          then={
+            <div className="fs-small">
+              {sprintf(_("%s file system"), filesystem?.type)}
+            </div>
+          }
+        />
+      </>
+    );
```
dgdavid, line 152: NP: I do not know exactly why, but I do not like this kind  of methods. I prefer a `Component` instead.

```suggestion
    const Content = () => {
      if (systems.length > 0) return systems.join(", ");
      if (device.filesystem?.isEFI) return _("EFI");

      return _("Not identified");
    };

    return (
      <>
        <div><Content />}</div>
        <If
          condition={filesystem}
          then={
            <div className="fs-small">
              {sprintf(_("%s file system"), filesystem?.type)}
            </div>
          }
        />
      </>
    );
```

But I'm aware that we're kind of _abusing_ of components here. Or maybe not.

Another option is to prefix the function with the `render` word. I.e., `renderContent` or `renderWhatever`. It makes the code more intuitive, IMHO.

Also wondering if this kind of "<DeviceContent />"  internal component could make sense in another place of the UI. E.g., in the DeviceSelector and also in another table. If so, it looks like a helper component to me, and consequently we should put it in a more central/common place to avoid repeating almost the same in multiple places.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+    const unused = device.size - used;
+
+    if (unused === 0) return null;
+
+    return (
+      <div className="fs-small">
+        {sprintf(_("%s unused"), deviceSize(unused))}
+      </div>
+    );
+  };
```
dgdavid, line 177: NP: Same here, just for feeding thoughts for another iteration: I wonder if that `UnusedSize` could make sense outside of this section, as a "property" of the disk or so.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+  const RecoverableSize = () => {
+    const size = device.recoverableSize;
+    let text;
+
+    if (size === 0)
+      text = _("No recoverable space");
+    else
+      text = sprintf(_("%s recoverable"), deviceSize(device.recoverableSize));
+
+    return <div className="fs-small">{text}</div>;
```
dgdavid, line 188: Note for another iteration: try something more than simply smaller text. Also check how these "less important" information should be _structured_ in terms of a11y.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+    <FormSelect
+      value={action}
+      isDisabled={isDisabled}
+      onChange={changeAction}
+      aria-label="Space action selector"
+    >
+      <FormSelectOption value="force_delete" label={_("Delete")} />
+      <FormSelectOption value="resize" label={_("Allow resize")} />
+      <FormSelectOption value="keep" label={_("Do not modify")} />
+    </FormSelect>
```
dgdavid, line 222: Note for another iteration, and related with a discussion I had with @ancorgs yesterday: stop using form controls for actions that are going to be executed/applied right away on user interaction. This is a bit more difficult than exchanging a checkbox by a switch, but for sure we will find something.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+            isHidden={!isExpanded}
+            onCollapse={onCollapse}
+            onChange={onChange}
+          />
+        );
+      });
+    });
+
+    return rows;
+  };
```
dgdavid, line 367: Note for another iteration: as we were talking, most probably this deserves a core/TreeTable wrapper or so. Specially given the fact we are adding an extra feature to the component provided by the library (the state saved into localStorage).

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,457 @@
...
+    <Section title={_("Find Space")} className="flex-stack">
+
+      <If
+        condition={isLoading && settings.spacePolicy === undefined}
+        then={<SectionSkeleton numRows={4} />}
+        else={
+          <>
+            <p>
+              {_("Indicate how to make free space in the selected disks for allocating the file systems:")}
+            </p>
```
dgdavid, line 446: Note for myself: I start feeling that core/Section should re-introduce the `description` or `helpText` prop for this kind of _inline help_. We will see.

dgdavid:
@joseivanlopez,

I've added a few test for the new storage/ProposalSpacePolicySection UI component.

Please, let me know if you miss some more.

Thanks!

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,472 @@
...
+
+/** @type {SpacePolicy[]} */
+const SPACE_POLICIES = [
+  {
+    name: "delete",
+    label: N_("Delete current content"),
+    description: N_("All partitions will be removed and any data in the disks will be lost.")
+  },
+  {
+    name: "resize",
```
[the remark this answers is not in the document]
joseivanlopez: We are using the same as the backend, to avoid translating the value.

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: The backend already gives that. I mean, all the drives implement the `Drive` D-Bus interface. Maybe we could add the `isDrive` attribute to the `StorageDevice` objects created by the storage js client. I would rethink this again when we start exporting the LVM volume groups.

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: Added `#unpartitionedSize` to the device object.

service/lib/agama/dbus/storage/device.rb
```diff
@@ -77,14 +79,16 @@ def storage_device=(value)
         attr_reader :tree
 
         # Adds the required interfaces according to the storage object
-        def add_interfaces # rubocop:disable Metrics/CyclomaticComplexity
+        def add_interfaces # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
jreidinger, line 82: well, this growing kind of call for a bit of refactoring. Something like having in interface class method like `apply?(storage_device)` and now have here something like
```
require "agama/dbus/storage/interfaces"

...
   interfaces = Interfaces.constants.select {|c| Interfaces.const_get(c).is_a?(Class)}
   interfaces.select! { |i| i.apply?(storage_device) }
```

This way it will automatically recognize new interfaces if they are added to that summary interfaces file. It also moves responsibility of knowing what is requirement of interface to that file and keep this file much simplier and without breaking metrics that already indicate that you are doing something wrong here.

jreidinger: ah, interfaces are components and not classes so simply replace `Class` with `Module`

service/lib/agama/dbus/storage/interfaces/component.rb
```diff
@@ -0,0 +1,89 @@
...
+          # Type of component.
+          #
+          # @return [String] Possible values:
+          #   "physical_volume"
+          #   "md_device"
+          #   "raid_device"
+          #   "multipath_wire"
+          #   "bcache_device"
+          #   "bcache_cset_device"
+          #   "md_btrfs_device"
```
jreidinger, line 45: possible value is also empty string

service/lib/agama/dbus/storage/interfaces/component.rb
```diff
@@ -0,0 +1,89 @@
...
+          def component_type
+            types = {
+              lvm_vg:      "physical_volume",
+              md:          "md_device",
+              dm_raid:     "raid_device",
+              multipath:   "multipath_wire",
+              bcache:      "bcache_device",
+              bcache_cset: "bcache_cset_device",
+              btrfs:       "md_btrfs_device"
```
jreidinger, line 54: do we want to have those string localized or not?

service/lib/agama/dbus/storage/interfaces/filesystem.rb
```diff
@@ -0,0 +1,61 @@
...
+          private_constant :FILESYSTEM_INTERFACE
+
+          # File system type.
+          #
+          # @return [String] e.g., "ext4"
+          def filesystem_type
+            storage_device.filesystem.type.to_s
+          end
+
+          # Whether the file system contains an EFI.
```
jreidinger, line 42: to be honest I find this explanation confusing  and prefer original one at https://github.com/yast/yast-storage-ng/blob/7157f8f78921ca688a41618f362ea2e04a0d4423/src/lib/y2storage/filesystems/base.rb#L225

EFI is Extensible Firmware Interface and ESP is EFI system partition, so later is what can file system contain.

web/cspell.json
```diff
@@ -75,6 +75,7 @@
         "textinput",
         "tkip",
         "udev",
+	      "unpartitioned",
```
jreidinger, line 78: indentation looks strange on github

service/test/agama/dbus/storage/interfaces/md_examples.rb
```diff
@@ -45,9 +45,13 @@
       end
     end
 
-    describe "#md_members" do
-      it "returns the name of the MD members" do
-        expect(subject.md_members).to contain_exactly("/dev/sda1", "/dev/sda2")
+    describe "#md_devices" do
+      it "returns the D-Bus path of the MD components" do
```
jreidinger, line 49: BTW all those changes to dbus API makes me worry that I do not see update of documentation, but looks like device interfaces are not documented at all.

web/src/client/storage.js
```diff
@@ -117,25 +118,33 @@ class DevicesManager {
    * @property {string} [transport]
    * @property {boolean} [sdCard]
    * @property {boolean} [dellBOOS]
-   * @property {string[]} [devices] - RAID devices (only for "raid" type)
+   * @property {string[]} [devices] - RAID devices (only for "raid" and "md" types)
```
jreidinger, line 121: I would not call it `RAID devices` and maybe instead `member devices` ?

web/src/client/storage.js
```diff
@@ -117,25 +118,33 @@ class DevicesManager {
...
    *
-   * @typedef {object} PartitionTableData
+   * @typedef {object} PartitionTable
    * @property {string} type
+   * @property {StorageDevice[]} partitions
+   * @property {number} unpartitionedSize - Total size not assigned to any partition
+   *
+   * @typedef {object} Filesystem
+   * @property {string} type
+   * @property {boolean} isEFI
```
jreidinger, line 142: after this change I have to read https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html to learn how it works and what means that square brackets :+1:

web/src/components/core/OptionsPicker.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+import React from "react";
+
+/**
+ * Wrapper for OptionsPicker options
+ * @component
+ *
+ * @param {object} props
+ * @param {string} [props.title] - Text to be used as option title
+ * @param {string} [props.body] - Text to be used as option body
```
jreidinger, line 30: why title and body is in square brackets? it looks like mandatory params to me and not optional.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+ * @component
+ *
+ * @param {object} props
+ * @param {StorageDevice} props.device
+ */
+const DeviceContentColumn = ({ device }) => {
+  const PartitionTableContent = () => {
+    return (
+      <div>
+        {sprintf(_("%s partition table"), device.partitionTable.type.toUpperCase())}
```
jreidinger, line 109: I think comment for translators would be nice here as it is hard to quess without code what that `%s` means.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+  };
+
+  const BlockContent = () => {
+    const renderContent = () => {
+      const systems = device.systems;
+      if (systems.length > 0) return systems.join(", ");
+
+      const filesystem = device.filesystem;
+      if (filesystem?.isEFI) return _("EFI");
+      if (filesystem) return sprintf(_("%s file system"), filesystem?.type);
```
jreidinger, line 121: I am not sure if that "file system" suffix won't be confusing. E.g. "Btrfs file system" will mean "B tree file system file system" or ntfs or even nfs will be strange.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+    );
+  };
+
+  const BlockContent = () => {
+    const renderContent = () => {
+      const systems = device.systems;
+      if (systems.length > 0) return systems.join(", ");
+
+      const filesystem = device.filesystem;
+      if (filesystem?.isEFI) return _("EFI");
```
jreidinger, line 120: I think more correct word here will be `ESP`

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+      if (filesystem) return sprintf(_("%s file system"), filesystem?.type);
+
+      const component = device.component;
+      switch (component?.type) {
+        case "physical_volume":
+          return sprintf(_("LVM physical volume of %s"), component.deviceNames[0]);
+        case "md_device":
+          return sprintf(_("Member of RAID %s"), component.deviceNames[0]);
```
jreidinger, line 128: again without translators comment it will be hard to guess. If I do not check code I will expect that there will be `0,1,5` as %s

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+ * @component
+ *
+ * @param {object} props
+ * @param {StorageDevice} props.device
+ */
+const DeviceSizeColumn = ({ device }) => {
+  return <div>{deviceSize(device.size)}</div>;
+};
+
+const DeviceDetailsColumn = ({ device }) => {
```
jreidinger, line 151: documentation of method?

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+  const UnusedSize = () => {
+    if (device.filesystem) return null;
+
+    const unused = device.partitionTable?.unpartitionedSize || 0;
+
+    return (
+      <div>
+        {sprintf(_("%s unused"), deviceSize(unused))}
```
jreidinger, line 159: I think this also deserve translators comment as without context it is hard to translate it properly.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+
+    return (
+      <div>
+        {sprintf(_("Shrinkable by %s"), deviceSize(device.recoverableSize))}
+      </div>
+    );
+  };
+
+  return (
+    <If condition={device.isDrive} then={<UnusedSize />} else={<RecoverableSize /> } />
```
jreidinger, line 177: inconsistent spacing in `{<RecoverableSize /> }`

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+ * @param {object} props
+ * @param {StorageDevice} props.device
+ * @param {string} props.action - Possible values: "force_delete", "resize" or "keep".
+ * @param {boolean} [props.isDisabled=false]
+ * @param {(action: SpaceAction) => void} [props.onChange]
+ */
+const DeviceActionColumn = ({ device, action, isDisabled = false, onChange = noop }) => {
+  const changeAction = (_, action) => onChange({ device: device.name, action });
+
+  const value = (device.isDrive && action === "resize") ? "keep" : action;
```
jreidinger, line 194: I think this code deserves comment why it is needed as from my first look it looks like we have some bug in underlaying code that it can return such action.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+      <Td dataLabel={columnNames.content}><DeviceContentColumn device={device} /></Td>
+      <Td dataLabel={columnNames.size}><DeviceSizeColumn device={device} /></Td>
+      <Td dataLabel={columnNames.details}><DeviceDetailsColumn device={device} /></Td>
+      <Td dataLabel={columnNames.action}>
+        <If
+          condition={showAction}
+          then={
+            <DeviceActionColumn
+              device={device}
+              action={spaceAction?.action || "keep"}
```
jreidinger, line 272: should we decide default action at this UI level in table? Maybe we should be strict and return action always from backend code?

web/src/components/core/OptionsPicker.jsx
dgdavid: No, they are not. At least one of them should be present, of course. 

So, maybe we can go for making title mandatory and body optional.

Thanks for catching it. Good to see you reviewing these pieces of code too.

web/src/components/storage/ProposalSpacePolicySection.jsx
dgdavid: Agree

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -0,0 +1,444 @@
...
+  return (
+    <OptionsPicker>
+      {SPACE_POLICIES.map((policy) => {
+        return (
+          <OptionsPicker.Option
+            key={policy.id}
+            title={policy.label}
+            body={policy.description}
+            onClick={() => onChange(policy.id)}
+            isSelected={currentPolicy?.id === policy.id}
```
jreidinger, line 393: I see `?.` here unnecessary

web/src/components/core/OptionsPicker.jsx
jreidinger: well, it is partly learning new stuff as I am not much familiar with jsdoc, so I need to read first how it works.

web/src/components/storage/ProposalSpacePolicySection.jsx
dgdavid: I prefer keeping it if there is any chance to have a proposal without a selected policy. Now or in the future.

service/lib/agama/dbus/storage/interfaces/component.rb
joseivanlopez: No, they are a type of id.

service/test/agama/dbus/storage/interfaces/md_examples.rb
joseivanlopez: Exactly, we don't have documentation yet for that interfaces. In fact, I don't know how to fit them in our system for documenting D-Bus, because those interfaces are not associated to a specific object but to an arbitrary object in the tree of devices. Maybe @mvidner could help.

web/src/client/storage.js
joseivanlopez: I have avoided `member` term because libstorage-ng uses it for a kind of BIOS RAID, see https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/md_member.rb#L35.

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: Well, the column is describing the content.

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: Good catch!

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: Well, if the backend does not specify an action then there is no action (it will do nothing with that device). Using "keep" for representing "no action" is something in the side of the UI code.

web/src/components/storage/ProposalSpacePolicySection.jsx
joseivanlopez: I think it is ok "Btrfs file system" or "ntfs file system". At the end, we use "Btrfs", "Xfs", etc as a name.

web/cspell.json
joseivanlopez: It is strange, I see it correctly in the code.

service/lib/agama/dbus/storage/device.rb
joseivanlopez: I like it! Thanks for your suggestion.

dgdavid (approved):
LGTM

joseivanlopez:
@jreidinger we are going to merge because almost all your suggestions were implemented. Anyway, don't hesitate to add more comments if you still miss something. Thanks!

PR #1029, opened 2024-02-07
[PoC] Import AutoYaST profiles

imobachgs:
Agama offers a mechanism to perform [unattended installations](../autoinstallation/). However, we would like AutoYaST users to be able to use their AutoYaST profiles in Agama.

This pull request contains:

* A document describing how Agama could support, to some extent, such profiles.
* A proof-of-concept that includes support for dynamic profiles. It gets activated when the profile name does not end in `xml`, `erb` or `/` (for rules).
* 
NOTE: Tests are failing because the `yast2-schema` package is missing.

## To do

- [x] Add a changelog entry.
- [x] Build a testing image.
- [x] Drop the D-Bus method.


service/lib/agama/autoyast/converter.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+
+      def tmp_profile_path
+        @tmp_profile_path ||= File.join(
+          Yast::AutoinstConfig.profile_dir,
+          "autoinst.xml"
+        )
+      end
+
+      # @return [Hash] D-Installer profile
```
teclator, line 112:
```suggestion
      # @return [Hash] Agama profile
```

doc/autoyast.md
```diff
@@ -0,0 +1,219 @@
...
+At this point, Agama only supports defining a list of connections that could correspond with the
+AutoYaST interfaces list. We might need to extend Agama to support `dns`, `net-udev`, etc.
+
+About `keep_install_network` and `setup_before_proposal`, we should not implement them to keep
+things simple.
+
+### `partitioning`
+
+By far, the most complex part of an AutoYaST profile. We can import the AutoYaST `partitioning`
+section as is because the partitioning is handled by the same code in Agama and AutoyaST.
```
teclator, line 122:
```suggestion
section as it is because the partitioning is handled by the same code in Agama and AutoyaST.
```

teclator (approved):
LGTM

PR #1033, opened 2024-02-09
[doc] Document the new Agama architecture

imobachgs:
Add a document explaining the proposal for the new Agama architecture. Let's consider the document as a starting point for further discussion.

You can see the [rendered version](https://github.com/openSUSE/agama/blob/new-architecture/doc/new_architecture.md) if you prefer.

PS: please, forgive me for those diagrams 😅

### DIFF
--- a/doc/new_architecture.md
+++ b/doc/new_architecture.md
@@ -152,17 +152,6 @@ user/password. For that reason, there must be a mechanism to skip the authentica
 Live could run a special service that generates a valid token and injects such a token into the
 server, the CLI and the web browser.
 
-## The (optimistic) plan
-
-1. Build an HTTP API to access the stuff in `agama-dbus-server`.
-2. Rename `agama-dbus-server` to `agama-core`.
-3. Move the manager to `agama-core`.
-4. Drop `agama-core` D-Bus interface, except the bits required by `rubygem-agama`.
-5. Build an HTTP API to access `rubygem-agama`.
-6. Adapt the CLI to use the HTTP API.
-7. Update `agama-live` to skip the authentication when not needed.
-8. Adapt the Web UI to use the HTTP API.
-
 ## Links
 
 * https://bugzilla.suse.com/show_bug.cgi?id=1219688
--- a/doc/new_architecture.md
+++ b/doc/new_architecture.md
@@ -1,0 +1,172 @@
+# Agama's 2024 architecture
+
+This document describes the proposal for the new Agama architecture. The reasons for introducing
+these changes are recorded in [a discussion in Agama's repository][drop-cockpit].
+
+[drop-cockpit]: https://github.com/openSUSE/agama/discussions/1000
+
+But before describing how the architecture should look, let's quickly look at the current status.
+
+## The current architecture
+
+At this point, Agama is composed of three high-level components:
+
+* **Agama service**: implements the logic to perform the installation. It is the core component of
+Agama and it offers a D-Bus interface. Actually, it is composed of two services: `rubygem-agama` and
+`agama-dbus-server`.
+
+* **Web user interface (`cockpit-agama`)**: a web-based interface that plays the role of a GUI when
+using Agama live.
+
+* **Command Line Interface (`agama-cli`)**: it allows to interact with Agama core and drives the
+auto-installation process.
+
+In addition to those components, we need to consider Cockpit, which plays a vital role:
+
+* It makes communication between the web UI and the D-Bus services possible.
+* It makes the web UI code available to the browser.
+* It takes care of authenticating the user when connecting remotely. Again, it is only relative to
+the web UI.
+
+```mermaid
+flowchart LR
+  subgraph Clients
+    Browser
+    CLI
+  end
+
+  subgraph Cockpit
+    UI["Web UI"]
+  end
+
+  subgraph Agama Service
+    Rust[Agama Rust]
+    Ruby[Agama Ruby]
+  end
+
+  Browser <---> UI
+  UI <--D-Bus--> Rust
+  UI <--D-Bus--> Ruby
+  CLI <--D-Bus-->  Rust
+  CLI <--D-Bus--> Ruby
+  Rust <--D-Bus---> Ruby
+```
+
+## The new architecture
+
+The proposed architecture is not that different from the current one, but it tries to meet these
+goals:
+
+* Drop our dependency on Cockpit.
+* Implement a higher-level API to be consumed by the clients, partially replacing the D-Bus.
+
+### Components
+
+With those goals in mind, we are considering the following components:
+
+* **Agama core (old `agama-dbus-server`)**: implements the installation logic. It relies heavily on
+the Agama YaST service.
+
+* **Agama YaST service (old `rubygem-agama`)**: it is written in Ruby and has direct access to YaST
+libraries. Complex parts, like storage and software handling, are implemented in this component.
+
+* **HTTP and WebSocket API**: implements the API the clients should use to communicate with Agama.
+
+* **Web user interface (old `cockpit-agama`)**: Agama's graphical user interface. The web server
+makes this React application available to the browsers.
+
+* **Command Line Interface (`agama-cli`)**: it allows interaction with Agama core and drives the
+auto-installation process. With the new architecture, connecting through the network might be
+possible without SSH.
+
+The following diagram could be better, but it represents the main components and their interactions.
+
+```mermaid
+flowchart LR
+  subgraph Clients
+    Browser
+    CLI
+  end
+
+  subgraph Agama Core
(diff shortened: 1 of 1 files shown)


doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+
+* **Agama service**: implements the logic to perform the installation. It is the core component of
+Agama and it offers a D-Bus interface. Actually, it is composed of two services: `rubygem-agama` and
+`agama-dbus-server`.
+
+* **Web user interface (`cockpit-agama`)**: a web-based interface that plays the role of a GUI when
+using Agama live.
+
+* **Command Line Interface (`agama-cli`)**: it allows to interact with Agama core and drives the
+auto-installation process.
```
jreidinger, line 22: I would add here to architecture also auto installation which is based on CLI. So we have complete picture.

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+  Rust <--D-Bus---> Ruby
+```
+
+## The new architecture
+
+The proposed architecture is not that different from the current one, but it tries to meet these
+goals:
+
+* Drop our dependency on Cockpit.
+* Implement a higher-level API to be consumed by the clients, partially replacing the D-Bus.
```
jreidinger, line 61: what means that `partially replacing`?

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+
+With those goals in mind, we are considering the following components:
+
+* **Agama core (old `agama-dbus-server`)**: implements the installation logic. It relies heavily on
+the Agama YaST service.
+
+* **Agama YaST service (old `rubygem-agama`)**: it is written in Ruby and has direct access to YaST
+libraries. Complex parts, like storage and software handling, are implemented in this component.
+
+* **HTTP and WebSocket API**: implements the API the clients should use to communicate with Agama.
```
jreidinger, line 73: I would mention here that it uses DBus as RPC to communicate with core and YaST service.

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+
+* **Web user interface (old `cockpit-agama`)**: Agama's graphical user interface. The web server
+makes this React application available to the browsers.
+
+* **Command Line Interface (`agama-cli`)**: it allows interaction with Agama core and drives the
```
jreidinger, line 78: it is confusing to use `interaction with Agama core` as you will interact also with YaST service, not?

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+
+* Drop our dependency on Cockpit.
+* Implement a higher-level API to be consumed by the clients, partially replacing the D-Bus.
+
+### Components
+
+With those goals in mind, we are considering the following components:
+
+* **Agama core (old `agama-dbus-server`)**: implements the installation logic. It relies heavily on
+the Agama YaST service.
```
jreidinger, line 68: now I am a but confused. Is this that rust part of dbus interface? also does it have to run as root or can it be dedicated different user?

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+    end
+
+    Web <--"Channel?" --> Rust
+  end
+  Rust <-- "D-Bus" --> YaST["Agama YaST"]
+
+  Browser --> UI
+  Browser --> API
+  Browser <---> WS
+  CLI --> API
```
jreidinger, line 106: why it is one side to API? it is a bit confusing to me.

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+flowchart LR
+  subgraph Clients
+    Browser
+    CLI
+  end
+
+  subgraph Agama Core
+    subgraph Web["Web Server"]
+      direction TB
+      UI["Web UI"]
```
jreidinger, line 94: it is kind of strange to have in one layer UI, API and WS. Also I am a bit confused why websocket and API is used by browser and not by UI? I kind of expect that browser communicate with UI and UI ( aka js code ) do that calls to API and WS.

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+the list of available products, request a new storage proposal or start the installation. About the
+approach, something REST-like is the most suitable. Switching to GraphQL or gRPC do not seem to be
+needed in our case (todo: write why).
+
+* **WebSocket**: Agama will use WebSockets to notify clients about any relevant event: progress,
+configuration changes, etc. The event's format is not defined yet, but a JSON event containing the
+`type` and the `payload/details` should be enough[^topics].
+
+[^topics]: Ideally, the clients should be able to subscribe to the topics they are interested in.
+    But that feature can wait.
```
jreidinger, line 124: that can be interesting to some kind of installation observer client that can report not just progress but also all changes done.

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+About the certificate, Agama will use a self-signed certificate unless the user injects its own
+certificate (through a kernel command line option).
+
+NOTE: under discussion.
+
+### Authentication
+
+The HTTP interface should allow authentication specifying a user and password that will be checked
+against PAM. It is not clear yet, but we might need to check whether the logged user has permissions
+(most probably through Policy Kit).
```
jreidinger, line 140: Do we need to allow non root user for installation? Does it make sense at all?

doc/new_architecture.md
```diff
@@ -0,0 +1,161 @@
...
+
+On successful authentication, the server generates a [JSON Web Token][jwt] that the client will
+include in the subsequent requests. The web client stores the token in an HTTP-only
+cookie[^http-only] and the CLI uses a file with restricted permissions.
```
jreidinger, line 144: Do we need authentication for local browser and CLI? I think it should be possible to skip it otherwise it will confuse users.

jreidinger: ah, it is below :+1:

doc/new_architecture.md
imobachgs: I use the term `partially` because we still use D-Bus for IPC. But it is true that, in this context, that's just a detail. I will improve it.

doc/new_architecture.md
imobachgs: OK, let's use "Agama" only. But I do not expect the CLI to directly talk to Agama Ruby at all (agama-core should act as proxy so we always use HTTP/WS).

doc/new_architecture.md
imobachgs: IMHO we can drop the D-Bus interface of Agama core and replace it with HTTP. I do not have an interest in maintaining both interfaces. And yes, we need it to run as root.

doc/new_architecture.md
imobachgs: Because it is always a request/response. In the WS case, once you establish the connection, you might end up getting messages in both directions.

doc/new_architecture.md
imobachgs: Well, it depends where you put the bar :-) Actually, the browser itself is handling the HTTP/WS stuff. You are serving the UI (a static HTML/JavaScript/CSS file set), the HTTP API and the WebSocket.

doc/new_architecture.md
imobachgs: Yes, we could get a lot of information from there.

doc/new_architecture.md
imobachgs: No, I do not think so. But I wondered whether the right question is: "do you have permissions?" instead of "are you root?". However, I only have a little experience with Polkit (which reminds me that the correct name is *Polkit*).

jreidinger: Well, as user of installation you can do almost anything to system. So using just root makes it easier from my POV. For me better question is if we want to allow e.g. openssh keys instead of root password and enforce it. So customer can disable root password and enforce usage of ssh keys.

imobachgs: In my opinion, that's a different (but related) topic. Anyway, we can keep the *most probably through Polkit* text and re-evaluate before implementing.

imobachgs: And, yes, absolutely we should allow the users to inject their own SSH public keys.

doc/new_architecture.md
```diff
@@ -21,6 +21,9 @@ using Agama live.
 * **Command Line Interface (`agama-cli`)**: it allows to interact with Agama core and drives the
 auto-installation process.
 
+* **Auto-installation (`autoinstallation`)**: it is composed by a Systemd service (`agama-auto`) and
+an script that relies in `agama-cli`.
```
jreidinger, line 25: I am not sure  if "relies in" is correct. I would use `relies on`

jreidinger (approved):

PR #1039, opened 2024-02-14
initial global btrfs option

jreidinger:
## Problem

Follow up of:

- [https://trello.com/c/IK8R3qCu/3525-8-concept-for-btrfs-related-features](https://trello.com/c/IK8R3qCu/3525-8-concept-for-btrfs-related-features "smartCard-inline")
- [https://trello.com/c/JnSKpQnu/554-3-storage-modify-basic-attributes-fstype-snapshots-etc-for-pre-defined-volumes](https://trello.com/c/JnSKpQnu/554-3-storage-modify-basic-attributes-fstype-snapshots-etc-for-pre-defined-volumes "smartCard-inline")

Right now Btrfs snapshots are shown in the UI as an special kind of filesystem type for the root volume in the table of file-systems.

But later we realized is something more relevant because it affects what you can do regarding booting (no separate `/boot` partition, for example), the volume sizes or the approach to btrfs subvolumes.

- trello https://trello.com/c/BnqdAWC4/3547-3-agama-snapshots-as-global-setting-in-the-ui

## Solution

Make btrfs with snapshots as global option that modify how root volume looks like.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

How it looks for product that require btrfs with snapshots

btrfs-required

Edit form for root when btrfs is not with snapshots
btrfs-multi

Edit form for root when btrfs is with snapshots
btrfs-only

Global option for btrfs with snapshots
btrfs_with_snapshot


jreidinger:
First screenshot. I plan to improve locations. But open questions are:

1. should btrfs stay with that "with snapshot" type? or global option kills it?
2. is wording good enough?

btrfs_with_snapshot

ancorgs:
1a. If you refer to the table, don't put too much work on it, we need to rework it anyway.
1b. If you refer to the form to edit the entry, then "btrfs with snapshots" should be the only option (ie. not even editable) if snapshots are chosen above and should not be an option otherwise.

2. IMHO, the second sentence in the screenshots is the useful part. I would get rid of the first sentence. So final result could be something like this (very preliminary version): "_Allows to boot to a previous version of the system after configuration changes or software upgrades_". But likely @mvidner or @shundhammer can help better. My point is that I will explain nothing about the root volume and that stuff. I would likely not mention the impact in size either, since the "auto-size" thingy should already make that obvious.

joseivanlopez:
We already have a short explanation in the filesystem selector: "Allows rolling back any change done to the system and restoring its previous state", see https://github.com/openSUSE/agama/blob/master/web/src/components/storage/VolumeForm.jsx#L212.

Maybe we could reuse it.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -448,6 +448,50 @@ const EncryptionSettingsForm = ({
...
+  const [isChecked, setIsChecked] = useState(isCheckedProp);
+
+  const switchState = (checked) => {
+    setIsChecked(checked);
+    onChange(checked);
+  };
+
+  if (isLoading) return <Skeleton width="25%" />;
+
+  const explanation = _("Btrfs snapshots for root partition enforce root volume to use btrfs with snapshots enabled. It is useful for returning to previous snapshot of system e.g. after software upgrade.");
```
shundhammer, line 477: Broken English, missing a lot of articles. This should be:

"Btrfs snapshots for **the** root partition enforce **the** root volume to use btrfs with snapshots enabled. ~~It~~ This is useful for returning to **a** previous snapshot of **the** system **,** e.g. after **a** software upgrade."

Maybe also add something about problems created by the admin to point out that it's not usually problems that our upgrades introduced:

"...or administrator actions that turned out to be problematic."

or

"...or manual configurations that turned out to be problematic."

ancorgs:
About the translation from settings to UI that was being discussed on IRC:

- Whether snapshots are enabled or not is dictated by the property `Snapshots` of the root volume.
- Whether the user can change the default value of that property or not is dictated by the property `SnapshotsConfigurable` of the outline corresponding to the root volume.

So:

If `Snapshots` is true and `SnapshotsConfigurable` is false, then the property `Filesystem` is surely set to "btrfs" (otherwise the product settings make no sense) and snapshots are mandatory. Not sure how to represent that in the UI, I guess the switcher makes no sense and a sentence should be presented instead.

If `Snapshots` is false and `SnapshotsConfigurable` is false, then it's impossible to configure snapshots. Not sure how to represent that in the UI, I guess the switcher makes no sense and we should likely show nothing regarding snapshots. This case is theoretical, but...

Last but not least, if `SnapshotsConfigurable` is true the switcher is there and allows to:
  - Select true:
    - Sets `Snapshots` to true for the root volume
    - Sets the fstype of the root volume to btrfs
    - Gives not option to change the fstype in the edit form of the root volume. Instead it shows a sentence like "Btrfs with snapshots" or "Using btrfs to provide snapshots" or something similar.
  - Select false:
    - Sets `Snapshots` to true for the root volume
    - In the edit form of the root volume, allows to use any of the values specified at the property `Filesystems` of the outline corresponding to the root volume.

As a cherry on top, during a recent conversation with @dgdavid we concluded the explanation could be extended a bit to "_Uses btrfs for the root file system allowing to boot to a previous version of the system after configuration changes or software upgrades_".

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -448,6 +448,65 @@ const EncryptionSettingsForm = ({
...
+    </div>
+  );
+
+  if (rootVolume.outline.snapshotsConfigurable) {
+    return selector;
+  } else if (rootVolume.fsType === "Btrfs" && rootVolume.snapshots) {
+    return enabledStatement;
+  } else { // strange situation, should not happen
+    return (<></>);
+  }
```
dgdavid, line 507: Since [React 18](https://github.com/facebook/react/blob/main/CHANGELOG.md#react-2), this is no longer needed.

Now React allows returning `undefined`, and this is the default return value from a JavaScript function when you returns nothing. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#description.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -448,6 +448,65 @@ const EncryptionSettingsForm = ({
...
+      <div>
+        {explanation}
+      </div>
+    </div>
+  );
+  const enabledStatement = (
+    <div>
+      {_("Btrfs snapshots required by product.")}
+    </div>
+  );
```
dgdavid, line 499: NP: I think it is better to use _internal components_ or to change the approach instead of this one. Don't get me wrong, it is ok in this case, but I believe that it is always _executing_ the code inside these parenthesis, no matter if finally you end up using _the component_ or not.

so, I would do something like

```js
const configurableSnapshots = rootVolume.outline.snapshotsConfigurable;
const forcedSnapshots = !configurableSnapshots && rootVolume.fsType === "Btrfs" && rootVolume.snapshots;

const SnapshotsToggle = () => (
    <>
      <Switch
        id="snapshots"
        label={_("Use Btrfs Snapshots")}
        isReversed
        isChecked={isChecked}
        onChange={switchState}
      />
      <div>
        {explanation}
      </div>
    </>
);

return (
  <div>
    <If condition={forcedSnapshots) then={_("Btrfs snapshots required by product.")} />
    <If condition={configurableSnapshots) then={<SnapshotsToggle />} />
  </div>
);
```

jreidinger:
Now with fixed rendering here is how edit volume looks like after changes:

btrfs with snapshots enavbled
btrfs-only

btrfs without snapshots
btrfs-multi

web/src/components/storage/VolumeForm.test.jsx
```diff
@@ -95,12 +95,12 @@ beforeEach(() => {
 });
 
 it("renders a control for displaying/selecting the file system type", async () => {
-  const { user } = plainRender(<VolumeForm {...props} />);
+  // use home which is not snapshotted
```
dgdavid, line 98: CSpell so picky here. Maybe we can go for 

```suggestion
  // use home which is not using snapshots
```

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -112,6 +112,65 @@ const EncryptionSettingsForm = ({
...
+ * @callback onChangeFn
+ * @param {object} settings
+ */
+const SnapshotsField = ({
+  settings,
+  onChange = noop
+}) => {
+  const rootVolume = (settings.volumes || []).find((i) => i.mountPath === "/");
+
+  const initialChecked = rootVolume !== undefined && rootVolume.fsType === "Btrfs" && rootVolume.snapshots;
```
dgdavid, line 132: Not for now, but I think we should go for something similar to https://github.com/openSUSE/agama/blob/76c930c06e1ea4bb24fa11c023cd9a94e09337fd/web/src/client/storage.js#L55-L58 for having the Fstypes in a "constant".

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -112,6 +112,65 @@ const EncryptionSettingsForm = ({
...
+  const switchState = (_, checked) => {
+    setIsChecked(checked);
+    onChange({ value: checked, settings });
+  };
+
+  const configurableSnapshots = rootVolume.outline.snapshotsConfigurable;
+  const forcedSnapshots = !configurableSnapshots && rootVolume.fsType === "Btrfs" && rootVolume.snapshots;
+
+  const SnapshotsToggle = () => {
+    const explanation = _("Uses btrfs for the root file system allowing to boot to a previous version of the system after configuration changes or software upgrades.");
```
dgdavid, line 149: Here I would change two things:

* btrfs -> Btrfs
* Use the `<abbr>` wrapping `Btrfs`. you can see examples at https://github.com/openSUSE/agama/blob/76c930c06e1ea4bb24fa11c023cd9a94e09337fd/web/src/components/storage/ProposalDeviceSection.jsx#L409-L418 and https://github.com/openSUSE/agama/blob/76c930c06e1ea4bb24fa11c023cd9a94e09337fd/web/src/components/storage/ProposalSettingsSection.jsx#L80-L89

dgdavid (commented):
In general changes looks good. Just a few minor comments.

Regard to screenshots in the PR description, as agreed in IRC we can go ahead without updating them to the latest changes introduced in #1045.

web/src/components/storage/ProposalSettingsSection.jsx
jreidinger: yeap, having in translatable string html tags is definitively something nice....changing name will happen in few minutes

web/src/components/storage/ProposalSettingsSection.jsx
ancorgs, line 132: I would feel safer if that comparison of `fsType` would be case insensitive.

dgdavid: > I would feel safer if that comparison of fsType would be case insensitive.

Then, please move it to a `isBtrfs` util method.

dgdavid: >  please move it to a isBtrfs util method

https://github.com/openSUSE/agama/blob/master/web/src/components/storage/utils.js looks as a good location for placing it

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -112,6 +112,65 @@ const EncryptionSettingsForm = ({
...
+    return <Skeleton width="25%" />;
+  }
+
+  const switchState = (_, checked) => {
+    setIsChecked(checked);
+    onChange({ active: checked, settings });
+  };
+
+  const configurableSnapshots = rootVolume.outline.snapshotsConfigurable;
+  const forcedSnapshots = !configurableSnapshots && rootVolume.fsType === "Btrfs" && rootVolume.snapshots;
```
ancorgs, line 146: Same here about "Btrfs". The capital "B" is just a side effect of how we represent the selector in the UI.

web/src/components/storage/utils.js
```diff
@@ -135,6 +135,20 @@ const deviceLabel = (device) => {
   return size ? `${name}, ${deviceSize(size)}` : name;
 };
 
+/**
+ * Checks if volume uses given fs. This method works same as in backend
+ * case insensitive.
+ *
+ * @param {import(~/clients/storage).Volume} volume
+ * @param {string} fs filesystem name to check
+ * @returns {boolean} true when volume uses given fs
```
dgdavid, line 144: NP: if I'm not wrong, we're using dash and _title notation_. I.e., 

```suggestion
 * @param {string} fs - Filesystem name to check.
 * @returns {boolean} true when volume uses given fs
```

web/src/components/storage/utils.test.js
```diff
@@ -86,3 +86,17 @@ describe("splitSize", () => {
...
+describe("hasFS", () => {
+  it("returns true if volume has given filesystem", () => {
+    expect(hasFS({ fsType: "Btrfs" }, "Btrfs")).toBe(true);
+  });
+
+  it("returns true if volume has given filesystem regarding different case", () => {
+    expect(hasFS({ fsType: "btrfs" }, "Btrfs")).toBe(true);
+  });
+
+  it("returns false  if volume has different filesystem", () => {
```
dgdavid, line 99: NP: I guess Eslint is gonna complain, but anyway

```suggestion
  it("returns false if volume has different filesystem", () => {
```

dgdavid (approved):
LGTM

ancorgs (approved):
LGTM

PR #1045, opened 2024-02-21
[web] Storage page: break settings into multiple sections

dgdavid:
Break the storage settings section to accomplish the organization defined in the [storage UI](https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md) document. Thus, proposed changes make the page to have _Device_, _Settings_, and _File systems_ sections instead of only _Settings_.

Also, it includes a last minute change about moving the LVM setting to the new _Device_ section, because it is more

> consistent with the future we want to go (in which LVM is part of the device selection)
> 
>  — @ancorgs at #yast IRC channel

|Before | After |
|-|-|
| Screen Shot 2024-02-22 at 13 12 15 | Screen Shot 2024-02-22 at 16 01 12 |

---

Related to https://trello.com/c/SK4uBRnw (internal link)


web/src/components/storage/ProposalDeviceSection.jsx
```diff
@@ -0,0 +1,209 @@
...
+  const DeviceContent = ({ device }) => {
+    return (
+      <Button variant="link" isInline onClick={openForm}>
+        {device ? deviceLabel(device) : _("No device selected yet")}
+      </Button>
+    );
+  };
+
+  if (isLoading) {
+    return <Skeleton screenreaderText={_("Loading selected device")} width="25%" />;
```
dgdavid, line 128: Note: I realized that we are rendering these loading skeletons without an accessible text. Not sure if the one chosen here is the way to go, but definitely we should think on a pattern for describing all of them.

web/src/components/storage/index.js
```diff
@@ -23,6 +23,7 @@ export { default as ProposalPage } from "./ProposalPage";
 export { default as ProposalPageMenu } from "./ProposalPageMenu";
 export { default as ProposalSettingsSection } from "./ProposalSettingsSection";
 export { default as ProposalSpacePolicySection } from "./ProposalSpacePolicySection";
+export { default as ProposalDeviceSection } from "./ProposalDeviceSection";
```
dgdavid, line 26: I start wondering about the convenience of exporting this _internal_ components. We are doing so just for importing them in the test file (!!), but we can do it without polluting the index file.

I.e., 

```js
import ProposalDeviceSection from "~/components/storage/ProposalDeviceSection";
```

instead of 

```js
import { ProposalDeviceSection } from "~/components/storage";
```

joseivanlopez: I agree. For example, I would say that *components/storage/index.js* should export the pages and the components that might be reused from pages out of storage. In that way the index file helps to know what is "public" in the storage namespace.

dgdavid: > I agree. For example, I would say that _components/storage/index.js_ should export the pages and the components that might be reused from pages out of storage. In that way the index file helps to know what is "public" in the storage namespace.

And this is extensible to all namespaces.

web/src/components/storage/ProposalDeviceSection.jsx
dgdavid: What does not fit to me here is the "Loading" word. It's actually not "loading" such a devices, but _gathering the_ or _waiting for_ information to know which device is selected.

web/src/components/storage/ProposalDeviceSection.jsx
```diff
@@ -0,0 +1,208 @@
...
+            isDisabled={devices.length === 0}
+          >
+            {_("Accept")}
+          </Popup.Confirm>
+          <Popup.Cancel onClick={closeForm} />
+        </Popup.Actions>
+      </Popup>
+    </>
+  );
+};
```
dgdavid, line 168: Note: most probably, these internal components does not make sense now that the device selector it's in its own section. I.e., most of the code could be _merged_ into the main component. But it is out of the scope of this PR, especially taking into account that there are more planned changes ahead that could affect this component too. So, let's do it in a _re-factorization round_.

web/src/components/storage/ProposalFileSystemsSection.jsx
```diff
@@ -0,0 +1,80 @@
...
+    return volumeTemplates.filter(t => (
+      t.mountPath.length > 0 && !mountPaths.includes(t.mountPath)
+    ));
+  };
+
+  const encryption = settings.encryptionPassword !== undefined && settings.encryptionPassword.length > 0;
+
+  return (
+    <Section title={_("File systems")}>
+      <ProposalVolumes
```
dgdavid, line 71: This PR does not directly integrate the `<ProposalVolumes>` code here because there is a plan for reworking almost the whole thing in a separated PR.

web/src/components/core/Section.jsx
```diff
@@ -86,7 +88,11 @@ export default function Section({
     const headerText = !path?.trim() ? title : <Link to={path}>{title}</Link>;
 
     return (
-      <h2 id={headerId}>{headerIcon}<span>{headerText}</span></h2>
+      <header>
+        <h2 id={headerId}>{headerIcon}<span>{headerText}</span></h2>
+        <p>{description}</p>
```
dgdavid, line 93: By now, I'm not adding an `aria-describedby` to the section pointing to this. I'm not sure if

* It's actually needed, since the description will be visible and I guess it will be read aloud
* It will be read twice
* It's needed to add an aria-hidden here too.

So, let's postpone this small detail just a bit, until we have time to test these a11y things.

web/src/components/storage/ProposalDeviceSection.jsx
ancorgs: I think that "Waiting for information about" is a good pattern for those texts.

dgdavid: Thanks a lot!

ancorgs:
About the section descriptions, I would only add these:

- Device: _Select the disk or LVM Volume Group for installation_. LVM is still not at this section, but we could already introduce the sentence with or without the LVM part. I'm not sure about adding something about the device being "where file systems will be allocated by default". It may help to understand the rest of the UI... but it may be a bit to technical as the first sentence. I'm also not sure about explaining LVM here, at most I would do it in the corresponding pop-up.
- Find Space: _Allocating the file systems might need to find free space in the devices listed below. Choose how to do it_
- Planned Actions: _Actions to create the file systems and to ensure the new system boots_

dgdavid:
> * . It may help to understand the rest of the UI... but it may be a bit to technical as the first sentence.

I agree. Too technical.

> I would not explain what LVM is here, at most I would do it in the corresponding pop-up.

But I think we should use the `<abbr>` right here. Long explanation in the dialog, of course, but _unwrapping_ the acronym from the very beginning. Don't make me think :)

web/src/components/storage/ProposalDeviceSection.jsx
```diff
@@ -0,0 +1,440 @@
...
+ * @typedef {import ("~/client/storage").DevicesManager.StorageDevice} StorageDevice
+ */
+
+/**
+ * Form for selecting the installation device.
+ * @component
+ *
+ * @param {object} props
+ * @param {string} props.id - Form ID.
+ * @param {StorageDevice|undefined} [props.current] - Currently selected device, if any.
```
joseivanlopez, line 50: NP: is `|undefined` needed? The param is optional.

web/src/components/storage/ProposalSpacePolicySection.jsx
```diff
@@ -447,7 +447,12 @@ export default function ProposalSpacePolicySection({
   const currentPolicy = SPACE_POLICIES.find(p => p.id === settings.spacePolicy) || SPACE_POLICIES[0];
 
   return (
-    <Section title={_("Find Space")}>
+    <Section
+      // TRANSLATORS: The storage "Find Space" section's title
+      title={_("Find Space")}
+      // TRANSLATORS: The storage "Find space" sections's description
+      description={_("Allocating the file systems might need to find free space in the devices listed below. Choose how to do it")}
```
joseivanlopez, line 454: NP: Do we have any rule for limiting the line length?

joseivanlopez, line 454: NP: having the setence without dot at the end looks strange to me.

joseivanlopez (approved):

web/src/components/storage/ProposalSpacePolicySection.jsx
dgdavid: Something that I myself asked in the past. We should have both, 

* a maximum line size
* a way of splitting them (I mean, the same way across the code)

dgdavid: I just followed the [@ancorgs's comment](https://github.com/openSUSE/agama/pull/1045#issuecomment-1959066761), which does not includes the dot at the end of these sentences. Should we add them?

ancorgs: The comment was not intended to be taken so literally. I have nothing against dots. Some of my best friends are dots.

joseivanlopez: We are splitting with back-slash in other parts of the code.

dgdavid: >  We are splitting with back-slash in other parts of the code.

I wanted to change it because it forces to break indentation... but agama-i18n eslint plugin [complains](https://github.com/openSUSE/eslint-plugin-agama-i18n/blob/262dd24a52cf5c140e7263f62e32e007172b06bd/string-literals.js#L11) if  `+` is used.

@lslezak told me that 

>  unfortunately xgettext does not support _("foo" + "bar") properly, the created POT file only contains the "foo" part but not the "bar" :-/

So, I've used back-slash too.

joseivanlopez (approved):

PR #1046, opened 2024-02-21
Add clang-devel and pam-devel in the setup services script

imobachgs:
Add missing `clang-devel` and `pam-devel` dependencies to the setup script.


mvidner (approved):
Yup, it works

I was a bit surprised that clang was needed but apparently it is a deep dependency for PAM

PR #1048, opened 2024-02-21
[web] Drop intrusive warning shown in the storage proposal page

dgdavid:
Since the Agama UI's regular workflow is to keep the system unchanged until installation begins, warning users about it is actually not needed. Thus, we agreed to use these intrusive warnings only in cases when/where Agama will take immediate action.

| Before | After |
|-|-|
|Screen Shot 2024-02-21 at 17 39 07 | Screen Shot 2024-02-21 at 17 38 48 |


ancorgs (approved):
LGTM

PR #1052, opened 2024-02-23
A little fix in the "About Agama" text

kobliha:
- Added missing "is"
- Using better phrase "we recommend using"

### DIFF
--- a/web/src/components/core/About.jsx
+++ b/web/src/components/core/About.jsx
@@ -51,9 +51,9 @@ export default function About() {
           {
             // TRANSLATORS: content of the "About" popup (1/2)
             _("Agama is an experimental installer for (open)SUSE systems. It \
-still under development so, please, do not use it in \
+is still under development so, please, do not use it in \
 production environments. If you want to give it a try, we \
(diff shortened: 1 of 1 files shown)


web/src/components/core/About.jsx
```diff
@@ -51,9 +51,9 @@ export default function About() {
           {
             // TRANSLATORS: content of the "About" popup (1/2)
             _("Agama is an experimental installer for (open)SUSE systems. It \
-still under development so, please, do not use it in \
+is still under development so, please, do not use it in \
 production environments. If you want to give it a try, we \
-recommend to use a virtual machine to prevent any possible \
+recommend using a virtual machine to prevent from any possible \
```
mvidner, line 56:
```suggestion
recommend using a virtual machine to prevent any possible \
```
"from" is only used before an "-ing" verb

mvidner (changes requested):
Thanks, but... :slightly_smiling_face:

web/src/components/core/About.jsx
kobliha: ok

mvidner (approved):

PR #1061, opened 2024-02-29
Initial break cockpt

jreidinger:
## Problem

According to new architecture for 2024 there is decision to use http based API instead of using cockpit and public dbus API.

## Solution

Start new architecture 2024 branch and use this as initial change that remove some specific cockpit stuff. Goal is not to have it in working state, but `npm build run` and `npm build server` still has to work and serve some content so we have starting point for more changes.

### DIFF
--- a/rust/agama-lib/src/software/store.rs
+++ b/rust/agama-lib/src/software/store.rs
@@ -18,10 +18,6 @@ impl<'a> SoftwareStore<'a> {
 
     pub async fn load(&self) -> Result<SoftwareSettings, ServiceError> {
         let patterns = self.software_client.user_selected_patterns().await?;
-        a = Some(Ok(1));
-if let Some(b) = a && let Ok(c) = b {
-    eprint!(c);
-}
         Ok(SoftwareSettings { patterns })
     }
 


web/webpack.config.js
```diff
@@ -8,13 +8,11 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const HtmlMinimizerPlugin = require("html-minimizer-webpack-plugin");
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
-const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
 const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");
```
imobachgs, line 11: Perhaps we can get rid of this one too.

jreidinger: yeap, on my list to check what it exactly do and why we need it.

jreidinger: I also checked if we need that manifests.js and manifest.json , but I worry it enlist list of provided localization. So probably I will skip it for now...but rsync looks ready for review as it is mainly for cockpit plugin development https://github.com/cockpit-project/cockpit/commit/e224a0296cd25a448d70810a4c114bac574b82dd

imobachgs: Yes, we still need the `manifest.json` file. In the future we could have a customized version, but we need at least the languages list.

rust/agama-lib/src/software/store.rs
```diff
@@ -18,6 +18,10 @@ impl<'a> SoftwareStore<'a> {
 
     pub async fn load(&self) -> Result<SoftwareSettings, ServiceError> {
         let patterns = self.software_client.user_selected_patterns().await?;
+        a = Some(Ok(1));
```
imobachgs, line 21: A left-over?

imobachgs (approved):

PR #1062, opened 2024-02-29
HTTPS support for the Agama web server

lslezak:
## Problem

- The communication between the user and the Agama web server needs to be encrypted

## Solution

- Enable HTTPS support using the standard openSSL library
- Automatically redirect all external HTTP requests to HTTPS
- Allow using HTTP locally (`http://localhost`)

## Testing

- Tested manually

## TODO

- [ ] Update the documentation

For later:

- [ ] Use the [gethostname crate](https://crates.io/crates/gethostname) and include the current host name in the generated self-signed certificate
- [ ] Read the SSL certificate from some default location (something like `/etc/agama.d/ssl/{certificate,key}.pem` ?)

### DIFF
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -142,12 +142,12 @@ fn https_redirect() -> Router {
         let request_host = req.headers().get("host");
         match request_host {
             // missing host in the request header, we cannot build the redirection URL, return error 400
-            None => return Ok(redirect_error()),
+            None => Ok(redirect_error()),
             Some(host) => {
                 let host_string = host.to_str();
                 match host_string {
                     // invalid host value in the request
-                    Err(_) => return Ok(redirect_error()),
+                    Err(_) => Ok(redirect_error()),
                     Ok(host_str) => return Ok(redirect_https(host_str, req.uri())),
                 }
             }
@@ -238,8 +238,8 @@ async fn start_server(address: String, service: Router, ssl_acceptor: SslAccepto
 
 /// Start serving the API.
 async fn serve_command(
-    address: &String,
-    address2: &String,
+    address: &str,
+    address2: &str,
     cert: &String,
     key: &String,
 ) -> anyhow::Result<()> {
@@ -256,14 +256,14 @@ async fn serve_command(
     let mut servers = vec![];
     if !address.is_empty() {
         servers.push(tokio::spawn(start_server(
-            address.clone(),
+            address.to_owned(),
             service.clone(),
             ssl_acceptor.clone(),
         )));
     }
     if !address2.is_empty() {
         servers.push(tokio::spawn(start_server(
-            address2.clone(),
+            address2.to_owned(),
             service.clone(),
             ssl_acceptor.clone(),
         )));
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -1,5 +1,3 @@
-use std::{path::PathBuf, pin::Pin};
-// use agama_lib::connection;
 use agama_dbus_server::{
     l10n::helpers,
     web::{self, run_monitor},
@@ -14,6 +12,7 @@ use hyper::body::Incoming;
 use hyper_util::rt::{TokioExecutor, TokioIo};
 use openssl::ssl::{Ssl, SslAcceptor, SslFiletype, SslMethod};
 use std::process::{ExitCode, Termination};
+use std::{path::PathBuf, pin::Pin};
 use tokio::sync::broadcast::channel;
 use tokio_openssl::SslStream;
 use tower::Service;
@@ -24,14 +23,14 @@ use utoipa::OpenApi;
 enum Commands {
     /// Start the API server.
     Serve {
-        // Address to listen to (":::3000" listens for both IPv6 and IPv4
+        // Address/port to listen on (":::3000" listens for both IPv6 and IPv4
         // connections unless manually disabled in /proc/sys/net/ipv6/bindv6only)
         #[arg(long, default_value = ":::3000", help = "Primary address to listen on")]
         address: String,
         #[arg(
             long,
             default_value = "",
-            help = "Optional secondary address to listen to"
+            help = "Optional secondary address to listen on"
         )]
         address2: String,
         #[arg(
@@ -155,11 +154,8 @@ fn https_redirect() -> Router {
     });
 
     Router::new()
-        .route_service(
-            // the wildcard path below does not match an empty path, we need to match it explicitly
-            "/",
-            redirect_service,
-        )
+        // the wildcard path below does not match an empty path, we need to match it explicitly
+        .route_service("/", redirect_service)
         .route_service("/*path", redirect_service)
 }
 
@@ -254,13 +250,13 @@ async fn serve_command(
     let ssl_acceptor = create_ssl_acceptor(cert, key);
 
     let mut servers = vec![];
-    if !address.is_empty() {
-        servers.push(tokio::spawn(start_server(
-            address.to_owned(),
-            service.clone(),
-            ssl_acceptor.clone(),
-        )));
-    }
+    servers.push(tokio::spawn(start_server(
+        address.to_owned(),
+        service.clone(),
+        ssl_acceptor.clone(),
+    )));
+
+    // optionally listen on the secondary address/port
     if !address2.is_empty() {
         servers.push(tokio::spawn(start_server(
             address2.to_owned(),
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -1,11 +1,22 @@
-use std::process::{ExitCode, Termination};
-
+use std::{path::PathBuf, pin::Pin};
+// use agama_lib::connection;
 use agama_dbus_server::{
     l10n::helpers,
     web::{self, run_monitor},
 };
+use axum::{
+    http::{Request, Response},
+    Router,
+};
 use clap::{Parser, Subcommand};
+use futures_util::pin_mut;
+use hyper::body::Incoming;
+use hyper_util::rt::{TokioExecutor, TokioIo};
+use openssl::ssl::{Ssl, SslAcceptor, SslFiletype, SslMethod};
+use std::process::{ExitCode, Termination};
 use tokio::sync::broadcast::channel;
+use tokio_openssl::SslStream;
+use tower::Service;
 use tracing_subscriber::prelude::*;
 use utoipa::OpenApi;
 
@@ -13,10 +24,28 @@ use utoipa::OpenApi;
 enum Commands {
     /// Start the API server.
     Serve {
-        // Address to listen on (":::3000" listens for both IPv6 and IPv4
+        // Address to listen to (":::3000" listens for both IPv6 and IPv4
         // connections unless manually disabled in /proc/sys/net/ipv6/bindv6only)
-        #[arg(long, default_value = ":::3000")]
+        #[arg(long, default_value = ":::3000", help = "Primary address to listen on")]
         address: String,
+        #[arg(
+            long,
+            default_value = "",
+            help = "Optional secondary address to listen to"
+        )]
+        address2: String,
+        #[arg(
+            long,
+            default_value = "",
+            help = "Path to the SSL private key file in PEM format"
+        )]
+        key: String,
+        #[arg(
+            long,
+            default_value = "",
+            help = "Path to the SSL certificate file in PEM format"
+        )]
+        cert: String,
     },
     /// Display the API documentation in OpenAPI format.
     Openapi,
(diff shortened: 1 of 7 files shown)


rust/agama-server/tests/common/mod.rs
```diff
@@ -41,6 +41,12 @@ pub trait ServerState {}
 impl ServerState for Started {}
 impl ServerState for Stopped {}
 
+impl Default for DBusServer<Stopped> {
+    fn default() -> Self {
```
lslezak, line 45: This was automatically added by the clippy check, no idea why... :thinking:

rust/package/agama.changes
```diff
@@ -1,3 +1,16 @@
+-------------------------------------------------------------------
+Thu Feb 29 09:49:18 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
+
+- Web server:
+  - Accept also IPv6 connections
```
lslezak, line 5: This was done in #1057

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+            .unwrap();
+
+        tracing::info!("Loading PEM key: {}", key_file);
+        tls_builder
+            .set_private_key_file(PathBuf::from(key_file), SslFiletype::PEM)
+            .unwrap();
+    }
+
+    // check that the key belongs to the certificate
+    tls_builder.check_private_key().unwrap();
```
jreidinger, line 109: in general all those unwrap means hard failure. I think for final code we should at least some of them propagate and e.g. allow ot have http only when building it failed.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+}
+
+// build a router for the HTTP -> HTTPS redirection
+// if the redirection URL cannot be built from the original request it returns error 400
+// instead of the redirection
+fn https_redirect() -> Router {
+    // see https://docs.rs/axum/latest/axum/routing/struct.Router.html#example
+    let redirect_service = tower::service_fn(|req: axum::extract::Request| async move {
+        let request_host = req.headers().get("host");
+        match request_host {
```
jreidinger, line 142: to avoid so much nesting you can use `if let syntax`

```
if let Some(host) = request_host {
  let host_string = host.to_str();
  if let Ok(host_str) = host_string {
     Ok(redirect_https(host_str, req.uri())
  } else {
    Ok(redirect_error())
      } else {
  Ok(redirect_error())

  }
```

rust/agama-server/src/cert.rs
```diff
@@ -0,0 +1,63 @@
...
+use openssl::x509::{X509NameBuilder, X509};
+
+use openssl::asn1::Asn1Time;
+use openssl::bn::{BigNum, MsbOption};
+use openssl::hash::MessageDigest;
+use openssl::x509::extension::{BasicConstraints, SubjectKeyIdentifier};
+
+// Generate a self-signed SSL certificate
+// see https://github.com/sfackler/rust-openssl/blob/master/openssl/examples/mk_certs.rs
+pub fn create_certificate() -> Result<(X509, PKey<Private>), ErrorStack> {
```
imobachgs, line 14: This code could be part of the `Certificate` struct I mentioned before.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+                    tracing::error!("Error during TSL handshake from {}: {}", addr, err);
+                }
+
+                let stream = TokioIo::new(tls_stream);
+                let hyper_service =
+                    hyper::service::service_fn(move |request: Request<Incoming>| {
+                        tower_service.clone().call(request)
+                    });
+
+                let ret = hyper_util::server::conn::auto::Builder::new(TokioExecutor::new())
```
imobachgs, line 199: This looks like too much qualification, to be honest 😅. Bear in mind that you can even rename the elements you import (see [The use declaration](https://doc.rust-lang.org/rust-by-example/mod/use.html)).

imobachgs (changes requested):
First of all, congrats for your first Rust contribution! Great work :-)

I have added a few comments, and I would like to have a closer look at the `start_server` as it is quite complex.

imobachgs:
I did some tests, and this change seems to break the WebSocket support. At least [websocat](https://github.com/vi/websocat) does not work (I tried `ws` and `wss`).

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+    let msg = "HTTP protocol is not allowed for external requests, please use HTTPS.\n";
+    builder.body(String::from(msg)).unwrap()
+}
+
+// build a router for the HTTP -> HTTPS redirection
+// if the redirection URL cannot be built from the original request it returns error 400
+// instead of the redirection
+fn https_redirect() -> Router {
+    // see https://docs.rs/axum/latest/axum/routing/struct.Router.html#example
+    let redirect_service = tower::service_fn(|req: axum::extract::Request| async move {
```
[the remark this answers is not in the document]
jreidinger: agreed. It is kind of confusing...for me it is also not easy to decide. Usually when code is short, it is easy to keep all import in my mind, but when file is longer I often forgot what belongs to what...and another thing is if two crates uses similar name, then we should really at least rename it....btw can you rename it e.g. to `axum::Request` or only `axum_request` is allowed?

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+
+// start the web server
+async fn start_server(address: String, service: Router, ssl_acceptor: SslAcceptor) {
+    tracing::info!("Starting Agama web server at {}", address);
+
+    // see https://github.com/tokio-rs/axum/blob/main/examples/low-level-openssl/src/main.rs
+    // how to use axum with openSSL
+    let listener = tokio::net::TcpListener::bind(&address)
+        .await
+        .unwrap_or_else(|_| panic!("could not listen on {}", &address));
```
jreidinger, line 170: will it also print details why it cannot listed like port occupied or insuficient permissions?

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+
+    let redirector = https_redirect();
+
+    loop {
+        let tower_service = service.clone();
+        let redirector_service = redirector.clone();
+        let tls_acceptor = ssl_acceptor.clone();
+
+        // Wait for a new tcp connection
+        let (tcp_stream, addr) = listener.accept().await.unwrap();
```
jreidinger, line 182: unwrap here also means immediate kill. Not sure if we can handle it more nicely.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+
+        tokio::spawn(async move {
+            if is_ssl_stream(&tcp_stream).await {
+                // handle HTTPS connection
```
jreidinger, line 186: to be honest, I would split this into own sub-methods to handle https and http connections. Method is already quite big and rust has no issue with private methods.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+                let ssl = Ssl::new(tls_acceptor.context()).unwrap();
+                let mut tls_stream = SslStream::new(ssl, tcp_stream).unwrap();
+                if let Err(err) = SslStream::accept(Pin::new(&mut tls_stream)).await {
+                    tracing::error!("Error during TSL handshake from {}: {}", addr, err);
```
jreidinger, line 190: so here it will just continue when handshake failed?

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
+                        }
+                    });
+
+                let ret = hyper_util::server::conn::auto::Builder::new(TokioExecutor::new())
+                    .serve_connection_with_upgrades(stream, hyper_service)
+                    .await;
+
+                if let Err(err) = ret {
+                    tracing::error!("Error serving connection from {}: {}", addr, err);
+                }
```
jreidinger, line 229: I see two identical ending of condition. Only difference is what stream and hyper_service contains.

rust/agama-server/src/cert.rs
```diff
@@ -0,0 +1,63 @@
+use openssl::error::ErrorStack;
+use openssl::pkey::{PKey, Private};
+use openssl::rsa::Rsa;
+use openssl::x509::extension::{KeyUsage, SubjectAlternativeName};
+use openssl::x509::{X509NameBuilder, X509};
+
+use openssl::asn1::Asn1Time;
+use openssl::bn::{BigNum, MsbOption};
+use openssl::hash::MessageDigest;
+use openssl::x509::extension::{BasicConstraints, SubjectKeyIdentifier};
```
jreidinger, line 10: here it use two `use` sharing same `openssl::x509::extension::` which is quite confusing.

rust/agama-server/src/cert.rs
```diff
@@ -0,0 +1,63 @@
...
+        serial.rand(159, MsbOption::MAYBE_ZERO, false)?;
+        serial.to_asn1_integer()?
+    };
+    builder.set_serial_number(&serial_number)?;
+    builder.set_subject_name(&x509_name)?;
+    builder.set_pubkey(&key)?;
+
+    let not_before = Asn1Time::days_from_now(0)?;
+    builder.set_not_before(&not_before)?;
+    let not_after = Asn1Time::days_from_now(365)?;
```
jreidinger, line 36: so basically we create certificate that is valid from today for a next year. Question is are we sure we have correct date in agama? if not, then it can be issue with remote browser, that will have different date.

rust/agama-server/src/cert.rs
```diff
@@ -0,0 +1,63 @@
...
+            .critical()
+            .key_cert_sign()
+            .crl_sign()
+            .build()?,
+    )?;
+
+    builder.append_extension(
+        SubjectAlternativeName::new()
+            .dns("agama")
+            .dns("agama.local")
```
jreidinger, line 51: I would document here why it is needed  as not everyone follows so closely agama developement.

rust/agama-server/tests/common/mod.rs
jreidinger: https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#new_without_default

rust/package/agama.changes
jreidinger: so please mention it in change with `gh#openSUSE/agama#1057` annotation.

rust/agama-server/src/agama-web-server.rs
mchf: it was also my thought .... may be you're looking (as i was) for things like ```?```, ```is_ok()```, ```ok_or```, ```and_then``` and others (depending whether you're dealing with Option or Result, Option also contains some funny methods for converting into a Result directly)

imobachgs:
> I did some tests, and this change seems to break the WebSocket support. At least [websocat](https://github.com/vi/websocat) does not work (I tried `ws` and `wss`).

Well, that's not true, I am having problems even with a regular `authenticate` call. So probably it is a problem on my side.

imobachgs:
> I did some tests, and this change seems to break the WebSocket support. At least [websocat](https://github.com/vi/websocat) does not work (I tried `ws` and `wss`).

That's not true, it works just fine 😅

rust/agama-server/src/agama-web-server.rs
```diff
@@ -32,23 +60,213 @@ struct Cli {
...
-
     let (tx, _) = channel(16);
     run_monitor(tx.clone()).await?;
 
     let config = web::ServiceConfig::load().unwrap();
     let service = web::service(config, tx);
-    axum::serve(listener, service)
-        .await
-        .expect("could not mount app on listener");
+    let ssl_acceptor = create_ssl_acceptor(cert, key);
```
imobachgs, line 250: Setting up the SSL Acceptor looks like a job for the `start_server` function. If you had a `Certificate` struct it would feel kind of natural to pass it as an argument to the `start_server` function.

rust/agama-server/src/agama-web-server.rs
imobachgs: It is not trivial: bear in mind that streams have different types. It is doable, sure, but we need to think a little about it.

rust/agama-server/src/cert.rs
```diff
@@ -0,0 +1,63 @@
+use openssl::error::ErrorStack;
+use openssl::pkey::{PKey, Private};
+use openssl::rsa::Rsa;
+use openssl::x509::extension::{KeyUsage, SubjectAlternativeName};
+use openssl::x509::{X509NameBuilder, X509};
+
```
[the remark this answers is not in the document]
lslezak: Generating a self-signed certificate manually [is not trivial](https://stackoverflow.com/questions/10175812/how-to-generate-a-self-signed-ssl-certificate-using-openssl).

I thought it would be a good idea to reuse the code for some separate tool like `agama-gen-cert`. You would call `agama-gen-cert mycert.pem mykey.pem` to obtain your certificate easily. (Probably with few optional configuration settings like `--hostname myagama.mydomain`.)

Then you would just copy them to a running Agama instance.

rust/agama-server/src/agama-web-server.rs
lslezak: If the server cannot listen on the port then we cannot do much about it. :worried:
But normally this should not happen, when running from Live ISO the port should not be used by anything else and the server should have enough permissions.

This could possibly happen only during development where a crash is acceptable.

rust/agama-server/src/agama-web-server.rs
lslezak: Yes, the problem is that it uses different data types in each branch. I have no idea how solve that, we probably need to use some templates...

rust/agama-server/src/cert.rs
lslezak: Ah, good point. On the other hand the certificate is self-signed so the browser will complain anyway. Wrong date will make the problem just a little bit more serious. :smile:

rust/agama-server/src/cert.rs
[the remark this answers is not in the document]
imobachgs: Sorry, I meant the blank-line 😅

rust/agama-server/src/agama-web-server.rs
imobachgs: You cannot do that much, but you can report an error. In this case it just crashes.

rust/agama-server/src/agama-web-server.rs
imobachgs: Yes, most probably generics is the way to go. However, splitting this stuff into smaller functions (one for SSL and the other for non-SSL) would make the thing more readable.

rust/agama-server/src/agama-web-server.rs
mchf: just a question? Why do you consider ```unwrap_or_else``` better then ```expect``` I put there in previous commit? I know that documentation states that "Because this function may panic, its use is generally discouraged." But as you use ```panic!``` macro anyway you more less mimic expect here.

rust/agama-server/src/cert.rs
lslezak: We could use something like
```rust
// Jan 1 1970 00:00 UTC
let not_before = Asn1Time::from_unix(0)?;
// Jan 1 2040 00:00 UTC 
let not_after = Asn1Time::from_str_x509("20400101000000Z")?;
```
This would cover bigger range.

Anyway, I guess this will be reviewed by the security team later so this might change...

lslezak: Uh, wrong date would even make troubles connecting to other remote https resources like SCC. As a solution we should probably run NTP synchronization on the Live ISO...

rust/agama-server/src/agama-web-server.rs
```diff
@@ -39,26 +40,270 @@ struct Cli {
...
 
+#[derive(Args, Debug)]
+struct ServeArgs {
+    // Address/port to listen on (":::3000" listens for both IPv6 and IPv4
+    // connections unless manually disabled in /proc/sys/net/ipv6/bindv6only)
+    #[arg(long, default_value = ":::3000", help = "Primary address to listen on")]
+    address: String,
+    #[arg(
+        long,
+        default_value = "",
```
imobachgs, line 51: Please, do not use "empty" values to mean "nothing". Prefer using an `Option<T>`.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -39,26 +40,270 @@ struct Cli {
...
+    fn ssl_acceptor(&self) -> Result<SslAcceptor, openssl::error::ErrorStack> {
+        let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
+
+        if self.cert.is_empty() && self.key.is_empty() {
+            let (cert, key) = agama_server::cert::create_certificate()?;
+
+            tls_builder.set_private_key(&key)?;
+            tls_builder.set_certificate(&cert)?;
+
+            // for debugging you might dump the certificate to a file:
```
imobachgs, line 83: I do not think this belongs here. It is kind of hidden.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -39,26 +40,270 @@ struct Cli {
...
+    #[arg(long, default_value = "unix:path=/run/agama/bus")]
+    dbus_address: String,
+}
+
+impl ServeArgs {
+    /// Builds an SSL acceptor using a provided SSL certificate or generates a self-signed one
+    fn ssl_acceptor(&self) -> Result<SslAcceptor, openssl::error::ErrorStack> {
+        let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
+
+        if self.cert.is_empty() && self.key.is_empty() {
```
imobachgs, line 77: Please, prefer `Option<T>`.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -39,26 +40,270 @@ struct Cli {
...
     run_monitor(tx.clone()).await?;
 
     let config = web::ServiceConfig::load()?;
     let dbus = connection_to(&args.dbus_address).await?;
     let service = web::service(config, tx, dbus).await?;
-    axum::serve(listener, service)
-        .await
-        .expect("could not mount app on listener");
+
+    let ssl_acceptor = if let Ok(ssl_acceptor) = args.ssl_acceptor() {
```
imobachgs, line 283: There is no need to change it now (time pressure), but it is weird that `ServeArgs` is responsible for creating the SSL acceptor.

imobachgs (commented):
Just a few comments. It is looking better now.

rust/agama-server/src/agama-web-server.rs
lslezak: If the server listens on two ports then it would be a good idea to use the same generated self-signed certificate on both ports. So the SSL acceptor needs to be created early.

imobachgs (approved):

PR #1067, opened 2024-03-04
Implementation of Server {Login|Logout} commands

mchf:
## Problem

We need authorization token to be able to proceed with subsequent commands

## Solution

```agama server login``` asks for JWT and stores it into a well-known file
```agama server logout``` deletes stored JWT

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -24,12 +24,8 @@ pub enum AuthCommands {
 /// Main entry point called from agama CLI main loop
 pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
     match subcommand {
-        AuthCommands::Login(options) => {
-            login(LoginArgs::proceed(options).password()?).await
-        }
-        AuthCommands::Logout => {
-            logout()
-        }
+        AuthCommands::Login(options) => login(LoginArgs::proceed(options).password()?).await,
+        AuthCommands::Logout => logout(),
     }
 }
 
@@ -193,7 +189,7 @@ async fn login(password: String) -> anyhow::Result<()> {
 fn logout() -> anyhow::Result<()> {
     // mask if the file with the JWT doesn't exist (most probably no login before logout)
     if !Path::new(DEFAULT_JWT_FILE).exists() {
-        return Ok(())
+        return Ok(());
     }
 
     Ok(std::fs::remove_file(DEFAULT_JWT_FILE)?)
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -41,7 +41,7 @@ pub fn jwt() -> anyhow::Result<String> {
         }
     }
 
-    Err(anyhow::anyhow!("JWT not available"))
+    Err(anyhow::anyhow!("Authentication token not available"))
 }
 
 /// Stores user provided configuration for login command
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -33,6 +33,7 @@ pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
     }
 }
 
+/// Reads stored JWT token and returns it
 pub fn jwt() -> anyhow::Result<String> {
     if let Some(file) = jwt_file() {
         if let Ok(token) = read_line_from_file(&file.as_path()) {
@@ -205,15 +206,14 @@ async fn login(password: String) -> anyhow::Result<()> {
     // 2) if successful store the JWT for later use
     if let Some(path) = jwt_file() {
         if let Some(dir) = path.parent() {
-            std::fs::create_dir_all(dir)?;
+            fs::create_dir_all(dir)?;
         } else {
             return Err(anyhow::anyhow!("Cannot store the JWT token"));
         }
 
-        std::fs::write(path.as_path(), res)?;
+        fs::write(path.as_path(), res)?;
         set_file_permissions(path.as_path())?;
     }
-
 
     Ok(())
 }
@@ -222,19 +222,25 @@ async fn login(password: String) -> anyhow::Result<()> {
 fn logout() -> anyhow::Result<()> {
     let path = jwt_file();
 
-    if !&path.clone().is_some_and(|p| p.as_path().exists()) {
+    if !&path.clone().is_some_and(|p| p.exists()) {
         // mask if the file with the JWT doesn't exist (most probably no login before logout)
         return Ok(());
     }
 
+    // panicking is right thing to do if expect fails, becase it was already checked twice that
+    // the path exists
     let file = path.expect("Cannot locate stored JWT");
 
-    return Ok(std::fs::remove_file(file)?)
+    Ok(fs::remove_file(file)?)
 }
 
 /// Shows stored JWT on stdout
 fn show() -> anyhow::Result<()> {
-    println!("{}", jwt()?);
+    // we do not care if jwt() fails or not. If there is something to print, show it otherwise
+    // stay silent
+    if let Ok(token) = jwt() {
+        println!("{}", token);
+    }
 
     Ok(())
 }
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -1,4 +1,5 @@
 use clap::{arg, Args, Subcommand};
+use home;
 use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
 use std::fs;
 use std::fs::File;
@@ -7,7 +8,7 @@ use std::io::{BufRead, BufReader};
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 
-const DEFAULT_JWT_FILE: &str = "/tmp/agama-jwt";
+const DEFAULT_JWT_FILE: &str = ".agama/agama-jwt";
 const DEFAULT_AUTH_URL: &str = "http://localhost:3000/authenticate";
 const DEFAULT_FILE_MODE: u32 = 0o600;
 
@@ -33,13 +34,13 @@ pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
 }
 
 pub fn jwt() -> anyhow::Result<String> {
-    let jwt = read_line_from_file(&Path::new(DEFAULT_JWT_FILE));
-
-    if let Ok(token) = jwt {
-        return Ok(token);
-    } else {
-        return Err(anyhow::anyhow!("JWT not available"));
+    if let Some(file) = jwt_file() {
+        if let Ok(token) = read_line_from_file(&file.as_path()) {
+            return Ok(token);
+        }
     }
+
+    return Err(anyhow::anyhow!("JWT not available"));
 }
 
 /// Stores user provided configuration for login command
@@ -106,6 +107,11 @@ impl Credentials for MissingCredentials {
     }
 }
 
+/// Path to file where JWT is stored
+fn jwt_file() -> Option<PathBuf> {
+    Some(home::home_dir()?.join(DEFAULT_JWT_FILE))
+}
+
 /// Reads first line from given file
 fn read_line_from_file(path: &Path) -> io::Result<String> {
     if !&path.exists() {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: rust/agama-cli/src/server.rs [hunk 1]
```diff
@@ -0,0 +1,195 @@
...
+    },
+    /// Release currently stored JWT
+    Logout,
+}
+
+/// Main entry point called from agama CLI main loop
+pub async fn run(subcommand: ServerCommands) -> anyhow::Result<()> {
+    match subcommand {
+        ServerCommands::Login { password, file } => {
+            // actions to do:
```
imobachgs, line 28: This comment might be useful as part of the command documentation. Here is not that useful.

rust/agama-cli/src/server.rs
```diff
@@ -0,0 +1,195 @@
...
+            // 1) somehow obtain credentials (interactive, commandline, from a file)
+            // credentials are handled in this way (in descending priority)
+            // "command line options" -> "read from file" -> "ask to the user"
+            // 2) pass credentials to the web server
+            // 3) receive JWT
+            // 4) store the JWT in a well known location
+
+            let options = LoginOptions {
```
imobachgs, line 36: Clap can do this for you:

```rust
#[derive(Debug, clap::Args)]
pub enum ServerCommands {
  Login(LoginOptions)
  Logout
}

struct LoginOptions {
  #[clap(long, short = 'p')]
  password: Option<String>,
  #[clap(long, short = 'f')]
  file: Option<String>
}
```

np: and if you follow that approach, you could rename it to `LoginArgs` :-)

rust/agama-cli/src/server.rs
```diff
@@ -0,0 +1,195 @@
...
+/// Stores user provided configuration for login command
+struct LoginOptions {
+    password: Option<String>,
+    file: Option<PathBuf>,
+}
+
+impl LoginOptions {
+    /// Transforms user provided options into internal representation
+    /// See Credentials trait
+    fn parse(options: LoginOptions) -> Box<dyn Credentials> {
```
imobachgs, line 60: np: usually `parse` refers to processing something like a string.

rust/agama-cli/src/server.rs
```diff
@@ -0,0 +1,195 @@
+use clap::Subcommand;
+use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
+use std::fs::File;
+use std::io;
+use std::io::{BufRead, BufReader};
+use std::path::PathBuf;
+
+const DEFAULT_JWT_FILE: &str = "/tmp/agama-jwt";
```
imobachgs, line 8: I would not save the token in `/tmp`. I would rather prefer to save it in some private place, like `$HOME/.agama/...`.

rust/agama-cli/src/server.rs
```diff
@@ -0,0 +1,195 @@
...
+    Err(anyhow::anyhow!("Failed to get JWT token"))
+}
+
+/// Logs into the installation web server and stores JWT for later use.
+async fn login(password: String) -> anyhow::Result<()> {
+    // 1) ask web server for JWT
+    let res = get_jwt(DEFAULT_AUTH_URL.to_string(), password).await?;
+
+    // 2) if successful store the JWT for later use
+    std::fs::write(DEFAULT_JWT_FILE, res)?;
```
imobachgs, line 187: Please, restrict file permissions (0400 would be enough).

imobachgs (changes requested):
I made a few comments (nothing critical), although I would probably only allow reading the password from `stdin` or, at most, from an environment variable. Of course, as a fallback, I would ask interactively (as you did). I guess it would be more flexible and simpler to implement :-)

rust/agama-cli/src/server.rs
```diff
@@ -0,0 +1,195 @@
...
+
+const DEFAULT_JWT_FILE: &str = "/tmp/agama-jwt";
+const DEFAULT_AUTH_URL: &str = "http://localhost:3000/authenticate";
+
+#[derive(Subcommand, Debug)]
+pub enum ServerCommands {
+    /// Login with defined server. Result is JWT stored and used in all subsequent commands
+    Login {
+        #[clap(long, short = 'p')]
+        password: Option<String>,
```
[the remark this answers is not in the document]
mchf: well ... ways how to obtain password was explicitly specified in the PBI, so I didn't improvise ;-)

rust/agama-cli/src/commands.rs
```diff
@@ -35,4 +36,7 @@ pub enum Commands {
     /// Collects logs
     #[command(subcommand)]
     Logs(LogsCommands),
+    /// Request an action on the web server like Login / Logout
+    #[command(subcommand)]
+    Server(ServerCommands),
```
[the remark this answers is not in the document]
mchf: I put it under ```Server``` with idea that one day we might be login to a remote server and then something like ```agama server --addr <ip> login ....``` looks reasonable, however if you want to separate server management (e.g. reload ...) and user commands (e.g. login) I can rename to something else easily

rust/agama-cli/src/server.rs
mchf: it was a kind of my todo list, but if it makes sense I can move it to the command doc.

rust/agama-cli/src/server.rs
mchf: in the end i managed to get it working ;-)

rust/agama-cli/src/server.rs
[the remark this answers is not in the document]
imobachgs: Yes, sure. It is just that, looking at the final result, I think stdin could be better. Anyway, it is something we can change later.

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
+    match subcommand {
+        AuthCommands::Login(options) => login(LoginArgs::proceed(options).password()?).await,
+        AuthCommands::Logout => logout(),
+        AuthCommands::Show => show(),
+    }
+}
+
+/// Reads stored JWT token and returns it
+pub fn jwt() -> anyhow::Result<String> {
```
imobachgs, line 37: No need to be public.

```suggestion
fn jwt() -> anyhow::Result<String> {
```

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+    if let Some(file) = jwt_file() {
+        if let Ok(token) = read_line_from_file(&file.as_path()) {
+            return Ok(token);
+        }
+    }
+
+    return Err(anyhow::anyhow!("JWT not available"));
```
imobachgs, line 44:
```suggestion
    Err(anyhow::anyhow!("JWT not available"))
```

Actually, if you run `cargo clippy`, you should get a bunch of suggestions.

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+    #[arg(long, short = 'p')]
+    password: Option<String>,
+    #[arg(long, short = 'f')]
+    file: Option<PathBuf>,
+}
+
+impl LoginArgs {
+    /// Transforms user provided options into internal representation
+    /// See Credentials trait
+    fn proceed(options: LoginArgs) -> Box<dyn Credentials> {
```
imobachgs, line 59:
```rust
    fn proceed(options: LoginArgs) -> Box<dyn Credentials> {
        if let Some(password) = options.password {
            Box::new(KnownCredentials { password })
        } else if let Some(path) = options.file {
            Box::new(FileCredentials { path })
        } else {
            Box::new(MissingCredentials {})
        }
    }
```

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/// Path to file where JWT is stored
+fn jwt_file() -> Option<PathBuf> {
+    Some(home::home_dir()?.join(DEFAULT_JWT_FILE))
+}
+
+/// Reads first line from given file
+fn read_line_from_file(path: &Path) -> io::Result<String> {
+    if !&path.exists() {
```
imobachgs, line 118: No need to use the `&`, it is already a reference.

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+        ));
+    }
+
+    if let Ok(file) = File::open(&path) {
+        // cares only of first line, take everything. No comments
+        // or something like that supported
+        let raw = BufReader::new(file).lines().next();
+
+        if let Some(line) = raw {
+            return Ok(line?);
```
imobachgs, line 131:
```suggestion
            return Ok(line);
```

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+async fn login(password: String) -> anyhow::Result<()> {
+    // 1) ask web server for JWT
+    let res = get_jwt(DEFAULT_AUTH_URL.to_string(), password).await?;
+
+    // 2) if successful store the JWT for later use
+    if let Some(path) = jwt_file() {
+        if let Some(dir) = path.parent() {
+            fs::create_dir_all(dir)?;
+        } else {
+            return Err(anyhow::anyhow!("Cannot store the JWT token"));
```
imobachgs, line 211: I would omit "JWT" (it is redundant, as "T" stands for "token"). Something more meaningful for the user could be "authentication token" or "authentication credentials".

imobachgs (changes requested):

rust/agama-cli/src/auth.rs
mchf: No, this won't work because ```BufReader::lines::next``` returns ```Option<Result<...>>``` si line is ```Result``` at this place.

So, it should either stay as is (with ```?```) or ```return line;``` At time of writing I picked the first option to make it kind of obvious that reading of line could fail.

rust/agama-cli/src/auth.rs
mchf: ah yes ... a direct consequence of redoing things multiple times as things evolved ;-)

rust/agama-cli/src/auth.rs
mchf: ah yes ... I was surprised that the match passed through the first round ;-)

I really don't know if the ```if ... else``` version is much better, but will adapt.

rust/agama-cli/src/auth.rs
mchf: ok ... messages are definitely not my strong skill

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
+use std::fs;
+use std::fs::File;
+use std::io;
+use std::io::{BufRead, BufReader};
+use std::os::unix::fs::PermissionsExt;
+use std::path::{Path, PathBuf};
+
+const DEFAULT_JWT_FILE: &str = ".agama/agama-jwt";
+const DEFAULT_AUTH_URL: &str = "http://localhost:3000/authenticate";
```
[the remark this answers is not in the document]
mchf: ok ... will update and rebase and retest.

rust/agama-cli/src/auth.rs
imobachgs: We can make it public when we need it to be public 😉. If we want to make it reusable, perhaps it should be in `agama-lib`.

rust/agama-cli/src/auth.rs
imobachgs: Yes, sorry, I made a mistake. I meant `return line`.

imobachgs: Actually, `cargo clippy` suggests this change too.

rust/agama-cli/src/auth.rs
```diff
@@ -0,0 +1,243 @@
...
+
+/// Reads stored token and returns it
+pub fn jwt() -> anyhow::Result<String> {
+    if let Some(file) = jwt_file() {
+        if let Ok(token) = read_line_from_file(&file.as_path()) {
+            return Ok(token);
+        }
+    }
+
+    Err(anyhow::anyhow!("JWT not available"))
```
imobachgs, line 44: I would change this to "Authentication token..." too.

imobachgs (approved):
It looks good overall (although I would like to have the text reviewed by @mvidner). But that's something we can do when we review the CLI's overall help, which is far from ideal.

As a note, I really think we should read stuff from stdin, but this PR implements what it was in the PBI. So I will create another one for the future.

Last but not least, please, do not "force-push" during a review. It makes it harder to review the changes (for instance, the [Compare](https://github.com/openSUSE/agama/compare/6ef471d1cc5128b4af39e7e6c134e01a88bdc63a..9042a193e9f1d3c4a0a687c710b309bfd627e274) link from GitHub might be useless then).

imobachgs (approved):

imobachgs (approved):

PR #1068, opened 2024-03-04
New storage settings for different target devices 

joseivanlopez:
## Problem

There are three main different scenarios to be supported by the storage proposal:

* a) Install in a target disk device (e.g., disk, RAID, DASD, etc).
* b) Install in a new LVM volume group.
* c) Reuse an existing LVM volume group.

Moreover, the proposal should be able to indicate the following boot options:

* a) Whether to configure or not separate partitions for booting.
* b) What device to use for creating the booting partitions.

## Solution

Adapt the storage settings to support all the scenarios described above:

~~~
Target                 s (options: "disk", "newLvmVg", "reusedLvmVg")
TargeDevice            s (optional: does not make sense if Target is "newLvmVg")
TargetPVDevices        as (optional: only makes sense if Target is "newLvmVg")
ConfigureBoot          b
BootDevice             s
~~~

## Testing

* Added and adapted unit tests
* Tested manually

### DIFF
--- a/service/lib/agama/storage/proposal.rb
+++ b/service/lib/agama/storage/proposal.rb
@@ -19,18 +19,32 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "y2storage"
 require "agama/issue"
 require "agama/storage/actions"
+require "agama/storage/device_settings"
 require "agama/storage/proposal_settings_conversion"
+require "yast"
+require "y2storage"
 
 module Agama
   module Storage
     # Backend class to calculate a storage proposal.
     class Proposal
+      include Yast::I18n
+
+      # Settings used for calculating the proposal.
+      #
+      # @note Some values are recoverd from Y2Storage, see
+      #   {ProposalSettingsConversion::FromY2Storage}
+      #
+      # @return [ProposalSettings, nil] nil if no proposal has been calculated yet.
+      attr_reader :settings
+
       # @param config [Config] Agama config
       # @param logger [Logger]
       def initialize(config, logger: nil)
+        textdomain "agama"
+
         @config = config
         @logger = logger || Logger.new($stdout)
         @on_calculate_callbacks = []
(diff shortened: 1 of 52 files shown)


web/src/client/storage.js
```diff
@@ -23,8 +23,8 @@
 // cspell:ignore ptable
 
 import DBusClient from "./dbus";
+import { compact, hex, uniq } from "~/utils";
```
dgdavid, line 26: NP: I'd move it even one more line up :P

web/src/components/storage/ProposalDeviceSection.jsx
```diff
@@ -397,12 +397,11 @@ export default function ProposalDeviceSection({
   isLoading = false,
   onChange = noop
 }) {
-  // FIXME: we should work with devices objects ASAP
-  const { bootDevice } = settings;
+  const targetDevice = settings.targetDevice;
```
dgdavid, line 400: NP: although the FIXME comment has been drop, I still thinking we should pass the targetDevice, not only its name, to the "BootdeviceSelectorDialog" or whatever component we're going to create.

joseivanlopez: Ok, let's do it as part of the other PR, https://github.com/openSUSE/agama/pull/1106.

service/lib/agama/storage/proposal.rb
```diff
@@ -186,13 +201,17 @@ def boot_device_issue
...
-        return unless settings.boot_device
-        return if available_devices.map(&:name).include?(settings.boot_device)
+        available = available_devices.map(&:name)
+        missing = settings.installation_devices.reject { |d| available.include?(d) }
 
-        Issue.new("Selected device is not found in the system",
+        return if missing.none?
+
+        Issue.new(
+          format(_("The following selected devices are not found in the system: %{devices}"),
```
ancorgs, line 210: Should we use `_n` here?

joseivanlopez: Probably yes because this issue is intended to be reported to the final user.

PR #1069, opened 2024-03-05
Expose the software service through the HTTP/JSON API

imobachgs:
Trello: https://trello.com/c/2MjaulMd/3566-3-expose-the-software-api-over-http

This PR exposes the public[^1] part of the software API through the HTTP/JSON interface. It includes:

* `/software/patterns`: list of patterns, including whether they are selected (and by whom).
* `/software/products`: list of products.
* `/software/probe`: starts the software probing.
* `/software/config`: describing the software configuration.

```json
{
  "patterns": [
    "gnome"
  ],
  "product": "Tumbleweed"
}
```

Additionally, it exposes the following events through the websocket:

* `PatternsChanged`, containing the patterns and who selected them ('user' or 'auto').
* `ProductChanged`, with the name of the new product.

[^1] The part of the D-Bus API that it is used by the web UI.

## Implementation details

In this phase of the development, we are still deciding the best way to implement these HTTP/JSON interfaces. The implementation has two parts:

* The [HTTP/JSON interface itself](https://github.com/openSUSE/agama/blob/http-software-srv/rust/agama-server/src/software/web.rs#L104), which is implemented as a [Router](https://docs.rs/axum/latest/axum/struct.Router.html) and a set of small functions (one per each endpoing+verb).
* An [events stream](https://github.com/openSUSE/agama/blob/http-software-srv/rust/agama-server/src/software/web.rs#L60) which emits Event values (see [stream](https://tokio.rs/tokio/tutorial/streams) in the Tokio documentation).

About the service status, the zbus proxies are cached and can be cloned, so it looks like a good idea to keep our clients as part of the state.

## In the future

There are a few improvements we could consider in the future:

* Split the events in different types depending on the service (`SoftwareEvent`, `ManagerEvent`), so those services only know about their types.
* Consolidate all errors under a common one (e.g., `Error::Software(SoftwareError)`, `Error::Manager(ManagerError)`, etc.) to have a single `IntoResponse` implementation while keeping modules isolation.

### DIFF
--- a/rust/agama-lib/src/software/client.rs
+++ b/rust/agama-lib/src/software/client.rs
@@ -1,6 +1,7 @@
 use super::proxies::Software1Proxy;
 use crate::error::ServiceError;
 use serde::Serialize;
+use std::collections::HashMap;
 use zbus::Connection;
 
 /// Represents a software product
@@ -20,7 +21,26 @@ pub struct Pattern {
     pub order: String,
 }
 
+/// Represents the reason why a pattern is selected.
+#[derive(Clone, Copy)]
+pub enum SelectionReason {
+    /// The pattern was selected by the user.
+    User = 0,
+    /// The pattern was selected automatically.
+    Auto = 1,
+}
+
+impl From<u8> for SelectionReason {
+    fn from(value: u8) -> Self {
+        match value {
+            0 => Self::User,
+            _ => Self::Auto,
+        }
+    }
+}
+
 /// D-Bus client for the software service
+#[derive(Clone)]
 pub struct SoftwareClient<'a> {
     software_proxy: Software1Proxy<'a>,
 }
@@ -55,14 +75,25 @@ impl<'a> SoftwareClient<'a> {
 
     /// Returns the ids of patterns selected by user
     pub async fn user_selected_patterns(&self) -> Result<Vec<String>, ServiceError> {
-        const USER_SELECTED: u8 = 0;
         let patterns: Vec<String> = self
             .software_proxy
             .selected_patterns()
             .await?
             .into_iter()
-            .filter(|(_id, reason)| *reason == USER_SELECTED)
+            .filter(|(_id, reason)| *reason == SelectionReason::User as u8)
             .map(|(id, _reason)| id)
+            .collect();
+        Ok(patterns)
+    }
+
+    /// Returns the selected pattern and the reason each one selected.
+    pub async fn selected_patterns(
+        &self,
+    ) -> Result<HashMap<String, SelectionReason>, ServiceError> {
+        let patterns = self.software_proxy.selected_patterns().await?;
+        let patterns = patterns
+            .into_iter()
+            .map(|(id, reason)| (id, reason.into()))
             .collect();
         Ok(patterns)
     }
--- a/rust/agama-lib/src/software/client.rs
+++ b/rust/agama-lib/src/software/client.rs
@@ -112,6 +112,9 @@ impl<'a> SoftwareClient<'a> {
         }
     }
 
+    /// Returns the required space for installing the selected patterns.
+    ///
+    /// It returns a formatted string including the size and the unit.
     pub async fn used_disk_space(&self) -> Result<String, ServiceError> {
         Ok(self.software_proxy.used_disk_space().await?)
     }
--- a/rust/agama-lib/src/software/client.rs
+++ b/rust/agama-lib/src/software/client.rs
@@ -111,4 +111,8 @@ impl<'a> SoftwareClient<'a> {
             Ok(())
         }
     }
+
+    pub async fn used_disk_space(&self) -> Result<String, ServiceError> {
+        Ok(self.software_proxy.used_disk_space().await?)
+    }
 }
--- a/rust/agama-lib/src/software/client.rs
+++ b/rust/agama-lib/src/software/client.rs
@@ -115,4 +115,9 @@ impl<'a> SoftwareClient<'a> {
     pub async fn used_disk_space(&self) -> Result<String, ServiceError> {
         Ok(self.software_proxy.used_disk_space().await?)
     }
+
+    /// Starts the process to read the repositories data.
+    pub async fn probe(&self) -> Result<(), ServiceError> {
+        Ok(self.software_proxy.probe().await?)
+    }
 }
--- a/rust/agama-lib/src/software/client.rs
+++ b/rust/agama-lib/src/software/client.rs
@@ -22,19 +22,22 @@ pub struct Pattern {
 }
 
 /// Represents the reason why a pattern is selected.
-#[derive(Clone, Copy)]
-pub enum SelectionReason {
+#[derive(Clone, Copy, Debug, Serialize)]
+pub enum SelectedBy {
     /// The pattern was selected by the user.
     User = 0,
     /// The pattern was selected automatically.
     Auto = 1,
+    /// The pattern has not be selected.
+    None = 2,
 }
 
-impl From<u8> for SelectionReason {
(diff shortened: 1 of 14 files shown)


rust/agama-lib/src/software/client.rs
```diff
@@ -20,7 +21,29 @@ pub struct Pattern {
...
+    /// The pattern has not be selected.
+    None = 2,
+}
+
+impl From<u8> for SelectedBy {
+    fn from(value: u8) -> Self {
+        match value {
+            0 => Self::User,
+            1 => Self::Auto,
+            _ => Self::None,
```
jreidinger, line 40: I do not get this. Why on one side we convert "none" to 2 and here we accept anything? This looks kind of strange. I expect some kind of conversion error instead of silently accepting it.

rust/agama-server/src/software/web.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[derive(Clone, Serialize, Deserialize)]
+struct SoftwareConfig {
+    patterns: Option<Vec<String>>,
+    product: Option<String>,
+}
+
+#[derive(Error, Debug)]
+pub enum SoftwareError {
+    #[error("Service error: {0}")]
```
jreidinger, line 44: I am not sure if we should not specify which service like `Software Service Error: {0}`

rust/agama-server/src/software/web.rs
```diff
@@ -0,0 +1,242 @@
...
+    Error(#[from] ServiceError),
+}
+
+impl IntoResponse for SoftwareError {
+    fn into_response(self) -> Response {
+        let body = json!({
+            "error": self.to_string()
```
jreidinger, line 51: maybe also specify kind of error in response for easier introspect?

rust/agama-server/src/software/web.rs
```diff
@@ -0,0 +1,242 @@
...
+
+async fn product_changed_stream(
+    dbus: zbus::Connection,
+) -> Result<impl Stream<Item = Event>, Error> {
+    let proxy = SoftwareProductProxy::new(&dbus).await?;
+    let stream = proxy
+        .receive_selected_product_changed()
+        .await
+        .then(|change| async move {
+            if let Ok(id) = change.get().await {
```
jreidinger, line 75: Can it happen that change result is not `Ok`? And if so, in which case? I would at least document it here.

rust/agama-server/src/software/web.rs
```diff
@@ -0,0 +1,242 @@
...
+        .route("/config", put(set_config).get(get_config))
+        .route("/probe", post(probe))
+        .with_state(state);
+    Ok(router)
+}
+
+/// Returns the list of available products.
+///
+/// * `state`: service state.
+#[utoipa::path(get, path = "/software/products", responses(
```
jreidinger, line 121: should we somehow define/document also that potential error state?

rust/agama-server/src/software/web.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// a number and a unit (e.g., "GiB").
+    size: String,
+}
+
+/// Returns the proposal information.
+///
+/// At this point, only the required space is reported.
+#[utoipa::path(
+    get, path = "/software/proposal", responses(
+        (status = 200, description = "Software proposal")
```
jreidinger, line 222: does it also somehow document how returning json will look like? Will it link it with SoftwareProposal struct automatic?

rust/agama-server/src/web.rs
```diff
@@ -43,13 +53,29 @@ pub fn service(config: ServiceConfig, events: EventsSender) -> Router {
...
+
+    Ok(())
+}
+
+/// Emits the events from the system streams through the events channel.
+///
+/// * `connection`: D-Bus connection.
+/// * `events`: channel to send the events to.
+pub async fn run_events_monitor(dbus: zbus::Connection, events: EventsSender) -> Result<(), Error> {
+    let stream = software_stream(dbus).await?;
```
jreidinger, line 74: so if I get it, here in the end will be list of all our streams and we will all  mix them, right?

rust/agama-server/src/web/docs.rs
```diff
@@ -3,13 +3,27 @@ use utoipa::OpenApi;
...
     components(
         schemas(super::http::PingResponse),
         schemas(crate::l10n::LocaleEntry),
         schemas(crate::l10n::web::LocaleConfig),
         schemas(crate::l10n::Keymap),
-        schemas(crate::l10n::TimezoneEntry)
+        schemas(crate::l10n::TimezoneEntry),
+        schemas(agama_lib::software::Pattern),
+        schemas(agama_lib::product::Product),
+        schemas(crate::software::web::PatternEntry)
```
jreidinger, line 26: will it report if I forgot to add here some part that is marked by derive?

rust/agama-lib/src/software/client.rs
imobachgs: Yes, you are right, I will handle other cases.

rust/agama-server/src/web.rs
imobachgs: Yes, that's the idea. We combine our streams and emit then through the websocket.

rust/agama-server/src/web/docs.rs
imobachgs: No, it does not. I already forgot about a few things in the previous PR :-/

rust/agama-server/src/software/web.rs
imobachgs: No, it does not :-(

rust/agama-server/src/software/web.rs
imobachgs: Given that I added the "Software" prefix, it should be clear now.

rust/agama-server/src/software/web.rs
imobachgs: If it fails, it is because it cannot deserialize the information coming from D-Bus. Our proxy could be outdated, for instance. But I do not think we should put a comment about that in each place where we use this function.

See https://docs.rs/zbus/latest/src/zbus/proxy/mod.rs.html#196-200 (because [get documentation](https://docs.rs/zbus/latest/zbus/proxy/struct.PropertyChanged.html#method.get)) is not enough.

In general, any conversion problem is, most probably, because our proxies are outdated.

rust/agama-server/src/software/web.rs
jreidinger: well, I mean it for easier scripting support, but probably just error is fine now and we can extend it later.

rust/agama-server/src/software/web.rs
jreidinger: ok, so basically internal problem and not user side.

rust/agama-lib/src/software/client.rs
```diff
@@ -32,12 +32,18 @@ pub enum SelectedBy {
...
+impl TryFrom<u8> for SelectedBy {
+    type Error = UnknownSelectedBy;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
         match value {
-            0 => Self::User,
-            1 => Self::Auto,
-            _ => Self::None,
+            0 => Ok(Self::User),
+            1 => Ok(Self::Auto),
```
jreidinger, line 45: now, I am confused, why is not there `2 => Ok(Self::None)`

imobachgs: Because it is not expected to get a `2` from the API (it does not exist).

jreidinger: ah, ok.

jreidinger (approved):

PR #1074, opened 2024-03-06
Allow accessing the web UI from webpack

imobachgs:
Trello: https://trello.com/c/8fSBUP9P/3594-3-improve-agama-web-server-integration-with-the-web-ui

## Accessing Agama server through webpack

When developing the web UI, you need to access Agama web server. Similarly to what we implemented for Cockpit, you can access it directly by setting the `AGAMA_SERVER` environment variable. By default, it is set to `localhost:3000`.

```sh
AGAMA_SERVER=localhost:4567 npm run server
```

The API is available under the `/api` path.

```javascript
const ping = await fetch("/api/ping");
```

## Service the web UI in `agama-web-server`

The new Agama web server will expose the web UI in the root (`/`) path. By default, it uses `$HOME/.local/agama/web-ui`, falling back to `/usr/share/agama/web-ui`. However, you can override that route using the new `--public-dir` argument.

```sh
agama-web-server serve --public-dir /srv/agama/web-ui
```

NOTE: Additionally, we could implement support for reading the public directory from an environment variable (e.g., `AGAMA_PUBLIC_DIR`). In that case, you could use `dotenv` (or your favorite alternative) to set it automatically to `../web/dist`.

# Tasks

- [x] Access Agama server from webpack
- [x] Serve the web UI through `agama-web-server`
- [x] Move the API to `/api`
- [x] Update the documentation

### DIFF
--- a/rust/agama-server/src/web/service.rs
+++ b/rust/agama-server/src/web/service.rs
@@ -6,25 +6,46 @@ use axum::{
     routing::{get, post},
     Router,
 };
-use std::convert::Infallible;
+use std::{
+    convert::Infallible,
+    path::{Path, PathBuf},
+};
 use tower::Service;
-use tower_http::{compression::CompressionLayer, trace::TraceLayer};
+use tower_http::{compression::CompressionLayer, services::ServeDir, trace::TraceLayer};
 
+/// Builder for Agama main service.
+///
+/// It is responsible for building an axum service which includes:
+///
+/// * A static assets directory (`public_dir`).
+/// * A websocket at the `/ws` path.
+/// * An authentication endpointg at `/authenticate`.
+/// * A 'ping' endpoint at '/ping'.
+/// * A number of authenticated services that are added using the `add_service` function.
 pub struct MainServiceBuilder {
     config: ServiceConfig,
     events: EventsSender,
-    router: Router<ServiceState>,
+    api_router: Router<ServiceState>,
+    public_dir: PathBuf,
 }
 
 impl MainServiceBuilder {
-    pub fn new(events: EventsSender) -> Self {
-        let router = Router::new().route("/ws", get(super::ws::ws_handler));
+    /// Returns a new service builder.
+    ///
+    /// * `events`: channel to send events through the WebSocket.
+    /// * `public_dir`: path to the public directory.
+    pub fn new<P>(events: EventsSender, public_dir: P) -> Self
+    where
+        P: AsRef<Path>,
+    {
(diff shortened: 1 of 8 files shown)


web/README.md
```diff
@@ -1,25 +1,18 @@
...
-There are basically two ways how to develop the Agama fronted. You can
-override the original Cockpit plugins with your own code in your `$HOME` directory
-or you can run a development server which works as a proxy and sends the Cockpit
-requests to a real Cockpit server.
-
-The advantage of using the development server is that you can use the
-[Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/)
-feature for automatically updating the code and stylesheet in the browser
-without reloading the page.
+The easiest way to work on the Agama Web UI is to the development server to
```
dgdavid, line 8: I think there are some missing words.

web/webpack.config.js
```diff
@@ -95,17 +96,20 @@ module.exports = {
       // TODO: modify it to not depend on cockpit
       // forward the manifests.js request and patch the response with the
       // current Agama manifest from the ./src/manifest.json file
-      "/manifests.js": {
-        target: cockpitTarget + "/cockpit/@localhost/",
-        // ignore SSL problems (self-signed certificate)
+      // "/manifests.js": {
+      //   target: cockpitTarget + "/cockpit/@localhost/",
```
dgdavid, line 100: Question: why keep comments with Cockpit references. Nothing against Cockpit, of course. I just think these lines are no longer useful.

imobachgs: It is not because of cockpit, but because of the `manifest.js`. We can remove them when we handle the manifest properly.

web/README.md
imobachgs: Ooopps, yes, you are right.

rust/agama-server/src/web/service.rs
```diff
@@ -6,40 +6,65 @@ use axum::{
...
-use tower_http::{compression::CompressionLayer, trace::TraceLayer};
+use tower_http::{compression::CompressionLayer, services::ServeDir, trace::TraceLayer};
 
+/// Builder for Agama main service.
+///
+/// It is responsible for building an axum service which includes:
+///
+/// * A static assets directory (`public_dir`).
+/// * A websocket at the `/ws` path.
+/// * An authentication endpointg at `/authenticate`.
```
dgdavid, line 22: typo

```suggestion
/// * An authentication endpoint at `/authenticate`.
```

dgdavid (approved):
LGTM, just a typo

PR #1078, opened 2024-03-06
WIP: Storage Result Section

dgdavid:
## :construction: WIP :construction: 

This PR is part of the work for a new presentation of the storage proposal result.

It will be better documented once the PR is more mature. Meanwhile, check https://trello.com/c/7TTVVquM (internal link) for more details about the progress.

---

Screenshot from 2024-03-06 17-38-35

### DIFF
--- a/service/lib/agama/dbus/storage/interfaces/device.rb
+++ b/service/lib/agama/dbus/storage/interfaces/device.rb
@@ -33,6 +33,7 @@ end
 
 require "agama/dbus/storage/interfaces/device/block"
 require "agama/dbus/storage/interfaces/device/component"
+require "agama/dbus/storage/interfaces/device/device"
 require "agama/dbus/storage/interfaces/device/drive"
 require "agama/dbus/storage/interfaces/device/filesystem"
 require "agama/dbus/storage/interfaces/device/lvm_vg"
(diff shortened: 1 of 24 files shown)


joseivanlopez:
I like how it looks! And definitely is going to help a lot! We still have to think a proper wording for the delete warning (not only partitions can be deleted) and how to show the list of deleted devices. And, looking at the table, I think having the "New" prefix for things that are going to be created is very useful (e.g., "New XFS partition").

dgdavid:
Proposals for listing what is going to be deleted, 

| Using a table hiding headers | Using a list |
|-|-|
|Screenshot from 2024-03-07 09-32-17 | Screenshot from 2024-03-07 09-39-06 |

ancorgs:
Looks quite ok in general. I just want to see more clearly:

- Which devices (partitions, volume groups, etc.) are new
- Which devices has been resized

I'm also considering whether it would be beneficial to somehow empathize the devices that are part of the new system, so they are identified more easily. That would include devices with a mount point, physical volumes, devices created for booting like PReP or bios_boot... But I fear it would be too much visual information and I trust I good visualization of what is new and what is resized should be enough to get the full picture.

dgdavid:
Closing it in favor of #1088

PR #1082, opened 2024-03-08
Extend storage D-Bus API

joseivanlopez:
## Problem

More information is required to properly represent the storage proposal result:

* Device description
* Filesystem label

This is a follow-up of https://github.com/openSUSE/agama/pull/1071.

## Solution

Add new D-Bus interfaces `Device` , `Partition`, `LVM.LogicalVolume` and adapt `Filesystem` interface.

This changes require https://github.com/yast/yast-storage-ng/pull/1373.

NOTE: This PR goes to a feature branch instead of master.

## Testing

* Added unit tests
* Tested manually


service/lib/agama/dbus/storage/interfaces/device/device.rb
```diff
@@ -0,0 +1,82 @@
...
+            # @return [Integer]
+            def device_sid
+              storage_device.sid
+            end
+
+            # Name of the device.
+            #
+            # @return [String] e.g., "/dev/sda".
+            def device_name
+              storage_device.name
```
ancorgs, line 58: I guess this is safe for now because we only use subtypes of `Y2Storage::Device` that actually define the method name. But take into account that the base class `Y2Storage::Device` does not implement `#name`, only `#display_name` (which can return nil, so not really much safer).

service/lib/agama/dbus/storage/interfaces/device/lvm_vg.rb
```diff
@@ -75,7 +68,6 @@ def lvm_vg_lvs
             def self.included(base)
               base.class_eval do
                 dbus_interface VOLUME_GROUP_INTERFACE do
-                  dbus_reader :lvm_vg_name, "s", dbus_name: "Name"
```
ancorgs, line 78: Note this name is not the same than the device name. For example, the device name could be `/dev/vg0` and this would be `vg0`

joseivanlopez: Actually it uses the device name.

ancorgs (approved):
LGTM

PR #1088, opened 2024-03-11
web: add Result section to Storage page

dgdavid:
Replace the `Actions` section in the storage proposal for a `Result` one which presents the proposal result in a more appealing way  displaying how the storage would look after installation instead of just the list of actions. Among others bits, it hints the user about things like file-systems are going to be created, resized actions and a subtle emphasis in deletions, if any.

The full, plain list of actions is now available in a modal dialog linked right before the new introduced table.

### Some screenshots

| Proposal with errors | Proposal without deletion actions |
|-|-|
| Screenshot from 2024-03-14 09-50-07 | Screenshot from 2024-03-14 09-46-47 |

| Proposal with deletion actions | List of actions dialog |
|-|-|
| Screenshot from 2024-03-14 09-51-36 | Screenshot from 2024-03-14 09-51-41 |

:warning: The screenshots above are just a bit updated: since the collapse/expand feature is not yet implemented for the Result table, the arrow toggles has been hiding with a nasty CSS hack. The plan could be either, to add the collapsible feature to the table or to see how to tell PF/TreeTable to render the structure without the collapsible feature... if that's is possible.
 

---

Replaces #1078


web/src/components/storage/DevicesManager.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /**
+   * Devices deleted.
+   *
+   * @note The devices are extracted from the actions.
+   *
+   * @returns {StorageDevice[]}
+   */
+  deletedDevices() {
```
dgdavid, line 155: I wonder if that method could receive a flag for excluding subvolumes. Use case when rendering the amount of planned delete actions. What is more, I wonder why we have to filter the subvolume actions there.

joseivanlopez: > I wonder if that method could receive a flag for excluding subvolumes. Use case when rendering the amount of planned delete actions. 

This method cannot be used for counting the amount of delete actions. It returns the deleted devices (e.g., partitions, VGs, MD RAIDs, etc). The number of delete actions can be more than the number of deleted devices.

To know the number of delete actions:

`devicesManager.actions.filter(a => a.delete).length`

And if you want to exclude the subvolume actions, then:

`devicesManager.actions.filter(a => a.delete && !a.subvol).length`

> What is more, I wonder why we have to filter the subvolume actions there.

I don't get this. Where are the subvolume actions filtered out?

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+      <div className="split">
+        <span>{name}</span>
+      </div>
+    );
+  };
+
+  const renderNewLabel = (item) => {
+    if (!item.sid) return;
+
+    // FIXME New PVs over a disk is not detected as new.
```
ancorgs, line 124: We would need a Trello card collecting all the shortcuts taken and the corner-cases found on this pull request.

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+  staging = [],
+  actions = [],
+  errors = [],
+  isLoading = false
+}) {
+  if (isLoading) errors = [];
+
+  const description = sprintf(
+    // TRANSLATORS: %d will be replaced by the amount actions
+    _("During installation, %d actions will be performed to configure the system as displayed below"),
```
ancorgs, line 259: No `_n` here?

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+    actions.length
+  );
+
+  return (
+    <Section
+      // TRANSLATORS: The storage "Result" section's title
+      title={_("Result")}
+      // TRANSLATORS: The storage "Result" section's description
```
ancorgs, line 267: ¿A message for translators without a text to translate?

web/src/components/storage/DevicesManager.js
```diff
@@ -0,0 +1,207 @@
...
+    return compact(devices);
+  }
+
+  /**
+   * Systems deleted.
+   *
+   * @returns {string[]}
+   */
+  deletedSystems() {
+    const systems = this.deletedDevices()
```
ancorgs, line 169: This can add the same system twice if there is an action on the partition and on the disk (eg. deleting the partition table). Since both devices reference the same system.

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+      </>
+    );
+  };
+
+  const renderResizedLabel = (item) => {
+    if (!item.sid || !devicesManager.isShrunk(item)) return;
+
+    return (
+      <Tag variant="orange">
+        {sprintf(_("Resized %s"), deviceSize(devicesManager.shrinkSize(item)))}
```
ancorgs, line 163: It would be good to add a comment for translators here explaining what `%s` is.

web/src/components/core/Reminder.jsx
```diff
@@ -62,18 +62,20 @@ const ReminderTitle = ({ children }) => {
  * @param {object} props
  * @param {string} [props.icon] - The name of desired icon.
  * @param {JSX.Element|string} [props.title] - The content for the title.
- * @param {string} [props.role="status"] - The reminder's role, "status" by
+ * @param {string} [props.role="status"] - The reminder's role, "status" by default.
+ * @param {("subtle")} [props.variant] - The reminder's variant, none by default.
```
ancorgs, line 66: ¿subtle?

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+  if (total === 0) return;
+
+  // TRANSLATORS: %d will be replaced by the amount of destructive actions
+  const warningTitle = sprintf(n_(
+    "There is %d destructive action planned",
+    "There are %d destructive actions planned",
+    total
+  ), total);
+
+  // FIXME: Use the Intl.ListFormat instead of the `join(", ")` used below.
```
ancorgs, line 60: Something to add to the mentioned follow-up card.

web/src/components/storage/ProposalPage.test.jsx
```diff
@@ -76,6 +76,7 @@ const vdb = {
   size: 1e+6
 };
 
+// FIXME: needed to be reviewed/adapted after latest changes
```
ancorgs, line 79: I don't get what this comment refers to.

web/src/components/storage/ProposalResultSection.jsx
ancorgs: Thinking it twice, I don't think that can ever happen. The proposal only create PVs on partitions. The `GuidedProposal` is partition-driven. It can reuse a given VG, LV or partition, but it will never turn a block device into a PV without partitioning it.

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,286 @@
...
+  // Most probably, a `listFormat` or similar wrapper should live in src/i18n.js or so.
+  // Read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat
+  return (
+    <Reminder title={warningTitle} variant="subtle">
+      <If
+        condition={systems.length > 0}
+        then={
+          <p>
+            {_("Including the deletion of")} <strong>{systems.join(", ")}</strong>
```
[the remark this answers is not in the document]
dgdavid: The thing is that I'd like to avoid abusing of `dangerouslySetInnerHTML` and that's the reason I did not use a replacement there.

Is it enough for you something like?

```suggestion
            {_("Affecting")} <strong>{systems.join(", ")}</strong>
```

web/src/components/storage/ProposalPage.test.jsx
dgdavid: It is a left over. I added it at the very beginning of this sprint because I marked all tests as `skip`ed.

web/src/components/storage/ProposalResultSection.jsx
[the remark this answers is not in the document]
ancorgs: Good enough for now, yes.

web/src/components/storage/ProposalResultSection.jsx
ancorgs: Done https://trello.com/c/6UDnDvdU/365-agama-several-fixes-improvements

web/src/components/storage/DevicesManager.js
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Devices deleted.
+   *
+   * @note The devices are extracted from the actions.
+   *
+   * @returns {StorageDevice[]}
+   */
+  deletedDevices() {
+    return this.#deleteActionsDevice().filter(d => !d.isDrive);
```
ancorgs, line 156: I'm not sure this is the most general fix for the problem it tries to address.

I would say it makes more sense (although likely is most complex) to get the systems from the disk and filter out the subdevices (partitions) of that disk.

joseivanlopez: Note that `#deletedDevices` is not used for gettng the deleted systems. This method is for getting the list of deleted devices (partitions, VGs, etc). And it is filtering drives out to not include them as deleted (i.e., when a partition table is deleted or a file system is removed from a direclty formatted drive). Maybe we could simplify the detection of deleted devices by simply returning the system devices that do not appear in staging (instead of find them in the actions).

Then, for the list of systems, the strategy is similar to what you suggested, but the other way around. That is, the devices with partition table are not considered (because the systems come from the partitions), see `#deletedSystems`.

ancorgs: Discussed offline. Let's leave it as it is for now.

dgdavid:
@ancorgs, @joseivanlopez 

I've sent commit f13fd87 to fix the _Device_ column alignment. The whole thing about the CSS overrides deserves a review in later iterations, but now it looks better.

| Before | After| 
|-|-|
| Screenshot from 2024-03-15 17-02-20 | Screenshot from 2024-03-15 17-02-08 |

dgdavid:
I've "fixed" the problem with not needed "cuts" in column headers. See screenshot below forcing to show the `Resize` label to check that `Mount point` column does not get affect.

As said above, all of this needs a refinement and the addition of the "fit-content" utility class to all columns could be a prop of the core/TreeTable component instead of adding it manually always.

But for now I perfer merging these styles as they are and apply all the improvement all together in a specific commit.
Screenshot from 2024-03-15 17-33-06

web/src/components/storage/DevicesManager.js
dgdavid: > > What is more, I wonder why we have to filter the subvolume actions there.
> 
> I don't get this. Where are the subvolume actions filtered out?

Nevermind. I meant that we have a method for filtering the subvolumes from actions in the ProposalResultSection.jsx at line ~232. But we have a "DevicesManager" object that provides a "deleteDevices" method and, in my opinion, could provide a "deleteSystems" too. Although I konw, it is a "DeviceManager" not a "SystemsManager".

web/src/components/storage/ProposalResultSection.jsx
```diff
@@ -0,0 +1,292 @@
...
+  // FIXME: Use the Intl.ListFormat instead of the `join(", ")` used below.
+  // Most probably, a `listFormat` or similar wrapper should live in src/i18n.js or so.
+  // Read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat
+  return (
+    <Reminder title={warningTitle} variant="subtle">
+      <If
+        condition={systems.length > 0}
+        then={
+          <p>
+            {_("Affecting %s")} <strong>{systems.join(", ")}</strong>
```
ancorgs, line 69: NP: this likely needs a comment for the translators, although I guess the string is not definitive anyways so we can live with it as it is.

ancorgs (approved):
LGTM, I would just add another comment for translators.

web/src/components/storage/ProposalResultSection.jsx
dgdavid: Good you comment it, because there is a `%s` leftover :disappointed_relieved:

ancorgs (approved):
LGTM

PR #1090, opened 2024-03-12
WIP: SpacePolicy as pop-up

ancorgs:
Just to upload the progress.


web/src/components/storage/ProposalSpacePolicyField.jsx
```diff
@@ -48,22 +48,48 @@ const SPACE_POLICIES = [
...
   },
   {
     id: "custom",
     label: N_("Custom"),
-    description: N_("Select what to do with each partition.")
+    description: N_("Select what to do with each partition."),
+    summaryLabels: [
+      // TRANSLATORS: This is presented next to the label "Find space", so the whole sentence
+      // would read as "Find space performing a custom set of actions".
+      N_("performing a custom set of actions")
```
dgdavid, line 91: NP: I have asked about the need of using `N_` and `_` here instead of directly `_`. See https://github.com/openSUSE/agama/pull/726#discussion_r1524383469

Asking because I've used directly `_(` in other places and, unless I'm wrong, it works.

web/src/components/storage/ProposalSpacePolicyField.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+
+const openPopup = async (props = {}) => {
+  const allProps = { policy, devices, actions, ...props };
+  const { user } = plainRender(<ProposalSpacePolicyField { ...allProps } />);
+  const button = screen.getByRole("button");
+
+  await user.click(button);
+  const dialog = screen.getByRole("dialog", { name: "Find Space" });
+  return { user, dialog };
+};
```
dgdavid, line 113: Cool! :ok_hand:

web/src/components/storage/ProposalSpacePolicyField.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+      within(picker).getByRole("option", { name: /custom/i });
+    });
+
+    describe("when there are no installation devices", () => {
+      beforeEach(() => {
+        devices = [];
+      });
+
+      it("does not render the policy actions", async () => {
+        const { dialog } = await openPopup();
```
dgdavid, line 159: That saves a lot of duplicated lines! :tada:

web/src/components/storage/ProposalSpacePolicyField.jsx
dgdavid: According to the @lslezak answer, you do not need '`N_` here, unless we decide to start wrapping `Object.freeze` to these objects.

dgdavid (commented):
So far, it looks good. I really like the simple, but clever, trick you did to avoid repeating same lines in the test examples. I wonder why I didn't think on it before :stuck_out_tongue_winking_eye: I'm gonna think that I'm simply less lazy than you :smile:

dgdavid:
> Move the space policy configuration to a popup.

Just  a note for people reading this. It has been moved not only for avoid having too much tables in the same page. We also found a kind of a pattern in that page which is, more or less, to move to a dialog things that are not applied immediately or things that are actually related with the real status of the current system, not the proposal that is going on.

Of course, we can elaborate this a bit more.

web/src/components/storage/ProposalSpacePolicyField.jsx
```diff
@@ -260,6 +237,18 @@ const DeviceRow = ({
   onCollapse = noop,
   onChange = noop
 }) => {
+  // Generates the action value according to the policy.
+  const action = () => {
```
ancorgs, line 241: I'm not sure if this (together with the `useEffect` below) is the best (ie. most usable) behavior, but I find it reasonable enough to get this merged. We can re-evaluate in the future based on user's feedback.

web/src/components/storage/ProposalSpacePolicyField.jsx
```diff
@@ -428,45 +418,137 @@ const SpacePolicyPicker = ({ currentPolicy, onChange = noop }) => {
...
+  const [customUsed, setCustomUsed] = useState(false);
 
-  const changeSpaceActions = (spaceAction) => {
-    const spaceActions = settings.spaceActions.filter(a => a.device !== spaceAction.device);
+  // The selectors for the space action have to be initialized always to the same value
+  // (e.g., "keep") when the custom policy is selected for first time. The following two useEffect
+  // ensures that.
+
+  // Stores whether the custom policy has been used.
+  useEffect(() => {
```
ancorgs, line 438: See https://github.com/openSUSE/agama/pull/1090/files#r1526086667. As said there, I'm not sure this will be the final behavior, but it's not a stopper to merge (is actually quite good).

ancorgs (commented):
I approve changes introduced by @joseivanlopez, but I cannot effectively approve the PR since I'm the original author.

PR #1091, opened 2024-03-13
first draft of questions API in http

jreidinger:
## Problem

Questions are not exposed in new architecture 2024

## Solution

Add path to enlist questions and also announce when question is answered or new appear on websocket.

## Testing

- *Tested manually*


rust/agama-server/src/questions.rs
```diff
@@ -94,7 +95,7 @@ trait AnswerStrategy {
     ///
     /// I gets as argument the question to answer. Returned value is pair
     /// of `answer` and `password` properties. If `None` is used in any
-    /// position it means that this object does not respond to given property.
```
imobachgs, line 97: np: typo

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,151 @@
+//! This module implements the web API for the software module.
```
imobachgs, line 1: copy&paste error ;-)

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,151 @@
...
+    routing::{get, put},
+    Json, Router,
+};
+use tokio_stream::{Stream, StreamExt};
+use zbus::fdo::ObjectManagerProxy;
+use thiserror::Error;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+
+// TODO: move to lib
```
imobachgs, line 26: :+1:

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,151 @@
...
+        .with_state(state);
+    Ok(router)
+}
+
+pub async fn questions_stream(dbus: zbus::Connection) -> Result<impl Stream<Item = Event>, Error> {
+    let proxy = ObjectManagerProxy::new(&dbus).await?;
+    let add_stream = proxy
+        .receive_interfaces_added()
+        .await?
+        .then(|_| async move {
```
imobachgs, line 127: I do not know what your plan is here, but to simplify things a bit, you could include the question number (taken from the D-Bus path) and let the client (on the JavaScript side) retrieve the question.

imobachgs (commented):
It is looking good, although I know that quite some work still remains. I just added a comment about how to implement the stream.

About answering a question, we could implement a D-Bus signal that it is more convenient than having to listen for properties changes. Is it worth it?

rust/agama-server/src/questions/web.rs
jreidinger: yeah, you find my inspiration :)

rust/agama-server/src/questions/web.rs
jreidinger: well, current API does not allow to retrieve single question. That is different to DBus.
Basically my idea about API is:
1. list all unanswered questions at `/questions`
2. answer specific question according to its id with `/questions/:id/answer`
3. get event that list of unanswered questions changed
4. and now when I am thinking about it, we need also way to provide answers file, so some API for it.

Reason why I design it this way is that common workflow is question arise, answer is provided and then another question arise. I think that having multiple unanswered question is quite rare situation, so I optimize API for that single case, but do not prevent in future to extend it e.g. to also get single question if there will be many questions.

jreidinger:
> It is looking good, although I know that quite some work still remains. I just added a comment about how to implement the stream.
> 
> About answering a question, we could implement a D-Bus signal that it is more convenient than having to listen for properties changes. Is it worth it?

Well, as said, issue is that it is not about properties, but about object manager...which makes sense for dbus, but working with it in signals is not so easy...especially when it talks about interfaces, so we had that known race condition when you have question with multiple interfaces. So maybe having specific signal on Questions interface that will just say questions changed or two specialized like `NewQuestion` and `QuestionAnswered` can help and maybe makes sense even for http API?

rust/agama-server/src/questions/web.rs
jreidinger: only tricky part with moving to lib is that structs for web does not live there...There are GenericQuestion struct, so after move I will need to implement some `Into`

rust/agama-server/src/questions.rs
jreidinger: thanks for catch. I have to say that sometime vscode together with middle button of mouse do serious mess.

rust/agama-server/src/questions/web.rs
imobachgs: OK, I get the idea. About the answers file, we are not using it right now, so you can skip that part.

imobachgs:
> > It is looking good, although I know that quite some work still remains. I just added a comment about how to implement the stream.
> > About answering a question, we could implement a D-Bus signal that it is more convenient than having to listen for properties changes. Is it worth it?
> 
> Well, as said, issue is that it is not about properties, but about object manager...which makes sense for dbus, but working with it in signals is not so easy...especially when it talks about interfaces, so we had that known race condition when you have question with multiple interfaces. So maybe having specific signal on Questions interface that will just say questions changed or two specialized like `NewQuestion` and `QuestionAnswered` can help and maybe makes sense even for http API?

If adding those signals could help, I am all for it.

rust/agama-server/src/questions/web.rs
imobachgs: Well, actually `GenericQuestion` can be reused in other contexts, right? So I do not see the problem of moving that struct to lib too.

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,210 @@
...
+
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
+pub struct Question {
+    generic: GenericQuestion,
+    with_password: Option<QuestionWithPassword>,
+}
+
+/// Facade of agama_lib::questions::GenericQuestion
+/// For fields details see it.
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
```
imobachgs, line 136: Why do you need a facade? If it is because of the `utoipa::ToSchema`, it is fine with me. I have derived the `utoipa::ToSchema` for some structs in lib and it feels wrong to me.

However, having to use a facade just because of our documentation library feels wrong too.

jreidinger: No, it is not related to utoipa at all. It is because I do not want to break dbus code and I like more the approach with generic question and composition of question parts. Original code contain QuestionWithPassword that contain link to GenericQuestion, which I do not like much.

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,271 @@
...
+            .build()
+            .await?;
+        dbus_generic
+            .set_answer(answer.generic.answer.as_str())
+            .await?;
+        Ok(())
+    }
+}
+
+#[derive(Error, Debug)]
```
imobachgs, line 143: This error do not bring any value. As we already did in the software layer, you can directly use [crate::error::Error](https://github.com/openSUSE/agama/blob/architecture_2024/rust/agama-server/src/error.rs#L12).

rust/agama-server/src/questions/web.rs
imobachgs: Well, as soon as you do not break the D-Bus external API, you can refactor the internals if you wish. If you decide to keep the facade, please, write down the reason in the comment.

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,271 @@
...
+    class: String,
+    text: String,
+    options: Vec<String>,
+    default_option: String,
+    data: HashMap<String, String>,
+}
+
+/// Facade of agama_lib::questions::WithPassword
+/// For fields details see it.
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
```
imobachgs, line 183: Same than above.

rust/agama-server/src/questions/web.rs
```diff
@@ -0,0 +1,271 @@
...
+pub async fn questions_stream(
+    dbus: zbus::Connection,
+) -> Result<Pin<Box<dyn Stream<Item = Event> + Send>>, Error> {
+    let question_path = OwnedObjectPath::from(
+        ObjectPath::try_from("/org/opensuse/Agama1/Questions")
+            .context("failed to create object path")?,
+    );
+    let proxy = ObjectManagerProxy::builder(&dbus)
+        .path(question_path)
+        .context("Failed to create object manager path")?
```
imobachgs, line 225: To be honest, I am still not a big fan of using `anyhow` in libraries. But I can leave with that :-)

imobachgs (approved):

rust/agama-server/src/questions/web.rs
jreidinger: well, I will write it to comment. In DBus we have attributes and in question is included also answer attribute. Which is not what I want to http API. I have there two parts: 1. question and 2. answer. So original one struct is split into two and the first one is used as output and the second as expected input. At least that is my idea. As said I will write it to comment

rust/agama-server/src/questions/web.rs
jreidinger: me neither, but it is much faster then adding each zvariant error :)

PR #1092, opened 2024-03-13
Expose the Progress interface through the HTTP/JSON API

imobachgs:
Trello: https://trello.com/c/2RvcZBFR/3564-5-expose-the-progress-api-over-http

This PR exposes the `Progress` API for the Manager and Software services through the HTTP/JSON interface. It is implemented as a pair of functions that can be reused in all services that want to implement such an interface.

## A `/SERVICE/progress` endpoint

The `progress_router` function allows to add a `/SERVICE/progress` route that exposes the current progress:

```json
{
  "current_step": 4,
  "max_steps": 4,
  "current_title": "Calculating the software proposal",
  "finished": false
}
```

## The events stream

The `progress_stream` builds an events stream that emits a new event when the `current_step` changes:

```json
{
  "type": "Progress",
  "service": "org.opensuse.Agama.Software1",
  "current_step": 4,
  "max_steps": 4,
  "current_title": "Calculating the software proposal",
  "finished": false
}
```

## Enabling `ServiceStatus` and `Progress` for the Software service

Additionally, the PR enables the `ServiceStatus` (implemented https://github.com/openSUSE/agama/pull/1089) and the `Progress` for the Software service too.


rust/agama-server/src/software/web.rs
```diff
@@ -61,8 +67,24 @@ impl IntoResponse for SoftwareError {
...
+                "/org/opensuse/Agama/Software1",
+            )
+            .await?,
+        ),
+        progress_stream(
+            dbus,
+            "org.opensuse.Agama.Software1",
+            "/org/opensuse/Agama/Software1",
+        )
+        .await,
```
jreidinger, line 87: well, tokio explicitelly mention to not do multiple merge. See documentation at https://dtantsur.github.io/rust-openstack/tokio/stream/trait.StreamExt.html#method.merge
They suggest to use StreamMap for such case https://docs.rs/tokio-stream/latest/tokio_stream/struct.StreamMap.html

rust/agama-server/src/web/common.rs
```diff
@@ -108,3 +116,120 @@ async fn build_service_status_proxy<'a>(
...
+}
+
+#[derive(Clone)]
+struct ProgressState<'a> {
+    proxy: ProgressProxy<'a>,
+}
+
+async fn progress(State(state): State<ProgressState<'_>>) -> Json<Progress> {
+    let proxy = state.proxy;
+    let progress = Progress::from_proxy(&proxy).await.unwrap();
```
jreidinger, line 168: so it should crash when there is issue with dbus? I think that having some error status would be nicer.

rust/agama-server/src/web/common.rs
```diff
@@ -108,3 +116,120 @@ async fn build_service_status_proxy<'a>(
...
+    Json(progress)
+}
+
+#[pin_project]
```
jreidinger, line 172: I need to read some documentation about that pinning.

rust/agama-server/src/web/common.rs
```diff
@@ -108,3 +116,120 @@ async fn build_service_status_proxy<'a>(
...
+    pub async fn new(proxy: ProgressProxy<'a>) -> Self {
+        let stream = proxy.receive_current_step_changed().await;
+        ProgressStream {
+            inner: stream,
+            proxy,
+        }
+    }
+}
+
+impl<'a> Stream for ProgressStream<'a> {
```
jreidinger, line 200: own stream? nice

rust/agama-server/src/web/common.rs
```diff
@@ -108,3 +116,120 @@ async fn build_service_status_proxy<'a>(
...
+}
+
+async fn build_progress_proxy<'a>(
+    dbus: &zbus::Connection,
+    destination: &str,
+    path: &str,
+) -> Result<ProgressProxy<'a>, zbus::Error> {
+    let proxy = ProgressProxy::builder(&dbus)
+        .destination(destination.to_string())?
+        .path(path.to_string())?
```
jreidinger, line 231: I am not sure if there is advantage of using `&str` when you inside always needs to create new string. If you have String in method call already, then it can be passed or created outside and just passed.

rust/agama-server/src/web/event.rs
```diff
@@ -8,13 +8,28 @@ use tokio::sync::broadcast::{Receiver, Sender};
...
+    InstallationPhaseChanged {
+        phase: InstallationPhase,
+    },
+    BusyServicesChanged {
+        services: Vec<String>,
+    },
+    ServiceStatusChanged {
+        service: String,
+        status: u32,
+    },
```
jreidinger, line 32: I expect that they usually come together that Busy and ServiceStatus change?

rust/agama-server/src/web/common.rs
imobachgs: Yes, sure, it is just a draft. There are a couple of `unwraps` here and there.

rust/agama-server/src/software/web.rs
imobachgs: Yes, I want to search for an alternative. Thanks for the hint, StreamMap looks good enough. I was also checking the [StreamExt of the futures crate](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html).

rust/agama-server/src/web/common.rs
imobachgs: Same for me. I need to dig deeper on that topic.

rust/agama-server/src/web/common.rs
imobachgs: Yes, it allows us to keep some additional state.

rust/agama-server/src/web/event.rs
imobachgs: They are different things: one is the list of busy services, which is only kept by the `Manager`. The `ServiceStatusChanged` belongs to each separate service. However, I am thinking that perhaps we are not interested in the `BusyServicesChanged` event from the web UI. I will check.

rust/agama-server/src/web/common.rs
imobachgs: Well, using `&str` in the boundaries looks more flexible to me. And, after all, from outside you do not need to know what is going to happen inside. But yes, I could reconsider it.

rust/agama-server/src/web/common.rs
jreidinger: And I did it. Quite interesting topic. Basically it is about guarantees about memory ( location and content ). I think it is mainly useful if you need to do some unsafe code, to be more safer :)

rust/agama-server/src/web.rs
```diff
@@ -72,13 +73,51 @@ pub async fn run_monitor(events: EventsSender) -> Result<(), ServiceError> {
...
+        .await?,
+    );
+    stream.insert(
+        "software-progress",
+        progress_stream(
+            dbus.clone(),
+            "org.opensuse.Agama.Software1",
+            "/org/opensuse/Agama/Software1",
+        )
+        .await,
```
jreidinger: :+1: looks better.

jreidinger (approved):
LGTM

PR #1094, opened 2024-03-15
Adapt product selection and handling ofl10n settings with the HTTP/JSON server

imobachgs:
Although the new HTTP/JSON interface is not ready yet, we have started adapting part of the JavaScript code. This pull request introduces quite some changes to move from the login page to the overview, although most of the sections are disabled.

<details>
<summary>Login, selecting a product and changing l10n settings</summary>

overview
</summary>
</details>

## Changes

* Synchronize the l10n models we use in the backend and the frontend.
* Adapt the ManagerClient to the HTTP/JSON API and uncomment the related code (e.g., start probing after selecting a product).
* Adjust the L10nClient to recent changes.
* Adapt the L10nSection and L10nPage components.

### DIFF
--- a/web/src/client/manager.js
+++ b/web/src/client/manager.js
@@ -22,7 +22,7 @@
 // @ts-check
 
 import DBusClient from "./dbus";
-import { WithStatus, WithProgress } from "./mixins";
+import { WithProgress, WithStatus } from "./mixins";
 import cockpit from "../lib/cockpit";
 
 const MANAGER_SERVICE = "org.opensuse.Agama.Manager1";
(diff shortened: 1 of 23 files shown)


web/src/client/manager.js
```diff
@@ -108,33 +100,37 @@ class ManagerBaseClient {
...
 }
 
 /**
-  * Client to interact with the Agama manager service
-  */
-class ManagerClient extends WithProgress(WithStatus(ManagerBaseClient, MANAGER_PATH), MANAGER_PATH) { }
+ * Client to interact with the Agama manager service
+ */
+class ManagerClient extends WithProgress(
+  WithStatus(ManagerBaseClient, "/manager/status", "org.opensuse.Agama.Manager1"),
```
dgdavid, line 132: Why not having a variable for "org.opensuse.Agama.Manager1" too? Is it always in sync with MANAGER_PATH?

I mean, MANAGER_PATH.replace("/", ".") should work always?

dgdavid (approved):
LGTM, even the Rust part :wink:

joseivanlopez (commented):
LGTM

web/src/client/manager.js
imobachgs: I don't know if it will work in all cases (apart from the fact that you need to remove the leading `/`).

PR #1095, opened 2024-03-15
Simplify server errors and drop some unused code

imobachgs:
* Simplify the error handling in the `agama-server` crate, unifying many.
* Remove some functions from the `Locale` D-Bus API that are irrelevant for inter-process communication.
* +88/-215 lines and removed some redundant `IntoResponse` implementations.

### DIFF
--- a/rust/agama-server/src/l10n/web.rs
+++ b/rust/agama-server/src/l10n/web.rs
@@ -9,20 +9,16 @@ use crate::{
 use agama_locale_data::{InvalidKeymap, LocaleId};
 use axum::{
     extract::State,
-    http::StatusCode,
-    response::{IntoResponse, Response},
     routing::{get, put},
     Json, Router,
 };
 use serde::{Deserialize, Serialize};
-use serde_json::json;
 use std::{
     process::Command,
     sync::{Arc, RwLock},
 };
-use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(thiserror::Error, Debug)]
 pub enum LocaleError {
     #[error("Unknown locale code: {0}")]
     UnknownLocale(String),
@@ -30,19 +26,8 @@ pub enum LocaleError {
     UnknownTimezone(String),
     #[error("Invalid keymap: {0}")]
     InvalidKeymap(#[from] InvalidKeymap),
-    #[error("Cannot translate: {0}")]
-    OtherError(#[from] Error),
-    #[error("Cannot change the local keymap: {0}")]
-    CouldNotSetKeymap(#[from] std::io::Error),
-}
-
-impl IntoResponse for LocaleError {
-    fn into_response(self) -> Response {
-        let body = json!({
-            "error": self.to_string()
-        });
-        (StatusCode::BAD_REQUEST, Json(body)).into_response()
-    }
+    #[error("Could not apply the changes")]
+    Commit(#[from] std::io::Error),
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


rust/agama-server/src/l10n/web.rs
```diff
@@ -119,14 +104,14 @@ async fn keymaps(State(state): State<LocaleState>) -> Json<Vec<Keymap>> {
 async fn set_config(
     State(state): State<LocaleState>,
     Json(value): Json<LocaleConfig>,
-) -> Result<Json<()>, LocaleError> {
+) -> Result<Json<()>, Error> {
```
jreidinger, line 107: generic comment for whole this PR.
1. I like reduction of anyhow errors
2. I do not like this change in method signature. I think it is better to return specific error and use upper `Error` only when it can return more types of error.

rust/agama-server/src/l10n/web.rs
```diff
@@ -119,14 +104,14 @@ async fn keymaps(State(state): State<LocaleState>) -> Json<Vec<Keymap>> {
...
     let mut data = state.locale.write().unwrap();
     let mut changes = LocaleConfig::default();
 
     if let Some(locales) = &value.locales {
         for loc in locales {
             if !data.locales_db.exists(loc.as_str()) {
-                return Err(LocaleError::UnknownLocale(loc.to_string()));
+                return Err(LocaleError::UnknownLocale(loc.to_string()))?;
```
jreidinger, line 114: why it needs `?` here? it looks kind to strange to return Err and using `?`for returning it

imobachgs: Because it returns a `LocaleError`, not a `Error` itself. In the next iteration, my idea is to put most of this code in a separate struct, and keep this function to just do the arguments lifting. At that point, we could get rid of this `return Err` and the code will look better.

rust/agama-server/src/l10n/web.rs
imobachgs: However, if you prefer, I could keep the `LocaleError` and implement `IntoResponse` only for that kind.

jreidinger: you cannot have IntoResponse for whole Enum? If it is implemented for Error, I kind of expect it can process also element of Enum.

jreidinger: or maybe some kind of trait can help there?

imobachgs: It is implemented for the whole enum (the [crate::error::Error](https://github.com/openSUSE/agama/blob/better-errors/rust/agama-server/src/error.rs#L12)) and that's the reason I am returning `Error` instead of `LocaleError`. After all, `LocaleError` knows nothing about `Error`.

Adding any trait is overcomplicating things, in my opinion. So I would go for:

1) keep it as it is (returning `Error`) or...
2) returning `LocaleError` and implementing `IntoResponse`.

jreidinger: Well, reason why I do not like it is because now it really see only use case for respond, but if it will be used in other parts of rust, then user of this function needs to handle any generic error, while we know that it can product only LocaleError.
So how I do see it is, that if we want to keep extensibility and allow in future to other types of Error in this function, lets change it to Error. But if we know that we always produce only LocaleError ( which I kind of expect ), then I prefer to return minimal abstraction it uses.

But ( and it is big BUT ) I do not care much in web functions that is short and really serve only as connection to real code and do some error translation. But this specific function is quite big and I would say many of its part is not web specific. So maybe if it lives elsewhere and this web function just call other method and translate its errortype into Error. Then it would much nicer.

And to conclude it, I still see this PR as improvement, so feel free to merge it.

jreidinger (approved):
as written in https://github.com/openSUSE/agama/pull/1095#discussion_r1529228567 feel free to merge it

rust/agama-server/src/l10n/web.rs
imobachgs: Yes, I [fully agree](https://github.com/openSUSE/agama/pull/1095#discussion_r1526457385), this code does not belong here but it is a step in the right direction.

I have added a note to [this Trello card](https://trello.com/c/npg6apH0/351-make-l10n-http-json-and-d-bus-interfaces-share-the-same-state) so it can be addressed later.

PR #1100, opened 2024-03-18
Add support to expose the Issues API

imobachgs:
Trello: https://trello.com/c/7SUGZlbi/349-expose-the-issues-interface-over-http

* It enables the `org.opensuse.Agama1.Issue` interface over D-Bus.
* It fixes a problem when handling status changes (`WithStatus`).

### DIFF
--- a/rust/agama-server/src/software/web.rs
+++ b/rust/agama-server/src/software/web.rs
@@ -114,12 +114,12 @@ fn reason_to_selected_by(
 pub async fn software_service(dbus: zbus::Connection) -> Result<Router, ServiceError> {
     const DBUS_SERVICE: &'static str = "org.opensuse.Agama.Software1";
     const DBUS_PATH: &'static str = "/org/opensuse/Agama/Software1";
+    const DBUS_PRODUCT_PATH: &'static str = "/org/opensuse/Agama/Software1/Product";
 
     let status_router = service_status_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
     let progress_router = progress_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
     let software_issues = issues_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
-    let product_issues =
-        issues_router(&dbus, DBUS_SERVICE, "/org/opensuse/Agama/Software1/Product").await?;
+    let product_issues = issues_router(&dbus, DBUS_SERVICE, DBUS_PRODUCT_PATH).await?;
 
     let product = ProductClient::new(dbus.clone()).await?;
     let software = SoftwareClient::new(dbus).await?;
--- a/rust/agama-server/src/software/web.rs
+++ b/rust/agama-server/src/software/web.rs
@@ -8,7 +8,7 @@
 use crate::{
     error::Error,
     web::{
-        common::{progress_router, service_status_router},
+        common::{issues_router, progress_router, service_status_router},
         Event,
     },
 };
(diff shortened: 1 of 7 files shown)


rust/agama-server/src/software/web.rs
```diff
@@ -129,6 +132,8 @@ pub async fn software_service(dbus: zbus::Connection) -> Result<Router, ServiceE
         .route("/probe", post(probe))
         .merge(status_router)
         .merge(progress_router)
+        .nest("/issues/product", product_issues)
+        .nest("/issues/software", software_issues)
```
jreidinger, line 136: to be honest I am not sure if I expect subtree under issue or other way `/software/issues`. Probably later makes more sense to me. Or having general `/issues` call. Would it be possible?

rust/agama-server/src/web.rs
```diff
@@ -114,6 +114,24 @@ async fn run_events_monitor(dbus: zbus::Connection, events: EventsSender) -> Res
...
+        "software-issues",
+        issues_stream(
+            dbus.clone(),
+            "org.opensuse.Agama.Software1",
+            "/org/opensuse/Agama/Software1",
+        )
+        .await?,
+    );
+    stream.insert(
+        "software-product-issues",
```
jreidinger, line 127: similar here. I would either expect some id in issue report that can be used to make it specific or at least use just `product-issues`, otherwise naming looks inconsistent

web/src/client/mixins.js
```diff
@@ -226,8 +210,8 @@ const WithProgress = (superclass, progress_path, service_name) =>
      * @return {import ("./dbus").RemoveFn} function to disable the callback
      */
     onProgressChange(handler) {
-      return this.client.onEvent("Progress", (progress) => {
-        if (progress?.service === service_name) {
+      return this.client.onEvent("Progress", ({ service, ...progress }) => {
+        if (service === service_name) {
```
jreidinger, line 214: :+1:

web/src/client/software.js
```diff
@@ -355,4 +355,6 @@ class ProductClient {
   }
 }
 
+class ProductClient extends WithIssues(ProductBaseClient, "/issues/product", PRODUCT_PATH) {}
```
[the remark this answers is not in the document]
imobachgs: I would say it is a mistake: it should be `/software/issues/product`. Having said that, the difference (regarding issues) between the "manager" and "products" services is that the former has only one issues list (in `/org/opensuse/Agama/Manager1`) while the latter has two (in `/org/opensuse/Agama/Software1` and `/org/opensuse/Agama/Software1/Product`).

rust/agama-server/src/web.rs
imobachgs: I am not using those IDs for anything, so they aren't relevant. I decided to, somehow, match the D-Bus structure (`/org/opensuse/Agama/Software1` and `/org/opensuse/Agama/Software1/Product`).

However, if we use it as an ID, we could get rid of the `service`, hiding to the client the fact that it is talking to a D-Bus service:

```javascript
WithIssues("software/issues")
```

instead of

```javascript
WithIssues("software/issues", "/org/opensuse/Agama/Software1"
```

In that case, I would adjust the events-related types to something like:

```rust
pub type EventsScope = String;
pub type EventsSender = Sender<(Scope, Event)>;
pub type EventsReceiver = Receiver<(Scope, Event)>;
```

Does it make sense?

imobachgs: On the other hand, keeping the scope out of the event introduces some small inconsistencies, like being able to send a `LocaleChanged` event from any unrelated scope. But I guess I can live with that.

I have been also thinking about using something like:

```rust
pub enum Event {
  Locale(LocaleEvent),
  Software(SoftwareEvent),
};
```

And then each service defines its event types:

```rust
pub enum LocaleEvent {
  LocaleChanged { locale: String },
  IssuesChanged(IssuesChangedEvent),
  ProgressChanged(ProgressChangedEvent)
}
```

In this case, the scope is implicit.

I would like to think a little bit about that (and other options we could think of). For now, I will just update the IDs to make them look better and more consistent.

jreidinger: makes sense to me to think a bit about it and what is more important document decision, so we will be consistent.

web/src/client/software.js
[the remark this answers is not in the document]
jreidinger: ok, then it makes sense to me.

jreidinger (approved):

rust/agama-server/src/web.rs
imobachgs: Just a note about the second proposal: if you can emit a `IssuesChange` in several places from a single service, you still need a way to differentiate them.

PR #1101, opened 2024-03-18
Add mtu for network connections

jcronenberg:
## Problem

No mtu support for network connections

## Solution

Add mtu for network connections

## Testing

- *Modified existing unit tests*
- *Tested manually*

### DIFF
--- a/rust/agama-server/src/network/nm/dbus.rs
+++ b/rust/agama-server/src/network/nm/dbus.rs
@@ -71,17 +71,30 @@ pub fn connection_to_dbus<'a>(
     result.insert("match", match_config_to_dbus(&conn.match_config));
 
     if conn.is_ethernet() {
-        let ethernet_config = HashMap::from([(
-            "assigned-mac-address",
-            Value::new(conn.mac_address.to_string()),
-        )]);
+        let ethernet_config = HashMap::from([
+            (
+                "assigned-mac-address",
+                Value::new(conn.mac_address.to_string()),
+            ),
+            ("mtu", Value::new(conn.mtu)),
+        ]);
         result.insert(ETHERNET_KEY, ethernet_config);
     }
 
     match &conn.config {
         ConnectionConfig::Wireless(wireless) => {
             connection_dbus.insert("type", WIRELESS_KEY.into());
-            let wireless_dbus = wireless_config_to_dbus(wireless, &conn.mac_address);
+            let mut wireless_dbus = wireless_config_to_dbus(wireless);
+            if let Some(wireless_dbus_key) = wireless_dbus.get_mut(WIRELESS_KEY) {
+                wireless_dbus_key.extend(HashMap::from([
+                    ("mtu", Value::new(conn.mtu)),
+                    (
+                        "assigned-mac-address",
+                        Value::new(conn.mac_address.to_string()),
+                    ),
+                ]));
+            }
+
             result.extend(wireless_dbus);
         }
         ConnectionConfig::Bond(bond) => {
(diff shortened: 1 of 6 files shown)


rust/agama-server/src/network/nm/dbus.rs
```diff
@@ -66,17 +66,31 @@ pub fn connection_to_dbus<'a>(
     result.insert("match", match_config_to_dbus(&conn.match_config));
 
     if conn.is_ethernet() {
-        let ethernet_config = HashMap::from([(
-            "assigned-mac-address",
-            Value::new(conn.mac_address.to_string()),
-        )]);
+        let ethernet_config = HashMap::from([
+            (
+                "assigned-mac-address",
```
teclator, line 71: Wondering if we could do this assignation generally and not based in the specific connection type taking into account that in case of empty it could be removed later when the connection is cleaned up just to avoid the code duplication...

teclator:
In general LGTM

rust/agama-server/src/network/nm/dbus.rs
jcronenberg: Maybe something like this?
```rust
    match &conn.config {
        ConnectionConfig::Wireless(wireless) => {
            connection_dbus.insert("type", WIRELESS_KEY.into());
            result.extend(wireless_config_to_dbus(wireless));
        }
        // ...
    }

    let conn_config = HashMap::from([
        (
            "assigned-mac-address",
            Value::new(conn.mac_address.to_string()),
        ),
        ("mtu", Value::new(conn.mtu)),
    ]);
    if conn.is_ethernet() {
        result.insert(ETHERNET_KEY, conn_config);
    } else if let ConnectionConfig::Wireless(_) = conn.config {
        result.get_mut(WIRELESS_KEY).unwrap().extend(conn_config);
    }
```
or like this?
```rust
    match &conn.config {
        ConnectionConfig::Wireless(wireless) => {
            connection_dbus.insert("type", WIRELESS_KEY.into());
            result.extend(wireless_config_to_dbus(wireless));
        }
        // ...
    }

    add_base_conn_values(conn, &mut result);
    // ...
}

fn add_base_conn_values(conn: &Connection, conn_result: &mut NestedHash) {
    let conn_type = if conn.is_ethernet() {
        ETHERNET_KEY
    } else if let ConnectionConfig::Wireless(_) = conn.config {
        WIRELESS_KEY
    } else {
        return;
    };
    let conn_config = HashMap::from([
        (
            "assigned-mac-address",
            Value::new(conn.mac_address.to_string()),
        ),
        ("mtu", Value::new(conn.mtu)),
    ]);
    if let Some(conn_entry) = conn_result.get_mut(conn_type) {
        conn_entry.extend(conn_config);
    } else {
        conn_result.insert(conn_type, conn_config);
    }
}
```
Generally without identifying connection type I don't think is possible, because it needs to be inserted at the `*_KEY` key.
The above solutions at least avoid the duplicate HashMap generation code. I personally don't really have a preference :slightly_smiling_face:

rust/agama-server/src/network/nm/dbus.rs
```diff
@@ -71,17 +71,31 @@ pub fn connection_to_dbus<'a>(
...
     }
 
     match &conn.config {
         ConnectionConfig::Wireless(wireless) => {
             connection_dbus.insert("type", WIRELESS_KEY.into());
-            let wireless_dbus = wireless_config_to_dbus(wireless, &conn.mac_address);
+            let mut wireless_dbus = wireless_config_to_dbus(wireless);
+            wireless_dbus
+                .get_mut(WIRELESS_KEY)
+                .unwrap()
```
imobachgs, line 90: If possible, I would avoid unwrapping (and potentially crashing). You can use a default value or ignore it completely.

imobachgs (commented):
In general it looks good. Just a minor comment about a call to `unwrap`.

rust/agama-server/src/network/nm/dbus.rs
jcronenberg: I mean currently this is impossible, but for future-proofing, sure.

imobachgs (approved):

PR #1102, opened 2024-03-18
Merge `master` into the `architecture_2024` branch

imobachgs:
Keep the `architecture_2024` in sync with the `master` branch.


mchf:
LGTM (but the React part ... but that is my personal problem with React ;-))

PR #1103, opened 2024-03-19
Adapt to the HTTP-based Progress API

imobachgs:
Adapt the `WithProgress` to the HTTP-based implementation of the D-Bus progress interface (#1092).

### DIFF
--- a/web/src/client/mixins.js
+++ b/web/src/client/mixins.js
@@ -192,12 +192,14 @@ const WithStatus = (superclass, status_path, service_name) =>
  */
 
 /**
- * Extends the given class with methods to get and track the progress over D-Bus
- * @param {string} object_path - object_path
+ * Extends the given class with methods to get and track the service progress
+ *
+ * @template {!WithHTTPClient} T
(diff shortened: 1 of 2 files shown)


web/src/client/mixins.js
```diff
@@ -222,13 +226,16 @@ const WithProgress = (superclass, object_path) =>
      * @return {import ("./dbus").RemoveFn} function to disable the callback
      */
     onProgressChange(handler) {
-      return this.client.onObjectChanged(object_path, PROGRESS_IFACE, (changes) => {
-        const { TotalSteps, CurrentStep, Finished } = changes;
-        if (TotalSteps === undefined && CurrentStep === undefined && Finished === undefined) {
-          return;
+      return this.client.onEvent("Progress", (progress) => {
+        if (progress?.service === service_name) {
```
jreidinger, line 230: is it possible to get undefined service? Is it worth logging it?

imobachgs: It should not happen, but I wanted to avoid a crash as we cannot control what comes through the wire.

PR #1105, opened 2024-03-19
WIP: RFC

ancorgs:
Draft PR to discuss some aspects.

I modeled this as a new class `VolumeLocation` with two attributes:

- device: the name of the target device, the exact meaning depends on the other attribute
- target: which specifies where and how to allocate the volume
  - default: nothing special, new partition in the installation disk or new LV in the system VG. `#device` is ignored.
  - new partition: new partition at a specific disk indicated by `#device`.
  - new VG: new LV in a new dedicated VG created at a specific disk indicated by `#device`.
  - device: the existing block device specified by `device` is used and reformatted.
  - filesystem: the existing filesystem on the existing device specified by `device` is mounted

The new class looks like this, which is very similar to `SpaceSettings` and its policies:

```ruby
class VolumeLocation
  TARGETS = [:default, :new_partition, :new_vg, :device, :filesystem].freeze
  private_constant :TARGETS
  
  # @return [Symbol] see .targets
  attr_accessor :target
  
  # @return [String, nil]
  attr_accessor :device 
  
  # All possible values for #target
  #
  # @return [Array<Symbol>]
  def self.targets
    TARGETS
  end 
    
  def initialize
    @target = :default
  end
end 
```

But I don't feel 100% comfortable using just symbols there instead of something closer to an ENUM that makes harder to mistype the value.

On the other hand, I have a question regarding conversion from/to DBus:


service/lib/agama/storage/volume_location.rb
```diff
@@ -0,0 +1,78 @@
...
+  module Storage
+    # Settings specifying what device should be used for a given Volume and how
+    class VolumeLocation
+      # @see .targets
+      TARGETS = [:default, :new_partition, :new_vg, :device, :filesystem].freeze
+      private_constant :TARGETS
+
+      # What to do to allocate the volume
+      #
+      # @return [Symbol] see .targets
```
joseivanlopez, line 32: I think it needs brackets, see {.targets}

joseivanlopez (commented):
Looks good, but I miss a changelog entry. The only bad thing is that I will eat a lot of conflicts with https://github.com/openSUSE/agama/pull/1068.

doc/dbus/org.opensuse.Agama.Storage1.Proposal.Calculator.doc.xml
```diff
@@ -18,8 +18,8 @@
       <!--
         MountPath               s
         MountOptions            as
-        TargetDevice            s
-        TargetVG                s
+        Target                  s
+        TargetDevice            s (only makes sense if Target is not default)
```
joseivanlopez, line 22: The equivalent for the default device are called `Target` and `Device`. I am not sure if we should follow the same naming or not.

ancorgs: I like the current names because it shows the relationship between both fields. I mean, `TargetDevice` is actually some kind of "details" field for the `Target` one.

joseivanlopez: Ok, makes sense. Then, I will adapt my PR to follow the same approach.

joseivanlopez (approved):
LGTM

PR #1106, opened 2024-03-19
New device and boot selection dialogs

dgdavid:
This PR implements the new UI selectors for the target installation device and for the boot device, see  https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md.

For the installation device it offers the following options:

* Select a target disk (the file systems are created over new partitions).
* Create a new LVM volume group (the file systems are crated over logical volumes).
* The option for reusing LVM volume groups is not offered yet.

For the boot device it offers the following options:

* To use the selected disk for installation.
* To use a different disk.
* To not configure boot partitions.

## Screenshots

<details>
<summary>Toggle</summary>

localhost_8080_ (33)

localhost_8080_ (32)

</details>


web/src/components/core/ExpandableSelector.jsx
```diff
@@ -0,0 +1,240 @@
...
+ * @param {object[]} props.items - Collection of items to be rendered.
+ * @param {string} [props.itemIdKey="id"] - The key for retrieving the item id.
+ * @param {(item: object) => Array<object>} [props.itemChildren=() => []] - Lookup method to retrieve children from given item.
+ * @param {(item: object) => boolean} [props.itemSelectable=() => true] - Whether an item will be selectable or not.
+ * @param {(item: object) => (string|undefined)} [props.itemClassNames=() => ""] - Callback that allows adding additional CSS class names to item row.
+ * @param {object[]} [props.itemsSelected=[]] - Collection of selected items.
+ * @param {string[]} [props.initialExpandedKeys=[]] - Ids of initially expanded items.
+ * @param {(selection: Array<object>) => void} [props.onSelectionChange=noop] - Callback to be triggered when selection changes.
+ * @param {object} [props.tableProps] - Props for {@link https://www.patternfly.org/components/table/#table PF/Table}.
+ */
```
dgdavid, line 113: Add a note why this cannot be build with core/TreeTable.

Try to unify them as much as possible (after current milestone)

dgdavid:
Since this afternoon I had problems with Proxy 

```
dbus.js:103 Could not create a proxy for org.opensuse.Agama.Storage1.Proposal undefined error= 
[]
```

I cheated by importing fake devices located at _components/storage/test-data/..._.  This lead me to have devices with "Transport Unknown".

Screenshot from 2024-03-25 18-47-08

I guess we could simply show nothings if we do not have meaning information for such an _attribute_. Also, as discussed offline, the `Transport` word is not necessary when we know the transport. It just a matter of data and its representation that we can polish little by little, w/o blocking the main goal of the PR IMHO.

dgdavid:
@ancorgs, see https://github.com/openSUSE/agama/pull/1106#issuecomment-2018751811

joseivanlopez:
> Since this afternoon I had problems with Proxy
> 
> ```
> dbus.js:103 Could not create a proxy for org.opensuse.Agama.Storage1.Proposal undefined error= 
> []
> ```

Did you check whether a proposal exists in D-Bus? I think that error is expected if the installer is not in config phase. Maybe you reached the storage page too early? Did you select a product? Just guessing.

web/src/components/storage/BootSelectionDialog.jsx
```diff
@@ -0,0 +1,168 @@
...
+      <Form id="boot-form" onSubmit={onSubmit}>
+        <fieldset className="stack">
+          <legend className="split">
+            <RadioOption id={BOOT_AUTO_ID} defaultChecked={isBootAuto} onChange={() => selectBootAuto()}>
+              {_("Automatic")}
+            </RadioOption>
+          </legend>
+          <div>
+            {_("Additional partitions to boot the system will be configured at /dev/vdc, \
+            the device selected for installing the system.")}
```
dgdavid, line 125: Actually, the idea here is to go for a `%s` and then replace it by the selected _installationDevice_

```suggestion
            {_("Additional partitions to boot the system will be configured in \
            the device selected for installing the system, currently %s")}
```

But I'm not sure how to make it work. I mean, when _target_ is disk is pretty clear, just to put the _device.name_ or _deviceLabel_ there. But what is the right value when the _target_ is a new LVM thing (see https://github.com/openSUSE/agama/pull/1106/files#diff-bd6c15ca365f9adcb6bb3d76300f7729cfbd21274b895e1504c5956443e8a99cR53-R61)

Let's talk about it on Monday.

ancorgs (approved):
Content-wise it LGTM, but I guess we need to squash those 49 WIP commits(!) into something more reasonable.

PR #1107, opened 2024-03-19
web: update dependencies (2024-03-19)

dgdavid:
Update web dependencies to their latest versions via `npm update` and `npm install package-name@latest` when needed.

---

Previous update: https://github.com/openSUSE/agama/pull/985


imobachgs (commented):
I am not sure whether it is worth it. It introduces some conflicts with the architecture_2024 branch and, after all, we should be merging them in one month.

I would only patch important security issues.

dgdavid:
> I am not sure whether it is worth it. It introduces some conflicts with the architecture_2024 branch and, after all, we should be merging them in one month.
> 
> I would only patch important security issues.

I do not see conflicts in just to files a big problem. I can take care of them.

That said, I can live postponing the update until the mentioned branch is merged. But please, keep the PR open to not forget about it.

imobachgs:
> I do not see conflicts in just two files a big problem. I can take care of them.
> 
> That said, I can live postponing the update until the mentioned branch is merged. But please, keep the PR open to not forget about it.

I like to avoid lockfiles merges if possible :-) I am not against this PR, but I do not see the benefit. I would rather prefer to update the dependencies in the `architecture_2024` branch instead.

imobachgs (approved):

dgdavid:
> 
> I like to avoid lockfiles merges if possible :-) I am not against this PR, but I do not see the benefit. I would rather prefer to update the dependencies in the `architecture_2024` branch instead.

Ok, let's wait

dgdavid:
Postponing the update until `architecture_2024` branch is ready. With a bit of luck we are not going to forget it.

PR #1109, opened 2024-03-20
[web] Fix inconsistencies in storage volumes

### LINKED ISSUE #1108: Inconsistencies in the volumes
Before merging #1104 

not-broken

After merging that pull request

not-not-broken

a) It's not longer possible to add a `/home` volume although it's in the TW configuration
b) The Btrfs switcher is not in sync with the table of volumes

ancorgs:
This fixes #1108, both the problems and the implemented solutions are described there in depth.

Tested manually to check both problems are indeed fixed.

### DIFF
--- a/web/src/components/storage/ProposalPage.jsx
+++ b/web/src/components/storage/ProposalPage.jsx
@@ -232,6 +232,7 @@ export default function ProposalPage() {
       <ProposalSettingsSection
         availableDevices={state.availableDevices}
         encryptionMethods={state.encryptionMethods}
+        volumeTemplates={state.volumeTemplates}
         settings={state.settings}
         onChange={changeSettings}
         isLoading={state.loading}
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
LGTM

web/src/components/storage/ProposalPage.jsx
```diff
@@ -232,6 +232,7 @@ export default function ProposalPage() {
       <ProposalSettingsSection
         availableDevices={state.availableDevices}
         encryptionMethods={state.encryptionMethods}
+        volumeTemplates={state.volumeTemplates}
```
dgdavid, line 235: In a talk with @imobachgs, we concluded that this kind of errors could be avoid with the type checking. To do so, the  `volumeTemplates` prop should be required and not have a default value. The consumer is the responsible of giving the empty array if it's the case.

dgdavid: Just a comment for bearing it in mind when activating the ts-check for the whole .jsx files.

PR #1110, opened 2024-03-21
Skip the IRC notification if the integration test is started manually

lslezak:
## Problem

- When debugging a failing integration test you have to trigger it manually
- If the test fails it sends notification to IRC
- If you need to run the test several times to find the problem then you are spamming the IRC channel

## Solution

- Skip the IRC notification if the integration test is started manually

## Rationale

- If you run the test manually you are very likely watching the result, it should send the IRC notification only when started automatically via the cron setting.


mvidner (approved):
Makes sense, LGTM

PR #1116, opened 2024-03-25
Adapt the Network UI to the HTTP/JSON API

teclator:
## Problem

Recently, we have been adding an HTTP/JSON layer on top of the network configuration. Using this new API we should be able to:

Display the active connections in the [NetworkSection](https://github.com/openSUSE/agama/blob/master/web/src/components/overview/NetworkSection.jsx) and the [ConnectionsTable](https://github.com/openSUSE/agama/blob/master/web/src/components/network/ConnectionsTable.jsx).
Allow editing the configuration of those connections.

- [*Trello Card*](https://trello.com/c/chdpfc4S/3619-8-basic-network-configuration-handling)

## Solution

The UI has been adapted using the HTTP/JSON API and additing or updating a connections using it is now allowed. Although it was out of the scope, adding a Wireless connection is also allowed but in that case we are not listening to the DBUS events and not reacting properly, therefore, a refresh would be needed in order to see the added connection.


rust/agama-lib/src/network/types.rs
```diff
@@ -55,6 +55,43 @@ pub enum DeviceType {
     Bridge = 6,
 }
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
imobachgs, line 59: We should use `camelCase`. It won't make a difference in this case, but for consistency...

rust/agama-server/src/network/model.rs
```diff
@@ -559,15 +568,17 @@ impl TryFrom<Connection> for NetworkConnection {
         let mac = conn.mac_address.to_string();
         let method4 = Some(conn.ip_config.method4.to_string());
         let method6 = Some(conn.ip_config.method6.to_string());
-        let mac_address = (!mac.is_empty()).then(|| mac);
-        let nameservers = conn.ip_config.nameservers.into();
-        let addresses = conn.ip_config.addresses.into();
-        let gateway4 = conn.ip_config.gateway4.into();
-        let gateway6 = conn.ip_config.gateway6.into();
-        let interface = conn.interface.into();
+        let mac_address = (!mac.is_empty()).then_some(mac);
```
imobachgs, line 571: :+1:

rust/agama-server/src/network/nm/client.rs
```diff
@@ -231,6 +243,8 @@ impl<'a> NetworkManagerClient<'a> {
         if conn.is_up() {
             self.activate_connection(path).await?;
         } else {
+            //FIXME: This is wrong, deactivate receive an active connection path and this is a
```
imobachgs, line 246: Is this comment still a problem?

imobachgs (commented):
Just a partial review; I will continue in a while.

rust/agama-server/src/network/nm/client.rs
```diff
@@ -288,16 +302,24 @@ impl<'a> NetworkManagerClient<'a> {
...
         let proxy = NetworkManagerProxy::new(&self.connection).await?;
-        match proxy.deactivate_connection(&path.as_ref()).await {
-            Err(e) => {
-                // Ignore ConnectionNotActive error since this just means the state is already correct
-                if e.to_string().contains("ConnectionNotActive") {
-                    Ok(())
-                } else {
-                    Err(ServiceError::DBus(e))
+
+        if let Some(active_connection) = self.settings_active_connection(path.clone()).await? {
```
imobachgs, line 306: What about something like:

```rust
        if let Some(active_connection) = self.settings_active_connection(path.clone()).await? {
            if let Err(e) = proxy
                .deactivate_connection(&active_connection.as_ref())
                .await
            {
                // Ignore ConnectionNotActive error since this just means the state is already correct
                if !e.to_string().contains("ConnectionNotActive") {
                    return Err(ServiceError::DBus(e));
                }
            }
        }
        Ok(())
```

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
+    state
+        .actions
+        .send(Action::GetConnection(id.clone(), tx))
+        .unwrap();
+    let current_conn = rx.await.unwrap();
     let mut conn = Connection::try_from(conn)?;
-    conn.uuid = id;
+    match current_conn {
+        Some(current) => {
+            if current.id != id.clone() {
```
imobachgs, line 235: No need to clone.

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
-    Path(id): Path<Uuid>,
+    Path(id): Path<String>,
     Json(conn): Json<NetworkConnection>,
 ) -> Result<Json<()>, NetworkError> {
+    let (tx, rx) = oneshot::channel();
+    state
+        .actions
+        .send(Action::GetConnection(id.clone(), tx))
+        .unwrap();
+    let current_conn = rx.await.unwrap();
```
imobachgs, line 231: What about using `old_conn` or `orig_conn`? Or use `conn` here and rename `conn` to `updated_conn`.

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
     let mut conn = Connection::try_from(conn)?;
-    conn.uuid = id;
+    match current_conn {
```
imobachgs, line 233: What about something like:

```rust
    let current = current_conn.ok_or_else(|| NetworkError::UnknownConnection(id.clone()))?;
    if current.id != id {
        return Err(NetworkError::UnknownConnection(id));
    } else {
        conn.uuid = current.uuid;
    }
```

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
+    Path(id): Path<String>,
+) -> Result<Json<()>, NetworkError> {
+    let (tx, rx) = oneshot::channel();
+    state
+        .actions
+        .send(Action::GetConnection(id.clone(), tx))
+        .unwrap();
+
+    if let Some(mut current_conn) = rx.await.unwrap() {
+        current_conn.set_up();
```
imobachgs, line 267: Would not be better to put this logic as part of the action? IMHO, as much as possible, the handler should be responsible of parsing/validating the request/response and call the right method (or dispatch the right action).

However, if we have followed that approach elsewhere, we can keep it as it is and reevaluate in the short term (when we drop the D-Bus code and clean-up the actions).

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
+    State(state): State<NetworkState>,
+    Path(id): Path<String>,
+) -> Result<Json<()>, NetworkError> {
+    let (tx, rx) = oneshot::channel();
+    state
+        .actions
+        .send(Action::GetConnection(id.clone(), tx))
+        .unwrap();
+
+    if let Some(mut current_conn) = rx.await.unwrap() {
```
imobachgs, line 266: You could return early like this (not tried, there might be some syntax error):

```rust
let Some(mut current_conn) = rx.await.unwrap() {
  return Err(NetworkError::UnknownConnection(id);
}

/// the rest goes here
```

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
+    Path(id): Path<String>,
+) -> Result<Json<()>, NetworkError> {
+    let (tx, rx) = oneshot::channel();
     state
         .actions
-        .send(Action::UpdateConnection(Box::new(conn)))
+        .send(Action::GetConnection(id.clone(), tx))
         .unwrap();
 
+    if let Some(mut current_conn) = rx.await.unwrap() {
```
imobachgs, line 298: Early return, in the same way than the `connect()` function.

web/src/client/network.test.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2022] SUSE LLC
+ * Copyright (c) [2024] SUSE LLC
```
imobachgs, line 2:
```suggestion
 * Copyright (c) [2022-2024] SUSE LLC
```

imobachgs (commented):
Already reviewed the Rust code. Let's go with the JavaScript part.

rust/agama-server/src/network/web.rs
```diff
@@ -167,67 +175,142 @@ async fn connections(State(state): State<NetworkState>) -> Json<Vec<NetworkConne
...
         .send(Action::RemoveConnection(id, tx))
         .unwrap();
 
-    let _ = rx.await.unwrap();
-    let (tx, rx) = oneshot::channel();
-    state.actions.send(Action::Apply(tx)).unwrap();
-    let _ = rx.await.unwrap();
-
-    Ok(Json(()))
+    Ok(Json(rx.await.unwrap()?))
```
imobachgs, line 215: It should return a StatusCode::NO_CONTENT (204).

web/src/client/http.js
```diff
@@ -150,6 +150,27 @@ class HTTPClient {
...
+  async delete(url) {
+    const response = await fetch(`${this.baseUrl}/${url}`, {
+      method: "DELETE",
+      headers: {
+        "Content-Type": "application/json",
+      },
+    });
+
+    try {
+      return await response.json();
```
imobachgs, line 167: The answer from the backend is empty, so we should the status code 204 (no content). In that case, there is not need to parse the response for a JSON body.

imobachgs (commented):
Finished with reviewing the JavaScript code too.

rust/agama-server/src/network/nm/client.rs
teclator: no, a leftover ;)

imobachgs (approved):

PR #1117, opened 2024-03-26
Adapt the users service to the HTTP/JSON API

jreidinger:
## Problem

Users section in architecture_2024 is disabled

## Solution

Readd it using http API.

## Testing

- *Tested manually*

Only issue found during testing is that some put/delete request does not respond body in this PR and js part complain as it always expect json.

## Screenshots

new_users_overview
new_users


rust/agama-server/src/users/web.rs
```diff
@@ -0,0 +1,253 @@
...
+    const DBUS_PATH: &str = "/org/opensuse/Agama/Users1";
+
+    let users = UsersClient::new(dbus.clone()).await?;
+    let state = UsersState { users };
+    let validation_router = validation_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
+    let status_router = service_status_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
+    let router = Router::new()
+        .route("/config", get(get_info))
+        .route("/user", put(set_first_user).delete(remove_first_user))
+        .route(
```
imobachgs, line 134: I think that having a single `/root` resource would be more straightforward (and extensible). Let's say that, tomorrow, we want to add another attribute. 

We could have the following routes (just an idea):

* GET `/root` to get the full user.
* PUT (or PATCH) `/root` to update any of the attributes (just specify the one that changes).
* DELETE `/root/password` to delete the password.
* DELETE `/root/sshkey` to delete the SSH key.

Alternatively, we could use `PUT` to replace the whole resource and `PATCH` to update only one of the elements.

rust/agama-server/src/users/web.rs
```diff
@@ -0,0 +1,253 @@
...
+pub struct UsersInfo {
+    user: Option<FirstUser>,
+    root: RootInfo,
+}
+
+#[utoipa::path(put, path = "/users/config", responses(
+    (status = 200, description = "Configuration for users including root and the first user", body = UsersInfo),
+    (status = 400, description = "The D-Bus service could not perform the action"),
+))]
+async fn get_info(State(state): State<UsersState<'_>>) -> Result<Json<UsersInfo>, Error> {
```
imobachgs, line 237: Here is where the API looks confusing to me. If I need to update the user, I do call `PUT` on the `/users/user` route. But, if I want to get the current user, I need to get the info from `/users/config`.

rust/agama-server/src/web/common.rs
```diff
@@ -352,3 +352,108 @@ async fn build_issues_proxy<'a>(
...
+    };
+    Ok(Json(validation))
+}
+
+#[derive(Clone)]
+struct ValidationState<'a> {
+    proxy: ValidationProxy<'a>,
+}
+
+/// Builds a stream of the changes in the the `org.opensuse.Agama1.Issues`
```
imobachgs, line 418:
```suggestion
/// Builds a stream of the changes in the the `org.opensuse.Agama1.Validation`
```

rust/agama-server/src/web/event.rs
```diff
@@ -21,6 +23,13 @@ pub enum Event {
     ProductChanged {
         id: String,
     },
+    FirstUserChanged(FirstUser),
+    RootPasswordChanged {
```
imobachgs, line 27: Why not merge both root actions?

```rust
RootUserChanged: {
  sshkey: Option<String>,
  has_password: Option<bool>
}
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
...
 
+Example of running from different machine:
+
+```
+  # backend machine
+  # using ip of machine instead of localhost is important to be network accessible
+  agama-web-server serve --address 10.100.1.1:3030
+
+  # frontend machine
+  # ESLINT=0 is useful during development when there is some styling issues
```
imobachgs, line 46:
```suggestion
  # ESLINT=0 is useful to ignore linter problems during development
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
...
+  ESLINT=0 AGAMA_SERVER=10.100.1.1:3000 npm run server
+```
+
+### Debugging Hints
+
+There are more locations where to look when something does not work and require debugging.
```
imobachgs, line 52:
```suggestion
There are several places to look when something does not work and requires debugging.
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
...
+The first location to check when something does not work is browser console which can give
```
imobachgs, line 53:
```suggestion
The first place is the browser's console which can give
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
...
+some hints. The second location to check for errors or warnings is output of `npm run server`
+where some issues when communicating with backend can be seen. And last but on least is
```
imobachgs, line 55:
```suggestion
where you can find issues when communicating with the backend. And last but on least is
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
...
+journal on backend machine where is logged backend activity `journalctl -b`.
```
imobachgs, line 56: Thanks a lot for documenting this. I would add: `busctl monitor --address /run/agama/bus` as another source of information.

web/src/client/http.js
```diff
@@ -150,6 +150,23 @@ class HTTPClient {
     }
   }
 
+  /**
+   * @param {string} url - Endpoint URL (e.g., "/l10n/config").
+   * @return {Promise<object>} Server response.
+   */
+  async delete(url) {
```
imobachgs, line 157: :+1:

web/src/client/mixins.js
```diff
@@ -259,12 +255,12 @@ const WithValidation = (superclass, object_path) => class extends superclass {
     let errors;
```
imobachgs, line 255: What about renaming `errors` to something like `response`? Basically to avoid the `errors.errors.map` below.

imobachgs (changes requested):
Implementation-wise, it looks good. However, I am not convinced with the API design. But I am not a big fan of users D-Bus API, so it is probably just a consequence.

rust/agama-server/src/users/web.rs
```diff
@@ -0,0 +1,253 @@
...
+struct UsersState<'a> {
+    users: UsersClient<'a>,
+}
+
+/// Returns streams that emits users related events coming from D-Bus.
+///
+/// It emits the Event::RootPasswordChange, Event::RootSSHKeyChanged and Event::FirstUserChanged events.
+///
+/// * `connection`: D-Bus connection to listen for events.
+pub async fn users_streams(
```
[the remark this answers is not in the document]
jreidinger: yeap, it would be consistent

rust/agama-server/src/users/web.rs
jreidinger: yeap, that is something I mention when I speak about need to have some guide to have consistent API. Both ways is possible. When talking about single resource I even think about having `/users/config` and methods get and patch. ( put is problematic as root password get is boolean, but patch needs value and encryption ).

rust/agama-server/src/users/web.rs
jreidinger: yeap, that is why I mention that we can have that `/users/config` and patch when need to change it. Question is how much it will be consistent with other APIs we have.

rust/agama-server/src/web/event.rs
jreidinger: well, we can merge even all of them and have UsersChanged and have complete get content there. Or as in question simple UsersChanged and force user to reread values with get. This way it just follows dbus signals and use value it provides.

web/README.md
jreidinger: well, I think that this busctl monitor and other useful tools for dbus observing belongs more to backend documentation. Either service or rust dir. Here I try to focus mostly on debugging issues with web UI.

rust/agama-server/src/users/web.rs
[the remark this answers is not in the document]
jreidinger: btw as all our streams return this quite complex result I think it makes sense to create type in crate::web for easier sharing.

jreidinger:
@imobachgs regarding API I think it would be useful to have call tomorrow. I check current http API we have to see if it is consistent or not.

So far it looks quite good and consistent ( sadly users has some specific methods especially for that root password which makes get/put with different data which looks bad ).
Inconsistent API I found:

- in general returning empy json replies in methods that do not require to have response looks kind of strange.
- network/system/apply uses `put` instead of `post` which I would expect
- questions use for answer path `put` and probably `post` will fit better ( yes, I know I am author :)

web/README.md
imobachgs: Yes, sure, but given that you mentioned `journalctl` too, I think it might be useful.

jreidinger: well, I mention journalctl mainly because you see there if request from web UI lands to backend. If it is correctly send there, then I would continue with backend debugging. But maybe we can link backend debugging document from this one.

jreidinger:
Just do document result of call:
http API will look like:

```
/users/root -> get/patch method, where patch allows to set key, password or unset it
/users/initial -> get/put/delete method where put will put whole config for user
```

and events:
```
split to two events one when initial user change and one when root user config change
```

rust/agama-server/src/users/web.rs
```diff
@@ -0,0 +1,229 @@
...
+    dbus: zbus::Connection,
+) -> Result<Vec<(&'static str, Pin<Box<dyn Stream<Item = Event> + Send>>)>, Error> {
+    const FIRST_USER_ID: &str = "first_user";
+    const ROOT_PASSWORD_ID: &str = "root_password";
+    const ROOT_SSHKEY_ID: &str = "root_sshkey";
+    // here we have three streams, but only two events. Reason is
+    // that we have three streams from dbus about property change
+    // and munify two root user properties into single event to http API
```
imobachgs, line 41: I am not sure whether we need a long explanation. Anyway, let's fix the typo.

```suggestion
    // and unify two root user properties into single event to http API
```

rust/agama-server/src/users/web.rs
```diff
@@ -0,0 +1,229 @@
...
+    (status = 200, description = "Removes the first user"),
+    (status = 400, description = "The D-Bus service could not perform the action"),
+))]
+async fn remove_first_user(State(state): State<UsersState<'_>>) -> Result<(), Error> {
+    state.users.remove_first_user().await?;
+    Ok(())
+}
+
+#[utoipa::path(put, path = "/users/first", responses(
+    (status = 200, description = "User values are set"),
```
imobachgs, line 158:
```suggestion
    (status = 200, description = "Sets the first user"),
```

web/README.md
```diff
@@ -35,6 +35,26 @@ running Agama server instance. This is especially useful if you use the Live ISO
 which does not contain any development tools, you can develop the web frontend
 easily from your workstation.
 
+Example of running from different machine:
+
+```
+  # backend machine
+  # using ip of machine instead of localhost is important to be network accessible
+  agama-web-server serve --address 10.100.1.1:3030
```
imobachgs, line 43:
```suggestion
  agama-web-server serve --address 10.100.1.1:3000
```

web/README.md
imobachgs: Well, I would avoid having different documents about debugging. All the information about that matter should be in a single place.

web/src/client/http.js
```diff
@@ -142,27 +142,36 @@ class HTTPClient {
...
+  /**
+   * @param {string} url - Endpoint URL (e.g., "/l10n/config").
+   * @return {Promise<Response>} Server response.
+   */
+  async delete(url) {
+    const response = await fetch(`${this.baseUrl}/${url}`, {
+      method: "DELETE",
+    });
+
+    return response;
```
imobachgs, line 157: I am not sure whether we should return the at all. At some point we could wrap the standard responses objects with our own information.

In any case, if we want to, we should do that in all methods to be consistent.

web/src/client/users.js
```diff
@@ -165,16 +152,21 @@ class UsersBaseClient {
    * @return {import ("./dbus").RemoveFn} function to disable the callback
    */
   onUsersChange(handler) {
-    return this.client.onObjectChanged(USERS_PATH, USERS_IFACE, changes => {
-      if (changes.RootPasswordSet) {
+    return this.client.ws.onEvent((event) => {
+      if (event.type === "RootChanged") {
+        const res = {};
```
imobachgs, line 157: What `res` means?

imobachgs (commented):
In general it looks good. Just a question: are `UsersSection` and `UsersPage` tests working again?

web/README.md
jreidinger: and that reminds me that I should update that documentation now with ssl enabled

web/README.md
jreidinger: ok, so lets add it.

web/src/client/http.js
jreidinger: well, response is important if we want to check if call succeed or failed ( method `ok` on response or if response is added as it is optional for non-get calls). What I do not like even on current `get` is that it returns `true/false` if response is not json, which is quite bad API. So providing whole response object is more clear for me.

web/src/client/users.js
jreidinger: well, I use it as result of data in event. Maybe I should use to something like `changed_data`?

jreidinger:
Good question about tests. It is not enabled, so lets do it and do proper mocking in it

imobachgs (approved):

web/src/client/users.js
imobachgs: or just "changes",

PR #1119, opened 2024-03-26
Adapt agama-live to work with the new architecture

imobachgs:
## Changes summary

* Adds a `--generate-token` option to `agama-web-server`. Using such a token you can skip the authentication (given that you have access to the token).
* Adds a new Firefox start-up script that injects the token, skipping the authentication when running locally.
* Defines an `agama-web-server` systemd service.
* Temporarily disables the `no-unused-vars` linter.
* Fixes `agama-web-ui.spec`.
* Adjusts the KIWI definition.

## To do

- [x] Replace the sqlite3 hack with support in the backend.
- [x] Use standard HTTP and HTTPS ports.
- [x] Add uncompressed version of the files to the package.

## agama-live sources in the repo

In the future, we plan to move the source we use to build `agama-live` to this repository.


live/startup
```diff
@@ -0,0 +1,39 @@
...
+    CONSTRAINT moz_uniqueid UNIQUE (name, host, path, originAttributes));"
+else
+  sqlite3 $COOKIES_DB "DELETE FROM moz_cookies WHERE name = 'agamaToken';"
+fi
+
+NOW=$(date +%s)
+VALUE=$(cat $TOKEN_FILE)
+EXPIRATION=$(date --date '1 day' +%s)
+
+sqlite3 $COOKIES_DB "INSERT INTO moz_cookies \
```
lslezak, line 34: I do not like hacking the internal Firefox cookies database directly. That's very fragile and works only with Firefox.

What about creating a `/run/agama/token.html` file with root-only permissions with content like 

```html
<!DOCTYPE html>
<html>
<body>
  <script>
    document.location = "http://localhost/login?token=" + <JS-escaped-token-value>;
  </script>
</body>
</html>
```

and implement `/login` endpoint which would set the `agamaToken` cookie and redirect to `/`.

And then use that file in the `firefox` command below like

```sh
firefox --kiosk --profile $HOME/.mozilla/firefox/profile /run/agama/token.html
```

That special file would avoid exposing the token in the process list (in contrast when using the token directly in the firefox command).

The advantage would not hacking the SQLite internals and would be generic for all browsers.

Note: Originally I wanted to put something like 
```js
 document.cookie = "agamaToken=" + tokenValue + ";domain=localhost;path=/api";
```
into that HTML file. But for security reasons it is not possible to set cookie for a different domain...

imobachgs: Of course, I am open to other approaches and I see your point. However bear in mind that we only need to support Firefox and, if something changes, we can detect soon enough in our CI.

To be honest, I would rather prefer some kind of API to import the cookie and do not have to extend the backend (and add a redirection). But I am not against this solution.

rust/agama-server/src/web/service.rs
```diff
@@ -84,7 +84,7 @@ impl MainServiceBuilder {
             .route("/ping", get(super::http::ping))
             .route("/auth", post(login).get(session).delete(logout));
 
-        let serve = ServeDir::new(self.public_dir);
+        let serve = ServeDir::new(self.public_dir).precompressed_gzip();
```
lslezak, line 87: > Both the precompressed version and the uncompressed version are expected to be present in the directory.

https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html#method.precompressed_gzip

That means we will later probably need to update the packaging of the web frontend files to contain both files.

imobachgs: Yes, exactly.

live/startup
imobachgs: Thinking about your proposal, given that we need to extend the backend anyway, we can get rid of the `token.html` file by just setting the cookie *and* redirecting the user. The following code seems to work:

```rust
pub async fn get_cookie(
    State(state): State<ServiceState>,
    Query(params): Query<GetCookieParams>,
) -> impl IntoResponse {
    let decoding = DecodingKey::from_secret(state.config.jwt_secret.as_ref());
    let mut headers = HeaderMap::new();

    if jsonwebtoken::decode::<TokenClaims>(&params.token, &decoding, &Validation::default()).is_ok()
    {
        let cookie = format!("agamaToken={}; HttpOnly", params.token);
        headers.insert(
            SET_COOKIE,
            cookie.parse().expect("could not build a valid cookie"),
        );
    }

    headers.insert(LOCATION, "/".parse().unwrap());
    (StatusCode::PERMANENT_REDIRECT, headers)
}
```

If it is OK for you, I will clean-up the code a bit and update the PR.

imobachgs:
I have update the PR to follow @lslezak's approach. About automating this process, we already have a card for it: https://trello.com/c/1G6TeRzy/337-agama-put-agama-live-sources-in-a-github-repository. Perhaps it is something to plan for in the following sprint.

rust/agama-server/src/web/http.rs
```diff
@@ -69,6 +72,33 @@ pub async fn login(
...
+
+    if TokenClaims::from_token(&params.token, &state.config.jwt_secret).is_ok() {
+        let cookie = format!("agamaToken={}; HttpOnly", params.token);
+        headers.insert(
+            SET_COOKIE,
+            cookie.parse().expect("could not build a valid cookie"),
+        );
+    }
+
+    headers.insert(LOCATION, "/".parse().unwrap());
```
lslezak, line 98: You can use `HeaderValue::from_static("/")` which is IMHO better...

rust/agama-server/src/web/http.rs
```diff
@@ -69,6 +72,33 @@ pub async fn login(
...
+    (status = 301, description = "Injects the authentication cookie if correct and redirects to the web UI")
+))]
+pub async fn login_from_query(
+    State(state): State<ServiceState>,
+    Query(params): Query<LoginFromQueryParams>,
+) -> impl IntoResponse {
+    let mut headers = HeaderMap::new();
+
+    if TokenClaims::from_token(&params.token, &state.config.jwt_secret).is_ok() {
+        let cookie = format!("agamaToken={}; HttpOnly", params.token);
```
lslezak, line 91: There is no expiration set, how long is the cookie valid by default?

rust/agama-server/src/web/http.rs
```diff
@@ -69,6 +72,33 @@ pub async fn login(
...
+        headers.insert(
+            SET_COOKIE,
+            cookie.parse().expect("could not build a valid cookie"),
+        );
+    }
+
+    headers.insert(LOCATION, "/".parse().unwrap());
+    (StatusCode::PERMANENT_REDIRECT, headers)
```
lslezak, line 99: Permanent redirect means that the browser should remember that and the future requests should directly go to the redirected location skipping the original request.

I'd rather use `StatusCode::TEMPORARY_REDIRECT` here because the redirection has a side effect of setting the cookie. If the browser is restarted or reloaded for whatever reason later with the `/login` path it could cause some nasty surprise...

lslezak (commented):
Just some minor questions and suggestions...

rust/agama-server/src/web/http.rs
imobachgs: Not having an expiration date makes it a "session cookie". It should be gone once you close the browser. You can find more information in the [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#expiresdate) docs.

lslezak: Ah, when there is no expiration then the cookie should be valid for the current session (until the browser is closed). That's IMHO a correct behavior in our use case.

lslezak: Probably worth adding a comment into the code...

rust/agama-server/src/web/http.rs
imobachgs: Good point!

rust/agama-server/src/web/http.rs
imobachgs: Oh, I did not know about that option. Much better, yes. Thanks!

live/startup
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+# Start a browser to connect to Agama's web user interface skipping the authentication.
+
+TOKEN_FILE=/run/agama/token
+TOKEN=$(cat $TOKEN_FILE)
+
+firefox --kiosk --profile $HOME/.mozilla/firefox/profile "http://localhost/login?token=$TOKEN"
```
lslezak, line 7: The code is simpler without using a helper `token.html` file on the other hand the token is then visible in the `ps` output. I'm not sure how serious is this problem... :thinking:

lslezak: Yes, once you are `root` you can do everything...

The problem is that for running simple `ps` you do not need to be root. If e.g. the avahi daemon running as `avahi` user is vulnerable and allows arbitrary code execution then you could still get the root token if you force it to run `ps` and get the output although it's not running as `root`.

imobachgs: OK, let's give the homepage setting a try.

lslezak (approved):
LGTM

Thanks, that looks much better than the SQLite hack! :+1:

imobachgs:
Yes, you are right. Now the next step is to submit all this stuff automatically. But that's another story. 😄

PR #1120, opened 2024-03-26
Synchronize HTTP and D-Bus localization interfaces

imobachgs:
Trello: https://trello.com/c/npg6apH0/351-make-l10n-http-json-and-d-bus-interfaces-share-the-same-state

At this point, two different (and disconnected) interfaces handle localization settings. The preferred one is HTTP-based, but a D-Bus one is still needed for internal communication.

This pull request tries reorganizing the code and keeping both APIs in sync.

## Tasks

- [x] Keep both APIs in sync
- [x] Reorganize the code to decouple the D-Bus interface from the original `Locale` struct.
- [x] Move common logic to the `Locale` struct (now `L10n`).
- [ ] Find a better name for the `L10n` struct.

### DIFF
--- a/rust/agama-lib/src/localization.rs
+++ b/rust/agama-lib/src/localization.rs
@@ -6,5 +6,6 @@ mod settings;
 mod store;
 
 pub use client::LocalizationClient;
+pub use proxies::LocaleProxy;
 pub use settings::LocalizationSettings;
 pub use store::LocalizationStore;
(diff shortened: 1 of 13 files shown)


dgdavid (approved):

lslezak (approved):
The change looks reasonable, but I'm not a Rust expert... :stuck_out_tongue_closed_eyes:

imobachgs:
Thanks!

PR #1121, opened 2024-03-27
Do not use Cockpit manifest for storing supported languages

lslezak:
## Problem

- Continue with dropping the Cockpit dependency

## Solution

- Do not use the Cockpit manifest for storing the list of supported languages
- Use a separate `languages.json` file instead of the `manifest.json`
- Import the JSON file directly into the JS code
- The workflow is not changed, the update script just generates a different file
- The change is backward compatible, works with both Cockpit and the new server

## Testing

- Updated unit tests
- Tested manually, the language selection still works fine


web/share/update-languages.py
```diff
@@ -124,32 +118,31 @@ def update(self, po_files, lang2territory: str, threshold: int):
...
             include_territory = languages.count(locale.language) > 1
-            self.content["locales"][locale.code()] = locale.name(
-                include_territory)
+            self.content[locale.code()] = locale.name(include_territory)
 
     def write(self):
         with open(self.path, "w+") as file:
-            json.dump(self.content, file, indent=4, ensure_ascii=False)
+            json.dump(self.content, file, indent=4, ensure_ascii=False,
+                      sort_keys=True)
```
lslezak, line 128: The `sort_keys=True` option ensures the list is sorted and avoids random reordering of the items.

web/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "noEmit": true,
     "target": "esnext",
     "moduleResolution": "node",
+    "resolveJsonModule": true,
```
lslezak, line 8: To support loading JSON files.

web/webpack.config.js
```diff
@@ -112,16 +111,6 @@ module.exports = {
         // ignore SSL problems (self-signed certificate)
         secure: false,
       },
-      // forward the manifests.js request and patch the response with the
-      // current Agama manifest from the ./src/manifest.json file
-      "/manifests.js": {
```
lslezak, line 117: We do not need to read  the manifests from the server anymore.

web/share/update-languages.py
```diff
@@ -124,32 +118,31 @@ def update(self, po_files, lang2territory: str, threshold: int):
...
 
 
 if __name__ == "__main__":
-    parser = ArgumentParser(prog="locales.py")
-    parser.set_defaults(func=update_manifest)
+    parser = ArgumentParser(prog="update-languages.py")
+    parser.set_defaults(func=update_languages)
     parser.add_argument(
-        "manifest", type=str, help="Path to the manifest file",
+        "file", type=str, help="Path to the output file",
```
jreidinger, line 145: I would call it output instead of file as it is more similar to input file.

web/cspell.json
```diff
@@ -3,6 +3,7 @@
   "language": "en",
   "allowCompoundWords": false,
   "ignorePaths": [
+    "src/languages.json",
```
jreidinger, line 6: maybe add it also to `.gitignore`?

.github/workflows/weblate-merge-po.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         if: steps.check_changes.outputs.po_updated == 'true'
         working-directory: ./agama
         run: |
-          web/share/update-manifest.py web/src/manifest.json
+          web/share/update-languages.py web/src/languages.json
```
jreidinger, line 95: why not generate it to stdout by default and allow redirect?

web/share/update-languages.py
jreidinger: or maybe even better write to stdout by default and let user redirect it where he needs?

dgdavid (commented):
In general it looks good. I guess it deserves an entry in the changes file too.

dgdavid (approved):
:+1:

web/share/update-languages.py
```diff
@@ -105,14 +109,16 @@ def update(self, po_files, lang2territory: str, threshold: int):
...
+                    file=sys.stderr
                 )
             elif po_file.coverage() < threshold:
                 print(
                     "not enough coverage for '{language}' ({coverage}%)"
                     .format(
                         language=locale.code(),
-                        coverage=po_file.coverage())
+                        coverage=po_file.coverage()),
+                    file=sys.stderr
```
jreidinger, line 121: :+1: for writting errors to stderr

PR #1122, opened 2024-03-27
Added keyboard support when navigating suggested usernames

### LINKED ISSUE #1070: Add keyboard support for auto-suggested usernames
The list of suggested usernames introduced on #1022 lacks of keyboard support. A [know issue](https://github.com/openSUSE/agama/pull/1022#discussion_r1495704161) that is going to be addressed when time permits.

Initially we agreed on not opening an specific issue for it and refer to #112 instead.

However, I have though that could be useful to left here some useful articles/docs that might help to implement it. See comments below.

balsa-asanovic:
## Problem

In #1022 the username auto-suggest feature was introduced, but the dropdown with usernames didn't support keyboard navigation.

This PR aims to solve this, It introduces navigation with arrow keys (up and down) and sets the value on the press of Enter.

fixes #1070 

## Solution

Now checking for keyboard events on the username input field and on the arrow presses changing the index of the selected menu item, highlighting the appropriate one. On the press of Enter, again using the current value of index, the text input is populated with a value from the usernames array.

## Testing

- Tested manually

## Screenshots

image


web/src/components/users/FirstUser.jsx
```diff
@@ -220,11 +223,38 @@ export default function FirstUser() {
...
 
+  const handleKeyDown = (event) => {
+    switch (event.key) {
+      case 'ArrowDown':
+        event.preventDefault(); // Prevent page scrolling
+        setFocusedIndex((prevIndex) => (prevIndex + 1) % suggestions.length);
+        break;
+      case 'ArrowUp':
+        event.preventDefault(); // Prevent page scrolling
+        setFocusedIndex((prevIndex) => (prevIndex - 1 + suggestions.length) % suggestions.length);
```
dgdavid, line 246: With this code, when pressing the <kbd>Arrow Up</kbd> when suggestion does not have the focus yet, the next-to-last suggestion get focused instead of the last one.

For example, having 4 suggestions:

```
FirstUser.jsx:247 prevIndex: -1
FirstUser.jsx:248 suggestions.length: 4
FirstUser.jsx:249 result of (prevIndex - 1 + suggestions.length) % suggestions.length: 2
```

dgdavid (commented):
Ey @balsa-asanovic,

Thank you for this one. I have had a quick look and added two minor comments, but I'd like to think a about it a bit more to see if there is another approach not using these two new internal pieces of state in the component. Nothing again them, but maybe we can avoid the extra re-renders when the focused suggestion change.

Meanwhile, I think we can go ahead adding the tests for this feature because it does not depend on the implementation. Let me know if you need help with these tests.

dgdavid:
> I have had a quick look and added two minor comments

Looks like I have missed one, will add it again.

dgdavid:
@balsa-asanovic Sorry for not catching this before, but while giving a try to this PR I have found two issues introduced in #1022. Please, check https://github.com/openSUSE/agama/issues/1124 if you don't mind.

web/src/components/users/FirstUser.jsx
```diff
@@ -220,11 +223,38 @@ export default function FirstUser() {
...
+        break;
+      case 'Enter':
+        if (focusedIndex >= 0) {
+          onSuggestionSelected(suggestions[focusedIndex]);
+        }
+        break;
+      default:
```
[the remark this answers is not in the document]
balsa-asanovic: When the dropdown is closed on Escape (and the username input is still empty), should the dropdown be reopened on arrow up/down (while not losing focus)? If the focus leaves the field and then comes back, the usernames will be shown anyhow.

balsa-asanovic:
Hey @dgdavid 

I've addressed your comments and issues mentioned in #1124 

Next, I'll be getting to writing tests. I'll ping you if I need any help with that.

web/src/components/users/FirstUser.jsx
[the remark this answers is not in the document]
balsa-asanovic: Currently set it up so that arrow up/down shows the suggestions again, even if if the focus never left the input.

dgdavid (commented):
Thanks for adding tests! I have left some nitpicking comments :innocent:, please check them.

Feel free to reject some of the suggestions, but please include the two missing examples.

balsa-asanovic:
> Thanks for adding tests! I have left some nitpicking comments :innocent:, please check them.
> 
> 
> 
> Feel free to reject some of the suggestions, but please include the two missing examples.

Hey David,

Thanks for the comments.

I'll be checking them out.

I actually meant to add more tests, I've just started with a few last night, that's why I didn't ping you here 😄

dgdavid:
> I actually meant to add more tests, I've just started with a few last night, that's why I didn't ping you here 😄

Ah, excuse me then. In any case, I'll be happy if this early review saves you some keystrokes :) :)

balsa-asanovic:
> Ah, excuse me then. In any case, I'll be happy if this early review saves you some keystrokes :) :)

Oh, no problem, your comments helped. 😄 

Okay, so I think I addressed everything you mentioned, added missing tests.

I've also added one more using keyboard strokes. This is maybe not needed, so check it out.

web/src/components/users/FirstUser.test.jsx
```diff
@@ -276,3 +286,104 @@ describe("when the user has been modified", () => {
...
+    await user.tab();
+
+    const menuItem = screen.getByText('janed');
+    const usernameInput = within(dialog).getByLabelText("Username");
+    
+    await act(async () => {
+      menuItem.click();
+    });
+
+    expect(usernameInput.value).toBe("janed");
```
dgdavid, line 368: CSpell complaining here and a few lines above :facepalm: 

You can do either, to add the `/* cspell:disable-next-line */` comment before these lines... or to cheat it by filling the full name input with "User Name" and then clicking on "username" :sweat_smile: :scream:

web/src/components/users/FirstUser.test.jsx
```diff
@@ -276,3 +286,104 @@ describe("when the user has been modified", () => {
...
+    const fullNameInput = within(dialog).getByLabelText("Full name");
+    await user.type(fullNameInput, "Jane Doe");
+
+    await user.tab();
+
+    const menuItem = screen.getByText('janed');
+    const usernameInput = within(dialog).getByLabelText("Username");
+    
+    await act(async () => {
+      menuItem.click();
```
dgdavid, line 365: NP: I'd use below instead

```suggestion
      await user.click(menuItem);
```

I think it is _semantically_ more close to what you're testing although the result is the same.

web/src/components/users/FirstUser.test.jsx
```diff
@@ -41,6 +41,16 @@ let setUserFn = jest.fn().mockResolvedValue(setUserResult);
...
 
+const openUserForm = async () => {
+  const { user } = installerRender(<FirstUser />);
+  await screen.findByText("No user defined yet.");
+  const button = await screen.findByRole("button", { name: "Define a user now" });
+  await user.click(button);
+  const dialog = await screen.findByRole("dialog");
+
+  return { user, dialog }
+}
```
dgdavid, line 52: NP: Missing semicolons, 

```suggestion
  return { user, dialog };
};
```

web/src/components/users/FirstUser.test.jsx
```diff
@@ -276,3 +286,104 @@ describe("when the user has been modified", () => {
...
+    let menuItems = screen.queryAllByText("Use suggested username");
+    expect(menuItems.length).toBe(4)
+
+    const usernameInput = within(dialog).getByLabelText("Username");
+    // the user now types something
+    await user.type(usernameInput, "John Smith");
+
+    // checking if suggestions are gone
+    menuItems = screen.queryAllByText("Use suggested username");
+    expect(menuItems.length).toBe(0)
```
dgdavid, line 350:
```suggestion
    expect(menuItems.length).toBe(0);
```

web/src/components/users/FirstUser.jsx
```diff
@@ -220,11 +223,44 @@ export default function FirstUser() {
...
+  useEffect(() => {
+    if (displaySuggestions) {
+      setFocusedIndex(-1);
+      setSuggestions(suggestUsernames(formValues.fullName));
+    }
+  }, [displaySuggestions, formValues.fullName]);
+
   const onSuggestionSelected = (suggestion) => {
     setInsideDropDown(false);
     setFormValues({ ...formValues, userName: suggestion });
```
dgdavid, line 235: One thing more.

While doing a manual test, I've realized that the focus is now lost when a suggestion is chosen by clicking on it. It works like a charm if using the keyboard instead.

I think would be enough to create a `ref` (let's say, usernameInputRef) and use it to ensure the focus comes back here again.  Just saying, maybe there is a better, more elegant solution.

```suggestion
    setFormValues({ ...formValues, userName: suggestion });
    usernameInputRef.current?.focus()
```

web/src/components/users/FirstUser.test.jsx
dgdavid, line 368: In order to check that the focus is not lost, let's add one more expectation here.

```suggestion
    expect(usernameInput).toHaveFocus();
    expect(usernameInput.value).toBe("janed");
```

dgdavid (commented):
In manual testing I have found a little problem with focus when using the mouse for selecting a suggestion. Please check my comments.

web/src/components/users/FirstUser.test.jsx
dgdavid: Also, add such an expectation in other examples if it makes sense.

balsa-asanovic:
Hey @dgdavid 

Addressed the comments.

Sorry for the big delay in dealing with these small issues.

> Just saying, maybe there is a better, more elegant solution.

I went with the ref choice, since I didn't come up with anything better.
I am not sure here, why does it lose focus in the first place when using click action, but with the keyboard it remains.. 🤔

dgdavid (approved):
@balsa-asanovic 

Once more, **thanks a ton!**

I've fixed minor ESLint complains and... will merge as soon as everything is green in CI.

dgdavid:
> will merge as soon as everything is green in CI.

Ouch!

Do you mind add an entry in the changes file? Or do you prefer that I add it?

balsa-asanovic:
> > will merge as soon as everything is green in CI.
> 
> Ouch!
> 
> Do you mind add an entry in the changes file? Or do you prefer that I add it?

Don't mind at all, I'll add it soon.

balsa-asanovic:
@dgdavid Added entry to changes file.

dgdavid:
Great! Thanks!

ancorgs:
ntioned in the [announcement of Agama 8](https://yast.opensuse.org/blog/2024-05-17/agama-8).

Thanks for your contributions!

PR #1112, opened 2024-03-22
Adapt the software UI to the new HTTP/JSON API

imobachgs:
The initial idea was to adapt the software page and the patterns selector to the new HTTP-based API. However, we took the chance to revamp the software page to make it more similar to other Agama parts.

<details>
  <summary>New software page</summary>

software-page
</details>

<details>
  <summary>New patterns selector</summary>

software-selection
</details>

## Changes summary

* Introduce a new software page that lists the patterns to install.
* Move the patterns selector to a pop-up.
* Introduce small changes to the software API.
* Update the JavaScript tests and enable the type checking on the new code.

## What is missing

- [ ] Distinguish between user and automatically selected patterns.

## Out of scope

* Improve the wording.

PR #1123, opened 2024-04-01
web: Improve Patterns selector UI

dgdavid:
In #1112, the _Patterns selector_ was changed to be more consistent with the rest of the UI selection dialogs. However, the _core/Selector_ component used under the hood wasn't ready to handle auto selection. A partial adaptation was made by allowing overriding how it behaves when an option is clicked (see 61b0c0f), but the UI part is being addressed here by

* Improving how the information is layout
* Letting the user know which patterns are auto selected
  - Using a different _accent-color_ for the checkbox.
    
    After a few attemps, @imobachgs and me decided to go for a white accent color by now since it was the less out of line. We tested using the primary color (too dark, a bit confusing), only the border (quite confusing), using a dotted bottom border (it does not provide too much information nowadays, honestly), a solid square (weird). In any case, this is just an starting point, we can improve these details later, when we have time for playing a bit and/or giving love to the whole UI radio/checkbox.

  - Explicitly adding the "auto selected" text below the checkbox, since a just a different checkbox style (whatever we choose) is not enough for letting the user know that it has been auto selected pattern in a quick way. And no, a tooltip is not a good solution for this use case: it is hidden until the user focus or hover the widget.

| Before | After |
|-|-|
| Screenshot from 2024-04-01 21-01-07 | Screenshot from 2024-04-01 21-01-36 |

The "auto selected" text might looks a bit longer, but honestly "auto" does not provide too much value there: _auto what?_ Maybe we can put it in another place (as a label in the bottom right corner of the item. or in the top left. Whatever. In any case, **please accept this change as it is** unless you find a very strong stopper. As said above, we can (and should) improve this later, when we can invest the deserved time for this kind of things. And, IMHO, the improvement should start in the back-end by allowing the user to get back to the auto selected state. As it is now is ok (I guess) for YaST user, but it quite confusing. Of course, I know that it isn't something trivial.

### DIFF
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -248,7 +248,7 @@ ul[data-type="agama/list"] {
 
   li {
     border: 2px solid var(--color-gray-dark);
-    padding: var(--spacer-normal);
+    padding: var(--spacer-small);
     text-align: start;
     background: var(--color-gray-light);
     margin-block-end: 0;
@@ -276,7 +276,6 @@ ul[data-type="agama/list"] {
   // not belongs to a listbox or grid list ul.
   li[aria-selected] {
     background: var(--color-gray-dark);
-    font-weight: 700;
 
     &:not(:last-child) {
       border-bottom-color: white;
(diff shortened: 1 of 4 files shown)


web/src/assets/styles/blocks.scss
```diff
@@ -386,6 +403,18 @@ ul[data-type="agama/list"][role="grid"] {
   }
 }
 
+ul[data-items-type="agama/patterns"] {
```
dgdavid, line 406: NOTE: this change reveals that this part, the core/Selector styles, can be simplified / improved. Let's wait a bit until selectors are more defined (we're working in new storage selectors that might have impact in such simplification if finally we get rid of storage/device-utils/DeviceSelector.

imobachgs (approved):

web/src/components/software/PatternSelector.jsx
```diff
@@ -134,14 +134,16 @@ function PatternSelector({ patterns, onSelectionChanged = noop }) {
...
+        <div>
+          <b>{pattern.summary}</b>
+        </div>
+        <div>{pattern.description}</div>
       </div>
-      <div>{pattern.description}</div>
-    </>
-  );
+    );
+  }
```
imobachgs, line 146:
```suggestion
  };
```

PR #1125, opened 2024-04-03
[web] Changes in the volumes table

ancorgs:
## Problem

The [storage_ui document](https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md) shows a certain format for the table of volumes (a.k.a. file systems), with four columns including a "location" one.

But Agama still included an old format for that table.

## Solution

Implement the table described in the document (with minor changes to the strings):

volumes

## Testing

- Added unit tests
- Tested manually  (see screenshots above) hacking a bit the info from the backend (since there is no UI yet to configure alternative locations)

### DIFF
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -131,7 +131,7 @@ const ZFCP_DISK_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Disk";
  * @typedef {object} Volume
  * @property {string} mountPath
  * @property {string} target
- * @property {string} [targetDevice]
+ * @property {StorageDevice} [targetDevice]
  * @property {string} fsType
  * @property {number} minSize
  * @property {number} [maxSize]
@@ -430,7 +430,8 @@ class ProposalManager {
    */
   async defaultVolume(mountPath) {
     const proxy = await this.proxies.proposalCalculator;
-    return this.buildVolume(await proxy.DefaultVolume(mountPath));
+    const systemDevices = await this.system.getDevices();
+    return this.buildVolume(await proxy.DefaultVolume(mountPath), systemDevices);
   }
 
   /**
@@ -505,7 +506,7 @@ class ProposalManager {
           spaceActions: dbusSettings.SpaceActions.v.map(a => buildSpaceAction(a.v)),
           encryptionPassword: dbusSettings.EncryptionPassword.v,
           encryptionMethod: dbusSettings.EncryptionMethod.v,
-          volumes: dbusSettings.Volumes.v.map(vol => this.buildVolume(vol.v)),
+          volumes: dbusSettings.Volumes.v.map(vol => this.buildVolume(vol.v, systemDevices)),
           // NOTE: strictly speaking, installation devices does not belong to the settings. It
           // should be a separate method instead of an attribute in the settings object.
           // Nevertheless, it was added here for simplicity and to avoid passing more props in some
@@ -559,6 +560,8 @@ class ProposalManager {
         MinSize: { t: "t", v: volume.minSize },
         MaxSize: { t: "t", v: volume.maxSize },
         AutoSize: { t: "b", v: volume.autoSize },
+        Target: { t: "s", v: volume.target },
+        TargetDevice: { t: "s", v: volume.targetDevice?.name },
         Snapshots: { t: "b", v: volume.snapshots },
         Transactional: { t: "b", v: volume.transactional },
       });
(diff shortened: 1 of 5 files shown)


web/src/client/storage.js
```diff
@@ -645,8 +650,8 @@ class ProposalManager {
     };
 
     return {
-      target: dbusVolume.Target.v,
-      targetDevice: dbusVolume.TargetDevice?.v,
+      target: dbusVolume.Target?.v || "default",
```
joseivanlopez, line 653: NP: I guess we can assume this value always comes from D-Bus.

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -203,25 +206,48 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
...
+      if (transactional)
+        return _("Transactional Btrfs");
+      if (snapshots)
+        return _("Btrfs with snapshots");
+
+      return volume.fsType;
+    };
+
+    return (
+      <div className="split">
```
joseivanlopez, line 226: Is this `div` needed?

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -203,25 +206,48 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
...
+      if (volume.target === "device" || volume.target === "filesystem")
+        return volume.targetDevice?.name || "";
+      if (options.lvm)
+        return _("Logical volume at system LVM");
+
+      return _("Partition at installation disk");
+    };
 
     return (
       <div className="split">
```
joseivanlopez, line 249: The same for this one.

joseivanlopez:

web/src/components/storage/ProposalVolumes.jsx
dgdavid: No, it's not.

web/src/client/storage.js
ancorgs: Fine with me

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -203,25 +206,48 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
...
+        return _("Transactional Btrfs");
+      if (snapshots)
+        return _("Btrfs with snapshots");
+
+      return volume.fsType;
+    };
+
+    return (
+      <div className="split">
+        <span>{text()}</span>
```
dgdavid, line 227: I would not use this kind of function. As far as I can see, you can use a `let text;` and then fill it's value conditionally and finally use it directly.

```js
let text;

if (volume.target === "filesystem") {
  // TRANSLATORS: %s will be replaced by a file-system type like "Btrfs" or "Ext4"
  text =  sprintf(_("Reused %s"), volume.targetDevice?.filesystem?.type || "");
} else if (transactional) {
   text =  _("Transactional Btrfs");
} else if (snapshots) {
   text = _("Btrfs with snapshots");
};
```

Or whatever other technique but not a definition and evaluation of a function.

dgdavid: The above is because I expect this internal component to be re-rendered as soon as its volume prop changes.

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -203,25 +206,48 @@ const VolumeRow = ({ columns, volume, options, isLoading, onEdit, onDelete }) =>
...
     return (
       <div className="split">
-        <span>{text}</span>
-        {/* TRANSLATORS: filesystem flag, it uses an encryption */}
-        <If condition={options.encryption} then={<Em icon={lockIcon}>{_("encrypted")}</Em>} />
-        {/* TRANSLATORS: filesystem flag, it allows creating snapshots */}
-        <If condition={snapshots && !transactional} then={<Em icon={snapshotsIcon}>{_("with snapshots")}</Em>} />
-        {/* TRANSLATORS: flag for transactional file system  */}
-        <If condition={transactional} then={<Em icon={transactionalIcon}>{_("transactional")}</Em>} />
+        <span>{text()}</span>
```
dgdavid, line 250: The same here, do not use a function for this please.

dgdavid (approved):
LGTM, thank you!

joseivanlopez (approved):
LGTM

PR #1126, opened 2024-04-03
Provide the frontend translations via the /po.js path

lslezak:
## Problem

- Reimplement the translation support for the web frontend

## Solution

- Route the `/po.js` path to the code which redirects to the requested translations or returns an empty JS file if the requested language is not found.

### DIFF
--- a/rust/agama-server/src/web/http.rs
+++ b/rust/agama-server/src/web/http.rs
@@ -5,11 +5,13 @@ use super::{
     state::ServiceState,
 };
 use axum::{
+    body::Body,
     extract::State,
-    http::{header::SET_COOKIE, HeaderMap},
+    http::{header, HeaderMap, HeaderValue, StatusCode},
     response::IntoResponse,
     Json,
 };
+use axum_extra::extract::cookie::CookieJar;
 use pam::Client;
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
@@ -62,7 +64,7 @@ pub async fn login(
     let mut headers = HeaderMap::new();
     let cookie = format!("agamaToken={}; HttpOnly", &token);
     headers.insert(
-        SET_COOKIE,
+        header::SET_COOKIE,
         cookie.parse().expect("could not build a valid cookie"),
     );
 
@@ -76,7 +78,7 @@ pub async fn logout(_claims: TokenClaims) -> Result<impl IntoResponse, AuthError
     let mut headers = HeaderMap::new();
     let cookie = "agamaToken=deleted; HttpOnly; Expires=Thu, 01 Jan 1970 00:00:00 GMT".to_string();
     headers.insert(
-        SET_COOKIE,
+        header::SET_COOKIE,
         cookie.parse().expect("could not build a valid cookie"),
     );
     Ok(headers)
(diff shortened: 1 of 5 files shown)


rust/agama-server/src/web/service.rs
```diff
@@ -85,8 +88,64 @@ impl MainServiceBuilder {
...
+        async fn po(jar: CookieJar) -> impl IntoResponse {
+            let mut response_headers = HeaderMap::new();
+
+            if let Some(cookie) = jar.get("agamaLanguage") {
+                let mut target_file = String::new();
+                let mut found = false;
+                // FIXME: this does not work, the public_dir setting is not accessible :-/
+                // when using something like PathBuf::from("/usr/share/cockpit/agama") here
+                // it works just fine....
+                let prefix = self.public_dir;
```
lslezak, line 104: Here I need to somehow access the `self.public_dir` setting. I have no idea how to do that. Use a closure? Implement an axum service?

Any idea Rust experts?

rust/agama-server/src/web/service.rs
```diff
@@ -85,8 +88,64 @@ impl MainServiceBuilder {
             .route("/auth", post(login).get(session).delete(logout));
 
         let serve = ServeDir::new(self.public_dir);
+
+        // handle the /po.js request
+        // the requested language (locale) is sent in the "agamaLanguage" HTTP cookie
+        // this reimplements the Cockpit translation support
+        async fn po(jar: CookieJar) -> impl IntoResponse {
```
lslezak, line 95: BTW this should be probably moved elsewhere. To `web/http.rs`? Any suggestion?

lslezak (commented):
Questions

rust/agama-server/src/web/service.rs
imobachgs: Yes, `http.rs` looks good. And if it grows, it could have its own module.

imobachgs (approved):
LGTM

rust/agama-server/src/web/http.rs
```diff
@@ -90,3 +92,61 @@ pub async fn logout(_claims: TokenClaims) -> Result<impl IntoResponse, AuthError
...
+                let target_file = format!("po.{}.js", lang);
+                if state.public_dir.join(&target_file).exists() {
+                    return redirect_to_file(&target_file);
+                };
+            }
+        } else {
+            // use the cookie as is
+            let target_file = format!("po.{}.js", cookie.value());
+            if state.public_dir.join(&target_file).exists() {
+                return redirect_to_file(&target_file);
```
jreidinger, line 138: I see three times almost identical code. Maybe you can use something like

```
fn redirect_to_file(state, file) -> Option<impl IntoResponse>

// and here in code use something like
if Some(response) = redirect_to_file(&state, &file) {
  return response
} 
```

this way logic with testing file existence in public dir is implemented just once.

jreidinger: maybe name can be `serve_file_if_exist`

lslezak: Tried that, it did not look any better, the added Some/None handling made the code less readable IMHO...

PR #1128, opened 2024-04-04
web: Improve find space

joseivanlopez:
Improvements in the dialog to find space:

* Includes the unused space.
* Show devices and slots sorted by position.
* The colums *Device* and *Details show similar information to the Result table.
* Remove *Size details* in favor of *Shrinkable* column.

Code improvements:

* Remove `ProposalSpacePolicyField` in favor of `SpacePolicyDialog` component, following the same apprach that the dialogs for boot and device configuration.
* Enable type checking in several components.
* Add more options to `TreeTable` (expandable rows, `expadedItems` props). 

## Screenshots

localhost_8080_ (34)


web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -438,9 +503,9 @@ export default function ProposalSettingsSection({
           isLoading={isLoading}
           onChange={changeBoot}
         />
-        <ProposalSpacePolicyField
-          policy={settings.spacePolicy}
-          actions={settings.spaceActions}
+        <SpacePolicyField
```
dgdavid, line 506: I have been thinking on drop the `Proposal` prefix from these fields too. In fact, I talked about it with @ancorgs in a call discussing about the PBI I'm working on. However, I believe would be better to move it field to its own file. We can see it in the context of my PBI.

dgdavid (approved):
Great job, thanks! :+1:

PR #1130, opened 2024-04-05
Fix webpack configuration to load CockpitPoPlugin

imobachgs:
Add a missing line to load the CockpitCoPlugin in webpack.


lslezak (approved):
Thanks! Looks like I made a mistake when merging master...

PR #1136, opened 2024-04-08
adapt debugging hints

jreidinger:
## Problem

With SSL changes debugging tips no longer reflect reality.

## Solution

Adapt -> succeed

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -27,7 +27,7 @@ This can work also remotely, with a Agama instance running in a different
 machine (a virtual machine as well). In that case run
 
 ```
-    AGAMA_SERVER=<IP> npm run server -- --open
+    AGAMA_SERVER=https://<IP>:<port> npm run server -- --open
 ```
 
 Where  `AGAMA_SERVER` is the IP address, the hostname or the full URL of the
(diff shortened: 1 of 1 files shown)


web/README.md
```diff
@@ -54,6 +55,8 @@ The first place is the browser's console which can give
 some hints. The second location to check for errors or warnings is output of `npm run server`
 where you can find issues when communicating with the backend. And last but on least is
 journal on backend machine where is logged backend activity `journalctl -b`.
+If journal does not provide required info, there is also option to see dbus communication
```
imobachgs, line 58:
```suggestion
If the journal does not contain the required info, you can inspect the D-Bus communication
```

web/README.md
```diff
@@ -54,6 +55,8 @@ The first place is the browser's console which can give
...
+which can give hint about data flow. Command is `busctl monitor --address /run/agama/bus`
```
imobachgs, line 59: I think the address is `unix:path=/run/agama/bus`

jreidinger: right

lslezak (approved):

PR #1138, opened 2024-04-08
web: Revamp storage proposal page

dgdavid:
The Storage Page has become a bit difficult to navigate and a bit overwhelming at first glance, making it more likely to scare users than help them.

The changes in this PR address several weak points we identified: dynamic button/link text, inconsistent widget placement/alignment, and too much information exposed in the first render. As such, the UI proposed is more straightforward, but it is still relevant and helpful for everyone, including advanced and basic users. It keeps contextual help visible and close to the widget/action it belongs to, without diverting user attention from the important bits.

At the time of introducing these changes, the rest of Agama was also taken into consideration so that we could easily maintain consistency by just using the new widgets when needed.

For a better understanding of the changes made, please check commit by commit.

---

Related to https://trello.com/c/czpTfm3y/3628-5-agama-polish-storage-section (internal link)


web/src/components/core/Field.jsx
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @typedef {Omit<FieldProps, 'icon'>} FieldPropsWithoutIcon
+ */
+
+/**
+ * Component for laying out a page field
+ *
+ * @param {FieldProps} props
+ */
+const Field = ({
```
dgdavid, line 52: I've called it _Field_ but definitely is not the best name. A field usually has a label and value, which is the case for most of the use cases it will have. It might happen that a field does not have a _simple value_ but a complex one, which is the case of the "Partitions and file systems" (see attached screenshot in the PR's description). In this case, the value is actually rendered in the _content_ part, a.k.a _children_. 

So, this looks like more a _subsection_ or so, but I'm not sure. Anyway, if you have a better naming in mind... please speak up. If not, never will be late for running the [sed](https://linux.die.net/man/1/sed) command against the whole code base :stuck_out_tongue_winking_eye:

web/src/assets/styles/blocks.scss
```diff
@@ -743,6 +743,51 @@ section [data-type="agama/reminder"] {
...
+    }
+  }
+
+  hr {
+    margin-block: var(--spacer-normal);
+    border: 0;
+    border-bottom: thin dashed var(--color-gray);
+  }
+}
+
```
dgdavid, line 790: Let's start with this. Later we can try to go for a better solution with grids and subgrids or other _advanced_ CSS stuff.

web/src/assets/styles/blocks.scss
```diff
@@ -743,6 +743,51 @@ section [data-type="agama/reminder"] {
...
+  > div:nth-child(n+3) {
+    margin-block-start: var(--spacer-small);
+  }
+
+  &.on {
+    button {
+      fill: var(--color-link-hover);
+    }
+  }
```
dgdavid, line 782: Something not writing in stone, of course.

lslezak:
Looking at the screenshot, I'm not sure if the "Encryption" and "Snapshots" should be the top level proposal sections. I'd probably put them below the "Installation Device" part as subsections...

IMHO those setting are less important (less used) so they should probably less stand out. I see too many top level options, it is more difficult to find out the important parts.

ancorgs:
> Looking at the screenshot, I'm not sure if the "Encryption" and "Snapshots" should be the top level proposal sections.

I'm convinced "Encryption" IS a top-level section. We really want to encourage encryption... to the point of publishing pre-encrypted images for our non-installable distributions (like SLE-Micro) and having developed a whole infrastructure to re-encrypt them on first boot and so on.

I'm more open to the discussion about "Btrfs Snapshots". Now that we have "Partitions and file systems" as a collapsible section, maybe we could move that setting inside.

> I'd probably put them below the "Installation Device" part as subsections...

If you mean putting there somehow inside "Installation Device" I would strongly disagree. The do not belong there.

> IMHO those setting are less important (less used) so they should probably less stand out. I see too many top level options, it is more difficult to find out the important parts.

I don't think those settings (especially encryption) are really less used.

dgdavid:
@lslezak, @ancorgs 

Based on your comments, that's the new proposal:

| Snapshots disabled | Snapshots enabled |
|-|-|
| Screenshot from 2024-04-09 15-59-48 | Screenshot from 2024-04-09 15-59-51 | 


<details>
<summary>Click to show/hide other ideas</summary>

---

Just a NP comment: we can even stop mentioning `Btrfs` in the snapshots label and put it in the explanation.


Screenshot from 2024-04-09 16-00-52


Also, we can tweak the styles, if needed, for these internal fields. But later, when every thing is stable. Note below that I simply removed the _left margin_ of the explanation for them.

Screenshot from 2024-04-09 15-59-34
</details>

web/src/components/storage/utils.js
```diff
@@ -215,7 +215,7 @@ const deviceLabel = (device) => {
 const deviceChildren = (device) => {
   const partitionTableChildren = (partitionTable) => {
     const { partitions, unusedSlots } = partitionTable;
-    const children = partitions.concat(unusedSlots);
+    const children = partitions.concat(unusedSlots).filter(i => !!i);
```
dgdavid, line 218: I detected that sometimes this ends up with [partition, partition, undefined]. We should add a test for this util.

web/src/components/storage/SpacePolicyField.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+  policy: keepPolicy,
+  isLoading: false,
+  onChange: jest.fn(),
+  actions: [
+    { device: "/dev/sda", action: "force_delete" },
+  ]
+};
+
+describe("SpacePolicyField", () => {
+  it("renders a button for opening the space policy dialog", async () => {
```
dgdavid, line 65: TODO: add a test for the value too, which reveals that maybe a test for SPACE_POLICIES should be added in storage-utils.test.js.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -25,8 +25,9 @@ import React, { useEffect, useState } from "react";
 import { Button, Checkbox, Form, Skeleton, Switch, Tooltip } from "@patternfly/react-core";
 
 import { sprintf } from "sprintf-js";
-import { _, n_ } from "~/i18n";
-import { BootSelectionDialog, ProposalVolumes, SpacePolicyDialog } from "~/components/storage";
+import { _ } from "~/i18n";
+import { BootSelectionDialog, ProposalVolumes } from "~/components/storage";
+import SpacePolicyField from "~/components/storage/SpacePolicyField";
```
dgdavid, line 30: Although we wanted to do stop exporting _internal namespace components_ at once, I've started here. We can adapt the other components later, but let's start with new ASAP.

lslezak:
| Current State | Proposed Simplification |
|---|---|
| agama_storage_original | agama_storage_redesign |

I tried to do some simplification, esp. removing or hiding the unnecessary text.

- Removed text "Select the main disk or LVM Volume Group for installation."
- The Btrfs snaphost description moved to the "i" icon
- "Find space" moved up to the other settings
- Make the "File systems to create" label a header (make it bold), removed the description (IMHO not needed)
- Simplified the boot loader config link, removed the confusing icon
- Removed the description in the Result section, moved the number of actions to the link

What do you think about it? Should we go this way? I can add my commit here if you want to improve it somehow...

imobachgs:
> I tried to do some simplification, esp. removing or hiding the unnecessary text.
> 
>     * Removed text "Select the main disk or LVM Volume Group for installation."
> 
>     * The Btrfs snaphost description moved to the "i" icon

In general, I like the idea of having those texts. They help to understand the interface and, in general, they are a much better approach than the old YaST "Help" button. Additionally, it looks a little inconsistent to remove one of the texts and use an "i" icon for the other.

>     * "Find space" moved up to the other settings
> 
>     * Make the "File systems to create" label a header (make it bold), removed the description (IMHO not needed)

Please, check @dgdavid's proposal. AFAIU there will be a "Partitions and file systems" (similar to your idea) section. Actually, the Btrfs will be moved to that section following [your proposal](https://github.com/openSUSE/agama/pull/1138#issuecomment-2044682226) to make it less prominent.

> 
>     * Simplified the boot loader config link, removed the confusing icon
> 
>     * Removed the description in the Result section, moved the number of actions to the link

Again, I would keep the text. What "28 planned actions" means?

> What do you think about it? Should we go this way? I can add my commit here if you want to improve it somehow...

lslezak:
> In general, I like the idea of having those texts. They help to understand the interface and, in general, they are a much better approach than the old YaST "Help" button.

OK, maybe I was too eager with the cleanup. I'm not against additional texts but in some cases I have a feeling that we are just wasting the space.

"Select the main disk or LVM Volume Group for installation."
"Installation device /dev/sda, 20GB"

Why not merging that to a single

"Selected installation device /dev/sda, 20GB" or something similar? 

> Additionally, it looks a little inconsistent to remove one of the texts and use an "i" icon for the other.

I felt this is a very long description just for a small switch. After looking at the screen again the real problem is that the additional texts are not unified.

- The Btrfs snapshot description uses normal text style (standard size and color)
- "Select main disk ..." uses standard size with lighter color
- The "Find space" text uses smaller size and lighter color

IMHO that's a bit confusing, all such texts should look the same.

> Again, I would keep the text. What "28 planned actions" means?

Probably too much simplification again, but I wanted to merge 

"During installation 28 actions will be performed...."
"Check all planned actions"

Maybe something like "28 planned actions for changing the system storage configuration" with "28 planned actions" as a link.

lslezak:
> Please, check @dgdavid's proposal. AFAIU there will be a "Partitions and file systems" (similar to your idea) section. Actually, the Btrfs will be moved to that section following [your proposal](https://github.com/openSUSE/agama/pull/1138#issuecomment-2044682226) to make it less prominent.

I based my change on the current code in this branch, that change is not commited here.

lslezak:
Great, the current state looks much better! What about this still small simplification?

| Current | Proposal |
|---|---|
| agama_storage_result_current | agama_storage_result_proposal |

web/src/components/core/Field.jsx
ancorgs: "Field" is fine for now.

ancorgs (approved):
Of course there is a lot of room for improvement, but I see no reason for merging this into master. At least three developers have been playing around with it and we all agree is a clear step forward.

Thus, approving (although this likely deserves a changelog entry).

dgdavid:
> Great, the current state looks much better! What about this still small simplification?
> 
> Current	Proposal
> agama_storage_result_current	agama_storage_result_proposal


Hi @lslezak,

First of all, let me skip the [previous proposal](https://github.com/openSUSE/agama/pull/1138#issuecomment-2051036144) since as far as I could see it there was a kind of mis-understanding or lack of sync.

Regarding this one, let me **thank you twice**: for the proposal itself and for using the Agama infrastructure to make it possible. No matter whether the proposal is finally adopted or not, playing with Agama to show what you had in mind is hands-down a full win-win since it comes with benefits for all: you, the project, and the team. At least that is what I believe, because it gives you the opportunity to get familiar with the stack, to improve your front-end skills, and even to spot something to fix or improve.

As I said in IRC on Friday, at first sight the proposal LGTM. But still prefer the current status because I feel it could be more meaningful for a broader set of users. To start with, the "During installation" clarification. Probably too obvious for you, and it might seem obvious to a large audience. But IMHO it does not hurt and can help other users. I'm pretty sure there is a lot of for improvements in the UI, but at this moment I think it has a good balance of options and text/inline-help.

In the second place, believe it or not, the "Check all planned actions" was placed there on purpose after quite a few tests and discussions with @ancorgs  and @joseivanlopez . In short, there is an intended sequence

  1. Let the user know in a "less disruptive way" the concept of actions and how many will be taken.
  2. Show a "not too much aggressive warning" about destructive actions, if any.
  3. Provide the link to "invite" users to check all the planned actions.

For me still makes sense in that order. Of course, nothing is written in stone and, as said, the UI still has room for improvements here and there. Moreover, proposals like yours are more than welcome as they could come with fresh ideas and a different POV.

About "simplification" I'd like to ask you guys to have no fear of adding needed texts to the UI. I might be wrong, but without neglecting the importance of concision, I do not think we will have a better and cleaner UI by neither reducing texts nor trying to make it fit in "just one screen". We're managing complex concepts and users deserve all the help we can provide: from a fluid layout now that we're delivering a Web product to not hiding explanations when they are needed and not hindering those who do not need them. Storage Proposal is a very good example of a complex part of Agama that, I believe, is now in a shape that doesn't scare and guides the user. What is more, it fits now in the @ancorgs screen but falling back to scrollable UI when the user has a smaller screen, more complex setup, or just using a higher zoom level.

Talking about the scroll, I'd like to encourage all the team to embrace it. All modern UI are based on vertical scroll. Timelines, social network app, book apps, and so on. Users are more than used to using scroll and IMHO we should not deny it when laying out Agama pages. It could lead us to make mistakes for a concern that might not have sense in a web UI.

Finally, although I have a lot to learn about UI/UX/a11y/Design, I generally try to consider all kinds of users when working with the Agama UI. As an example, I tend to consider the focus/tab behavior when trying to create a more keyboard friendly user interface. That's might explain why I tend to avoid certain resources or to propose to do something in a different way we are used to.

dgdavid:
> About "simplification" I'd like to ask you guys to have no fear of adding needed texts to the UI. I might be wrong, but without neglecting the importance of concision, I do not think we will have a better and cleaner UI by neither reducing texts nor trying to make it fit in "just one screen". We're managing complex concepts and users deserve all the help we can provide: from a fluid layout now that we're delivering a Web product to not hiding explanations when they are needed and not hindering those who do not need them. Storage Proposal is a very good example of a complex part of Agama that, I believe, is now in a shape that doesn't scare and guides the user. What is more, it fits now in the @ancorgs screen 😝 but _falling back_ to an scrollable UI when the user has a smaller screen, more complex setup, or just using a higher zoom level.


I forget to add a note about _wasting space_. Please, do not see explanations, helps, spaces and whatever other resource as a waste of space :pleading_face: I really don't think that there is too much _wasted space_ currently in Agama. As said multiple times already, I might be wrong but instead I think that such feeling comes from keeping the _mental model_ that we had in YaST/libyui where saving space all the time is/was something crucial. That's not the case anymore. Of course, we should not waste space just because _it's the Web, it's fluid_, but in the same way we have to try not to confuse a resource that will help the user in one way or another with such a thing.

dgdavid:
Ey @balsa-asanovic :wave: 

Although I actually do not know how further you are following the Agama development, I'm pinging you here for two reasons:

- To kindly ask you to check out to the new Storage page.
- To let you know we're speeding-up the activation of type-checking in the front-end because it helps to deliver a more quality code. This and https://github.com/openSUSE/agama/pull/1141 are the latest PR were we have put more effort on it. I would appreciate it if you could take a look. As you can see, we're [relying on JSDoc](https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax) to do so, although it is requires quite more keystokes compared to just using TypeScript which we're not using (yet?) due to historical reasons. Just out of curiosity, would TypeScript be a hindrance to your contributions? Or would be, instead, an encouragement?

ancorgs (approved):
Better than on Friday. LGTM.

balsa-asanovic:
> Ey @balsa-asanovic 👋
> 
> Although I actually do not know how further you are following the Agama development, I'm pinging you here for two reasons:
> 
>     * To kindly ask you to check out to the new Storage page.
> 
>     * To let you know we're speeding-up the activation of type-checking in the front-end because it helps to deliver a more quality code. This and [web: Allow changing file system location #1141](https://github.com/openSUSE/agama/pull/1141) are the latest PR in which we have put more effort on it. I would appreciate it if you could take a look. As you can see, we're [relying on JSDoc](https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax) to do so, although it requires quite more keystokes compared to just using TypeScript which we're not using (yet?) due to historical reasons. Just out of curiosity, would TypeScript be a hindrance to your contributions? Or would be an encouragement instead?

Hey @dgdavid,

I am still following the project, so thank you for the ping.

I've just checked out the new Storage page and compared it a bit with the old one.
I definitely prefer the new one.
Somehow the design of the sections and their separation is done better.
It is easier to understand what is being done and what to do.

As for the type-checking, kudos for pushing it more. As you know I've worked on a couple of PRs in which the usage of JSDoc was present, so this does not come as a surprise to me.
JSDoc, as you say, seems to be a more verbose choice compared to Typescript, but gives the advantage of not needing to migrate the whole front-end code to Typescript, by merely introducing JSDoc descriptions in the vanilla JavaScript code.

> Just out of curiosity, would TypeScript be a hindrance to your contributions? Or would be an encouragement instead?

Definitely not a hindrance. But also, honestly, I wouldn't give it much advantage over the current situation. (at least for me :smile: )

dgdavid:
@balsa-asanovic, thanks for the feedback.

PR #1139, opened 2024-04-09
unify interface of http helpers

jreidinger:
## Problem

Http helpers has inconsistent API.

## Solution

Unify it on standard js Response object.


web/src/client/http.js
```diff
@@ -95,21 +95,21 @@ class HTTPClient {
 
   /**
    * @param {string} url - Endpoint URL (e.g., "/l10n/config").
-   * @return {Promise<object>} Server response.
+   * @return {Promise<Response>} Server response.
```
dgdavid, line 98: Honest question, what is `Response`? https://developer.mozilla.org/en-US/docs/Web/API/Response, https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.response.html, or anything else?

dgdavid: Nevermind, I see it is the first one, which is the what https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API returns.

web/src/client/users.js
```diff
@@ -64,13 +64,12 @@ class UsersBaseClient {
    * @return {Promise<User>}
    */
   async getUser() {
-    const user = await this.client.get("/users/first");
-
-    if (user === null) {
+    const response = await this.client.get("/users/first");
+    if (!response.ok) {
+      console.log("Failed to get first user config: ", response);
       return { fullName: "", userName: "", password: "", autologin: false, data: {} };
```
lslezak, line 70: Later we should probably improve the error handling. This basically silently hides the error, users usually do no check the console logs.

The problem is when some of the defaults gets back to the server as a changed value. This can potentially loose some settings on the server accidentally if there is some temporary network problem (i.e. the GET call fails but POST later succeeds).

I think there should be some automatic retry with several attempts. If it still fails we should tell the user and probably do not allow to continue, or offer a full page reload to get into the consistent state...

web/src/client/users.js
```diff
@@ -64,13 +64,12 @@ class UsersBaseClient {
    * @return {Promise<User>}
    */
   async getUser() {
-    const user = await this.client.get("/users/first");
-
-    if (user === null) {
+    const response = await this.client.get("/users/first");
+    if (!response.ok) {
+      console.log("Failed to get first user config: ", response);
```
lslezak, line 69: I think `console.error` would be better here...

lslezak (commented):
Just some minor comments

web/src/client/users.js
jreidinger: yeah, error reporting is something we can definitively improve.

web/src/client/users.js
jreidinger: agreed

web/src/client/l10n.js
```diff
@@ -89,7 +89,7 @@ class L10nClient {
   async getUIKeymap() {
     const response = await this.client.get("/l10n/config");
     if (!response.ok) {
-      console.log("Failed to get localizaation config: ", response);
+      console.error("Failed to get localizaation config: ", response);
```
imobachgs, line 92: typo: `localization` instead of `localizaation`.

imobachgs (commented):
Just a typo. Otherwise, LGTM.

PR #1141, opened 2024-04-09
web: Allow changing file system location

joseivanlopez:
By default, all the file systems are created as partitions at the disk selected for installation. This PR adds an optiton for changing the location of a file system, allowing to select a target disk or to follow the installation device. Moreover, it offers an option to create a dedicated LVM volume group. For more details see  https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md#file-systems.

Bonus: This PR introduces quite some changes to enable type checking for several components.

## Screenshots

localhost_8080_ (37)

### DIFF
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -111,7 +111,7 @@ const ZFCP_DISK_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Disk";
  * @property {boolean} delete
  *
  * @typedef {object} ProposalSettings
- * @property {string} target
+ * @property {ProposalTarget} target
  * @property {string} [targetDevice]
  * @property {string[]} targetPVDevices
  * @property {boolean} configureBoot
@@ -124,13 +124,15 @@ const ZFCP_DISK_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Disk";
  * @property {Volume[]} volumes
  * @property {StorageDevice[]} installationDevices
  *
+ * @typedef {keyof ProposalTargets} ProposalTarget
+ *
  * @typedef {object} SpaceAction
  * @property {string} device
  * @property {string} action
  *
  * @typedef {object} Volume
  * @property {string} mountPath
- * @property {string} target
+ * @property {VolumeTarget} target
  * @property {StorageDevice} [targetDevice]
  * @property {string} fsType
  * @property {number} minSize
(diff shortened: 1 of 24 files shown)


web/src/components/storage/VolumeLocationDialog.jsx
```diff
@@ -0,0 +1,184 @@
...
+            <RadioOption id={LOCATION_AUTO_ID} defaultChecked={isLocationAuto} onChange={selectAutoOption}>
+              {_("Automatic")}
+            </RadioOption>
+          </legend>
+          <div>
+            {autoText()}
+          </div>
+        </fieldset>
+
+        <fieldset className="stack">
```
dgdavid, line 148: I haven't tested it yet, but looks like the (Firefox) problem described at https://github.com/openSUSE/agama/pull/354 has been solved (see https://bugzilla.mozilla.org/show_bug.cgi?id=1829874)

So, If it's the case, we could disable the fieldset instead of going field by field. Should work natively.

web/src/client/storage.js
```diff
@@ -140,15 +144,44 @@ const ZFCP_DISK_IFACE = "org.opensuse.Agama.Storage1.ZFCP.Disk";
...
+  DISK: "disk",
+  NEW_LVM_VG: "newLvmVg",
+  REUSED_LVM_VG: "reusedLvmVg"
+});
+
+/**
+ * Enum for the possible volume targets.
+ *
+ * @readonly
+ */
```
dgdavid, line 176: NP: talk about this

web/src/components/storage/ProposalPage.jsx
```diff
@@ -216,6 +216,13 @@ export default function ProposalPage() {
     calculate(newSettings).catch(console.error);
   };
 
+  /**
+   * @todo Enable type checking and ensure the components are called with the correct props.
+   *
+   * @note The default value for `settings` should be `undefined` instead of an empty object, and
```
dgdavid, line 222: NP: if so, we should keep the "?." operator.n

web/src/components/storage/ProposalPage.test.jsx
```diff
@@ -76,32 +91,84 @@ const vdb = {
...
+    staging: {
+      getDevices: jest.fn().mockResolvedValue([vda])
+    },
+    getErrors: jest.fn().mockResolvedValue([]),
+    isDeprecated: jest.fn().mockResolvedValue(false),
+    onDeprecate: jest.fn(),
+    onStatusChange: jest.fn()
+  };
+
+  // @ts-expect-error
```
dgdavid, line 171: NP: adds a hint why a ts error is expected

web/src/components/storage/ProposalVolumes.jsx
```diff
@@ -370,53 +431,43 @@ const VolumesTable = ({ volumes, options, isLoading, onVolumesChange }) => {
...
-          onVolumesChange={onVolumesChange}
-        />
+        {renderVolumes()}
       </Tbody>
     </Table>
   );
 };
 
 /**
+ * @todo This component should be restructured to use the same approach as other newer components:
```
dgdavid, line 441: 👍

dgdavid (approved):
Thanks @joseivanlopez!

Code wise it looks good to me. You can go ahead and merfe it is nobody opposes.

I have, however, not blocking doubts I'd like to discuss in a short meeting.

BTW, excuse me for not noticing in the sync calls you were adding type checking in files that will collide (a lot?) with #1138 . And big thank you for joining to the adventure of enabliy the type checking. I really believe it will be a good thing for Agama.

web/src/components/storage/ProposalPage.jsx
joseivanlopez: Right now it cannot be undefined. If we initialize settings to undefined, then I would change the rest of the code a bit instead of having `.?` everywhere.

web/src/components/storage/ProposalPage.test.jsx
dgdavid: Thanks a lot for adding it. I found a way to fix it by not expecting an error, see https://gist.github.com/dgdavid/99916d7571697b9b3b733d71e6191fd3

PR #1142, opened 2024-04-10
Fixed setting unlimited maximum partition size

lslezak:
## Problem

- https://github.com/openSUSE/agama/issues/1065
- The maximum partition size cannot be set to "unlimited", using empty value results in the default maximum size

**Example:** Changing the default 2GB max size is not possible, it is always restored back:
agama_max_size_broken

## Details

- The problem is that the "unlimited" value is represented using the `undefined` value in the web UI.
- But `undefined` cannot be sent over D-Bus because D-Bus does not have concept for `undefined`, `nil` or `NULL` values, Iin that case the `MaxSize` is simply missing the sent D-Bus data.
- When the D-Bus service creates a `Volume` object it first initializes it with the default data from the YAML product configuration file. Then it updates the attributes using the D-Bus data. Obviously when a value is missing in the D-Bus then it keeps the default value from the config. And that's the problem.

## Solution

- Set the "unlimited" value when the `MaxSize` attribute is missing in the D-Bus data.

## Questions

- The fix has a drawback, the current code allows sending only partial data set and ensures the defaults from the config file are used for missing values.
- This won't work any more for the maximum size as a missing value won't be considered as a "use the default" but as "use the unlimited size".
- Can this have some bad consequences?
- The code works, but is it safe in all scenarios?

## Alternative Solution

- If this fix is not correct then we should probably introduce a new key in the D-Bus data, something like `MaxSizeUnlimited` with `true` / `false` values which can be transferred via D-Bus...

## Testing

- Tested manually

Now it is possible to unset  the maximum size:
agama_max_size_fixed

### DIFF
--- a/service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
+++ b/service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
@@ -51,9 +51,19 @@ module Agama
             dbus_volume_issues.each { |i| logger.warn(i) }
 
             builder = Agama::Storage::VolumeTemplatesBuilder.new_from_config(config)
-            builder.for(dbus_volume["MountPath"] || "").tap do |target|
+            ret = builder.for(dbus_volume["MountPath"] || "").tap do |target|
               valid_dbus_properties.each { |p| conversion(target, p) }
             end
+
+            # The `undefined` value is used as a value for unlimited size in the
+            # UI but D-Bus cannot send `undefined`, `nil` or `NULL` values. In
+            # that case the value is missing in the D-Bus data. Override the
+            # default from the config when the max size is missing.
+            if ret.max_size && !ret.max_size.unlimited? && !dbus_volume.key?("MaxSize")
+              ret.max_size = Y2Storage::DiskSize.unlimited
+            end
+
+            ret
           end
 
         private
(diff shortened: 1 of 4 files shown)


joseivanlopez:
IMHO, we can assume unlimited max size if the value does not come from D-Bus, so I think the fix is ok.

service/lib/agama/dbus/storage/volume_conversion/from_dbus.rb
```diff
@@ -51,9 +51,19 @@ def convert
...
+
+            # The `undefined` value is used as a value for unlimited size in the
+            # UI but D-Bus cannot send `undefined`, `nil` or `NULL` values. In
+            # that case the value is missing in the D-Bus data. Override the
+            # default from the config when the max size is missing.
+            if ret.max_size && !ret.max_size.unlimited? && !dbus_volume.key?("MaxSize")
+              ret.max_size = Y2Storage::DiskSize.unlimited
+            end
+
+            ret
```
joseivanlopez, line 66: NP: I would simplify the code a bit:

```suggestion
            builder.for(dbus_volume["MountPath"] || "").tap do |target|
              valid_dbus_properties.each { |p| conversion(target, p) }
              target.max_size = Y2Storage::DiskSize.unlimited unless dbus_volume.key?("MaxSize")
            end
```

joseivanlopez (approved):
LGTM

PR #1143, opened 2024-04-11
Agama server logging

mchf:
## Problem

We want to have an insight into incoming requests and corresponding responses
- https://trello.com/c/erXAzTxp

## Solution

Configured [TraceLayer](https://docs.rs/tower-http/0.5.2/tower_http/trace/struct.TraceLayer.html) to write some basic info into journald

## Testing

- *Tested manually*

## Screenshots

agama-web-server_logs

### DIFF
--- a/rust/agama-server/src/web/service.rs
+++ b/rust/agama-server/src/web/service.rs
@@ -97,13 +97,16 @@ impl MainServiceBuilder {
             .route("/login", get(login_from_query))
             .route("/po.js", get(super::http::po))
             .nest("/api", api_router)
-            .layer(TraceLayer::new_for_http()
-                .on_request(|request: &Request<Body>, _span: &Span| {
-                    tracing::info!("request: {} {}", request.method(), request.uri().path())
-                })
-                .on_response(|response: &Response<Body>, latency: Duration, _span: &Span| {
-                    tracing::info!("response: {} {:?}", response.status(), latency)
-                })
+            .layer(
+                TraceLayer::new_for_http()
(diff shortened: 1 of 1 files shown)


mvidner (approved):
The code and the resulting screenshot look good, thanks :)

rust/agama-server/src/agama-web-server.rs
```diff
@@ -239,6 +239,8 @@ async fn start_server(address: String, service: Router, ssl_acceptor: SslAccepto
 
     pin_mut!(listener);
 
+    println!("Agama web server listening at: {}", address);
```
lslezak, line 242: I'd remove this. It might mess the console when starting automatically via systemd. The address is printed by the tracing call above.

imobachgs:
Code-wise looks good. However, I would have expected this change against the `architecture_2024` branch and without a changelog (we will create a big entry when we merge).

rust/agama-server/src/web/service.rs
```diff
@@ -1,18 +1,22 @@
 use super::http::{login, login_from_query, logout, session};
 use super::{auth::TokenClaims, config::ServiceConfig, state::ServiceState, EventsSender};
 use axum::{
+    body::Body,
     extract::Request,
     middleware,
     response::IntoResponse,
+    response::Response,
```
imobachgs, line 8: np: Why not merging both lines: `response::{IntoResponse, Response}`?

imobachgs (commented):
Now it looks better. Thanks!

imobachgs (approved):

PR #1144, opened 2024-04-11
Fix users JavaScript tests

jreidinger:
## Problem

After change to new architecture users client js test start failing

## Solution

Adapt it.

### DIFF
--- a/web/src/client/users.test.js
+++ b/web/src/client/users.test.js
@@ -64,21 +64,20 @@ jest.mock("./http", () => {
   };
 });
 
-const FirstUser = {
+const firstUser = {
   fullName: "Jane Doe",
   userName: "jane",
   password: "12345",
   autologin: false,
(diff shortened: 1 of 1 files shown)


web/src/client/users.test.js
```diff
@@ -21,71 +21,91 @@
...
+        get: mockGetFn,
+        put: mockPutFn,
+        delete: mockDeleteFn,
+        patch: mockPatchFn,
+      };
+    }),
+  };
 });
 
+const FirstUser = {
```
imobachgs, line 67: Please, use camelCase. For classes or React components it is right to use Pascal case, but not for regular variables and functions.

imobachgs (changes requested):
It looks almost good. Just adapt some variable names, please.

imobachgs (approved):

PR #1146, opened 2024-04-12
Adapt product registration to the new architecture

jreidinger:
Another change in set that switch web UI from using dbus to HTTP API. As part of this change the rust code is added to provide that HTTP API and also adapted web UI. It also fixes all relevant js tests.

Things discussed but postponed (due to size of change and time constraints ) in this change was:

1. move registration requirement from registration to product as it is highly coupled with it
2. provide all changed data in websocket signal, so client does not need to do another http call to get correct data

Matching trello card: https://trello.com/c/2bTZOnpB

### DIFF
--- a/rust/agama-lib/src/product/client.rs
+++ b/rust/agama-lib/src/product/client.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 
 use crate::error::ServiceError;
 use crate::software::proxies::SoftwareProductProxy;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use zbus::Connection;
 
 use super::proxies::RegistrationProxy;
@@ -18,6 +18,7 @@ pub struct Product {
     pub description: String,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema)]
 pub enum RegistrationRequirement {
     /// Product does not require registration
     NotRequired = 0,
@@ -32,9 +33,15 @@ impl TryFrom<u32> for RegistrationRequirement {
 
     fn try_from(v: u32) -> Result<Self, Self::Error> {
         match v {
-            x if x == RegistrationRequirement::NotRequired as u32 => Ok(RegistrationRequirement::NotRequired),
-            x if x == RegistrationRequirement::Optional as u32 => Ok(RegistrationRequirement::Optional),
-            x if x == RegistrationRequirement::Mandatory as u32 => Ok(RegistrationRequirement::Mandatory),
+            x if x == RegistrationRequirement::NotRequired as u32 => {
+                Ok(RegistrationRequirement::NotRequired)
+            }
+            x if x == RegistrationRequirement::Optional as u32 => {
+                Ok(RegistrationRequirement::Optional)
+            }
+            x if x == RegistrationRequirement::Mandatory as u32 => {
+                Ok(RegistrationRequirement::Mandatory)
+            }
             _ => Err(()),
         }
     }
--- a/rust/agama-lib/src/product/client.rs
+++ b/rust/agama-lib/src/product/client.rs
@@ -24,7 +24,7 @@ pub enum RegistrationRequirement {
     NotRequired = 0,
     /// Product has optional registration
     Optional = 1,
-    /// It is mandatory to register product
+    /// It is mandatory to register the product
     Mandatory = 2,
 }
 
--- a/rust/agama-lib/src/product/client.rs
+++ b/rust/agama-lib/src/product/client.rs
@@ -18,6 +18,28 @@ pub struct Product {
     pub description: String,
 }
 
+pub enum RegistrationRequirement {
+    /// Product does not require registration
+    NotRequired = 0,
+    /// Product has optional registration
+    Optional = 1,
+    /// It is mandatory to register product
+    Mandatory = 2,
+}
+
+impl TryFrom<u32> for RegistrationRequirement {
+    type Error = ();
+
+    fn try_from(v: u32) -> Result<Self, Self::Error> {
+        match v {
+            x if x == RegistrationRequirement::NotRequired as u32 => Ok(RegistrationRequirement::NotRequired),
+            x if x == RegistrationRequirement::Optional as u32 => Ok(RegistrationRequirement::Optional),
+            x if x == RegistrationRequirement::Mandatory as u32 => Ok(RegistrationRequirement::Mandatory),
+            _ => Err(()),
+        }
+    }
+}
+
 /// D-Bus client for the software service
 #[derive(Clone)]
 pub struct ProductClient<'a> {
@@ -86,6 +108,13 @@ impl<'a> ProductClient<'a> {
         Ok(self.registration_proxy.email().await?)
     }
 
+    pub async fn registration_requirement(&self) -> Result<RegistrationRequirement, ServiceError> {
+        let requirement = self.registration_proxy.requirement().await?;
+        // unknown number can happen only if we do programmer mistake
+        let result: RegistrationRequirement = requirement.try_into().unwrap();
+        Ok(result)
+    }
+
     /// register product
     pub async fn register(&self, code: &str, email: &str) -> Result<(u32, String), ServiceError> {
         let mut options: HashMap<&str, zbus::zvariant::Value> = HashMap::new();
@@ -93,5 +122,10 @@ impl<'a> ProductClient<'a> {
             options.insert("Email", zbus::zvariant::Value::new(email));
         }
         Ok(self.registration_proxy.register(code, options).await?)
+    }
+
+    /// de-register product
+    pub async fn deregister(&self) -> Result<(u32, String), ServiceError> {
+        Ok(self.registration_proxy.deregister().await?)
     }
 }
--- a/rust/agama-server/src/software/web.rs
+++ b/rust/agama-server/src/software/web.rs
@@ -8,13 +8,13 @@
 use crate::{
     error::Error,
     web::{
-        common::{issues_router, progress_router, service_status_router},
+        common::{issues_router, progress_router, service_status_router, Streams},
         Event,
     },
 };
 use agama_lib::{
     error::ServiceError,
-    product::{Product, ProductClient, RegistrationRequirement},
+    product::{Product, ProductClient, RegistrationRequirement, proxies::RegistrationProxy},
     software::{
         proxies::{Software1Proxy, SoftwareProductProxy},
         Pattern, SelectedBy, SoftwareClient, UnknownSelectedBy,
@@ -26,7 +26,7 @@ use axum::{
     Json, Router,
 };
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, pin::Pin};
+use std::collections::HashMap;
 use tokio_stream::{Stream, StreamExt};
 
 #[derive(Clone)]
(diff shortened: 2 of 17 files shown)


rust/agama-server/src/software/web.rs
```diff
@@ -49,14 +49,33 @@ pub struct SoftwareConfig {
...
+            "patterns_changed",
+            Box::pin(patterns_changed_stream(dbus.clone()).await?),
+        ),
+        (
+            "product_changed",
+            Box::pin(product_changed_stream(dbus.clone()).await?),
+        ),
+        (
+            "registration_requirement_changed",
+            Box::pin(registration_requirement_changed_stream(dbus.clone()).await?),
```
imobachgs, line 66: If we move the registration requirement to the product itself (after all, that's part of the product, right?), we do not need this signal.

rust/agama-server/src/software/web.rs
```diff
@@ -99,6 +118,57 @@ async fn patterns_changed_stream(
...
+async fn registration_email_changed_stream(dbus: zbus::Connection,
+) -> Result<impl Stream<Item = Event>, Error> {
+    let proxy = RegistrationProxy::new(&dbus).await?;
+    let stream = proxy
+        .receive_email_changed()
+        .await
+        .then(|change| async move {
+            if let Ok(_id) = change.get().await {
+                // unwrap is safe as possible numbers is send by our controlled dbus
+                return Some(Event::RegistrationChanged);
```
imobachgs, line 147: So we need to ask for the registration information again, right? Perhaps it would be better to send the registration information along with the event.

rust/agama-lib/src/product/client.rs
```diff
@@ -86,6 +115,13 @@ impl<'a> ProductClient<'a> {
         Ok(self.registration_proxy.email().await?)
     }
 
+    pub async fn registration_requirement(&self) -> Result<RegistrationRequirement, ServiceError> {
+        let requirement = self.registration_proxy.requirement().await?;
+        // unknown number can happen only if we do programmer mistake
+        let result: RegistrationRequirement = requirement.try_into().unwrap();
```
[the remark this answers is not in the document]
jreidinger: problem with `?` is  that it is not ServiceError, so I will need to convert that error somehow ( doable as I own error ), but I think it is not worth effort.

rust/agama-server/src/software/web.rs
jreidinger: well, it is possible, I will just need to recreate info again from proxy as I cannot get it from signal as it contain just one param and not all of them.

rust/agama-server/src/software/web.rs
jreidinger: that is right, but value is computed in registration interface and not product one (on dbus side).

imobachgs: JFTR, we agreed on postponing this change.

rust/agama-server/src/software/web.rs
imobachgs: We already solve a problem like this with the [ProgressStream](https://github.com/openSUSE/agama/blob/architecture_2024/rust/agama-server/src/web/common.rs#L173-L220) API, that includes the full struct in the events. We can folllow a similar approach here.

However, we decided to keep it simple by now, but we have plans to improve it later.

jreidinger:
I get this warning in FF using arch branch. @lslezak @imobachgs does it ring any bells to you?
```
Cookie “agamaToken” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```

imobachgs:
> I get this warning in FF using arch branch. @lslezak @imobachgs does it ring any bells to you?
> 
> ```
> Cookie “agamaToken” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
> ```

I would say we do not need to send the cookies in third-party contexts. We could easily set the `SameSite` value in the [atuh_cookie_from_token function](https://github.com/openSUSE/agama/blob/architecture_2024/rust/agama-server/src/web/http.rs#L132).

rust/agama-lib/src/product/client.rs
```diff
@@ -18,6 +18,35 @@ pub struct Product {
     pub description: String,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema)]
+pub enum RegistrationRequirement {
+    /// Product does not require registration
+    NotRequired = 0,
+    /// Product has optional registration
+    Optional = 1,
+    /// It is mandatory to register product
```
imobachgs, line 27:
```suggestion
    /// It is mandatory to register the product
```

rust/agama-server/src/software/web.rs
```diff
@@ -99,6 +116,62 @@ async fn patterns_changed_stream(
     Ok(stream)
 }
 
+async fn registration_requirement_changed_stream(
+    dbus: zbus::Connection,
+) -> Result<impl Stream<Item = Event>, Error> {
+    // TODO: move registration requirement to product in dbus and so just one event will be needed.
```
imobachgs, line 122: As we agreed, I will create a card for this.

rust/agama-server/src/software/web.rs
```diff
@@ -153,6 +230,74 @@ async fn products(State(state): State<SoftwareState<'_>>) -> Result<Json<Vec<Pro
...
+    /// Registration key.
+    key: String,
+    /// Registration email.
+    email: String,
+}
+
+/// Register product
+///
+/// * `state`: service state.
+#[utoipa::path(post, path = "/software/registration", responses(
```
imobachgs, line 277: I would most probably use PUT, as you are replacing (or creating) the resource.

rust/agama-server/src/software/web.rs
```diff
@@ -153,6 +230,74 @@ async fn products(State(state): State<SoftwareState<'_>>) -> Result<Json<Vec<Pro
     Ok(Json(products))
 }
 
+/// Information about registration configuration (product, patterns, etc.).
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct RegistrationInfo {
+    /// Registration key. Empty value mean key not used or not registered.
+    key: String,
```
imobachgs, line 238: Why not use `Option<String>`?

rust/agama-server/src/software/web.rs
```diff
@@ -153,6 +230,74 @@ async fn products(State(state): State<SoftwareState<'_>>) -> Result<Json<Vec<Pro
...
+#[utoipa::path(post, path = "/software/registration", responses(
+    (status = 200, description = "registration successfull"),
+    (status = 400, description = "The D-Bus service could not perform the action")
+))]
+async fn register(
+    State(state): State<SoftwareState<'_>>,
+    Json(config): Json<RegistrationParams>,
+) -> Result<Json<()>, Error> {
+    state.product.register(&config.key, &config.email).await?;
+    Ok(Json(()))
```
imobachgs, line 286: Following your own suggestion, I would return `Ok(StatusCode::NO_CONTENT)` (you need to change the return type to `Result<impl IntoResponse, Error>`). We avoid then returning a useless `null`. The same for the `deregister` method below.

rust/agama-server/src/web/common.rs
```diff
@@ -17,6 +17,8 @@ use crate::error::Error;
 
 use super::Event;
 
+pub type Streams = Vec<(&'static str, Pin<Box<dyn Stream<Item = Event> + Send>>)>;
```
imobachgs, line 20: I like this but, what about using `EventsStream` as a more descriptive name?

web/src/Main.jsx
```diff
@@ -21,17 +21,15 @@
...
-  // return (
-  //   <>
-  //     <Questions />
-  //     <Outlet />
-  //   </>
-  // );
-  return <Outlet />;
+  return (
+    <>
+      <Questions />
```
imobachgs, line 29: :+1:

web/src/client/software.js
```diff
@@ -87,106 +85,6 @@ const SelectedBy = Object.freeze({
  * @property {string} icon - Icon name (not path or file name!).
  */
 
-/**
- * Product manager.
- * @ignore
- */
-class BaseProductManager {
```
imobachgs, line 94: :+1:

web/src/client/software.js
```diff
@@ -348,6 +246,73 @@ class ProductBaseClient {
...
+
+  /**
+   * Returns the registration of the selected product.
+   *
+   * @return {Promise<Registration>}
+   */
+  async getRegistration() {
+    const response = await this.client.get("/software/registration");
+    if (!response.ok) {
+      console.log("Failed to get registration config: ", response);
```
imobachgs, line 258: No need to add a whitespace there, `console.log` takes care.

```suggestion
      console.log("Failed to get registration config:", response);
```

web/src/client/software.js
```diff
@@ -348,6 +246,73 @@ class ProductBaseClient {
...
+      return { requirement: "unknown", code: null, email: null };
+    }
+    const config = await response.json();
+
+    const requirement = config.requirement;
```
imobachgs, line 263: You can use destructuring to get the date like this:

```
    const { requirement, email, code } = config;
```

web/src/client/software.js
```diff
@@ -348,6 +246,73 @@ class ProductBaseClient {
...
+   * @param {string} code
+   * @param {string} [email]
+   * @returns {Promise<ActionResult>}
+   */
+  async register(code, email = "") {
+    const response = await this.client.post("/software/registration", { key: code, email });
+
+    return {
+      success: response.ok,
+      message: "", // TODO: pass message and code in body
```
imobachgs, line 286: Why? Any chance we can fix this before merging?

web/src/client/software.js
```diff
@@ -348,6 +246,73 @@ class ProductBaseClient {
...
+   * Tries to deregister the selected product.
+   *
+   * @returns {Promise<ActionResult>}
+   */
+  async deregister() {
+    const response = await this.client.delete("/software/registration");
+
+    return {
+      success: response.ok,
+      message: "", // TODO: how to get message?
```
imobachgs, line 300: What does it mean?

imobachgs (commented):

rust/agama-server/src/software/web.rs
jreidinger: well, if I use PUT, I will have to create it idempotent, which it is currently not.

rust/agama-server/src/software/web.rs
jreidinger: I am considering it, but in the end I use same values as in dbus API. Reason is mainly to prevent accidental sending just email, that can mean that system is not registered.

rust/agama-server/src/software/web.rs
jreidinger: well, for me reason is that POST has mandatory response body according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST

And that reminds me that what is still missing from old API is more detailed error codes. I am not sure if we can have something like `Ok(Json(scc_error))`?

rust/agama-server/src/web/common.rs
jreidinger: in that case I would use `EventStreams` as it is not just single stream, but array of streams.

web/src/client/software.js
jreidinger: you are right, as conversion for requirement is no longer there, it is not needed

web/src/client/software.js
jreidinger: well, it calls delete. And if it failed we need to somehow get error code. I am not sure if way would be using body in Delete and 200 code?

rust/agama-server/src/software/web.rs
imobachgs: Well, I am not sure. If you perform the same action with the same data several times, what changes? I would expect the result to be the same. But I am fine with using POST.

rust/agama-server/src/software/web.rs
imobachgs: I still prefer using `Option<T>` that the typical "empty means null" handling. In D-Bus we are far more limited. Of course, I am not blocking the merge because of this, just mentioning.

rust/agama-server/src/web/common.rs
imobachgs: Yes, you are right, that's even better.

rust/agama-server/src/software/web.rs
jreidinger: well, the change will be that after first successful registration the next one fails. Also we do not handle if you use different data. You have to first deregister and then can register with different data. We handle that automatic only when you change product.

web/src/client/software.js
imobachgs: Sure, we need the error code. My question was more about why is this a TODO? If we cannot do it right now, then it is fine.

About the answer, I would use "422 Unprocessable Content" (perhaps there is a better code) and put the error code in the body.

jreidinger:
JFYI testing of error response succeed:

registration_failure

imobachgs (approved):

PR #1147, opened 2024-04-13
Refactor the NetworkSystem API

imobachgs:
## The problem

While working in introducing an API to get the runtime configuration of the network, we decided that we were not happy with the current `NetworkSystem` API. There was quite some manual to do when using the API, like dealing with channels and having to explictly move the server to a separate Tokio task.

## The solution

This change splits the API into smaller pieces:

* `NetworkSystem` is still the main entry point.
* `NetworkSystemClient` wraps the `actions` channel and offers a higher-level API by hiding the message passing handling.
* `NetworkSystemServer` keeps the network state and dispatches the actions. Beware that this struct is private and the `NetworkSystemClient` should be the only way to communicate with it.
* 
You can get a better idea of how it looks like through the following example (included in the documentation):

```rust
let adapter = NetworkManagerAdapter::from_system()
    .await
    .expect("Could not connect to NetworkManager.");
let dbus = connection()
    .await
    .expect("Could not connect to Agama's D-Bus server.");
let network = NetworkSystem::new(dbus, adapter);

// Start the networking service and get the client for communication.
let actions = network.start()
    .await
    .expect("Could not start the networking configuration system.");

// Perform some action, like getting the list of devices.
let devices = client.get_devices().await.unwrap();
```

## To do

* [ ] Get rid of `NetworkService` as it is not needed anymore. Should we rename `NetworkSystem*`?.
* [x] Adapt the HTTP/JSON interface to the new API (partially done).

### DIFF
--- a/rust/agama-server/src/network/system.rs
+++ b/rust/agama-server/src/network/system.rs
@@ -1,6 +1,6 @@
 use super::{
     error::NetworkStateError,
-    model::{Device, StateConfig},
+    model::{AccessPoint, Device, StateConfig},
     NetworkAdapterError,
 };
 use crate::network::{
@@ -8,10 +8,7 @@ use crate::network::{
     model::{Connection, GeneralState},
     Action, Adapter, NetworkState,
 };
-use agama_lib::{
-    error::ServiceError,
-    network::{settings::NetworkConnection, types::DeviceType},
-};
+use agama_lib::{error::ServiceError, network::types::DeviceType};
 use std::{error::Error, sync::Arc};
 use tokio::sync::{
     mpsc::{self, error::SendError, UnboundedReceiver, UnboundedSender},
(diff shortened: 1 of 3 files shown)


rust/agama-server/src/network/system.rs
```diff
@@ -1,67 +1,223 @@
...
+/// # tokio_test::block_on(async {
+/// let adapter = NetworkManagerAdapter::from_system()
+///     .await
+///     .expect("Could not connect to NetworkManager.");
+/// let dbus = connection()
+///     .await
+///     .expect("Could not connect to Agama's D-Bus server.");
+/// let network = NetworkSystem::new(dbus, adapter);
+///
+/// // Start the networking service and get the channel for communication.
```
teclator, line 57:
```suggestion
/// // Start the networking service and get the client for communication.
```

teclator:
LGTM: about renaming.. for me makes sense

teclator (approved):
LGTM

PR #1150, opened 2024-04-15
[Experimental DRAFT] web: Uses the same look & feel in all selection dialogs

dgdavid:
This PR is just an experiment to see how further we can/desire go by unification the UI of "WhateverDialog" in storage page.

In general it makes everything looks better, with the exception of the complex DeviceSelectionDialog which still needing tweaks to make it worth it which proposed approach.

It also reduces a lot the complexity for both, the developer and the user.

PLEASE DO NOT MERGE. Open as much as needed new PRs and use this as source of inspiration.


dgdavid:
> In general it makes everything looks better, with the exception of the complex DeviceSelectionDialog which still needing tweaks to make it worth it which proposed approach.

Actually it can be a complex for the SpacePolicyDialog too, because these two dialog has something in common: the device selection table, which can be long enough to make the approach confusing.

dgdavid:
> Actually it can be a complex for the SpacePolicyDialog too, 

Unless we find a way to only show the selector when the user is using the "Custom" option. Is it really needed to show it in all other shortcuts? I'd say no, because we have the "Result" table in which the user is gonna see how the shortcut applies in a cleaner way than "crowding" these options with the potentially long table that is available in "Custom"

dgdavid:
Let's keep it close until we can continue with the idea.

PR #1152, opened 2024-04-16
web: Use meaningful icons in Storage proposal page

dgdavid:
During the presentation of #1138, it was seen that the _shadow_ icon chosen for some of the actions opening a dialog created confusion in some attendees when other no-iconized actions opened a dialog too.

Somehow, it was forgotten one of the principles followed in the Agama UI until now: an action (link, button, button dressed up as a link, menu item, etc) should be easily recognizable by users as an element they can interact with in the way they use to do it. I.e., it's not needed to identify if the action opens a dialog, navigates to another page, to another section of the same page, opens a menu, shows or hides a sidebar, etc, since the same action that today opens a dialog in the future might navigate to another page instead.

Proposed changes here fix the issue by replacing these icons with ones that represent the concept of the action or by none when it would hamper more than help. **Keep in mind that icons are merely cosmetic aids for sight people, reason why the meaning of the actions must not be determined by them**. The UI must still be meaningful if, for whatever reason, icons disappear.

Of course, the ones proposed here can be changed at any point when better suggestions arise. **But remember that at this moment we're mainly relying on https://fonts.google.com/icons**. We can change our mind in the future, but choosing a consistent library of icons instead of grabbing them from here and there. For example, [Lucid icons](https://lucide.dev/icons/) could be a good candidate for a library replacement, but in a future **PLEASE**.

| Before | After |
|-|-|
| Storage proposal page using shadow icon | Storage proposal page using better icons |

### DIFF
--- a/web/package/cockpit-agama.changes
+++ b/web/package/cockpit-agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Apr 16 11:15:13 UTC 2024 - David Diaz <dgonzalez@suse.com>
+
+- Use better icons in storage proposal page
+  (gh#openSUSE/agama#1152).
+
+-------------------------------------------------------------------
 Mon Apr 15 10:52:14 UTC 2024 - David Diaz <dgonzalez@suse.com>
 
 - Stop using `<abbr>` tag because "its exposure continues to be
--- a/web/src/components/core/Field.jsx
+++ b/web/src/components/core/Field.jsx
@@ -108,13 +108,12 @@ const SwitchField = ({ isChecked = false, highlightContent = false, ...props }) 
 };
 
 /**
- * @param {Omit<FieldProps, 'icon'> & {isExpanded: boolean}} props
+ * @param {FieldProps & {isExpanded: boolean}} props
  */
 const ExpandableField = ({ isExpanded, ...props }) => {
-  const iconName = isExpanded ? "collapse_all" : "expand_all";
   const className = isExpanded ? "expanded" : "collapsed";
 
-  return <Field {...props} icon={iconName} className={className} />;
+  return <Field {...props} className={className} />;
 };
 
 export default Field;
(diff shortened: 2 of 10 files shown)


joseivanlopez (approved):
LGTM

ancorgs:
I like all the changes

I'm just not sure about the icon for "Find Space". I admit it's a very challenging one, but I believe this one would fit better (absolutely subjective opinion):

tune

Or maybe also any of these other options:

others

I understand none of them is great, but the one on this pull request looks too much like a searching functionality... which "Find Space" is not.

dgdavid:
Yeah, it was suspicious to get the PR merged as quick as it was :wink:

To be honest, for this case I like none of your proposals. But, I do not like the magnifying glass neither. Moreover, I believe a I even do not like the "Find space" label neither, nor the final implementation we have in that dialog.

But, I just wanted to go ahead having in mind that icons should be just an (small, insignificant) aid. Without any intention of play down the importance they deserves, I wouldn't like to focus more than needed now in the tiny details.

That said, let me explain why I choosed that icon even when I'm not fully happy with it:

* If you search "scan", you can see that Qr and barcode scan icons are enclosed in the same square

  Screenshot from 2024-04-16 14-13-03



* I actually wanted to use "Flex Wrap" or "Flex No Wrap", which looks more meaningful to me since they represents empty and full spaces. But, I though adding one of them would raise a discussion :smiling_face_with_tear: 

  Screenshot from 2024-04-16 14-12-29

* What is more, I liked one icon from Lucid Icons, but I feel I should not invest my time right now in a proposal for switching from one library to another, since I'd like to think slowly about it before make the movement.
   Screenshot from 2024-04-16 14-13-45


But, please, do not hesitate on making a PR if you feel the "Tune" one will be less confusing for the users. You'll have my LGTM for getting ride of the current one :wink:

dgdavid:
> * I actually wanted to use "Flex Wrap" or "Flex No Wrap", which looks more meaningful to me since they represents empty and full spaces. But, I though adding one of them would raise a discussion

I also though on using the "Qr scan" one, but it's too tied to a widely known concept for using it.

PR #1153, opened 2024-04-16
web: Overrides few rules for PF/Table headers

dgdavid:
## Problem

Sometimes table headers with an space in their content are cut and uses the `text-overflow` value of CSS property, even when looking at it _seems_ that there is enough space. Actually, there isn't because of `margin`s `padding`s and so.

## Solution

To overrides the PF/Table styles for setting the `white-space` CSS property from `nowrap` to `normal`. 

Forcing a size per column it's not a solution because 

* it does not warrant that it's the right size for the same content in different languages, and
* it is preferred to let the browser laying out the table according to the available space (among others)

## Testing

Tested manually.

## Note for reviewers

Although it has a little visual impact, I do not believe this deserves an entry in the changelog. But let me know if you think it should be added.

## Screenshots

| Before | After |
|-|-|
| Screen Shot 2024-04-16 at 15 40 56 | Screen Shot 2024-04-16 at 15 32 32 |

### DIFF
--- a/web/src/assets/styles/blocks.scss
+++ b/web/src/assets/styles/blocks.scss
@@ -471,14 +471,7 @@ ul[data-type="agama/list"][role="grid"] {
 
 table[data-type="agama/tree-table"] {
   th:first-child {
-    block-size: fit-content;
     padding-inline-end: var(--spacer-normal);
-  }
-
-  th.fit-content {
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
LGTM

PR #1154, opened 2024-04-16
web: Add file systems for arbitrary mount points

joseivanlopez:
The storage UI only allows adding file systems for the mount points predefined by the product (typicaly */home*, *swap* or */var*). This PR extends the storage UI making possible to add file systems for any mount point.

Main changes:

* The *Add file system* button is now a menu button, which allows to select either a predefined mount point (if the mount point was not added yet) or an arbitrary one (the *Other* option).
* The *Add file system* button turns into a regular button if only the *Other* option is available.
* The *Add file system* button is disabled if the product does not admit arbitrary mount points (i.e., transactional system) and there is no pending predefined mount point to be added (i.e., all the predefined mount points were already added).
* The *Add file system* is hidden if the product does not admit optional file systems (i.e., transactinal system without predefined mount points).
* The file system form validates the mount point:
  * Validates mount point presence on accept.
  * Validates mount point format on accept.
  * Validates whether the entered mount point is already added and offers to edit such a file system.
  * Validates whether the entered mount point maches any of the predefined mount points and offers to add the predefined mount point.
 
Note:

This PR introduces a new validation pattern which was previoulsy discussed and agreed. The form validates its fields on accept and the field error is gone when the field value changes (independently on the field is now valid or not). The rationale of this behavior is:

* The form should not complain on the fly (while typing) because the user has not finished until clicking on accept. 
* If the field shows an error, then remove the error with the first change on its value. We can consider the user has started fixing the field, but the form will not bother the user untill pressing on accept again.  

In the future we could improve these validations. For example, a wrong field could show a "green check" if the user edit its value and the new value is correct. 

<details>
<summary>Screenshots</summary>

localhost_8080_ (40)

localhost_8080_ (43)

localhost_8080_ (42)

</details>


dgdavid:
@joseivanlopez, please check commit https://github.com/openSUSE/agama/pull/1154/commits/6c561e993bb2b4fcb0989d9e7e9d233ff08ee313

I've added a component there for not using an HTML `<label>` when a form control is not shown for a value. As you can see in the below screenshots, the result is visually the same but with a different markup, more aligned with [HTML spec](https://html.spec.whatwg.org/multipage/forms.html#categories)

What do you think?

| Before | After |
|-|-|
| label_markup | span_markup |

web/src/components/storage/PartitionsField.jsx
```diff
@@ -494,18 +461,21 @@ const VolumesTable = ({ volumes, devices, target, targetDevice, isLoading, onVol
...
     onVolumesChange(newVolumes);
   };
 
+  /** @type {(volume: Volume) => void} */
   const deleteVolume = (volume) => {
     const newVolumes = volumes.filter(v => v.mountPath !== volume.mountPath);
     onVolumesChange(newVolumes);
   };
 
+  /** @type {() => React.ReactElement[]|React.ReactElement} */
```
dgdavid, line 478: Wondering if [`ReactNode`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/67ed83d31b2c8c0381b964b0897fc2457b874d11/types/react/index.d.ts#L448-L489) fits better here, 

```suggestion
  /** @type {() => React.ReactNode} */
```

web/src/components/storage/PartitionsField.jsx
```diff
@@ -603,15 +637,69 @@ const Advanced = ({
...
+   * Whether to show the button for adding a volume.
+   * @type {() => boolean}
+   */
+  const showAddVolume = () => {
+    const hasOptionalVolumes = () => {
+      return templates.find(t => t.mountPath.length && !t.outline.required) !== undefined;
+    };
+
+    return !isTransactionalSystem(templates) || hasOptionalVolumes();
+  };
```
dgdavid, line 696: You already explained to me the reason why you tend to use functions for a better logic encapsulation. But still, I think that sometimes we're abusing a bit of their usage.

Here, I can see the benefit of having `hasOptionalVolumes` defined as an small function in `showAddVolume` function. It prevents to evaluate the `find` if it's actually not needed because is not a transactional system. But maybe `hasOptionalVolumes` also deserves a place in storage/utils and should recive templates as an argument instead of getting them from the scope, as `isTransactionalSysmte` alrady does. Thus, the `showAddVolume` would be not needed at all, and instead you can go for a `<If condition={!isTransactionalSystem(templates) || hasOptionalVolumes(templates)} ... />` below. Much clear and less indirection IMHO. Plus, developers can reuse `hasOptionalVolumes` when needed by taking a look at what we have in storage/utils, avoiding to create yet another function to do the same.

Anyway, I know that this is _just a matter of taste_.

web/src/components/storage/VolumeDialog.jsx
```diff
@@ -0,0 +1,743 @@
...
+    const volume = this.findVolume();
+    const path = this.mountPath === "/" ? "root" : this.mountPath;
+
+    return (
+      <div className="split">
+        <span>{sprintf(_("There is already a file system for %s."), path)}</span>
+        <Button variant="link" isInline onClick={() => onClick(volume)}>
+          {_("Do you want to edit it?")}
+        </Button>
+      </div>
```
dgdavid, line 277: Just a note: it LGTM for this iteration, but I plan to make "assisted errors" to look a bit different than "format errors" in a future. The former should have a less "red" (Not be part of FormValidationError) and more prominent look and feel. I would came with a proposal when time permits.

web/src/components/storage/VolumeDialog.jsx
```diff
@@ -0,0 +1,743 @@
...
+    const template = this.findTemplate();
+    const path = this.mountPath === "/" ? "root" : this.mountPath;
+
+    return (
+      <div className="split">
+        <span>{sprintf(_("There is a predefined file system for %s."), path)}</span>
+        <Button variant="link" isInline onClick={() => onClick(template)}>
+          {_("Do you want to add it?")}
+        </Button>
+      </div>
```
dgdavid, line 324: Same here, looks good for now, but the plan is to make it looks better.

web/src/components/storage/VolumeFields.jsx
```diff
@@ -461,271 +444,56 @@ const SIZE_OPTION_LABELS = Object.freeze({
...
+              />
+            );
+          })}
+        </div>
+
+        <div aria-live="polite" className="highlighted-live-region">
+          <If condition={sizeMethod === SIZE_METHODS.AUTO} then={<SizeAuto {...sizeWidgetProps} />} />
+          <If condition={sizeMethod === SIZE_METHODS.RANGE} then={<SizeRange {...sizeWidgetProps} />} />
+          <If condition={sizeMethod === SIZE_METHODS.MANUAL} then={<SizeManual {...sizeWidgetProps} />} />
+        </div>
```
dgdavid, line 493: Intentionally I marked tests for this as "todo" because I plan to change this approach a bit in order to improve the form. Out of scope of this PR.

web/src/components/storage/VolumeFields.jsx
```diff
@@ -157,9 +127,10 @@ const FsSelectOption = ({ fsOption }) => {
  * @param {string} props.id - Widget id.
  * @param {string} props.value - Currently selected file system.
  * @param {Volume} props.volume - The selected storage volume.
- * @param {onChangeFn} props.onChange - Callback for notifying input changes.
+ * @param {boolean} props.isDisabled
+ * @param {(data: object) => void} props.onChange - Callback for notifying input changes.
  */
-const FsSelect = ({ id, value, volume, onChange }) => {
+const FsSelect = ({ id, value, volume, isDisabled, onChange }) => {
```
dgdavid, line 133: Question: below, in the line 154 (former 174), the onChange callback sent `{ fsType: option, snapshots: false }`. Still `snapshots: false` being relevant here?

web/src/components/storage/VolumeFields.test.jsx
```diff
@@ -0,0 +1,195 @@
...
+    );
+    const group = screen.getByRole("radiogroup", { name: "Size" });
+    const rangeOption = within(group).getByRole("radio", { name: "Range" });
+    await user.click(rangeOption);
+    expect(callbackFn).toHaveBeenCalledWith({ sizeMethod: SIZE_METHODS.RANGE });
+  });
+
+  it.todo("test SizeAuto internal component");
+  it.todo("test SizeManual internal component");
+  it.todo("test SizeRange internal component");
```
dgdavid, line 145: As said in a comment, left as "todo" because these components are going to change in a short term.

dgdavid (approved):
It LGTM. I just added a few comments for the record.

web/src/components/storage/VolumeDialog.jsx
```diff
@@ -0,0 +1,743 @@
...
+  const isProductDefined = volume.outline.productDefined;
+  const mountPath = volume.mountPath === "/" ? "root" : volume.mountPath;
+
+  if (isNewVolume && isProductDefined) return sprintf(_("Add %s file system"), mountPath);
+  if (!isNewVolume && isProductDefined) return sprintf(_("Edit %s file system"), mountPath);
+
+  return isNewVolume ? _("Add file system") : _("Edit file system");
+};
+
+class MissingMountPathError {
```
lslezak, line 94: I'd expect that the `*Error` names are user for some exceptions. In this case the names should be rather `*Check` as the class does not represent an error but rather a check for error which might or might not happen in the end.

So rather use `MissingMountPathCheck`, `InvalidMountPathCheck`, ...

lslezak (commented):
Names?

web/src/components/storage/PartitionsField.jsx
joseivanlopez: It sound good. Feel free to move it.

web/src/components/storage/VolumeFields.jsx
joseivanlopez: I good say it is even wrong. Snapshots should be configured only by means of the snapshots field. Good catch!

web/src/components/storage/VolumeDialog.jsx
joseivanlopez: I am not sure if `*Check` is a good name. Those classes are helpers for both to check and to render the error. Maybe we could use something like `*Manager`, that is, `MissingMountPathManager`:

~~~
const errorManager = new MissingMountPathManager("");
errorManager.check() ? errorManager.render(); 
~~~

dgdavid:
BTW, not directly related to this PR since it actually comes from https://github.com/openSUSE/agama/pull/543 (https://github.com/openSUSE/agama/pull/543/commits/00a5eba602cea964b99079ea76df76a17a806e2e) and https://github.com/openSUSE/agama/pull/590 (https://github.com/openSUSE/agama/commit/94cc4e1f52cea3a4f51afb1678d7bc560877e394) , but we have 


https://github.com/openSUSE/agama/blob/1b4bb22d0bf63f4d2787ea72dcef39be36e0989e/web/src/components/storage/VolumeFields.jsx#L262-L273


which make me think [again](https://github.com/openSUSE/agama/blob/da1840ab50bb2db639bc90738c4a3702e3c7d65c/web/src/components/storage/ProposalResultSection.jsx#L60-L62) that we should start using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat instead.

What do you think?

lslezak:
> which make me think [again](https://github.com/openSUSE/agama/blob/da1840ab50bb2db639bc90738c4a3702e3c7d65c/web/src/components/storage/ProposalResultSection.jsx#L60-L62) that we should start using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat instead.
> 
> What do you think?

Yes, I think it is a good idea. I didn't know about these functions when I implemented this.

web/src/components/storage/index.js
```diff
@@ -31,7 +31,8 @@ export { default as DASDFormatProgress } from "./DASDFormatProgress";
 export { default as ZFCPPage } from "./ZFCPPage";
 export { default as ZFCPDiskForm } from "./ZFCPDiskForm";
 export { default as ISCSIPage } from "./ISCSIPage";
-export { DeviceList, DeviceSelector, DeviceContentInfo, DeviceExtendedInfo, FilesystemLabel } from "./device-utils";
+export { DeviceContentInfo, DeviceExtendedInfo, FilesystemLabel } from "./device-utils";
+export { default as VolumeForm } from "./VolumeForm";
```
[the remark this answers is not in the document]
dgdavid: Rigth! A mistake from my side while merging master for solving a conflict. Good you catch it.

web/src/components/storage/PartitionsField.jsx
dgdavid: The PR is already merged. Let's move it to utils in the next opportunity. Making a note in my todo list.

web/src/components/storage/VolumeDialog.jsx
```diff
@@ -91,6 +91,19 @@ const renderTitle = (volume, volumes) => {
...
+ *  represent an error but a helper to check and render an error. It would be a better approach to
+ *  have something like a volume checker which generates errors:
+ *
+ *  For example:
+ *
+ *  const checker = new VolumeChecker(volume, volumes, templates);
+ *  const error = checker.existingMountPathError();
+ *  const message = error?.render(onClick);
+*/
+
```
dgdavid, line 106: :+1:

PR #1155, opened 2024-04-17
Include the encryption status in the device description

lslezak:
## Problem

- The device summary does not include the  encryption status in the device labels

## Solution

- Pass a new flag to `yast2-libstorage-ng` so the encryption status is included in the labels
- See https://github.com/yast/yast-storage-ng/pull/1379

## Testing

- Tested manually


joseivanlopez (approved):
LGTM

PR #1156, opened 2024-04-19
web: Allow opening popups with "stable" size

dgdavid:
Although the (ab)use of popup/modal dialog has been avoided as much as possible, they still being a valid resource for some use cases. However, they can become a bit annoying when holding _dynamic_ content and users interactions makes them change their size. 

To bypass such a problem, two props has been added to the Agama core/Popup component: `blockSize` and `inlineSize`. In a first iteration, both accepts "auto", "small", "medium", and "large" values, setting a _stable size_ on the corresponding axis that it's calculated with respect to the size of viewport (except when using "auto", which actually means _no stable size_).

As a side effect, in order to have more control on the sizing, the PF/Modal `variant` prop is being ignored.

## Notes for reviewers

* It makes no sense to add screenshots here. Please, jump into the branch and test it manually if you don't mind.
* On purpose, this PR **does not** enable type checking on touched files since changes are more related to CSS than JavaScript.
* Again, I believe this change does not deserve an entry in the changelog but I happily can add one if you think it is actually needed.

### DIFF
--- a/web/src/components/storage/SpacePolicyDialog.jsx
+++ b/web/src/components/storage/SpacePolicyDialog.jsx
@@ -149,7 +149,8 @@ in the devices listed below. Choose how to do it.");
       title={_("Find space")}
       description={description}
       isOpen={isOpen}
-      variant="medium"
+      blockSize="medium"
+      inlineSize="large"
       {...props}
     >
       <Form id="space-policy-form" onSubmit={onSubmit}>
--- a/web/src/components/storage/SpacePolicyDialog.jsx
+++ b/web/src/components/storage/SpacePolicyDialog.jsx
@@ -149,7 +149,7 @@ in the devices listed below. Choose how to do it.");
       title={_("Find space")}
       description={description}
       isOpen={isOpen}
-      blockSize="medium"
+      blockSize="large"
       inlineSize="large"
       {...props}
     >
--- a/web/src/assets/styles/app.scss
+++ b/web/src/assets/styles/app.scss
@@ -55,20 +55,6 @@
   }
 }
 
-.auto-modal-popup {
-  block-size: auto;
-}
-
-.medium-modal-popup {
-  min-block-size: 55vh;
-  max-block-size: 55vh;
-}
-
-.large-modal-popup {
-  min-block-size: 85vh;
-  max-block-size: 85vh;
-}
-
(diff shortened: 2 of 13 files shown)


imobachgs (approved):

imobachgs (approved):

web/src/components/storage/SpacePolicyDialog.jsx
```diff
@@ -149,7 +149,8 @@ in the devices listed below. Choose how to do it.");
       title={_("Find space")}
       description={description}
       isOpen={isOpen}
-      variant="medium"
+      blockSize="medium"
```
ancorgs, line 152: I would use "large" for the blockSize here. In my system only two lines of the table are visible which is not enough to make sense out of the table.

not-enough

Sometimes there are several disks involved at the installation. Or maybe just one disk but with many partitions (bear in mind the list of partitions is expanded automatically when clicking on "custom"). In all those cases, a lot of information is not visible at first sight which makes it harder to understand how to use the whole thing.

If you think "large" is too much, maybe we can use an approach similar to the one in the list of actions, deciding the size based on the number of disks and partitions.

ancorgs:
I miss the corresponding change in the pop-up to create/edit a storage volume (which is actually one of the most annoying cases). I guess that's because we are reworking that part and you expect the size of that form to become stable. But I wanted to mention it in case it was an overlook.

not-fixed

dgdavid:
> I miss the corresponding change in the pop-up to create/edit a storage volume (which is actually one of the most annoying cases). I guess that's because we are reworking that part and you expect the size of that form to become stable. But I wanted to mention it in case it was an overlook.
> 

Exactly, I didn't want to touch it here. I should have mentioned it. Thanks for the reminder.

web/src/components/storage/SpacePolicyDialog.jsx
dgdavid: > If you think "large" is too much, maybe we can use an approach similar to the one in the list of actions, deciding the size based on the number of disks and partitions.

I prefer this option, but bearing in mind that it is not so straightforward and I expect more changes in this dialog in the short/mid term, let's go for `blockSize="large"` as you suggested.

Good catch, BTW.

ancorgs (approved):
LGTM

PR #1160, opened 2024-04-22
Add Live ISO sources to Git

lslezak:
## Problem

The Live ISO sources are not maintained at GitHub, that makes the development complicated:

- Cannot track the changes easily, working with history in OBS is limited and more complicated than with Git
- Changing files in tarballs is difficult, you have to unpack it, modify a file and then manually compress back with correct parameters
- Building the Live ISO locally is not trivial

## Solution

- Move the sources into the `/live` subdirectory
- Add GitHub action for automatic submission from `master` to the [systemsmanagement:Agama:Staging](https://build.opensuse.org/project/show/systemsmanagement:Agama:Staging) OBS project
- Allow running the ISO build locally via `make build`

## Changes

The individual files were not touched, I just copied them from the current OBS project. I only moved the `fw_cleanup.rb` script to the `/tmp` directory as the script is temporary and it is deleted at the end of the ISO build.

- All Kiwi tarballs were unpacked to Git
- Simple `make` builds the sources to the `dist` subdirectory and builds the tarballs
- The ISO can be built locally using `make build`
- Updated/extended the documentation in README.md (There are two README.md files, the main in `/live` which documents the details and the `/live/src/README.md` which is copied to OBS. Mostl importantly it says "do not touch this in OBS, use GitHub".) 

## Testing

- Tested manually
- The autosubmission works (see the [GitHub action log](https://github.com/lslezak/agama/actions/runs/8780447151/job/24090382501))
- The built ISO works (tested in my [agama-live branch](https://build.opensuse.org/package/show/home:lslezak:branches:systemsmanagement:Agama:Staging/agama-live) project)
- Local build works as well


live/README.md
```diff
@@ -0,0 +1,240 @@
...
+## Layout
+
+This directory contains a set of files that are used to build the Agama Live ISO
+image.
+
+- [src](src) subdirectory contains all source files which are copied unmodified
+  to the OBS project
+- [root](root) subdirectory contains files which are added to the Live ISO root
+  system (inside the squashfs image)
+- [root-ALP-PXE](root-ALP-PXE) subdirectory contains specific files for the ALP
```
imobachgs, line 36: I guess we will need to drop this ALP reference at some point. But it is out of scope of this PR.

live/README.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+## Image Definition
+
+The [KIWI](https://github.com/OSInside/kiwi) image builder is used by OBS to
+build the Live ISO. See the [Kiwi documentation](
+https://osinside.github.io/kiwi/index.html) for more details about the build
+workflow and the `.kiwi` file format.
+
+### Kiwi Files
```
imobachgs, line 109: np: I would say the name of the project is *KIWI* (all capital letters) (see [KIWI homepage](https://osinside.github.io/kiwi/)).

imobachgs (commented):
It looks great. Thanks!

It would be even better if we update (or remove) the following references:

* https://github.com/openSUSE/agama/blob/master/doc/live_iso.md (obsolete, right?)
* https://github.com/openSUSE/agama/blob/master/PACKAGING.md#the-live-iso

live/root-ALP-PXE/etc/dracut.conf.d/90-brd.conf
```diff
@@ -0,0 +1 @@
+add_drivers+=" brd "
```
kobliha, line 1: Just a comment, **ALP** is dead

live/PXE.md
```diff
@@ -0,0 +1,46 @@
...
+
+- TFTP/DHCP running (usually dnsmasq)
+- FTP server running (usually vsftp)
+
+## Setup
+
+Extract the Linux kernel and the initrd from the archive:
+
+```shell
+osc getbinaries images x86_64 -M ALP-PXE
```
kobliha, line 15: ALP -> FO? Or call it ... call it something else?

live/root-ALP-PXE/etc/dracut.conf.d/90-brd.conf
lslezak: As @imobachgs mentioned above, let's solve that separately later.

imobachgs (approved):
Some *Kiwi* capitalization... otherwise, LGTM!

PR #1161, opened 2024-04-22
Use "DIAG" (upper case) label, added translators comment

lslezak:
## Problem

- The "Diag" label is confusing for the translators (see the [weblate discussion](https://l10n.opensuse.org/translate/agama/agama-web/en/?q=has:comment#comments))
- Usually it is written in the upper case form

## Solution

- Use upper case
- Added a comment for translators with more details and a hint that usually it is not translated


teclator (approved):
LGTM

PR #1162, opened 2024-04-23
GitHub Action: bind mount /usr/bin/gh into the containers

lslezak:
## Problem

- GitHub action which merges the translations from Weblate failed ([Action log](https://github.com/openSUSE/agama/actions/runs/8769617276/job/24065296950#step:3:27))
- The problem is that the repository changed the GPG signature and the `gh` package cannot be installed

## Solution

- Do not install it, it already is pre-installed in the default Ubuntu host system
- The `gh` tool is a statically linked binary without any dependencies so we can simply just bind mount it from the host into the container
- We already use this solution in the `.github/workflows/weblate-merge-service-po.yml` file, let's use it everywhere

### DIFF
--- a/.github/workflows/weblate-merge-po.yml
+++ b/.github/workflows/weblate-merge-po.yml
@@ -22,18 +22,20 @@ jobs:
 
     container:
       image: registry.opensuse.org/opensuse/tumbleweed:latest
+      volumes:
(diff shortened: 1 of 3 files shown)


mvidner (approved):
LGTM, thanks for the explanation

PR #1163, opened 2024-04-23
Delete not needed "dangerouslySetInnerHTML" parts

lslezak:
## Problem

- We should avoid using `dangerouslySetInnerHTML` in the code
- In this case it is actually not needed anymore as the text messages do not contain any HTML markup

## Solution

- Just delete the `dangerouslySetInnerHTML` parts


dgdavid:
Thanks @lslezak 

For others: more information at https://github.com/openSUSE/agama/pull/1151

dgdavid (approved):
LGTM

PR #1164, opened 2024-04-24
Adapt to recent changes on Y2Storage::GuidedProposal

ancorgs:
## Problem

Agama uses the `Y2Storage` strategy to find space called `:bigger_resize`. That strategy was found to be inconsistent, so it was slightly modified.

See more details at https://github.com/yast/yast-storage-ng/pull/1382

## Solution

This pull request adapts Agama to the new approach, in which the space actions are only expected for partitions (and maybe in the future for logical volumes), but not for whole disks.

1) The actions generated for the pre-defined policies do not longer include the disks, only partitions.

2) The "Find Space" form only offers the action selector for the partitions. In the case of disks without partitions, some static texts are added to help the user understand what can happen with the content of the disk.

partition_actions

Note we plan to rethink the whole interface of "Find Space" in the short term. This pull request only aims to keep the UI consistent with the new logic in the proposal, while the bigger changes arrive.

## Testing

- Unit tests adapted.
- Tested manually together with https://github.com/yast/yast-storage-ng/pull/1382

### DIFF
--- a/web/src/components/storage/SpaceActionsTable.jsx
+++ b/web/src/components/storage/SpaceActionsTable.jsx
@@ -27,7 +27,7 @@ import { FormSelect, FormSelectOption } from "@patternfly/react-core";
 import { _ } from "~/i18n";
 import { FilesystemLabel } from "~/components/storage";
 import { deviceChildren, deviceSize } from '~/components/storage/utils';
-import { If, Tag, TreeTable } from "~/components/core";
+import { Tag, TreeTable } from "~/components/core";
 import { sprintf } from "sprintf-js";
 
 /**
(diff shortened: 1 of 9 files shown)


web/src/components/storage/SpaceActionsTable.jsx
```diff
@@ -122,16 +120,45 @@ const DeviceAction = ({ device, action, isDisabled = false, onChange }) => {
...
+        onChange={onChange}
+      />
+    );
+  }
+
+  if (device.filesystem || device.component)
+    return _("The content may be deleted");
+
+  if (!device.partitionTable || device.partitionTable.partitions.length === 0)
+    return _("No content found");
```
dgdavid, line 157: NP: I still thinking that "No content found" is a bad content for an "Action" column. But I do not have a good proposal at the time of writing.

dgdavid (approved):

PR #1165, opened 2024-04-24
web: Reuse devices

joseivanlopez:
Implement feature for reusing an existing device (i.e., allow formatting a block device or mounting a file system). Main changes:

* Redesign the dialog for changing location:
  *  Shows a table with all the devices that can be selected as location for the volumes.
  * Allows selecting how to allocate the file system at selected device (create partition, create dedicated LVM, format, mount).
* Add a menu option in the table of volumes for resetting the location (i.e., to use the installation device instead of a custom location).

Note: We have detected some problems with the UI for changing location implemented in this PR:

* The table is scrollable. If there are too many devices, then there is a chance of overlooking some devices if you don't realize about the scroll.
* Having a table and group of radios at the end looks quite overwhelming. The very same information could be provided in a simpler way.
* Ideally, this change location option should be integrated in the form of the volume.

Check some ideas in the thread of this PR and at https://github.com/openSUSE/agama/discussions/1176.

<details>
 
localhost_8080_ (44)

localhost_8080_ (45)

</details>

### DIFF
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -3,6 +3,10 @@
   margin-block-end: var(--stack-gutter);
 }
 
+.stack.small {
+  --stack-gutter: var(--spacer-smaller);
+}
+
 .flex-stack {
   display: flex;
   flex-direction: column;
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -47,6 +47,20 @@
   }
 }
 
+.location-layout table button[id^="expand-toggle"] {
+  display: none;
+}
+
+.location-layout table tbody tr:not(:first-child) {
+  td:nth-child(2) {
+    padding-inline-start: 2ch;
+  }
+
+  td:nth-child(3) {
+    padding-inline-start: 5ch;
+  }
+}
+
 [data-state="reversed"] {
   flex-direction: row-reverse;
 }
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -32,11 +32,14 @@
     display: flex;
     flex-direction: column;
     flex: 1 1 0;
+    gap: 0
   }
 
-  form > div:first-child {
-    min-block-size: 120px;
+  form > div:nth-child(2) {
     overflow-y: auto;
+    min-block-size: 120px;
+    margin-block-end: var(--spacer-medium);
+    table { background: transparent; }
   }
 
   form > div:last-child {
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -3,10 +3,6 @@
   margin-block-end: var(--stack-gutter);
 }
 
-.stack.small {
-  --stack-gutter: var(--spacer-smaller);
-}
-
 .flex-stack {
   display: flex;
   flex-direction: column;
@@ -39,6 +35,7 @@
   }
 
   form > div:first-child {
+    min-block-size: 120px;
     overflow-y: auto;
   }
 
@@ -47,15 +44,12 @@
   }
 }
 
+// FIXME: Teporary solution to hide expand button. The button should be removed instead.
 .location-layout table button[id^="expand-toggle"] {
   display: none;
 }
 
 .location-layout table tbody tr:not(:first-child) {
-  td:nth-child(2) {
-    padding-inline-start: 2ch;
-  }
-
   td:nth-child(3) {
     padding-inline-start: 5ch;
   }
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -47,7 +47,7 @@
   }
 }
 
-// FIXME: Teporary solution to hide expand button. The button should be removed instead.
+// FIXME: Temporary solution to hide expand button. The button should be removed instead.
 .location-layout table button[id^="expand-toggle"] {
   display: none;
 }
--- a/web/src/assets/styles/composition.scss
+++ b/web/src/assets/styles/composition.scss
@@ -24,6 +24,25 @@
   flex-wrap: wrap;
 }
 
+// TODO: make it less specific.
+.location-layout > div {
+  display: flex;
+
+  form {
+    display: flex;
+    flex-direction: column;
+    flex: 1 1 0;
+  }
+
+  form > div:first-child {
+    overflow-y: auto;
+  }
+
+  form > div:last-child {
+    min-block-size: min-content;
+  }
+}
+
 [data-state="reversed"] {
   flex-direction: row-reverse;
 }
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -454,11 +454,18 @@ class ProposalManager {
   /**
    * Gets the devices that can be selected as target for a volume.
    *
+   * A device can be selected as target for a volume if either it is an available device for
+   * installation or it is a device built over the available devices for installation. For example,
+   * a MD RAID is a possible target only if all its members are available devices or children of the
+   * available devices.
+   *
    * @returns {Promise<StorageDevice[]>}
    */
   async getVolumeDevices() {
+    /** @type {StorageDevice[]} */
     const availableDevices = await this.getAvailableDevices();
 
+    /** @type {(device: StorageDevice) => boolean} */
     const isAvailable = (device) => {
       const isChildren = (device, parentDevice) => {
         const partitions = parentDevice.partitionTable?.partitions || [];
@@ -471,6 +478,7 @@ class ProposalManager {
       );
     };
 
+    /** @type {(device: StorageDevice[]) => boolean} */
     const allAvailable = (devices) => devices.every(isAvailable);
 
     const system = await this.system.getDevices();
@@ -576,6 +584,9 @@ class ProposalManager {
           return device;
         };
 
+        // Only consider the device assigned to a volume as installation device if it is needed
+        // to find space in that device. For example, devices directly formatted or mounted are not
+        // considered as installation devices.
         const volumes = dbusSettings.Volumes.v.filter(vol => (
           [VolumeTargets.NEW_PARTITION, VolumeTargets.NEW_VG].includes(vol.v.Target.v))
         );
--- a/web/src/client/storage.js
+++ b/web/src/client/storage.js
@@ -452,6 +452,35 @@ class ProposalManager {
   }
 
   /**
+   * Gets the devices that can be selected as target for a volume.
+   *
+   * @returns {Promise<StorageDevice[]>}
+   */
+  async getVolumeDevices() {
+    const availableDevices = await this.getAvailableDevices();
+
+    const isAvailable = (device) => {
+      const isChildren = (device, parentDevice) => {
+        const partitions = parentDevice.partitionTable?.partitions || [];
+        return !!partitions.find(d => d.name === device.name);
+      };
+
+      return (
+        !!availableDevices.find(d => d.name === device.name) ||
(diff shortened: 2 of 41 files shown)


web/src/assets/styles/composition.scss
```diff
@@ -24,6 +28,39 @@
...
+.location-layout table tbody tr:not(:first-child) {
+  td:nth-child(2) {
+    padding-inline-start: 2ch;
+  }
+
+  td:nth-child(3) {
+    padding-inline-start: 5ch;
+  }
+}
+
```
dgdavid, line 63: I forgot to add a FIXME comment here. This should be redo in a better way.

dgdavid:
@ancorgs, @joseivanlopez

I've sent the latest changes we were talking about this afternoon, please check https://github.com/openSUSE/agama/pull/1165/commits/e2ae65d36b07b1d64d1fd61442e5f52da37d8531 and screenshots below.

| Table not scrolled | Table scrolled |
|-|-|
| Screenshot from 2024-05-02 17-54-12 |  Screenshot from 2024-05-02 17-54-17 |

Although making it work was a bit more difficult than I though, I think it is enough for **mitigating** the inherited problems from an _scrollable HTML node_ in the middle of the UI because the browsers hiding the scroll bars.

However, I still thinking we have chosen the wrong layout here. If I'm able to find time, I will try to come up with an example with a different one to see if I'm right and, if so, I'm able to convince you about it.

In any case, let's considered this _scrollbox thingy implementation_ as temporary workaround that must be improved if finally it stays.

dgdavid:
> Although making it work was a bit more difficult than I though

BTW, while scrolling you will see how these chips in the "Usage" column overlaps the _scroll shadow_. I'm sorry but at this time I do not know how to _fix it_. I was playing with the `z-index` property without success. I'll try it again early in the morning, with a fresh mind :)

dgdavid:
> I will try to come up with an example

Here you have a rough test. Of course, need refinements (as use just one column in very small devices), but I hope it can be used for evaluation.

<details>

---

Kind of medium resolution

Screenshot from 2024-05-02 19-37-23
Screenshot from 2024-05-02 19-37-26
Screenshot from 2024-05-02 19-37-18


Bigger resolution

Screenshot from 2024-05-02 19-37-42

</details>

Changes can be found at https://github.com/openSUSE/agama/commit/ca0ad81e7254fe1e3b03d0825496077f3f6ea59a

IMHO, it provides a more guided, clear, and less prone to surprises experience.

dgdavid:
> > I will try to come up with an example
> 
> Here you have a rough test. Of course, need refinements (as use just one column in very small devices), but I hope it can be used for evaluation.
> 
> ...

Let's continue feeding thoughts.

How could fit the use of a select instead of a collection of radios?

<details>

---

Screenshot from 2024-05-02 22-15-20
Screenshot from 2024-05-02 22-15-25
Screenshot from 2024-05-02 22-15-35
</details>

Draft code available at https://github.com/openSUSE/agama/commit/dd3c5feaaf046331ff76d514f92079cafd2c3f3b

 PS: note that selector already comes with keyboard support.

web/src/components/storage/VolumeLocationSelectorTable.jsx
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Returns what (volumes, installation device) is using a device.
+ * @function
+ *
+ * @param {PartitionSlot|StorageDevice} item
+ * @param {StorageDevice[]} targetDevices
+ * @param {Volume[]} volumes
+ * @returns {string[]}
+ */
+const deviceUsers = (item, targetDevices, volumes) => {
```
dgdavid, line 50: NP: IMHO, the term _Users_ is a bit confusing.

web/src/components/storage/VolumeLocationSelectorTable.jsx
```diff
@@ -0,0 +1,121 @@
...
+  return users.concat(volumeUsers.map(v => v.mountPath));
+};
+
+/**
+ * @component
+ *
+ * @param {object} props
+ * @param {string[]} props.users
+ */
+const DeviceUsage = ({ users }) => {
```
dgdavid, line 69: NP: IMHO, naming the prop _users_ is a bit confusing.

web/src/components/storage/VolumeLocationSelectorTable.jsx
```diff
@@ -0,0 +1,121 @@
...
+  if (!device) return [];
+
+  const isTargetDevice = !!targetDevices.find(d => d.name === device.name);
+  const volumeUsers = volumes.filter(v => v.targetDevice?.name === device.name);
+
+  const users = [];
+  if (isTargetDevice) users.push(_("Installation device"));
+
+  return users.concat(volumeUsers.map(v => v.mountPath));
+};
```
dgdavid, line 61: NP: IMHO, the term _users_ here is a bit confusing.

dgdavid:
> 
> Let's continue feeding thoughts.
> 
> How could fit the use of a select instead of a collection of radios?
> ...

And now stop using the table for selecting devices and goes for a select too...

<details>

---

Screenshot from 2024-05-02 23-09-14
</details>

But with better format for the disk content information (it was just stolen from the table for the test). See https://github.com/openSUSE/agama/commit/09ad6b43d4e0b812bb526e1405fe287d534a019a

I'm going to stop here, but please take your time for thinking about this before outright rejecting any of the proposals (for future iterations, sure). Specially, try to imagine the last one with better styles and horizontal alignment

```
A good label

| Format ....  ⌄ |           | /dev/vdc1....  ⌄ |
```

Apart from a good, simple, keyboard navigation... it could be easily integrated in the VolumeDialog as yet another field.

Also, the device selector could use the https://www.patternfly.org/components/menus/select#typeahead feature.

Disclaimer: I'm not saying to go for it. Just to, at least, arrange time for evaluating it. Maybe we don't need tables everywhere.

ancorgs:
Talking about ideas... https://github.com/openSUSE/agama/discussions/1176

web/src/components/storage/VolumeLocationSelectorTable.jsx
joseivanlopez: It is quite internal, and the component explains what it means. Anyway, any suggestion?

dgdavid: > It is quite internal, 

True, but when reading the code it was a bit confusing to me because Users is a whole concept Agama manages.

> and the component explains what it means.

Right.

 > Anyway, any suggestion?

First of all, sorry for complaining without them :pray: I guess `deviceUsedBy` is too generic, so ignore the NP comment and keep the code as it's. We have much more interesting things to address right now :)

ancorgs:
Some notes about review.

We agreed we want to merge the UI implemented in this pull request... just as a first approach with close expiration date. We need to merge it in order to proceed with other important tasks.

The idea is to substitute this with an UI based on drop-downs (as sketched in a previous comment by @dgdavid). I would suggest to start with that next week.

I tested this pull request and works as expected. So from a functional POV, it LGTM.

But code-wise I would prefer @dgdavid to review it (taking into account some parts would be short-lived, as explained above).

web/src/client/storage.js
```diff
@@ -451,6 +451,43 @@ class ProposalManager {
...
+    /** @type {(device: StorageDevice) => boolean} */
+    const isAvailable = (device) => {
+      const isChildren = (device, parentDevice) => {
+        const partitions = parentDevice.partitionTable?.partitions || [];
+        return !!partitions.find(d => d.name === device.name);
+      };
+
+      return (
+        !!availableDevices.find(d => d.name === device.name) ||
+        !!availableDevices.find(d => isChildren(device, d))
```
dgdavid, line 477: NP: I wonder why only one `find` is not enough.

web/src/client/storage.js
```diff
@@ -451,6 +451,43 @@ class ProposalManager {
...
+   *
+   * @returns {Promise<StorageDevice[]>}
+   */
+  async getVolumeDevices() {
+    /** @type {StorageDevice[]} */
+    const availableDevices = await this.getAvailableDevices();
+
+    /** @type {(device: StorageDevice) => boolean} */
+    const isAvailable = (device) => {
+      const isChildren = (device, parentDevice) => {
```
dgdavid, line 470: NP: If I'm not wrong, we already have a `isChildren` function somewhere. Anyway, this kind of functions looks like a potential util that should live elsewhere.

web/src/components/core/ExpandableSelector.jsx
```diff
@@ -93,23 +98,26 @@ const sanitizeSelection = (selection, allowMultiple) => {
...
- * @param {object[]} props.items - Collection of items to be rendered.
- * @param {string} [props.itemIdKey="id"] - The key for retrieving the item id.
- * @param {(item: object) => Array<object>} [props.itemChildren=() => []] - Lookup method to retrieve children from given item.
- * @param {(item: object) => boolean} [props.itemSelectable=() => true] - Whether an item will be selectable or not.
- * @param {(item: object) => (string|undefined)} [props.itemClassNames=() => ""] - Callback that allows adding additional CSS class names to item row.
- * @param {object[]} [props.itemsSelected=[]] - Collection of selected items.
- * @param {string[]} [props.initialExpandedKeys=[]] - Ids of initially expanded items.
- * @param {(selection: Array<object>) => void} [props.onSelectionChange=noop] - Callback to be triggered when selection changes.
- * @param {object} [props.tableProps] - Props for {@link https://www.patternfly.org/components/table/#table PF/Table}.
+ * @typedef {object} ExpandableSelectorBaseProps
```
dgdavid, line 106: NP: Just a note not related with this PR. We are following this pattern of defining ComponentBaseProp and for now it is ok. But I wonder if, since it is something internal, we could call them just BaseProp instead. I see two problems, however:

* When a single file can end up with more than one BaseProp. Weird.
* `@typedef` auto exports types defined at top level, https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax#exporting-types

web/src/components/storage/VolumeLocationSelectorTable.jsx
```diff
@@ -0,0 +1,121 @@
...
+import { Chip } from '@patternfly/react-core';
+
+import { _ } from "~/i18n";
+import {
+  DeviceName, DeviceDetails, DeviceSize, toStorageDevice
+} from "~/components/storage/device-utils";
+import { ExpandableSelector } from "~/components/core";
+
+/**
+ * @typedef {import("../core/ExpandableSelector").ExpandableSelectorColumn} ExpandableSelectorColumn
```
dgdavid, line 34: NP: Did `~/components/core/ExpandableSelector` not work? Or maybe did you though it is too long? By the way, maybe `ExpandableSelectorColumn` can be names just `Column` in `core/ExpandableSelector.jsx` You have to rename it anyway when _importing_ it.

```
 * @typedef {import("~/components/core/ExpandableSelector").Column} ExpandableSelectorColumn
```

Again, not related with this PR but this kind of things makes me tend to think that we are actually running away of TypeScript with no strong reason.

dgdavid (approved):
LGTM, what a big and nice job. Thank you.

Please, do not take into consideration my NP comments. Think on them as just reminders for myself to discuss a few things in a near future.

From my side, you can go ahead with the merge :muscle:

web/src/components/storage/VolumeLocationSelectorTable.jsx
joseivanlopez: Good catch! The path was automatically added by my editor :).

web/src/client/storage.js
joseivanlopez: You are right. I guess it was at different places in some preliminary code and then it ends so.

PR #1169, opened 2024-04-24
Storage adapt

jreidinger:
It is the first step in having HTTP based API for storage. As it is really big API we decided to split it into series of patches for easier review.


rust/agama-server/src/storage/web.rs
```diff
@@ -0,0 +1,51 @@
...
+    const DBUS_PATH: &str = "/org/opensuse/Agama/Storage1";
+
+    let status_router = service_status_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
+    let progress_router = progress_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
+    let issues_router = issues_router(&dbus, DBUS_SERVICE, DBUS_PATH).await?;
+
+    let client = StorageClient::new(dbus.clone()).await?;
+    let state = StorageState { client };
+    let router = Router::new()
+        .route("/devices/dirty", get(devices_dirty))
```
imobachgs, line 41: I wonder whether it might be useful to have a more generic endpoint which could have more "status-like" attributes (apart from `dirty`). But if there is no need...

rust/agama-lib/src/storage/client.rs
```diff
@@ -140,4 +155,57 @@ impl<'a> StorageClient<'a> {
...
+    async fn build_device(
+        &self,
+        object: &(
+            OwnedObjectPath,
+            HashMap<OwnedInterfaceName, HashMap<std::string::String, OwnedValue>>,
+        ),
+    ) -> Result<Device, ServiceError> {
+        Ok(Device {
+            device_info: self.build_device_info(&object.0).await?,
+            component: None,
```
imobachgs, line 168: Perhaps we could implement `Default` or, if not possible, have some kind of constructor.

imobachgs (commented):
The approach looks good. Let's continue :-)

rust/agama-server/src/storage/web.rs
jreidinger: well, more generic will be probably that proposal settings. Currently for devices I am aware only about this dirty flag that signal that re probe is needed. ( and staging can be no longer valid )

rust/agama-lib/src/storage/client.rs
jreidinger: I am thinking that in the end I will add to `Device` struct method `try_from` that will construct itself from that Object Manager struct. and here will be just `Ok(object.into())`

rust/agama-lib/src/storage/client.rs
```diff
@@ -16,18 +21,112 @@ pub struct StorageDevice {
     description: String,
 }
 
+/// Represent single change action done to storage
```
imobachgs, line 24:
```suggestion
/// Represents a single change action done to storage
```

rust/agama-lib/src/storage/client.rs
```diff
@@ -16,18 +21,112 @@ pub struct StorageDevice {
...
+pub struct Action {
+    device: String,
+    text: String,
+    subvol: bool,
+    delete: bool,
+}
+
+/// Represents value for target key of Volume
+#[derive(Debug, Clone, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "snake_case")]
```
imobachgs, line 36: I guess you are using `snake_case` because it comes like that from storage-ng, right? I would add a note there.

rust/agama-lib/src/storage/client.rs
```diff
@@ -16,18 +21,112 @@ pub struct StorageDevice {
...
+            snapshots_configurable: get_property(&mvalue, "SnapshotsConfigurable")?,
+            snaphosts_affect_sizes: get_property(&mvalue, "SnapshotsAffectSizes")?,
+            size_relevant_volumes: get_property(&mvalue, "SizeRelevantVolumes")?,
+        };
+
+        Ok(res)
+    }
+}
+
+/// Represents single volume
```
imobachgs, line 93:
```suggestion
/// Represents a single volume
```

rust/agama-lib/src/storage/device.rs
imobachgs, line 1: We should consider putting these structs (and others in the `client` module) into a unique module (`storage::model` or something like that).

imobachgs (commented):
Just a few suggestions to reorganize the code a bit, but it can be postponed.

rust/agama-lib/src/storage/client.rs
jreidinger: yes, I try to use same values...in fact currently on D-Bus it is just string...so I think it would be nicer to have enum from it.

rust/agama-lib/src/storage/client.rs
```diff
@@ -140,4 +276,110 @@ impl<'a> StorageClient<'a> {
...
+        let properties = interfaces.get(&interface);
+        // All devices has to implement device info, so report error if it is not there
+        if let Some(properties) = properties {
+            Ok(DeviceInfo {
+                sid: get_property(properties, "SID")?,
+                name: get_property(properties, "Name")?,
+                description: get_property(properties, "Description")?,
+            })
+        } else {
+            Err(ServiceError::Anyhow(anyhow!(
```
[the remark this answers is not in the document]
jreidinger: Sadly, we do not know which device is missing it as that information is in Device interface ( name of device, its SID )...and this method does not get object path ( but if needed, I can add it as parameter just for error purpose ).

rust/agama-lib/src/dbus.rs
```diff
@@ -1,7 +1,48 @@
...
+            "Failed to find property '{}'",
+            name
+        )))?
+        .into();
+
+    T::try_from(value).map_err(|e| e.into())
+}
+
+/// It is similar helper like get_property with difference that name does not need to be in HashMap.
+/// In such case `None``is returned, so type has to be enclosed in `Option`.`
```
imobachgs, line 33:
```suggestion
/// In such case `None` is returned, so type has to be enclosed in `Option`.
```

imobachgs (approved):

PR #1170, opened 2024-04-25
A few sentences about how agama deals with security

mchf:
## Problem

Security audit

- *Bugzilla link*](https://bugzilla.suse.com/show_bug.cgi?id=1219688)

## Solution

a few sentences about security mechanism in agama's web server


doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
+Agama Concepts
```
imobachgs, line 1: Let's try to be consistent and use the `#`-based markdown notation. Not a big issue, but most of our documents are written in that way.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+==============
+
+Agama's functionality is divided into backend and frontend. Communication between two parts is done through REST api. Most of the api requires an authorization.
```
imobachgs, line 4: I would say there is some missing information here. The architecture is not so simple (yet). Agama's core is a service which is splitted into a Ruby and a Rust part, and they communicate through D-Bus (at this point, D-Bus works as our internal IPC).

Then, that's true, the "frontend" (let's consider the CLI a frontend too) communicate with the service through a dedicated API. Beware that it is not a REST API, just HTTP/JSON + Websocket.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+Authorization
+-------------
+
+Authorization is done via password. To get authorized user is asked for a password of root on backend's machine. The password is validated through PAM [1]. Once the authorization succeeds, backend generates an authorization token and passes it back to frontend / user. Agama uses JWT as authorization token [2]. All subsequent calls to the API has to be done together with the token.
```
imobachgs, line 9: What "To get authorized user" means? There is not another user apart from the root.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+To make local use (frontend and backend running on same machine) a bit easier agama implements option ```--generate-token```. When this option is used, agama-web-server generates valid JWT automatically on start. The token is pushed into web browser's cache by agama.
```
imobachgs, line 11: What is the purpose of "make local use a bit easier"? I would try to rephrase this section to make clear the motivation (skipping the authentication when working locally on Agama Live).

imobachgs, line 11: What JWT means? Please, add a link to https://jwt.io/.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+JWT
```
imobachgs, line 13:
```suggestion
Contents of the JWT
```

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+---
+
+The token carries just one claim - the expiration date. Token's lifetime is currently set to one day. The token is provided in an encrypted form. Security key is either automatically created random string [3] of length at least 30 characters. However, security key can be provided via ```/etc/agama.d/server.yaml``` using ```jwt_secret``` option. Content of this option is expected to be a string but no checks are done.
```
imobachgs, line 16: I would explain that, to generate a token, you need to have a key. Otherwise it is not clear what "security key" is exactly.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+Communication between the frontend and the backend
+--------------------------------------------------
+
+If both components run locally, communication can be done over http even https. However, in case when both run on different machines the https is mandatory. In such case all http requests are automatically redirected to https. A http response with code 308 (permanent redirect) is returned in such case.
+
+Https certificates
```
imobachgs, line 23: Please, use "HTTP" and "HTTPS" (instead of "http" and "https").

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+JWT
+---
+
+The token carries just one claim - the expiration date. Token's lifetime is currently set to one day. The token is provided in an encrypted form. Security key is either automatically created random string [3] of length at least 30 characters. However, security key can be provided via ```/etc/agama.d/server.yaml``` using ```jwt_secret``` option. Content of this option is expected to be a string but no checks are done.
+
+Communication between the frontend and the backend
+--------------------------------------------------
+
+If both components run locally, communication can be done over http even https. However, in case when both run on different machines the https is mandatory. In such case all http requests are automatically redirected to https. A http response with code 308 (permanent redirect) is returned in such case.
```
imobachgs, line 21: What about the websocket? We should mention how it behaves too.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+Https certificates
+------------------
+
+SSL communication is secured either by self-signed certificate which is automatically generated by agama if no certificate was provided by user. If agama should use particular custom certificate agama's web server provides options --cert and --key for path to certificate respectively to private key (in PEM format).
+
+... TBD ...
+
+External Projects in use
+========================
+[1] Rust pam crate
```
imobachgs, line 32: Please, add links.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+Communication between the frontend and the backend
+--------------------------------------------------
+
+If both components run locally, communication can be done over http even https. However, in case when both run on different machines the https is mandatory. In such case all http requests are automatically redirected to https. A http response with code 308 (permanent redirect) is returned in such case.
+
+Https certificates
+------------------
+
+SSL communication is secured either by self-signed certificate which is automatically generated by agama if no certificate was provided by user. If agama should use particular custom certificate agama's web server provides options --cert and --key for path to certificate respectively to private key (in PEM format).
```
imobachgs, line 26: Please, use "Agama" instead of "agama".

doc/agama-security.md
imobachgs, line 11: I would explain how we inject the cookie into the browser. See https://github.com/openSUSE/agama/blob/86afb33098786e8f1c505b78dfa25efcb4fd5264/live/README.md for further details (and the related PR https://github.com/openSUSE/agama/pull/1119).

imobachgs (changes requested):
Thanks! It is a good starting point. However, I think we need to explain more how things are organized and, even more important, the motivation for some decisions (like skipping the authentication).

doc/agama-security.md
mchf: Well. I intentionally didn't go in too much details about architecture as i somehow considered it doesn't belong here.

doc/agama-security.md
mchf: ok, will re-think it

mchf: I'd bet that I put a reference to RFC in the document. But obviously not. Will improve it.

doc/agama-security.md
mchf: ok.

doc/agama-security.md
mchf: ok

doc/agama-security.md
mchf: ok, i planned to do markdown fine as last step

doc/agama-security.md
imobachgs: You do not need to go into many details, but in my opinion it makes sense to talk a little bit about the architecture because it is crucial to understand how the security model is implemented.

doc/agama-security.md
mchf: yeah websockets ... that piece i need to study ... never used == completely forgot of that

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Authorization
+
+Authorization is done via password. To get authorized frontend has to provide a root password (root on backend's machine). The password is validated through PAM [1]. Once the authorization succeeds, backend generates an authorization token and passes it back to frontend / user. Agama uses JWT [2] as authorization token [3]. All subsequent calls to the API has to be done together with the token. In case of web UI the token is stored in session cookie.
+
+To make local use (frontend and backend running on same machine) with respect to agama-live use case more friendly and allow skipping explicit login in web UI Agama implements option ```--generate-token```. When this option is used, Agama's web server service generates valid JWT automatically on start. The token is stored locally [4] and then imported into web browser's internal database by Agama provided startup [5]. The script prepares custom profile with predefined homepage pointing to Agama's login page with the generated token as get parameter in the homepage url. Then the firefox browser is started in kiosk mode.
```
imobachgs, line 11: Actually, the condition to skip the authentication is not just that they run in the same system. Otherwise if you have Agama (not expected, but...) installed in your system (and running), anyone could access it.

I would replace this paragraph with something like this:

```
However, there is an special scenario where we need to skip the authentication: using the browser or the CLI included in the installation media (Agama Live) so it feels like it is using a desktop application. Of course, connecting remotely to the installation media would ask for authentication, but not when using the local browser.

To skip the authentication, the frontend gets a valid token injected. Such a token is generated passing the option `--generate-token` to `agama-web-server`, which writes a valid token at `/run/agama/token` (only the root can read that file) at boot.

How is the token injected in the browser and the CLI?

* [Firefox startup script](https://github.com/openSUSE/agama/blob/architecture_2024/live/root/.icewm/startup) points the browser to an special URL including the token in the request. As part of the response, the token is stored as `httpOnly` cookie.
* In the CLI case, things are easier because it has access to the token (it runs as root in Agama Live).
```

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+The token carries just one claim - the expiration date. Token's lifetime is currently set to one day. The token is provided in an encrypted form. Security key is either automatically created random string [6] of length at least 30 characters. However, security key can be provided via ```/etc/agama.d/server.yaml``` using ```jwt_secret``` option. Content of this option is expected to be a string but no checks are done.
+
+### Communication between the frontend and the backend
+
+If both components run locally, communication can be done over HTTP even HTTPS. However, in case when both run on different machines the HTTPS is mandatory. In such case all HTTP requests are automatically redirected to HTTPS. A HTTP response with code 308 (permanent redirect) is returned in such case.
+
+For notifications on changes from backend Agama uses WebSocket technology. Typically backend notificates about installation progress or network configuration changes this way.
+
+### Https certificates
```
imobachgs, line 23:
```suggestion
### HTTPS certificates
```

doc/agama-security.md
mchf: I took your paragraph as an inspiration

mchf:
> I have done a bunch of suggestions. Please, review the footnotes because some of the might have change after applying some of the changes (especially the suggestion about skipping the authentication).

I didn't use all changes exactly as suggested. References seems ok.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+SSL communication is secured either by self-signed certificate which is automatically generated by Agama if no certificate was provided by user. If Agama should use particular custom certificate Agama's web server provides options --cert and --key for path to certificate respectively to private key (in PEM format).
+
+## Links to external sources
+
+- [1] Rust pam crate, https://crates.io/crates/pam
+- [2] RFC 7519, http://jwt.io
+- [3] Rust jsonwebtoken crate, https://crates.io/crates/jsonwebtoken
+- [4] Backend's machine at /run/agama/token
+- [5] [Firefox startup script] See Agama's git repository path - live/root/.icewm/startup
```
imobachgs, line 33: Missing links? You can point to https://github.com/openSUSE/agama/blob/master/live/root/root/.icewm/startup because it will be updated with the proper code after mering `architecture_2024` into `master`.

doc/agama-security.md
```diff
@@ -0,0 +1,34 @@
...
+
+For notifications on changes from backend Agama uses WebSocket technology. Typically backend notificates about installation progress or network configuration changes this way.
+
+### HTTPS certificates
+
+SSL communication is secured either by self-signed certificate which is automatically generated by Agama if no certificate was provided by user. If Agama should use particular custom certificate Agama's web server provides options --cert and --key for path to certificate respectively to private key (in PEM format).
+
+## Links to external sources
+
+- [1] Rust pam crate, https://crates.io/crates/pam
```
imobachgs, line 29:
```suggestion
- [1] Rust PAM crate, https://crates.io/crates/pam
```

imobachgs (approved):
Just two fixes but, otherwise, LGTM. Feel free to merge when you apply the fixes.

PR #1171, opened 2024-04-26
Storage: show skeletons only when needed

lslezak:
## Problem

- The loading skeletons are displayed always, even when not needed
- The installation device selection popup is empty when clicked quickly before loading all data

### Loading Skeletons

- Pass the name of the component which is changed by user to parent
- Send back the changing component to all children
- There is mapping which defined whether a child needs refresh when some setting is being changed

### Installation Device Progress

- Use the sent data when when the stored values are `undefined`
- Show a simple skeleton

### screenreaderText Properties

- I have removed all `screenreaderText` properties for skeletons and keep it only for the result section
- The reason is that it does not make sense to notify users about 5 values in progress, it usually takes about 2 seconds to refresh the data so the screen reader would not have enough time to read all messages
- Use it only the result section which is refreshed always, moreover the data for the other values will be read together with the result data, they will always take the same time

## Questions

- Do we need more granularity with changing values? I.e. if the Btrfs snapshots are enabled/disabled then the partition list below is refresh as well. (IMHO fine to solve this later in a separate card / PR.)
- Better progress for the installation device loading? (Note: do not overengineer here, the progress is displayed only when you click the link very quickly after opening the storage page, in most cases it won't be used at all.)

## Testing

- Tested manually

## Screenshots

When enabling/disabling the Btrfs snapshots all data was refreshed, now the installation device and the space policy does not show the progress skeletons.

| Original | Fixed |
|---|---|
| storage_skeletons_broken | storage_skeletons_fixed |

Another fix was empty installation device screen when the "Installation device" link was clicked quickly, before loading the data finished.

| Original | Fixed |
|---|---|
| storage_loading_progress_broken | storage_loading_progress2  |

### DIFF
--- a/web/src/components/core/If.jsx
+++ b/web/src/components/core/If.jsx
@@ -60,15 +60,13 @@
  *
  * @param {object} props
  * @param {boolean} props.condition
- * @param {(JSX.Element|function():JSX.Element)} [props.then=null] - the content to be rendered when the condition is true
- * @param {(JSX.Element|function():JSX.Element)} [props.else=null] - the content to be rendered when the condition is false
+ * @param {JSX.Element} [props.then=null] - the content to be rendered when the condition is true
+ * @param {JSX.Element} [props.else=null] - the content to be rendered when the condition is false
  */
 export default function If ({
   condition,
   then: positive = null,
   else: negative = null
 }) {
-  const ret = condition ? positive : negative;
-  // if the parameter is a function then evaluate it
-  return (typeof ret === "function") ? ret() : ret;
+  return condition ? positive : negative;
 }
--- a/web/src/components/core/If.jsx
+++ b/web/src/components/core/If.jsx
@@ -60,13 +60,15 @@
  *
  * @param {object} props
  * @param {boolean} props.condition
- * @param {JSX.Element} [props.then=null] - the content to be rendered when the condition is true
- * @param {JSX.Element} [props.else=null] - the content to be rendered when the condition is false
+ * @param {(JSX.Element|function():JSX.Element)} [props.then=null] - the content to be rendered when the condition is true
+ * @param {(JSX.Element|function():JSX.Element)} [props.else=null] - the content to be rendered when the condition is false
  */
 export default function If ({
   condition,
   then: positive = null,
   else: negative = null
 }) {
-  return condition ? positive : negative;
+  const ret = condition ? positive : negative;
+  // if the parameter is a function then evaluate it
+  return (typeof ret === "function") ? ret() : ret;
 }
--- a/web/src/components/core/Popup.jsx
+++ b/web/src/components/core/Popup.jsx
@@ -25,6 +25,7 @@ import React from "react";
 import { Button, Modal } from "@patternfly/react-core";
 import { _ } from "~/i18n";
 import { partition } from "~/utils";
+import { Loading } from "~/components/layout";
 
 /**
  * @typedef {import("@patternfly/react-core").ModalProps} ModalProps
@@ -190,12 +191,14 @@ const AncillaryAction = ({ children, ...actionsProps }) => (
  * @typedef {object} PopupBaseProps
  * @property {"auto" | "small" | "medium" | "large"} [blockSize="auto"] - The block/height size for the dialog. Default is "auto".
  * @property {"auto" | "small" | "medium" | "large"} [inlineSize="medium"] - The inline/width size for the dialog. Default is "medium".
+ * @property {boolean} [isLoading=false] - Whether the data is loading, if yes it displays a loading indicator instead of the requested content
  * @typedef {Omit<ModalProps, "variant" | "size"> & PopupBaseProps} PopupProps
  *
  * @param {PopupProps} props
  */
 const Popup = ({
   isOpen = false,
+  isLoading = false,
   showClose = false,
   inlineSize = "medium",
   blockSize = "auto",
@@ -214,7 +217,7 @@ const Popup = ({
       actions={actions}
       className={`${className} block-size-${blockSize} inline-size-${inlineSize}`.trim()}
     >
-      {content}
+      {isLoading ? <Loading text={_("Loading data...")} /> : content}
     </Modal>
   );
 };
--- a/web/src/components/core/Popup.jsx
+++ b/web/src/components/core/Popup.jsx
@@ -192,6 +192,7 @@ const AncillaryAction = ({ children, ...actionsProps }) => (
  * @property {"auto" | "small" | "medium" | "large"} [blockSize="auto"] - The block/height size for the dialog. Default is "auto".
  * @property {"auto" | "small" | "medium" | "large"} [inlineSize="medium"] - The inline/width size for the dialog. Default is "medium".
  * @property {boolean} [isLoading=false] - Whether the data is loading, if yes it displays a loading indicator instead of the requested content
+ * @property {string} [loadingText="Loading data..."] - Text displayed when `isLoading` is set to `true`
  * @typedef {Omit<ModalProps, "variant" | "size"> & PopupBaseProps} PopupProps
  *
  * @param {PopupProps} props
@@ -199,6 +200,8 @@ const AncillaryAction = ({ children, ...actionsProps }) => (
 const Popup = ({
   isOpen = false,
   isLoading = false,
+  // TRANSLATORS: progress message
+  loadingText = _("Loading data..."),
   showClose = false,
   inlineSize = "medium",
   blockSize = "auto",
@@ -217,7 +220,7 @@ const Popup = ({
       actions={actions}
       className={`${className} block-size-${blockSize} inline-size-${inlineSize}`.trim()}
     >
-      {isLoading ? <Loading text={_("Loading data...")} /> : content}
+      {isLoading ? <Loading text={loadingText} /> : content}
     </Modal>
   );
 };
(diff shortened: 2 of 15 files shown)


ancorgs:
I see the form for selecting the installation device is now protected from "quick clickers"[1]. 

But that's the only pop-up that got that protection. What about these?

- Encryption - if the current settings have encryption enabled but I click on "Encryption" before data is loaded, then I get a form that makes me think encryption is not enabled. I expected a "loading" message/spinner/something while the data loads and then getting the form correctly initialized.
- Find space - if I click before the data is loaded, I get a Javascript error "policy is undefined".

[1] I don't using a skeleton within the pop-up instead of some kind of "spinner", but that's another topic for another comment.

web/src/components/storage/DeviceSelectionDialog.jsx
```diff
@@ -118,63 +125,70 @@ devices.").split(/[[\]]/);
...
-              itemSelectable={isDeviceSelectable}
-              onSelectionChange={setTargetPVDevices}
-              variant="compact"
-            />
-          </Panels.Panel>
-        </Panels>
-      </Form>
+      <If
+        condition={isLoading}
+        then={ <SectionSkeleton numRows={6} /> }
```
ancorgs, line 130: Honestly, this skeleton here looks really weird.

As commented during the planning meeting of the card and again today during review, the goal of the skeletons is to "reserve space" in the UI for the future content, so things don't jump up and down as that content is loaded into its place.

In case of a pop-up that is going to load at once as soon the content arrives, that multi-line skeleton looks like some kind of render error. I expected the pop-up to open kind of blank but with some "loading" message and/or a spinner (I personally expected the same three circles we have when loading the whole UI).

web/src/components/storage/ProposalPage.jsx
```diff
@@ -96,6 +99,30 @@ const reducer = (state, action) => {
...
+// mapping of not affected values for settings components
+// key:   component name
+// value: list of items which can be changed without affecting
+//        the state of the component
+export const NOT_AFFECTED = {
+  // the EncryptionField shows the skeleton only during initial load,
+  // it does not depend on any changed item and does not show skeleton later.
+  // the ProposalResultSection is refreshed always
+  InstallationDeviceField: [CHANGING.ENCRYPTION, CHANGING.BOOT, CHANGING.POLICY, CHANGING.VOLUMES],
+  PartitionsField: [CHANGING.ENCRYPTION],
```
ancorgs, line 122: Shouldn't this include `CHANGING.POLICY`?

web/src/components/storage/ProposalPage.jsx
```diff
@@ -96,6 +99,30 @@ const reducer = (state, action) => {
...
+  SpacePolicyField: [CHANGING.ENCRYPTION, CHANGING.BOOT, CHANGING.VOLUMES, CHANGING.TARGET],
```
ancorgs, line 123: Just a comment about this. Being fully strict, the fact is that the content of the sentence displayed right after "Find Space" can actually be affected by changes in the target disk(s), in the booting disk and in the list of volumes.

For example, the sentence can initially be "deleting all content of the installation disk". But then I change the target and decide to install on an LVM over sdb and sdc. Then the sentence changes to "deleting all content of the 2 selected disks". Then I decide that, in addition to that, the system will boot from sda so the sentence changes to "deleting all content of the 3 selected disks". And for extra fun I decide to allocate /home as a new partition at sdd. Then the sentence changes again to "deleting all content of the 4 selected disks".

I don't know if those changes in the sentence really deserve a skeleton. I just wanted to mention there is SOME influence and as such it should be documented next to the code.

ancorgs:
In fact, `BootConfigField` is never affected by changes in other settings. It only makes sense to display it as a skeleton when the whole page is loading. But once the value is set it can only be changed via its very own "Change boot options" button.

Watching it become a skeleton so often feels quite wrong.

web/src/components/storage/ProposalSettingsSection.jsx
```diff
@@ -40,6 +41,18 @@ import SpacePolicyField from "~/components/storage/SpacePolicyField";
...
 
+/**
+ * A helper function to decide whether to show the progress skeletons or not
+ * for the specified component
+ * @param {boolean} loading loading status
+ * @param {string} component name of the component
+ * @param {symbol} changing the item which is being changed
+ * @returns {boolean} true if the skeleton should be displayed, false otherwise
+ */
+const ShowSkeleton = (loading, component, changing) => {
```
dgdavid, line 52: I would not use PascalCase here. It looks like a component but actually is not a component.

web/src/components/core/If.jsx
```diff
@@ -60,13 +60,15 @@
...
  */
 export default function If ({
   condition,
   then: positive = null,
   else: negative = null
 }) {
-  return condition ? positive : negative;
+  const ret = condition ? positive : negative;
+  // if the parameter is a function then evaluate it
+  return (typeof ret === "function") ? ret() : ret;
```
lslezak, line 73: This allows lazy evaluation of the branches. The problem is that `<If>` evaluates *both* branches and then returns one of them.

So for example this
```jsx
<If condition={data === undefined} then={<Loading/>} else={<Foo foo={data.foo}>} />
```

crashes when `data` is `undefined` because it evaluates the `else` branch too early.

With this improvement you can pass a function which can be evaluated only when the result is needed. (Passing functions is quite common in Javascript to allow the lazy evaluation.)

Then the fix is trivial, just add `() =>` before the value:
```jsx
<If condition={data === undefined} then={<Loading/>} else={() => <Foo foo={data.foo}>} />
```

@dgdavid are you OK with this change?

I'll add some example into the `<If>` documentation if you agree with this improvement.

dgdavid: Thanks @lslezak 

Let's say that I have no strong opinion about it. At first sight, it looks reasonable to prevent not needed evaluations. But on the other hand it's weird to me to accept a function as content of the branches. IMHO it should be limited to [ReactNode](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/778ad7a80e9876f258270d84d1db4e6e1d5e3796/types/react/index.d.ts#L448-L489) (Although technically a component is a function too).


That said, I fully understand the proposal looking at the code written at https://github.com/openSUSE/agama/pull/1171/files#diff-994f1f4ff00fecec1797a2aaa799dbbeddc4e8fb2c1ae666fedbb95b7c499770R155-R178

```jsx
  <If
        condition={isLoading}
        then={
          <Loading text={_("Loading data...")} />
        }
        else={
          () =>
            <Form id="space-policy-form" onSubmit={onSubmit}>
              <SpacePolicyPicker currentPolicy={policy} onChange={setPolicy} />
              <If
                condition={devices.length > 0}
                then={
                  <SpaceActionsTable
                    devices={devices}
                    expandedDevices={expandedDevices}
                    deviceAction={deviceAction}
                    isActionDisabled={policy.id !== "custom"}
                    onActionChange={changeActions}
                  />
                }
              />
            </Form>
        }
      />
```

I would go for a different implementation instead. For me, what is in the else should be a component that React should be able to understand. Let's say `<SpacePolicyForm />`. Then, the component should be aware about what it needs for return either, an output or nothing.

```jsx
const SpacePolicyForm = ({ prop1, prop2 }) => {
  if (!prop1 || !prop2) return;

  ...

  return (
    <div><b>{prop1}:</b>{prop2}</div>
  );
}
```

I.e., the key for me is to write more robust components.

Of course, I might be wrong, so maybe another view here could help to decide what way to go.

joseivanlopez: I was assuming the `else` branch is not evaluated if the condition is true. So, probably, I have written unsafe code.

dgdavid: BTW, as far as I can see I think that this set of changes reveals another change we could address: to add the `isLoading` prop to core/Popup and make it show the spinner or its children based on the prop value. That way, we could reduce the `<If condition={isLoading} then={...` repetition and have more clean WhateverDialog.jsx components.

lslezak: After reading [this](https://medium.com/doctolib/the-subtle-aspects-of-react-re-rendering-part-1-functions-calls-and-lazy-evaluation-6136b0e31995) I decided to revert the `<If>` change. It seems it is not safe and it can have some nasty side effects in some cases.

web/src/components/core/Popup.jsx
```diff
@@ -214,7 +217,7 @@ const Popup = ({
       actions={actions}
       className={`${className} block-size-${blockSize} inline-size-${inlineSize}`.trim()}
     >
-      {content}
+      {isLoading ? <Loading text={_("Loading data...")} /> : content}
```
dgdavid, line 220: By any means I want to make your PR more longer than needed. But what do you think about taking the opportunity for adding the "isLoadingText" prop to make possible customize it when needed. Maybe there is a place when we want to be explicit about what data is the dialog loading. E.g., "Loading available devices...". Or to add more information, "Loading encryption data. Be patient, this can take a while".

Please, feel free to not address it here since I fully understand is completely out of topic.

dgdavid: BTW, I meant a prop with a default value, of course `isLoadingText = _("Loading data...")`

web/src/components/core/If.jsx
imobachgs: @joseivanlopez Actually I do not think you have written "unsafe" code. The `Form` itself is not executed (@dgdavid and me has done some tests with another example). The problem is the `data.foo` when `data` is undefined (that's the code that gets executed).

dgdavid (approved):
LGTM, thanks.

PR #1175, opened 2024-05-02
Adapt storage proposal to the new architecture

jreidinger:
This PR implements an HTTP/JSON API for managing the storage proposal and all the support methods needed by the corresponding web interface.

This works as a proxy over the real API which is implemented in D-Bus. That comes with some problems (especially regarding performance) that will need to be fixed in the future.

## Testing

- Existing unit tests adapted to the new API, except those related to signals (`onDeprecate` and `onIssuesChange`).
- Tested manually

## References

https://trello.com/c/HOkBiVHp/


rust/agama-lib/src/storage/client.rs
```diff
@@ -1,111 +1,25 @@
...
-    target: VolumeTarget,
-    target_device: Option<String>,
-    min_size: u64,
-    max_size: Option<u64>,
-    auto_size: bool,
-    snapshots: Option<bool>,
-    transactional: Option<bool>,
-    outline: Option<VolumeOutline>,
-}
+type DbusObject = (
```
imobachgs: We often use `DBusObject` instead.

rust/agama-lib/src/storage/client.rs
```diff
@@ -127,19 +41,14 @@ impl<'a> StorageClient<'a> {
                 .path("/org/opensuse/Agama/Storage1")?
                 .build()
                 .await?,
-            proposal_proxy: ProposalProxy::new(&connection).await?,
+            proposal_proxy: ProposalProxy::builder(&connection)
+                .cache_properties(zbus::CacheProperties::No)
```
imobachgs, line 45: :+1: Perhaps we should document somewhere why we are doing this.

rust/agama-lib/src/storage/client.rs
```diff
@@ -178,20 +81,20 @@ impl<'a> StorageClient<'a> {
...
-            transactional: get_optional_property(&volume_hash, "Transactional")?,
-            outline: get_optional_property(&volume_hash, "Outline")?,
-        };
-
-        Ok(volume)
+
+        Ok(volume_hash.try_into()?)
+    }
+
+    pub async fn product_mount_points(&self) -> Result<Vec<String>, ServiceError> {
```
imobachgs, line 88: Add to your personal TODO list to document these methods :-) Actually, I expect some refactoring (because we might need an HTTP-based client) so it can be part of that change.

rust/agama-lib/src/storage/client.rs
```diff
@@ -379,4 +271,159 @@ impl<'a> StorageClient<'a> {
...
+            }))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn build_drive(&self, object: &DbusObject) -> Result<Option<Drive>, ServiceError> {
+        let properties = self.get_interface(object, "org.opensuse.Agama.Storage1.Drive");
+
+        if let Some(properties) = properties {
```
imobachgs, line 295: np: This is perfectly fine, but I would have used a [let-else](https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html) construction. Nothing to change today, anyway.

imobachgs (commented):
I have finished checking the Rust part. There are a few things that I am pretty sure Clippy will complain, so we can address them in the short term. But by now it is fine.

rust/agama-lib/src/storage/client.rs
joseivanlopez: Fixed.

web/src/client/storage.js
```diff
@@ -245,172 +247,135 @@ class DevicesManager {
...
+        addRaidInfo(device, info);
       };
 
-      /** @type {(device: StorageDevice, partitionProperties: object) => void} */
-      const addPartitionProperties = (device, partitionProperties) => {
+      /** @type {(device: StorageDevice, info: object) => void} */
+      const addPartitionInfo = (device, info) => {
         device.type = "partition";
-        device.isEFI = partitionProperties.EFI.v;
+        device.isEFI = info.EFI;
```
imobachgs, line 293: Is the capitalization right (`EFI`)?

web/src/client/storage.js
```diff
@@ -245,172 +247,135 @@ class DevicesManager {
...
    */
   async getDevices() {
-    /** @type {(path: string, dbusDevices: object[]) => StorageDevice} */
-    const buildDevice = (path, dbusDevices) => {
-      /** @type {(device: StorageDevice, deviceProperties: object) => void} */
-      const addDeviceProperties = (device, deviceProperties) => {
-        device.sid = deviceProperties.SID.v;
-        device.name = deviceProperties.Name.v;
-        device.description = deviceProperties.Description.v;
+    const buildDevice = (jsonDevice, jsonDevices) => {
```
imobachgs, line 250: I have the impression that we are abusing inline functions. AFAIS, this function does not have any closure and, moreover, it defines many other internal functions. I would add this to our list of things to discuss.

web/src/client/storage.test.js
```diff
@@ -682,475 +662,480 @@ const contexts = {
...
+    },
+    {
+      deviceInfo: {
+        sid: 60,
+        name: "/dev/sda1",
+        description: ""
       },
-      "org.opensuse.Agama.Storage1.Partition": {
-        EFI: { t: "b", v: false }
+      partition: { EFI: false },
```
imobachgs, line 707: If capitalization is OK, then nothing to say. Otherwise, we might need to adjust it accordingly.

imobachgs (commented):
Just a few comments, nothing serious.

web/src/client/storage.js
joseivanlopez: It is wrong. Good catch!

rust/agama-lib/src/storage/client.rs
```diff
@@ -332,17 +209,35 @@ impl<'a> StorageClient<'a> {
...
         let interfaces = &object.1;
-        let interface: OwnedInterfaceName =
-            InterfaceName::from_static_str_unchecked("org.opensuse.Agama.Storage1.Device").into();
-        let properties = interfaces.get(&interface);
+        interfaces.get(&interface)
+    }
+
+    async fn build_device(&self, object: &DBusObject) -> Result<Device, ServiceError> {
+        Ok(Device {
+            block_device: self.build_block_device(object).await?,
```
joseivanlopez, line 224: The conversion of other objects (e.g., `ProposalSettings`) is done in the model by implementing the `TryFrom` trait. For the devices, the objects are directly created in the client. Probably it would be better to follow the same approach and implement `TryFrom` for the needed types in the model.

imobachgs (approved):

PR #1178, opened 2024-05-03
Network cli

teclator:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -9,6 +9,7 @@ use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 
 const DEFAULT_JWT_FILE: &str = ".agama/agama-jwt";
+const DEFAULT_AGAMA_TOKEN_FILE: &str = "/run/agama/token";
 const DEFAULT_AUTH_URL: &str = "http://localhost:3000/api/auth";
 const DEFAULT_FILE_MODE: u32 = 0o600;
 
@@ -33,8 +34,19 @@ pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
     }
 }
 
+/// Reads stored agama-live token and returns it
+pub fn agama_token() -> anyhow::Result<String> {
+    if let Some(file) = agama_token_file() {
+        if let Ok(token) = read_line_from_file(file.as_path()) {
+            return Ok(token);
+        }
+    }
+
+    Err(anyhow::anyhow!("Authentication token not available"))
+}
+
 /// Reads stored token and returns it
-fn jwt() -> anyhow::Result<String> {
+pub fn jwt() -> anyhow::Result<String> {
     if let Some(file) = jwt_file() {
         if let Ok(token) = read_line_from_file(file.as_path()) {
             return Ok(token);
@@ -108,6 +120,10 @@ impl Credentials for MissingCredentials {
 /// Path to file where JWT is stored
 fn jwt_file() -> Option<PathBuf> {
     Some(home::home_dir()?.join(DEFAULT_JWT_FILE))
+}
+/// Path to agama-live token file
+fn agama_token_file() -> Option<PathBuf> {
+    Some(home::home_dir()?.join(DEFAULT_AGAMA_TOKEN_FILE))
 }
 
 /// Reads first line from given file
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -1,6 +1,7 @@
 use clap::{arg, Args, Subcommand};
 
 use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
+use rpassword::{prompt_password, read_password};
 use std::fs;
 use std::fs::File;
 use std::io;
@@ -153,12 +154,9 @@ fn read_line_from_file(path: &Path) -> io::Result<String> {
 
 /// Asks user to provide a line of input. Displays a prompt.
 fn read_credential(caption: String) -> io::Result<String> {
-    let mut cred = String::new();
-
-    println!("{}: ", caption);
-
-    io::stdin().read_line(&mut cred)?;
-    if cred.pop().is_none() || cred.is_empty() {
+    let caption = format!("{}: ", caption);
+    let cred = prompt_password(caption.clone()).unwrap();
+    if cred.is_empty() {
         return Err(io::Error::new(
             io::ErrorKind::Other,
             format!("Failed to read {}", caption),
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -1,3 +1,4 @@
+use crate::auth;
 use crate::error::CliError;
 use crate::printers::{print, Format};
 use agama_lib::connection;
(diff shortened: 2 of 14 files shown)


rust/agama-cli/src/config.rs
```diff
@@ -31,41 +32,57 @@ pub enum ConfigAction {
     Load(String),
 }
 
+fn token() -> Option<String> {
+    match auth::jwt() {
```
imobachgs, line 36: What about:

```rust
auth::jwt().or_else(|_| auth::agama_token()).ok()
```

rust/agama-cli/src/auth.rs
```diff
@@ -109,6 +122,10 @@ impl Credentials for MissingCredentials {
 fn jwt_file() -> Option<PathBuf> {
     Some(home::home_dir()?.join(DEFAULT_JWT_FILE))
 }
+/// Path to agama-live token file
+fn agama_token_file() -> Option<PathBuf> {
+    Some(home::home_dir()?.join(DEFAULT_AGAMA_TOKEN_FILE))
```
imobachgs, line 127: What about:

```rust
home::home_dir().map(|p| p.join(DEFAULT_AGAMA_TOKEN_FILE)
```

So you do not need to wrap the expresion (which contains a `?`) on a `Some`. But I guess it is just a matter of taste.

rust/agama-cli/src/auth.rs
```diff
@@ -137,12 +154,9 @@ fn read_line_from_file(path: &Path) -> io::Result<String> {
...
 /// Asks user to provide a line of input. Displays a prompt.
 fn read_credential(caption: String) -> io::Result<String> {
-    let mut cred = String::new();
-
-    println!("{}: ", caption);
-
-    io::stdin().read_line(&mut cred)?;
-    if cred.pop().is_none() || cred.is_empty() {
+    let caption = format!("{}: ", caption);
+    let cred = prompt_password(caption.clone()).unwrap();
```
imobachgs, line 158: In this case, where we have many small functions in this file, I would most probably keep the prefix:

```
rpassword::prompt_password(...)
```

rust/agama-cli/src/config.rs
```diff
@@ -1,3 +1,4 @@
+use crate::auth;
```
imobachgs, line 1: It might be a good opportunity to merge all the crate uses.

```rust
use crate::{
  auth, error::CliError, printers::{print, Format}
};
```

rust/agama-cli/src/config.rs
```diff
@@ -31,41 +32,57 @@ pub enum ConfigAction {
...
+        Err(_) => match auth::agama_token() {
+            Ok(token) => return Some(token),
+            Err(_) => return None,
+        },
+    }
+}
+
 pub async fn run(subcommand: ConfigCommands, format: Format) -> anyhow::Result<()> {
-    let store = SettingsStore::new(connection().await?).await?;
+    if let Some(token) = token() {
```
imobachgs, line 46: If you are not interested in a potential error coming from `token()`, you could use `let else` (to avoid indenting the whole function).

```rust
let Some(token) = token() {
  println!("You need to login for generating a valid token");
  return Ok(());
}
```

Having said that, you should not return `Ok(())` when it is actually an error.

imobachgs (changes requested):
I have not finished with the review yet, just a partial one.

rust/agama-cli/src/auth.rs
```diff
@@ -109,6 +121,10 @@ impl Credentials for MissingCredentials {
 fn jwt_file() -> Option<PathBuf> {
     Some(home::home_dir()?.join(DEFAULT_JWT_FILE))
 }
+/// Path to agama-live token file.
+fn agama_token_file() -> Option<PathBuf> {
+    home::home_dir().map(|p| p.join(DEFAULT_AGAMA_TOKEN_FILE))
```
imobachgs, line 126: There is nothing like a default Agama token file. Actually, the `serve` token allows to pass a path through the `--generate-token` option.

rust/agama-lib/src/error.rs
```diff
@@ -18,6 +18,8 @@ pub enum ServiceError {
     // specific error will be printed too
     #[error("Error: {0}")]
     Anyhow(#[from] anyhow::Error),
+    #[error("Network client error: '{0}'")]
+    NetworkClientError(String),
```
imobachgs, line 22: A little bit generic, isn't it? Let's keep it as it is but...

rust/agama-lib/src/lib.rs
```diff
@@ -41,13 +41,34 @@ mod store;
 pub use store::Store;
 pub mod questions;
 use crate::error::ServiceError;
+use reqwest::cookie::Jar;
```
imobachgs, line 44: Merge this line and the next one, please.

rust/agama-lib/src/lib.rs
```diff
@@ -41,13 +41,34 @@ mod store;
...
+use std::sync::Arc;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 pub async fn connection() -> Result<zbus::Connection, ServiceError> {
     connection_to(ADDRESS).await
 }
 
+pub fn http_client(token: String) -> Result<reqwest::Client, ServiceError> {
+    let cookie_store = Arc::new(Jar::default());
```
imobachgs, line 55: Are you using the cookie store for anything? If not, please, disable it.

rust/agama-lib/src/network/store.rs
```diff
@@ -1,17 +1,16 @@
...
 /// Loads and stores the network settings from/to the D-Bus service.
-pub struct NetworkStore<'a> {
-    network_client: NetworkClient<'a>,
+pub struct NetworkStore {
+    network_client: NetworkClient,
 }
 
-impl<'a> NetworkStore<'a> {
-    pub async fn new(connection: Connection) -> Result<NetworkStore<'a>, ServiceError> {
+impl NetworkStore {
```
imobachgs, line 10: No lifetimes, nice!

rust/agama-lib/src/store.rs
```diff
@@ -17,19 +18,19 @@ use zbus::Connection;
...
+    network: NetworkStore,
     product: ProductStore<'a>,
     software: SoftwareStore<'a>,
     storage: StorageStore<'a>,
     localization: LocalizationStore<'a>,
 }
 
 impl<'a> Store<'a> {
-    pub async fn new(connection: Connection) -> Result<Store<'a>, ServiceError> {
+    pub async fn new(connection: Connection, client: Client) -> Result<Store<'a>, ServiceError> {
```
imobachgs, line 29: What is `client: Client`? A little bit generic. I would rename the parameter to `http_client`.

rust/agama-lib/src/lib.rs
```diff
@@ -41,13 +41,34 @@ mod store;
...
+use reqwest::{header, Client};
+use std::sync::Arc;
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 pub async fn connection() -> Result<zbus::Connection, ServiceError> {
     connection_to(ADDRESS).await
 }
 
+pub fn http_client(token: String) -> Result<reqwest::Client, ServiceError> {
```
imobachgs, line 54: np: It is fine as it is. But I would consider defining our own `HTTPClient` on top of `reqwest::Client` (using the [new type idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)) to hide the fact that we are using reqwest. But it just a matter of personal taste. What do you think?

rust/agama-server/src/network/web.rs
```diff
@@ -158,6 +160,24 @@ async fn devices(
     Ok(Json(state.network.get_devices().await?))
 }
 
+#[utoipa::path(get, path = "/network/connections/:id", responses(
+  (status = 200, description = "Get connection given by its id", body = NetworkConnection)
```
imobachgs, line 164:
```suggestion
  (status = 200, description = "Get connection given by its ID", body = NetworkConnection)
```

imobachgs (changes requested):

rust/agama-lib/src/lib.rs
teclator: No, we are not using it, but as commented in the description we could save the cookies to a file and load it when obtaining the client, but as for this PBI the focus was local access we could omit it by now.

rust/agama-lib/src/lib.rs
teclator: Do not see its benefits

rust/agama-cli/src/auth.rs
teclator: we could ignore it then, I just followed the PBI suggestions which mentioned the /run/agama/token file

imobachgs: As said, we can leave it as it is. Benefits? Well, 1) encapsultation and 2) defining our own simplified API). We follow a similar approach in the JavaScript side. But let's keep it as it is until the benefits become more clear.

rust/agama-lib/src/lib.rs
imobachgs: As said, we can leave it as it is. Benefits? Well, 1) encapsultation and 2) defining our own simplified API). We follow a similar approach in the JavaScript side. But let's keep it as it is until the benefits become more clear.

rust/agama-cli/src/auth.rs
imobachgs: For the live media, I guess we will inject that token into root's home directory. Or we can decide that we are going to use always that file (we could even have a AGAMA_RUN_DIR environment variable).

imobachgs (commented):
In general it looks rather good. We just need to agree on how to handle the token path.

rust/agama-lib/src/error.rs
imobachgs: Well, thinking again about this, I would say it is wrong. We are just throwing away the cause of the error for no good reason. Please, have a look to the [Error trait](https://doc.rust-lang.org/std/error/trait.Error.html) definition.

rust/agama-lib/src/lib.rs
```diff
@@ -41,13 +41,29 @@ mod store;
...
 use crate::error::ServiceError;
+use reqwest::{header, Client};
 
 const ADDRESS: &str = "unix:path=/run/agama/bus";
 
 pub async fn connection() -> Result<zbus::Connection, ServiceError> {
     connection_to(ADDRESS).await
 }
 
+pub fn http_client(token: String) -> Result<reqwest::Client, ServiceError> {
```
imobachgs, line 52: np: Please, reorder the functions:

```rust
pub fn connection() {} // D-Bus related
pub connection_to {} // D-Bus related
pub http_client() {} // HTTP
```

teclator:
Adapted

imobachgs (approved):

imobachgs (changes requested):
Oh, I forgot about something: did you remove the `--generate-token` option from the `.service` file? I mean [this line](https://github.com/openSUSE/agama/blob/master/rust/share/agama-web-server.service#L7).

teclator:
> Oh, I forgot about something: did you remove the `--generate-token` option from the `.service` file? I mean [this line](https://github.com/openSUSE/agama/blob/master/rust/share/agama-web-server.service#L7).

Good catch ;), removed

imobachgs (approved):

PR #1188, opened 2024-05-09
Documentation clean-up

imobachgs:
Agama's repository contains many documents related to different parts of the project. As we do not have a documentation site (yet), it is rather easy to forget about them. As a consequence, a few of them are really outdated.

Things got worse when we merged the new architecture into the `master` branch. This PR tries to update (or remove, in some cases) those documents. Additionally, we took the opportunity to reduce the size of the `README.md,` which was simply too long (check [the result](https://github.com/openSUSE/agama/blob/docs-cleanup/README.md)).


teclator (approved):
LGTM

doc/architecture.md
```diff
@@ -0,0 +1,98 @@
...
+  CLI <---> WS
+  auto --> CLI
+```
+
+## Encryption
+
+In the case of a remote installation, the communication between the clients and the server must be
+encrypted. Connecting to port 80 (HTTP) redirects the client to port 443 (HTTPS).
+
+About the certificate, Agama uses a self-signed certificate unless the user injects its own.
```
kobliha, line 73: It would be great to link to a place where this is documented.

imobachgs: You are right. The problem is that the mechanism to inject a certificate is not working yet. But @mchf is working on it :-)

doc/architecture.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Skipping the authentication
+
+When using Agama locally in the installation media, it would be unpleasant to ask for a password.
+For that reason, Agama implements a mechanism to skip the authentication step. This mechanism is
+documented in the [security document](./agama-security.md).
+
+## Links
+
+* https://bugzilla.suse.com/show_bug.cgi?id=1219688
```
kobliha, line 94: The bug links to https://github.com/openSUSE/agama/blob/architecture_2024/doc/agama-security.md - you might want to update bugzilla too when it's merged.

imobachgs: Good point. Let's update the bug because it is already merged.

PR #1199, opened 2024-05-13
Drop network D-Bus server completely

teclator:
Once we have moved the UI and the CLI to a JSON/HTTP API for handling the network configuration the D-Bus server can be removed completely.


imobachgs (changes requested):
Wow, +16/-1,384! Please, add an entry to the changes log clarifying that we are only dropping the *network D-Bus service*.

imobachgs (approved):

PR #1200, opened 2024-05-13
Fix Questions component useEffect

imobachgs:
Fix Questions component useEffect.


ancorgs (approved):
This fixed the problem for me as well. Thanks.

PR #1208, opened 2024-05-14
Storage improvements

joseivanlopez:
Several improvements:

* Small changes in HTTP API
  * _/storage/proposal/usable_devices_ (get): returns a list of SIDs instead of device names.
  * _/storage/proposal/settings_ (put): returns whether the proposal was successful calculated.
* Document storage HTTP API


rust/agama-lib/src/storage/client.rs
```diff
@@ -67,19 +66,18 @@ impl<'a> StorageClient<'a> {
         Ok(result)
     }
 
-    /// Returns the available devices
-    ///
-    /// These devices can be used for installing the system.
-    pub async fn available_devices(&self) -> Result<Vec<StorageDevice>, ServiceError> {
-        let devices: Vec<_> = self
+    pub async fn available_devices(&self) -> Result<Vec<DeviceSid>, ServiceError> {
```
imobachgs, line 69: Please, bring back some documentation.

rust/agama-server/src/storage/web.rs
```diff
@@ -150,24 +207,79 @@ async fn product_params(
...
+    responses(
+        (status = 200, description = "List of actions", body = Vec<Action>),
+        (status = 400, description = "The D-Bus service could not perform the action")
+    )
+)]
+async fn actions(State(state): State<StorageState<'_>>) -> Result<Json<Vec<Action>>, Error> {
+    Ok(Json(state.client.actions().await?))
+}
+
+/// Gets the SID (Storage ID) of the devices usable for the installation.
```
imobachgs, line 233: np: it looks weird having two "names" for the same thing: "Storage ID" vs "Device ID". But we can keep it as it is.

imobachgs (approved):

rust/agama-server/src/storage/web.rs
joseivanlopez: Maybe we could rename `DeviceSid` as `StorageId`. Anyway, I would even prefer to simply have `id` property in the devices, hidding the `sid` concept (even in the D-Bus layer).

imobachgs: Sure, no problem, just a comment.

imobachgs (approved):

PR #1211, opened 2024-05-15
remove dolomite as product won't be realized

jreidinger:
## Problem

ALP is renamed and Dolomite product won't be realized.

## Solution

Remove definition for no longer available Dolomite product.


ancorgs (approved):
LGTM. Thanks.

PR #1213, opened 2024-05-15
Do not crash if the file-based locales database is empty

imobachgs:
Improve #1205 so that it does not crash if there are no valid locales in the `/etc/agama.d/locales` file.


ancorgs (approved):
LGTM

joseivanlopez (approved):
LGTM

PR #1216, opened 2024-05-15
Added manager route for downloading logs

teclator:
## Problem

In the Cockpit times, Agama allowed to download the logs using the web UI. However, after the switch to the HTTP/JSON API, that’s not the case anymore. Agama relied on Cockpit’s file API which, for obvious reasons, is not available anymore.

Therefore the option to download the logs using our agama-web-server should be implemented bringing the functionality back.

- Trello card: - https://trello.com/c/xExmF6a2/3660-3-allow-downloading-the-logs-using-the-web-ui

## Solution

The agama-web-server has been adapted adding the /logs route in the manager HTTP/JSON API. The Web UI has been also adapted bringing back the fetchLogs() method.

Last but not least the options for showing the logs and the terminal have been removed as them were also relying on Cockpit and therefore are currently broken.


rust/agama-server/src/manager/web.rs
```diff
@@ -163,3 +167,36 @@ async fn installer_status(
...
+}
+
+async fn generate_logs() -> Result<String, Error> {
+    let random_name: String = Alphanumeric.sample_string(&mut rand::thread_rng(), 8);
+    let path = format!("/run/agama/logs_{random_name}");
+
+    Command::new("agama")
+        .args(["logs", "store", "-d", path.as_str()])
+        .status()
+        .expect("Cannot generate logs");
```
imobachgs, line 198: I think that if the call fails the client will not get any response. Or am I missing something?

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue May 16 10:48:42 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Allow to download Agama los throgh the manager HTTP API
```
imobachgs, line 4:
```suggestion
- Allow to download Agama los through the manager HTTP API
```

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu May 16 10:28:27 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix the download logs action in the web UI and drop the broken
```
imobachgs, line 4: I would put both changes in the next entry.

web/src/components/core/ShowLogButton.jsx
```diff
@@ -1,61 +0,0 @@
...
-        onClick={onClick}
-        isDisabled={isLogDisplayed}
-        icon={<Icon name="description" size="s" />}
-      >
-        {/* TRANSLATORS: button label */}
-        {_("Show Logs")}
-      </Button>
-
-      { isLogDisplayed &&
-        <FileViewer
```
imobachgs, line 51: I did not meant to drop the buttons completely (just removing the from the menu). But I am not against dropping that code. In that case, you should remove `FileViewer` too.

web/src/client/http.js
```diff
@@ -106,6 +106,15 @@ class HTTPClient {
     return response;
   }
 
+  /**
+   * @param {string} url - Endpoint URL (e.g., "/l10n/config").
+   * @return {Promise<Response>} Server response.
+   */
+  async getRaw(url) {
```
imobachgs, line 113: Instead of extending the API (which is expected to be used only here), you could just use fetch. WDYT?

teclator: In fact originally it was download, and we (@dgdavid and me) were evaluating all the options you mentioned even adding an option for specifying the content type as a parameter for the get method but didn't liked any of them at all.

By now it is only used in this case so, just using fetch should be enough.

web/src/components/core/ShowLogButton.jsx
teclator: We were also not sure, that is why a I put it in a separate commit just to revert in case.. :) Will remove it only from the menu by now as it does not hurt and probably we will bring them back soon.

imobachgs (approved):

PR #1218, opened 2024-05-16
Disable HMR for some modules (example code)

lslezak:
- This is just an example how to explicitly disable HMR
- Originally it was designed as a workaround for #1214
- But as the code might be useful in some other cases I'm keeping it here for reference


lslezak:
Closing, the real fix is in #1214

PR #1220, opened 2024-05-16
Do not fail when the storage proposal is missing

imobachgs:
* The CLI does not fails when there is not storage proposal.
* Properly detect when LVM is activated (from the CLI).


teclator (approved):
LGTM

PR #1221, opened 2024-05-16
Re-create the clients when they are updated via HMR

lslezak:
## Problem

- Updating a client via HMR does not crash anymore but the updated code is not used, you need to manually reload the page to see the changes.

## Solution

- Add a hook which re-creates the clients when the clients module is updated via HMR.

## Testing

- Tested manually with changing the `clients/storage.js` file.

### DIFF
--- a/web/src/context/installer.jsx
+++ b/web/src/context/installer.jsx
@@ -87,10 +87,27 @@ function InstallerClientProvider({
         return;
       }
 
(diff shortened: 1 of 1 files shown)


imobachgs (approved):
Really nice! It works for me just fine.

ancorgs (approved):
It also works for me

PR #1222, opened 2024-05-16
Restarting agama.service restarts the web server

imobachgs:
Restarting the `agama.servce` (through `systemctl`) causes the `agama-web-server` to fail because I gets disconnected from the D-Bus daemon. Declaring `agama-web-server` PartOf `agama.service` forces the web server to be restarted too.


ancorgs (approved):
LGTM

PR #1223, opened 2024-05-16
Tumbleweed configuration: more sensible min for generic volumes

ancorgs:
The min size for the generic volume (the one without mount point) is only relevant to populate the initial value at the UI form when the user clicks on "Add file system -> Other".

zero

Zero is a weird value for that because is not even a valid min. All the other minimums in the product (/, /home, swap...) are in the scale of Gigabytes. So 1GiB looks more sensible as starting point for the user to adjust it.

onegib

### DIFF
--- a/products.d/tumbleweed.yaml
+++ b/products.d/tumbleweed.yaml
@@ -191,6 +191,7 @@ storage:
     - filesystem: xfs
       size:
         auto: false
+        min: 1 GiB
       outline:
         required: false
         filesystems:


ancorgs:
> I do not like that now the maximum is smaller than the minimum. But if you think it is better, code-wise looks good.

Yes, that's a general problem of the form that happens always when there is no max (like when adding a `/home` volume). That form is crap... but is what we have now.

PR #1226, opened 2024-05-16
fix: Avoid implicit storage probing

joseivanlopez:
## Problem

Storage devices were implictly probed. As side effect, the UI logs some errors when it asks for the available devices for first time.

Trace:

* The *Storage* service starts and exports the D-Bus object _/org/opensuse/Agama/Storage1_.
* The method `#available_devices` is called in order to populate the corresponding D-Bus property.
* The system is automatically probed as side effect of calling to `Y2Storage::DiskAnalizer`.
* The SIDs of the reported available devices will not exist anymore once the storage probing is explicitly requested by the *Manager* service. 

## Solution

Avoid implicit probing.

### DIFF
--- a/service/lib/agama/storage/proposal.rb
+++ b/service/lib/agama/storage/proposal.rb
@@ -66,7 +66,7 @@ module Agama
       #
       # @return [Array<Y2Storage::Device>]
       def available_devices
-        disk_analyzer.candidate_disks
+        disk_analyzer&.candidate_disks || []
       end
 
       # Calculates a new proposal.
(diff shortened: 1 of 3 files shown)


service/lib/agama/storage/proposal.rb
```diff
@@ -170,15 +172,19 @@ def calculate_proposal(settings)
...
       def disk_analyzer
+        return nil unless storage_manager.probed?
+
         storage_manager.probed_disk_analyzer
       end
 
       # Devicegraph representing the system
       #
-      # @return [Y2Storage::Devicegraph]
+      # @return [Y2Storage::Devicegraph, nil]
```
jreidinger, line 184: please document when it can return nil

jreidinger (approved):

PR #1227, opened 2024-05-17
Fix download logs button

imobachgs:
## Problem

The button to download logs does not work. There are two problems:

* `response.blob()` is an asynchronous function, so `createObjectURL` receives a promise instead of a Blob.
* Fetch returns a successful promise even if the response is not OK. So the `catch` call does not handle that kind of problem. And, what is worse, it is hiding the real issue.

## Solution

Wait for `response.blob()` before using its result and log errors in `catch`.


teclator (approved):
LGTM

PR #1228, opened 2024-05-17
Minor improvements to HTTPS handling in Agama's web server

mchf:
## Problem

Purpose is to make https support more "user" friendly. Currently self-generated certificate contains only a generic name and is generated repeatedly on each start (if needed). Also need of explicit specification of user's own certificate on each (re)start is not much friendly.

## Solution

- real hostname is written into self-signed certificate
- generated self-signed certificate is stored for later use
- defined default location where to search for certificates on start

+ refactoring of some old pieces. Created struct to encapsulate Certificate related stuff.

## Testing

I (mvidner) have tested manually that the certificate (+key) is saved, and reused on service restart, and contains the host name.

## Screenshots

agama-hostname-in-certificate

### DIFF
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -79,23 +79,14 @@ struct ServeArgs {
     #[arg(long, default_value = None)]
     address2: Option<String>,
 
-    #[arg(
-        long,
-        default_value = "/etc/agama.d/ssl/key.pem",
-    )]
+    #[arg(long, default_value = "/etc/agama.d/ssl/key.pem")]
     key: Option<PathBuf>,
 
-    #[arg(
-        long,
-        default_value = "/etc/agama.d/ssl/cert.pem",
-    )]
+    #[arg(long, default_value = "/etc/agama.d/ssl/cert.pem")]
     cert: Option<PathBuf>,
 
     // Agama D-Bus address
-    #[arg(
-        long,
-        default_value = "unix:path=/run/agama/bus",
-    )]
+    #[arg(long, default_value = "unix:path=/run/agama/bus")]
     dbus_address: String,
 
     // Directory containing the web UI code
@@ -132,12 +123,14 @@ impl ServeArgs {
             let _ = certificate.write();
 
             Ok(certificate)
-        }
+        };
     }
 }
 
 /// Builds an SSL acceptor using a provided SSL certificate or generates a self-signed one
-fn ssl_acceptor(certificate: &agama_server::cert::Certificate) -> Result<SslAcceptor, openssl::error::ErrorStack> {
+fn ssl_acceptor(
+    certificate: &agama_server::cert::Certificate,
+) -> Result<SslAcceptor, openssl::error::ErrorStack> {
     let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
 
     tls_builder.set_private_key(&certificate.key)?;
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -101,6 +101,7 @@ impl ServeArgs {
     fn ssl_acceptor(&self) -> Result<SslAcceptor, openssl::error::ErrorStack> {
         let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
 
+        // use default or explicitly provided certificate if any
         if let (Some(cert), Some(key)) = (self.cert.clone(), self.key.clone()) {
             tracing::info!("Loading PEM certificate: {}", cert);
             tls_builder.set_certificate_file(PathBuf::from(cert), SslFiletype::PEM)?;
@@ -108,7 +109,12 @@ impl ServeArgs {
             tracing::info!("Loading PEM key: {}", key);
             tls_builder.set_private_key_file(PathBuf::from(key), SslFiletype::PEM)?;
         } else {
+            tracing::info!("Creating self-signed certificate");
+
+            // create a self-signed certificate if needed and store it for later use
             let (cert, key) = agama_server::cert::create_certificate()?;
+
+            agama_server::cert::write_certificate(cert.clone(), key.clone());
 
             tls_builder.set_private_key(&key)?;
             tls_builder.set_certificate(&cert)?;
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -129,9 +129,7 @@ impl ServeArgs {
 }
 
 /// Builds an SSL acceptor using a provided SSL certificate or generates a self-signed one
-fn ssl_acceptor(
-    certificate: &Certificate,
-) -> Result<SslAcceptor, openssl::error::ErrorStack> {
+fn ssl_acceptor(certificate: &Certificate) -> Result<SslAcceptor, openssl::error::ErrorStack> {
     let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
 
     tls_builder.set_private_key(&certificate.key)?;
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -6,6 +6,7 @@ use std::{
 
 use agama_lib::{auth::AuthToken, connection_to};
 use agama_server::{
+    cert::Certificate,
     l10n::helpers,
     logs::init_logging,
     web::{self, run_monitor},
@@ -97,25 +98,25 @@ struct ServeArgs {
 impl ServeArgs {
     /// Returns true of given path to certificate points to an existing file
     fn valid_cert_path(&self) -> bool {
-        self.cert.clone().is_some_and(|c| Path::new(&c).exists())
+        self.cert.as_ref().is_some_and(|c| Path::new(&c).exists())
     }
 
     /// Returns true of given path to key points to an existing file
     fn valid_key_path(&self) -> bool {
-        self.key.clone().is_some_and(|k| Path::new(&k).exists())
+        self.key.as_ref().is_some_and(|k| Path::new(&k).exists())
     }
 
     /// Takes options provided by user and loads / creates Certificate struct according to them
-    fn to_certificate(&self) -> anyhow::Result<agama_server::cert::Certificate> {
-        return if self.valid_cert_path() && self.valid_key_path() {
+    fn to_certificate(&self) -> anyhow::Result<Certificate> {
+        if self.valid_cert_path() && self.valid_key_path() {
             let cert = self.cert.clone().unwrap();
             let key = self.key.clone().unwrap();
 
             // read the provided certificate
-            agama_server::cert::Certificate::read(cert.as_path(), key.as_path())
+            Certificate::read(cert.as_path(), key.as_path())
         } else {
             // ask for self-signed certificate
-            let certificate = agama_server::cert::Certificate::new()?;
+            let certificate = Certificate::new()?;
 
             // write the certificate for the later use
             // for now do not care if writing self generated certificate failed or not, in the
@@ -123,13 +124,13 @@ impl ServeArgs {
             let _ = certificate.write();
 
             Ok(certificate)
-        };
+        }
     }
 }
 
 /// Builds an SSL acceptor using a provided SSL certificate or generates a self-signed one
 fn ssl_acceptor(
-    certificate: &agama_server::cert::Certificate,
+    certificate: &Certificate,
 ) -> Result<SslAcceptor, openssl::error::ErrorStack> {
     let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
 
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -21,7 +21,7 @@ use futures_util::pin_mut;
 use hyper::body::Incoming;
 use hyper_util::rt::{TokioExecutor, TokioIo};
 use hyper_util::server::conn::auto::Builder;
-use openssl::ssl::{Ssl, SslAcceptor, SslFiletype, SslMethod};
+use openssl::ssl::{Ssl, SslAcceptor, SslMethod};
 use tokio::sync::broadcast::channel;
 use tokio_openssl::SslStream;
 use tower::Service;
@@ -91,8 +91,11 @@ struct ServeArgs {
     )]
     cert: Option<String>,
 
-    /// The D-Bus address for connecting to the Agama service
-    #[arg(long, default_value = "unix:path=/run/agama/bus")]
+    // Agama D-Bus address
+    #[arg(
+        long,
+        default_value = "unix:path=/run/agama/bus",
+    )]
     dbus_address: String,
 
     // Directory containing the web UI code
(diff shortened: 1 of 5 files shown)


rust/agama-server/src/agama-web-server.rs
```diff
@@ -103,15 +103,22 @@ impl ServeArgs {
...
 
             tracing::info!("Loading PEM key: {}", key);
             tls_builder.set_private_key_file(PathBuf::from(key), SslFiletype::PEM)?;
         } else {
+            tracing::info!("Creating self-signed certificate");
+
+            // create a self-signed certificate if needed and store it for later use
             let (cert, key) = agama_server::cert::create_certificate()?;
 
+            // tries to write generated self-signed certificate. Nobody cares if it fails
```
imobachgs, line 119: It is not a big issue, but I think it deserves a log entry.

rust/agama-server/src/cert.rs
```diff
@@ -1,9 +1,12 @@
+use anyhow;
+use gethostname::gethostname;
 use openssl::asn1::Asn1Time;
 use openssl::bn::{BigNum, MsbOption};
 use openssl::error::ErrorStack;
 use openssl::hash::MessageDigest;
 use openssl::pkey::{PKey, Private};
 use openssl::rsa::Rsa;
+use std::{ fs, path::Path };
```
imobachgs, line 9:
```suggestion
use std::{fs, path::Path};
```

imobachgs (commented):
In general it looks good (and even better when you fix the conflict) :-) And, please, update the PR's description.

rust/agama-server/src/agama-web-server.rs
mchf: sure ... once i merge upstream ;-)

rust/agama-server/src/cert.rs
```diff
@@ -24,15 +27,35 @@ use openssl::x509::{X509NameBuilder, X509};
 //     pub write(...)
 // }
 
+pub const DEFAULT_CERT_DIR: &str = "/etc/agama.d/ssl";
```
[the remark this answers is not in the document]
imobachgs: Well, if I take openSSH as an example, `/etc/agama.d/ssl` looks like a good option.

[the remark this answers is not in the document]
mchf: Well if you go through the commits, you'll see that it originally was ```/run/....```` ;-)

Problem is that there was also request to save the cert into location which is searched by default even for user provided cert to make agama re-read it automatically e.g. on restart. In such case ```/etc/...``` makes more sense in my POV.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -103,15 +103,26 @@ impl ServeArgs {
     fn ssl_acceptor(&self) -> Result<SslAcceptor, openssl::error::ErrorStack> {
         let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
 
-        if let (Some(cert), Some(key)) = (self.cert.clone(), self.key.clone()) {
-            tracing::info!("Loading PEM certificate: {}", cert);
-            tls_builder.set_certificate_file(PathBuf::from(cert), SslFiletype::PEM)?;
+        // use default or explicitly provided certificate if any
+        if self.cert.clone().is_some_and(|c| Path::new(&c).exists()) && self.key.clone().is_some_and(|k| Path::new(&k).exists()) {
```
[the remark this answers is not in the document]
mchf: this piece is more about verifying users input which is path to certificate (```Option<String>```)  ... but i'll try to do something to make it more beautiful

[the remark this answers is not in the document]
imobachgs: Right, it depends on how you think about this function. I will try to explain.

In general, I do not like the idea of having `ServeArgs` as responsible for creating the SSL acceptor. IMHO, `ServeArgs` should be responsible only for parsing/validating. But let's assume that we will not change that as part of this PR.

About validating that the file exists, you might also want to validate (not by now but you know what I mean) that it is a valid certificate or any other stuff. IMHO, it should belong to the same `Certificate` struct. You could also have a constructor that receives the paths and returns `Result<Certificate, E>` (replacing `E` with the error type).

Additionally, encapsulating the certificate-related logic would make it easier to test (and more reusable).

rust/agama-server/src/agama-web-server.rs
```diff
@@ -76,13 +76,13 @@ struct ServeArgs {
...
         long,
-        default_value = None,
+        default_value = "/etc/agama.d/ssl/key.pem",
         help = "Path to the SSL private key file in PEM format"
     )]
     key: Option<String>,
     #[arg(
         long,
-        default_value = None,
+        default_value = "/etc//agama.d/ssl/cert.pem",
```
mvidner, line 85:
```suggestion
        default_value = "/etc/agama.d/ssl/cert.pem",
```

rust/agama-server/src/agama-web-server.rs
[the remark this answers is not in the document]
mvidner: I've made this to make it easier to read (aguably :sweat_smile: ) and avoid clones:
```rs
        let path_exists = |opt_s: &Option<String>| opt_s.as_ref().is_some_and(|s| Path::new(&s).exists());
        if path_exists(&self.cert) && path_exists(&self.key) {
```

rust/agama-server/src/cert.rs
```diff
@@ -24,15 +27,35 @@ use openssl::x509::{X509NameBuilder, X509};
...
+pub fn write_certificate(cert: X509, key: PKey<Private>) -> anyhow::Result<()> {
+    // check and create default dir if needed
+    if ! Path::new(DEFAULT_CERT_DIR).is_dir() {
+        std::fs::create_dir_all(DEFAULT_CERT_DIR)?;
+    }
+
+    if let Ok(bytes) = cert.to_pem() {
+        fs::write(Path::new(DEFAULT_CERT_DIR).join("cert.pem"), bytes)?;
+    }
+    if let Ok(bytes) = key.public_key_to_pem() {
```
mvidner, line 42: So , `key` is supposed to be the _private_ key, yet we write it with a "_public_ key to PEM" method??

mchf: already fixed in local branch, but thanks for pointing that ;-)

rust/agama-server/src/agama-web-server.rs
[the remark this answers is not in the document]
mchf: this piece is subject of refactoring (moving stuff from ServeArgs to new Certificate struct) but I'll use as much as possible

rust/agama-server/src/cert.rs
mvidner: Okay... why don't you push what you have then? No one likes reviewing outdated code :smiling_face_with_tear:

mchf: done

rust/agama-server/src/.agama-web-server.rs.swp
mvidner, line 1: This file should not be here :slightly_smiling_face:

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+    pub fn write(&self) -> anyhow::Result<()> {
+        // check and create default dir if needed
+        if !Path::new(DEFAULT_CERT_DIR).is_dir() {
+            std::fs::create_dir_all(DEFAULT_CERT_DIR)?;
+        }
+
+        if let Ok(bytes) = self.cert.to_pem() {
+            fs::write(Path::new(DEFAULT_CERT_DIR).join("cert.pem"), bytes)?;
+        }
+        if let Ok(bytes) = self.key.private_key_to_pem_pkcs8() {
```
mvidner, line 30: JFYI, for my reference, this command can parse the key and show the secret numbers `openssl pkey -in key.pem -text`

mchf: as you can see in the screenshot in the description ;-)

rust/agama-server/src/agama-web-server.rs
```diff
@@ -100,30 +98,53 @@ struct ServeArgs {
 }
 
 impl ServeArgs {
-    /// Builds an SSL acceptor using a provided SSL certificate or generates a self-signed one
-    fn ssl_acceptor(&self) -> Result<SslAcceptor, openssl::error::ErrorStack> {
-        let mut tls_builder = SslAcceptor::mozilla_modern_v5(SslMethod::tls_server())?;
+    /// Returns true of given path to certificate points to an existing file
+    fn valid_cert_path(&self) -> bool {
+        self.cert.clone().is_some_and(|c| Path::new(&c).exists())
```
imobachgs, line 103:
```suggestion
        self.cert.as_ref().is_some_and(|c| Path::new(&c).exists())
```

There is not need to clone the value in this case.

rust/agama-server/src/agama-web-server.rs
```diff
@@ -100,30 +98,53 @@ struct ServeArgs {
...
+    }
 
-        if let (Some(cert), Some(key)) = (self.cert.clone(), self.key.clone()) {
-            tracing::info!("Loading PEM certificate: {}", cert);
-            tls_builder.set_certificate_file(PathBuf::from(cert), SslFiletype::PEM)?;
+    /// Returns true of given path to key points to an existing file
+    fn valid_key_path(&self) -> bool {
+        self.key.clone().is_some_and(|k| Path::new(&k).exists())
```
imobachgs, line 108:
```suggestion
        self.key.as_ref().is_some_and(|k| Path::new(&k).exists())
```

rust/agama-server/src/agama-web-server.rs
```diff
@@ -100,30 +98,53 @@ struct ServeArgs {
...
+    }
 
-            tracing::info!("Loading PEM key: {}", key);
-            tls_builder.set_private_key_file(PathBuf::from(key), SslFiletype::PEM)?;
-        } else {
-            let (cert, key) = agama_server::cert::create_certificate()?;
+    /// Takes options provided by user and loads / creates Certificate struct according to them
+    fn to_certificate(&self) -> anyhow::Result<agama_server::cert::Certificate> {
+        return if self.valid_cert_path() && self.valid_key_path() {
```
imobachgs, line 113: No need to write the `return` if you remove the semicolon in line 129.

imobachgs, line 113: This condition is actually readable :+1:

rust/agama-server/src/agama-web-server.rs
```diff
@@ -100,30 +98,53 @@ struct ServeArgs {
...
+            let cert = self.cert.clone().unwrap();
+            let key = self.key.clone().unwrap();
 
-            tls_builder.set_private_key(&key)?;
-            tls_builder.set_certificate(&cert)?;
-        }
+            // read the provided certificate
+            agama_server::cert::Certificate::read(cert.as_path(), key.as_path())
```
imobachgs, line 118: np: Why not importing the `Certificate` instead of using the full path?

rust/agama-server/src/agama-web-server.rs
```diff
@@ -308,7 +329,7 @@ async fn serve_command(args: ServeArgs) -> anyhow::Result<()> {
     let service = web::service(config, tx, dbus, web_ui_dir).await?;
     // TODO: Move elsewhere? Use a singleton? (It would be nice to use the same
     // generated self-signed certificate on both ports.)
-    let ssl_acceptor = if let Ok(ssl_acceptor) = args.ssl_acceptor() {
+    let ssl_acceptor = if let Ok(ssl_acceptor) = ssl_acceptor(&args.to_certificate()?) {
```
imobachgs, line 332: I would say this is a good opportunity to use let-else:

```rust
    let Ok(ssl_acceptor) = ssl_acceptor(&args.to_certificate()?) else {
        return Err(anyhow::anyhow!("SSL initialization failed"));
    };
```

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+            fs::write(Path::new(DEFAULT_CERT_DIR).join("key.pem"), bytes)?;
+        }
+
+        Ok(())
+    }
+
+    /// Reads cert from given path
+    pub fn read(cert: &Path, key: &Path) -> anyhow::Result<Self> {
```
imobachgs, line 38: Here feel free to ignore the folllowing suggestion. But you might want to use the [AsRef trait](https://doc.rust-lang.org/std/convert/trait.AsRef.html):

```rust
pub fn read<P: AsRef<Path>>(cert: P, key: P) -> anyhow::Result<Self> {
```

Now it works with anything that can be converted into a reference to a `Path` (e.g., `String`, `&str`, `PathBuf`, etc.). But, as said, you can ignore this comment.

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+        let cert_bytes = std::fs::read(cert)?;
+        let key_bytes = std::fs::read(key)?;
+
+        let cert = X509::from_pem(&cert_bytes.as_slice());
+        let key = PKey::private_key_from_pem(&key_bytes.as_slice());
+
+        if let (Ok(c), Ok(k)) = (cert, key) {
```
imobachgs, line 45: np: I would say using a `match` here is more idiomatic (https://rust-lang.github.io/rfcs/0160-if-let.html). But I have to admit that this one is a pedantic comment 😅

imobachgs (commented):
IMHO the code looks much better now. Just a few comments, but nothing really relevant. Nice work! Thanks!

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+        // check and create default dir if needed
+        if !Path::new(DEFAULT_CERT_DIR).is_dir() {
+            std::fs::create_dir_all(DEFAULT_CERT_DIR)?;
+        }
+
+        if let Ok(bytes) = self.cert.to_pem() {
+            fs::write(Path::new(DEFAULT_CERT_DIR).join("cert.pem"), bytes)?;
+        }
+        if let Ok(bytes) = self.key.private_key_to_pem_pkcs8() {
+            fs::write(Path::new(DEFAULT_CERT_DIR).join("key.pem"), bytes)?;
```
mvidner, line 31: This will leave the key file world-readable.
It seems we should use https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.mode to set the file mode before writing to it

rust/agama-server/src/cert.rs
mvidner: Well no. The screenshot shows the *certificate* which is in a *different* format.

rust/agama-server/src/agama-web-server.rs
mchf: happens to me sometimes ;-)

rust/agama-server/src/.agama-web-server.rs.swp
mchf: I'll remove it once the review is over as i'll need to fix very first commit and force push. Question is how it sneaked in to it ;-)

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+        builder.set_not_before(&not_before)?;
+        let not_after = Asn1Time::days_from_now(365)?;
+        builder.set_not_after(&not_after)?;
+
+        builder.append_extension(BasicConstraints::new().critical().ca().build()?)?;
+
+        builder.append_extension(
+            SubjectAlternativeName::new()
+                // use the default Agama host name
+                // TODO: use the gethostname crate and use the current real hostname
```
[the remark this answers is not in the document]
mchf: well it partly is : https://github.com/openSUSE/agama/pull/1228/files#diff-280a7d38feb9a8c0c477b0903c149026ea053d231fa7ebe1326b4fd76eec07ecR62

may be @lslezak can say something about this ?

rust/agama-server/src/cert.rs
```diff
@@ -1,84 +1,108 @@
...
+    /// Creates a self-signed certificate
+    pub fn new() -> anyhow::Result<Self> {
+        let rsa = Rsa::generate(2048)?;
+        let key = PKey::from_rsa(rsa)?;
+
+        let hostname = gethostname()
+            .into_string()
+            .unwrap_or(String::from("localhost"));
+        let mut x509_name = X509NameBuilder::new()?;
+        x509_name.append_entry_by_text("O", "Agama")?;
```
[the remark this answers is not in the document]
mchf: ok, thanks

rust/agama-server/src/agama-web-server.rs
mchf: you know ... if you learned to use hammer .... ;-)

rust/agama-server/src/cert.rs
mchf: ok, this is at the edge of my rust skills ... will read a bit.

rust/agama-server/src/cert.rs
mchf: do we care for self generated certificate for installer which becomes useless once the installation is over?

rust/agama-server/src/cert.rs
mchf: studied ...  another piece into rust way how to handle "classes" polymorphism and so on niece to get it into my toolbox

rust/agama-server/src/cert.rs
imobachgs: I agree with @mvidner in this case. I will not hurt to set the correct permissions in this case. We are already doing that for the token so, why not?

mvidner: Ah, thanks for the tip, this method shows how it is done: https://github.com/openSUSE/agama/blob/8bebd79d8990ec690ef76ca8575c086b3044b654/rust/agama-lib/src/auth.rs#L98

service/Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    agama-yast (8)
```
imobachgs, line 4: Please, do not update the `Gemfile.lock`.

mchf:
sorry ... rebased and force pushed to get rid of of that *.swp file from the first commit

this https://github.com/openSUSE/agama/pull/1228/commits/6d1d1375721e720755fe9e9b241cc469a73bde88 piece is new (slightly modified @mvidner's approach)

imobachgs (changes requested):
Sorry, but I still see changes in the `Gemfile.lock` file.

mchf:
> Sorry, but I still see changes in the `Gemfile.lock` file.

sorry ... found one more commit which contained it ... eliminated.

imobachgs (approved):

imobachgs:
Oh, it looks like @joseivanlopez was faster and now you have a conflict to solve ;-)

mchf:
> Oh, it looks like @joseivanlopez was faster and now you have a conflict to solve ;-)

no problem ... just another small point into work stats ;-)

PR #1233, opened 2024-05-17
Remove Dolomite from agama-products-opensuse.spec

imobachgs:
After removing Dolomite, we forgot to update the spec file. See #1211.


jreidinger (approved):
Thanks, my fault

imobachgs:
No problem.

PR #1236, opened 2024-05-17
fix: Emit signal when storage propeties change

joseivanlopez:
### Problem

The value of the D-Bus properties *ProductMountPoints* and *EncryptionMethods* from the interface *org.opensuse.Agama.Storage1.Proposal.Calculator* depend on the selected product. Nevertheless, the *PropertiesChanged* signal is not emitted when the product changes. Therefore, proxies clients never update such values.

This implies that the option for FDE (Full Disk Encryption) might not be shown. See https://github.com/openSUSE/agama/pull/1004.

### Solution

* For the *ProductMountPoints* property: emit a *PropertiesChanged* signal when the product changes.
* For the *EncryptionMethods* property: emit a *PropertiesChaged* signal when software is probed. Note that the availability of certain packages is checked in order to know whether FDE is a possible encryption method.

### DIFF
--- a/service/lib/agama/dbus/clients/base.rb
+++ b/service/lib/agama/dbus/clients/base.rb
@@ -97,8 +97,8 @@ module Agama
           return if callbacks.size > 1
 
           interface_proxy = dbus_object[interface]
-          interface_proxy.on_signal(signal) do |iface, changes, invalid|
-            callbacks.each { |c| c.call(iface, changes, invalid) }
+          interface_proxy.on_signal(signal) do |*args|
+            callbacks.each { |c| c.call(*args) }
           end
         end
 
--- a/service/lib/agama/dbus/clients/base.rb
+++ b/service/lib/agama/dbus/clients/base.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2022-2023] SUSE LLC
+# Copyright (c) [2022-2024] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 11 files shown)


service/lib/agama/dbus/clients/base.rb
```diff
@@ -68,24 +68,37 @@ def service
...
-          return if @on_properties_change_callbacks[dbus_object.path].size > 1
+          return if callbacks.size > 1
 
-          dbus_properties = dbus_object["org.freedesktop.DBus.Properties"]
-          dbus_properties.on_signal("PropertiesChanged") do |interface, changes, invalid|
-            callbacks = @on_properties_change_callbacks[dbus_object.path]
-            callbacks.each { |c| c.call(interface, changes, invalid) }
+          interface_proxy = dbus_object[interface]
+          interface_proxy.on_signal(signal) do |iface, changes, invalid|
+            callbacks.each { |c| c.call(iface, changes, invalid) }
```
imobachgs, line 101: The signature looks like the one for PropertiesChanged (or am I missing something?).

service/lib/agama/dbus/clients/software.rb
```diff
@@ -36,17 +36,10 @@ class Software < Base
...
-
-          @dbus_object = service["/org/opensuse/Agama/Software1"]
-          @dbus_object.introspect
-
-          @dbus_product = service["/org/opensuse/Agama/Software1/Product"]
-          @dbus_product.introspect
-
-          @dbus_proposal = service["/org/opensuse/Agama/Software1/Proposal"]
-          @dbus_proposal.introspect
+        # @note This client is sigleton because ruby-dbus does not work properly with several
```
imobachgs, line 39:
```suggestion
        # @note This client is singleton because ruby-dbus does not work properly with several
```

service/lib/agama/dbus/clients/base.rb
joseivanlopez: Yes, let's simply pass the signal arguments to its handler.

imobachgs (approved):

PR #1237, opened 2024-05-17
Document profiling

jreidinger:
## Problem

There is no help when agama is slow and developer wants to profile it.

## Solution

Start with document that aim to summarize how to profile different layers.


jreidinger:
yeah, with move from dbus and also move to http service, this needs revisit and document new parts how to profile and identifiy what takes the most time. So lets move it back to draft when I find time to improve it.

imobachgs:
I guess we should move this stuff to the developer's documentation of the documentation site.

PR #1242, opened 2024-05-20
inject complete lshw and adapt example to work with whole lshw

jreidinger:
## Problem

There is only limited subset of hardware information for jsonnet.

## Solution

Add complete json tree from lshw and adapt example how to search for the biggest disk. Also add utils library for easier search in lshw output and enhance example.

## Testing

- *Tested manually*

### DIFF
--- a/autoinstallation/README.md
+++ b/autoinstallation/README.md
@@ -84,21 +84,11 @@ local findBiggestDisk(disks) =
   local sizedDisks = std.filter(function(d) std.objectHas(d, 'size'), disks);
   local sorted = std.sort(sizedDisks, function(x) x.size);
   sorted[0].logicalname;
-local selectClass(lshw, class) =
-  local selectClass_(parent, class) =
-    if std.objectHas(parent, 'class') && parent.class == class then
-      [ parent ]
-    else if std.objectHas(parent, 'children') then
-      std.flattenArrays(std.prune(std.map(function(x) selectClass_(x, class), parent.children )))
-    else
-      [];
-
-  local result = selectClass_(lshw, class);
-  result;
+local memory = agama.findID(agama.lshw, 'memory').size;
 
 {
   software: {
-    product: 'ALP-Dolomite',
+    product: if memory < 8000000000 then 'MicroOS' else 'Tumbleweed',
   },
   root: {
     password: 'nots3cr3t',
@@ -108,7 +98,7 @@ local selectClass(lshw, class) =
     language: 'en_US',
   },
   storage: {
-    bootDevice: findBiggestDisk(selectClass(agama.lshw, 'disk')),
+    bootDevice: findBiggestDisk(agama.selectClass(agama.lshw, 'disk')),
   },
 }
 ```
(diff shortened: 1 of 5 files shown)


autoinstallation/README.md
```diff
@@ -84,6 +84,17 @@ local findBiggestDisk(disks) =
   local sizedDisks = std.filter(function(d) std.objectHas(d, 'size'), disks);
   local sorted = std.sort(sizedDisks, function(x) x.size);
   sorted[0].logicalname;
+local selectClass(lshw, class) =
```
imobachgs, line 87: IMHO, we should offer this as some kind of helper. I would not expect the users to have to include this on their own profiles.

jreidinger: yeap, that is open question. Basically I think we should provide something like jsonnet library that is also injected into jsonnet during evaluation. At least two helpers will be useful..`selectClass` for class filtering and `findID` to find one specific id ( useful to e.g. check physical memory ).

rust/share/agama.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+    else if std.objectHas(parent, 'children') then
+      std.flattenArrays(std.prune(std.map(function(x) selectClass_(x, class), parent.children )))
+    else
+      [];
+
+  local result = selectClass_(lshw, class);
+  result,
+
+// function go throught lshw output and returns object with given "id" or null if not found.
+findID(lshw, id)::
```
imobachgs, line 16: About the name of the function, you are not "finding" and ID, but an object. So I would rename it to `findById` or something similar (or `findObject` or whatever).

rust/share/agama.libsonnet
```diff
@@ -0,0 +1,26 @@
+// function go throught lshw output and enlist only given class.
+// Basically it is same as calling `lshw -class <class>`.
+selectClass(lshw, class)::
```
imobachgs, line 3: `selectClass` looks good enough. But perhaps to be consistent with my suggestion below we could use `findClass` or something similar.jsdo

rust/share/agama.libsonnet
imobachgs: Please, document the arguments (type and so on).

rust/share/agama.libsonnet
imobachgs: Please, document the arguments.

imobachgs (commented):
I like the solution but I think we need better names for those helpers as they are part of the public API.

imobachgs (approved):

imobachgs (changes requested):
Please, update the changes file.

rust/package/agama.changes
```diff
@@ -7,7 +13,7 @@ Fri May 17 09:52:25 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 Tue May 16 12:48:42 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
 
 - Allow to download Agama los throgh the manager HTTP API
```
imobachgs, line 15: I know, it does not belong to this PR but let's fix this :-)

```suggestion
- Allow to download Agama los through the manager HTTP API
```

imobachgs (approved):

PR #1243, opened 2024-05-21
MicroOS Desktop community does not support installations via Agama

sysrich:
## Problem

The MicroOS Desktop community (aka Kalpa/Aeon) do not support installations via Agama

The only installer that is currently supported for Aeon (formerly MicroOS Desktop GNOME) is 'tik' (github.com/sysrich/tik)

Kalpa will be retaining YaST as it's only installer until it decides its future.

## Solution

Remove references to the MicroOS Desktop (a product name no longer used) in Agama - do not replace with Kalpa or Aeon


imobachgs (commented):
It looks good but, please, could you add an entry to [the changes file](https://github.com/openSUSE/agama/blob/master/products.d/agama-products-opensuse.changes)?

Thanks!

imobachgs (approved):

ancorgs:
> The only installer that is currently supported for Aeon (formerly MicroOS Desktop GNOME) is 'tik' (github.com/sysrich/tik)

So I guess you should also remove this role. https://github.com/yast/skelcd-control-MicroOS/blob/master/control/control.MicroOS.xml#L258

PR #1244, opened 2024-05-21
Better AutoYaST import support

imobachgs:
* Extend the converter to read some missing sections: `networking`, `language`, `keyboard`, `timezone`, and `suse_register`.
* Document how supported sections are handled, including what is missing.
* Improve test coverage.

## Tasks

- [x] Read and document the `networking` section
- [x] Read and document the `language` section
- [x] Read and document the `keyboard` section
- [x] Read and document the `timezone` section
- [x] Read and document the `suse_register` section

### DIFF
--- a/doc/autoyast.md
+++ b/doc/autoyast.md
@@ -372,6 +372,18 @@ About the `slp_discoverty` element, Agama does not support [SLP] at all?
 
 [SLP]: https://documentation.suse.com/sles/15-SP5/single-html/SLES-administration/#cha-slp
 
+| AutoYaST                         | Supported | Agama                     | Comment                                         |
+| -------------------------------- | --------- | ------------------------- | ----------------------------------------------- |
+| do_registration                  | Never     |                           | The system is registered if there is a reg code |
+| email                            | Yes       | product.registrationEmail |                                                 |
+| install_updates                  | Never     |                           |                                                 |
+| reg_code                         | Yes       | product.registrationCode  |                                                 |
+| reg_server                       | Planned   |                           |                                                 |
+| reg_server_cert                  | Undecided |                           | It is not clear how addons will work            |
+| reg_server_cert_fingerprint      | Planned   |                           |                                                 |
+| reg_server_cert_fingerprint_type | Planned   |                           |                                                 |
+| slp_discovery                    | Never     |                           |                                                 |
+
 ### `timezone`
 
 | AutoYaST | Supported | Agama         | Comment |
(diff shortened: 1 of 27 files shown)


doc/autoyast.md
```diff
@@ -186,34 +383,53 @@ About the `slp_discoverty` element, Agama does not support [SLP] at all?
...
 [SLP]: https://documentation.suse.com/sles/15-SP5/single-html/SLES-administration/#cha-slp
 
+| AutoYaST                         | Supported | Agama                     | Comment                              |
+| -------------------------------- | --------- | ------------------------- | ------------------------------------ |
+| do_registration                  | Yes       |                           | The section is ignored if `false`    |
+| email                            | Yes       | product.registrationEmail |                                      |
+| install_updates                  | Never     |                           |                                      |
+| reg_code                         | Yes       | product.registrationCode  |                                      |
+| reg_server                       | Planned   |                           |                                      |
+| reg_server_cert                  | Undecided |                           | It is not clear how addons will work |
```
jreidinger, line 393: I am not sure how addones from comment is related to registration server certificate. I think it is more related to RMT

service/lib/agama/autoyast/converter.rb
```diff
@@ -111,41 +116,19 @@ def tmp_profile_path
...
+        product = Agama::AutoYaST::ProductReader.new(profile)
+        l10n = Agama::AutoYaST::L10nReader.new(profile)
+        storage = Agama::AutoYaST::StorageReader.new(profile)
+
+        user.read
+          .merge(root.read)
+          .merge(software.read)
+          .merge(product.read)
+          .merge(storage.read)
+          .merge(l10n.read)
```
jreidinger, line 131: NP, just matter of taste, but I would write it in more easier way for extend like:
```
READERS = ["user", "root", "software", ... ] # extend list to add support for more sections
def export_profile(profile)
  sections.each_with_object({}) do |section, result|
    require "agama/autoyast/#{section}_reader}"
    klass = "#{section}_reader".split("_").map(&:capitalize).join
    reader = Agama::AutoYaST.get_const(klass).new(profile)
    result.merge(reader.read)
   end
end
```

this way it is really just single line to extend it with new section as opposite to current code which require change in three places.

jreidinger: btw I can even reduce it to no line in converter if we have something like `agama/autoyast` file that contain all classes and then filter it by known readers.

doc/autoyast.md
imobachgs: Ouch, wrong line 😅

service/lib/agama/autoyast/converter.rb
imobachgs: Good point! I decided to go with the `READERS` variable approach because it is more explicit.

jreidinger (approved):

PR #1246, opened 2024-05-22
Fix development scripts

lslezak:
## Problem

- The development script did not work anymore after dropping Cockpit

## Solution

- Update the scripts

## Testing

- Tested manually in a Tumbleweed instance
- After running the scripts I could properly access running Agama

### DIFF
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -32,11 +32,12 @@ sudosed() {
 
 # if agama is already running -> stop it
 $SUDO systemctl list-unit-files agama.service &>/dev/null && $SUDO systemctl stop agama.service
+$SUDO systemctl list-unit-files agama-web-server.service &>/dev/null && $SUDO systemctl stop agama-web-server.service
 
 # Ruby services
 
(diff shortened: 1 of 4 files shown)


setup-services.sh
```diff
@@ -32,11 +32,12 @@ sudosed() {
 
 # if agama is already running -> stop it
 $SUDO systemctl list-unit-files agama.service &>/dev/null && $SUDO systemctl stop agama.service
+$SUDO systemctl list-unit-files agama-web-server.service &>/dev/null && $SUDO systemctl stop agama-web-server.service
 
 # Ruby services
 
 # Packages required for Ruby development (i.e., bundle install).
-$SUDO zypper --non-interactive --gpg-auto-import-keys install \
```
lslezak, line 39: The `--gpg-auto-import-keys` option is only needed when installing packages from unknown 3rd party repositories. It is not needed when using the official openSUSE repositories. (It could actually accidentally accept some forged GPG key...)

imobachgs (approved):

PR #1248, opened 2024-05-23
improve patterns selector

jreidinger:
## Problem

There is broken counter how in search box for pattern selector. Also there is no clear indication that nothing is matched.

## Solution

See screenshots how it is improved:

Old:
  counter:
broken_counter

  empty result:
broken_empty

New:
  counter:
new_counter

  empty result (attempt 1):
new_empty

   empty result (attempt 2):
new_empty

### DIFF
--- a/web/src/components/software/PatternSelector.jsx
+++ b/web/src/components/software/PatternSelector.jsx
@@ -20,7 +20,7 @@
  */
 
 import React, { useCallback, useEffect, useState } from "react";
-import { SearchInput } from "@patternfly/react-core";
+import { SearchInput, Label } from "@patternfly/react-core";
 
 import { Section, Selector } from "~/components/core";
 import { _ } from "~/i18n";
@@ -143,7 +143,7 @@ function PatternSelector({ patterns, onSelectionChanged = noop }) {
     </div>
   );
 
-  const selector = sortGroups(groups).map((groupName) => {
+  let selector = sortGroups(groups).map((groupName) => {
     const selectedIds = groups[groupName].filter((p) => p.selectedBy !== SelectedBy.NONE).map((p) =>
       p.name
     );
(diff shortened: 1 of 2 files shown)


web/src/components/software/PatternSelector.jsx
```diff
@@ -166,6 +166,12 @@ function PatternSelector({ patterns, onSelectionChanged = noop }) {
     );
   });
 
+  if (selector.length === 0) {
+    selector = (
+      <Label>{_("No results")}</Label>
```
dgdavid, line 171: I'would not use the `<Label>` component for this kind of things. Not even the Patternfly guys [would use it either](https://www.patternfly.org/components/label/design-guidelines#usage) :wink:

Not saying we have to follow strictly their design guidelines, but in this case I'm more aligned with them.

Use just plain old [HTML](https://html.spec.whatwg.org/multipage/) like [`<span>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element), [`<b>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-b-element) or even [`<strong>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-strong-element) if we considered it is something should be emphasise.

Also, do not hesitate of adding more text to make it clear. I cannot help too much with texts, but "No matches" looks a bit poor to me.

web/src/components/software/PatternSelector.jsx
```diff
@@ -166,6 +166,12 @@ function PatternSelector({ patterns, onSelectionChanged = noop }) {
     );
   });
 
+  if (selector.length === 0) {
+    selector = (
+      <b>{_("None of the patterns match the text.")}</b>
+    );
+  }
```
dgdavid, line 173: NP:

```suggestion
  if (selector.length === 0) {
    selector = (
      <b>{_("None of the patterns match the filter.")}</b>
    );
  }
```

dgdavid, line 173: NP:

Sorry, I didn't realize before, but it is actually not needed to redefine the selector variable. You can just change `{selector}` to `{selector.length ? selector : <NoResultFoundThing />}` below. But don't worry ant keep it as it is to minimize the conflicts with the new-ui branch.

dgdavid (approved):
LGTM.

I just added to NP comments for the record, not asking for changing them now.

Thanks for taking care.

PR #1249, opened 2024-05-23
Use a timeout when calling `setxkbmap`

lslezak:
## Problem

- When calling `setxkbmap` in development the `root` user might not have access to the `:0` X display.
- Maybe there is XDM running or another user is logged in. In that case `setxkbmap` prints
  ```
  Authorization required, but no authorization protocol specified
  ```
  error message in an infinite loop and gets stuck.
- Additionally the $DISPLAY might be already set so we should honor that value (e.g. when you manually start the web server from terminal)

## Solution

- Use the [subprocess](https://crates.io/crates/subprocess) crate to limit the execution time
- Use the current `DISPLAY` value when already set

### SSH X Forwarding Problem

We need to set the display for the `setxkbmap` call  when running on the Live ISO because when running Agama from systemd service the `DISPLAY` is not set but we need to access the locally running web browser so we have to explicitly set the `:0` value.

When the web server is started manually from a terminal we should keep the current value if it is already set. But using the current `DISPLAY` value might behave a bit unexpectedly when you login to the system via SSH with X forwarding enabled. In that case the `setxkbmap` call will query and set (!) keyboard of the remote X server. That might be quite confusing. 

Sure, that is a corner case,  but if you select some exotic layout with complete different mapping like Dvorak then it might not be trivial to restore your original layout. When using X forwarding the `DISPLAY` is set to something like `localhost:10.0`, so maybe we should use it only when it starts with `:` character which means the local X server? :thinking: 

## Testing

- Tested manually

### DIFF
--- a/rust/agama-server/src/l10n/l10n.rs
+++ b/rust/agama-server/src/l10n/l10n.rs
@@ -1,5 +1,4 @@
 use std::env;
-use std::ffi::OsStr;
 use std::io;
 use std::process::Command;
 use std::time::Duration;
@@ -31,7 +30,7 @@ const SETXKBMAP_TIMEOUT: u64 = 3;
 
 // helper function which runs a command with timeout and collects it's standard
 // output
-fn run_with_timeout(cmd: &[impl AsRef<OsStr>], timeout: u64) -> Result<Option<String>, PopenError> {
+fn run_with_timeout(cmd: &[&String], timeout: u64) -> Result<Option<String>, PopenError> {
     // start the subprocess
     let mut process = Popen::create(
         cmd,
@@ -47,7 +46,7 @@ fn run_with_timeout(cmd: &[impl AsRef<OsStr>], timeout: u64) -> Result<Option<St
         .wait_timeout(Duration::from_secs(timeout))?
         .is_none()
     {
-        tracing::warn!("Command timed out!");
+        tracing::warn!("Command {:?} timed out!", cmd);
         // if the process is still running after the timeout then terminate it,
         // ignore errors, there is another attempt later to kill the process
         let _ = process.terminate();
@@ -65,7 +64,7 @@ fn run_with_timeout(cmd: &[impl AsRef<OsStr>], timeout: u64) -> Result<Option<St
     let (out, err) = process.communicate(None)?;
 
     if let Some(err_str) = err {
-        if err_str.len() > 0 {
+        if !err_str.is_empty() {
             tracing::warn!("Error output size: {}", err_str.len());
         }
     }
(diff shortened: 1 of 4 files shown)


rust/agama-server/src/l10n/l10n.rs
```diff
@@ -21,6 +25,60 @@ pub struct L10n {
...
+            ..Default::default()
+        },
+    )?;
+
+    // wait for it to finish or until the timeout is reached
+    if process
+        .wait_timeout(Duration::from_secs(timeout))?
+        .is_none()
+    {
+        tracing::warn!("Command timed out!");
```
imobachgs, line 50: which command?

rust/agama-server/src/l10n/l10n.rs
```diff
@@ -21,6 +25,60 @@ pub struct L10n {
...
+        }
+
+        return Err(PopenError::LogicError("Timeout reached"));
+    }
+
+    // get the collected stdout/stderr
+    let (out, err) = process.communicate(None)?;
+
+    if let Some(err_str) = err {
+        if err_str.len() > 0 {
```
imobachgs, line 68: Why not use the [is_empty](https://doc.rust-lang.org/std/string/struct.String.html#method.is_empty) function?

imobachgs (changes requested):
It looks good. Thanks!

However, I would prefer that function to live in some kind of "helpers" module because it is not related to localization and it is not that "small" (apart from bringing quite some elements into scope).

rust/agama-server/src/l10n/l10n.rs
lslezak: Ah, still getting familiar with Rust... :smiley:

lslezak:
> However, I would prefer that function to live in some kind of "helpers" module 

I was thinking about that as well, but then I decided to keep it here. It is quite specialized, if we want to move it we should make it more generic and support more use cases (like reading stderr, getting the exit code, etc...).

imobachgs:
> > However, I would prefer that function to live in some kind of "helpers" module
> 
> I was thinking about that as well, but then I decided to keep it here. It is quite specialized, if we want to move it we should make it more generic and support more use cases (like reading stderr, getting the exit code, etc...).

Well, it looks generic enough for our use case :wink: But I am fine with keeping it there for a while. Just let's try to not forget that it is there so we do not implement something similar again 😅

imobachgs (commented):
When working with arguments, it is usually preferred to use `&str` instead of `&String`. With that approach, you could simplify things a bit. Please, see https://github.com/openSUSE/agama/commit/8301f3e5953b4c320f1cec1173d7ca950a8d8519 and feel free to cherry-pick it if you want.

imobachgs (approved):

imobachgs (approved):

PR #1254, opened 2024-05-24
Reconnect the HTTP client websocket automatically

teclator:
## Problem

When the current WebSocket is closed or there is an error the frontend is not aware of it and does not react properly just stop receiving notifications. It would be nice to reconnect silently reporting in case that it is impossible to reconnect after a while.

- Trello Card: https://trello.com/c/Kqaez0RH/3632-2-agama-adapt-connection-error-handling

## Solution

When the WebSocket is closed we will try to reconnect during a while. From that point in advance the state of the socket will be unrecoverable needing a user manual reconnection.

### DIFF
--- a/web/src/client/http.js
+++ b/web/src/client/http.js
@@ -27,6 +27,23 @@
  */
 
 /**
+ * Enum for the WebSocket states.
+ *
+ *
+ */
+
+const SocketStates = Object.freeze({
+  CONNECTED: 0,
+  CONNECTING: 1,
+  CLOSING: 2,
+  CLOSED: 3,
+  UNRECOVERABLE: 4,
+});
+
+const MAX_ATTEMPTS = 10;
+const ATTEMPT_INTERVAL = 2000;
+
+/**
  * Agama WebSocket client.
  *
  * Connects to the Agama WebSocket server and reacts on the events.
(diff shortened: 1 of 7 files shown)


web/src/client/http.js
```diff
@@ -38,11 +55,65 @@ class WSClient {
...
+    return (this.wsState() !== SocketStates.UNRECOVERABLE);
+  }
+
+  isConnected() {
+    return (this.wsState() === SocketStates.CONNECTED);
+  }
+
+  subscribe_to_events() {
+    this.client.onopen = () => {
+      console.log("Connected websocket");
```
jreidinger, line 85: I would say other way it will be better "Websocket connected"

web/src/client/http.js
```diff
@@ -38,11 +55,65 @@ class WSClient {
...
+      this.reconnectAttempts = 0;
+      if (this.timeout !== undefined) clearTimeout(this.timeout);
```
jreidinger, line 87: I think that condition is not needed. See https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout Just call it always

web/src/client/http.js
```diff
@@ -38,11 +55,65 @@ class WSClient {
...
+      return this.dispatchOpenEvent();
+    };
+
     this.client.onmessage = (event) => {
       this.dispatchEvent(event);
     };
-    this.handlers = [];
+
+    this.client.onclose = () => {
+      console.log(`Closed WebSocket`);
```
jreidinger, line 97: also here I would write "Websocket closed"

web/src/client/http.js
```diff
@@ -38,11 +55,65 @@ class WSClient {
...
+
+    this.client.onerror = (e) => {
+      console.error("WebSocket error:", e);
+      this.dispatchErrorEvent();
+    };
+  }
+
+  connect(attempt = 0) {
+    this.reconnectAttempts = attempt;
+    if (attempt > 10) {
```
jreidinger, line 110: here it should use constant defined above

teclator: yep, it was temporal and already done ;)

web/src/components/core/ServerError.jsx
```diff
@@ -28,19 +28,19 @@ import { locationReload } from "~/utils";
...
         <EmptyState variant="xl">
           <EmptyStateHeader
-            titleText={_("Cannot connect to D-Bus")}
+            titleText={_("Cannot connect to Agama server")}
             headingLevel="h2"
             icon={<EmptyStateIcon icon={ErrorIcon} />}
           />
           <EmptyStateBody>
-            {_("Could not connect to the D-Bus service. Please, check whether it is running.")}
+            {_("Could not connect to the Agama server. Please, check whether it is running.")}
```
imobachgs, line 43: I would say this part of the message is not useful at all. The first sentence is basically a repetition of the bigger "Cannot connect to...". And, as a user, how can I check whether the service is running?

By now, I would remove the first sentence. About the second part, we should link to some page explaining how to proceed. But that's out of the scope of this card.

teclator: I agree

PR #1255, opened 2024-05-24
fix: use sudo to create files in setup-services.sh

imobachgs:
Use `sudo` to create PAM and products configuration in the `setup-services.sh` script.


jreidinger (approved):

lslezak (approved):
Ouch, I forgot about the sudo case...

Thanks for the fix!

PR #1263, opened 2024-05-27
feat(schema): guided and AutoYaST storage proposals

joseivanlopez:
Schema definition for the guided and AutoYaST storage proposals:

~~~json
{
  "storage": {
    "guided": { }
  },
  "legacyAutoyastStorage": [ ]
}
~~~

Based on [auto_storage.md](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md) which is expected to change over time.

Bonus: This includes a commit revisiting how `title` and `description` are used, to be more compliant with https://json-schema.org/understanding-json-schema/reference/annotations

Note for reviewers: you can easily check your JSON config files by using `agama profile validate config.json`

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -271,7 +271,6 @@
               "title": "Target device",
               "oneOf": [
                 {
-                  "type": "string",
                   "enum": ["disk", "newLvmVg"]
                 },
                 {
@@ -320,19 +319,16 @@
                 },
                 "method": {
                   "title": "Method used to create the encryption devices",
-                  "type": "string",
                   "enum": ["luks2", "tpm_fde"]
                 },
                 "pbkdFunction": {
                   "title": "Password-based key derivation function to use for LUKS2",
-                  "type": "string",
                   "enum": ["pbkdf2", "argon2i", "argon2id"]
                 }
               }
             },
             "space": {
               "title": "Policy to find space for the new partitions",
-              "type": "string",
               "enum": [ "delete", "resize", "keep" ]
             },
             "volumes": {
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
   "$id": "https://github.com/openSUSE/agama/blob/master/rust/agama-lib/share/profile.schema.json",
   "title": "Profile",
   "description": "Profile definition for automated installation",
@@ -96,8 +96,7 @@
                 "type": "array",
                 "items": {
                   "title": "Connection addresses",
-                  "type": "string",
-                  "additionalProperties": false
+                  "type": "string"
                 }
               },
               "nameservers": {
@@ -105,8 +104,7 @@
                 "items": {
                   "title": "Nameservers",
                   "description": "IPv4 and/or IPv6 are allowed.",
-                  "type": "string",
-                  "additionalProperties": false
+                  "type": "string"
                 }
               },
               "wireless": {
@@ -149,8 +147,7 @@
                     "type": "array",
                     "items": {
                       "title": "A list of the interfaces or connections to be bonded",
-                      "type": "string",
-                      "additionalProperties": false
+                      "type": "string"
                     }
                   }
                 }
@@ -164,32 +161,28 @@
                     "type": "array",
                     "items": {
                       "title": "A list of kernel command line arguments to match",
-                      "type": "string",
-                      "additionalProperties": false
+                      "type": "string"
                     }
                   },
                   "interface": {
                     "type": "array",
                     "items": {
                       "title": "A list of interface names to match",
-                      "type": "string",
-                      "additionalProperties": false
+                      "type": "string"
                     }
                   },
                   "driver": {
                     "type": "array",
                     "items": {
                       "title": "A list of driver names to match",
-                      "type": "string",
-                      "additionalProperties": false
+                      "type": "string"
                     }
                   },
                   "path": {
                     "type": "array",
                     "items": {
                       "title": "A list of paths to match against the ID_PATH udev property of devices",
-                      "type": "string",
-                      "additionalProperties": false
+                      "type": "string"
                     }
                   }
                 }
@@ -288,7 +281,7 @@
                       "items": { "type": "string" }
                     }
                   },
-                  "required": ["newLvmVg"] 
+                  "required": ["newLvmVg"]
                 }
               ]
             },
@@ -329,7 +322,64 @@
             },
             "space": {
               "title": "Policy to find space for the new partitions",
-              "enum": [ "delete", "resize", "keep" ]
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "policy": {
+                  "enum": ["delete", "resize", "keep", "custom"]
+                },
+                "actions": {
+                  "type": "array"
+                }
+              },
+              "if": {
+                "properties": {
+                  "policy": { "const": "custom" }
+                }
+              },
+              "then": {
+                "required": ["policy", "actions"],
+                "properties": {
+                  "actions": {
+                    "type": "array",
+                    "items": {
+                      "anyOf": [
+                        {
+                          "type": "object",
+                          "required": ["forceDelete"],
+                          "additionalProperties": false,
+                          "properties": {
+                            "forceDelete": {
+                              "title": "Device to delete",
+                              "type": "string"
+                            }
+                          }
+                        },
+                        {
+                          "type": "object",
+                          "required": ["resize"],
+                          "additionalProperties": false,
+                          "properties": {
+                            "resize": {
+                              "title": "Device to allow resizing",
+                              "type": "string"
+                            }
+                          }
+                        }
+                      ]
+                    }
+                  }
+                }
+              },
+              "else": {
+                "required": ["policy"],
+                "properties": {
+                  "actions": {
+                    "type": "array",
+                    "maxItems": 0
+                  }
+                }
+              }
             },
             "volumes": {
               "title": "Set of volumes (file systems) to create",
@@ -345,8 +395,8 @@
                     "required": ["path"],
                     "properties": {
                       "path": {
-                      "title": "Mount path",
-                      "type": "string"
+                        "title": "Mount path",
+                        "type": "string"
                       },
                       "options": {
                         "title": "Options to add to the fourth field of fstab",
@@ -392,19 +442,13 @@
                       { "$ref": "#/$defs/sizeString" },
                       { "$ref": "#/$defs/sizeInteger" },
                       {
+                        "description": "Lower size limit and optionally upper size limit",
                         "type": "array",
+                        "items": {
+                          "$ref": "#/$defs/sizeValue"
+                        },
                         "minItems": 1,
-                        "maxItems": 2,
-                        "prefixItems": [
-                          {
-                            "title": "Mandatory lower size limit",
-                            "$ref": "#/$defs/sizeValue"
-                          },
-                          {
-                            "title": "Optional upper size limit",
-                            "$ref": "#/$defs/sizeValue"
-                          }
-                        ]
+                        "maxItems": 2
                       },
                       {
                         "type": "object",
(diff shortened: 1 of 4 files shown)


rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+          "properties": {
+            "target": {
+              "title": "Target device",
+              "description": "Can be 'disk', 'newLvmVg' or an object with exactly one key-value pair",
+              "type": ["string", "object"],
+              "additionalProperties": false,
+              "properties": {
+                "disk": { "type": "string" },
+                "newLvmVG": { "type": "string" }
+              }
```
joseivanlopez, line 271: "oneOf": [
      { "type": "string", "pattern": "^(disk|newLvm)$" },
      { "type": "object", "properties": { "disk": "string" } },
      { "type": "object", "properties": { "newLvm": "string" } }
    ]

joseivanlopez, line 271:
```suggestion
          "oneOf": [
            {
              "type": "string",
              "pattern": "^(disk|newLvm)$"
            },
            {
              "type": "object",
              "properties": { "disk": "string" },
              "required": ["disk"],
              "additionalProperties": false
            },
            {
              "type": "object",
              "properties": { "newLvm": "string" },
              "required": ["newLvm"],
              "additionalProperties": false
            }
          ]
```

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+              "properties": {
+                "configure": {
+                  "title": "Whether to configure partitions for booting",
+                  "type": "boolean"
+                },
+                "device": {
+                  "title": "Device to use for booting",
+                  "type": "string"
+                }
+              }
```
joseivanlopez, line 286:
```suggestion
              },
              "required": ["configure"],
              "additionalProperties": false
```

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+                "disk": { "type": "string" },
+                "newLvmVG": { "type": "string" }
+              }
+            },
+            "boot": {
+              "title": "Configuration of the boot settings",
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "configure": {
```
joseivanlopez, line 278: Should configure be required? Or could we assume `true` by default?

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+                    "additionalProperties": false,
+                    "required": ["path"],
+                    "properties": {
+                      "path": {
+                      "title": "Mount path",
+                      "type": "string"
+                      },
+                      "options": {
+                        "title": "Options to add to the fourth field of fstab",
+                        "type": "string"
```
joseivanlopez, line 334: Should be an array?

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+                      }
+                    }
+                  },
+                  "filesystem": {
+                    "title": "File system of the volume",
+                    "type": ["string", "object"],
```
joseivanlopez, line 340: IMHO, it is more clear and easiest to document if we use `oneOf`.

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,156 @@
...
+                      "max": {
+                        "title": "Optional upper size limit",
+                        "type": ["integer", "string"]
+                      }
+                    }
+                  },
+                  "target": {
+                    "title": "Location of the resulting file system",
+                    "description": "Can be 'default' or an object with exactly one key-value pair",
+                    "type": ["string", "object"],
```
joseivanlopez, line 385: This is another candidate for `oneOf`.

rust/agama-lib/share/profile.schema.json
ancorgs: Done. Mostly the same (but I overlooked your comment so I did "push -f" instead of accepting your suggestion, sorry).

rust/agama-lib/share/profile.schema.json
ancorgs: I though we assumed "true" by default.

rust/agama-lib/share/profile.schema.json
ancorgs: Fixed in an amended commit.

rust/agama-lib/share/profile.schema.json
ancorgs: Done in the amended commit (as said, I force-pushed because I overlooked your comments).

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,194 @@
...
+                "pbkdFunction": {
+                  "title": "Password-based key derivation function to use for LUKS2",
+                  "type": "string",
+                  "enum": ["pbkdf2", "argon2i", "argon2id"]
+                }
+              }
+            },
+            "space": {
+              "title": "Policy to find space for the new partitions",
+              "type": "string",
```
joseivanlopez, line 328: I think `type` can be omitted here.

rust/agama-lib/share/profile.schema.json
```diff
@@ -251,22 +251,194 @@
...
+                        "additionalProperties": false,
+                        "properties": {
+                          "snapshots": {
+                            "title": "Whether Btrfs snapshots should be configured",
+                            "type": "boolean"
+                          }
+                        }
+                      }
+                    }
+                  },
```
joseivanlopez, line 372:
```suggestion
                  "filesystem": {
                    "title": "File system of the volume",
                    "oneOf": [
                      {  "enum": ["btrfs", "xfs", "ext3", ...] },
                      {
                        "type": "object",
                        "additionalProperties": false,
                        "properties": {
                          "btrfs": {
                            "title": "Specification of a Btrfs file system",
                            "type": "object",
                            "additionalProperties": false,
                            "properties": {
                              "snapshots": {
                                "title": "Whether Btrfs snapshots should be configured",
                                "type": "boolean"
                              }
                            }
                          }
                        }
                      }
                    ]
                  },
```

rust/agama-lib/share/profile.schema.json
ancorgs: If it does not hurt, I think I prefer to have it explicit.

ancorgs: Oh, you meant it's redundant with "enum". That's indeed true.

ancorgs: Done (also in some other places).

rust/agama-lib/share/profile.schema.json
```diff
@@ -219,54 +221,222 @@
...
+                      "min": {
+                        "title": "Mandatory lower size limit",
+                        "type": ["integer", "string"]
+                      },
+                      "max": {
+                        "title": "Optional upper size limit",
+                        "type": ["integer", "string"]
+                      }
+                    }
+                  },
```
joseivanlopez, line 397:
```suggestion
                  "size": {
                    "title": "Size limits",
                    "description": "It can be 'auto', a fixed size (as string or integer) or a set of size limits",
                    "oneOf": [
                      { "const": "auto" },
                      { "$ref": "#/$defs/units" },
                      { "$ref": "#/$defs/bytes" },
                      {
                        "type": "array",
                        "items": { "#/$defs/value" },
                        "minItems": 1,
                        "minItems": 2
                      },
                      {
                        "type": "object",
                        "additionalProperties": false,
                        "required": ["min"],
                        "properties": {
                          "min": {
                            "title": "Mandatory lower size limit",
                            "type": { "#/$defs/value" }
                          },
                          "max": {
                            "title": "Optional upper size limit",
                            "type": { "#/$defs/value" }
                          }
                        }
                      }
                    ],
                    "$defs": {
                      "units": {
                        "type": "string",
                        "pattern": "^\d+(\.\d+)?\s*(GiB|MiB|Kib)"
                      },
                      "bytes": {
                        "type": "integer",
                        "minimum": 0
                      },
                      "value": {
                        "oneOf": [
                          { "$ref": "#/$defs/units" },
                          { "$ref": "#/$defs/bytes" }
                        ]
                      }
                    }
                  },
```

mvidner:
examples/profile_tw.json fails to validate with this: `bootDevice` was unexpected. Should we update the profile or the schema?

mvidner:
... and now that I actually read the schema, it is obvious that the profile needs updating :)

mvidner:
Do you have any example profiles demonstrating the new schema parts which should be added?

rust/agama-lib/share/profile.schema.json
```diff
@@ -224,54 +219,314 @@
...
+              ]
+            },
+            "boot": {
+              "title": "Configuration of the boot settings",
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "configure": {
+                  "title": "Whether to configure partitions for booting",
+                  "type": "boolean"
```
mvidner, line 295:
```suggestion
                  "type": "boolean",
                  "default": "true"
```

https://json-schema.org/understanding-json-schema/reference/annotations

rust/agama-lib/share/profile.schema.json
```diff
@@ -224,54 +219,314 @@
...
+                  "policy": { "const": "custom" }
+                }
+              },
+              "then": {
+                "required": ["policy", "actions"],
+                "properties": {
+                  "actions": {
+                    "type": "array",
+                    "items": {
+                      "anyOf": [
```
mvidner, line 346: In this schema we're using both `anyOf` and `oneOf`, with `or` and `xor` semantics respectively, as I've just learned from https://json-schema.org/understanding-json-schema/reference/combining
AFAICT either of these would work for any place in our schema, so we should be consistent.

rust/agama-lib/share/profile.schema.json
```diff
@@ -224,54 +219,314 @@
...
+                        "required": ["btrfs"]
+                      }
+                    ]
+                  },
+                  "size": {
+                    "title": "Size limits",
+                    "oneOf": [
+                      { "const": "auto" },
+                      { "$ref": "#/$defs/sizeString" },
+                      { "$ref": "#/$defs/sizeInteger" },
```
mvidner, line 443: We could replace these 2 lines with `sizeValue`, right?

rust/agama-lib/share/profile.schema.json
```diff
@@ -224,54 +219,314 @@
...
+                      {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "properties": { "device": { "type": "string" } },
+                        "required": ["device"]
+                      },
+                      {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "properties": { "filesystem": { "type": "string" } },
```
mvidner, line 497: Filesystem as a target? To a storage non-expert like me this does not make sense. We should explain it in `title` or `description`

```json
                    "target": {
                        "filesystem": "..."
                    }                
```

mvidner (changes requested):
After printing the schema out and reading all of it, it makes sense, thanks!

Definitely add some examples please.

rust/agama-lib/share/profile.schema.json
joseivanlopez: I think using `anyOf` vs `oneOf` has really meaning if the object is an array. And you are right, in our case we can replace all `oneOf` by `anyOf`. Let's be consistent. Thanks!

rust/agama-lib/share/profile.schema.json
joseivanlopez: Yes. Thanks!

joseivanlopez:
> Do you have any example profiles demonstrating the new schema parts which should be added?

Added. Thanks!

joseivanlopez:
> And maybe add validating the examples to our CI tests too?

Good idea. If you don't mind, I will create a separate card for that. IMHO, adding it to CI is out-of-scope in this PR.

rust/agama-lib/share/profile.schema.json
```diff
@@ -1,61 +1,62 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
```
mvidner, line 2:
```suggestion
  "$comment": "based on doc/auto_storage.md",
  "$schema": "https://json-schema.org/draft/2019-09/schema",
```

mvidner (approved):
Thanks!

PR #1264, opened 2024-05-27
service: Avoid reading product mount points too early

joseivanlopez:
## Problem

Changes in https://github.com/openSUSE/agama/pull/1236 introduced a bug reading the product mount points.

The list of mount points was read just after selecting a new product. At that time, the *storage* service has not updated its config yet, so it still uses the previous product.

Note that the *manager* service calls to storage `#probe` when a new product is selected, and the *storage* service updates its config as part of the call to `#probe`. For that reason, the product mount points cannot be updated until probe is done.

https://bugzilla.suse.com/show_bug.cgi?id=1225348

## Solution

Do not update the product mount points when the product changes. The list of product mount points is automatically refreshed once the probing is done.

### DIFF
--- a/service/lib/agama/dbus/storage/manager.rb
+++ b/service/lib/agama/dbus/storage/manager.rb
@@ -60,7 +60,6 @@ module Agama
         def initialize(backend, logger)
           super(PATH, logger: logger)
           @backend = backend
-          @product_mount_points = read_product_mount_points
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

kobliha:
Your change has done the job. Now only / or /var can be used for Micro OS. Thx!

PR #1265, opened 2024-05-27
feat(rust): read the password from stdin

imobachgs:
The `agama auth` command is more complex than it should be. Instead of having a "switch" to read from a file or from the CLI itself, it now reads the password from the stdin and, if that's not possible, asks the user.

### DIFF
--- a/rust/agama-cli/src/auth.rs
+++ b/rust/agama-cli/src/auth.rs
@@ -3,10 +3,12 @@ use clap::{arg, Args, Subcommand};
 use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
 use std::fs;
 use std::fs::File;
-use std::io;
+use std::io::{self, IsTerminal};
 use std::io::{BufRead, BufReader};
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
+
+use crate::error::CliError;
 
 const DEFAULT_JWT_FILE: &str = ".agama/agama-jwt";
 const DEFAULT_AGAMA_TOKEN_FILE: &str = "/run/agama/token";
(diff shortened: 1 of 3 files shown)


rust/agama-cli/src/auth.rs
```diff
@@ -3,32 +3,32 @@ use clap::{arg, Args, Subcommand};
...
 const DEFAULT_AGAMA_TOKEN_FILE: &str = "/run/agama/token";
 const DEFAULT_AUTH_URL: &str = "http://localhost/api/auth";
 const DEFAULT_FILE_MODE: u32 = 0o600;
 
 #[derive(Subcommand, Debug)]
 pub enum AuthCommands {
-    /// Login with defined server. Result is JWT stored locally and made available to
-    /// further use. Password can be provided by commandline option, from a file or it fallbacks
-    /// into an interactive prompt.
+    /// Authenticate with Agama's server and store the credentials
```
jreidinger, line 20: well, it is nice to have help, but then there misterous LoginArgs...I would also kill it to not confuse users and readers of manual.

rust/agama-cli/src/auth.rs
```diff
@@ -1,34 +1,34 @@
...
 const DEFAULT_FILE_MODE: u32 = 0o600;
 
 #[derive(Subcommand, Debug)]
 pub enum AuthCommands {
-    /// Login with defined server. Result is JWT stored locally and made available to
-    /// further use. Password can be provided by commandline option, from a file or it fallbacks
-    /// into an interactive prompt.
-    Login(LoginArgs),
-    /// Release currently stored JWT
+    /// Authenticate with Agama's server and store the credentials
```
jreidinger, line 20: I would document that it can read password from stdin or it will ask on tty. with example like `agama auth login < /tmp/root_pwd`

jreidinger (approved):

PR #1266, opened 2024-05-28
Storage AutoYaST backward compatibility

joseivanlopez:
Agama is going to offer new settings for configuring the storage devices, see https://github.com/openSUSE/agama/pull/1256. That new settings will be applied by loading a config file (e.g., `$ agama config load myconfig.json`). Nevertheless, Agama is also going to temporary support the SLE15 AutoYaST schema to some extent, in order to make the transition smoother. For that, the json config file can indicate a "legacyAutoyastStorage" key equivalent to the partitioning section of the AutoYaST profile:

~~~
"legacyAutoyastStorage": [
  {
    "device": "/dev/vda",
    "use": "all",
    "partitions": [...]
  }
]
~~~ 

This PR implements the support for the "legacyAutoyastStorage" settings, making possible to calculate an AutoYaST proposal. Converting the "partitioning" section from a XML AutoYaST profile to a "legacyAutoyastStorage" section of a json config file is out of the scope and it will be addressed in a follow-up.

Note:  The AutoYaST guided partitioning will not be supported, see https://doc.opensuse.org/projects/autoyast/#CreateProfile-Guided-Partitioning.


ancorgs:
Closed in favor of #1284 (same content but cleaner history)

PR #1270, opened 2024-05-28
feat(CLI): Agama profile import command

jreidinger:
## Problem

There are many steps in autoinstallation processing that is splitted over multiple command line arguments. It is problematic if user use shell script based autoinstallation and need to all of those steps.

## Solution

Introduce new "agama profile import" command that do all autoinstallation processing. Only install part is not done. This allows user to use hooks or tuning before real installation begins.


rust/agama-cli/src/profile.rs
```diff
@@ -45,15 +56,55 @@ fn validate(path: String) -> anyhow::Result<()> {
...
     let evaluator = ProfileEvaluator {};
     evaluator
-        .evaluate(Path::new(&path))
+        .evaluate(Path::new(&path), stdout())
         .context("Could not evaluate the profile".to_string())?;
     Ok(())
 }
 
-pub fn run(subcommand: ProfileCommands) -> anyhow::Result<()> {
+async fn import(url: String, dir: Option<PathBuf>) -> anyhow::Result<()> {
```
imobachgs, line 69: The other commands (`download` and `evaluate`) rely on a separate struct. If you do not want to build a struct for this, it is fine, but at least I would move the logic to the same separate module.

I would like to keep the functions here to handle the details that are specific to the commands (setting default values, etc.).

rust/agama-cli/src/profile.rs
```diff
@@ -45,15 +56,55 @@ fn validate(path: String) -> anyhow::Result<()> {
...
+        evaluator
+            .evaluate(&output_path, fd)
+            .context("Could not evaluate the profile".to_string())?;
+        output_path = output_dir.join("profile.json");
+    }
+
+    let output_path_str = output_path.to_str().context("Failed to get output path")?.to_string();
+    // TODO: optional skip of validation
+    validate(output_path_str.clone())?;
+    crate::config::run(crate::config::ConfigCommands::Load { path: output_path_str }, crate::printers::Format::Json).await?;
```
imobachgs: Oh, it looks like we're calling the same binary again with different options. I'm not sure I like this approach.

Perhaps we might think about building a `Config` API that can be used in `config::run`. Let's keep it by now but, please, add a FIXME.

rust/agama-cli/src/profile.rs
```diff
@@ -45,15 +56,55 @@ fn validate(path: String) -> anyhow::Result<()> {
...
+    let reader = ProfileReader::new(url.as_str())?;
+    let contents = reader.read()?;
+    let mut output_path = output_dir.join(output_file);
+    let mut output_fd = File::create(output_path.clone())?;
+    output_fd.write_all(contents.as_bytes())?;
+    if output_file.ends_with(".sh") {
+        let err = Command::new("bash")
+        .args([output_path.to_str().context("Wrong path to shell script")?])
+        .exec();
+        println!("Error: {}", err);
```
imobachgs, line 88: So it will always print this string, right?

rust/agama-cli/src/profile.rs
```diff
@@ -45,15 +61,64 @@ fn validate(path: String) -> anyhow::Result<()> {
...
+    let output_dir = dir.unwrap_or_else(|| tmpdir.into_path());
+    let reader = ProfileReader::new(url.as_str())?;
+    let contents = reader.read()?;
+    let mut output_path = output_dir.join(output_file);
+    let mut output_fd = File::create(output_path.clone())?;
+    output_fd.write_all(contents.as_bytes())?;
+    if output_file.ends_with(".sh") {
+        let err = Command::new("bash")
+            .args([output_path.to_str().context("Wrong path to shell script")?])
+            .exec();
```
imobachgs, line 87: `output()` would give you `stdout`, `stderr`, etc.

rust/agama-cli/src/profile.rs
```diff
@@ -45,15 +61,64 @@ fn validate(path: String) -> anyhow::Result<()> {
...
+    if output_file.ends_with(".jsonnet") {
+        let fd = File::create(output_dir.join("profile.json"))?;
+        let evaluator = ProfileEvaluator {};
+        evaluator
+            .evaluate(&output_path, fd)
+            .context("Could not evaluate the profile".to_string())?;
+        output_path = output_dir.join("profile.json");
+    }
+
+    let output_path_str = output_path
```
imobachgs, line 100: I would expect something with the `_str` suffix to be of type `&str` :wink:

imobachgs (commented):
Just a few comments. I have the impression that splitting the `import` function into smaller steps would make things easier follow.

rust/agama-cli/src/profile.rs
```diff
@@ -13,6 +15,15 @@ pub enum ProfileCommands {
 
     /// Evaluate a profile, injecting the hardware information from D-Bus
     Evaluate { path: String },
+
+    /// Import autoinstallation profile from given location into agama configuration
+    /// 
+    /// This is top level command that do all autoinstallation processing beside starting
+    /// installation. Unless there is a need to inject additional commands between processing
+    /// use this command instead of set of underlaying commands.
```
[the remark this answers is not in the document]
jreidinger: did you mean `underlying`?

rust/agama-cli/src/profile.rs
jreidinger: in fact it is not same binary...I just call method that implements `config load`. Otherwise I will call just `agama config load` via Command, but I think this will be faster?

rust/agama-cli/src/profile.rs
jreidinger: no. It is printed only if `exec()` failed. Exec replace current process with the new one...and only if it failed, it continues with execution.

rust/agama-cli/src/profile.rs
jreidinger: it should not be needed as exec keep current fds so identical stdin, stdout and stderr. It basically replace process.

rust/agama-cli/src/profile.rs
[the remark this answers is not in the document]
jreidinger:
```suggestion
    /// use this command instead of set of underlying commands.
```

rust/agama-cli/src/profile.rs
jreidinger: yeah, reason is that config load has it inside this run command - https://github.com/openSUSE/agama/blob/master/rust/agama-cli/src/config.rs#L89

imobachgs: Yes, I know the reason :-) and I dislike it. I will add this to the list of things to fix in the Rust codebase.

imobachgs (approved):

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed May 29 07:15:37 UTC 2024 - Josef Reidinger <jreidinger@suse.com>
+
+- CLI: Add new command "agama profile import" that do whole
```
imobachgs, line 4:
```suggestion
- CLI: Add new command "agama profile import" that does the whole
```

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
...
+  autoinstallation processing and loads configuration.
```
```suggestion
  autoinstallation processing and loads the configuration.
```

imobachgs (approved):

imobachgs (approved):

PR #1271, opened 2024-05-28
refactor(rust): replace rpassword with inquire

imobachgs:
[Inquire](https://github.com/mikaelmello/inquire) is a library for building interactive prompts. It offers many "widgets" (check the demo!) and it looks like a better solution for our use case.

This PR replaces [rpassword](https://github.com/conradkleinespel/rpassword) with Inquire.


mvidner:
I wanted to review this and started by trying out `agama auth show`...

I found it quite confusing how it shows `~/.local/agama/token` after I've used `agama auth login` but falls back to `/run/agama/token` (after `agama auth logout`), which looks the same but doesn't actually provide access: `agama config show` says Invalid authentication token: ExpiredSignature.

imobachgs:
I would open a new issue to improve the `agama auth` command. Some ideas:

* `auth login` should not generate a new token if the user is already logged in (the token is valid).
* `auth show` should display not only the token, but perhaps where it comes from.

Perhaps @mchf have some additional ideas.

Having said that, I think this PR is unrelated to those issues.

rust/agama-cli/src/auth.rs
```diff
@@ -37,15 +38,35 @@ pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
...
+            Ok(Validation::Invalid("The password cannot be blank.".into()))
+        } else {
+            Ok(Validation::Valid)
+        }
+    };
+
+    Password::new("Please, introduce the root password:")
+        .with_validator(validator)
+        .without_confirmation()
+        .with_help_message("Press <esc> to exit.")
```
mvidner, line 65: I don't see the purpose of this. It helps the user to distinguish entering an empty password from not entering a password?

rust/agama-cli/src/auth.rs
```diff
@@ -37,15 +38,35 @@ pub async fn run(subcommand: AuthCommands) -> anyhow::Result<()> {
...
         buffer
     };
     Ok(password)
 }
 
+/// Asks interactively for the password.
+fn ask_password() -> Result<String, CliError> {
+    let validator = |input: &str| {
+        if input.is_empty() {
+            Ok(Validation::Invalid("The password cannot be blank.".into()))
```
mvidner, line 56: This is reasonable for *setting* a password (along with more nontrivial validations).
I don't see the purpose for this, for *verifying* a password.

mvidner (commented):
Sorry, I don't see how this helps. :smiling_face_with_tear: 
Well, the prompt ends up being colored, that is nice,
and there is a help text but I argue that the feature which is describes is superfluous.

rust/agama-cli/src/auth.rs
imobachgs: We are just making sure that the password is not empty.

rust/agama-cli/src/auth.rs
imobachgs: Sorry, I do not get the question

imobachgs: The main point is to switch from `rpassword` to `inquire`, which is a more general library for handling user interaction (`rpassword` is only for passwords).

mvidner: I am nitpicking on the existence of this help message. When I remove the help, I can still cancel the operation (and get a CliError::InteractivePassword).

I, as a user, am used to canceling password input by simply entering a garbage or empty password. (Unless the counterparty is a Bank that penalizes all incorrect attempts by $100 and a 3 hour delay until the next attempt, this is a work of fiction :rofl: )

For my personal UX, it is better to hide the help message, because it just draws my attention away towards a useless feature.

rust/agama-cli/Cargo.toml
```diff
@@ -25,8 +25,9 @@ zbus = { version = "3", default-features = false, features = ["tokio"] }
 tokio = { version = "1.33.0", features = ["macros", "rt-multi-thread"] }
 async-trait = "0.1.77"
 reqwest = { version = "0.11", features = ["json"] }
-rpassword = "7.3.1"
 url = "2.5.0"
+home = "0.5.9"
```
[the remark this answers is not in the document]
imobachgs: Good point, it is an unwanted side-effect of rebasing the changes. It is used in `agama-lib`, so no need to be included here.

mvidner (approved):

PR #1272, opened 2024-05-29
fix(rust): avoid a long-lived probing request

imobachgs:
The `Probe` D-Bus function is a blocking one. It causes the call `POST /api/manager/probe` to be a long-lived request which is causing some problems (see #1254).

The proposed fix is to return immediately and run the request as a separate Tokio task. After all, we will get an update on the status through a WebSocket. If needed, we could also emit an error through the WebSocket.


teclator (approved):
LGTM

PR #1274, opened 2024-05-29
Show a different message error when the password is wrong

teclator:
## Problem

The login form does not differentiate between an authentication error or an server error.

## Solution

Show a different error depending on the response status code.

## Screenshots

|Authentication Error | Server error | 
| --- | --- |
| image | Screenshot from 2024-06-03 08-39-40 |

### DIFF
--- a/web/src/context/auth.jsx
+++ b/web/src/context/auth.jsx
@@ -37,12 +37,19 @@ function useAuth() {
   return context;
 }
 
+const AuthErrors = Object.freeze({
+  SERVER: "server",
+  AUTH: "auth",
+  OTHER: "other"
+});
+
 /**
  * @param {object} props
  * @param {React.ReactNode} [props.children] - content to display within the provider
  */
 function AuthProvider({ children }) {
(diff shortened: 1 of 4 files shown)


web/src/context/auth.jsx
```diff
@@ -37,22 +37,37 @@ function useAuth() {
...
 
   const login = useCallback(async (password) => {
     const response = await fetch("/api/auth", {
       method: "POST",
       body: JSON.stringify({ password }),
       headers: { "Content-Type": "application/json" },
     });
+
     const result = response.status === 200;
+    if ((response.status >= 500) && (response.status < 600)) {
```
imobachgs, line 62: I would say there is nothing like a 6xx error code.

imobachgs (commented):
It looks good. But, please, update the changes file.

web/src/context/auth.jsx
teclator: Them are not, but we only wanted 5XX error codes according to the standard definition... :)

PR #1282, opened 2024-06-03
fix(dbus): refresh AvailableProducts property

imobachgs:
The HTTP/JSON API uses a proxy to access the products interface. When the language changes, the `AvailableProducts` property is silently updated (not really updated because it gets calculated each time we try to access it).

The proxy does not know that this property changed and returns the old value. We can:

* Stop caching those values (in the HTTP/JSON API).
* Fix the D-Bus service to report that the `AvailableProducts` changed (this one).
* Decide that `AvailableProperties` is a method instead of a property.

This PR is a tentative fix I have not tried in a real ISO.

## To do

- [ ] After logging in, the wrong translations still appear.
- [ ] Fix unit tests

### DIFF
--- a/rust/agama-lib/src/proxies.rs
+++ b/rust/agama-lib/src/proxies.rs
@@ -182,3 +182,13 @@ trait Issues {
     #[dbus_proxy(property)]
     fn all(&self) -> zbus::Result<Vec<(String, String, u32, u32)>>;
 }
+
+#[dbus_proxy(
+    interface = "org.opensuse.Agama1.Locale",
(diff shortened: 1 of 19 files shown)


mvidner:
Testing this in a container...
It kind of works, the product descriptions are shown in the new language, but only if I change the language from the main page.
If I change it in the Product page, it stays in the old language even after going back and forth, or changing the product.

imobachgs:
Thanks for testing. It is still a work in progress, but I will work on that after the new UI is merged.

PR #1286, opened 2024-06-05
testing_using_container: fix port forwarding

mvidner:
--network=host tries to use privileged ports on the host so it fails when run as user. Go back to port forwarding.

Also, let podman pick a number and print it.


mvidner:
Ah, also the doc file should be updated like in #1268

mvidner:
@jreidinger you still use this, right?
@lslezak if you want to find out why port forwarding does not work for you I am happy to help

jreidinger (approved):

PR #1287, opened 2024-06-05
refactor(rust): extract token handling to a agama_lib::auth

imobachgs:
This PR aims to bring the token related code to a single place. It introduces an `AuthToken` struct
that offers a simplistic API to token management.

Additionally, it introduces a behaviour change: if the /run/agama/token is available and readable,
the CLI uses such a token.

### DIFF
--- a/rust/agama-server/tests/service.rs
+++ b/rust/agama-server/tests/service.rs
@@ -1,6 +1,7 @@
 pub mod common;
 
-use agama_server::web::{generate_token, MainServiceBuilder, ServiceConfig};
+use agama_lib::auth::AuthToken;
+use agama_server::web::{MainServiceBuilder, ServiceConfig};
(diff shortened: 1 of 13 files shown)


rust/agama-server/tests/service.rs
```diff
@@ -77,8 +78,8 @@ async fn test_access_protected_route() -> Result<(), Box<dyn Error>> {
 // TODO: The following test should belong to `auth.rs`.
```
jreidinger, line 78: is this TODO still valid?

jreidinger (approved):

rust/agama-server/tests/service.rs
imobachgs: Yes, it is still valid. It is unrelated to this change.

PR #1288, opened 2024-06-05
Live ISO: set the root password from boot command line or interactively

lslezak:
## Problem

- The Live ISO uses a static well known root password, that is insecure

## Solution

This adds several boot command line options for setting the root password:

- `agama.password=<password>` - set a plain text password, not really secure as every process can read it from `/proc/cmdline` but it is still far better than a predefined known password
- `agama.password_hash=<password_hash>` - set a hashed password (generated by `mkpasswd` or `openssl passwd`), this is more secure but difficult to type manually as the hashed password is usually quite long, suitable for environments where you can prepare the boot command line before booting (like in PXE boot)
- `agama.password_systemd` - interactively ask for the password during boot using the `systemd-ask-password` tool (simple prompt but should work also over a serial console)
- `agama.password_dialog` - interactively ask for the password during boot using an interactive dialog (more fancy, but might not work well over a serial console)

## Testing

- Tested manually in VirtualBox and KVM virtual machines
- The requested password is set properly and can be used in both Agama and SSH

## Additional Fixes

- Use `set -e` in `config.sh`, there was a hidden problem with activating the `spicevdagend` service, it was failing unnoticed for quite some time... :scream: 
- Disable the snapper snapshot cleanup services, if you left the password dialog running for long time enough then starting the timer services would mess the screen. Moreover these are not need and it is potentially dangerous to run such services from a Live ISO.
- Removed the `chown` workarounds, not needed anymore as we build the tarball using the `--owner=0 --group=0` options so the owner is properly set already in the archive.
- Disabled the YaST units, not needed at all, that might make the boot process a tiny bit faster :smiley: 

(But that gave me an idea about using the YaST Firstboot package for the Agama setup. We have almost full YaST in the Live ISO already. Using `dialog` is fine for simple things but if we ever need to do some more complicated setup then using YaST Firstboot might be a possible solution...)

## Notes

- In the interactive cases the SSH and Agama web server are blocked until you enter the password, that avoids (mis)using the default predefined password before a new one is set.
- Starting an interactive session via systemd is quite tricky, the boot process normally continues and the systemd progress messages scroll over the console possibly destroying the displayed content. The workaround is to run the interactive service as late as possible, ideally after all unblocked services are finished. That is achieved by using the `After=` option. If the screen later gets broken by another service the solution should be to add it into the `After=` list.
- Additionally printing the kernel messages to the console is disabled when the dialog is running
- The dialog contains the "Press Ctrl+L to refresh the screen" hint at the top in case the output is messed up by some background process
- The `systemd-ask-password` tool has the same problem so the same workaround is applied there as well. (I hoped it is better integrated into the systemd boot process, but there is nothing special about it regarding the terminal handling.)

## TODO

- [ ] Add documentation (will be added later in a separate PR)
- [ ] Update changes (the same here)

### Entering a password using dialog
agama_password_dialog

### Entering a password using systemd
agama_password_systemd

### DIFF
--- a/live/root/etc/systemd/system/agama-password-dialog.service
+++ b/live/root/etc/systemd/system/agama-password-dialog.service
@@ -1,0 +1,45 @@
+[Unit]
+Description=Interactively set the Agama/root password in a dialog
+
+# before starting the SSH and Agama server so they use the new password
+Before=sshd.service
+Before=agama-web-server.service
+# before X11 because it switches the terminal to VT7
+Before=x11-autologin.service
+
+# copied from YaST2-Second-Stage.service
+Before=getty@tty1.service
+Before=getty@tty2.service
+Before=getty@tty3.service
+Before=getty@tty4.service
+Before=getty@tty5.service
+Before=getty@tty6.service
+Before=serial-getty@hvc0.service
+Before=serial-getty@sclp_line0.service
+Before=serial-getty@ttyAMA0.service
+Before=serial-getty@ttyS0.service
+Before=serial-getty@ttyS1.service
+Before=serial-getty@ttyS2.service
+Before=serial-getty@ttysclp0.service
+
+# start at the end to avoid overwriting the screen with systemd messages
+After=agama.service
+After=modprobe@drm.service
+
+# kernel command line option
+ConditionKernelCommandLine=agama.password_dialog
+
+[Service]
+Type=oneshot
+Environment=TERM=linux
+ExecStartPre=dmesg --console-off
+ExecStart=agama-password --dialog
+ExecStartPost=dmesg --console-on
+TTYReset=yes
+TTYVHangup=yes
+StandardInput=tty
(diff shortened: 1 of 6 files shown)


live/root/etc/systemd/system/agama-password-dialog.service
```diff
@@ -0,0 +1,45 @@
+[Unit]
+Description=Interactively set the Agama/root password in a dialog
+
+# before starting the SSH and Agama server so they use the new password
+Before=sshd.service
+Before=agama-web-server.service
```
imobachgs, line 6: np: I would try to be consistent with the style in `agama-password-cmdline.service` and use:

```
Before=sshd.service agama-web-server.service
```

or

```
Before=sshd.service
Before=agama-web-server.service
```

Whatever you prefer, but the same.

imobachgs (approved):

PR #1290, opened 2024-06-05
feat(live): Set the root password from ISO metadata

lslezak:
## Problem

Avoid using the well known default root password in the Live medium.

## Solution

Allow modifying the default root password in the ISO medium by users.  A new default root password can be embedded into the ISO file metadata using this command:
```
tagmedia --add-tag "agama_password=$((openssl passwd -6) | base64 -w 0)" <agama.iso>
```
This will embed a SHA512 hashed password into the ISO application area. This area is already used for storing the medium checksum, if you run the `checkmedia` command to verify the medium integrity then the expected checksum is read from this metadata. You can check its content by running `tagmedia <agama.iso>`.

The ISO application area is pretty small (512 bytes), but there is still enough space to embed a hashed root password.  Unfortunately it needs to be Base64 encoded otherwise `tagmedia` then reports an error about unsupported format.

If the root password is set by other options on the boot command line (see #1288) then they will override this ISO file default.

The advantage of this solution is that it does not need any special tweaks and you do not root permissions to modify the ISO image. This solution is suitable for mass deployment using the same physical medium. You can  modify the ISO, dump it on an USB stick and then install several machines using your specific root password.

## Testing

- Tested manually, the embedded password is correctly used by Agama and SSH.
- If the password is set by the other boot options from #1288 then they will take precedence.

## TODO

Later:

- [ ] Create documentation
- [ ] Update changes

### DIFF
--- a/live/root/etc/systemd/system/agama-password-cmdline.service
+++ b/live/root/etc/systemd/system/agama-password-cmdline.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Set the Agama/root password from kernel command line
+
 # before starting the SSH and Agama server so they use the new password
 Before=sshd.service
 Before=agama-web-server.service
--- a/live/root/etc/systemd/system/agama-password-iso.service
+++ b/live/root/etc/systemd/system/agama-password-iso.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Set the Agama/root password from ISO application area
+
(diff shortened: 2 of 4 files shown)


imobachgs (approved):
LGTM. I actually checked that `tagmedia` (from the `checkmedia` package) is included in the Live image. Thanks!

PR #1291, opened 2024-06-06
fix(live): Fixed RPM cleanup

lslezak:
## Problem

- ISO build on S390 fails.

## Solution

- The removed packages might not be installed on all architectures, ignore the failures to not fail whole ISO build.


teclator (approved):
LGTM

PR #1292, opened 2024-06-06
feat(live): Automatically generate the root password

lslezak:
## Problem

- Using a well known default `linux` password is insecure

## Solution

- Generate a random password during boot
- Print it to the console

## Testing

- Tested manually

## Notes

- Updated documentation

## Screenshots

agama_password_random


doc/live_iso.md
```diff
@@ -49,3 +70,158 @@ Notes:
...
+
+```sh
+mkpasswd
+```
+
+and then enter the password on the command line prompt.
+
+#### OpenSSL
+
+Alternatively you can use the `openssl passwd` command from the openSSL package.
```
imobachgs, line 210: openssl or OpenSSL

doc/live_iso.md
```diff
@@ -49,3 +70,158 @@ Notes:
...
+It offers less encryption methods but on the other hand it should be basically
+installed in every system.
+
+:warning: *Warning: By default it uses a weak encryption method (DES or MD5
```
imobachgs, line 214: GitHub implements a nicer warning message (we are using them already):

```
> [!WARNING]
> Put your message here...
```

imobachgs (commented):
Just a few comments. Otherwise, it looks good.

doc/live_iso.md
```diff
@@ -49,3 +70,158 @@ Notes:
...
+be suitable for a different use case.
+
+### Boot Command Line
+
+You can define the password directly on the boot command line. There are two
+options:
+
+* Use `agama.password=<password>` with a plain text password. This is the
+  easiest way but on the other hand it is not secure as every process running on
+  the machine can read the password from the `/proc/cmdline` file. But for
```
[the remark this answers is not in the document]
ancorgs: Agreed. I wouldn't say the note about this being insecure or restricted to test deployments is a bit over the top.

doc/live_iso.md
```diff
@@ -49,3 +70,158 @@ Notes:
...
+# replace the agama.iso name with your image name
+tagmedia --remove-tag agama_password agama.iso
+```
+
+:information_source: *Note: The image usually already contains some other tags,
+like the checksums for verifying the medium integrity. Do not touch them!*
+
+### Random Password as a Fallback
+
+When no password is specified or the entering the password interactively was
```
ancorgs, line 171: "or the entering the password" sounds weird. I guess there is a left-over "the".

ancorgs:
I added a few minor comments, but I have to say the whole password management you implemented looks AWESOME. :clap: 

That being said, I have doubts about using the `agama` namespace for it. Being precise, this has nothing to do with Agama itself, is setting the root password of the int-sys. Moreover, we tried really hard in the past to draw a line on people's minds between the installer (Agama/YaST) and the inst-sys (Agama-Live/Linuxrc+Installation-Images). This blurs the line again.

What about using `agama-live` (or any other name that really refers to the int-sys) as namespace for all this?

ancorgs:
> What about using `agama-live` (or any other name that really refers to the int-sys) as namespace for all this?

Thinking it twice, I wouldn't even use `agama-live` because this is so generic and reusable for any other live image out there that I don't think we should use "Agama" as part of the name.

doc/live_iso.md
lslezak: Nice, there are supported even some more info boxes.
https://github.com/orgs/community/discussions/16925

lslezak:
I made the boot scripts less tied to Agama and use the `live` prefix for the boot options.

I have retested everything again, all works fine.

## Updated Screenshots

### Password Dialog
agama_password_dialog2

### Systemd Propmpt
agama_password_systemd2

### Random Password
agama_password_random2

PR #1297, opened 2024-06-10
New terminal

jreidinger:
## Problem

From cockpit times we still miss terminal in web UI.

## Solution

Try to add it back.

## Testing

- *Tested manually*

## Screenshots


dgdavid (commented):
From the UI POV, I'd say it is enough as it is having in mind that we need to cook a more appealing Terminal component. But let's do it once the `new-ui-proposal` branch gets merged into `master`.

dgdavid:
We should keep an eye on https://github.com/patternfly/react-console which might have the UI pieces we will need. See https://github.com/patternfly/react-console/blob/main/packages/module/src/components/SerialConsole/SerialConsole.tsx and https://github.com/patternfly/react-console/blob/main/packages/module/src/components/SerialConsole/XTerm.tsx

Check https://www.patternfly.org/extensions/react-console too

dgdavid:
This PR is way behind master branch. Let's close it for now and consider opening a new one in the future if we decide to reimplement a terminal in the installer web interface.

PR #1298, opened 2024-06-10
fix(ui): use Label component

joseivanlopez:
Drop `Tag` component in favor of patternfly `Label`.

Screenshot from 2024-06-10 15-46-03


web/src/components/storage/ProposalResultTable.jsx
```diff
@@ -89,21 +90,21 @@ const DeviceCustomSize = ({ item, devicesManager }) => {
   const sizeBefore = isResized ? devicesManager.systemDevice(device.sid).size : item.size;
 
   return (
-    <div className="split">
+    <Flex direction={{ default: "row" }} gap={{ default: "gapXs" }}>
+      <DeviceSize item={item} />
       <If
         condition={isResized}
         then={
-          <Tag variant="orange">
```
dgdavid, line 96: Thus, <Tag> component is now only used by storage/device-utils.jsx. Something to check and drop

web/src/assets/styles/blocks.scss
```diff
@@ -510,31 +510,6 @@ table.proposal-result {
     );
   }
   /** End of temporary hack */
-
-  @media (width > 768px) {
-    th.details-column {
-      padding-inline-start: calc(60px + var(--spacer-smaller) * 2);
-    }
-
-    td.details-column {
```
dgdavid, line 519: Do you mind dropping only the reference to these CSS classes in ProposalResultSection.jsx when defining the const `columns`? Thanks in advance.

dgdavid (commented):
LGTM, just minor comments

web/src/components/storage/ProposalResultTable.jsx
joseivanlopez: Dropped.

dgdavid (approved):

PR #1301, opened 2024-06-10
feat(web): add real data to the overview page

imobachgs:
The data shown in the overview in the new UI branch is just hard-coded. This PR aims to add some real data to it. At this point we are only considering three sections: localization, storage and software.

<details>
<summary>The overview page displaying real data</summary>

Captura desde 2024-06-11 00-03-24
</details>

## (Not) Adopting a state management library

As part of this PR, we experimented with [Jotai](https://jotai.org/). However, we decided to postpone the adoption of a state management library because we might need to do some further research and consider other options. If you are curious, you can checkout to commit cfe8d820864af51031c9c4f2bfe990c5a65e9813.

We need to decide how much logic we want to move out of our React code. For having kind of a "shared" `useState`, Jotai is great. If you want to put some logic to load and synchronize the state, then perhaps we should go with another option (e.g., [zustand](https://docs.pmnd.rs/zustand/getting-started/introduction), [recoil](https://recoiljs.org/), or even [react-query](https://tanstack.com/query/v4/docs/framework/react/overview)).


dgdavid (approved):
Thanks Imo

mvidner:
What about the failing tests?
The test log is cluttered with `console.warn` and `console.error` and then a FAIL:
> TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Partitions and file systems"

imobachgs:
> What about the failing tests? The test log is cluttered with `console.warn` and `console.error` and then a FAIL:
> 
> > TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Partitions and file systems"

Tests are out of scope. I fixed/added some tests related to the code I changed, but nothing else.

The plan is to disable the tests that are failing to make sure we do not break more tests.

PR #1302, opened 2024-06-11
Adapt network to the new UI

ancorgs:
This includes the pure UI changes:

 - Turning sections into cards
 - Moving the "Connect to Wi-Fi" button from the special menu to the header

network

But in my experience, there are quite some functional (beyond UI) aspects that are currently broken:

- I would not expect two entries on "WiFi connections" since I only have one WiFi network card. Looks like all connections related to the card (both active and non-active) are displayed.
- `wifiScanSupported` is always false, so I had to tweak some conditions to make the "connect" button visible at the screenshot above.
- If I click on any of the available networks at the WiFi scanning pop-up, I get the following error.

cannot


web/src/components/network/NetworkPage.jsx
```diff
@@ -177,15 +177,40 @@ export default function NetworkPage() {
...
-        {ready ? <WifiConnections /> : <Skeleton />}
-      </Section>
-
-      <NetworkPageMenu wifiScanSupported={wifiScanSupported} openWifiSelector={openWifiSelector} />
+      <Page.Header>
+        <Flex>
+          <FlexItem>
+            <h2>{_("Network")}</h2>
+          </FlexItem>
+          <If
```
dgdavid, line 185: Note, I'm killing the `<If>` component too. There were no few complains about it in the past :)

dgdavid: Also, another note: for the shake of reducing nodes, I'm using Flex with props when possible instead of Flex with FlexItems. But this is just a comment, please keep the change as it is.

web/src/components/network/NetworkPage.jsx
```diff
@@ -177,15 +177,40 @@ export default function NetworkPage() {
...
+            condition={wifiScanSupported}
+            then={
+              <FlexItem align={{ default: "alignRight" }}>
+                <Button onClick={openWifiSelector}>{_("Connect to a Wi-Fi network")}</Button>
```
dgdavid, line 189: What about using `variant="secondary"` here?

dgdavid (approved):
Looks good, just a suggestions about the button look & feel. But feel free to merge it as it is since it is good enough for the task we're addressing this week.

web/src/components/network/NetworkPage.jsx
ancorgs: Updated

PR #1308, opened 2024-06-11
feat(web): product selection progress

joseivanlopez:


web/src/Root.jsx
```diff
@@ -70,6 +71,9 @@ const Header = () => {
 };
 
 const Sidebar = () => {
+  const navigate = useNavigate();
+  const { products } = useProduct();
+
```
dgdavid, line 76: Now or in the future, I think la the logic for displaying the `ChangeProductButton` should live in it's own component.

web/src/MainLayout.jsx
```diff
@@ -69,6 +70,21 @@ const Header = () => {
   );
 };
 
+const ChangeProductButton = () => {
```
dgdavid, line 73: Thanks!

I actually meant to put only the button and it's logic in its own file somewhere, without the PageSidebarBody. The idea is that such a button could be mounted in other places if needed.

But don't worry by now and keep it as it is now. There still a lot work to do regarding the UI. I can relocate it later.

web/src/components/product/ProductSelectionProgress.jsx
```diff
@@ -0,0 +1,162 @@
...
+      setStorageProgress(p => ({ ...p, start: progress.start }));
+    };
+
+    cancellablePromise(software.getProgress()).then(updateProgress);
+
+    return software.onProgressChange(updateProgress);
+  }, [cancellablePromise, setSoftwareProgress, software]);
+
+  return (
+    <Center style={{ blockSize: "100vh" }}>
```
dgdavid, line 139: Just a reminder for myself: this and other `style` attributes should be dropped.

dgdavid (approved):
LGTM

web/src/components/product/ProductSelectionProgress.jsx
```diff
@@ -0,0 +1,162 @@
...
+      <Grid hasGutter>
+        <GridItem sm={8} smOffset={2}>
+          <Card>
```
dgdavid, line 142: I'd use isPlain here

```suggestion
          <Card isPlain>
```

dgdavid (approved):
Everything is green, go ahead :wink:

imobachgs (approved):

PR #1309, opened 2024-06-11
Small network fixes

teclator:
## Problem

- Set settings when load the network page as otherwise wireless will be set to not supported.
- Fixed set of device routes from the  API when no netmask is defined.


dgdavid (approved):
LGTM, thanks

PR #1311, opened 2024-06-11
feat(Live): Display more info in the console

lslezak:
## Problem

- Similar to printing the root password to the console the users should know more details about the system, esp.security related data like the Agama SSL certificate fingerprint

## Solution

- Add several issue generators which contain the required details

## Displayed Info

- Agama welcome message, it would be nice to welcome the users :handshake: 
- The welcome message contains the Agama version (8+227 means version 8 + additional 227 commits from Git, for the public release it will display only the plain version like 9)
- SSH host key fingerprints, useful when connecting via SSH to verify that you are really connecting to the right machine
- Agama SSL certificate fingerprint, when connecting remotely you should verify that the same fingerprint is reported by the browser
- Prints Agama URLs for connection, including the mDNS name
- (The generated root password was already printed at the console before.)

## Details

- The mDNS name (e.g. "agama.local") is directly obtained from the Avahi server, if you start a second Agama instance it will use the "agama-2.local" name to avoid conflicts. If the Avahi daemon does not run or is stopped then the item is removed from the list.
- Also the SSL certificate fingerprint is automatically refreshed whenever the certificate file is updated or removed.
- Using the udev rule ensures that the dynamically added/removed network devices will be handled correctly
- The version in the welcome message is static (not refreshed after creating it), later when we support some kind of self-update we could possibly regenerate it as well
- The same applies for the SSH host key fingerprints , currently they are static

## Testing

- Tested manually

## Screenshots

agama_console_messages

### DIFF
--- a/live/root/usr/lib/udev/rules.d/80-agama-connect-issue.rules
+++ b/live/root/usr/lib/udev/rules.d/80-agama-connect-issue.rules
@@ -1,0 +1,5 @@
+# udev rules for generating the Agama access URLs displayed at the console
+ACTION=="add", SUBSYSTEM=="net", RUN+="/usr/bin/agama-issue-generator --network add $env{INTERFACE}"
+ACTION=="remove", SUBSYSTEM=="net", RUN+="/usr/bin/agama-issue-generator --network rm $env{INTERFACE}"
+ACTION=="move", SUBSYSTEM=="net", RUN+="/usr/bin/agama-issue-generator --network add $env{INTERFACE}"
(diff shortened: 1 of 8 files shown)


dgdavid (approved):
Code-wise I can't help too much, but the overall approach LGTM.

Thanks for such a cool and complete welcome! :heart:

live/root/usr/lib/udev/rules.d/80-agama-connect-issue.rules
```diff
@@ -0,0 +1,5 @@
+# udev rules for generating the Agama access URLs displayed at the console
```
imobachgs, line 1: Nice trick :-)

imobachgs (approved):
Great work! Thanks a lot.

PR #1312, opened 2024-06-11
web: restore the storage devices tech menu

dgdavid:
Restores the storage devices tech menu, placing it at the end of the device selection page. 

Think on it as a temporary change. We're going to improve both, the look & feel and the location ASAP.

Screenshot from 2024-06-11 17-31-15


ancorgs (approved):
As a temporary change, LGTM.

PR #1314, opened 2024-06-11
feat(cli): drop "agama config add/set" subcommands

imobachgs:
## Problem

Agama CLI supports setting specific values directly via the CLI. Check `agama config --help` to learn more. However, it has a few problems:

* It is not finished yet, especially when it comes to handle collections.
* How useful is for real use cases (instead of just using config load)? We have some examples in the documentation, but nothing critical.
* Users could manipulate the configuration by using a tool like `jq`.

## Solution

Just drop the `agama config add/set` subcommands.


dgdavid (approved):

joseivanlopez (approved):
LGTM

PR #1316, opened 2024-06-12
feat(web): users adjustments

dgdavid:
Among other changes (see commits) this PR restores the text for the "Edit password too" in order to avoid introducing a new translation just because. Of course, we can change texts as much as needed in the future, but let's preserve them when possible now.


teclator (approved):
LGTM

PR #1317, opened 2024-06-12
fix(dbus): fix users issues detection

imobachgs:
Recently, we implemented the Issues API to users management (which was using the old validation API). However, the manager and the D-Bus service where using separate `Users` instances.


teclator (approved):
LGTM

mvidner (approved):
Thanks for fixing the docs too

PR #1202, opened 2024-05-13
feat(web): remake the user interface

dgdavid:
As Agama has progressed, the original idea of a hyper-minimalist interface has completely vanished. We have moved far away from the first _SPA interactive interface_. The installation summary has lost all its value and it is now a sort of navigation menu or index. Furthermore, it is impossible to start with a valid configuration straight away due to several reasons, such as

* The creation of a user it's mandatory since it cannot be relegated to the first boot.
* The storage proposal algorithm [does not perform as many attempts](https://github.com/openSUSE/agama/issues/1159#issuecomment-2082701802) as it used to do in YaST.

Last but not least, it is way weird landing in a page with a big, green, and enabled <kbd>Install</kbd> button that will prompt an error when clicked before any other user interaction.

There is a [proposal to improve](https://github.com/openSUSE/agama/discussions/778) that first initial screen, but having in mind the Agama development progression I believe that the current approach does not scale. Regrettably, looks like a dead end.

With this in mind, I have been thinking a bit about an alternative that I had on my to-do list for the future, which consisted of converting the summary screen into a panel on the left and loading the content of each section to the right (which would overlap the first on small devices). It would solve two problems in one shot by stop forcing the user to navigate back and forth to change between sections and stop wasting available space on large screens. The key was to make better use of react-router and embrace [nested routes](https://reactrouter.com/en/main/start/overview#nested-routes) as designed instead of fighting against them. Something perfectly doable.

However, as soon as I started writing some code to play with, I realized that also embracing more heavily PatternFly would be enough to start making it possible. Moreover, I concluded that it could even help to solve many of the problems we currently have with the interface at many levels.

So I got to work to carry out a small proof of concept with, among others, following ideas in mind,

* Use PatternFly as much as possible to the point of looking familiar with Cockpit UI but keeping some bits of Agama's identity (like typography and colors, the absence of wizard,  to allow the user to move as freely as possible, etc)
* Reduce the number of components developed by us and help to improve the existing ecosystem instead. After all, at this moment it is not realistic to think that we can also develop and maintain a design and components system (although I would like to do so, of course :P). 
* Sensibly embrace the router: use a [data router](https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis), [nested routes](https://reactrouter.com/en/main/start/overview#nested-routes), [outlet context](https://reactrouter.com/en/main/hooks/use-outlet-context), etc.
[...]

PR #1328, opened 2024-06-13
chore(web): improve product selection

joseivanlopez:
* Do not preselect a product.
* Show product progress inside the simple layout which allows changing the installer settings (e.g., language).
* Fix routes.

### DIFF
--- a/web/src/components/overview/OverviewPage.jsx
+++ b/web/src/components/overview/OverviewPage.jsx
@@ -43,7 +43,7 @@ import SoftwareSection from "./SoftwareSection";
 import { _ } from "~/i18n";
 
 const SCOPE_HEADERS = {
-  user: _("Users"),
+  users: _("Users"),
   storage: _("Storage"),
   software: _("Software")
 };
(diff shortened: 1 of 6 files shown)


web/src/components/overview/OverviewPage.jsx
```diff
@@ -43,7 +43,7 @@ import SoftwareSection from "./SoftwareSection";
 import { _ } from "~/i18n";
 
 const SCOPE_HEADERS = {
-  user: _("Users"),
+  users: _("Users"),
```
dgdavid, line 46: Good catch!

web/src/components/product/ProductSelectionProgress.jsx
```diff
@@ -136,26 +137,28 @@ function ProductSelectionProgress() {
...
-                  storageProgress={storageProgress}
-                  softwareProgress={softwareProgress}
-                />
-              </Stack>
-            </CardBody>
-          </Card>
-        </GridItem>
-      </Grid>
-    </Center>
+    <SimpleLayout showOutlet={false} showInstallerOptions={false}>
```
dgdavid, line 140: Why do you want to hide the installer options here? What if I need to download the logs because something?

dgdavid (approved):
LGTM, however I've left a comment about installer options. Please have a look but feel free to merge it as it is. We can polish these details against `master` branch once #1202 gets merged.

dgdavid (approved):

PR #1329, opened 2024-06-13
fix(zypp): Reuse the repositories, use libzypp in "chroot"

lslezak:
## Problem

- When changing the product from openSUSE Tumleweed to openSUSE MicroOS the Tumbleweed repository is refreshed again
- The code initializes the libzypp in the Live ISO system, there are some workarounds to avoid reusing the Live ISO repositories

Related cards:
- https://trello.com/c/liPX29Cu/3695-5-research-improve-software-management-performance
 

## Solution

- When the changed product uses the same repositories then just reuse them, do not create the repositories again
- Run libzypp in `/run/agama/zypp` repository to not mess with the package management from the Live ISO, removed the related workarounds

## Notes

- The reusing is simple all-or-nothing, i.e. the products must use the very same repositories. In the future we could improve it to allow reusing just a subset of the repositories.
- Maybe it will need some adoption when the system is registered, but as there is no real product to register I'm leaving this for the future.
- I put the zypp lock to `/run/agama/zypp` as well, that means now you can run `zypper` in the Live system while Agama is running, yay! :star_struck: 
- That workaround with switching `Yast::Stage` is not needed anymore, it was needed when we run the proposal using the YaST code, it is not needed with the new refactored code.

## Testing

- Tested manually (mvidner: in container, switched products 3 times, without hitting Install)

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -39,7 +39,6 @@ Yast.import "Package"
 Yast.import "Packages"
 Yast.import "PackageCallbacks"
 Yast.import "Pkg"
-Yast.import "Stage"
 
 module Agama
   module Software
@@ -72,6 +71,11 @@ module Agama
       PROPOSAL_ID = "agama-user-software-selection"
       private_constant :PROPOSAL_ID
 
+      # create the libzypp lock and the zypp caches in a special directory to
+      # not be affected by the Live system package management
+      TARGET_DIR = "/run/agama/zypp"
+      private_constant :TARGET_DIR
+
       attr_accessor :languages
 
       # Available products for installation.
@@ -97,6 +101,7 @@ module Agama
         @user_patterns = []
         @selected_patterns_change_callbacks = []
         on_progress_change { logger.info(progress.to_s) }
+        initialize_target
       end
 
       # Selects a product with the given id.
@@ -112,8 +117,10 @@ module Agama
 
         raise ArgumentError unless new_product
 
+        update_repositories(new_product)
+
         @product = new_product
-        repositories.delete_all
+
         update_issues
         true
       end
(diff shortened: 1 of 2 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -140,12 +142,16 @@ def probe
...
 
-      def initialize_target_repos
-        Yast::Pkg.TargetInitialize("/")
+      def initialize_target
+        # create the zypp lock also in the target directory
+        ENV["ZYPP_LOCKFILE_ROOT"] = TARGET_DIR
+        # cleanup the previous content (after service restart or crash)
+        FileUtils.rm_rf(TARGET_DIR)
+        FileUtils.mkdir_p(TARGET_DIR)
+        Yast::Pkg.TargetInitialize(TARGET_DIR)
```
mvidner, line 154: I was wondering whether this could break the actual RPM installation, but apparently it's OK as `TargetInitialize` will be also called with the destination directory later:
```console
60eaee65c80e:/ # grep TargetInit /var/log/YaST2/y2log
2024-06-13 12:31:25 <1> 60eaee65c80e(2007) [Pkg] software/manager.rb:154 Pkg Builtin called: TargetInitialize
2024-06-13 12:33:59 <1> 60eaee65c80e(2007) [Pkg] software/proposal.rb:157 Pkg Builtin called: TargetInitialize
2024-06-13 12:35:06 <1> 60eaee65c80e(2007) [Pkg] software/proposal.rb:157 Pkg Builtin called: TargetInitialize
2024-06-13 12:36:07 <1> 60eaee65c80e(2007) [Pkg] software/proposal.rb:157 Pkg Builtin called: TargetInitialize
```

lslezak: Yes, later the target is later moved to `/mnt`. We can do this *only after* the installation is started, after the partitioning step.

mvidner (approved):

PR #1330, opened 2024-06-13
Add more dns options to network settings

jcronenberg:
## Problem

Network model is missing DNS search domains and ignore auto DNS toggle.

## Solution

Add to network settings

## Testing

- *Added to existing test*
- *Tested manually*
- *Tested with the wicked migration*


rust/agama-server/src/network/nm/dbus.rs
```diff
@@ -704,6 +736,10 @@ fn ip_config_from_dbus(conn: &OwnedNestedHash) -> Option<IpConfig> {
         }
     }
 
+    // ensure only unique entries in dns_searchlist
+    ip_config.dns_searchlist.sort();
```
mvidner, line 740: But it's a list, not a set, so this would defeat the user supplied ordering

teclator (approved):
LGTM

PR #1331, opened 2024-06-13
fix(web): do not render the overview page when selecting a product

dgdavid:
## Problem

When selecting a product, there is a small chance for the UI displaying the MainLayout/OverviewPage while waiting for phase/status signals changes. It happens because several reasons, among them the fact of a few miss placed navigations/redirections.

## Solution

To add a new route for the product selection progress and relocate navigation logic, mainly centralising it in the src/App.jsx _intermediate router_.

## Testing

- Adapted existing tests
- Tested manually


dgdavid:
In order to mitigate the problem until we found a better solution, I've created #1333 which can be seen as a subset of this PR

joseivanlopez (approved):
LGTM

PR #1333, opened 2024-06-13
fix(web): products selector redirection

dgdavid:
## Problem

We've found an extra, not needed, even blocking redirection to the product selector once the product is configured.

We think this happens because the condition, which wrongly leaves in the OverviewPage component for historical reasons (it was the main component in the previous UI / routes schema), is triggered before the component has the chance to run its useEffects.

## Solution

Place the condition in App.jsx, but inside the internal `Content` component AFTER the products list has been loaded as a way to ensure that useEffects has been already triggered. A hotfix, indeed.

## Testing

_To be tested with a live ISO_

## Follow up

Refactor the App.jsx component to drop its "router" behaviour and rely more, when possible, in React Router and route loaders or whatever other mechanism that feels better than current one.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Jun 13 16:05:53 UTC 2024 - David Diaz <dgonzalez@suse.com>
+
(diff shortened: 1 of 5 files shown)


dgdavid:
#1334 fixes the problem

PR #1338, opened 2024-06-14
doc(rust): remove references to "config add/set" subcommands

imobachgs:
The old "config add/set" subcommands are still mentioned in the CLI help.

## Open question (out of scope)

* Should we merge `config` and `profile` commands?


joseivanlopez (approved):
LGTM

PR #1341, opened 2024-06-14
Experimental Live ISO self-update

lslezak:
## Problem

- Agama cannot be easily updated on the Live medium
- For testing it is annoying to download the latest image again and again

## Solution

- Let's try a simple approach using `zypper dup`

## Notes

- By default it is disabled, can be enabled by the `agama.self_update` boot option or executed manually via the `agama-self-update` script in a running system.
- Compared to the YaST self-update it uses RPM DB and evaluates the dependencies, this should avoid installing inconsistent updates as we have seen in YaST recently.
- On the other hand YaST compares the individual files from the updated RPMs and overwrites only the changed files. That means it is more space efficient. But as for the initial PoC it is OK I think.

## Testing

- Tested manually

## Screenshots

Updated Agama web UI package during boot:

[agama-self-update-screen0.webm](https://github.com/openSUSE/agama/assets/907998/504a0305-ae13-486a-a416-541633c7cb50)

### DIFF
--- a/doc/live_iso.md
+++ b/doc/live_iso.md
@@ -8,6 +8,7 @@
 - [Agama Live ISO (*only for development and testing*)](#agama-live-iso-only-for-development-and-testing)
   - [Description](#description)
   - [Hardware Requirements](#hardware-requirements)
+- [Experimental Self-update](#experimental-self-update)
 - [The Access Password](#the-access-password)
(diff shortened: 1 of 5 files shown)


ancorgs (approved):
LGTM. Thanks!

PR #1349, opened 2024-06-17
fix(service): Fixed libzypp repository cache removal

lslezak:
## Problem

- The libzypp repository cache was not removed after changing the product

## Solution

- The Pkg bindings just mark the repository removal in memory, the repositories and their caches are really deleted after running the `Pkg.SourceSaveAll` call (kind of "commit" call).

## Testing

- Tested manually


ancorgs (approved):
LGTM

PR #1352, opened 2024-06-17
fix(zypp): Fixed moving libzypp target to `/mnt`

lslezak:
## Problem

- Moving the libzypp target to `/mnt` (the target installed system) happened too early, already in the "propose" step, it needs to happen after partitioning the disk.
- The side effect of this is that mounting the `/mnt` partition can shadow the libzypp cache so it could crash later. Also the next target used after the product change would be the `/mnt`  instead of the `/run/agama/zypp`.

## Solution

- Move the target change to the "install" step

## Notes

- I found out that the `Pkg.SourceCacheCopyTo` call does not work properly with target different than `/`. It always copies the data from `/` ignoring the current target. 
- It could copy some wrong data from the Live ISO to the target, as a quick solution I have commented that out. I'll fix it later in a separate PR.

## Testing

- Tested manually
- Updated unit tests

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -172,6 +172,11 @@ module Agama
 
       # Installs the packages to the target system
       def install
+        # move the target from the Live ISO to the installed system (/mnt)
+        Yast::Pkg.TargetFinish
+        Yast::Pkg.TargetInitialize(Yast::Installation.destdir)
+        Yast::Pkg.TargetLoad
+
         steps = proposal.packages_count
         start_progress(steps)
         Callbacks::Progress.setup(steps, progress)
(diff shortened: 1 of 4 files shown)


service/lib/agama/software/proposal.rb
```diff
@@ -153,12 +153,8 @@ def languages=(value)
 
       # Initializes the target, closing the previous one
       def initialize_target
-        Yast::Pkg.TargetFinish # ensure that previous target is closed
-        Yast::Pkg.TargetInitialize(Yast::Installation.destdir)
-        Yast::Pkg.TargetLoad
-
         preferred, *additional = languages
-        Yast::Pkg.SetPackageLocale(preferred) if preferred
+        Yast::Pkg.SetPackageLocale(preferred || "")
```
lslezak, line 157: Always set the locale, "" means no locale for libzypp. After removing the wrong target re-initialization above we need to reset the previously set locale if the current one is not defined.

service/lib/agama/software/manager.rb
```diff
@@ -196,7 +201,10 @@ def finish
         progress.step(_("Writing repositories to the target system")) do
           Yast::Pkg.SourceSaveAll
           Yast::Pkg.TargetFinish
-          Yast::Pkg.SourceCacheCopyTo(Yast::Installation.destdir)
+          # FIXME: Pkg.SourceCacheCopyTo works correctly only from the inst-sys
```
lslezak, line 204: I'll fix this in a separate PR, this is just a quick hot fix.

imobachgs (approved):
Please, either add an entry in the changes file now *or* when you finish the packaging fixes.

lslezak:
I'll add the changes in the next PR.

PR #1353, opened 2024-06-17
agama-server: add tun/tap connection config

cfconrad:
## Problem

No support for Tun/Tap connections

## Solution

Add TunTap ConnectionConfig to agama-server.

## Testing

- *Tested manually*


rust/agama-server/src/network/model.rs
```diff
@@ -663,6 +663,7 @@ pub enum ConnectionConfig {
     Vlan(VlanConfig),
     Bridge(BridgeConfig),
     Infiniband(InfinibandConfig),
+    Tun(TunConfig),
```
cfconrad, line 666: I used `Tun` as name, cause it is the name in the NetworkManager dbus model as well (see: https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html  search `tun setting`).

teclator: Do not have an strong opinion, both looks good to me.

rust/agama-server/src/network/model.rs
```diff
@@ -1313,6 +1314,25 @@ impl fmt::Display for InfinibandTransportMode {
...
+    Tun = 1,
+    Tap = 2,
+}
+
+#[derive(Default, Debug, PartialEq, Clone, Serialize)]
+pub struct TunConfig {
+    pub mode: TunMode,
+    pub group: Option<String>,
+    pub owner: Option<String>,
+    pub multi_queue: bool,
```
teclator, line 1330: I'm not very used to tunneling terminology and protocol, and specially this variable is not self descriptive to me, in fact in the NM Device Settings it is NoPi (no protocol information), apart of that, which is the wicked equivalent name?, it is also configurable in wicked?

teclator:
Once completely supported we do not have to forget to add the DeviceType as we already happened with Bridge type 

https://github.com/openSUSE/agama/blob/a6d85d73bfa02e7d1759173d5121773c906d19be/rust/agama-server/src/network/nm/model.rs#L87

rust/agama-server/src/network/model.rs
cfconrad: Wicked only support mode, group and owner. Everything else was added because of NetworkManager.

I would say multi_queue is this: `From version 3.8, Linux supports multiqueue tuntap which can uses multiple
file descriptors` (see [kernel.org/tuntap.txt](https://www.kernel.org/doc/Documentation/networking/tuntap.txt) )
And the same as you can configure with `ip`
```
# ip tuntap help
Usage: ip tuntap { add | del | show | list | lst | help } [ dev PHYS_DEV ]
	[ mode { tun | tap } ] [ user USER ] [ group GROUP ]
	[ one_queue ] [ pi ] [ vnet_hdr ] [ multi_queue ] [ name NAME ]
```

teclator: The thing is ... if we do not need it by now.. shall we add it to our model?

cfconrad: Ok I agree, we will omit them for now.

imobachgs (approved):

PR #1356, opened 2024-06-19
feat: unify progress handling

imobachgs:
Trello: https://trello.com/c/Miuvai6N/3706-5-better-progress-report-during-installation

The main goal of this PR is to implement a better mechanism for displaying progress in the web UI. It takes the product selection progress as inspiration but eliminates hardcoded values.

## Progress API problem

When we designed the initial progress API, Agama was quite different. While working on this PR, we have identified a relevant problem: there is no way to have a subprogress in the same service.

The manager service defines the main steps (e.g., "Probing Software") and the details are just the progress from software and storage services (e.g., "Refreshing repositories metadata").

As a consequence, we cannot have a detailed "Configuring system" step at the end of the installation with its details (e.g., "Writing users"). You can only have either 1) a "Configuring system" with no details or several separate steps, which is too much.

In general, we are telling people that they can use the WS to track the progress, so we should decide on a stable API as soon as possible.

## Screenshots

<details>
<summary>Installation progress</summary>

Captura desde 2024-06-19 15-07-24
</details>

<details>
<summary>Product selection selection progress</summary>

As you may see in the following screenshot, some texts have been changed. The reason is that now it uses the texts from the backend. We should consider adjusting them.
Captura desde 2024-06-19 15-06-57
</details>

### DIFF
--- a/rust/agama-server/src/web/common.rs
+++ b/rust/agama-server/src/web/common.rs
@@ -165,10 +165,21 @@ struct ProgressState<'a> {
     proxy: ProgressProxy<'a>,
 }
 
-async fn progress(State(state): State<ProgressState<'_>>) -> Result<Json<Progress>, Error> {
+/// Information about the current progress sequence.
+#[derive(Clone, Default, Serialize)]
+pub struct ProgressSequence {
+    /// Sequence steps if known in advance
+    steps: Vec<String>,
+    #[serde(flatten)]
+    progress: Progress,
+}
+
+async fn progress(State(state): State<ProgressState<'_>>) -> Result<Json<ProgressSequence>, Error> {
     let proxy = state.proxy;
     let progress = Progress::from_proxy(&proxy).await?;
-    Ok(Json(progress))
+    let steps = proxy.steps().await?;
+    let sequence = ProgressSequence { steps, progress };
+    Ok(Json(sequence))
 }
 
 #[pin_project]
--- a/rust/agama-server/src/web/common.rs
+++ b/rust/agama-server/src/web/common.rs
@@ -167,6 +167,7 @@ struct ProgressState<'a> {
 
 /// Information about the current progress sequence.
 #[derive(Clone, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct ProgressSequence {
     /// Sequence steps if known in advance
     steps: Vec<String>,
--- a/service/lib/agama/progress.rb
+++ b/service/lib/agama/progress.rb
@@ -192,7 +192,7 @@ module Agama
   private
 
     def description_for(step)
-      @descriptions[step - 1] || format(_("Step %s/%s"), step, total_steps)
+      @descriptions[step - 1] || ""
     end
   end
 end
--- a/service/lib/agama/progress.rb
+++ b/service/lib/agama/progress.rb
@@ -24,9 +24,15 @@ module Agama
   #
   # It allows to configure callbacks to be called on each step and also when the progress finishes.
   #
+  # In most cases all steps are known in advance (e.g., "probing software", "probing storage", etc.)
+  # but, in some situations, only the number of steps is known (e.g., "Installing package X").
+  #
+  # Use the Progress.with_descriptions to initialize a progress with known step descriptions and
+  # Progress.with_size when only the number of steps is known
+  #
   # @example
   #
-  #   progress = Progress.new(3)                    # 3 steps
+  #   progress = Progress.with_size(3)              # 3 steps
   #   progress.on_change { puts progress.message }  # configures callbacks
   #   progress.on_finish { puts "finished"  }       # configures callbacks
   #
@@ -45,6 +51,14 @@ module Agama
   #
   #   progress.finished?                            #=> true
   #   progress.current_step                         #=> nil
+  #
+  # @example Progress with known step descriptions
+  #
+  #   progress = Progress.with_descriptions(["Partitioning", "Installing", "Finishing"])
+  #   progress.step { partitioning }                      # next step
+  #   progress.current_step.description                   #=> "Partitioning"
+  #   progress.step("Installing packages") { installing } # overwrite the description
+  #   progress.current_step.description                   # "Installing packages"
   class Progress
     # Step of the progress
     class Step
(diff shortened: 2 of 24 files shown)


service/lib/agama/progress.rb
```diff
@@ -154,5 +188,11 @@ def to_s
 
       "#{current_step.description} (#{@counter}/#{total_steps})"
     end
+
+  private
+
+    def description_for(step)
+      @descriptions[step - 1] || format(_("Step %s/%s"), step, total_steps)
```
joseivanlopez, line 195: NP: I would not generate a default description. If no description is known for the step (e.g., `Propress.with_size` and `progress.step {}`), then I would create a `Step` with empty description.

Note that the steps counter is generated by the clients if needed (e.g., "Creating partitions (2/4)") from the info on D-Bus (`CurrentStep` and `TotalSteps`). With the default step description proposed here, the clients have no clue whether the counter is needed or it already comes in the description.

service/lib/agama/software/manager.rb
```diff
@@ -132,11 +132,11 @@ def probe
         logger.info "Probing software"
 
         if repositories.empty?
-          start_progress(3)
+          start_progress_with_size(3)
```
joseivanlopez, line 135: In this case the descriptions are known in advance. Should we use the `with_descritptions` constructor?

service/lib/agama/storage/manager.rb
```diff
@@ -107,7 +107,7 @@ def on_probe(&block)
 
       # Probes storage devices and performs an initial proposal
       def probe
-        start_progress(4)
+        start_progress_with_size(4)
```
joseivanlopez, line 110: I know in these case the steps are known in advance too, but this progress is only used for the "details". Anyway, I would use the `with_descriptions` constructor. Maybe they would be needed if we change the UI. Anyway, probably you want to change the overall approach of the progress API. So we could keep it as it is for now.

service/lib/agama/with_progress.rb
```diff
@@ -30,18 +30,34 @@ class NotFinishedProgress < StandardError; end
     # @return [Progress, nil]
     attr_reader :progress
 
+    # Creates a new progress with a give number of steps
```
joseivanlopez, line 33: *given*

service/lib/agama/with_progress.rb
```diff
@@ -30,18 +30,34 @@ class NotFinishedProgress < StandardError; end
...
+    # Creates a new progress with a given set of steps
+    #
+    # @param descriptions [Array<String>] Steps descriptions
+    def start_progress_with_descriptions(*descriptions)
+      start_progress(descriptions: descriptions)
+    end
+
     # Creates a new progress
     #
+    # Prefer using #start_progress_with_Size or #start_progress_with_descriptions.
```
joseivanlopez, line 49: Could the method be private?

service/test/agama/dbus/interfaces/progress_test.rb
```diff
@@ -133,15 +133,16 @@ class Backend
...
       progress.step("step 1")
     end
 
     it "returns de D-Bus properties of the progress interface" do
       expected_properties = {
         "TotalSteps"  => 2,
         "CurrentStep" => [1, "step 1"],
-        "Finished"    => false
+        "Finished"    => false,
+        "Steps"       => []
```
joseivanlopez, line 145: It would be nice to have a test in which `Steps` has some value.

rust/agama-server/src/web/common.rs
```diff
@@ -165,10 +165,21 @@ struct ProgressState<'a> {
     proxy: ProgressProxy<'a>,
 }
 
-async fn progress(State(state): State<ProgressState<'_>>) -> Result<Json<Progress>, Error> {
+/// Information about the current progress sequence.
+#[derive(Clone, Default, Serialize)]
+pub struct ProgressSequence {
```
joseivanlopez, line 171: Why not to add a `steps` field to the `Progress` struct and fill it in `#from_proxy`?

web/src/components/core/InstallationProgress.jsx
```diff
@@ -20,26 +20,24 @@
...
-          <GridItem sm={8} smOffset={2}>
-            <Card>
-              <CardBody>
-                <ProgressReport />
-              </CardBody>
-            </Card>
-          </GridItem>
-        </Grid>
-      </Center>
+    <SimpleLayout showOutlet={false} showInstallerOptions={false}>
```
joseivanlopez, line 39: NP: I would show the installer options. It could be useful to download logs while or after installing.

web/src/components/product/ProductSelectionProgress.jsx
```diff
@@ -21,152 +21,38 @@
...
-
-    cancellablePromise(software.getProgress()).then(updateProgress);
-
-    return software.onProgressChange(updateProgress);
-  }, [cancellablePromise, setSoftwareProgress, software]);
+  if (!selectedProduct) {
+    return;
+  }
 
   if (status === IDLE) return <Navigate to="/" replace />;
```
joseivanlopez, line 49: Probably we could move this navigation to `App`:

~~~

    if (status === IDLE && location.pathname === "/products/progress") {
      return <Navigate to="/" />;
    }
~~~

So App would contain all the "routing" logic until we will improve the router usage.

service/lib/agama/storage/manager.rb
imobachgs: Initially, I used the `with_description` constructor, but as it was not really needed, I decided to revert that change.

rust/agama-server/src/web/common.rs
imobachgs: Because `Progress` represents a progress change, it is used in the "Progress" signal, and it does not include the list of steps.

web/src/components/core/InstallationProgress.jsx
imobachgs: copy&paste error :-) Good point.

web/src/components/product/ProductSelectionProgress.jsx
imobachgs: I see your point but 1) it is not part of this PR *and* I am scared of breaking the navigation again. I would keep it as it is until we fix the router usage.

joseivanlopez (approved):
LGTM

PR #1357, opened 2024-06-19
fix(service): Copy the libzypp caches and credentials to the installed system

lslezak:
## Problem

- The libzypp caches and credentials are not copied to the installed system
- Installing any package or even simple `zypper refresh` triggers full repository reload

## Solution

- Just copy the files from the Live ISO zypp repository to the installed system

## Testing

- Added a new unit test
- Tested manually, after installation all caches are copied and `zypper refresh` does nothing

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jun 19 06:04:46 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
+
+- Use a different libzypp target for Agama, do not use the Live
+  system package management (gh#openSUSE/agama#1329)
+- Properly delete the libzypp cache when changing the products
+  (gh#openSUSE/agama#1349)
+
+-------------------------------------------------------------------
 Thu Jun 13 10:53:27 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Replace the Validations with the Issues API in the users-related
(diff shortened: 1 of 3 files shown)


service/test/agama/software/manager_test.rb
```diff
@@ -352,11 +356,52 @@
...
       expect(Yast::Pkg).to receive(:TargetFinish)
 
       subject.finish
     end
+
+    it "copies the libzypp cache and credentials to the target system" do
+      allow(Dir).to receive(:exist?).and_call_original
+
+      # copying the raw cache
+      expect(Dir).to receive(:exist?).with("/run/agama/zypp/var/cache/zypp/raw").and_return(true)
```
imobachgs, line 370: np: Although we might consider it good enough, this test just makes sure that you call the expected methods. What about a test that actually performs the copy and checks the result?

lslezak: To be honest I do not like unit tests which really touch the system. For integration tests it is OK but for unit tests I'd rather avoid that if not necessary. And it actually means you are testing the Ruby functionality, not our code.

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Jun 19 06:04:46 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
+
+- Use a different libzypp target for Agama, do not use the Live
+  system package management (gh#openSUSE/agama#1329)
+- Properly delete the libzypp cache when changing the products
+  (gh#openSUSE/agama#1349)
+
```
mvidner, line 8: OK so shouldn't there be an entry for *this* PR too?

service/test/agama/software/manager_test.rb
imobachgs: I can understand that, but I would say that we are not writing unit tests "by the book" and abusing mocks can produce tests that are not that useful after all. Actually, that test knows a lot about implementation details, which looks wrong to me.

However, as said, I will not block this PR just because of that.

lslezak: I see your point. In such case the test just tests the mocks themselves.

But if we test the real result then we are actually testing the external code and the test is more fragile. If the code basically only calls external code and nothing else then I think it is acceptable.

PR #1358, opened 2024-06-19
fix(Live): Improve interactive root password services

lslezak:
## Improvements

- Properly disable the systemd console output while the interactive dialog is displayed
- See the [systemd  signals documentation](https://www.freedesktop.org/software/systemd/man/latest/systemd.html#Signals)
- I used some workaround before (starting the service at the end) but that is not reliable, if some new service appears or if the service dependencies are changed then it might easily break.
- Clear/restore the screen after closing the interactive dialog so the dialog content is not displayed on the screen after leaving

## Testing

- Tested manually, all services work fine, no systemd output is printed over the interactive dialog

### DIFF
--- a/live/root/usr/bin/live-password
+++ b/live/root/usr/bin/live-password
@@ -16,6 +16,7 @@ TITLE="Set Login Password"
 
 # functions for entering the password in an interactive dialog
 confirm_exit() {
+  # --keep-tite is not a misspelling of "title"
   if dialog --keep-tite --backtitle "$BTITLE" --defaultno --yesno "Are you sure you want to cancel?" 5 40; then
     exit 1
   fi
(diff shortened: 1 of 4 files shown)


live/root/usr/bin/live-password
```diff
@@ -16,22 +16,22 @@ TITLE="Set Login Password"
 
 # functions for entering the password in an interactive dialog
 confirm_exit() {
-  if dialog --backtitle "$BTITLE" --defaultno --yesno "Are you sure you want to cancel?" 5 40; then
+  if dialog --keep-tite --backtitle "$BTITLE" --defaultno --yesno "Are you sure you want to cancel?" 5 40; then
```
mvidner, line 19:
```suggestion
  # --keep-tite is not a misspelling of "title"
  if dialog --keep-tite --backtitle "$BTITLE" --defaultno --yesno "Are you sure you want to cancel?" 5 40; then
```

mvidner (approved):

PR #1366, opened 2024-06-21
fix(web): do not render App on protected routes

imobachgs:
When visiting the root route (`/`), the App component is rendered even if the user is not logged in. Of course, this causes many errors that are not visible in the UI but in the network monitor.

Fixing this problem requires moving the code to handle phase and status changes to a higher place in the hierarchy because, now, `App` is rendered "too late" to observe status changes.

<details>
<summary>Failed connection attempts</summary>

Captura desde 2024-06-21 15-29-39
</details>

### DIFF
--- a/web/src/Protected.jsx
+++ b/web/src/Protected.jsx
@@ -27,6 +27,7 @@ import { AppProviders } from "./context/app";
 export default function Protected() {
   const { isLoggedIn } = useAuth();
 
+  // It is not known yet whether the user is logged or not.
   if (isLoggedIn === undefined) return;
 
   if (isLoggedIn === false) {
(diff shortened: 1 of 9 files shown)


web/src/Protected.jsx
```diff
@@ -27,6 +27,8 @@ import { AppProviders } from "./context/app";
 export default function Protected() {
   const { isLoggedIn } = useAuth();
 
+  if (isLoggedIn === undefined) return;
```
jreidinger, line 30: I think this deserve a bit of comment in code why we do it as it is not obvious especialyl with combination of condition below.

imobachgs: Well, it is a rather typical pattern in Reac/JavaScript, but a comment won't hurt.

jreidinger (approved):

PR #1372, opened 2024-06-25
Install from local DVD if it is present

lslezak:
## Problem

- Local DVD cannot be used for installation, offline installation is not possible
- We have a [special branch](https://github.com/openSUSE/agama/compare/install_from_2nd_DVD?expand=1) with this feature and build special ISO but I think it should work out of box with upstream Agama
- Sooner or later we will need to support installation from the local media anyway

## Solution

- A disk label defines the installation repository name for each product
- If such disk label is found then that disk is used as the installation repository
- If no label is found in the system it uses the online repositories
- If multiple labels are found then the DVD device is preferred

## TODO

- [x] Update unit tests
- [x] Add more comments
- [x] Update documentation

## Testing

- Tested manually, when the Tumbleweed DVD is present it is automatically used instead of the online repositories
- When the product is switched to MicroOS and the DVD is not present it uses the online repositories

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -445,7 +445,9 @@ module Agama
       # returns empty list if there is no device with the required label
       def disks_with_label(label)
         data = list_disks
-        disks = data.fetch("blockdevices", []).map { |device| device["kname"] if device["label"] == label }
+        disks = data.fetch("blockdevices", []).map do |device|
+          device["kname"] if device["label"] == label
+        end
         disks.compact!
         logger.info "Disks with the installation label: #{disks.inspect}"
         disks
@@ -457,7 +459,7 @@ module Agama
         # we need only the kernel device name and the label
         output = `lsblk --paths --json --output kname,label`
         JSON.parse(output)
-      rescue RuntimeError e
+      rescue RuntimeError => e
         logger.error "ERROR: Cannot read disk devices: #{e}"
         {}
       end
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -459,7 +459,7 @@ module Agama
         # we need only the kernel device name and the label
         output = `lsblk --paths --json --output kname,label`
         JSON.parse(output)
-      rescue RuntimeError => e
+      rescue StandardError => e
         logger.error "ERROR: Cannot read disk devices: #{e}"
         {}
       end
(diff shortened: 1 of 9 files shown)


imobachgs:
The approach looks good to me. However, I am not sure if we should stop using the online repositories at all (that's what I understood from the description). What if I have some updates on the online repos that are not present in the DVD?

lslezak:
1. For Tumbleweed the updates do not make much sense, the installation  repository is always rebuilt as a whole.
2. If we want to install updates we should explicitly ask the user to do so. There is a popup question in both Tumbleweed and Leap whether to use the online repositories. The same applies also for SLES after registration.

I consider this as the initial support, later we should think about all possible scenarios and decide which ones should be supported. IMHO installing with updates is a different problem and should be solved separately.

imobachgs:
1. Yes, I know, but I think that even in Tumbleweed there is some kind of updates channel for urgent things.
2. OK, no problem.

For me it is fine to postpone the decision a bit, but please, let's not forget about this (creating a card in Trello should suffice).

service/lib/agama/software/manager.rb
```diff
@@ -420,9 +421,66 @@ def import_gpg_keys
...
+        disks.compact!
+        logger.info "Disks with the installation label: #{disks.inspect}"
+        disks
+      end
+
+      # get list of disks, returns parsed data from the `lsblk` call
+      # @return [Hash] parsed data
+      def list_disks
+        # we need only the kernel device name and the label
+        output = `lsblk --paths --json --output kname,label`
```
imobachgs, line 460: `lsblk` is included in `util-linux-systemd`. Should we add a dependency?

service/lib/agama/software/manager.rb
```diff
@@ -420,9 +421,66 @@ def import_gpg_keys
...
+
+        # multiple installation media?
+        if disks.size > 1
+          # prefer optical media (/dev/srX) to disk so the disk can be used as
+          # the installation target
+          optical = disks.find { |d| d.match(/\A\/dev\/sr[0-9]+\z/) }
+          optical || disks.first
+        else
+          # none or just one disk
+          disks.first
```
imobachgs, line 480: I guess it is a corner case but, what happens if I select the disk for installing the system? Do we need any kind of filtering?

lslezak: The disk will not be available for partitioning/installation. But if you have more than one disk it should work (one for RPM packages, second as the installation target).

But I guess all these scenarios are rather theoretical and look a bit strange to me...

service/lib/agama/software/manager.rb
lslezak: I'll add that...

imobachgs (approved):

PR #1375, opened 2024-06-25
fix: set installer language and keyboard layout

imobachgs:
## Problem

The installer language/keymap selection does not work properly. Although most of the problems have existed for quite some time, the new UI makes them more evident.

## Solution

* Stop using signals to update the backend language. Now, there is an explicit `SetLocale` method, so we know exactly when it is finished.
* Use a proper form (in a popup) to change the language/keymap. You need to click the "Accept" button to apply the changes.
* Properly reload the screen after applying the changes. Disable the "Accept/Cancel" buttons while performing the changes.
* Display a progress screen while re-reading the repositories (as it can take quite some time).
* Move the "Download logs" to the sidebar.

## Testing

- Updated unit tests.
- Tested manually.

## Known problems

- Some unwanted flickering.


dgdavid (approved):
UI wise LGTM having in mind there will be another round of improvements when time permits.

service/lib/agama/dbus/interfaces/locale.rb
```diff
@@ -0,0 +1,52 @@
...
+      module Locale
+        include Yast::I18n
+        include Yast::Logger
+
+        LOCALE_INTERFACE = "org.opensuse.Agama1.Locale"
+
+        def self.included(base)
+          base.class_eval do
+            dbus_interface LOCALE_INTERFACE do
+              dbus_method :SetLocale, "in locale:s" do |locale|
```
mvidner, line 43: Names of locales differ, notably between Linux and browsers: `en_US.UTF-8` vs `en-us` so we should specify what the API expects or we will get subtle bugs

imobachgs: In this case, it expects an `en_US.UTF-8` kind of locale. That's the only locale that the backend understands. The transformation is performed in the [installerL10n.jsx context](https://github.com/openSUSE/agama/blob/master/web/src/context/installerL10n.jsx#L109-L140).

I have added a comment to clarify that. Thanks!

web/src/components/core/InstallationFinished.test.jsx
```diff
@@ -29,6 +29,7 @@ import { EncryptionMethods } from "~/client/storage";
 import InstallationFinished from "./InstallationFinished";
 
 jest.mock("~/client");
+jest.mock("~/components/core/InstallerOptions", () => () => <div>Instlaler Options</div>);
```
joseivanlopez, line 32: Typo: Instlaler

joseivanlopez (approved):
LGTM

PR #1383, opened 2024-06-26
Filter only external connections

teclator:
## Problem

During the installation the default NetworkManager profile tries to apply DHCP over the wired and connected interfaces but the profile is a volatile or only in memory one being skipped by agama.

In order to show the connection we should skip only the external configured ones.

## Solution

Only skip external configured connections by now.

## Screenshots

| Before the fix| After the fix |
| --- | --- |
| image | image |

### DIFF
--- a/rust/agama-server/src/network/nm/client.rs
+++ b/rust/agama-server/src/network/nm/client.rs
@@ -173,7 +173,7 @@ impl<'a> NetworkManagerClient<'a> {
                 .build()
                 .await?;
             let flags = proxy.flags().await?;
-            if flags > 0 {
+            if flags >= 8 {
                 log::warn!("Skipped connection because of flags: {}", flags);
                 continue;
             }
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

rust/agama-server/src/network/nm/client.rs
```diff
@@ -173,7 +173,7 @@ impl<'a> NetworkManagerClient<'a> {
                 .build()
                 .await?;
             let flags = proxy.flags().await?;
-            if flags > 0 {
+            if flags >= 8 {
```
mvidner, line 176:
```suggestion
            # https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMSettingsConnectionFlags
            if flags >= 8 {
```

mvidner (changes requested):
Please fix the test, as suggested.

NP: Also, the PR description does not mention the user visible problem, only some mechanism known to insiders. Copy the user problem from IRC?

mvidner (approved):

rust/agama-server/src/network/nm/client.rs
imobachgs: Please, beware that the comment should start with `//` instead of `#`.

imobachgs (approved):

PR #1385, opened 2024-06-26
fix(web): mark routes name for being translated

dgdavid:
Uses `N_` and `_` combination to get the sidebar routes translated.


imobachgs (approved):
Good catch! LGTM

PR #1394, opened 2024-06-26
fix(web): display the installation progress when connecting late

imobachgs:
## Problem

If you connect when the installation has started, the software service cannot tell ou what you are installing, hence the progress screen is not shown.

## Solution

Until we have a responsive software service, let's omit the product's name.

### DIFF
--- a/web/src/components/core/InstallationProgress.jsx
+++ b/web/src/components/core/InstallationProgress.jsx
@@ -25,12 +25,9 @@ import ProgressReport from "./ProgressReport";
 import SimpleLayout from "~/SimpleLayout";
 
 function InstallationProgress() {
-  console.log("INSTALATION ISSUE");
-
-  const title = _("Installing the system, please wait ...");
   return (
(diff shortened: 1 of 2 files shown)


web/src/components/core/InstallationProgress.jsx
```diff
@@ -20,21 +20,14 @@
...
 import React from "react";
-import { useProduct } from "~/context/product";
-import { sprintf } from "sprintf-js";
 import { _ } from "~/i18n";
 import ProgressReport from "./ProgressReport";
 import SimpleLayout from "~/SimpleLayout";
 
 function InstallationProgress() {
-  const { selectedProduct } = useProduct();
+  console.log("INSTALATION ISSUE");
```
dgdavid, line 28: is this needed?

dgdavid (approved):

web/src/components/core/InstallationProgress.jsx
imobachgs: ouch, fixed

PR #1406, opened 2024-06-27
enable checksum generation for s390x agama live images

wfeldt:
## Problem

For s390x live images no image embedded checksums are generated. This makes media checking impossible.

This also means that the build service will not sign the images.

For some reason this has been enabled for all architectures but s390x. It should be enabled for all.

## Solution

Add `mediacheck` attribute to kiwi config for s390x.

## Testing

Manually verified that built isos are ok after this change.


teclator (approved):
LGTM

Issue #1378, opened 2024-06-25
Use gzip (or another supported format) to generate the logs archive


imobachgs:
GitHub does not allow attaching `bzip2` files. In order to make error reporting easier, `agama logs store` should use a different format (e.g., `.gz` and `.tgz` are supported). Otherwise, you need to uncompress and compress the logs again before submitting them.

imobachgs:
Closed by #1408.

PR #1408, opened 2024-06-28
Use gzip (.gz) instead of bzip2 (.bz2) to compress logs

mvidner:
## Problem

Quoting #1378:

GitHub does not allow attaching `bzip2` files. In order to make error reporting easier, `agama logs store` should use a different format (e.g., `.gz` and `.tgz` are supported). Otherwise, you need to uncompress and compress the logs again before submitting them.

## Solution

Use `gzip` and `.gz` instead.

## Testing

- [x] *Tested manually*

## Screenshots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 /*.pot
 *.mo
 *.bz2
+# Do NOT ignore .github: for git this is a no-op
+# but it helps ripgrep (rg) which would otherwise ignore dotfiles and dotdirs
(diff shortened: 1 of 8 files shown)


imobachgs (commented):
Thanks a lot! Please, add a changelog entry referencing the original bug. Otherwise, it looks good to me!

PS: thanks for the ripgrep fix too :-)

imobachgs (approved):

Issue #1277, opened 2024-05-29
Size entry gets reset when switching from Range to Fixed


kobliha:
![Size gets reset](https://github.com/openSUSE/agama/assets/463995/4bb17923-5aa9-4140-9b27-48aa76583124)

I'm adding a separate /home filesystem.

1. Add filesystem - /home
2. First what I see is size, so I change the default _10 GiB_ to let's say _100 GiB_
3. Then I realize I really want exactly that size, but there is _Range_ selected by default
4. So I switch to _Fixed_
5. The size is now reset back to _10 GiB_

I know this might be a bit questionable, but I would expect the size not to reset back and I would expect the _Mininum_ to be the _Exact Size_ then.

PR #1412, opened 2024-06-28
VolumeDialog: share numeric value between Fixed and Range

mvidner:
## Problem

#1277 - "Size entry gets reset when switching from Range to Fixed"

## Solution

In Add/Edit file system dialog (VolumeDialog), keep the numeric value when switching between Fixed and Range sizing.

Do it by removing `size` and `sizeUnit` from `formData` and using `minSize`, `minSizeUnit` for both roles.

## Testing

- [x] *Tested manually*

## Screenshots

See #1277.


dgdavid (approved):
LGTM, thanks!

PR #1419, opened 2024-07-01
fix(storage): avoid segmentation fault when generating actions

joseivanlopez:
## Problem

The generation of the storage actions was fixed to avoid a segmentation fault while accessing to the `CompoundAction` objects, see https://github.com/openSUSE/agama/pull/1410. But there is still a chance to reference to an actiongraph removed by the ruby GC.

~~~ruby
target_graph.actiongraph.compound_actions.map do |action|
  GC.start
  Action.new(a)
end
~~~ 

The actiongraph object could be dropped by the garbage collector while iterating the list of compound actions.

## Solution

Keep a reference to the actiongraph, ensuring the actiongraph object is not removed while generating the list of actions.

Note: once the list of actions is generated, the actiongraph is not needed anymore because the `Action` objects do not keep a reference to the source compound action, see https://github.com/openSUSE/agama/pull/1410.

### DIFF
--- a/service/lib/agama/storage/actions_generator.rb
+++ b/service/lib/agama/storage/actions_generator.rb
@@ -29,7 +29,20 @@ module Agama
       # param target_graph [Y2Storage::Devicegraph]
       def initialize(system_graph, target_graph)
         @system_graph = system_graph
-        @target_graph = target_graph
+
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

ancorgs (approved):
LGTM

PR #1428, opened 2024-07-02
feat(storage): more accurate info about resizing

joseivanlopez:
## Problem

The information about resizing devices shown by the UI for finding free space is not 100% accurate. Sometimes it allows selecting to resize a device but the device cannot be resized. 

## Solution

* D-Bus provides more accurate information about resizing.
* The UI does not allow selecting "resize" if the device cannot be resized.
* The UI offers a button to show information about the device (either reasons why it cannot be shrunk or total shrinkable size). 

<details>

localhost_8080_ (15)

localhost_8080_ (16)

localhost_8080_ (17)

</details>


rust/agama-lib/src/storage/model.rs
```diff
@@ -504,14 +504,45 @@ pub struct DeviceInfo {
...
+        let mut info: Option<Self> = None;
+
+        if let Some(size) = get_optional_property(&hash, "Supported")? {
+            info = Some(Self::Supported(size));
+        }
+        if let Some(reasons) = get_optional_property(&hash, "Unsupported")? {
+            info = Some(Self::Unsupported(reasons));
+        }
+
+        if let Some(info_value) = info {
```
imobachgs, line 536: What about using [ok_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) or [ok_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or)?

```rust
info.ok_or_else(|| Self::Error::Message(format!("Wrong value for Shrinking: {}", value)))

```

(`ok_or` is shorter, but it builds a string that may be unnecessary).

Additionally, the call to `to_string` is not needed (the `format!` macro already returns a string).

service/lib/agama/storage/device_shrinking.rb
```diff
@@ -0,0 +1,187 @@
...
+
+    private
+
+      # @return [Y2Storage::BlkDevice]
+      attr_reader :device
+
+      # Reasons preventing shrinking.
+      SHRINKING_REASONS = [
+        :RB_SHRINK_NOT_SUPPORTED_BY_FILESYSTEM,
+        :RB_SHRINK_NOT_SUPPORTED_BY_MULTIDEVICE_FILESYSTE,
```
imobachgs, line 82: I think you are missing an `M`.

web/src/components/storage/SpaceActionsTable.test.jsx
```diff
@@ -0,0 +1,206 @@
...
+ * @typedef {import ("~/client/storage").StorageDevice} StorageDevice
+ * @typedef {import ("~/client/storage").Volume} Volume
+ * @typedef {import ("~/components/storage/SpaceActionsTable").SpaceActionsTableProps} SpaceActionsTableProps
+ */
+
+/**
+ * GiB to Bytes.
+ *
+ * @type {(value: number) => number } */
+const gib = (value) => value * (1024 ** 3);
```
dgdavid, line 39: NP: this should be in a utils file, no matter if it is (now) only used in tests.

service/test/agama/storage/device_shrinking_test.rb
```diff
@@ -0,0 +1,171 @@
...
+
+        context "and there is some reasons preventing to resize" do
+          let(:reasons) { [:RB_RESIZE_NOT_SUPPORTED_BY_DEVICE] }
+
+          it "returns false" do
+            expect(subject.supported?).to eq(false)
+          end
+        end
+
+        context "and there is not reasons" do
```
imobachgs, line 97: *there are no reasons" or "there is no reason"?

imobachgs (commented):
It looks good, just some minor comments (and a small bug).

ancorgs:
The rationale (and some historical explanations) for the logic implemented at `Storage::Agama::DeviceShrinking` comes directly from a couple of big comments I wrote at https://trello.com/c/YWTxEvpu

I would like that content (not necessarily literally) to be somewhere on this pull request. Maybe close to the `DeviceShinking` code, maybe as a separate file at `doc` or maybe as part of the PR description. But I think the comments contain a useful historical and logical view that I would like anybody could easily find in the future when reviewing our decisions.

ancorgs:
Shouldn't we also honor `DeviceShrinking` at the proposal settings conversion [when `:resize` is chosen](https://github.com/openSUSE/agama/blob/master/service/lib/agama/storage/proposal_settings_conversions/to_y2storage.rb#L129) as policy?

I fear that piece of code can generate potentially dangerous `:shrink` operations that wouldn't be accepted if configured manually. Or are we maybe safe because the proposal will ignore those? If that's the case, I would like to at least see a comment about it.

joseivanlopez:
> Shouldn't we also honor `DeviceShrinking` at the proposal settings conversion [when `:resize` is chosen](https://github.com/openSUSE/agama/blob/master/service/lib/agama/storage/proposal_settings_conversions/to_y2storage.rb#L129) as policy?
> 
> I fear that piece of code can generate potentially dangerous `:resize` actions that wouldn't be accepted if configured manually. Or are we maybe safe because the proposal will ignore those? If that's the case, I would like to at least see a comment about it.

I fear we are not safe in all scenarios. For example, I was able to resize a device with closed LUKS.

rust/agama-lib/src/storage/model.rs
joseivanlopez: Good point! Thanks.

service/lib/agama/storage/device_shrinking.rb
joseivanlopez: Goot catch!

ancorgs (approved):
LGTM

PR #1429, opened 2024-07-02
fix(web): simplify directories list in the spec file

imobachgs:
- Follow a suggestion to simplify the directories list in the spec file.


ancorgs (approved):
LGTM

PR #1430, opened 2024-07-03
fix(auto): remove scriptlets section

imobachgs:
`agama-auto` is a one-shot service and does not need an scriptlets section in the spec. Not to
mention that it was wrongly referring to the `agama-web-server` service.


ancorgs (approved):
LGTM

PR #1437, opened 2024-07-04
JSON schema: improve non-storage parts

mvidner:
## Problem

While reviewing #1263 I read the whole schema and saw room for improvement. Small things

## Solution

- disallow additional properties
- more specific type than string

## Testing

Minimal, just making sure the schema itself is valid, by

```console
$ agama profile validate rust/agama-lib/share/examples/profile_gnome.json
The profile is valid
```

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -13,16 +13,23 @@
       "properties": {
         "patterns": {
           "title": "List of patterns to install",
-          "type": "array"
+          "type": "array",
+          "items": {
+            "type": "string",
+            "examples": ["minimal_base"]
+          }
         }
       }
     },
     "product": {
       "title": "Product to install",
       "type": "object",
+      "additionalProperties": false,
+      "required": ["id"],
       "properties": {
         "id": {
           "title": "Product identifier",
+          "description": "The id field from a products.d/foo.yaml file",
           "type": "string"
         },
         "registrationCode": {
@@ -46,6 +53,9 @@
           "items": {
             "type": "object",
             "additionalProperties": false,
+            "required": [
+              "id"
+            ],
             "properties": {
               "id": {
                 "title": "Connection ID",
@@ -117,7 +127,8 @@
                     "type": "string"
                   },
                   "security": {
-                    "type": "string"
+                    "type": "string",
+                    "examples": ["? maybe this is actually an enum"]
                   },
                   "ssid": {
                     "type": "string"
@@ -139,7 +150,8 @@
                 "additionalProperties": false,
                 "properties": {
                   "mode": {
-                    "type": "string"
+                    "type": "string",
+                    "examples": ["? maybe this is actually an enum"]
                   },
                   "options": {
                     "type": "string"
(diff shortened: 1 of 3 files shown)


rust/agama-lib/share/profile.schema.json
```diff
@@ -188,17 +201,15 @@
...
-            ]
+            }
           }
         }
       }
     },
     "user": {
       "title": "First user settings",
       "type": "object",
+      "additionalProperties": false,
```
mvidner, line 212: feature request: I am surprised that we don't allow hashed passwords in the profile

rust/agama-lib/share/profile.schema.json
```diff
@@ -576,7 +605,8 @@
     "sizeString": {
       "title": "Human readable size (e.g., '2 GiB')",
       "type": "string",
-      "pattern": "^[0-9]+(\\.[0-9]+)?(\\s*([KkMmGgTtPpEeZzYy][iI]?)?[Bb])?$"
+      "pattern": "^[0-9]+(\\.[0-9]+)?(\\s*([KkMmGgTtPpEeZzYy][iI]?)?[Bb])?$",
+      "examples": ["2 GiB", "1.5 TB", "1TIB", "1073741824 b", "1073741824"]
```
mvidner, line 609: BTW, here the `examples` feature really shines because it provides an alternative to the dense regexp

Unfortunately, `agama profile validate` does not check whether the schema is internally consistent and uses only valid examples. Like, I could use a number here and it would not notice.

rust/agama-lib/share/profile.schema.json
```diff
@@ -117,7 +129,8 @@
                     "type": "string"
                   },
                   "security": {
-                    "type": "string"
+                    "type": "string",
+                    "examples": ["? maybe this is actually an enum"]
```
mvidner, line 133: FIXME, need to look up in the code what is valid here

mvidner: @teclator can you advise please?
[service/test/agama/autoyast/wireless_reader_test.rb](https://github.com/openSUSE/agama/blob/master/service/test/agama/autoyast/wireless_reader_test.rb) and 
[web/src/components/network/WifiConnectionForm.jsx](https://github.com/openSUSE/agama/blob/master/web/src/components/network/WifiConnectionForm.jsx) seem to have different sets of values

rust/agama-lib/share/profile.schema.json
imobachgs: Yes, you are right, that's something we need to implement. I have created a placeholder card in Trello.

rust/agama-lib/share/profile.schema.json
```diff
@@ -238,16 +271,18 @@
       "type": "object",
       "properties": {
         "language": {
-          "title": "System language ID (e.g., 'en_US')",
-          "type": "string"
+          "title": "System language ID",
+          "type": "string",
+          "examples": ["en_US"]
```
imobachgs, line 276: We support specifying the whole locale (`en_US.UTF-8`). If no encoding is given (e.g., `en_US`), we fall back to `UTF-8`.

imobachgs (approved):
I did not know about the `examples` element. Thanks!

imobachgs (approved):

PR #1438, opened 2024-07-04
Replace D-Bus with HTTP-based clients

mvidner:
## Problem

- https://trello.com/c/hvPtBtMD

When moving to the new HTTP-based architecture, we took some shortcuts. One of them was not using HTTP clients in the command-line interface. We need to adapt the following clients:

- [UsersClient](https://github.com/openSUSE/agama/blob/master/rust/agama-lib/src/users/client.rs)
- [LocalizationClient](https://github.com/openSUSE/agama/blob/master/rust/agama-lib/src/localization/client.rs)
- [ProductClient](https://github.com/openSUSE/agama/blob/master/rust/agama-lib/src/product/client.rs)
- [StorageClient](https://github.com/openSUSE/agama/blob/master/rust/agama-lib/src/storage/client.rs)
- [SoftwareClient](https://github.com/openSUSE/agama/blob/master/rust/agama-lib/src/software/client.rs)

## Solution

- `UsersClient` - still needed, because the HTTP service uses it to talk to the Ruby backend. `UsersHTTPClient` added.
- `LocalizationClient` replaced with `LocalizationHTTPClient`

The rest will be done in subsequent PRs.

## Testing

- Tested manually with various correct and incorrect configs via `agama config show` and `agama config load`. This is ripe for automation, even in this PR.
- aaand: added tests that
    -  set up a trivial HTTP server (using [httpmock](https://crates.io/crates/httpmock) - is the dependency size OK?),
    -  mocking the JSON data and
    -  asserting what the Store classes make out of it

## Screenshots


mvidner:
One odd thing, OpenAPI seems to omit the `/api` prefix for `/users/first`, but the server logs include it

mvidner:
@jreidinger do you want to have a quick look at this stage?

rust/agama-lib/src/base_http_client.rs
```diff
@@ -111,15 +116,89 @@ impl BaseHTTPClient {
...
+    /// * `object`: Object that can be serialiazed to JSON as body of request.
+    pub async fn patch(&self, path: &str, object: &impl Serialize) -> Result<(), ServiceError> {
+        let response = self
+            .request_response(reqwest::Method::PATCH, path, object)
+            .await?;
+        if response.status().is_success() {
+            Ok(())
+        } else {
+            Err(self.build_backend_error(response).await)
+        }
```
jreidinger, line 200: as those five lines is so often repeated, maybe we can also have helper method for it that gets response object and return basically Result.

mvidner:
Argh, my flawless tests fail CI with
> Error: NotAuthenticated
> test localization::store::test::test_getting_l10n ... FAILED

it looks like I mock the call changed in this PR but the auth call still goes to the live service?

rust/agama-lib/src/users/store.rs
```diff
@@ -53,10 +58,7 @@ impl<'a> UsersStore<'a> {
             password: settings.password.clone().unwrap_or_default(),
             ..Default::default()
         };
-        let (success, issues) = self.users_client.set_first_user(&first_user).await?;
-        if !success {
-            return Err(ServiceError::WrongUser(issues));
-        }
+        self.users_client.set_first_user(&first_user).await?;
```
mchf, line 61: why is suddenly possible to ignore failed case?

rust/agama-lib/src/base_http_client.rs
```diff
@@ -109,48 +112,108 @@ impl BaseHTTPClient {
...
     ///
     /// In general unless specific response handling is needed, simple post should be used.
     ///
     /// Arguments:
     ///
+    /// * `method`: for example `reqwest::Method::PUT`
     /// * `path`: path relative to HTTP API like `/questions`
     /// * `object`: Object that can be serialiazed to JSON as body of request.
-    pub async fn post_response(
+    pub async fn request_response(
```
imobachgs, line 183: Do we expect this to be part of the public API? It looks a little bit "low-level" compared with the rest (which actually rely on it). I would keep it private until it is needed.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -109,48 +112,108 @@ impl BaseHTTPClient {
...
             .map_err(|e| e.into())
     }
 
+    /// patch object at given path and report error if response is not success
+    ///
+    /// Arguments:
+    ///
+    /// * `path`: path relative to HTTP API like `/users/first`
+    /// * `object`: Object that can be serialiazed to JSON as body of request.
+    pub async fn patch(&self, path: &str, object: &impl Serialize) -> Result<(), ServiceError> {
```
imobachgs, line 203: I would expect a `patch_response` method too. Like `PATCH`, nothing prevents a response to a `PATCH` request from having a body.

rust/agama-lib/src/localization/http_client.rs
```diff
@@ -0,0 +1,26 @@
+use super::model::LocaleConfig;
+use crate::{base_http_client::BaseHTTPClient, error::ServiceError};
+
+pub struct LocalizationHTTPClient {
+    client: BaseHTTPClient,
+}
+
+impl LocalizationHTTPClient {
+    pub async fn new() -> Result<Self, ServiceError> {
```
imobachgs, line 9: We do not need this method to be async.

rust/agama-lib/src/localization/http_client.rs
```diff
@@ -0,0 +1,26 @@
...
+        Ok(Self {
+            client: BaseHTTPClient::new()?,
+        })
+    }
+
+    pub async fn new_with_base(base: BaseHTTPClient) -> Result<Self, ServiceError> {
```
imobachgs, line 15: No async needed.

rust/agama-lib/src/dbus.rs
```diff
@@ -58,7 +58,7 @@ macro_rules! property_from_dbus {
 /// NOTE: we could follow a different approach like building our own type (e.g.
 /// using the newtype idiom) and offering a better API.
 ///
-/// * `source`: hash map containing non-onwed values ([zbus::zvariant::Value]).
+/// * `source`: hash map containing non-onwed values ([enum@zbus::zvariant::Value]).
```
imobachgs, line 61: I did not know about that `enum@` prefix.

rust/agama-lib/src/localization/store.rs
```diff
@@ -1,51 +1,150 @@
...
+    localization_client: LocalizationHTTPClient,
 }
 
-impl<'a> LocalizationStore<'a> {
-    pub async fn new(connection: Connection) -> Result<LocalizationStore<'a>, ServiceError> {
+impl LocalizationStore {
+    pub async fn new() -> Result<LocalizationStore, ServiceError> {
         Ok(Self {
-            localization_client: LocalizationClient::new(connection).await?,
+            localization_client: LocalizationHTTPClient::new().await?,
```
imobachgs, line 16: When `LocalizationHTTPClient::new` is fixed, this method won't need to be async.

rust/agama-lib/src/localization/store.rs
```diff
@@ -1,51 +1,150 @@
...
         })
     }
 
+    pub async fn new_with_client(
```
imobachgs, line 20: No need to be async.

rust/agama-lib/src/users/http_client.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Returns the settings for first non admin user
+    pub async fn first_user(&self) -> Result<FirstUser, ServiceError> {
+        self.client.get("/users/first").await
+    }
+
+    /// Set the configuration for the first user
+    pub async fn set_first_user(&self, first_user: &FirstUser) -> Result<(), ServiceError> {
+        let result = self.client.put_void("/users/first", first_user).await;
+        if let Err(ServiceError::BackendError(422, ref issues_s)) = result {
```
imobachgs, line 28: np: perhaps we could use the [http::status::StatusCode](https://docs.rs/http/latest/http/status/struct.StatusCode.html) from the [HTTP crate](https://docs.rs/http/latest/http/).

rust/agama-lib/src/users/http_client.rs
```diff
@@ -0,0 +1,80 @@
...
+        value: &str,
+        encrypted: bool,
+    ) -> Result<u32, ServiceError> {
+        let rps = RootPatchSettings {
+            sshkey: None,
+            password: Some(value.to_owned()),
+            password_encrypted: Some(encrypted),
+        };
+        // TODO various errors
+        // current backend always returns 0
```
imobachgs, line 57: I would document this.

rust/agama-lib/src/users/http_client.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// SetRootSSHKey method
+    pub async fn set_root_sshkey(&self, value: &str) -> Result<u32, ServiceError> {
+        let rps = RootPatchSettings {
+            sshkey: Some(value.to_owned()),
+            password: None,
+            password_encrypted: None,
+        };
+        // TODO various errors
+        // current backend always returns 0
```
imobachgs, line 76: I would document this too.

imobachgs (commented):
Good job! In general, the PR looks good, although there are a few comments.

Checking the `BaseHTTPClient` client again and how it is used, I am still not convinced about those "raw" methods (`get_response`, `post_response`, etc.). I am not even sure if we will need them. Additionally, they leak the `Response` from `reqwest` and I am not sure whether I like that.

Perhaps it is out of the scope of this PR. Let's see if we need them at some point.

rust/agama-lib/src/users/store.rs
mvidner: The D-Bus client `set_first_user` returns a tuple `(success,issues)` in band, as `zbus::Result<(bool, Vec<String>)>`
https://github.com/openSUSE/agama/blob/b9a73b57fc71f6eb47b3f98154ec61a31f4d57d9/rust/agama-lib/src/users/client.rs#L84

But the HTTP client `set_first_user` returns a `Result<(), ...>` where all the errors are moved to the `Err` alternative, and the `WrongUser` is handled on that lower level:
https://github.com/openSUSE/agama/blob/3dca6b5ed6a30389b0023aea9a1ea031934a5859/rust/agama-lib/src/users/http_client.rs#L26

(and it is the `?` in this line that propagates the `Err`)

rust/agama-lib/src/dbus.rs
mvidner: It's a disambiguation suggested by `rustdoc` because there is also a `#[derive(Value)]`

rust/agama-lib/src/users/http_client.rs
mvidner: Yup, but then I can't use `if let` like this... that's fine, a `match` will look good too

imobachgs (commented):
It looks better now, although I am not convinced on the use case of those `_void` functions.

imobachgs (commented):
It looks good to me but, please, add a changes entry.

rust/agama-lib/src/localization/store.rs
```diff
@@ -142,9 +143,11 @@ mod test {
         };
         let result = store.store(&settings).await;
 
+        // main assertion
```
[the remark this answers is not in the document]
mvidner: The resulting behavior is of course identical.
Writing it in this way means to draw your attention to the main assertion of the test.

Actually I was bothered by the test code being obscure that I even looked up [Rust RSpec](https://github.com/rust-rspec/rspec) in the hope to write tests that better express the intent. Well, BDD.

imobachgs (approved):
LGTM. Thanks!

About the `*_void` methods, we agreed to review our HTTP API to be more consistent and, eventually, we could get rid of them. But it is out of the scope of this PR.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -109,48 +112,108 @@ impl BaseHTTPClient {
...
     }
 
-    /// post object to given path and returns server response. Reports error only if failed to send
+    /// put object to given path and report error if response is not success
+    ///
+    /// Arguments:
+    ///
+    /// * `path`: path relative to HTTP API like `/users/first`
+    /// * `object`: Object that can be serialiazed to JSON as body of request.
+    pub async fn put_void(&self, path: &str, object: &impl Serialize) -> Result<(), ServiceError> {
```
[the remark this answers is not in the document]
mvidner: Right, I've added a body-returning `put`

rust/agama-lib/src/base_http_client.rs
mvidner: Made this private. The public API is now dealing with deserialized objects (in the successful cases)

rust/agama-lib/src/base_http_client.rs
mvidner: Right. Have `patch` and `patch_void` now.

PR #1439, opened 2024-07-04
refactor(web): adopt TanStack Query for state management

imobachgs:
Trello: https://trello.com/c/8u1WOJz4/3723-8-agama-ui-adopt-a-state-management-solution

Agama's UI state management could be better. For that reason, and after evaluating a few alternatives (see the list of considered alternatives), we decided to adopt [Tanstack Query (former React Query)](https://tanstack.com/query/v5).

## Why TanStack query

* It offers a mechanism (based on promises) to fetch/update the state.
* It is not limited to the client state: it works great with the server state, too.
* It takes care of caching, re-fetching, error handling, etc.
* It is a mature, well-documented and widely known project.
* Plays nicely with [React Router](https://reactrouter.com/), although a [TanStack Router](https://tanstack.com/router/latest) project exits.

## What is included in this PR

There is quite some work to do so let's consider this pull request as the first step in this path. It introduces the following changes:

- Add TanStack query as a dependency.
- Replace `L10Provider` with a queries-based approach, including data mutation. Check [queries/l10n.js](https://github.com/openSUSE/agama/blob/react-query/web/src/queries/l10n.js) if you are curious.
- Drop the parts of the languages client that are not relevant anymore.

## Considered Alternatives

- [Redux](https://react-redux.js.org/)
- [Jotai](https://jotai.org/)
- [Zustand](https://docs.pmnd.rs/zustand/getting-started/introduction)
- [Tanstack Query (former React Query)](https://tanstack.com/query/v5)
- [Recoil](https://recoiljs.org/)

We had a look at others but did not consider them at all.


joseivanlopez (approved):
Great work!

joseivanlopez (changes requested):
Wait, changelog entry is missing.

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Jul  8 10:56:14 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Introduce TanStack Query to handle for data fetching and state
```
joseivanlopez, line 4: NP: "to handle for" sound wrong.

PR #1441, opened 2024-07-04
web: UI code pruning and clean up `round #1`

dgdavid:
## Problem

[Switching from the former UI to the current approach](https://github.com/openSUSE/agama/pull/1202) left behind a lot of things to review, dead components and a CSS clean up among them.

## Solution

Drop dead components and make as much clean-up rounds as needed. This can be considered `Round #1`, made in the middle of more urgent work.

## Testing

Tested manually

## Screenshots

Visual changes after this PR should keep at minimum. In any case, it's preferred that reviewers give the UI a shot taking the time they need.


web/src/components/storage/ProposalActionsDialog.jsx
```diff
@@ -29,17 +29,19 @@ const ActionsList = ({ actions }) => {
   // Some actions (e.g., deleting a LV) are reported as several actions joined by a line break
   const actionItems = (action, id) => {
     return action.text.split("\n").map((text, index) => {
+      const Wrapper = action.delete ? "strong" : "span";
```
dgdavid, line 32: Here I'd like to hear opinion from _storage people_. @ancorgs, to mention one.

Please, take your time to check https://html.spec.whatwg.org/#the-b-element and  https://html.spec.whatwg.org/#the-strong-element. What element do you think suits better for delete actions. In my view `strong` is the right one if my understanding about their importance is accurate.

imobachgs (approved):

web/src/components/storage/ProposalActionsDialog.jsx
ancorgs: Yes. Looks like `strong` is the right thing.

PR #1448, opened 2024-07-08
New storage proposal

ancorgs:
Temporary PR created only to illustrate some conversations I'm having with @joseivanlopez 

This depends on https://github.com/yast/yast-storage-ng/pull/1386


joseivanlopez:
Thinking aloud:

~~~json

"drives": [
  {
    "search": "/dev/vda",
    "partitions": [ "boot" ]
  }
]

~~~

service/lib/y2storage/proposal/agama_device_planner.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+
+      # @param planned [Planned::Disk, Planned::Partition]
+      # @param settings [Agama::Storage::Settings::Mount]
+      def configure_mount(planned, settings)
+        planned.mount_point = settings.path
+        planned.mount_by = settings.mount_by
+        planned.fstab_options = settings.options
+        # FIXME: Is this needed? Or #options is enough?
+        # planned.read_only = settings.read_only?
```
ancorgs, line 92: Checking how this `read_only` is used so far. For both the Guided and the AutoYaST proposals, this is used only for Btrfs planned filesystems, honoring `VolumeSpecification#btrfs_read_only`.

```ruby
class VolumeSpecification
  attr_accessor :btrfs_read_only
end

module Proposal
  class DevicesPlanner
    def adjust_btrfs(planned_device, volume)
      return unless planned_device.btrfs?
      
      #[...]
      planned_device.read_only = volume.btrfs_read_only?
    end 
  end 
end 

module Proposal
  class AutoinstDrivePlanner
    def add_filesystem_attrs(device, partition_section)
      #[...]
      device.read_only = read_only?(partition_section.mount)
    end

    def read_only?(mount_point)
      return false unless mount_point

      spec = VolumeSpecification.for(mount_point)
      !!spec && spec.btrfs_read_only?
    end
  end
end
```

When committing it to the system, the `read_only` flag is indeed translated to the `fstab_options`.

```ruby
module Planned
  module CanBeMounted
    attr_accessor :read_only
  end
end

module Planned
  module CanBeFormatted
    def fstab_options_for(mount_point)
      #[...]
      options.push("ro") if read_only && !options.include?("rw")
      #[...]
    end
  end
end
```

service/lib/y2storage/proposal/agama_device_planner.rb
```diff
@@ -0,0 +1,124 @@
...
+          planned_partition(partition_settings).tap { |p| p.disk = settings.found_device.name }
+        end
+      end
+
+      # @param settings [Agama::Storage::Settings::Partition]
+      # @return [Planned::Partition]
+      def planned_partition(settings)
+        Planned::Partition.new(nil, nil).tap do |planned|
+          planned.partition_id = settings.id
+          # FIXME: do we want to allow to force primary partitions? That was not really
```
ancorgs, line 115: I would remove `Partition.type` (string) from the schema and would totally let the algorithm decide whether a partition should be primary or logical.

service/lib/y2storage/proposal/agama_device_planner.rb
ancorgs: We moved the topic a bit forward to https://trello.com/c/4Cm7lh1j/492-storage-profile-filesystem-related-properties-especially-btrfs-ones

service/lib/y2storage/proposal/agama_device_planner.rb
ancorgs: Removed.

ancorgs:
Closing in favor of #1552

PR #1449, opened 2024-07-09
fix(web): postpone creation of the WSClient

imobachgs:
The web user interface tries to connect to the WebSocket from the very beginning (most likely
because of
[network/routes.js](https://github.com/openSUSE/agama/blob/master/web/src/components/network/routes.js#L32).

This PR introduces an HTTPClient#ws method that builds the WSClient lazily so the initialization is
postponed until it is required.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -2,7 +2,7 @@
 Tue Jul  9 08:51:34 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Do not try to connect to the WebSocket until the user is
-  logged in (gh#openSUSE/agama#).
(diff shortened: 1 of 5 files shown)


web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Jul  9 08:51:34 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not try to connect to the WebSocket until the user is
+  logged in (gh#openSUSE/agama#).
```
teclator, line 5:
```suggestion
  logged in (gh#openSUSE/agama#1449).
```

teclator (approved):
LGTM

PR #1450, opened 2024-07-09
Bring back PXE image for remote installations

teclator:
## Problem

We dropped PXE images when the ALP images were removed (see https://github.com/openSUSE/agama/pull/1181).

## Solution

Bring back PXE images for remote installations.


imobachgs:
Good catch! Please, update the changes file.

imobachgs (approved):

PR #1451, opened 2024-07-10
add post method to create new question

jreidinger:
## Problem

http interface currently does not have a way to create question with it. But it is needed to e.g. ask questions from autoyast converter script.

## Solution

Add it and together also remove password from question as it should not be preset.

### DIFF
--- a/rust/agama-server/src/questions/web.rs
+++ b/rust/agama-server/src/questions/web.rs
@@ -8,7 +8,7 @@
 use crate::{error::Error, web::Event};
 use agama_lib::{
     error::ServiceError,
-    proxies::{GenericQuestionProxy, QuestionWithPasswordProxy},
+    proxies::{GenericQuestionProxy, QuestionWithPasswordProxy, Questions1Proxy},
 };
 use anyhow::Context;
 use axum::{
@@ -16,6 +16,7 @@ use axum::{
     routing::{get, put},
     Json, Router,
 };
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, pin::Pin};
 use tokio_stream::{Stream, StreamExt};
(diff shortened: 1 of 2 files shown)


rust/agama-server/src/questions/web.rs
```diff
@@ -46,6 +49,46 @@ impl<'a> QuestionsClient<'a> {
...
+                    data,
+                )
+                .await?
+        };
+        let mut res = question.clone();
+        // we are sure that regexp is correct, so use unwrap
+        let id_matcher = Regex::new(r"/(?<id>\d+)$").unwrap();
+        let Some(id_cap) = id_matcher.captures(path.as_str()) else {
+            panic!("Id not in path")
+        }; // TODO: better error if path does not contain id
```
imobachgs, line 87: Does this code crash the server?

imobachgs (commented):
In general it looks OK. Perhaps we should plan to refactor the questions a bit so the D-Bus service and the HTTP API are just that... APIs. In the network area, we use message passing and I think for questions we should do something similar (and integrate the D-Bus and the HTTP services into a single service).

But that's out of scope.

rust/agama-server/src/questions/web.rs
jreidinger: yeap, it terminates...but it should never happen as new question if method succeed should always return id as part of path

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jul 10 10:01:18 UTC 2024 - Josef Reidinger <jreidinger@suse.com>
+
+- Add to HTTP API method POST for question to ask new question
+  (gh#openSUSE/agam#1451)
```
```suggestion
  (gh#openSUSE/agama#1451) 
```

imobachgs (commented):
Just a typo. Otherwise, LGTM.

imobachgs (approved):

PR #1453, opened 2024-07-10
feat(http): implement get answer and delete question

jreidinger:
## Problem

Currently there is no way how to get answer for asked question ( service do it invisible on dbus ) and also delete is not exposed ( same case that services do it hidden on dbus ).

## Solution

Implement `DELETE /questions/:id` and `GET /questions/:id/answer` HTTP calls.


imobachgs (commented):
Just some minor comments. Otherwise, it looks good to me.

PR #1455, opened 2024-07-11
feat(schema): basic storage settings

joseivanlopez:
Extend JSON schema to partially support the new *storage config* described in this document: https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md.

This PR only extends the schema for the most basic settings (i.e., *boot*, *drives* and *partitions*). Settings for advanced searches, delete devices, creeate RAIDs, etc are not considered yet.

Notes:

* The *search* section added in this PR is based on the following proposal: https://gist.github.com/joseivanlopez/4f2381eb1f6c3aca562aadc2ad920a72?permalink_comment_id=5121860#gistcomment-5121860. This will be revisited later.
* The *guided* schema added by https://github.com/openSUSE/agama/pull/1263 is not going to be offered (e.g., *$ agama config load* will not understand it). The *storage config* will cover all the use cases and features provided by the *guided config*.
* The *guided config* will be used only by clients (e.g., web UI) to easily configure a subset of the use cases supported by the *storage config*.

### DIFF
--- a/doc/auto_storage.md
+++ b/doc/auto_storage.md
@@ -126,6 +126,7 @@ Partition
   encryption [Encryption]
   filesystem [<Filesystem>]
   delete [<boolean=false>]
+  deleteIfNeeded [<boolean=false>]
 
 LogicalVolume
   search [<Search>]
@@ -139,7 +140,7 @@ LogicalVolume
   encryption [Encryption]
   filesystem [<Filesystem>]
   delete [<boolean=false>]
-
+  deleteIfNeeded [<boolean=false>]
 Encryption
   reuse <Boolean>
   type <EncryptionType>
@@ -255,6 +256,10 @@ into the following:
  - If the product does not specify a default volume, the behavior is still not defined (there are
    several reasonable options).
 
+It is also possible to specify "current" as a size value for partitions and logical volumes that
+already exist in the system. The usage of "current" and how it affects resizing the corresponding
+devices is explained at a separate section below.
+
 ### Under Discussion
 
 As explained, it should be possible to specify the sizes as "default", as a range or as a fixed
@@ -264,10 +269,11 @@ represent a size. The following two possibilities are also under consideration.
  - `{ "gib": 40 }`
  - `{ "value": 40, "units": "gib" }`
 
-Resizing is also an open topic. We may not be able to easily combine a resize operation with aspects
-like ranges (min and max) or automatic sizes. We also need to define how to specify something like
-"shink as much as needed to allocate the other partitions". See the section below about deleting and
-shrinking.
+## Partitions Needed for Booting
+
+Using a `boot` entry makes it possible to configure whether (and where, using an alias) Agama
+should calculate and create the extra partitions needed for booting. If the device is not
+specified, Agama will take the location of the root file system as a reference.
 
 ## Searching Existing Devices
 
@@ -548,11 +554,6 @@ system (so the same conditions can be matched by a disk, a partition, an LVM dev
 }
 ```
 
-## Partitions Needed for Booting
-
-Using a `boot` entry makes it possible to configure whether (and where, using an alias) Agama
-should calculate and create the extra partitions needed for booting.
-
 ## Keeping an Existing File System or Encryption Layer
 
 The entries for both `encryption` and `filesystem` contain a flag `reuse` with a default value of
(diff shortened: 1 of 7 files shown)


rust/agama-lib/share/examples/storage-guided.json
```diff
@@ -0,0 +1,63 @@
...
+          },
+          "target": {
+            "newVg": "/dev/vda"
+          }
+        },
+        {
+          "mount": {
+            "path": "swap"
+          },
+          "filesystem": "swap",
```
jreidinger, line 54: is it smart enough that if omited it will use swap fs automatically?

rust/agama-lib/share/examples/storage-guided.json
```diff
@@ -0,0 +1,63 @@
...
+          "mount": {
+            "path": "/",
+            "options": ["ro"]
+          },
+          "filesystem": {
+            "btrfs": {
+              "snapshots": true
+            }
+          },
+          "size": [1024, "5 Gib"],
```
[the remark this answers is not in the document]
joseivanlopez: The schema defines what it means: [min, max].

rust/agama-lib/share/examples/storage-guided.json
joseivanlopez: The behavior does not depend on the schema. We can decide in the implementation of the proposal what to do:

* Report an error.
* Assume swap fs if no fs is given.
* Assume swap fs if other fs is given.

One thing that could go to the schema is to force `"filesystem": "swap"` if mount is *swap* and vice versa. But, for now, I would not make the schema more complex.

ancorgs:
Some aspects of the current specification that would benefit from some discussion:

- Explicit `create` vs relying on the device type and the presence of `search`
- The need for `ifNotFound` (and its possible values)
  - Current values: `skip` (ignore device entry), `create` (turn the search into a creation), `error` (panic)
- Separate `format` + `mount` vs a single setting (eg. `filesystem`)
- Delete policies per device vs relying only on `partitions: [ search: {} ]`

doc/auto_storage.md
```diff
@@ -516,23 +537,53 @@ system (so the same conditions can be matched by a disk, a partition, an LVM dev
...
+specify any other usage (like declaring a file system on it). But resizing a partition is more on
+the grey area. Often some partitions or logical volumes are resized only to make space for the
+declared devices. But since resizing is not a destructive operation it can also make sense to
+declare a given partition must be resized and then formatted.
+
+The exact syntax to specify all the possible actions is still under discussion. The current document
+shows how a `delete` attribute could be used in combination with `search` to specify the devices to
+delete. An alternative could be to have separate section for each device to specify how to delete
+and/or shrink its partitions or logical volumes, instead of integrating that information into the
+corresponding `partitions` or `logical_volumes` attributes.
```
ancorgs, line 578: Regarding this, I think I would kill all the current `delete` flags. In fact, the current `spaceMaker` only deletes partitions (and maybe logical volumes in the future). Deleting MDs and VGs is not considered by the SpaceMaker or by AutoYaST nowadays.

Then I would add a "space_actions" (tentative name) to everything that could contain partitions (`Drive` and `MdRaid`) and maybe also in the future also to volume groups in order to decide what to do with the existing LVs.

Each entry of that section would consist on one `search` (same syntax and context than the corresponding `partitions` section), one (optional) `delete` and one (also optional) `shrink`.

Example - delete everything and create a root:

```json
"drives": [
   {
     "partitions": [
        { 
           "size": { "min": "1 GiB" },
           "filesystem": { "path": "/", "type": "btrfs" }
        }
     ],
     "space_actions": [
        { 
           "search": {},
           "delete": true
        }
      ]
   }
]
```

Example - try resizing partitions first and deleting them is that's not enough:

```json
"drives": [
   {
     "partitions": [
        { 
           "size": { "min": "1 GiB" },
           "filesystem": { "path": "/", "type": "btrfs" }
        }
     ],
     "space_actions": [
        { 
           "search": {},
           "delete": "onDemand",
           "shrink": "onDemand"
        }
      ]
   }
]
```


Example - Keep the first partition, resizing it if needed, and delete the rest

```json
"drives": [
   {
     "partitions": [
        { 
           "size": { "min": "1 GiB" },
           "filesystem": { "path": "/", "type": "btrfs" }
        }
     ],
     "space_actions": [
        { 
           "search": { "max": 1 },
           "shrink": "onDemand"
        },
        { 
           "search": {}
           "delete": true
        }
      ]
   }
]
```

Example - Keep the biggest partition, resizing it to a certain size, and delete the rest if needed

```json
"drives": [
   {
     "partitions": [
        { 
           "size": { "min": "1 GiB" },
           "filesystem": { "path": "/", "type": "btrfs" }
        }
     ],
     "space_actions": [
        { 
           "search": { "sort": { "property": "sizeKib", "order": "desc" }, "max": 1 },
           "shrink": "15 GiB"
        },
        { 
           "search": {}
           "delete": "onDemand"
        }
      ]
   }
]
```

ancorgs: Re-reading my proposal... I don't really like it. But I think some points can be taken as valid. For example:

>  Shrinking a partition to make space would be the only kind of resizing I would support. No support for enlarging existing partitions or for mounting an existing partition and also changing its size.

doc/auto_storage.md
```diff
@@ -192,13 +210,11 @@ it) to allocate two file systems.
...
                     "size":   { "min": "10 GiB" },
-                    "format": { "filesystem": "btrfs" },
-                    "mount":  { "path": "/" }
+                    "filesystem": { "path": "/", "type": "btrfs" }
                 },
                 {
                     "size":   "2 GiB",
-                    "format": { "filesystem": "swap" },
-                    "mount":  { "path": "swap" }
+                    "format": { "path": "swap", "type": "swap" }
```
joseivanlopez, line 217: s/format/filesystem

rust/agama-lib/share/examples/storage-guided.json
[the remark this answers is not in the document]
mvidner: @joseivanlopez true, but the schema also allows a `{"min": min, "max": max }` alternative. In light of @jreidinger 's comment, maybe we should remove the pair alternative, or discourage it.
https://github.com/openSUSE/agama/pull/1455/files#diff-e7dcc0a7344de41b0cc7ca87ba9624848c4160406f9d618a2ae025751f6c7bf7R601-R627

rust/agama-lib/share/examples/storage-guided.json
```diff
@@ -0,0 +1,63 @@
...
+          "target": "default"
+        },
+        {
+          "mount": {
+            "path": "/home"
+          },
+          "filesystem": "xfs",
+          "size": {
+            "min": "5 GiB",
+            "max": "20 GiB"
```
mvidner, line 44: ... which is illustrated here

rust/agama-lib/share/profile.schema.json
```diff
@@ -568,26 +560,227 @@
...
+      "title": "Password-based key derivation function to use for LUKS2.",
+      "enum": ["pbkdf2", "argon2i", "argon2id"]
+    },
+    "encryption": {
+      "title": "Encryption options.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["key"],
+      "properties": {
+        "key": {
```
ancorgs, line 675: I don't like the name "key" because is too ambiguous (not to say plain wrong). With LUKS there is a _master key_, several _user keys_,  potentially _FIDO2 security keys_ and more types. And guess what, none of them correspond to the passphrase. The passphrase is not a "key" (the actual user keys are derived from the passphrase using the PBKDF).

So the correct name here would be `password` or `passphrase`.

rust/agama-lib/share/profile.schema.json
```diff
@@ -568,26 +560,227 @@
...
+        "label": {
+          "title": "LUKS label for the encrypted device.",
+          "type": "string"
+        },
+        "cipher": {
+          "title": "Cipher used for LUKS encryption.",
+          "description": "The value must be compatible with the --cipher argument of the command cryptsetup.",
+          "type": "string"
+        },
+        "key_size": {
```
ancorgs, line 694: Is snake_case expected here? I thought we were using camelCase in all the schema (like pbkdFunction just some lines above).

rust/agama-lib/share/profile.schema.json
joseivanlopez: Ok. I have renamed it as `password`.

rust/agama-lib/share/profile.schema.json
```diff
@@ -568,26 +560,227 @@
...
+    },
+    "encryptionPbkdFunction": {
+      "title": "Password-based key derivation function to use for LUKS2.",
+      "enum": ["pbkdf2", "argon2i", "argon2id"]
+    },
+    "encryption": {
+      "title": "Encryption options.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["key"],
```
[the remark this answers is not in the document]
joseivanlopez: For now, the available encryption methods (LUKS2 and TPM) require a password. I would keep password as required property. If we add any other encryption method with optional password, then we could define separate schemas for each one.

rust/agama-lib/share/profile.schema.json
joseivanlopez: Good point.

rust/agama-lib/share/profile.schema.json
```diff
@@ -303,41 +303,98 @@
...
+                "properties": {
+                  "search": {
+                    "description": "The search is limited to drives scope.",
+                    "$ref": "#/$defs/search"
+                  },
+                  "encryption": {
+                    "$ref": "#/$defs/encryption"
+                  },
+                  "filesystem": {
+                    "description": "The partition table (if any) is deleted.",
```
ancorgs, line 329: Does this comment belong to `filesystem`?

rust/agama-lib/share/profile.schema.json
```diff
@@ -303,41 +303,98 @@
...
+              {
+                "title": "Partitioned drive",
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                  "search": {
+                    "description": "The search is limited to drives scope.",
+                    "$ref": "#/$defs/search"
+                  },
+                  "encryption": {
```
ancorgs, line 343: Does encryption make sense here?

rust/agama-lib/share/profile.schema.json
```diff
@@ -303,41 +303,98 @@
...
+                  }
+                }
+              }
+            ]
+          }
+        },
         "guided": {
-          "title": "Settings to execute a Guided Proposal",
+          "title": "Guided proposal settings",
+          "$comment": "Guided settings will be removed from this schema.",
```
ancorgs, line 361: Not sure whether the new title is better than the previous. But, in any case, the `$comment` could be more clear.

rust/agama-lib/share/profile.schema.json
```diff
@@ -625,27 +608,311 @@
...
+        {
+          "title": "Automatic size",
+          "description": "The size is auto calculated according to the product.",
+          "const": "auto"
+        },
+        {
+          "$ref": "#/$defs/sizeValue"
+        },
+        {
+          "title": "Size range (tuple systax)",
```
ancorgs, line 649: systax

rust/agama-lib/share/profile.schema.json
joseivanlopez: In fact, I would like to avoid the proposal word. The user of the config only has to know about configs, nothing about internals like proposal.

rust/agama-lib/share/profile.schema.json
joseivanlopez: Maybe not. I will remove it.

rust/agama-lib/share/profile.schema.json
joseivanlopez: Good point!

rust/agama-lib/share/profile.schema.json
joseivanlopez: Regarding the title, I am trying to use something short, and a description for longer explanations.

ancorgs (approved):
LGTM

PR #1456, opened 2024-07-11
feat(web): start using TypeScript

dgdavid:
## Problem

Agama web UI is using plain JavaScript but relying in TypeDoc for performing type checking. It is ok, but as the project grows looks like would be easier and straight forward to use TypeScript instead.

## Solution

Start using TypeScript and migrate the current code little by little as the files are touched.

## Testing

* Adapted Jest for working with TypeScript too.


web/src/queries/l10n.ts
```diff
@@ -123,26 +123,22 @@ const useL10nConfigChanges = () => {
...
-      configQuery(),
-      localesQuery(),
-      keymapsQuery(),
-      timezonesQuery()
-    ]
-  });
+  const [{ data: config }, { data: locales }, { data: keymaps }, { data: timezones }] =
+    useSuspenseQueries({
+      queries: [configQuery(), localesQuery(), keymapsQuery(), timezonesQuery()]
+    });
```
dgdavid, line 129: Ouch, I didn't realize I sent these format changes made by Zed editor which I'm testing. Will revert them in a new commit.

dgdavid: After a short discussion with @imobachgs, I will keep it as it is because we're gonna format all the code base for consistence.

dgdavid: BTW, the format change was not because Zed but because Zed applying the rules we have at .prettierrc file

imobachgs (approved):

web/src/queries/l10n.ts
dgdavid: > we're gonna format all the code base for consistence.

Done at https://github.com/openSUSE/agama/pull/1460

PR #1457, opened 2024-07-12
Ask cli

jreidinger:
## Problem

There is no CLI command to ask new question. It is needed to ask questions from autoyast convert script.
Also there is a need to share functionality between various CLI calls that will use HTTP API as current one in network has limitation and is not designed as generic http agent that fits agama.

## Solution

Implement needed functionality in questions http client.
Implement BaseHttpClient that shares code for doing http requests on agama http server.
Implement ask and list subcommand for questions to allow asking questions and also listing currently unanswered ones.


rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+        let client = Client::builder().default_headers(headers).build()?;
+
+        Ok(Self {
+            client,
+            base_url: API_URL.to_string(), // TODO: add support for remote server
+        })
+    }
+
+    const NO_TEXT: &'static str = "No text";
+    // Simple wrapper around Response to get target type.
```
imobachgs, line 34: What "target type" means?

rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+    // For more advanced usage use directly get method
+    pub async fn get_type<T: DeserializeOwned>(&self, path: &str) -> Result<T, ServiceError> {
+        let response = self.get(path).await?;
+        if !response.status().is_success() {
```
imobachgs, line 38: np: What about returning early if `is_success()`? I am saying just because the error handling seems to be longer.

rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    pub async fn get(&self, path: &str) -> Result<Response, ServiceError> {
+        self.client
+            .get(self.target_path(path))
+            .send()
+            .await
+            .map_err(|e| e.into())
+    }
+
+    fn target_path(&self, path: &str) -> String {
```
imobachgs, line 58: What about using `request_url`? (or just `url`)?

rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+        Ok(Self {
+            client,
+            base_url: API_URL.to_string(), // TODO: add support for remote server
+        })
+    }
+
+    const NO_TEXT: &'static str = "No text";
+    // Simple wrapper around Response to get target type.
+    // For more advanced usage use directly get method
+    pub async fn get_type<T: DeserializeOwned>(&self, path: &str) -> Result<T, ServiceError> {
```
imobachgs, line 36: I like the idea but `get_type` is kind of confusing (am I getting the "type"?).

In any case, I would call this method `get` and rename the `get` method below to something like `get_raw` as I expect us to use this more often.

rust/agama-lib/src/questions/model.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub with_password: Option<QuestionWithPassword>,
+}
+
+/// Facade of agama_lib::questions::GenericQuestion
+/// For fields details see it.
+/// Reason why it does not use directly GenericQuestion from lib
+/// is that it contain both question and answer. It works for dbus
+/// API which has both as attributes, but web API separate
+/// question and its answer. So here it is split into GenericQuestion
+/// and GenericAnswer
```
imobachgs, line 18: Perhaps we should consider refactoring the `GenericQuestion` from lib to be a structure composed of question and answer structs from this module. This is just an idea.

imobachgs (commented):
Apart from some minor naming issues (you know, naming is hard) and some missing docs, I think you are on the right track.

rust/agama-cli/src/main.rs
```diff
@@ -118,28 +118,28 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
     Ok(ManagerClient::new(conn).await?)
 }
 
-async fn run_command(cli: Cli) -> anyhow::Result<()> {
-    match cli.command {
+async fn run_command(cli: Cli) -> Result<(), ServiceError> {
+    Ok(match cli.command {
```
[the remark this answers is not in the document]
jreidinger: I can use `let result = match` and then `Ok(result)`

rust/agama-lib/src/questions/http_client.rs
```diff
@@ -0,0 +1,20 @@
+use crate::error::ServiceError;
+
+use super::model;
+
+pub struct HTTPClient {
```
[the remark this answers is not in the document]
jreidinger: BaseHTTPClient sounds the best for me.

rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+const API_URL: &str = "http://localhost/api";
+
+impl HTTPClient {
+    // if there is need for client without authorization, create new constructor for it
+    pub async fn new() -> Result<Self, ServiceError> {
+        let token = AuthToken::find().context("You are not logged in")?;
+
+        let mut headers = header::HeaderMap::new();
+        let value = header::HeaderValue::from_str(format!("Bearer {}", token).as_str())
+            .map_err(|e| ServiceError::NetworkClientError(e.to_string()))?;
```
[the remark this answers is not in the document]
jreidinger: ah, my fault. It should be generic one. Thanks for catch ( c&p error )

rust/agama-lib/src/http_client.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct HTTPClient {
+    client: Client,
+    pub base_url: String,
+}
+
+const API_URL: &str = "http://localhost/api";
+
+impl HTTPClient {
+    // if there is need for client without authorization, create new constructor for it
+    pub async fn new() -> Result<Self, ServiceError> {
```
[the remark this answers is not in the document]
jreidinger: right, I thought I see some await inside, but I was wrong.

rust/agama-lib/src/http_client.rs
jreidinger: it is basically to get object you expect from backend. Something like `let questions : Vec<Question> = client.get_type("/questions")`

rust/agama-lib/src/http_client.rs
jreidinger: agreed. I am even thinking about naming this one as `get_object` but it is right, that plain get here and maybe `get_response` would be better

rust/agama-lib/src/http_client.rs
jreidinger: BTW one more idea I had...I know that we use `400` for error and in body we use json for ServiceError...so do you think it make sense to have something like `ErrorFromBackend` that gets as argument `ServiceError` from response?

rust/agama-lib/src/http_client.rs
jreidinger: fine for me...`url` is probably better as it is internal only method..so shorter is better.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub async fn get<T: DeserializeOwned>(&self, path: &str) -> Result<T, ServiceError> {
+        let response = self.get_response(path).await?;
+        if response.status().is_success() {
+            response.json::<T>().await.map_err(|e| e.into())
+        } else {
+            Err(self.build_backend_error(response).await)
+        }
+    }
+
+    /// Calls GET method on given path and return Response that can be further
```
imobachgs, line 63:
```suggestion
    /// Calls GET method on the given path and returns [`Response`] that can be further
```

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+            .send()
+            .await
+            .map_err(|e| e.into())
+    }
+
+    fn url(&self, path: &str) -> String {
+        self.base_url.clone() + path
+    }
+
+    /// post object to given path and report error if response is not success
```
imobachgs, line 77: Please, describe the arguments.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+            Ok(())
+        } else {
+            Err(self.build_backend_error(response).await)
+        }
+    }
+
+    /// post object to given path and returns server response. Reports error only if failed to send
+    /// request, but if server returns e.g. 500, it will be in Ok result.
+    /// In general unless specific response handling is needed, simple post should be used.
+    pub async fn post_response(
```
imobachgs, line 90: Please, describe the arguments.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+    ) -> Result<Response, ServiceError> {
+        self.client
+            .post(self.url(path))
+            .json(object)
+            .send()
+            .await
+            .map_err(|e| e.into())
+    }
+
+    /// delete call on given path and report error if failed
```
imobachgs, line 103: Please, document the arguments.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// TODO: do not need variant with request body? if so, then create additional method.
+    pub async fn delete_response(&self, path: &str) -> Result<Response, ServiceError> {
+        self.client
+            .delete(self.url(path))
+            .send()
+            .await
+            .map_err(|e| e.into())
+    }
+
+    pub async fn build_backend_error(&self, response: Response) -> ServiceError {
```
imobachgs, line 125: Perhaps you might want to implement something like `From<Response>` (or `TryFrom`).

rust/agama-lib/src/error.rs
```diff
@@ -16,6 +16,8 @@ pub enum ServiceError {
     DBusProtocol(#[from] zbus::fdo::Error),
     #[error("Unexpected type on D-Bus '{0}'")]
     ZVariant(#[from] zvariant::Error),
+    #[error("Failed to communicate with HTTP backend '{0}'")]
```
imobachgs, line 19:
```suggestion
    #[error("Failed to communicate with the HTTP backend '{0}'")]
```

rust/agama-lib/src/questions.rs
```diff
@@ -1,8 +1,12 @@
 //! Data model for Agama questions
 
 use std::collections::HashMap;
+pub mod http_client;
+pub mod model;
 
 /// Basic generic question that fits question without special needs
+/// NOTE: structs below is for dbus usage and holds complete questions data
```
imobachgs, line 8:
```suggestion
/// NOTE: structs below is for D-Bus usage and holds complete questions data
```

imobachgs: Please, add a blank line to separate the `NOTE` from the rest (otherwise, everything gets rendered on the same line).

Additionally, I would drop the 'NOTE' prefix. Last but not least, "below" is inaccurate because each struct has a separate page.

imobachgs (commented):
In general it is OK. Just some comments about the documentation.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl BaseHTTPClient {
+    // if there is need for client without authorization, create new constructor for it
+    pub fn new() -> Result<Self, ServiceError> {
+        let token = AuthToken::find().context("You are not logged in")?;
+
+        let mut headers = header::HeaderMap::new();
+        // just use generic anyhow error here as Bearer format is constructed by us, so failures can come only from token
+        let value = header::HeaderValue::from_str(format!("Bearer {}", token).as_str())
+            .map_err(|e| anyhow::Error::new(e))?;
```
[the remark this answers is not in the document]
jreidinger: me neither...I basically use it for trash error which should not happen instead of unwrap

rust/agama-cli/src/questions.rs
```diff
@@ -19,6 +19,8 @@ pub enum QuestionsCommands {
         /// Path to a file containing the answers in YAML format.
         path: String,
     },
+    /// prints list of questions that is waiting for answer in YAML format
```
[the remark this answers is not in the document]
jreidinger: it is fine for me if agreed on it

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        let client = Client::builder().default_headers(headers).build()?;
+
+        Ok(Self {
+            client,
+            base_url: API_URL.to_string(), // TODO: add support for remote server
+        })
+    }
+
+    const NO_TEXT: &'static str = "No text";
```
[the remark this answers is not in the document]
jreidinger: To see that error from backend does not provide any text. For me less confusing to write explicitelly that no text received from backend.

rust/agama-lib/src/base_http_client.rs
jreidinger: Can it be asynchronous? As it is not simple conversion but it also ask for text.

rust/agama-cli/src/questions.rs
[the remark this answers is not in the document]
jreidinger: reason why I use yaml for it is that in answers file we used for automatic answers we also use yaml.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+    client: Client,
+    pub base_url: String,
+}
+
+const API_URL: &str = "http://localhost/api";
+
+impl BaseHTTPClient {
+    // if there is need for client without authorization, create new constructor for it
+    pub fn new() -> Result<Self, ServiceError> {
+        let token = AuthToken::find().context("You are not logged in")?;
```
mvidner, line 34: The error message is correct (I hope) but even better would be a hint how to fix the situation: `"... Try: agama auth login"`?

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+            .await
+            .map_err(|e| e.into())
+    }
+
+    pub async fn build_backend_error(&self, response: Response) -> ServiceError {
+        let code = response.status().as_u16();
+        let text = response
+            .text()
+            .await
+            .unwrap_or_else(|_| Self::NO_TEXT.to_string());
```
mvidner, line 168: "No text" -> `"(failed to extract error text from HTTP response)"` is perhaps too verbose but I would appreciate it when debugging it as an operator

rust/agama-lib/src/base_http_client.rs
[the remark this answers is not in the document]
mvidner: Please move the const definition closer to the method that uses it. That also makes it more apparent what it should say

rust/agama-lib/src/error.rs
```diff
@@ -33,6 +35,8 @@ pub enum ServiceError {
     UnsuccessfulAction(String),
     #[error("Unknown installation phase: '{0}")]
     UnknownInstallationPhase(u32),
+    #[error("Backend call failed with status '{0}' and text '{1}'")]
```
mvidner, line 40: NP, status is numeric
```suggestion
    #[error("Backend call failed with status {0} and text '{1}'")]
```

rust/agama-lib/src/error.rs
```diff
@@ -33,6 +35,8 @@ pub enum ServiceError {
     UnsuccessfulAction(String),
     #[error("Unknown installation phase: '{0}")]
```
mvidner, line 38:
```suggestion
    #[error("Unknown installation phase: {0}")]
```
unrelated code but related fix to one below

mvidner (commented):
Here reviewing just the BaseHTTPClient part

rust/agama-lib/src/base_http_client.rs
jreidinger: I check quite deeply when call to text can fail and to be honest I do not find it either in documentation neither from code. So lets use your suggestion.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,133 @@
...
+            .await
+            .map_err(|e| e.into())
+    }
+
+    fn url(&self, path: &str) -> String {
+        self.base_url.clone() + path
+    }
+
+    /// post object to given path and report error if response is not success
+    pub async fn post(&self, path: &str, object: &impl Serialize) -> Result<(), ServiceError> {
```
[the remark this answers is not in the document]
jreidinger: Well, response is optional in POST. I check what we have in `agama-web-server openapi` and it is like half to half if we return something or not...if we need it more often we can create variant that also return object...I am just not sure what will be good name :)

rust/agama-lib/src/base_http_client.rs
jreidinger: Also it is a bit against documentation of traits which mention that it should be obvious to convert...which I am not sure about this case.

rust/agama-lib/src/base_http_client.rs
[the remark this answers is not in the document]
imobachgs: Yes, a POST call does not need to return anything. I am not convinced about using "no text" string. Why not just return an empty body? In any case, we can evolve this API when we use it.

rust/agama-cli/src/questions.rs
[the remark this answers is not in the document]
imobachgs: Yes, I know, we use JSON for profiles and YAML for the answers file. I am not saying that we should fix it now, but let's have into account in the future.

rust/agama-cli/src/questions.rs
```diff
@@ -35,30 +39,54 @@ pub enum Modes {
...
         .await
-        .context("Failed to set answers from answers file")
+        .map_err(|e| e.into())
 }
 
-pub async fn run(subcommand: QuestionsCommands) -> anyhow::Result<()> {
+async fn list_questions() -> Result<(), ServiceError> {
+    let client = HTTPClient::new().await?;
+    let questions = client.list_questions().await?;
+    // FIXME: that conversion to anyhow error is nasty, but we do not expect issue
```
imobachgs, line 59: We could have a generic error for these cases. But let's keep it as it is now. At some point we should improve our error types.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        let client = Client::builder().default_headers(headers).build()?;
+
+        Ok(Self {
+            client,
+            base_url: API_URL.to_string(), // TODO: add support for remote server
+        })
+    }
+
+    /// Simple wrapper around Response to get object from response.
```
imobachgs, line 50:
```suggestion
    /// Simple wrapper around [`Response`] to get object from response.
```

rust/agama-lib/src/base_http_client.rs
imobachgs: OK, async could be a problem. Let's keep it as it is.

imobachgs (approved):
Just minor comments. Feel free to merge.

rust/agama-lib/src/base_http_client.rs
imobachgs: About when to implement those traits, I am usually unsure 😅

PR #1463, opened 2024-07-12
Update agama-products-opensuse.changes

stacheldrahtje:
Update changes as well


products.d/agama-products-opensuse.changes
```diff
@@ -1,3 +1,7 @@
+Tue Jul 12 12:37:15 UTC 2024 - Natasha Ament <stacheldrahtje@gmail.com>
+
+- Change leap_160.yaml to user distribution instead of repositories
```
jreidinger, line 3: btw type user -> use...also I would rephrase it a bit to not confuse users something like `leap_160.yaml - update repositories`

stacheldrahtje: I feel quite stupid but how do I incorporate these changes with retracting the whole thing again?

PR #1475, opened 2024-07-17
feat(live): add a SLE flavor to the live image

imobachgs:
This is an attempt to handle the openSUSE and SLE installation images using a single kiwi definition.

- Rename `agama-installer-openSUSE` to `agama-installer`.
- Add a `SLE` profile.
- Move specific packages to their own profiles.

## Todo

- [x] Inject the proper repositories in the live image (instead of Tumbleweed-based ones).
- [x] Do more testing.
- [x] Adapt the self-update script.


live/src/agama-installer.kiwi
```diff
@@ -134,53 +133,72 @@
...
         <archive name="root.tar.xz"/>
     </packages>
     <!-- additional packages for the openSUSE flavors -->
     <packages type="image" profiles="openSUSE,openSUSE-Playwright">
-      <package name="agama-products-opensuse"/>
-      <package name="openSUSE-build-key"/>
+        <package name="agama-products-opensuse"/>
+        <package name="grub2-branding-openSUSE" arch="aarch64,x86_64"/>
+        <package name="patterns-openSUSE-base"/>
+        <package name="ruby3.3-rubygem-agama-yast"/>
```
lslezak, line 165: This `ruby3.3` vs. `ruby3.2` packages looks ugly, but there is probably no better way... :worried:

lslezak (approved):
LGTM

live/src/agama-installer.kiwi
imobachgs: Yes, I agree. Quoting @jreidinger:

> I think plan for SLE16 is to have suseruby instead of ruby3.3 and it will mean ruby that is supported only for suse products

live/root/usr/bin/agama-self-update
```diff
@@ -6,17 +6,18 @@
...
+zypper lr | grep "^[1-9]" | cut -f1 -d\| | xargs zypper modifyrepo --disable
+zypper modifyrepo --enable agama-devel
 zypper refresh
 zypper --non-interactive dup --details --from agama-devel
 STATUS=$?
 
-# enable OSS back
-zypper modifyrepo --enable repo-oss
+# enable all repositories back
+zypper lr | egrep "^[1-9]" | cut -f1 -d\| | xargs zypper modifyrepo --enable
```
lslezak, line 18: Ah, `zypper` support the `--all` option, you can simplify it to

```suggestion
zypper modifyrepo --enable --all
```

live/root/usr/bin/agama-self-update
```diff
@@ -6,17 +6,18 @@
 # Agama Devel OBS project.
 
 
-# first try a quick and simple solution, refreshing the OSS repository takes a
+# first try a quick and simple solution, refreshing the distributions repository takes a
 # lot of time so try using only the agama-devel for update
-zypper modifyrepo --disable repo-oss
+zypper lr | grep "^[1-9]" | cut -f1 -d\| | xargs zypper modifyrepo --disable
```
lslezak, line 11: Use `-all` option:
```suggestion
zypper modifyrepo --disable --all
```

lslezak (commented):
Use the `--all` option.

live/src/config.sh
```diff
@@ -12,10 +12,21 @@ echo "Configure image: [$kiwi_iname]..."
...
+if [ -f "${REPO}.disabled" ]; then
+  mv "${REPO}.disabled" $REPO
+fi
+rm /etc/zypp/repos.d/*.disabled
+
 # configure the repositories in the Live system
 # import the OBS key for the systemsmanagement OBS project
 rpm --import /tmp/systemsmanagement_key.gpg
 rm /tmp/systemsmanagement_key.gpg
+# import the OBS key for the Devel:YaST:Agama:Head project
```
lslezak, line 27: At first I was confused here, then I realized it is the internal OBS (IBS).  Maybe it should be mentioned in the comment.

lslezak (approved):
LGTM

PR #1476, opened 2024-07-18
feat: add AutoYaST error reporting

jreidinger:
## Problem

Autoyast conversion script can try to open ui for reporting errors or ask questions.

## Solution

Monkey patch it to use agama question mechanism. This change tries to cover the most common cases.
It also needs to implement generic Question with Password react component.
And last but not least during manual testing few issues are revealed and fixed. Also as part of change listing of all question was fixed to really show only unanswered ones and optimize it to use ObjectManager methods.

## Testing
For testing CLI for asking this command is used:

```sh
agama questions ask < data.json
```

For testing autoyast questions it uses encrypted autoyast profile that ask for password and also report issue when password is wrong:

```sh
agama profile autoyast ftp://pong.suse.cz/autoinst_enc.xml
```


lslezak (commented):
:+1:

service/lib/agama/autoyast/report_patching.rb
```diff
@@ -0,0 +1,106 @@
...
+        text:           text,
+        options:        ["ok", "cancel"],
+        default_option: "cancel",
+        data:           {}
+      }
+      data = { generic: question, with_password: {} }.to_yaml
+      answer_yaml = Yast::Execute.locally!("agama", "questions", "ask", stdin: data, stdout: :capture)
+      answer = YAML.load(answer_yaml)
+      result = answer["generic"]["answer"].to_sym
+      return nil if result == "cancel"
```
mvidner, line 102:
```suggestion
      return nil if result == :cancel
```

jreidinger: thanks, good catch. Originally there was not to_sym, but yast popup API expects it and I forgot to update it also here.

web/src/components/questions/QuestionWithPassword.jsx
```diff
@@ -0,0 +1,69 @@
...
+    question.password = password;
+    question.answer = option;
+    answerCallback(question);
+  };
+
+  return (
+    <Popup
+      isOpen
+      title={_("Password Required")}
+      aria-label={_("Question")}
```
dgdavid, line 44: I guess you probably took another question component as an starting point. However [PF/Modal `aria-label` is only required when there is no `title`](https://github.com/patternfly/patternfly-react/blob/7836b87eb0dc557ad6bff78d6053f5b49128b730/packages/react-core/src/components/Modal/Modal.tsx#L191). Let's start dropping not needed props here :)

```suggestion
```

rust/agama-cli/src/questions.rs
```diff
@@ -71,11 +72,14 @@ async fn ask_question() -> Result<(), ServiceError> {
     let question = serde_json::from_reader(std::io::stdin())?;
 
     let created_question = client.create_question(&question).await?;
-    let answer = client.get_answer(created_question.generic.id).await?;
+    let Some(id) = created_question.generic.id else {
+        return Err(ServiceError::QuestionNotExist(0));
```
imobachgs, line 76: What does that '0' mean? It looks like the error is not exactly that the question does not exist.

rust/agama-cli/src/questions.rs
```diff
@@ -19,9 +20,9 @@ pub enum QuestionsCommands {
         /// Path to a file containing the answers in YAML format.
         path: String,
     },
-    /// prints list of questions that is waiting for answer in YAML format
+    /// prints list of questions that is waiting for answer in JSON format
     List,
-    /// Ask question from stdin in YAML format and print answer when it is answered.
+    /// Ask question from stdin in JSON format and print answer when it is answered.
```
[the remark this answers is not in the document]
jreidinger: I hope I can point to openapi json format...not sure if there is something like `agama-web-server openapi components.schemas.question`

rust/agama-cli/src/questions.rs
jreidinger: To be honest this should not happen as it is assigned in that create_question from created question path. I am not sure if we should have something like Programatic error or something that for cases which should not happen and user cannot do naything  with it

service/README.md
```diff
@@ -0,0 +1,21 @@
...
+to agama live ISO. There do this sequence of commands:
+
+```sh
+# ensure that only modified sources are installed
+gem uninstall agama-yast
+# install modified sources including proper binary names
+gem install --no-doc --no-format-executable <path to gem>
+```
+
+If change modifies also dbus parts, then restart related dbus services.
```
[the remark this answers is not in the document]
jreidinger: I hope that systemctl restart agama hopefully should be enough...at least restart of agama-web-service is not enough :)

service/lib/agama/autoyast/report_patching.rb
```diff
@@ -0,0 +1,116 @@
...
+require "yast"
+require "yast2/execute"
+require "ui/dialog"
+
+# :nodoc:
+# rubocop:disable Metrics/ParameterLists
+# rubocop:disable Lint/UnusedMethodArgument
+module Yast2
+  # :nodoc:
+  class Popup
```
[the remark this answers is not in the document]
jreidinger: yeah, I forget to remove it

service/lib/agama/autoyast/report_patching.rb
```diff
@@ -0,0 +1,116 @@
...
+  class PasswordDialog < Dialog
+    def new(label, confirm: false)
+      @label = label
+      # NOTE: implement confirm if needed
+    end
+
+    def run
+      # at first construct agama question to display.
+      text = @label
+      question = {
```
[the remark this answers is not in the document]
jreidinger: agreed.

web/src/components/questions/QuestionWithPassword.test.jsx
```diff
@@ -42,17 +42,10 @@ describe("QuestionWithPassword", () => {
...
-
-  it("contains a textinput for entering the password", async () => {
-    renderQuestion();
-
-    const passwordInput = await screen.findByLabelText("Password");
-    expect(passwordInput).not.toBeNull();
+    screen.queryByText(question.text);
   });
 
   describe("when the user selects one of the options", () => {
```
dgdavid, line 51: NP: Actually, "when the user enters the password"

imobachgs:
Commit fc249f87 introduces a few improvements to the tests:

* If you do not need side-effects (using useEffect or interacting with the UI), you do not need the tests to be `async`.
* If the test is not `async`, you should replace `find*` (which are async) with `query*`.
* Testing that the password input is there is kind of redundant (you are using it in the following test).
* If you do not need the client, prefer `plainRender` (specific to Agama).

web/src/components/questions/QuestionWithPassword.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+    };
+  });
+
+  it("renders the question text", () => {
+    renderQuestion();
+
+    screen.queryByText(question.text);
+  });
+
+  describe("when the user selects one of the options", () => {
```
dgdavid, line 51: NP:

```suggestion
  describe("when the user enters the password", () => {
```

imobachgs (approved):
Please, take into account @dgdavid's comment about the test description and you can merge.

web/src/components/questions/QuestionWithPassword.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+    it("calls the callback after setting both, answer and password", async () => {
```
dgdavid, line 52: NP
```suggestion
    it("calls the callback", async () => {
```

jreidinger:
There is some troubles with PRG cluster affecting also opensuse services. So tests are red due to failed to load testing container. Will merge even with it as previously it was green and change is just some text changes.

PR #1481, opened 2024-07-18
Live ISO: add Puppeteer, drop the Playwright flavor

lslezak:
- Replace Playwright with Puppeteer, include it in the default image (drop the Playwright ISO flavor)


imobachgs (changes requested):
Please, update the changes file.

imobachgs (approved):

PR #1486, opened 2024-07-19
Use "puppeteer-core" instead of full "puppeteer" NPM package

lslezak:
## Problem

- The full "puppeteer" NPM package also takes care of downloading and managing the web browsers.
- But as we use the system browser we do not need this functionality.

## Solution

- Installing just the "puppeteer-core" package makes the NPM bundle a little bit smaller.
- See more details in https://pptr.dev/guides/installation

## Testing

- Tested manually


mvidner (approved):
417 lines saved in `package-lock.json`, yay!

PR #1491, opened 2024-07-23
Updated CI integration test

lslezak:
## Problem

- The old integration test did not work and has been disabled

## Solution

- Switch the test from Playwright to Puppeteer

## Testing

- The [test run](https://github.com/openSUSE/agama/actions/runs/10053959779/job/27787662876) is green now
- If the test fails the screenshot and the HTML page dump is included in the uploaded artifacts archive

### DIFF
--- a/.github/workflows/ci-integration-tests.yml
+++ b/.github/workflows/ci-integration-tests.yml
@@ -29,17 +29,20 @@ jobs:
         fetch-tags: true
 
     - name: Created shared YaST log directory
-      run: mkdir -p /tmp/log/YaST2
+      run: mkdir -p /tmp/log/YaST2 /tmp/log/puppeteer
 
     - name: Start container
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-integration-tests.yml
```diff
@@ -68,29 +71,31 @@ jobs:
...
+            "cp /checkout/puppeteer/tests/test_root_password.js /usr/share/agama/integration-tests/tests &&
+            cp /checkout/puppeteer/agama-integration-tests /usr/bin/agama-integration-tests"
 
-    - name: Run the Playwright tests
-      # user authentication is not required when cockpit runs a local session
-      # run the tests in the Chromium browser
-      run:  podman exec agama bash -c "cd /checkout/playwright && SKIP_LOGIN=true playwright test --trace on --project chromium"
+    - name: Run the Puppeteer tests
+      # run the test
+      run:  podman exec agama bash -c "cd /var/log/puppeteer &&
```
jreidinger, line 82: this cd looks kind of strange. should not agama-integration-tests script ensure that artifacts will be in correct place itself?

lslezak: The script itself saves the log to `./log/` so it's up to the caller where the logs will be saved.

In this case we use `/var/log` as it is already used for the YaST logs, but it could be anywhere else.

jreidinger (approved):

PR #1495, opened 2024-07-24
feat(rust): allow passing the API URL through the CLI

mchf:
## Problem

We want to be able to use different remote installation servers 

## Solution

- Introduced global option for user to provide remote api url
- Used newly introduced BaseHTTPClient to handle urls.

Already done:
- auth, network, localization, users, software, product, storage, questions

## Testing

- *Tested manually*


rust/agama-cli/src/main.rs
```diff
@@ -26,6 +27,15 @@ use std::{
     time::Duration,
 };
 
+/// Agama's CLI global options
+#[derive(Args)]
+struct GlobalOpts {
+    #[clap(long)]
+    /// uri pointing to agama's remote api. If not provided, default https://localhost/api is
```
mvidner, line 34: Currently the option is required:
```console
 # agama config show
error: the following required arguments were not provided:
  --uri <URI>

Usage: agama --uri <URI> <COMMAND>

For more information, try '--help'.
```

mvidner:
I have tried this out, with this testing setup:

- Run agama in a container, with `/testing_using_container.sh`; this builds the Rust code in `rust/target/debug`, and exposes port 443 to 10443 on the host
- (Now I wanted to simply run the agama binary on the host, but this fails for me because 15.4 is too old and misses the new glibc and openssl that the TW container is using)
- Compile agama in release mode on the host, `cargo build -r` which puts the binaries to `rust/target/release`

Now I can try `rust/target/release/agama --api https://localhost:10443/api auth login`

But it does not work yet:

1. there seems to be a funny bug: BaseHTTPClient always wants an authentication token from a file, or else it fails (`NotAuthenticated` is printed). I needed to `mkdir -p $HOME/.local/agama; touch $HOME/.local/agama/token`
2.  the `--api` option seems to be ignored:
```console
$ ./target/release/agama --uri https://ohmu:10443/api auth login
> Please enter the root password: ********
Anyhow(error sending request for url (http://localhost/api/auth): error trying to connect: tcp connect error: Connection refused (os error 111)
```

rust/agama-cli/src/auth.rs
```diff
@@ -1,12 +1,57 @@
...
 use crate::error::CliError;
+use agama_lib::base_http_client::BaseHTTPClient;
 use inquire::Password;
 use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
 use std::io::{self, IsTerminal};
 
-const DEFAULT_AUTH_URL: &str = "http://localhost/api/auth";
+/// HTTP Client for auth queries
+struct AuthHTTPClient {
+    api: BaseHTTPClient,
```
jreidinger, line 12: I think this is wrong as BaseHTTPClient expects already authenticated entry. So auth queries should not set authentication. And also usage below does not use self.api...I expect that this case issue that @mvidner face.

rust/agama-cli/src/main.rs
jreidinger: reason is that it should be just ` pub uri: Option<String>,` so it does not need to be provided.

rust/agama-cli/src/main.rs
```diff
@@ -119,6 +132,8 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
 }
 
 async fn run_command(cli: Cli) -> Result<(), ServiceError> {
+    let client = BaseHTTPClient::new()?;
```
jreidinger, line 135: as mention above, BaseHTTPClient should be used only for commands that already require authenticated access. Or you need to modify it at https://github.com/openSUSE/agama/blob/895b81955c3ed882015ffc7fd4a683a09b4d5762/rust/agama-lib/src/base_http_client.rs#L31 as comment suggest to have constructor that does not set authentication token.
But then you should create that client only in calls where you know that authorization is not required.

rust/agama-cli/src/main.rs
mchf: yes ... during testing i used it as mandatory ... and it is already fixed in one of upcoming commits.

rust/agama-cli/src/auth.rs
mchf: I've introduced new version, which doesn't look for authentication token.

rust/agama-cli/src/main.rs
mchf: already done.  Sadly I pushed incomplete stuff before going to vacation :-/

rust/agama-lib/src/base_http_client.rs
```diff
@@ -28,6 +28,15 @@ pub struct BaseHTTPClient {
 const API_URL: &str = "http://localhost/api";
 
 impl BaseHTTPClient {
+    // Provides new client with no specialities like authorization headers or so.
+    // If needed, there is place for helpers like "self::to_authorized" or so.
+    pub fn bare() -> Result<Self, ServiceError> {
```
mvidner, line 33: Please consider using https://github.com/openSUSE/agama/pull/1517 instead: `BaseHTTPClient::default()` makes the same thing as your `bare()`

mchf: i'll fix it once your pr is merged

mvidner:
With #1438 being merged, this PR could be extended to work with `agama config load`, because this CLI command will ignore the parts of the API that are left unspecified in the JSON, and we can test only on the `user`, `root` and `localization` parts, that already have HTTP clients

mvidner:
Also note that in #1438 I needed to connect to a mock HTTP server and made client APIs for that, which should be pretty useful for your use case too :)

mvidner:
The current code looks good but the `auth` endpoint is special and IMHO we need to see how this works with a regular one too.

rust/agama-cli/src/main.rs
```diff
@@ -139,12 +151,24 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
+        BaseHTTPClient::default()
+    } else {
+        // this deals with authentication need inside
+        BaseHTTPClient::new()?
+    };
+
+    client.base_url = cli
+        .opts
+        .api
+        .strip_suffix("/")
```
mvidner, line 166: This returns `None` if the string has no such suffix, and then I keep wondering why my `--api` gets ignored
https://doc.rust-lang.org/nightly/core/primitive.str.html#method.strip_suffix

https://doc.rust-lang.org/nightly/core/primitive.str.html#method.trim_end_matches should work better

mvidner:
Yay! It gets somewhere... and then it fails on a self-signed certificate:

```console
$ ./rust/target/release/agama --api https://localhost:10443/api/ config show
Anyhow(Failed to communicate with the HTTP backend 'error sending request for url (https://localhost:10443/api/network/connections)'

Caused by:
    0: error sending request for url (https://localhost:10443/api/network/connections)
    1: client error (Connect)
    2: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1890: (self-signed certificate)
    3: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1890:)
```

mchf:
I was having some troubles with dbus in storage but for now at least local calls seem to be working. 
agama_config_show

rust/agama-lib/src/network/client.rs
```diff
@@ -98,32 +68,26 @@ impl NetworkClient {
...
         }
 
         Ok(())
     }
 
     /// Returns an array of network connections
     pub async fn apply(&self) -> Result<(), ServiceError> {
+        // trying to be tricky here. If something breaks then we need a put method on
+        // BaseHTTPClient which doesn't require a serialiable object for the body
+        let empty_body: [String; 0] = [];
```
mvidner, line 86: You may be interested in a shorter way to produce a dummy body:
https://github.com/agama-project/agama/commit/56335baf9e4ad8eeae78372c1a99ebe8696a6ec3

.github/workflows/ci-rust.yml
```diff
@@ -69,6 +69,7 @@ jobs:
 
     - name: Install required packages
       run: zypper --non-interactive install
+        dbus-1-daemon
```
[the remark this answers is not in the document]
mchf: i see ... merge / rebase didn't

rust/agama-cli/src/main.rs
```diff
@@ -138,13 +152,66 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
     Ok(ManagerClient::new(conn).await?)
 }
 
+#[derive(PartialEq)]
+enum InsecureApi {
+    Secure,         // Remote api is secure
+    Insecure,       // Remote api is insecure - e.g. self-signed certificate
+    Forbidden,      // Remote api is insecure and its use is forbidden (e.g. user decided not to use it)
+    Unrecheable     // Remote api is unrecheable
```
mvidner, line 160:
```suggestion
    Unreachable     // Remote api is unreachable
```

rust/agama-cli/src/main.rs
```diff
@@ -138,13 +152,66 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
+    let mut ping_client = BaseHTTPClient::default();
+    ping_client.base_url = api_url;
+
+    // decide whether access to remote site has to be insecure (self-signed certificate or not)
+    match ping_client.get::<HashMap<String, String>>("/ping").await {
+        Ok(res) => {
+            if res["status"] == "success" {
+                Ok(InsecureApi::Secure)
+            } else {
+                Ok(InsecureApi::Unrecheable)
```
mvidner, line 175:
```suggestion
                Ok(InsecureApi::Unreachable)
```

rust/agama-cli/src/lib.rs
```diff
@@ -138,13 +153,63 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
+    let mut ping_client = BaseHTTPClient::default();
+    ping_client.base_url = api_url;
+
+    // decide whether access to remote site has to be insecure (self-signed certificate or not)
+    match ping_client.get::<HashMap<String, String>>("/ping").await {
+        Ok(res) => {
+            if res["status"] == "success" {
+                Ok(InsecureApi::Secure)
+            } else {
+                Ok(InsecureApi::Unreachable)
```
mvidner, line 176: Now that the enum is spelled correctly, I realize that it is unused. The caller does not care if it is `Secure` or `Unreachable`, so we might as well `Ok(_) => { Ok(InsecureApi::Secure }`, no?

rust/agama-cli/src/auth.rs
```diff
@@ -18,15 +18,44 @@
...
 
 use crate::error::CliError;
+use agama_lib::base_http_client::BaseHTTPClient;
 use inquire::Password;
-use reqwest::header::{HeaderMap, HeaderValue, CONTENT_TYPE};
+use std::collections::HashMap;
 use std::io::{self, IsTerminal};
 
-const DEFAULT_AUTH_URL: &str = "http://localhost/api/auth";
+/// HTTP Client for auth queries
```
imobachgs, line 30:
```suggestion
/// HTTP client to handle authentication
```

Otherwise, it looks to me like it is for performing authenticated queries (well, perhaps it is just a personal impression).

rust/agama-cli/src/lib.rs
```diff
@@ -138,13 +153,63 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
+#[derive(PartialEq)]
+enum InsecureApi {
+    Secure,      // Remote api is secure
+    Insecure,    // Remote api is insecure - e.g. self-signed certificate
+    Forbidden, // Remote api is insecure and its use is forbidden (e.g. user decided not to use it)
+    Unreachable, // Remote api is unrecheable
+}
+
+/// Returns if insecure connection to remote api server is required and user allowed that
+async fn check_remote_api(api_url: String) -> Result<InsecureApi, ServiceError> {
```
imobachgs, line 165: I think we should fine a better home for this function and the accompanying `InsecureApi`.

imobachgs (approved):
First of all, thanks for taking care of this feature and having the patience to deal with multiple rebases due to changes in the codebase.

Now, I must admit that I am not convinced about how the `BaseHTTPClient` constructors look like. Let's put apart the fact that it lacks better naming and documentation. Passing that `insecure` argument here and there looks suspicious to me and it does not look that extensible.

Additionally, building a client implies performing a `ping`, so I think you could put all together and reach for the [builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html). I have not tried the following code (I am sure that the Rust compiler would be angry with me), but I think it illustrates the idea.

```rust
  struct BaseHTTPClientBuilder {
    url: String,
    insecure: boolean
  }

  impl BaseHTTPClientBuilder {
    pub fn new(url: &str) -> Self {
      Self {
        url: url.to_string(),
        insecure: false
      }
    }
    pub fn insecure(mut self) {
      self.insecure = true;
      self
    }

    pub async fn build() -> Result<Client, ClientError> {
      // You can do the "ping" here. If everything works,
      // it returns the client. Otherwise, it returns an error.
      todo!()
    }
  }
```

It offers an extensible API and the logic to build the client does not live in the generic `lib.rs`. Actually, there is no need for an `InsecureApi` enum.


```rust
let client = HTTPClientBuilder::new(url) # or you can use an `with_url` method.
  .build()
  .await;

// Now you can match on the result. If something went wrong, you can ask the user and build the client again with something like:

let client = HTTPClientBuilder::new(url)
  .insecure
  .build()
  .await;

```

About the overall approach, I would like to have an `--insecure` flag (or something similar) to skip that kind of checks if wanted (like `zypper` or `curl` do).

Last but not least, the [questions::HTTPClient](https://github.com/agama-project/agama/blob/api-option/rust/agama-lib/src/questions/http_client.rs#L35) is not adapted.

Thanks!

imobachgs (changes requested):
Sorry, I wrongly approved the PR 🤦🏻. Please, check the previous review.

rust/agama-cli/src/auth.rs
```diff
@@ -18,15 +18,44 @@
...
+    api: BaseHTTPClient,
+}
+
+impl AuthHTTPClient {
+    pub fn load(client: BaseHTTPClient) -> Result<Self, ServiceError> {
+        Ok(Self { api: client })
+    }
+
+    /// Query web server for JWT
+    pub async fn get_jwt(&self, password: String) -> anyhow::Result<String> {
```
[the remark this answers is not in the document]
mchf: well this is a piece of old code which only changed place ... but will try to find out better name. Another note is that get in the name has nothing to do with http's get

rust/agama-lib/src/base_http_client.rs
```diff
@@ -61,15 +62,30 @@ impl Default for BaseHTTPClient {
...
+        Self {
+            client: reqwest::Client::builder()
+                .danger_accept_invalid_certs(insecure)
+                .build()
+                .unwrap_or(default_client),
+            base_url: API_URL.to_owned(),
+        }
+    }
+    /// Uses `localhost`, authenticates with [`AuthToken`].
+    pub fn new_with_params(insecure: bool) -> Result<Self, ServiceError> {
```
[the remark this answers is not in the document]
mchf: parameters ;-)

rust/agama-lib/src/base_http_client.rs
```diff
@@ -61,15 +62,30 @@ impl Default for BaseHTTPClient {
...
+    }
+
+    pub fn bare(insecure: bool) -> Self {
+        let default_client = reqwest::Client::new();
+
+        Self {
+            client: reqwest::Client::builder()
+                .danger_accept_invalid_certs(insecure)
+                .build()
+                .unwrap_or(default_client),
```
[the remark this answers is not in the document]
mchf: According to the documentation:
This method fails if a TLS backend cannot be initialized, or the resolver cannot load the system configuration.

rust/agama-lib/src/network/client.rs
```diff
@@ -19,72 +19,42 @@
...
-        self.text_for(response).await
-    }
-
     /// Returns an array of network devices
     pub async fn devices(&self) -> Result<Vec<Device>, ServiceError> {
-        let text = self.get("/devices").await?;
-
-        let json: Vec<Device> = serde_json::from_str(&text)
-            .map_err(|e| ServiceError::NetworkClientError(e.to_string()))?;
+        let json = self.client.get::<Vec<Device>>("/network/devices").await?;
```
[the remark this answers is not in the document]
mchf: your pleasure, my happiness ;-)

rust/agama-lib/src/base_http_client.rs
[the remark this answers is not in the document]
imobachgs: I know it means `parameters` :-) I am saying that it says little about the intention (ok, having parameters :-/). But I commented more about the whole API int the review.

rust/agama-cli/src/lib.rs
```diff
@@ -138,13 +153,63 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
+
 pub async fn run_command(cli: Cli) -> Result<(), ServiceError> {
+    // somehow check whether we need to ask user for self-signed certificate acceptance
+    let api_url = cli.opts.api.trim_end_matches('/').to_string();
+    let insecure = check_remote_api(api_url.clone()).await? == InsecureApi::Insecure;
+
+    // we need to distinguish commands on those which assume that authentication JWT is already
+    // available and those which not (or don't need it)
+    let mut client = if let Commands::Auth(_) = cli.command {
+        BaseHTTPClient::bare(insecure)
```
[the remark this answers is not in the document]
mchf: > Well, since the client has the _insecure_ modification, I'd suggest renaming to `unauthenticated`

good point. thanks

rust/agama-cli/src/lib.rs
mchf: refactored a bit

rust/agama-cli/src/lib.rs
```diff
@@ -138,13 +157,50 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
 
+/// True if use of the remote API is allowed (yes by default when the API is secure, the user is
+/// asked if the API is insecure - e.g. when it uses self-signed certificate)
+async fn allowed_insecure_api(use_insecure: bool, api_url: String) -> Result<bool, ServiceError> {
+    // fake client used for remote site detection
+    let mut ping_client = BaseHTTPClient::default();
+    ping_client.base_url = api_url;
+
+    // decide whether access to remote site has to be insecure (self-signed certificate or not)
+    return match ping_client.get::<HashMap<String, String>>("/ping").await {
```
imobachgs, line 168: Please, remove the `return` keyword.

rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Oct 16 07:55:27 UTC 2024 - Michal Filka <mfilka@suse.com>
+
+- Implemented option for providing remote API address for the CLI
```
imobachgs, line 4: Please, add the `gh#agama-project/agama#1495` reference.

imobachgs (approved):

PR #1497, opened 2024-07-24
feat(puppeteer): Cleanup node_modules

lslezak:
## Problem

- The Puppeteer `node_modules` directory is huge

## Solution

- Delete the not needed content from `node_modules`
- Inspired by [node-prune](https://github.com/tj/node-prune/) tool (written in Go, not much suitable to run during RPM build, I rewrote that to shell)

## Testing

- Tested manually

## Numbers

- The `node_modules` directory size went down from 39MB to 15MB (-62%)
- The built RPM size went down from 3.5MB to 1.9MB (-46%)
- I expect the Live ISO size reduction similar to the RPM reduction

### DIFF
--- a/puppeteer/node-prune.sh
+++ b/puppeteer/node-prune.sh
@@ -58,16 +58,17 @@ FILES=(
   .coveralls.yml
   CHANGES
   changelog
-  LICENSE.txt
-  LICENSE
-  LICENSE-MIT
-  LICENSE.BSD
-  license
-  LICENCE.txt
-  LICENCE
-  LICENCE-MIT
-  LICENCE.BSD
-  licence
+  # keep the package licenses, it's unclear if we can legally delete them...
+  # LICENSE.txt
+  # LICENSE
+  # LICENSE-MIT
+  # LICENSE.BSD
+  # license
+  # LICENCE.txt
+  # LICENCE
+  # LICENCE-MIT
+  # LICENCE.BSD
+  # licence
   AUTHORS
   CONTRIBUTORS
   .yarn-integrity
--- a/puppeteer/node-prune.sh
+++ b/puppeteer/node-prune.sh
@@ -1,0 +1,158 @@
+#! /bin/bash
+
+# This script clean up the node_modules directory which is usually huge and
+# contains a lot of not needed files. It was inspired by the node-prune tool
+# (https://github.com/tj/node-prune).
+#
+# Usage:
+#
+#     node-prune.sh [path]
+#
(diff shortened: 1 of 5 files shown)


puppeteer/node-puppeteer-prune.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/sh
+
+# This is a helper script which deletes some not needed files from NPM packages.
+# This script is specific for Puppeteer installations.
+
+MODULES_PATH="${1:-./node_modules}"
+
+# delete Puppeteer CommonJS modules, we use the ES modules (in lib/esm)
+rm -rf "$MODULES_PATH/puppeteer-core/lib/cjs"
```
jreidinger, line 9: I kind of expect that it will run internally that node-prune.sh..so just one command will be enough to cleanup pupeteer modules.

puppeteer/package/agama-integration-tests.spec
```diff
@@ -47,6 +47,12 @@ outside.
 rm -f package-lock.json
 local-npm-registry %{_sourcedir} install --omit=optional --with=dev --legacy-peer-deps || ( find ~/.npm/_logs -name '*-debug.log' -print0 | xargs -0 cat; false)
 
+# node_modules cleanup
+%{_builddir}/agama/node-prune.sh
```
jreidinger, line 51: should we run it also for agama web directory?

lslezak: Agama-web uses webpack which already bundles only the needed files into the target JS file. It would not help there.

puppeteer/node-puppeteer-prune.sh
lslezak: Both commands are started in the `agama-integration-test` scripts which installs the NPM packages and runs the tests. Normally you should not need to run them separately.

puppeteer/node-prune.sh
```diff
@@ -0,0 +1,158 @@
...
+  LICENSE.txt
+  LICENSE
+  LICENSE-MIT
+  LICENSE.BSD
+  license
+  LICENCE.txt
+  LICENCE
+  LICENCE-MIT
+  LICENCE.BSD
+  licence
```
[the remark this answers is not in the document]
lslezak: Oh, good point. As we already run the deduplication in RPM we could keep them. I'll check how much the size changes...

puppeteer/package/agama-integration-tests.spec
lslezak: And we do obfuscation and minification as well...

PR #1502, opened 2024-07-25
fix(ci): avoid accidetally adding NPM TGZ source files to OBS

lslezak:
## Problem

- Sometimes a TGZ file is accidentally added to the OBS package sources, [see an example](https://build.opensuse.org/package/show/systemsmanagement:Agama:Devel/agama-integration-tests?rev=4)
- I guess that the `osc service` for some reason does not delete all downloaded NPM packages and leaves them there
- Then they are submitted with the usual package sources

## Solution

- To avoid this problem always delete all `*.tgz` files before submitting the package to OBS

### DIFF
--- a/.github/workflows/obs-staging-shared.yml
+++ b/.github/workflows/obs-staging-shared.yml
@@ -95,7 +95,7 @@ jobs:
         # sometimes the "osc service" run does not cleanup properly all
         # downloaded NPM package tarballs and they are accidentally added to the
         # OBS package, so delete any TGZ files present
-        run: rm -f *.tgz
+        run: rm -vf *.tgz
         working-directory: ./${{ vars.OBS_PROJECT }}/${{ inputs.package_name }}
 
       - name: Check status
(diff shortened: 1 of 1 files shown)


.github/workflows/obs-staging-shared.yml
```diff
@@ -91,6 +91,13 @@ jobs:
         run: osc service manualrun
         working-directory: ./${{ vars.OBS_PROJECT }}/${{ inputs.package_name }}
 
+      - name: Cleanup
+        # sometimes the "osc service" run does not cleanup properly all
+        # downloaded NPM package tarballs and they are accidentally added to the
+        # OBS package, so delete any TGZ files present
+        run: rm -f *.tgz
```
jreidinger, line 98: how it will look in ci output? will it print all deleted files? maybe adding `rm -vf` would help to see what is removed to verify if something unexpected happen.

jreidinger (approved):

PR #1503, opened 2024-07-26
fix(web): allow reloading the page during installation

imobachgs:
The new `useProduct` hook uses the Suspense API to display a "loading" message while retrieving the
list of products. Unfortunately, that's a problem during installation: the software service won't
response with the list of products and the UI will get stuck.

This PR fixes the problem by adding a `suspense` option `useProduct` so you can decide when to use
the Suspense API.

### DIFF
--- a/web/src/queries/software.ts
+++ b/web/src/queries/software.ts
@@ -119,17 +119,17 @@ const useConfigMutation = () => {
  */
 const useProduct = (
   options?: QueryHookOptions,
-): { products: Product[]; selectedProduct: Product | undefined } => {
+): { products?: Product[]; selectedProduct?: Product } => {
   const func = options?.suspense ? useSuspenseQueries : useQueries;
-  const [{ data: selected }, { data: products }] = func({
+  const [
+    { data: selected, isPending: isSelectedPending },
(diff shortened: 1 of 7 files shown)


web/src/queries/software.ts
```diff
@@ -113,14 +114,28 @@ const useConfigMutation = () => {
   return useMutation(query);
 };
 
+type QueryHookOptions = {
+  suspense: boolean;
+};
+
```
dgdavid, line 120: This is already declared in _types/queries.ts_. I guess it is a leftover and would be sufficient importing it.

web/src/types/queries.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+/**
+ * Typical options for our queries hooks.
+ */
+type QueryHookOptions = {
+  suspense: boolean;
```
dgdavid, line 26: Should `suspense` be optional?

dgdavid (approved):
Looks good to me, just two minor NP comments

PR #1504, opened 2024-07-26
refactor(web): use queries to track progress

imobachgs:
Follow-up of #1439 the goal of this PR is to change the progress track to use Tanstack Query.

### DIFF
--- a/web/src/queries/progress.ts
+++ b/web/src/queries/progress.ts
@@ -108,7 +108,7 @@ const useResetProgress = () => {
     return () => {
       queryClient.invalidateQueries({ queryKey: ["progress"] });
     };
-  }, []);
+  }, [queryClient]);
 };
 
 export { useProgress, useProgressChanges, useResetProgress, progressQuery };
--- a/web/src/queries/progress.ts
+++ b/web/src/queries/progress.ts
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) [2024] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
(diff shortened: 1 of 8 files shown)


web/src/queries/progress.ts
```diff
@@ -0,0 +1,91 @@
...
+    queryKey: ["progress", service],
+    queryFn: () =>
+      fetch(`/api/${service}/progress`)
+        .then((res) => res.json())
+        .then((body) => Progress.fromApi(body)),
+  };
+};
+
+type UseProgressOptions = {
+  suspense: boolean;
```
dgdavid, line 44: NP: should `suspense` be optional?

web/src/queries/progress.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+const useProgress = (service: string, options?: QueryHookOptions): Progress => {
```
dgdavid, line 47: NP: `QueryHookOptions`? Do you mean `useProgressOptions`?

dgdavid (approved):
LGTM, just minor comments

web/src/queries/progress.ts
imobachgs: No, I meant `QueryHookOptions`... `UseProgressOptions` is a left-over. Good catch!

web/src/queries/progress.ts
imobachgs: It is a left-over. I will remove it.

dgdavid (approved):
Thanks for documenting the queries. It looks good, you can merge it.

PR #1510, opened 2024-07-29
fix(live): One more PXE build fix

lslezak:
## Problem

The PXE build now fails with this error:
```
[  125s] + rpm --import '/usr/lib/rpm/gnupg/keys/*.asc'
[  125s] error: /usr/lib/rpm/gnupg/keys/*.asc: import read failed(2).
```

## Solution

- Install the same packages as for default Live ISO
- The fix is similar to #1505


teclator (approved):
LGTM

PR #1514, opened 2024-07-30
refactor(web): use queries to track the installer status

imobachgs:
Follow-up of #1439. The goal of this PR is to change installer status tracking to Tanstack Query. Additionally, it adjust the HTTP API a bit.

## Notes

The `WithStatus` mixin should not be relevant anymore. However, the storage UI still uses it to detect when the proposal changes. We should use a dedicated event instead.


web/src/App.jsx
```diff
@@ -28,9 +28,9 @@ import { useInstallerL10n } from "./context/installerL10n";
 import { useInstallerClientStatus } from "~/context/installer";
 import { useProduct, useProductChanges } from "./queries/software";
 import { CONFIG, INSTALL, STARTUP } from "~/client/phase";
-import { BUSY } from "~/client/status";
```
dgdavid, line 31: As discussed offline, please check if _client/status_ still needed.

web/src/App.jsx
```diff
@@ -64,15 +66,15 @@ function App() {
         </SimpleLayout>
       );
 
-    if ((phase === STARTUP && status === BUSY) || phase === undefined || status === undefined) {
```
dgdavid, line 67: I understand that `phase === undefined` is either, not possible anymore or nor relevant here. Could you please confirm it?

web/src/components/core/Installation.jsx
```diff
@@ -21,10 +21,9 @@
 
 import React from "react";
 import { InstallationProgress, InstallationFinished } from "~/components/core";
-import { IDLE } from "~/client/status";
 
-function Installation({ status }) {
-  return status === IDLE ? <InstallationFinished /> : <InstallationProgress />;
+function Installation({ isBusy }) {
+  return isBusy ? <InstallationFinished /> : <InstallationProgress />;
```
dgdavid, line 26: NP: out of the scope of the PR, but you might try to move small components and their tests to TypeScript.

web/src/components/overview/SoftwareSection.test.tsx
```diff
@@ -32,7 +32,7 @@ let mockTestingProposal: SoftwareProposal;
 jest.mock("~/queries/software", () => ({
   usePatterns: () => mockTestingPatterns,
   useProposal: () => mockTestingProposal,
-  useProposalChanges: jest.fn(),
+  useProposalChanges: () => jest.fn(),
```
dgdavid, line 35: good catch! :+1:

web/src/components/product/ProductSelectionProgress.jsx
```diff
@@ -35,15 +36,9 @@ import { PATHS } from "~/router";
...
-  const [status, setStatus] = useState();
+  const { isBusy } = useInstallerStatus({ suspense: true });
 
-  useEffect(() => {
-    manager.getStatus().then(setStatus);
-    return manager.onStatusChange(setStatus);
-  }, [manager, setStatus]);
-
-  if (status === IDLE) return <Navigate to={PATHS.root} replace />;
+  if (!isBusy) return <Navigate to={PATHS.root} replace />;
```
dgdavid, line 41: That was a good movement. I like to see how components are getting cleaner little by little.

web/src/types/status.ts
```diff
@@ -0,0 +1,45 @@
...
+type InstallerStatus = {
+  /** Whether the installer is busy */
+  isBusy: boolean;
+  /** Installation phase */
+  phase: InstallationPhase;
+  /** Whether the installation can be performed or not */
+  canInstall: boolean;
+  /** Whether the installer is running on Iguana */
+  useIguana: boolean;
+};
```
dgdavid, line 43: :ok_hand:

dgdavid (approved):
It looks really good, thanks!

web/src/App.jsx
imobachgs: Yes, that's it. We are using `suspense: true` to get the installer status.

web/src/App.jsx
imobachgs: It is still needed by the storage part. But `client/phase` is not, so I have removed it.

web/src/components/core/Installation.jsx
imobachgs: I would prefer to do that on a separate PR.

PR #1515, opened 2024-07-30
feat(live): Improve printing the Agama URL in the console

lslezak:
## Improvements

- Hide Avahi/mDNS URL if network is not configured (displaying `https://agama.local` URL is confusing if the network is not available, in that case the URL simply cannot work)
- Display only configured interfaces (original code printed empty `https://` URL for not configured devices, like not configured/not connected Wifi adapters)
- Refresh the URLs when a device is changed (it also detects cable connecting/disconnecting)
- Display a warning if the network is not available. Originally I wanted to display some hint like "Press Alt+F7 to switch to the local installer" but that does not work on some architectures (S390) or in some situations (serial console). So to avoid confusion I have dropped that feature.

## Notes

- Use the NetworkManager backend instead of udev. Udev works on the hardware level and does not detect network configuration changes (configuring Wifi) or changing the connection (unplugging/plugging in the network cable). The NetworkManager provides more details.

## Testing

- Tested manually

## Screenshots

- Video recording of a session where the virtual network cable in a VirtualBox VM was disconnected and then reconnected again

[Agama-screen0.webm](https://github.com/user-attachments/assets/44ea27b6-dede-4ec7-8231-d6de73815cbf)

### DIFF
--- a/live/root/etc/systemd/system/agama-url-issue.service
+++ b/live/root/etc/systemd/system/agama-url-issue.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Generate issue file for Agama URLs
+
+After=network-online.target
+
+[Service]
+ExecStart=agama-issue-generator --watch-network
+Type=simple
+
+[Install]
+WantedBy=default.target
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM, :ok_hand:

PR #1516, opened 2024-07-30
refactor(web): use queries for dealing with questions

dgdavid:
In the context of #1439, replace Questions client by its equivalent using TanStack Query.

Additionally, it includes two commits with changes out of the scope of the PBI we're working on. Namely:

  * Migrates the _components/questions_ files to TypeScript, https://github.com/openSUSE/agama/pull/1516/commits/560f953e9fac707688718fbddf8bf3e9dc249be5
  * Stop re-exporting everything in _components/questions/index.ts_, https://github.com/openSUSE/agama/pull/1516/commits/7c7b049b9ba2f2b5a25817e1f350506ad03ae207

Tested manually with a LuksActivationQuestion

Screenshot from 2024-08-02 13-45-36

---

Related to https://trello.com/c/8u1WOJz4 (internal link)


web/src/components/questions/Questions.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import GenericQuestion from "~/components/questions/GenericQuestion";
+import QuestionWithPassword from "~/components/questions/QuestionWithPassword";
+import LuksActivationQuestion from "~/components/questions/LuksActivationQuestion";
```
dgdavid, line 25: We've agreed on not exporting all components in index file ( see https://trello.com/c/c0GEUMQ8 (internal link) and https://github.com/openSUSE/agama/issues/1049). And this is a little step towards such an agreement. However, looking at it, two questions comes to my mind:

* Should we use `~/components/<namespace>/<Component>` or `./<Component>`. I see the first approach kind of _safe_ and _stable_, since it does not matter from where you are importing the component.
* If we use ``~/components/<namespace>/<Component>` does make sense to keep exporting _public_ components in the index file. For example, for importing `Questions.tsx` we could

  ```js
  import Questions from "~/components/questions/Questions";
  ```
  or 

  ```js
   import { Questions } from "~/components/questions";

Honestly, for namespaces exporting a single component there is not too much difference / advantages. Another story is when importing multiple components from a single namespace, like _core_. Similar to what happens when importing components from PatternFly: it would be really annoying if we have had to import each component from its path.

Just sharing thoughts. No strong opinion here (yet :stuck_out_tongue_closed_eyes:)
  ```

dgdavid, line 25: This three import also reminds me the article I shared on slack channel two weeks ago about naming: https://kyleshevlin.com/prefer-noun-adjective-naming

It isn't exactly the same case than the article, but maybe going for `QuestionGeneric`, `QuestionWithPassword`, and `QuestionLuksActivation` would make them more consistent from a naming POV. Again, no strong opinion.

dgdavid: Or maybe we could just drop the `Question` prefix/suffix since they are already in the `questions` namespace.

web/src/components/questions/LuksActivationQuestion.tsx
```diff
@@ -20,27 +20,41 @@
...
-import { QuestionActions } from "~/components/questions";
+import QuestionActions from "~/components/questions/QuestionActions";
 import { _ } from "~/i18n";
 
-const renderAlert = (attempt) => {
-  if (!attempt || attempt === 1) return null;
+/**
+ * Internal component for rendering an alert if given password failed
+ */
+const Alert = ({ attempt }: { attempt: string | undefined }): React.ReactNode => {
```
imobachgs, line 32: Should we use `string | undefined` or make the property optional?

web/src/components/questions/LuksActivationQuestion.tsx
```diff
@@ -20,27 +20,41 @@
...
+  if (!attempt || parseInt(attempt) === 1) return null;
 
   return (
     // TRANSLATORS: error message, user entered a wrong password
-    <Alert variant="warning" isInline isPlain title={_("Given encryption password didn't work")} />
+    <PFAlert
+      variant="warning"
+      isInline
+      isPlain
+      title={_("Given encryption password didn't work")}
```
imobachgs, line 41:
```suggestion
      title={_("The encryption password did not work")}
```

web/src/components/questions/QuestionActions.tsx
```diff
@@ -42,13 +39,23 @@ const label = (text) => `${text[0].toUpperCase()}${text.slice(1)}`;
  * React.Fragment (aka <>) here for wrapping the actions instead of directly using the Popup.Actions.
  *
  * @param {object} props - component props
- * @param {Array.<String>} props.actions - the actions to be shown
- * @param {String} [props.defaultAction] - the action to be shown as primary
- * @param {function} props.actionCallback - the function to be called when user clicks on action
- * @param {Object} [props.conditions={}] - an object holding conditions, like when an action is disabled
+ * @param props.actions - the actions to be shown
```
imobachgs, line 42:
```suggestion
 * @param props.actions - the actions show
```

web/src/components/questions/QuestionActions.tsx
```diff
@@ -42,13 +39,23 @@ const label = (text) => `${text[0].toUpperCase()}${text.slice(1)}`;
...
+ * @param props.defaultAction - the action to be shown as primary
```
imobachgs, line 43:
```suggestion
 * @param props.defaultAction - the action to show as primary
```

web/src/components/questions/QuestionActions.tsx
```diff
@@ -42,13 +39,23 @@ const label = (text) => `${text[0].toUpperCase()}${text.slice(1)}`;
...
+ * @param props.actionCallback - the function to be called when user clicks on action
```
imobachgs, line 44:
```suggestion
 * @param props.actionCallback - the function to call when the user clicks on the action
```

web/src/components/questions/Questions.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      expect(container).toBeEmptyDOMElement();
+    });
+  });
+
+  describe("when a question is answered", () => {
+    beforeEach(() => {
+      mockQuestions = [genericQuestion];
+    });
+
+    it("triggers the useQuestionMutationk", async () => {
```
imobachgs, line 77:
```suggestion
    it("triggers the useQuestionMutation", async () => {
```

imobachgs (commented):
It looks almost good. Thanks!

web/src/components/questions/LuksActivationQuestion.tsx
dgdavid: Good point. I guess making it optional would be better.

imobachgs (approved):

imobachgs (approved):

PR #1517, opened 2024-07-31
BaseHTTPClient: better constructor API, Default

mvidner:
to enable
- connecting to alternative servers, either production or test mocks
- unauthenticated calls, either for the initial auth or test mocks

This is needed for both #1438 and #1495

In the end, the API changes only by adding a `Default` trait implementation which does no authentication.
Connecting to a mock server is achieved by assigning to the public `base_url` field.

### DIFF
--- a/rust/agama-lib/src/base_http_client.rs
+++ b/rust/agama-lib/src/base_http_client.rs
@@ -1,4 +1,4 @@
-use reqwest::{header, Client, Response};
+use reqwest::{header, Response};
 use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{auth::AuthToken, error::ServiceError};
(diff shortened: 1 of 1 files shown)


rust/agama-lib/src/base_http_client.rs
```diff
@@ -21,15 +21,36 @@ use crate::{auth::AuthToken, error::ServiceError};
 ///   }
 /// ```
 pub struct BaseHTTPClient {
-    client: Client,
+    pub client: reqwest::Client,
```
imobachgs, line 24: Why is this field public? IMHO, `BaseHTTPClient` should offer the API that fulfills our needs, and the `reqwest::Client` should be just an implementation detail. If we want to replace [reqwest](https://docs.rs/reqwest/latest/reqwest/) with something else (e.g., [ureq](https://github.com/algesten/ureq)) we can do it easily (or if reqwest API changes it might be easy to adapt).

mvidner: It enables me [constructing  a mock client like this](https://github.com/openSUSE/agama/pull/1438/files#diff-e2b8db39f9a74a79051108962095c52672b5b2ff59630d24320b96780d9288b6R98-R101
):
```rs
        let mock_client =  BaseHTTPClient {
            base_url: mock_url,
            ..Default::default()
        };
```

I chose this way because I did not want to come up with a special name for it.

Alternatively we can make it private again, remove the Default trait and make a set of specialized constructors, like `new_unauthenticated_with_url` here?

mchf: I was playing with an idea to implement some traits here and than do necessary changes in behavior in uderlying structs (which can even cover differences in using authorization token or not)

imobachgs (approved):
LGTM. As @mvidner said, this `BaseHTTPClient` is pretty volatile and we need to experiment a bit more. But by now, it looks good enough.

rust/agama-lib/src/base_http_client.rs
```diff
@@ -43,12 +43,19 @@ impl BaseHTTPClient {
...
     pub fn new() -> Result<Self, ServiceError> {
         Ok(Self {
-            client: Self::authenticated_reqwest_client()?,
+            client: Self::authenticated_client()?,
             ..Default::default()
         })
     }
 
-    fn authenticated_reqwest_client() -> Result<reqwest::Client, ServiceError> {
+    pub fn new_unauthenticated_with_url(url: String) -> Result<Self, ServiceError> {
```
mchf, line 51: ```base_url``` is public, so no need to initialize that via constructor. Moreover it adds some overhead when you want unauthorized client but with a default API url (which for now might be the most expected case in my POV)

mvidner: You're right, it is not needed

PR #1518, opened 2024-07-31
Update nokogiri dependency to version 1.16

lslezak:
## Problem

The Live ISO does not build because of this problem:
```
nothing provides rubygem(ruby:3.3.0:nokogiri:1.15) >= 1.15.4 needed by ruby3.3-rubygem-agama-yast
```

## Solution

- The [nokogiri gem](https://build.opensuse.org/package/show/openSUSE:Factory/rubygem-nokogiri) has been updated to 1.16.6

## Testing

- Tested manually, I have updated the nokogiri gem manually in the Live ISO and Agama works fine with that


mvidner (approved):
LGTM, thank you!

PR #1519, opened 2024-08-01
refactor(web): replace NetworkClient with queries

teclator:
Adapt the network management code to use queries instead of NetworkCient. Related to https://github.com/openSUSE/agama/pull/1439.

Additionally, starts migrating the code to TypeScript and fixes bugs found by the type system.

---

Reviewers, please note that this ended up being a rather long PR, reason why is better to not only review the code but also to do manual testing if possible. Take your time, but any issue out of the scope of state management / queries migration should be reported as a new issue to be addressed in a new PBI.

---

Notes for creating follow-up work, 

* Fix WifiConnectionForm to avoid `"error": "Network system error: Network state error: Connection '<ssid>' already exists"` when connecting to hidden network that previously failed. In short, we should either, to drop the configuration created for a failed { hidden: true } connection or to implement a mechanism to know that such a configuration already exists and performs an _update_ instead of _add_.
* Review when it's possible to define a gateway and fix the UI information accordingly
* Check if we can use the type Address from ipaddr package instead of our own IPAddress type (most probably not)
* Add support for IPv6 since as @jreidinger [said](https://github.com/openSUSE/agama/pull/1519#discussion_r1714327402) its usage is becoming more common. Look for `FIXME: IPv6 is not supported yet.` comments


web/src/utils/network.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+
+/**
+ * @param flags - AP flags
+ * @param wpa_flags - AP WPA1 flags
+ * @param rsn_flags - AP WPA2 flags
+ * @return supported security protocols
+ */
+const securityFromFlags = (flags: number, wpa_flags: number, rsn_flags: number): SecurityProtocols[] => {
```
dgdavid, line 153: @teclator If I remember correctly, you told me that everything should be camelCase. Should we adapt this?

web/src/utils/network.ts
```diff
@@ -0,0 +1,209 @@
...
+  ApFlags,
+  ApSecurityFlags,
+  Connection,
+  ConnectionState,
+  Device,
+  IPAddress,
+  Route,
+  SecurityProtocols,
+} from "~/types/network";
+
```
dgdavid, line 34: I have a doubt with types in general that it is quite visible in the return types of functions in this file. Should we specify them always even when the typing system is able to infer it?

web/src/api/http.ts
```diff
@@ -55,7 +55,7 @@ const put = (url: string, data: object) => http.put(url, data);
  * @param url - endpoint URL
  * @param data - request payload
  */
-const post = (url: string, data: object) => http.post(url, data);
+const post = (url: string, data?: object) => http.post(url, data);
```
jreidinger, line 58: hmm, for both changes, does it make sense to have patch and post that does not provide data? It looks like bug for me.

web/src/api/network.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Apply network changes
+ */
+const applyChanges = () => post("/api/network/system/apply");
+
+/**
+ * Performs the connect action for connection matching given name
+ */
+const connect = (name: string) => patch(`/api/network/connections/${name}/connect`);
```
jreidinger, line 79: to be honest, I do not get here why it is patch..for patch it should contain patch structure that changes connection object. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH mentions that `A PATCH request is considered a set of instructions on how to modify a resource. Contrast this with PUT which is a complete representation of a resource.` So here I expect that for connect and disconnect it is more like action like apply that uses POST above.

web/src/api/http.ts
jreidinger: after seeing below usage I would say that post change makes sense...but still for patch I kind of expect instructions how to modify source and below usage is at least confusing for me.

rust/agama-server/src/network/web.rs
```diff
@@ -101,10 +101,10 @@ pub async fn network_service<T: Adapter + Send + Sync + 'static>(
                 .put(update_connection)
                 .get(connection),
         )
-        .route("/connections/:id/connect", get(connect))
-        .route("/connections/:id/disconnect", get(disconnect))
+        .route("/connections/:id/connect", patch(connect))
```
jreidinger, line 104: see below reasoning, but I think that post is better here then patch.

rust/agama-server/src/network/web.rs
```diff
@@ -101,10 +101,10 @@ pub async fn network_service<T: Adapter + Send + Sync + 'static>(
...
+        .route("/connections/:id/disconnect", patch(disconnect))
         .route("/devices", get(devices))
-        .route("/system/apply", put(apply))
+        .route("/system/apply", post(apply))
```
jreidinger, line 107: what is rationale for this change? for me the main difference between post and put is idempotency. So when I call apply twice will it do two actions and only the first apply do something?

web/src/components/network/AddressesDataList.tsx
```diff
@@ -39,15 +39,23 @@ import {
...
+type Address = IPAddress & { id?: number };
+
 export default function AddressesDataList({
   addresses: originalAddresses,
   updateAddresses,
   allowEmpty = true,
+}: {
+  addresses: Address[];
+  updateAddresses: (addresses: Address[]) => void;
+  allowEmpty: boolean;
```
jreidinger, line 58: I am not much familiar with ts so just curious. When `allowEmpty` is predefined as true. shouldn't here be used `allowEmpty?` to indicate that it is optional and will be replaced by true? or it applies only if it is checked later for presence and uses true dynamically?

web/src/components/network/AddressesDataList.tsx
```diff
@@ -60,25 +68,28 @@ export default function AddressesDataList({
...
 
-  const renderAddress = ({ id, address, prefix }) => {
+  const renderAddress = ({ id, address, prefix }: Address) => {
     const renderDeleteAction = () => {
       if (!allowEmpty && addresses.length === 1) return null;
+      const buttonId = `delete-address-${id}-button`;
 
       return (
-        <DataListAction>
+        /** @ts-expect-error: https://github.com/patternfly/patternfly-react/issues/9823 */
```
jreidinger, line 89: interesting ts feature..I expect I will also use it in future during testing :)

web/src/components/network/DnsDataList.tsx
```diff
@@ -55,36 +66,36 @@ export default function DnsDataList({ servers: originalServers, updateDnsServers
...
 
-  const deleteServer = (id) => {
+  const deleteServer = (id: number) => {
     const serverIdx = servers.findIndex((dns) => dns.id === id);
     servers.splice(serverIdx, 1);
     updateDnsServers(servers);
   };
 
-  const renderDns = ({ id, address }) => {
+  const renderDns = ({ id, address }: DNS) => {
```
jreidinger, line 81: hmm..when I check DNS type it can have optional id...what happens when id is not in object in this case?

web/src/components/network/IpAddressInput.tsx
```diff
@@ -20,17 +20,26 @@
...
 
-const IpAddressInput = ({ placeholder, onError = () => null, ...props }) => {
-  const [validated, setValidated] = useState("default");
+const IpAddressInput = ({
+  label = _("IP Address"),
+  onError = (value) => null,
+  ...props
+}: TextInputProps & { defaultValue?: string }) => {
+  const [validated, setValidated] = useState(
+    !props.defaultValue || props.defaultValue === "" || isValidIp(props.defaultValue)
```
jreidinger, line 33: that complex ternary operator with extra complex condition really kills my eyes and brain...and I see it as getting code worse then in previous state. Is it possible to somehow improve it? like at least moving that condition outside of method call.

web/src/components/network/IpPrefixInput.tsx
```diff
@@ -20,17 +20,26 @@
...
 
-const IpPrefixInput = ({ placeholder, onError = () => null, ...props }) => {
-  const [validated, setValidated] = useState("default");
+const IpPrefixInput = ({
+  label = _("Ip prefix or netmask"),
+  onError = (value) => null,
+  ...props
+}: TextInputProps & { defaultValue?: string | number }) => {
+  const [validated, setValidated] = useState(
+    !props.defaultValue || props.defaultValue === "" || isValidIpPrefix(props.defaultValue)
```
jreidinger, line 33: same applies here...maybe having it in own method can help like `emptyOrValidIPrefix(props.defaultValue)` or maybe even with returned string something like `IpPrefixState(props.defaultValue)`

web/src/components/network/IpSettingsForm.tsx
```diff
@@ -143,10 +138,14 @@ export default function IpSettingsForm() {
     );
   };
 
+  console.log("Method", method);
+  console.log("ConnectionMethod.AUTO", ConnectionMethod.AUTO);
+  console.log("ConnectionMethod.MANUAL", ConnectionMethod.MANUAL);
```
jreidinger, line 143: this looks like debug lines that should not be merged

web/src/components/network/NetworkPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { partition } from "~/utils";
+
+const WiredConnections = ({ connections, devices }) => {
+  const total = connections.length;
+
+  return (
+    <CardField label={total > 0 && _("Wired")}>
+      <CardBody>
+        {total === 0 && <EmptyState title={_("No wired connections found")} icon="warning" />}
+        {total !== 0 && <ConnectionsTable connections={connections} devices={devices} />}
```
jreidinger, line 40: when I see this pair of condition, cannot be have simple if here? or ternary as suggested at https://legacy.reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

web/src/components/network/WifiConnectionForm.tsx
```diff
@@ -51,60 +56,62 @@ const selectorOptions = security_options.map((security) => (
...
-
-export default function WifiConnectionForm({ network, onCancel, onSubmitCallback }) {
-  const { network: client } = useInstallerClient();
-  const queryClient = useQueryClient();
-  const [error, setError] = useState(false);
-  const [isConnecting, setIsConnecting] = useState(false);
-  const [ssid, setSsid] = useState(network?.ssid || "");
-  const [password, setPassword] = useState(network?.password || "");
-  const [security, setSecurity] = useState(securityFrom(network?.security || []));
+// FIXME: improve error handling. The errors props shuld have a key/value error
```
jreidinger, line 59: typo shuld

web/src/components/network/WifiConnectionForm.tsx
```diff
@@ -51,60 +56,62 @@ const selectorOptions = security_options.map((security) => (
...
-        <Alert variant="warning" isInline title={_("Something went wrong")}>
-          <p>{_("Please, review provided settings and try again.")}</p>
+    /** TRANSLATORS: accessible name for the WiFi connection form */
+    <Form onSubmit={accept} aria-label={_("WiFi connection form")}>
+      {showErrors && (
+        <Alert
+          variant="warning"
+          isInline
+          title={
+            // @ts-expect-error
```
jreidinger, line 100: reasoning?

web/src/components/network/WifiNetworksListPage.tsx
```diff
@@ -179,69 +188,65 @@ const NetworkListName = ({ network }) => {
...
- * @param {function} props.onCancelCallback - the function to trigger when user cancel dismiss before connecting to a network
  */
-function WifiNetworksListPage({ networks = [] }) {
-  const { data } = useSelectedWifi();
-  const selected =
-    data.ssid === undefined ? HIDDEN_NETWORK : networks.find((n) => n.ssid === data.ssid);
-  const changeSelected = useSelectedWifiChange();
+function WifiNetworksListPage() {
+  const networks: WifiNetwork[] = useWifiNetworks();
+  // @ts-expect-error
```
jreidinger, line 226: reasoning?

web/src/components/network/WifiNetworksListPage.tsx
```diff
@@ -44,21 +44,28 @@ import {
...
+  useRemoveConnectionMutation,
+  useSelectedWifi,
+  useSelectedWifiChange,
+  useWifiNetworks,
+} from "~/queries/network";
+import { slugify } from "~/utils";
+import { connect, disconnect } from "~/api/network";
 
-const HIDDEN_NETWORK = Object.freeze({ hidden: true });
+type HiddenNetwork = { hidden: boolean };
```
jreidinger, line 64: this type quite smell for me as I expect that hidden network contain more entries...why WifiNetwork cannot be used? or maybe instead of const variable method that do that simple check for hidden value on common WifiNetwork type can be used?

web/src/components/network/NetworkPage.tsx
dgdavid: Sure, kind of left over

web/src/components/network/IpSettingsForm.tsx
dgdavid: Right!

web/src/queries/network.ts
```diff
@@ -0,0 +1,330 @@
...
+const useSelectedWifi = (): { ssid?: string; needsAuth?: boolean } => {
+  const { data } = useQuery(selectedWiFiNetworkQuery());
+  return data || {};
+};
+
+const useSelectedWifiChange = () => {
+  type SelectedWifi = {
+    ssid?: string;
+    hidden?: boolean;
+    needsAuth?: boolean;
```
jreidinger, line 199: so selected wifi can be empty? it looks kind of strange. I expect that at least ssid is mandatory.

web/src/components/network/IpPrefixInput.tsx
dgdavid: Let's try to improve it, thanks for commenting it :)

web/src/components/network/AddressesDataList.tsx
dgdavid: > I am not much familiar with ts so just curious. When `allowEmpty` is predefined as true. shouldn't here be used `allowEmpty?` to indicate that it is optional and will be replaced by true?

Good catch. I'll check it tomorrow when on front of laptop and editor, but lost probably you're right

web/src/utils/network.ts
```diff
@@ -0,0 +1,209 @@
...
+    return parseInt(value);
+  } else {
+    return ipaddr.IPv4.parse(value).prefixLengthFromSubnetMask();
+  }
+};
+
+/**
+ *  Converts an IP given in decimal format to text format
+ *
+ * FIXME: IPv6 is not supported yet.
```
jreidinger, line 92: I think it is something that should deserve follow up card as ipv6 start being more common

jreidinger:
finished with code review. Manual test is currently ENOTIME.

web/src/api/http.ts
dgdavid: Which verbs use for which actions / request will be for sure a topic for discussions. I personally have to revisit all the spec for refreshing my mind about HTTP verbs and the best usage of them. But as per RFC, https://www.rfc-editor.org/rfc/rfc5789, patch does not require a payload. It's optional.

As you pasted in another comment, `A PATCH request is considered a set of instructions on how to modify a resource`. In the network case, we're using PATCH for request a `connect` or `disconnect` _instructions_ (aka actions) which tells the server how to modify the resource: connection or disconnecting it.

web/src/api/network.ts
dgdavid: As commented at https://github.com/openSUSE/agama/pull/1519#discussion_r1714857840, you can think in these actions as the instructions you're requesting for modifying the connection state.

As I said in linked comment, I still having pending to refresh the HTTP verbs/requests. But I tend to think in POST as a request for creating something or appending more data to a resource (although for the latest I'd use a PUT). Also, I expect that servers returns the resource representation or at least where to retrieve it. None of these are the case for connect and disconnect actions for which we are _just_ requiring a change in the connection state, a request _side-effect_.

rust/agama-server/src/network/web.rs
dgdavid: See https://github.com/openSUSE/agama/pull/1519#discussion_r1714878241

If you don't mind, we can postpone this discussion until @teclator and @imobachgs are back. As far as I know, this was a change agreed by them. So we can merge as it is and change our mind later if needed.

rust/agama-server/src/network/web.rs
dgdavid: If I understood what @imobachgs told me last week, the rationale here is that we're using POST for these kind of actions, like it's the case of _probing_.

web/src/utils/network.ts
dgdavid: For sure! Adding it to the pending things in the PR description for easing the follow-up creation. Thanks!

web/src/queries/network.ts
dgdavid: It's not that strange. You can connect to a _hidden_ Wi-Fi network.

web/src/components/network/WifiNetworksListPage.tsx
dgdavid: Let me revisit this and see how we can improve it.

web/src/components/network/WifiNetworksListPage.tsx
dgdavid: I've to re-check it.

web/src/components/network/DnsDataList.tsx
dgdavid: When missing, we're adding it at line 60 when creating an internal DNS collection based on the original ones. This id/index is needed to be used as key and avoid the component messing all up when the user adds and removes servers. Related to https://react.dev/learn/rendering-lists#why-does-react-need-keys.

Something that can be improved, for sure, but completely out of scope of this PR. This component somehow inherits a behavior we implemented for cockpit-widget long ago.

web/src/components/network/IpAddressInput.tsx
dgdavid: Just a note: previous state was not working as expected when, although likely not possible, mounting the component with a wrong value as `defaultValue`.

web/src/components/network/WifiNetworksListPage.tsx
dgdavid: Ok, I see now. WifiNetwork is an AccessPoint plus WifiNetworkStatus plus other optional values like `hidden`. So, HIDDEN_NETWORK cannot be a WifiNetwork sinde it does belong to none access point. A bit lost here, so I prefere to wait until @teclator can have a look.

web/src/queries/network.ts
dgdavid: Yeah, I know how hidden network works :) But the problem is that this is a "UI query" for keeping which network is/was selected by the user in order to show the proper information about it: the connection form or its configuration/state and actions.

But the user could select a "hidden network", for which the UI does not know the SSID in advance. That's why the ssid here is optional. Maybe a matter of naming or a matter or typing improvement (I'm just landing in the typed languages world :P)

How would you improve it?

jreidinger: at least document that info in the type as it is confusing for reader. So mention that for hidden network ssid is not known in advance. I am not sure if you can have something in ts like that hidden is true or ssid is specified.

dgdavid: >  I am not sure if you can have something in ts like that hidden is true or ssid is specified.

Interesting point. I'm going to see if that it's possible. Thanks a ton!

dgdavid: Argh! Actually is a bit more tricky than I though because I forgot to mention a small detail in previous explanations: we're updating the information in the query in a moment in which we don't know the ssid but only that the authentication failed. So, we assumed that authentication failed for the selected connection. See https://github.com/openSUSE/agama/pull/1519/files#diff-f1b1879261cce23d74bb1716c65fbbcb487ee658acf074133fc6d6def7288736R223-R263

So, to make it work without TypeScript complaints the changes would be something like 


```diff
diff --git a/web/src/queries/network.ts b/web/src/queries/network.ts
index 00754d6d..d8fecc7a 100644
--- a/web/src/queries/network.ts
+++ b/web/src/queries/network.ts
@@ -193,16 +193,25 @@ const useSelectedWifi = (): { ssid?: string; needsAuth?: boolean; hidden?: boole
 };
 
 const useSelectedWifiChange = () => {
-  type SelectedWifi = {
-    ssid?: string;
-    hidden?: boolean;
-    needsAuth?: boolean;
+  type VisibleWifi = {
+    ssid: string;
+    hidden?: false | never;
   };
 
+  type HiddenWifi = {
+    ssid?: never;
+    hidden: true;
+  };
+
+  type KnownWifi = VisibleWifi & { needsAuth?: boolean };
+
+  type SelectedWifi = KnownWifi | HiddenWifi;
+
   const queryClient = useQueryClient();
 
   const mutation = useMutation({
-    mutationFn: async (data: SelectedWifi): Promise<SelectedWifi> => Promise.resolve(data),
+    mutationFn: async (data: Partial<SelectedWifi>): Promise<Partial<SelectedWifi>> =>
+      Promise.resolve(data),
     onSuccess: (data: SelectedWifi) => {
       queryClient.setQueryData(["wifi", "selected"], (prev: SelectedWifi) => ({
         ssid: prev.ssid,
```

On one hand, from the _typing POV_ would be more clear that a hidden wifi is _unknown_ until the user enters the ssid and became it in a _KnownWifi_. On the other hand... although safer and robust code looks a bit complicated. What do you think?

dgdavid:
@teclator,  please be aware that I've merged master on this one to keep it up-to-date. See https://github.com/openSUSE/agama/pull/1519/commits/02220c0fdaeda507a63a3be8a82140a0ddf4c0af

jreidinger (approved):

PR #1520, opened 2024-08-01
fix(web): replace exfat-utils with exfatprogs

imobachgs:
Use `exfatprogs` instead of `exfat-utils`.


ancorgs (approved):
LGTM.

For completeness, libstorage-ng will keep working because it uses the binaries, that are the same for both packages.

See https://github.com/yast/yast-storage-ng/issues/1214

PR #1522, opened 2024-08-01
feat(live): Display QR codes at the console for easier accessing Agama with smartphones

lslezak:
## Problem

- Make accessing Agama easier for smartphones by scanning an QR code
- And this is a pretty cool feature for advertising Agama!  :smiley: 

## Solution

- Use the `qrencode` tool for generating the QR codes as text (it uses special UTF-8 graphical symbols)
- Display the codes side-by-side with URL below them
- If there is not enough space the QR codes are displayed only for the subset of the URLs, the other URLs are displayed just as text  without QR code. That means the QR codes always displayed in a single line.
- When running on serial console or when on the VGA plain text console the QR codes are not displayed (the terminal width must be greater than 80 to display them)
- The codes are refreshed whenever the network configuration is changed (see the video at the end)
- Tested manually

## Notes

The root password is randomly generated so the system actually knows the password. I was thinking about rendering URLs with embedded password (something like `https://1.2.3.4?password=PASSWORD`) so then using a smartphone would be even easier.

But that has some security drawbacks. The terminal already contains the root password, but you would need to be close to the screen to get it. On the other hand QR codes are designed for robustness, the QR code might be blurred, some part can be missing but the scanner will still be able to get the full content. That means you do not need to be next to the screen, the risk of eavesdropping for QR is higher than for a plain text password.

We could discuss that with the security team...

## Screenshots

| Standard resolution, 1 network device  | |
| -- | -- |
| qr_1dev | |
| Standard resolution, 4 network devices  | Smaller resolution, 4 network devices, only 3 QR codes displayed |
| qr_4devs | qr_4devs_800  |
| Serial console (80x24), 4 network devices, no QR code displayed | VGA text mode (80x25), 4 network devices, no QR code displayed |
| qr_4devs_serial | qr_4devs_text |

Dynamically adding network connections (activating 3 more network interfaces)

[Agama-screen0.webm](https://github.com/user-attachments/assets/c17c9c3a-c74b-4fd3-ac6a-87bbc2298e0f)

### DIFF
--- a/live/root/usr/bin/agama-issue-generator
+++ b/live/root/usr/bin/agama-issue-generator
@@ -139,6 +139,91 @@ generate_avahi_url() {
     done
 }
 
+# function for centering text
+# $1 - the text
+# $2 - requested width
+function center_text() {
+  LEN=${#1}
+  PADDING_LEN=$(($2 - LEN))
+  PADDING_LEN=$((PADDING_LEN / 2))
+  PADDING="$(printf '%*s' $PADDING_LEN)"
+  echo "$PADDING$1$PADDING"
+}
+
+# generate QR codes for the access URLs
+function create_qr_codes() {
+    ADDRESSES=("$@")
+    # width of the generated QR code
+    QR_WIDTH=30
+
+    # check if serial console is used, get the terminal size (width)
+    TERM_WIDTH=$(stty -F /dev/ttyS0 size 2> /dev/null | cut -d " " -f 2)
+
+    # otherwise check the first console size
+    if [ -z "$TERM_WIDTH" ]; then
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
Awesome :heart: , thanks a lot for both, the feature and the detailed description. It's useful for users and... to encourage us to have in mind a bit more the mobile version :stuck_out_tongue_closed_eyes:

mvidner:
@lslezak BTW I wondered how my phone would deal with multiple QR codes, and it turns out pretty well: when it picks the wrong one I just move it back and forth and it refocuses on the one in the middle. So multiple codes close together are not a problem.

PR #1530, opened 2024-08-05
refactor(web): use `axios` instead of `fetch`

imobachgs:
* Use [axios](https://axios-http.com/) instead of `fetch` as it implements a promise-based API which is better for our use case.
* Move the logic related to interacting with the API to an `API` module.
* Handle the `ServiceStatusChanged` signal properly.

## To do

- [x] software
- [x] l10n
- [x] users
- [x] issues
- [x] progress
- [x] status

### DIFF
--- a/web/src/api/issues.ts
+++ b/web/src/api/issues.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) [2024] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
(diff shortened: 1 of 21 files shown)


web/src/api/issues.ts
```diff
@@ -0,0 +1,37 @@
...
+  product: "software/issues/product",
+  software: "software/issues/software",
+  users: "users/issues",
+  storage: "storage/issues",
+};
+
+/**
+ * Return the issues of the given scope.
+ */
+const fetchIssues = (scope: IssuesScope) => get(`/api/${URLS[scope]}`);
```
dgdavid, line 35: NP/Question: what about adding the return type?


```suggestion
const fetchIssues = (scope: IssuesScope): Promise<Issue[]> => get(`/api/${URLS[scope]}`);
```

or 

```diff
- const fetchIssues = (scope: IssuesScope) => get(`/api/${URLS[scope]}`);
+ const fetchIssues = (scope: IssuesScope): Issue[] => get(`/api/${URLS[scope]}`);
```

Not sure now if all gets are promises or only these we're `await`_ing for_.

dgdavid (approved):
Thanks Imo, looks much much better than before.

I just added there a **non-blocking* NP/Question comments that applies to all `fetch`methods in these src/api namespaces. Feel free to add these returns types or to simply go ahead with the code as it is now.

dgdavid (approved):

PR #1532, opened 2024-08-11
feat(rust): add an HTTP/JSON API to DASD operations

imobachgs:
Trello: https://trello.com/c/2LC8JEZR/3743-8-agama-adapt-dasd-configuration-to-current-architecture

This PR is the first step to bring back support for DASD devices on s390x architectures. It includes:

* Extending the HTTP API to expose DASD devices and formatting jobs.
* Adding DASD-related events (adding, updating and changing DASD devices and format progress).

After some discussion, we decided to retain the original intention of the D-Bus API at HTTP level.

> [!WARNING]
> The code has barely been tested, and it may be incomplete or contain bugs. But it should be a good enough starting point.

## Examples

This section contains a few examples you can use to explore the API. For the ones using `curl`, it is required to have a `headers.txt` file containing the credentials (similar to the following example, but replacing `TOKEN` with your actual token):

```
Content-Type: application/json
Authorization: Bearer TOKEN
```

<details>
<summary>DASD operations</summary>

Remember that the `-k` disables the SSL certificate checking.

### Probing for DASD devices

```bash
curl -k -H @headers.txt -X POST https://$AGAMA_SERVER/api/storage/dasd/probe
```

### Listing DASD devices

```bash
$ curl -k --silent -H @headers.txt -X GET https://$AGAMA_SERVER/api/storage/dasd/devices | jq
[
  {
    "id": "0.0.0160",
    "enabled": true,
    "device_name": "dasda",
    "formatted": true,
    "diag": false,
    "status": "active",
    "device_type": "ECKD",
    "access_type": "rw",
    "partition_info": ""
  }
]
```

### Enabling DASD devices

> [!NOTE]
> We might consider using `PATCH` or even `POST` for enabling, disabling, setting diag and formatting the device.

```bash
$ curl -k -H @headers.txt -X POST https://$AGAMA_SERVER/api/storage/dasd/enable \
  -d '{"devices": ["0.0.0160"]}'
```

### Disabling devices

```bash
$ curl -k -H @headers.txt -X POST https://$AGAMA_SERVER/api/storage/dasd/disable \
  -d '{"devices": ["0.0.0160"]}'
```

### Setting the DIAG attribute

```bash
$ curl -k -H @headers.txt -X PUT https://$AGAMA_SERVER/api/storage/dasd/diag \
  -d '{"devices": ["0.0.0160"], "diag": true}'
```

### Formatting devices

```bash
$ curl -k -H @headers.txt -X PUT https://$AGAMA_SERVER/api/storage/dasd/diag \
  -d '{"devices": ["0.0.0160"], "diag": true}'
```
</details>

<details>
<summary>Jobs operation</summary>

### Listing jobs

```bash
$ curl -k -H @headers.txt -X GET https://$AGAMA_SERVER/api/storage/jobs | jq
[
  {
    "id": "/org/opensuse/Agama/Storage1/jobs/3",
    "running": false,
    "exit_code": 0
  },
  {
    "id": "/org/opensuse/Agama/Storage1/jobs/2",
    "running": false,
    "exit_code": 0
  }
]
```

</details>

<details>
<summary>Events</summary>

The PR adds the following events:

* `DASDDeviceAdded`, `DASDDeviceChaned` and `DASDDeviceRemoved`.
* `JobAdded`, `JobChanged` and `JobRemoved`.
* `DASDFormatProgress`.

> [!WARNING]
> The `Finished` signal of the `org.opensuse.Agama.Storage1.Job` is not handled.

Let's see a few examples:

## Formatting progress

```json
{"type":"DASDFormatJobChanged","job_id":"/org/opensuse/Agama/Storage1/jobs/9","total":30050,"step":30050,"done":true}
```

### DASD device changes

```
{"type":"DASDDeviceChanged","device":{"id":"0.0.0160","enabled":true,"device_name":"dasda","formatted":true,"diag":false,"status":"active","device_type":"ECKD","access_type":"rw","partition_info":""}
```

</details>

## A notes about signals

We have discussed whether we are abusing standard D-Bus signals a bit. For instance, we rely a lot on `PropertiesChanged`, `InterfacesAdded`, `InterfacesRemoved`, etc., even for meaningful and important system events. Then, the client code needs to figure out what those signals actually mean. Also, unless you use a proxy (and that's not always possible), you cannot rely on the type checking that the introspection offers.

Of course, that's out of the scope of this pull request.

## Code repetition

As you might notice, there is quite some repetitive code. At this point, I just wanted to have some working code. We can reduce the repetition later. Take into account that improving our signals would help reduce the amount of code on the client side.

### DIFF
--- a/rust/agama-lib/src/storage/client/dasd.rs
+++ b/rust/agama-lib/src/storage/client/dasd.rs
@@ -1,0 +1,100 @@
+//! Implements a client to access Agama's D-Bus API related to DASD management.
+
+use zbus::{
+    fdo::ObjectManagerProxy,
+    zvariant::{ObjectPath, OwnedObjectPath},
+    Connection,
+};
+
+use crate::{
+    error::ServiceError,
+    storage::{model::dasd::DASDDevice, proxies::DASDManagerProxy},
+};
+
+/// Client to connect to Agama's D-Bus API for DASD management.
+#[derive(Clone)]
+pub struct DASDClient<'a> {
+    manager_proxy: DASDManagerProxy<'a>,
+    object_manager_proxy: ObjectManagerProxy<'a>,
+}
+
+impl<'a> DASDClient<'a> {
+    pub async fn new(connection: Connection) -> Result<DASDClient<'a>, ServiceError> {
+        let manager_proxy = DASDManagerProxy::new(&connection).await?;
+        let object_manager_proxy = ObjectManagerProxy::builder(&connection)
+            .destination("org.opensuse.Agama.Storage1")?
+            .path("/org/opensuse/Agama/Storage1")?
+            .build()
+            .await?;
+        Ok(Self {
+            manager_proxy,
+            object_manager_proxy,
+        })
+    }
+
+    pub async fn probe(&self) -> Result<(), ServiceError> {
+        Ok(self.manager_proxy.probe().await?)
+    }
+
+    pub async fn devices(&self) -> Result<Vec<(OwnedObjectPath, DASDDevice)>, ServiceError> {
+        let managed_objects = self.object_manager_proxy.get_managed_objects().await?;
+
+        let mut devices: Vec<(OwnedObjectPath, DASDDevice)> = vec![];
+        for (path, ifaces) in managed_objects {
+            if let Some(properties) = ifaces.get("org.opensuse.Agama.Storage1.DASD.Device") {
+                match DASDDevice::try_from(properties) {
+                    Ok(device) => {
+                        devices.push((path, device));
+                    }
+                    Err(error) => {
+                        log::warn!("Not a valid DASD device: {}", error);
+                    }
+                }
+            }
+        }
+        Ok(devices)
+    }
+
+    pub async fn format(&self, ids: &[&str]) -> Result<(), ServiceError> {
+        let selected = self.find_devices(ids).await?;
+        let references = selected.iter().collect::<Vec<&ObjectPath<'_>>>();
+        self.manager_proxy.format(&references).await?;
+        Ok(())
+    }
+
+    pub async fn enable(&self, ids: &[&str]) -> Result<(), ServiceError> {
+        let selected = self.find_devices(ids).await?;
+        let references = selected.iter().collect::<Vec<&ObjectPath<'_>>>();
+        self.manager_proxy.enable(&references).await?;
+        Ok(())
+    }
+
+    pub async fn disable(&self, ids: &[&str]) -> Result<(), ServiceError> {
+        let selected = self.find_devices(ids).await?;
+        let references = selected.iter().collect::<Vec<&ObjectPath<'_>>>();
+        self.manager_proxy.disable(&references).await?;
+        Ok(())
+    }
+
+    pub async fn set_diag(&self, ids: &[&str], diag: bool) -> Result<(), ServiceError> {
+        let selected = self.find_devices(ids).await?;
+        let references = selected.iter().collect::<Vec<&ObjectPath<'_>>>();
+        self.manager_proxy.set_diag(&references, diag).await?;
+        Ok(())
+    }
+
+    async fn find_devices(&self, ids: &[&str]) -> Result<Vec<ObjectPath<'_>>, ServiceError> {
+        let devices = self.devices().await?;
+        let selected: Vec<ObjectPath> = devices
+            .into_iter()
+            .filter_map(|(path, device)| {
+                if ids.contains(&device.id.as_str()) {
+                    Some(path.into_inner())
+                } else {
+                    None
+                }
+            })
+            .collect();
+        Ok(selected)
+    }
+}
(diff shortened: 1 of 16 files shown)


rust/agama-lib/src/storage/model/dasd.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::collections::HashMap;
+
+use serde::Serialize;
+use zbus::zvariant::OwnedValue;
+
+use crate::{dbus::get_property, error::ServiceError};
+
+/// Represents a DASD device (specific to s390x systems).
+#[derive(Clone, Debug, Serialize, Default, utoipa::ToSchema)]
+pub struct DASDDevice {
```
imobachgs, line 11: Some of the fields in this struct (`status`, `device_type` and/or `access_type`) might deserve an enum.

rust/agama-server/src/storage/web/dasd.rs
```diff
@@ -0,0 +1,196 @@
...
+    Ok(Json(()))
+}
+
+/// Enables a set of devices.
+#[utoipa::path(
+    post,
+    path="/enable",
+    context_path="/api/storage/dasd",
+    responses(
+        (status = OK, description = "The formatting process started.")
```
jreidinger, line 132: c&p description

rust/agama-server/src/storage/web/dasd.rs
```diff
@@ -0,0 +1,196 @@
...
+    Ok(Json(()))
+}
+
+/// Disables a set of devices.
+#[utoipa::path(
+    post,
+    path="/disable",
+    context_path="/api/storage/dasd",
+    responses(
+        (status = OK, description = "The formatting process started.")
```
jreidinger, line 149: c&p description

rust/agama-server/src/storage/web/dasd.rs
```diff
@@ -0,0 +1,196 @@
...
+    Ok(Json(()))
+}
+
+/// Sets the diag property for a set of devices.
+#[utoipa::path(
+     put,
+     path="/diag",
+     context_path="/api/storage/dasd",
+     responses(
+         (status = OK, description = "The formatting process started.")
```
jreidinger, line 166: c&p description

rust/agama-server/src/web/common/jobs.rs
```diff
@@ -0,0 +1,186 @@
...
+    let state = JobsState { client };
+    Ok(Router::new().route("/jobs", get(jobs)).with_state(state))
+}
+
+#[derive(Clone)]
+struct JobsState<'a> {
+    client: JobsClient<'a>,
+}
+
+async fn jobs(State(state): State<JobsState<'_>>) -> Result<Json<Vec<Job>>, Error> {
```
jreidinger, line 38: where is openAPI documentation for this route?

rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -0,0 +1,110 @@
...
+                    Err(error) => {
+                        log::warn!("Not a valid DASD device: {}", error);
+                    }
+                }
+            }
+        }
+        Ok(devices)
+    }
+
+    pub async fn format(&self, ids: &[&str]) -> Result<String, ServiceError> {
```
jreidinger, line 64: API looks strange, it allows formating multiple DASDs but it return only one path for job.

imobachgs: That's what I deduced from https://github.com/openSUSE/agama/blob/master/service/lib/agama/dbus/storage/interfaces/dasd_manager.rb#L139.

teclator: there is only one job, that is ok, but then we need to check the status for each device, I think the missing part is the job Summary call on formatProgress callback.

https://github.com/openSUSE/agama/blob/master/web/src/client/storage.js#L891

teclator: I say nothing, it already uses the Summary but the DASD path is not sent at all...

see ... https://github.com/openSUSE/agama/pull/1532/files#diff-481b9b86ced79110f138995e6097a0f0cde62016924c72f1c5c6397940566654R210
and ... https://github.com/openSUSE/agama/pull/1532/files#diff-481b9b86ced79110f138995e6097a0f0cde62016924c72f1c5c6397940566654R220

rust/agama-server/src/web/common/jobs.rs
imobachgs: It is an unsolved problem: the jobs can appear under different URLs, so we are not documenting them (check the [issues API](https://github.com/openSUSE/agama/blob/master/rust/agama-server/src/web/common.rs#L288)). Something to solve in the future.

rust/agama-server/src/web/event.rs
```diff
@@ -17,7 +17,7 @@ use tokio::sync::broadcast::{Receiver, Sender};
 use super::common::Issue;
 
 #[derive(Clone, Debug, Serialize)]
-#[serde(tag = "type")]
+#[serde(rename_all = "camelCase", tag = "type")]
```
jreidinger, line 20: good, I recheck that it affects only dasd format event. So it is safe.

jreidinger (approved):

PR #1534, opened 2024-08-12
less strict nokogiri dependency

jreidinger:
## Problem

TW has nokogiri 1.16 and SLFO has 1.15. Due to our strict dependencies we cannot build against both.

## Solution

nokogiri follows semver, so it should be safe to depends only on minor version and not on patch level.


ancorgs (commented):
LGTM.

PR #1535, opened 2024-08-12
Fix gpg import

jreidinger:
## Problem

In staging there is no opensuse keys for PXE build which causes build failure.

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:adi:34/agama-installer:openSUSE-PXE/images/x86_64

## Solution

Import keys only if there is any available.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


ancorgs (approved):
LGTM

PR #1537, opened 2024-08-14
fix(web): add forgotten ReactRouter#useHref mock

dgdavid:
https://github.com/openSUSE/agama/pull/1536 faced some problems in CI not reproducible locally resulting in a set of changes in the test suite to make CI happy.

But these CI complaints were right since a mock for ReactRouter#useHref hook was forgotten and not sent as part of the changes, reason why initial tests were not working.

Now we have to decide: to go ahead with this PR and continue mocking ReactRouter for these simple tests  or to keep previous tests which were not mocking mentioned `useHref` hook.  In any case, changes file does not need to have a new entry for this.

### DIFF
--- a/web/src/components/core/Link.test.tsx
+++ b/web/src/components/core/Link.test.tsx
@@ -21,20 +21,19 @@
 
 import React from "react";
 import { screen } from "@testing-library/react";
-import { installerRender } from "~/test-utils";
+import { plainRender } from "~/test-utils";
 import { Link } from "~/components/core";
 
 describe("Link", () => {
   it("renders an HTML `a` tag with the `href` attribute set to given `to` prop", () => {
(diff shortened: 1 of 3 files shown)


ancorgs:
Well, mocking `useHref` looks more consistent with all the other related tests. Isn't it?

dgdavid:
> Well, mocking `useHref` looks more consistent with all the other related tests. Isn't it?

Yes, it is.

Just to be clear, I'm all for the mocking here because I expect that behavior of useHref and others ReactRouter hooks are already tested by the library. So, I don't care about their output but that the component in which they are used use such a value. Thus, mocking is perfectly fine to me here. But I had to ask just in case you guys has a different view.

ancorgs (approved):
LGTM

PR #1494, opened 2024-07-23
refactor(web): UI code pruning and clean up `round #2`

dgdavid:
Related to  #1441, just another round of UI code improvements and clean up.

PR #1540, opened 2024-08-14
refactor(web): UI code pruning and clean up `round #3`

dgdavid:
**Apart from a bit of clean up, this PR is intended for start writing better core components** that has been on hold for a few months already.

It's the case of _core/Page_ component, which has been rewritten almost for scratch and now makes the weird _core/CardField transitioning component_ obsolete.

Please, note  that this set of changes **continues with the migration to TypeScript for touched files** and also **introduce a PatternFly/Flex wrapper** in order to ease the work with its responsive props. It's a bit complex because the (ab)use of advanced types but it does the job without introducing props unknown by PF/Flex. As said in the file comments, ideally

> would be better to add these responsive props shortcuts direclty in PF/Flex to allow the consumer to just set the `default` value when not needed to change it depending on the breakpoint. But at this moment we're a bit short of time for creating and testing such an elaborated PR against upstream.

**Reviewers**: Look at it (core/Flex.tsx) as a PoC to see if **we** are willing to introduce it and similar ones for the other PatternFly components using responsive props.

---

Related to #1441 and #1494


web/src/components/layout/Flex.tsx
```diff
@@ -0,0 +1,151 @@
...
+// NOTE: PF/Flex#order prop is missing "sm" breakpoint
+// NOTE: Don't know why these ommited props are being captured otherwise
+type ResponsiveFlexProps = {
+  [Key in keyof Omit<FlexProps, "download" | "inlist"> as FlexProps[Key] extends {
+    default?: unknown;
+  }
+    ? Key
+    : // @ts-ignore
+      never]: FlexProps[Key]["default"] | FlexProps[Key];
+};
```
dgdavid, line 48: Just in case of doubt, what this advanced type is doing is something like:

* Asking for all FlexProps and capturing these that somehow looks like a responsive one. I.e., capturing only the ones that are objects with a `default` optional prop. As said in the comment, don't know why "download" and "inlist" are captured although they does not look like mentioned object. So, just ignore them.
* If the prop was captured, uses an union of its own type and the type of the default key.
* If the prop was not captured, just ignore it in the new type by using never.

web/src/components/layout/Flex.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * the consumer to just set the `default` value when not needed to change it depending on
+ * the breakpoint. But at this moment we're a bit short of time for creating and testing such
+ * an elaborated PR against upstream.
+ *
+ * BTW, the lines for extracting an object from the type were borrowed from
+ * https://dev.to/scooperdev/generate-array-of-all-an-interfaces-keys-with-typescript-4hbf
+ */
+
+// NOTE: PF/Flex#order prop is missing "sm" breakpoint
+// NOTE: Don't know why these ommited props are being captured otherwise
```
imobachgs, line 40: I would say that you already know that, right? (considered as *any*)

web/src/components/core/PageNext.tsx
```diff
@@ -0,0 +1,219 @@
...
+  descriptionProps?: CardBodyProps;
+  headerLevel?: TitleProps["headingLevel"];
+  pfCardProps?: CardProps;
+  pfCardHeaderProps?: CardHeaderProps;
+  pfCardBodyProps?: CardBodyProps;
+};
+
+// FIXME: add a Page.Back for navigationg to -1 insetad of ".."?
+type PageActionProps = { navigateTo?: string | number } & ButtonProps;
+type PageSubmitActionProps = { form: string } & ButtonProps;
```
imobachgs: maybe `formId`?

web/src/components/core/PageNext.tsx
```diff
@@ -0,0 +1,219 @@
...
+  pfCardHeaderProps,
+  pfCardBodyProps,
+  children,
+}: React.PropsWithChildren<SectionProps>) => {
+  const renderTitle = !!title && title.trim() !== "";
+  const renderValue = React.isValidElement(value);
+  const renderDescription = !!description && description.trim() !== "";
+  const renderHeader = renderTitle || renderValue;
+
+  console.log("here");
```
imobachgs: leftover

web/src/components/core/PasswordAndConfirmationInput.tsx
```diff
@@ -19,34 +19,43 @@
...
 import { FormGroup } from "@patternfly/react-core";
 import { FormValidationError, PasswordInput } from "~/components/core";
 import { _ } from "~/i18n";
 
-// TODO: improve the component to allow working only in uncontrlled mode if
-// needed.
-// TODO: improve the showErrors thingy
+// TODO:
+//  * add documentation,
+//  * allow working only in uncontrlled mode if needed, and
```
imobachgs, line 29: typo: uncontrolled

web/src/components/core/Page.tsx
```diff
@@ -42,24 +42,44 @@ import { Flex } from "~/components/layout";
...
   "aria-label"?: string;
+  /** Part of the header that complements the title as a representation of the
+   * section state. E.g. "Encryption enabled", where "Encryption" is the title
+   * and "enabled" the value */
   value?: React.ReactNode;
-  description?: string;
+  /** Elements to be rendered in the section footer */
   actions?: React.ReactNode;
-  descriptionProps?: CardBodyProps;
+  /** As sort as possible yet as much as needed text for describing what the section is about, if needed */
```
imobachgs, line 62: typo: short

imobachgs (commented):
I like the new `Page` component :-)

web/src/components/layout/Flex.tsx
dgdavid: Right!

web/src/components/core/PageNext.tsx
dgdavid: Thanks!

dgdavid:
> I like the new `Page` component :-)

Thanks a lot! I think it's looking better than before, even though  there is a lot of work to do yet. But yes, I'm happy with this new Page version, much simpler than before. Also, having the "section" thingy as part of it instead of the "CardField" component helps to write Agama pages IMHO.

web/src/components/layout/Flex.tsx
```diff
@@ -37,7 +37,9 @@ import { Flex as PFFlex, FlexProps, FlexItem, FlexItemProps } from "@patternfly/
  */
 
 // NOTE: PF/Flex#order prop is missing "sm" breakpoint
-// NOTE: Don't know why these ommited props are being captured otherwise
+// NOTE: The ommited props match the extends constraint becuase they are typed
```
imobachgs, line 40:
```suggestion
// NOTE: The omitted props match the extends constraint because they are typed
```

imobachgs (approved):

PR #1544, opened 2024-08-19
Early finish fix

jreidinger:
## Problem

See https://github.com/openSUSE/agama/issues/1542

## Solution

After debugging it shows that queries work, but when query gets invalidated it reread data and it wrongly parse json data and set busy status as undefined.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/service/.gitignore
+++ b/service/.gitignore
@@ -49,7 +49,7 @@ build-iPhoneSimulator/
 # for a library or gem, you might want to ignore these files since the code is
 # intended to run in multiple environments; otherwise, check them in:
 # Gemfile.lock
-# .ruby-version
+.ruby-version
 # .ruby-gemset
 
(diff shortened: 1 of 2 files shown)


ancorgs (approved):
I would appreciate if anyone with more knowledge of the web UI would be around. But it's August. :-) Personally, this LGTM.

jreidinger:
If there is anyone with more web knowledge I would be happy to let him do request instead of me :joy:

jreidinger:
@ancorgs sadly fix is wrong...basically it looks like SLE16 uses old rust backend! the change was done by imo https://github.com/openSUSE/agama/commit/afef47b21071d7b3addad1dd6c8c6f8d641843a2 and this PR breaks opensuse build :( ... we have some issue with auto submission to IBS project :(

Issue #1474, opened 2024-07-17
openQA: product-testing please provide a way to inject a custom repository


lkocman:
We need to be able to change/inject repository for product (e.g. Leap 16) for installation in openQA

We need to be able to test against repositories from the new product build in OBS rather than previously published repo in https://download.opensuse.org/distribution/leap/16.0/repo/oss/.

I understand that we pass the repository as a boot param on Leap 15.

openQA sync's it from OBS  with openqa obs sync.

lkocman:
The initial test suite would be probably installing from https://download.opensuse.org/distribution/leap/16.0/repo/oss/ which is okay for pre-Alpha ... but we need to get that changed soon.

lkocman:
We've touched this issue with Dominique on our 1x1 when we were discussing testing, as the download-o-o repo is not acceptable for e.g. staging testing etc.

The URL edit is also improtant if we have download-o-o outage like we've had 2 days ago. Mirrors could have been still used.

lkocman:
Closing the issue! Thank you all for assistance.

lkocman:
Can be tested with https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-9.0.0-openSUSE-Build11.3.iso and newer

PR #1545, opened 2024-08-19
openQA: Allow overriding of install repos #1474

lkocman:
* Override urls by using agama.install_url=https://url1,https://url2 boot param
* Add logger so we can print some repo related debug messages that can be captured by agama logs store

## Problem
We need to be able to test installation of preproduction data. See details in #1474 

## Example usage

Add agama.install_url as a boot argument to the linux kernel. Multiple urls can be seprated by commas.

image

```
# cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-6.10.5-1-default root=UUID=4fd273b6-0453-4b36-8f58-a81e70bc9b9a splash=silent mitigations=auto quiet security=apparmor agama.install_url=https://download.opensuse.org/repositories/openSUSE:/Leap:/16.0/product/repo/Leap-16.0-aarch64-ppc64le-s390x-x86_64,https://download.opensuse.org/repositories/openSUSE:/Leap:/16.0/standard/x86_64/
```

This will now set https://download.opensuse.org/repositories/openSUSE:/Leap:/16.0/product/repo/Leap-16.0-aarch64-ppc64le-s390x-x86_64 as an installation_url to all products. This is extremely helpful for staging testing or any testing in openQA before publishing changes to the public repository.

Note: if you specify incorrect urls, you can at least the value it in the overview screen :-)
image

### DIFF
--- a/doc/boot_arguments.md
+++ b/doc/boot_arguments.md
@@ -15,6 +15,22 @@ See [URL handling in the
 installer](https://github.com/yast/yast-installation/blob/master/doc/url.md) to find more details
 about the supported URLs.
 
+## Custom Installation URL Configuration
+
+You can override the default `installation_url` set in the product files [here](https://github.com/openSUSE/agama/tree/master/products.d) by passing the `agama.install_url` parameter as a boot option in the bootloader.
+This is particularly useful for any pre-production testing in openQA. 
+
+**Note:** Setting this variable will impact all products. 
+
+### Example Usage
+
+To specify a custom installation URLs, pass following as a parameter to kernel in the bootloader.
+You can specify multiple URLs by separating them with commas.
+
+```
+agama.install_url=https://myrepo,https://myrepo2
+```
+
 ## Changing configuration values
 
 Instead of loading a full configuration file, you might be interested in adjusting just a few
--- a/service/lib/agama/software/product_builder.rb
+++ b/service/lib/agama/software/product_builder.rb
@@ -19,24 +19,29 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "agama/cmdline_args"
 require "agama/software/product"
+require "logger"
 
 module Agama
   module Software
     # Builds products from the information of a config file.
     class ProductBuilder
       # @param config [Agama::Config]
-      def initialize(config)
+      def initialize(config, logger: Logger.new($stdout))
         @config = config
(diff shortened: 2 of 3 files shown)


service/lib/agama/software/product_builder.rb
```diff
@@ -45,13 +50,19 @@ def build
...
         Agama::Software::Product.new(id).tap do |product|
           product.display_name = attrs["name"]
           product.description = attrs["description"]
           product.name = data[:name]
           product.version = data[:version]
-          product.repositories = data[:repositories]
+          # Issue#1474 Use agama.install_url=https://path-to-repodata
+          if cmdline_args.data.key?("install_url")
+            @logger.info "agama.install_url is set to " + cmdline_args.data["install_url"]
+            product.repositories = [cmdline_args.data["install_url"]]
```
jreidinger, line 62: in general looks good, but this allows only single installation repository, which I am not sure is enough. Will staging need more installation repositories? if so, then I will add there some `split(",")` to allow multiple repos.

lkocman: good point, will do! also any idea how to to transform the code so ABC score/rubocop is happy?

lkocman:
(squashed my changes)

doc/boot_arguments.md
```diff
@@ -15,6 +15,29 @@ See [URL handling in the
...
+agama.install_url=https://myrepo,https://myrepo2
+```
+
+### Multiple URLs
+
+
+### Example Usage
+
+To specify a custom installation URL, use the following syntax:
+
```
jreidinger, line 39: here it missing how syntax with multiple repositories will look like.

lkocman: Will fix.

jreidinger (approved):

doc/boot_arguments.md
```diff
@@ -15,6 +15,22 @@ See [URL handling in the
 installer](https://github.com/yast/yast-installation/blob/master/doc/url.md) to find more details
 about the supported URLs.
 
+## Custom Installation URL Configuration
+
+You can override the default `installation_url` set in the product files [here](https://github.com/openSUSE/agama/tree/master/products.d) by passing the `agama.install_url` parameter as a boot option in the bootloader.
```
imobachgs, line 20: Please, I would try to use the same name for the parameter, either `install_url` or `installation_url` but the same.

PR #1546, opened 2024-08-19
Revert "Early finish fix"

jreidinger:
Reverts openSUSE/agama#1544

The change is wrong as it was tested/debugged against old version of backend that does not have https://github.com/openSUSE/agama/commit/afef47b21071d7b3addad1dd6c8c6f8d641843a2 ... The result is that new sources are now broken.


ancorgs (approved):
LGTM

PR #1547, opened 2024-08-19
Fix opensuse build key

jreidinger:
## Problem

see https://github.com/openSUSE/agama/issues/1538

## Solution

Add beside staging keys also official openSUSE ones.

## Testing

- due to broken opensuse build not tested, only by manual comparison of working build and non-working one where the later one missing added package.


ancorgs (approved):
LGTM

PR #1548, opened 2024-08-20
`agama config load`/`store` for "software" uses the HTTP API

mvidner:
## Problem

**Patterns** part of the migration of the CLI from D-Bus API to HTTP API:

- https://trello.com/c/hvPtBtMD/3719-5-replace-d-bus-with-http-based-clients

## Solution

- Added `SoftwareHTTPClient`
- Kept (D-Bus) `SoftwareClient` because it serves as the backend for the above

## Testing

- Tested manually, via [`/testing_using_container.sh`](https://github.com/openSUSE/agama/blob/c65497c94ff1a1aa05d0f67a100e980c89e6ebfb/testing_using_container.sh)
- Added tests that mock the HTTP API

## Screenshots

No


rust/agama-lib/src/software/http_client.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::{base_http_client::BaseHTTPClient, error::ServiceError};
+use std::collections::HashMap;
+
+pub struct SoftwareHTTPClient {
+    client: BaseHTTPClient,
+}
+
+impl SoftwareHTTPClient {
+    pub fn new() -> Result<Self, ServiceError> {
+        Ok(Self {
```
jreidinger, line 11: you can use here `new_with_base` like `Ok(Self.new_with_base(BaseHTTPClient::new()?))`

rust/agama-lib/src/software/http_client.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn new_with_base(base: BaseHTTPClient) -> Self {
+        Self { client: base }
+    }
+
+    pub async fn get_config(&self) -> Result<SoftwareConfig, ServiceError> {
+        self.client.get("/software/config").await
+    }
+
+    pub async fn set_config(&self, config: &SoftwareConfig) -> Result<(), ServiceError> {
+        // FIXME: test how errors come out:
```
jreidinger, line 25: well, it can be improved by decoding body of error response and creating error from it, if it is also service error. I plan to do it, but other higher priorities come in the way.

jreidinger (approved):

rust/agama-lib/src/software/http_client.rs
mvidner: That's what I thought too, but 
1. in the end the code seeing these errors does no decisions, just shows them to the user
2. **if** we wanted to do some decisions, we would have to serialize them better than with `Display`/`Debug`

rust/agama-lib/src/software/http_client.rs
mvidner: I see your point... IMHO my way is still not enough duplication to need the indirection :)

PR #1549, opened 2024-08-21
feat(web): adapt to HTTP API of  DASD operations

jreidinger:
Trello: https://trello.com/c/2LC8JEZR/3743-8-agama-adapt-dasd-configuration-to-current-architecture

This PR is the following step to support for DASD devices on s390x architectures. It includes:

* adaptation of web API to changes done in https://github.com/openSUSE/agama/pull/1532


web/src/routes/storage.tsx
```diff
@@ -61,6 +62,11 @@ const routes = (): Route => ({
       element: <ISCSIPage />,
       handle: { name: N_("iSCSI") },
     },
+    {
```
jreidinger, line 65: add it only conditionally if dasd is supported ( will do it myself ).

web/src/types/dasd.ts
```diff
@@ -32,4 +32,11 @@ type DASDDevice = {
   hexId: number;
 };
 
-export type { DASDDevice };
+type FormatJob = {
+  job_id: string,
```
teclator, line 36: I guess it should be `id` or `jobId`

web/src/api/dasd.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Enables given list of devices
+ * @param devicesIDs array of device ids
+ */
+const DASDDisable = (devicesIDs: string[]) => post("/api/storage/dasd/disable", { devices: devicesIDs });
+
+/**
+ * Enables giag on given list of devices
+ * @param devicesIDs array of device ids
+ */
+const diagEnable = (devicesIDs: string[]) => put("/api/storage/dasd/diag", { devices: devicesIDs, diag: true });
```
imobachgs, line 63: I would suggest to use `enableDiag` to be consistent (e.g., `probeDASD`).

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+        data.push(formatJob);
+
+        queryClient.setQueryData(["dasd", "formatJobs", "running"], data);
+      }
+      if (
+        event.type === "JobChanged"
+      ) {
+        const { id, running } = event.job;
+        if (running) return;
+        let data = queryClient.getQueryData(["dasd", "formatJobs", "running"]) as FormatJob[];
```
imobachgs, line 111:
```suggestion
        const data = queryClient.getQueryData(["dasd", "formatJobs", "running"]) as FormatJob[];
```

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+  type SelectDASD = {
+    unselect?: boolean,
+    device?: DASDDevice,
+    devices?: DASDDevice[]
+  }
+
+  const queryClient = useQueryClient();
+
+  const mutation = useMutation({
+    mutationFn: async (data: SelectDASD): Promise<SelectDASD> => Promise.resolve(data),
```
imobachgs, line 161: You can remove the `async` word.

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+  });
+
+  return mutation;
+};
+/**
+ * Returns seleced DASD ids
+ */
+const selectedDASDQuery = () => ({
+  queryKey: ["dasd", "selected"],
+  queryFn: async () => {
```
imobachgs, line 139: Please, remove the word `async`.

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  return mutation;
+};
+
+/**
+ * Returns DASD filters
+ */
+const filterDASDQuery = () => ({
+  queryKey: ["dasd", "filter"],
+  queryFn: async () => {
```
imobachgs, line 182: Remove `async`.

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+  const { data } = useQuery(filterDASDQuery());
+
+  return data || { minChannel: "", maxChannel: "" };
+}
+
+const useFilterDASDChange = () => {
+  const queryClient = useQueryClient();
+
+  const mutation = useMutation({
+    mutationFn: async (data: FilterDASD): Promise<FilterDASD> => Promise.resolve(data),
```
imobachgs, line 198: Remove `async`.

web/src/queries/dasd.ts
```diff
@@ -0,0 +1,256 @@
...
+ */
+const useDASDDevicesChanges = () => {
+  const client = useInstallerClient();
+  const queryClient = useQueryClient();
+
+  React.useEffect(() => {
+    if (!client) return;
+
+    return client.ws().onEvent((event) => {
+      if (event.type === "DASDDeviceAdded") {
```
imobachgs, line 221: Why not using a `switch`?

web/src/routes/storage.tsx
imobachgs: How did you solve this?

web/src/types/dasd.ts
```diff
@@ -0,0 +1,51 @@
...
+ * find current contact information at www.suse.com.
+ */
+
+type DASDDevice = {
+  id: string;
+  enabled: boolean;
+  deviceName: string;
+  formatted: boolean;
+  diag: boolean;
+  status: string; // TODO: sync with rust when it switch to enum
```
imobachgs, line 28: What does it mean?

imobachgs (changes requested):
Good job! However, I have a bunch of comments.

web/src/api/dasd.ts
```diff
@@ -0,0 +1,72 @@
...
+ * probes DASD devices
+ */
+const probeDASD = () => post("/api/storage/dasd/probe");
+
+/**
+ * Start format job for given list of devices
+ * @param devicesIDs array of device ids
+ * @return id of format job
+ */
+const formatDASD = (devicesIDs: string[]): Promise<string> => post("/api/storage/dasd/format", { devices: devicesIDs }).then(({ data }) => data);
```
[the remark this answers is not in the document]
dgdavid: Plus find a way to automatically trigger the `npx prettier . -w` command to have a consistent code style :)

web/src/components/storage/DASDFormatProgress.tsx
```diff
@@ -0,0 +1,65 @@
...
+        </Stack>
+      );
+    }));
+
+  return (
+    <Popup title={_("Formatting DASD devices")} isOpen={runningJobs.length > 0} disableFocusTrap>
+      {runningJobs.map((job) => {
+        const progress = job.summary || {};
+
+        if (Object.keys(progress).length === 0) return;
```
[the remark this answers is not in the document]
dgdavid: > Instead of using an `if` block inside of the `map`, I would rather prefer combining a `filter` and a `map`).

NP: I see your point, but probably this was _my fault_. I tend to avoid iterating twice over a same collection. I know, premature optimization.

That said, I'm not against the change you proposed. Most probably, it will bring a more readable code, which is also important.

Thanks for catching it.

web/src/components/storage/DASDTable.tsx
```diff
@@ -144,36 +147,40 @@ const Actions = ({ devices, isDisabled }) => {
...
   const min = hex(from) || Math.min(...allChannels);
   const max = hex(to) || Math.max(...allChannels);
 
   return devices.filter((d) => d.hexId >= min && d.hexId <= max);
 };
 
-export default function DASDTable({ state, dispatch }) {
+export default function DASDTable() {
+  const devices = useDASDDevices();
+  const { mutate: changeFilter } = useFilterDASDChange();
```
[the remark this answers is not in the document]
dgdavid: :100: Totally agree with @imobachgs here.

web/src/components/storage/DASDTable.tsx
```diff
@@ -144,36 +147,40 @@ const Actions = ({ devices, isDisabled }) => {
...
-  const filteredDevices = filterDevices(state.devices, state.minChannel, state.maxChannel);
-  const selectedDevicesIds = state.selectedDevices.map((d) => d.id);
+  const filteredDevices = filterDevices(devices, minChannel, maxChannel);
+  const selectedDevicesIds = selectedDevices.map((d) => d.id);
 
   // Selecting
   const selectAll = (isSelecting = true) => {
-    const type = isSelecting ? "SELECT_ALL_DEVICES" : "UNSELECT_ALL_DEVICES";
-    dispatch({ type, payload: { devices: filteredDevices } });
+    changeSelected({ unselect: !isSelecting, devices: filteredDevices })
```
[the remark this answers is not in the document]
dgdavid: This is something we should talk about. If we actually desired this kind of behavior and the naming for these queries.

To keep such a behavior, I think we should bear in mind

* consistency, all the pages should behave in an similar way on that regard.
* usability, UI should provide a clear, usable way to both, let users know that there are filters applied and an easy to find reset / clear action
* naming, kind of DX, look for a naming strategy for these queries in order to quickly distinguish them from the API queries.

But please, go ahead with this change as it is. Just open, create a follow-up card (if needed)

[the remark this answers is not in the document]
imobachgs: Additionally, we might consider to use a single query for all the DASD filters.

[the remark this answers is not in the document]
dgdavid: > Additionally, we might consider using a single query for all the DASD filters.

Good idea, something to evaluate in next iteration I guess.

web/src/types/dasd.ts
teclator: @jreidinger ?

web/src/components/storage/DASDTable.tsx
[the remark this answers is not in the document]
dgdavid: > This is something we should talk about. If we actually desired this kind of behavior and the naming for these queries.
> 
> To keep such a behavior, I think we should bear in mind
> 
> * consistency, all the pages should behave in an similar way on that regard.
> * usability, UI should provide a clear, usable way to both, let users know that there are filters applied and an easy to find reset / clear action
> * naming, kind of DX, look for a naming strategy for these queries in order to quickly distinguish them from the API queries.
> 
> But please, go ahead with this change as it is. Just create a follow-up card (if needed)

Thinking about this twice, I believe now that would be better to

* Do not keep filters applied after navigating. It's kind of unexpected and it will not play well for all pages
* Rework this part before merging the PR. Don't worry, I can help here.

You may argue that it's a complex page and the user could benefit of landing in the page again with the previous filtering already applied. But we shouldn't compromise consistency among pages because a special use case. Instead, we should arrange time to work on improving the DASD page filters for offering easy to use yet powerful filtering options to _DASD users_.

So, I'll try to get rid of query for filtering here and use a stateful page approach instead.

dgdavid (changes requested):

web/src/components/storage/DASDFormatProgress.tsx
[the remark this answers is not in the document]
dgdavid: I was wrong in previous comment since I was thinking about this other `. map` 

https://github.com/openSUSE/agama/pull/1549/files#diff-73b1d201fc2f08a130c9630786cf958273bec4a1e072833919e0f3a541ff1456R87-R94

This one is returning a collection of React.Node or undefined, which is working because React is able to render nothing when `undefined` is given. It's ok since it is a way to ignore them in a _JSX iteration_ but could be done in the way you proposed too.

web/src/components/storage/DASDTable.tsx
```diff
@@ -37,15 +38,17 @@ import {
...
 import { Table, Thead, Tr, Th, Tbody, Td } from "@patternfly/react-table";
 import { Icon } from "~/components/layout";
-import { SectionSkeleton } from "~/components/core";
+import { CardField } from "~/components/core";
 import { _ } from "~/i18n";
 import { hex } from "~/utils";
 import { sort } from "fast-sort";
-import { useInstallerClient } from "~/context/installer";
+import { DASDDevice } from "~/types/dasd";
+import { disableDASD, disableDiag, enableDASD, enableDiag, formatDASD } from "~/api/dasd";
```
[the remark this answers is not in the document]
teclator: Ok, as we were not sending any change in the objects themselves at all but calling an action and relying in events for updating the state coming from the backend I preferred to not use mutations in this case, but can mix both, use a mutation sending the change we know about and receiving the rest from the backend...

web/src/components/storage/DASDFormatProgress.tsx
```diff
@@ -0,0 +1,67 @@
...
+            hasGutter
+            className="dasd-format-progress"
+          >
+            <ProgressContent key={`progress_content_${job.jobId}`} progress={job.summary} />
+          </Stack>
+        );
+      })}
+    </Popup>
+  );
+}
```
dgdavid, line 67: @teclator 

Sorry for the NP here. I've no way of testing below diff in which I have tried to improve just a bit this component by extracting the progress to an internal component outside of the component itself and ending up with just one `<Stack>`.

```diff
diff --git a/web/src/components/storage/DASDFormatProgress.tsx b/web/src/components/storage/DASDFormatProgress.tsx
index 61f2b4e6..bee2c2ac 100644
--- a/web/src/components/storage/DASDFormatProgress.tsx
+++ b/web/src/components/storage/DASDFormatProgress.tsx
@@ -24,7 +24,18 @@ import { Progress, Stack } from "@patternfly/react-core";
 import { Popup } from "~/components/core";
 import { _ } from "~/i18n";
 import { useDASDDevices, useDASDRunningFormatJobs } from "~/queries/dasd";
-import { FormatSummary } from "~/types/dasd";
+import { DASDDevice, FormatSummary } from "~/types/dasd";
+
+const DeviceProgress = ({ device, progress }: { device: DASDDevice, progress: FormatSummary }) => (
+  <Progress
+    size="sm"
+    max={progress.total}
+    value={progress.step}
+    title={`${device.id} - ${device.deviceName}`}
+    measureLocation="none"
+    variant={progress.done ? "success" : undefined}
+  />
+);
 
 export default function DASDFormatProgress() {
   const devices = useDASDDevices();
@@ -32,36 +43,19 @@ export default function DASDFormatProgress() {
     (job) => Object.keys(job.summary || {}).length > 0,
   );
 
-  const ProgressContent = ({ progress }: { progress: { [key: string]: FormatSummary } }) =>
-    Object.entries(progress).map(([id, { total, step, done }]) => {
-      const device = devices.find((d) => d.id === id);
-
-      return (
-        <Progress
-          key={`progress_${id}`}
-          size="sm"
-          max={total}
-          value={step}
-          title={`${device.id} - ${device.deviceName}`}
-          measureLocation="none"
-          variant={done ? "success" : undefined}
-        />
-      );
-    });
-
   return (
     <Popup title={_("Formatting DASD devices")} isOpen={runningJobs.length > 0} disableFocusTrap>
-      {runningJobs.map((job) => {
-        return (
-          <Stack
-            key={`formatting_progress_${job.jobId}`}
-            hasGutter
-            className="dasd-format-progress"
-          >
-            <ProgressContent key={`progress_content_${job.jobId}`} progress={job.summary} />
-          </Stack>
-        );
-      })}
+      <Stack
+        hasGutter
+        className="dasd-format-progress"
+      >
+        {runningJobs.flatMap((job) => (
+          Object.entries(job.summary).map(([id, progress]) => {
+            const device = devices.find((d) => d.id === id);
+            return <DeviceProgress key={`${id}-format-progress`} device={device} progress={progress} />;
+          }))
+        )}
+      </Stack>
     </Popup>
   );
 }
```

<details>
<summary>Click to show hide the whole file content</summary>

---

```tsx
/*
 * Copyright (c) [2023] SUSE LLC
 *
 * All Rights Reserved.
 *
 * This program is free software; you can redistribute it and/or modify it
 * under the terms of version 2 of the GNU General Public License as published
 * by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
 * more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program; if not, contact SUSE LLC.
 *
 * To contact SUSE LLC about this file by physical or electronic mail, you may
 */

import React from "react";
import DASDTable from "~/components/storage/DASDTable";
import DASDFormatProgress from "~/components/storage/DASDFormatProgress";
import { _ } from "~/i18n";
import { Page } from "~/components/core";
import { useDASDDevicesChanges, useDASDFormatJobChanges } from "~/queries/dasd";

export default function DASDPage() {
  useDASDDevicesChanges();
  useDASDFormatJobChanges();

  return (
    <Page>
      <Page.Header>
        <h2>{_("DASD")}</h2>
      </Page.Header>

      <Page.MainContent>
        <DASDTable />
        <DASDFormatProgress />
      </Page.MainContent>
    </Page>
  );
}
```
</details>

web/src/components/storage/DASDTable.tsx
```diff
@@ -144,36 +153,40 @@ const Actions = ({ devices, isDisabled }) => {
...
 
   return devices.filter((d) => d.hexId >= min && d.hexId <= max);
 };
 
-export default function DASDTable({ state, dispatch }) {
+export default function DASDTable() {
+  const devices = useDASDDevices();
+  const { mutate: changeFilter } = useFilterDASDMutation();
+  const { mutate: changeSelected } = useSelectedDASDChange();
+  const { minChannel, maxChannel } = useFilterDASD();
```
dgdavid, line 168: @teclator 

I know you're gonna taking care of stop using queries for filtering. But just in case it can help, below you have a diff removing them. Not sending a commit to avoid conflicts with changes you're doing.


```diff
diff --git a/web/src/components/storage/DASDTable.tsx b/web/src/components/storage/DASDTable.tsx
index 80beae41..804ac279 100644
--- a/web/src/components/storage/DASDTable.tsx
+++ b/web/src/components/storage/DASDTable.tsx
@@ -19,7 +19,7 @@
  * find current contact information at www.suse.com.
  */
 
-import React, { Dispatch, SetStateAction, useState } from "react";
+import React, { useState } from "react";
 import {
   Button,
   CardBody,
@@ -46,8 +46,6 @@ import { DASDDevice } from "~/types/dasd";
 import { disableDASD, disableDiag, enableDASD, enableDiag, formatDASD } from "~/api/dasd";
 import {
   useDASDDevices,
-  useFilterDASD,
-  useFilterDASDMutation,
   useSelectedDASD,
   useSelectedDASDChange,
 } from "~/queries/dasd";
@@ -161,13 +159,17 @@ const filterDevices = (devices: DASDDevice[], from: string, to: string): DASDDev
   return devices.filter((d) => d.hexId >= min && d.hexId <= max);
 };
 
+type FilterOptions = {
+  minChannel?: string;
+  maxChannel?: string;
+}
+
 export default function DASDTable() {
   const devices = useDASDDevices();
-  const { mutate: changeFilter } = useFilterDASDMutation();
   const { mutate: changeSelected } = useSelectedDASDChange();
-  const { minChannel, maxChannel } = useFilterDASD();
   const selectedDevices = useSelectedDASD();
 
+  const [{ minChannel, maxChannel }, setFilters] = useState<FilterOptions>({ minChannel: "", maxChannel: "" });
   const [sortingColumn, setSortingColumn] = useState(columns[0]);
   const [sortDirection, setSortDirection] = useState<"asc" | "desc">("asc");
 
@@ -201,22 +203,9 @@ export default function DASDTable() {
     };
   };
 
-  // Filtering
-  const onMinChannelFilterChange = (_event, value) => {
-    changeFilter({ minChannel: value });
-  };
-
-  const onMaxChannelFilterChange = (_event, value) => {
-    changeFilter({ maxChannel: value });
-  };
-
-  const removeMinChannelFilter = () => {
-    changeFilter({ minChannel: "" });
-  };
-
-  const removeMaxChannelFilter = () => {
-    changeFilter({ maxChannel: "" });
-  };
+  const updateFilter = (newFilters: FilterOptions) => {
+    setFilters((currentFilters) => ({ ...currentFilters, ...newFilters }));
+  }
 
   const Content = () => {
     return (
@@ -274,14 +263,14 @@ export default function DASDTable() {
                     type="text"
                     aria-label={_("Filter by min channel")}
                     placeholder={_("Filter by min channel")}
-                    onChange={onMinChannelFilterChange}
+                    onChange={(_, minChannel) => updateFilter({ minChannel })}
                   />
                   {minChannel !== "" && (
                     <TextInputGroupUtilities>
                       <Button
                         variant="plain"
                         aria-label={_("Remove min channel filter")}
-                        onClick={removeMinChannelFilter}
+                        onClick={() => updateFilter({ minChannel: "" })}
                       >
                         <Icon name="backspace" size="s" />
                       </Button>
@@ -296,14 +285,14 @@ export default function DASDTable() {
                     type="text"
                     aria-label={_("Filter by max channel")}
                     placeholder={_("Filter by max channel")}
-                    onChange={onMaxChannelFilterChange}
+                    onChange={(_, maxChannel) => updateFilter({ maxChannel })}
                   />
                   {maxChannel !== "" && (
                     <TextInputGroupUtilities>
                       <Button
                         variant="plain"
                         aria-label={_("Remove max channel filter")}
-                        onClick={removeMaxChannelFilter}
+                        onClick={() => updateFilter({ maxChannel: "" })}
                       >
                         <Icon name="backspace" size="s" />
                       </Button>
diff --git a/web/src/queries/dasd.ts b/web/src/queries/dasd.ts
index f629fdc2..bd07a995 100644
--- a/web/src/queries/dasd.ts
+++ b/web/src/queries/dasd.ts
@@ -171,39 +171,6 @@ const useSelectedDASDChange = () => {
   return mutation;
 };
 
-/**
- * Returns DASD filters
- */
-const filterDASDQuery = () => ({
-  queryKey: ["dasd", "filter"],
-  queryFn: () => {
-    return Promise.resolve({ minChannel: "", maxChannel: "" });
-  },
-  staleTime: Infinity,
-});
-
-const useFilterDASD = (): FilterDASD => {
-  const { data } = useQuery(filterDASDQuery());
-
-  return data || { minChannel: "", maxChannel: "" };
-}
-
-const useFilterDASDMutation = () => {
-  const queryClient = useQueryClient();
-
-  const mutation = useMutation({
-    mutationFn: (data: FilterDASD): Promise<FilterDASD> => Promise.resolve(data),
-    onSuccess: (data: FilterDASD) => {
-      queryClient.setQueryData(["dasd", "filter"], (prev: FilterDASD) => ({
-        ...prev,
-        ...data
-      }));
-    },
-  });
-
-  return mutation;
-};
-
 /**
  * Listens for DASD devices changes.
  */
@@ -255,6 +222,6 @@ const useDASDDevicesChanges = () => {
 };
 
 export {
-  useDASDDevices, useDASDDevicesChanges, useFilterDASDMutation, filterDASDQuery, useFilterDASD, useSelectedDASD, useSelectedDASDChange, selectedDASDQuery,
+  useDASDDevices, useDASDDevicesChanges, useSelectedDASD, useSelectedDASDChange, selectedDASDQuery,
   useDASDFormatJobChanges, useDASDRunningFormatJobs, useFormatDASDMutation
 };
```

web/src/routes/storage.tsx
jreidinger: I thought that @teclator did it with that new supported call that should do it.

web/src/types/dasd.ts
jreidinger: it is related to FIXME in rust code that wants to use enums for some data like `status`. So when it rust switch to enum, we should also change it here.

web/src/queries/dasd.ts
```diff
@@ -254,7 +245,96 @@ const useDASDDevicesChanges = () => {
...
+
+const useFormatDASDMutation = () => {
+  const queryClient = useQueryClient();
+  const query = {
+    mutationFn: formatDASD,
+    onSuccess: (data: string, deviceIds: string[]) => {
+      queryClient.setQueryData(["dasd", "formatJob", data], { jobId: data });
+      queryClient.setQueryData(["dasd", "devices"], (prev: DASDDevice[]) => {
+        const nextData = prev.map((dev) => {
+          if (deviceIds.includes(dev.id)) dev.formatted = false;
```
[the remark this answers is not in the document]
teclator: I was not sure about this change at all and if modify the devices, them should be updated when finished the format progress, probably we could omit any change at all ...

web/src/routes/storage.tsx
teclator: We only added it in the selector, showing the option in case that supported but not in the routes, then you commented it here, but we can leave it as it is as the button for navigating to the route will not be shown although modifying the url will allow to navigate there even if not supported

web/src/queries/dasd.ts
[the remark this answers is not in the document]
imobachgs: Yes, you are right, that code is wrong. You should avoid modifying the original objects. I guess you could something like:

```javascript
prev.map((dev) => {
  return {
    ...dev,
    formatted: deviceIds.includes(dev.id) ? false : dev.formatted
  };
});
```

Or just:

```javascript
prev.map((dev) => {
  const newDev = {...dev};
  if (deviceIds.includes(dev.id)) newDev.formatted = false;
  return newDev;
});
```

I have not tested any of them, but you get the idea.

web/src/components/storage/DASDTable.tsx
```diff
@@ -88,6 +92,11 @@ const columns = [
 ];
 
 const Actions = ({ devices, isDisabled }: { devices: DASDDevice[]; isDisabled: boolean }) => {
+  const { mutate: enableDASD } = useEnableDASDMutation();
```
imobachgs, line 91: I think using a parameter (instead of having a hook for enabling and another for disabling) might be a better idea and would help to reduce the amount of code.

web/src/components/storage/DASDTable.tsx
```diff
@@ -210,21 +215,8 @@ export default function DASDTable() {
...
-  };
-
-  const removeMinChannelFilter = () => {
-    changeFilter({ minChannel: "" });
-  };
-
-  const removeMaxChannelFilter = () => {
-    changeFilter({ maxChannel: "" });
+  const updateFilter = (newFilters: FilterOptions) => {
+    setFilters((currentFilters) => ({ ...currentFilters, ...newFilters }));
```
imobachgs, line 221: :+1:

imobachgs (changes requested):
Please, check the new mutations.

web/src/routes/storage.tsx
teclator: Done by https://github.com/openSUSE/agama/pull/1549/commits/a75354f1ddb8f9eff6a54689e75f69d15d9860fa

web/src/components/storage/DASDTable.tsx
teclator: Ok, will address it.

web/src/queries/dasd.ts
```diff
@@ -135,61 +138,28 @@ const useSelectedDASD = (): DASDDevice[] => {
...
-}
+};
 
 const useSelectedDASDChange = () => {
   type SelectDASD = {
-    unselect?: boolean,
-    device?: DASDDevice,
-    devices?: DASDDevice[]
-  }
+    unselect?: boolean;
```
dgdavid: NP: if I'm not wrong, there is somewhere an identical type called SelectionOptions.

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Sep  4 21:00:34 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Bring back DASD management support (gh#openSUSE/1549).
+
```
dgdavid, line 5: Wrong reference because a missing word,

`gh#openSUSE/agama#1549`

dgdavid (commented):
It looks good to me, but pease check the log entry.

Great job 👍 👍 👍  BTW

web/package/agama-web-ui.changes
dgdavid: A bit NP here, bit please check the PR description before merging too. Is the warning still relevant?

imobachgs (approved):

web/src/routes/storage.tsx
```diff
@@ -38,46 +39,42 @@ const PATHS = {
...
+      path: PATHS.iscsi,
+      element: <ISCSIPage />,
+      handle: { name: N_("iSCSI") },
+    },
+    {
+      path: PATHS.dasd,
+      element: <DASDPage />,
+      handle: { name: N_("DASD") },
+      loader: async () => {
+        if (!DASDSupported()) return redirect(PATHS.root);
```
dgdavid, line 72: Not sure if worth mentioning that loader will not be executed until loading the route. Probably not.

Also, we should add a mechanism to avoid the application crashing because trying to go to the route before login **but in another round of routes improvements, not now** :wink:

BTW, thanks a lot for taking the time of doing manual testing of this part and catching such a bug.

teclator: I can improve the message or add a note in the ZFCP PR just for not delay this one.

PR #1550, opened 2024-08-21
feature(web): make the product selection more appealing and accessible

### LINKED ISSUE #1415: Icons for individual distros
it would be nice to have product/distribution icons in the product selection screen.
I'd like to work on this one.

lkocman:
* Fixes #1415
* Supply icon for all products and copy them to _dist/assets/logos_ in final web build.
* Add icon to structures, pass it to dbus, etc (please do not mix product icon with pattern icon) fallback to default.svg in case that no icon is set

|Before | After|
|-|-|
|Screenshot of the previous product selection page|Screenshot of the new product selection page|

---

For more details please read https://github.com/openSUSE/agama/pull/1550#pullrequestreview-2283239736 (and maybe https://github.com/openSUSE/agama/pull/1550#pullrequestreview-2269021089)


lkocman:
The entire reason not to use radio buttons, is that images simply imply that they're clickable and that would lead to messed up user experience. Now the entire card is clickable, and confusion is gone.

lkocman:
Padding bottom fixes the wrap of the text. However after spending minutes on the screen I also removed duplicite product mentining on the beginning of the sentence since product name is already in bold font one line above. This feels much better.

service/lib/agama/software/product.rb
```diff
@@ -48,6 +48,13 @@ class Product
       # @return [String, nil] E.g., "1.0".
       attr_accessor :version
 
+      # Product icon. Please use specify filename with svg suffix and ensure referenced
+      # file exists inside agama/web/src/assests/product.
```
jreidinger, line 52: typo `assests`

service/lib/agama/software/product.rb
```diff
@@ -48,6 +48,13 @@ class Product
...
+      # default.svg unless be used nless specified otherwise.
+      #
+      # @return [String, "default.svg"] E.g., "1.0".
```
jreidinger, line 55:
```suggestion
      # @return [String] E.g. "leap.svg"
```

service/lib/agama/software/product.rb
```diff
@@ -48,6 +48,13 @@ class Product
       # @return [String, nil] E.g., "1.0".
       attr_accessor :version
 
+      # Product icon. Please use specify filename with svg suffix and ensure referenced
+      # file exists inside agama/web/src/assests/product.
+      # default.svg unless be used nless specified otherwise.
```
jreidinger, line 53:
```suggestion
      # `default.svg` will be used unless specified otherwise.
```

service/lib/agama/software/product_builder.rb
```diff
@@ -64,6 +64,7 @@ def initialize_product(id, data, attrs)
           product.description = attrs["description"]
           product.name = data[:name]
           product.version = data[:version]
+          product.icon = attrs["icon"] if attrs.key?("icon") && !attrs["icon"].nil?
```
jreidinger, line 67: default value for missing key in ruby is nil so check for nil is enough...and also nil is evaluated to false so you can use something like
```suggestion
          product.icon = attrs["icon"] if attrs["icon"]
```

another option is to define default icon as constant like `DEFAULT_ICON` and then use here
```suggestion
          product.icon = attrs["icon"] || DEFAULT_ICON
```

web/src/components/core/ChangeProductLink.test.tsx
```diff
@@ -30,11 +30,13 @@ const tumbleweedProduct = {
   id: "Tumbleweed",
   name: "openSUSE Tumbleweed",
   description: "Tumbleweed description...",
+  icon: "Tumbleweed.icon",
```
jreidinger, line 33: icon suffix?

web/src/components/product/ProductSelectionPage.jsx
```diff
@@ -46,34 +47,57 @@ function ProductSelectionPage() {
     }
   };
 
-  const Item = ({ children }) => {
+  const Item = ({ children }) => (
+    <GridItem sm={10} smOffset={1} lg={8} lgOffset={2} xl={6} xlOffset={3}>
```
jreidinger, line 51: I expect this padding in CSS, but sadly I am not css expert. Lets hope @dgdavid gets back soon

web/src/components/product/ProductSelectionPage.jsx
```diff
@@ -46,34 +47,57 @@ function ProductSelectionPage() {
...
+
     return (
-      <GridItem sm={10} smOffset={1} lg={8} lgOffset={2} xl={6} xlOffset={3}>
-        {children}
-      </GridItem>
+      <img
+        src={productIcon}
+        alt={alt}
+        width="80px"
+        style={{ height: 'auto', width: '10%', float: 'left', padding: '0 20px 20px 0' }}
```
jreidinger, line 65: same here. I expect icon style to be in css

web/src/components/product/ProductSelectionPage.jsx
```diff
@@ -46,34 +47,57 @@ function ProductSelectionPage() {
...
             {products.map((product, index) => (
               <Item key={index}>
-                <Card key={index} isRounded>
+                <Card
+                  key={index}
+                  isRounded
+                  onClick={() => handleCardClick(product)}
+                  style={{
+                    cursor: 'pointer',  // Change the cursor to indicate clickable
+                    border: nextProduct === product ? '2px solid #51c38d' : 'none',  // highlight selected card
```
jreidinger, line 89: also here I expect some class and style defined in css.

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Aug 21 19:07:28 UTC 2024 - Lubos Kocman <lubos.kocman@suse.com>
+
+- Add product icons. Resolves Issue #1415
```
jreidinger, line 4: note to all changes file. We use OBS style for github issues ( so OBS can link it properly ). So it should be `gh#openSUSE/agama#1415`

jreidinger:
@dgdavid after consulting with patternfly documentation, I would say that for product selection we maybe can use Tile component, what do you think? https://www.patternfly.org/components/tile#with-long-subtext

lkocman:
@jreidinger I was originally playing with the Branding component which would be a nice choice for a wide image (e.g. with Text) on top of the card, but above description instead of aside.

web/src/components/product/ProductSelectionPage.jsx
```diff
@@ -46,34 +47,57 @@ function ProductSelectionPage() {
...
-  const Item = ({ children }) => {
+  const Item = ({ children }) => (
+    <GridItem sm={10} smOffset={1} lg={8} lgOffset={2} xl={6} xlOffset={3}>
+      {children}
+    </GridItem>
+  );
+
+  const ProductIcon = ({ src, alt }) => {
+    // Ensure that we display something even if icon path is incorrect
+    const productIcon = require(`../../assets/products/${src}`);
```
joseivanlopez, line 58:
```suggestion
    const productIcon = require(`~/assets/products/${src}`);
```

web/src/components/core/ChangeProductLink.test.tsx
lkocman: I'm not in favor of four letter suffixes, plus the format is svg.

lkocman:
@joseivanlopez hmm ~assets/ doesn't get properly resolved. Shall I revert it back?
```
<w> [ReactRefreshPlugin] Hot Module Replacement (HMR) is not enabled! React Refresh requires HMR to function properly.
ERROR in ./src/components/product/ProductSelectionPage.jsx 107:22-63
Module not found: Error: Can't resolve '~/assets/products' in '/home/runner/work/agama/agama/web/src/components/product'

webpack compiled with 1 error
Error: Process completed with exit code 1.
```

dgdavid (commented):
Hi @lkocman!

First of all, thanks a ton for such a great contribution! :heart: It's a pleasure to see more people approaching Agama development :)

I have, however, some doubts about merging the proposal as it is. It looks pretty nice and I guess you had put a lot of effort into it, but I'm a bit concerned about introducing aesthetic changes that might defeat the a11y more than it is already. We are aware that Agama a11y is not in good shape at this moment and we really want to prioritize it, but somehow we are still in the middle of the UI migration we started a few weeks before the openSUSE Conference back in May.

Having that in mind, I believe it is not a good idea to remove the radio buttons from the selector since they are natively accessible by the keyboard which is not the case in your proposal. Of course, we could work on your changes to make them keyboard-friendly, but not long ago we decided to avoid cooking complex components because of a bunch of reasons, available resources and productivity among them. 

Another reason to keep the radio buttons is consistency. We would like to make all selectors look as similar as possible, bearing in mind that there will be subtle differences in special cases. For me, that means that removing radios from a selector implies removing radios from all of them, although I could, of course, be wrong.

Last but not least, think that radio buttons are widely known and quickly recognisable by users. I'd say there is almost zero cognitive effort to know you're in front of a selector when native components are there. Also, it's faster to distinguish between a single or multiple selector when using them (radios vs checkboxes).

So, I'd go for a combination of your changes and the previous state in order to keep the best of two approaches.

---

Screenshot of a product selector proposal

Screenshot of a product selector proposal with lower resolution and displaying the code in the developer tool


<details>
<summary>Click to show/hide a diff</summary>

---

```diff
diff --git a/web/src/assets/styles/app.scss b/web/src/assets/styles/app.scss
index c8dfab14..509cba4a 100644
--- a/web/src/assets/styles/app.scss
+++ b/web/src/assets/styles/app.scss
@@ -35,3 +35,11 @@ button.remove-link:hover {
   position: relative;
   width: 100%;
 }
+
+#productSelectionForm {
+  .pf-v5-c-radio input {
+    align-self: center;
+    width: 20px;
+    height: 20px;
+  }
+}
diff --git a/web/src/components/product/ProductSelectionPage.jsx b/web/src/components/product/ProductSelectionPage.jsx
index a15f9326..a8241e26 100644
--- a/web/src/components/product/ProductSelectionPage.jsx
+++ b/web/src/components/product/ProductSelectionPage.jsx
@@ -21,7 +21,7 @@
  */
 
 import React, { useState } from "react";
-import { Card, CardBody, Flex, Form, Grid, GridItem } from "@patternfly/react-core";
+import { Card, CardBody, Flex, Form, Grid, GridItem, Radio, Split, Stack } from "@patternfly/react-core";
 import { Page } from "~/components/core";
 import { Center } from "~/components/layout";
 import { useConfigMutation, useProduct } from "~/queries/software";
@@ -54,6 +54,7 @@ function ProductSelectionPage() {
   );
 
   const ProductIcon = ({ src, alt }) => {
+    console.log("SRC > ", src);
     // Ensure that we display something even if icon path is incorrect
     const productIcon = require(`../../assets/products/${src}`);
 
@@ -62,7 +63,6 @@ function ProductSelectionPage() {
         src={productIcon}
         alt={alt}
         width="80px"
-        style={{ height: 'auto', width: '10%', float: 'left', padding: '0 20px 20px 0' }}
       />
     );
   };
@@ -78,26 +78,33 @@ function ProductSelectionPage() {
       <Center>
         <Form id="productSelectionForm" onSubmit={onSubmit}>
           <Grid hasGutter>
             {products.map((product, index) => (
              <Item key={index}>
                 <Card
                   key={index}
                   isRounded
-                  onClick={() => handleCardClick(product)}
-                  style={{
-                    cursor: 'pointer',  // Change the cursor to indicate clickable
-                    border: nextProduct === product ? '2px solid #51c38d' : 'none',  // highlight selected card
-                  }}
                 >
                   <CardBody>
-                    <ProductIcon
-                      src={product.icon}
-                      alt={`${product.name} product icon`}
+                    <Radio
+                      key={index}
+                      name="product"
+                      id={product.name}
+                      label={
+                        <Split hasGutter>
+                          <ProductIcon
+                            src={product.icon}
+                            alt={`${product.name} product icon`}
+                          />
+                          <Stack hasGutter>
+                            <Label>{product.name}</Label>
+                            <p>{product.description}</p>
+                          </Stack>
+                        </Split>
+                      }
+                      isChecked={nextProduct === product}
+                      onChange={() => setNextProduct(product)}
                     />
-                    <div>
-                      <Label>{product.name}</Label>
-                      <p>{product.description}</p>
-                    </div>
                   </CardBody>
                 </Card>
               </Item>
@@ -118,8 +125,8 @@ function ProductSelectionPage() {
           </Grid>
         </Form>
       </Center>
-    </Page>
+    </Page >
   );
 }
 
-export default ProductSelectionPage;
\ No newline at end of file
+export default ProductSelectionPage;
```

</details>

---

That way, we can add or not the thin border for the selected option, although I prefer postponing it for a while. Please, don't get me wrong. I'm not against aesthetic improvements. I like them (a lot!) but IMHO we should think a bit more about these changes, their impact and how to make them as consistent as possible (i.e., adding the same visual aid to other selectors). Postponing them for the time Agama supports themes would be a good compromise from my point of view. But let's see how the rest of the team sees it.

BTW, manually testing your changes I've found an a11y bug on the master branch: the focus is lost because the entire selector is redrawn every time the selection changes. A bit technical, but this happens because it is a controlled form in order of being able to know if the selection matches the already selected product. This, along with a condition you used for adding a border or not depending on the same condition, made me think that it does not make sense to have the selected product as part of the selector. Of course, it should be on the selection page, but not as an option, although this is for another PR :wink:

dgdavid:
> to show/hide a not intended to be final diff

Of course, the `diff` added at https://github.com/openSUSE/agama/pull/1550#pullrequestreview-2269021089 is just a rough example of what we can do. Still needing some polish. But don't worry, we can take care of it if you agree.

lkocman:
@dgdavid absolutely feel free to! I can help in spare time too if needed.

lkocman:
I do like the centering @dgdavid!

lkocman:
@dgdavid oky then

dgdavid:
> BTW, manually testing your changes I've found an a11y bug on the master branch: the focus is lost because the entire selector is redrawn every time the selection changes. A bit technical, but this happens because it is a controlled form in order of being able to know if the selection matches the already selected product. This, along with a condition you used for adding a border or not depending on the same condition, made me think that it does not make sense to have the selected product as part of the selector. Of course, it should be on the selection page, but not as an option, although this is for another PR 😉

As said, not related to this PR but I took the opportunity and fixed it as part of it :) See https://github.com/openSUSE/agama/pull/1550/commits/b95eff61c80308c6dda21269fe612fc872df9821

Now users can select the product using the keyboard too, but still thinking that selected product should not be part of the offered options. A better layout is needed to inform the user not only about the currently selected product but also about the consequences of changing it... but, now for real, in another PR :smiley:

dgdavid:
> @dgdavid after consulting with patternfly documentation, I would say that for product selection we maybe can use Tile component, what do you think? https://www.patternfly.org/components/tile#with-long-subtext

Sorry @jreidinger, I had overlooked this comment.

No please, do not use PF/Tile... it will be deprecated in favor of PF/Card, see https://github.com/patternfly/patternfly-react/issues/8542. I mentioned it a couple of months ago when working in storage part, see https://github.com/openSUSE/agama/commit/8d432fb47ab1cfce05765aeca8f2fc069dff202b, which is also no longer avialable.

The idea, as commented in team meetings, is to keep component as simple as possible, closer to HTML the better. At least by now, since there a lot of things to do in a limited period of time :) In any case, the whole UI will be benefit from using simple, yet understandable elements or components.

Thanks a lot for having a look anyway. You rocks!

dgdavid (commented):
@lkocman, @jreidinger,

Here I go :roll_eyes: 

I've reworked the proposed changes for having these logos in the UI but keeping it as much accessible as possible with our (yet limited) knowledge to that regard.

But before entering in details, let me thanks @lkocman again for the contribution since it has been not only an step forward for having the product logos in the selection page, but hopefully a turning point regarding to a11y in Agama.

As you can see in the changes, I've removed the `onClick` prop from the PF/Card. If you go commit by commit, you'll see that I restored it at some point, but a11y Firefox tool complained about  https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#clickable_elements_must_be_focusable_and_should_have_interactive_semantics.

Firefox a11y tools complaining

Thus, I've used a CSS technique for expanding the area of the label to the whole card in addition to force the pointer cursor for it in this case (by default, labels are clickable but use the default cursor). That way, what @lkocman  was looking for is achievable without defeating the keyboard usability (I believe).

You may wonder why I didn't use the PF/Card with selectable and clickable properties. I simply did not have as much as PF knowledge as I'd like and found it rather complex at this moment.

I also tried to follow and accomplish the PF/Card a11y testing check list, https://www.patternfly.org/components/card/accessibility#testing

Since product descriptions are usually long, I have marked them as `aria-details` in order to prevent the screen reader reading the whole text before announcing the "not selected radio button" or "selected radio button" state, which would be a bit annoying IMHO. What is more, from my POV these descriptions perfectly fits as details.

Last but not least, you will see that I have added the "Select a product" label (actually a _legend_) there. I know it's tempting to remove it, but users need a context. Not you, not me, not even most of the guys we know... but there are users that need such a context. I'm saying that because it's true that an `aria` attribute will be enough for screen reader users, but I really believe that such a selector needs a bit of context when a user lands there. In other words, IMHO it is not superfluous but contributes.

You can check the visual result by yourself. And play the below videos to see how keyboard and screen reader are supported.


https://github.com/user-attachments/assets/ab1b6184-2aaf-404d-8abd-54fac863eec5


https://github.com/user-attachments/assets/9ca6696f-26dd-4040-a8e6-1407ea18ebc7



Please, look at it as just an step forward. Any other change or improvement can be properly addressed in a next iteration in order to not end up with an endless PR.

Thanks a lot!

service/lib/agama/software/product.rb
imobachgs, line 52:
```suggestion
      # file exists inside agama/web/src/assets/product.
```

service/package/rubygem-agama-yast.changes
```diff
@@ -27,6 +27,12 @@ Mon Aug 26 10:01:27 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 - Do not depend on f2fs-tools and nilfs-utils
   (jsc#PED-8669, gh#openSUSE/agama#1554).
 
+-------------------------------------------------------------------
+Wed Aug 21 19:07:28 UTC 2024 - Lubos Kocman <lubos.kocman@suse.com>
```
imobachgs, line 31: Leftover?

web/src/components/core/ChangeProductLink.test.tsx
imobachgs, line 33: why not simply use `svg` as example?

imobachgs (approved):

imobachgs (approved):

PR #1551, opened 2024-08-22
New Y2Storage proposal for Agama

ancorgs:
This is a result of wrong merges


ancorgs:
Wrong cherry-picks or something. To be ignored.

PR #1552, opened 2024-08-22
New Y2Storage proposal for Agama

ancorgs:
In order to implement the unattended behavior for storage described at [auto_storage.md](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md), we decided we need a new kind of proposal since we want to achieve things that would be impossible with `Y2Storage::GuidedProposal` or `Y2Storage::AutoinstProposal`.

This introduces a new `Y2Storage::AgamaProposal` that combines components from the mentioned existing proposals.

Replaces #1448


joseivanlopez (approved):
LGTM

joseivanlopez (changes requested):
Only a changelog entry is missing.

joseivanlopez (approved):

PR #1556, opened 2024-08-27
chore(rust): apply Clippy suggestions

imobachgs:
* Apply a bunch of Clippy suggestions.
* Update Rust minimal version to 1.75 because [IpAddr::to_canonical](https://doc.rust-lang.org/std/net/enum.IpAddr.html#method.to_canonical) is stable since then.


teclator (approved):
LGTM

mvidner (approved):

PR #1557, opened 2024-08-27
fix(service): add dbus-1-daemon as a dependency

imobachgs:
`dbus-daemon` is no longer included in the ISO or the containers. This is most probably related to the introduction of `dbus-broker` which caused some dependencies to change. We might need to do some further investigation, but this is a hotfix to have the ISO working again.

### DIFF
--- a/service/package/gem2rpm.yml
+++ b/service/package/gem2rpm.yml
@@ -26,6 +26,7 @@
     Provides:       agama-yast
     BuildRequires:  dbus-1-common
     Requires:       dbus-1-common
+    Requires:       dbus-1-daemon
     Requires:       suseconnect-ruby-bindings
     # YaST dependencies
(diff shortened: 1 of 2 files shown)


dgdavid (approved):

service/package/gem2rpm.yml
```diff
@@ -26,6 +26,7 @@
     Provides:       agama-yast
     BuildRequires:  dbus-1-common
     Requires:       dbus-1-common
+    Requires:       dbus-1-daemon
```
jreidinger, line 29: do we need it also for rust dbus server or deps live only here?

imobachgs: I decided to put it only there because the rust package depends on agama-yast already.

jreidinger (approved):

Issue #1561, opened 2024-08-29
Relying on dbus-daemon which is replaced by dbus-broker in Tumbleweed


mvidner:
Agama CI tests started failing, being unable to start `dbus-daemon`.

By default it is replaced by `dbus-broker`. [Tumbleweed snapshot 20240825 announcement][tw0825] quotes dbus-1.changes:

> - No longer start or offer starting dbus as a system service
>   dbus-broker will be the only supported system dbus. Although
>   the existing daemon will stay as some things (gdm) require
>   dbus-run-session

[tw0825]: https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/GUIJNW45DGP25M7RS4EPAKTEP7IXEHWU/

We need to make it work again:
1. install dbus-1-daemon.rpm as a workaround
2. figure out the setup for dbus-broker. For example the daemon  configuration file does not apply

mvidner:
I have just checked that https://github.com/yast/yast-storage-ng passes CI just fine with `dbus-broker` so it seems changing https://github.com/yast/ci-ruby-container won't be needed

mvidner:
Ah, the package part (rubygem-agama-yast.rpm) was done by Imo in #1557

PR #1562, opened 2024-08-29
Install dbus-1-daemon explicitly since Tumbleweed now uses dbus-broker (#1561)

mvidner:
## Problem

- Filed as issue #1561
[Tumbleweed snapshot 20240825 announcement][tw0825] quotes dbus-1.changes:

> - No longer start or offer starting dbus as a system service
>   dbus-broker will be the only supported system dbus. Although
>   the existing daemon will stay as some things (gdm) require
>   dbus-run-session

[tw0825]: https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/GUIJNW45DGP25M7RS4EPAKTEP7IXEHWU/

dbus-broker is not a drop-in replacement:
for example it does not share the config file, so eventual migration to use it is not trivial

## Solution

- Fix `/setup-services.sh`, used by `/setup.sh` and `testing_in_container.sh`
- No `.changes` entry as this fixes no package (yet. TODO)

## Testing

- *Tested manually*: `testing_in_container.sh` passes

## Screenshots

No

### DIFF
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -51,6 +51,7 @@ $SUDO $ZYPPER install \
 # TODO extract list from gem2rpm.yml
 $SUDO $ZYPPER install \
   dbus-1-common \
+  dbus-1-daemon \
   suseconnect-ruby-bindings \
   autoyast2-installation \
   yast2 \


lslezak (approved):
LGTM

PR #1563, opened 2024-08-30
 agama config load/store for "product" uses the HTTP API

mvidner:
## Problem

**Product** and **registration** part of the migration of the CLI from D-Bus API to HTTP API:

- https://trello.com/c/hvPtBtMD/3719-5-replace-d-bus-with-http-based-clients?filter=member:martinvidner

## Solution

- Added `ProductHTTPClient`
- Kept (D-Bus) `ProductClient` because it serves as the backend for the above
- TODO mention the trickier parts of using two clients at two levels

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

No

### DIFF
--- a/rust/agama-lib/src/product/http_client.rs
+++ b/rust/agama-lib/src/product/http_client.rs
@@ -1,0 +1,75 @@
+use crate::software::model::RegistrationInfo;
+use crate::software::model::RegistrationParams;
+use crate::software::model::SoftwareConfig;
+use crate::{base_http_client::BaseHTTPClient, error::ServiceError};
+
+pub struct ProductHTTPClient {
+    client: BaseHTTPClient,
+}
+
+impl ProductHTTPClient {
+    pub fn new() -> Result<Self, ServiceError> {
+        Ok(Self {
+            client: BaseHTTPClient::new()?,
+        })
+    }
+
+    pub fn new_with_base(base: BaseHTTPClient) -> Self {
+        Self { client: base }
+    }
+
+    // FIXME get_software_config ?
+    pub async fn get_config(&self) -> Result<SoftwareConfig, ServiceError> {
+        self.client.get("/software/config").await
+    }
+
+    pub async fn set_config(&self, config: &SoftwareConfig) -> Result<(), ServiceError> {
+        self.client.put_void("/software/config", config).await
+    }
+
+    /// Returns the id of the selected product to install
+    pub async fn product(&self) -> Result<String, ServiceError> {
+        let config = self.get_config().await?;
+        if let Some(product) = config.product {
+            Ok(product)
+        } else {
+            Ok("".to_owned())
+        }
+    }
+
+    /// Selects the product to install
+    pub async fn select_product(&self, product_id: &str) -> Result<(), ServiceError> {
+        let config = SoftwareConfig {
+            product: Some(product_id.to_owned()),
+            patterns: None,
+        };
+        self.set_config(&config).await
+    }
+
+    pub async fn get_registration(&self) -> Result<RegistrationInfo, ServiceError> {
+        self.client.get("/software/registration").await
+    }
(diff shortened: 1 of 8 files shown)


mvidner:
#1272 switched `/api/manager/probe` to not wait for the D-Bus reply

rust/agama-lib/src/manager/http_client.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    pub fn new_with_base(base: BaseHTTPClient) -> Self {
+        Self { client: base }
+    }
+
+    pub async fn probe(&self) -> Result<(), ServiceError> {
+        // BaseHTTPClient did not anticipate POST without request body
+        let dummy_body: Vec<u8> = vec![];
+        self.client
+            .post_void("/manager/probe_sync", &dummy_body)
```
imobachgs, line 22: Instead of a dummy `Vec<u8>`, you could use just the void type.

```suggestion
            .post_void("/manager/probe_sync", &())
```

rust/agama-server/src/manager/web.rs
```diff
@@ -136,7 +137,21 @@ async fn probe_action<'a>(State(state): State<ManagerState<'a>>) -> Result<(), E
 
 /// Starts the probing process.
```
imobachgs, line 138: Please, document why we need a `sync` alternative.

testing_using_container.sh
```diff
@@ -44,7 +44,7 @@ podman run --name ${CNAME?} \
 # shortcut for the following
 CEXEC="podman exec ${CNAME?} bash -c"
 
-${CEXEC?} "cd /checkout && ./setup.sh"
+${CEXEC?} "cd /checkout && ./setup.sh || true"
```
imobachgs, line 47: Why?

rust/agama-lib/src/product/http_client.rs
```diff
@@ -0,0 +1,75 @@
...
+        Ok(Self {
+            client: BaseHTTPClient::new()?,
+        })
+    }
+
+    pub fn new_with_base(base: BaseHTTPClient) -> Self {
+        Self { client: base }
+    }
+
+    // FIXME get_software_config ?
```
imobachgs, line 21: Do this FIXME still apply? Some clarification would help.

rust/agama-lib/src/product/http_client.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// registration code used to register product
+    pub async fn registration_code(&self) -> Result<String, ServiceError> {
+        let ri = self.get_registration().await?;
+        Ok(ri.key)
+    }
+
+    /// email used to register product
+    pub async fn email(&self) -> Result<String, ServiceError> {
```
imobachgs, line 60: For consistency, I would use `registration_email`. Anyway, I would get rid of `registration_code` and `registration_email` completely. They are using in `ProductStore::load` and it causes the `GET /software/registration` to happen twice:

```rust
let registration_code = self.product_client.registration_code().await?;
let email = self.product_client.email().await?;
```

rust/agama-lib/src/product/http_client.rs
```diff
@@ -0,0 +1,75 @@
...
+        let ri = self.get_registration().await?;
+        Ok(ri.email)
+    }
+
+    /// register product
+    pub async fn register(&self, code: &str, email: &str) -> Result<(u32, String), ServiceError> {
+        // note RegistrationParams != RegistrationInfo, fun!
+        let params = RegistrationParams {
+            key: code.to_owned(),
```
imobachgs, line 69: np: We should be consistent: `key` or `code`?

imobachgs (changes requested):

imobachgs (approved):

PR #1570, opened 2024-09-03
refactor(web): bring back zFCP support (new HTTP / JSON API, queries, and TypeScript)

jreidinger:
**Trello:** https://trello.com/c/KMtaVltF (internal link)

This PR could be seen as a follow up of https://trello.com/c/2LC8JEZR (internal link) but in this case bringing back the support for managing zFCP devices on s390x architectures.

It includes:
  - Extending the HTTP API to expose zFCP devices management and zFCP related events.
  - Adapting the WEB UI for using Tanstack queries and the new HTTP API.
  - Adopting Typescript
  - Move to route pages.

## HTTP / JSON API Examples
This section contains a few examples you can use to explore the API. For the ones using curl, it is required to have a headers.txt file containing the credentials.  The file can be generated with:

```bash
#!/bin/bash

SERVER=myServer
PASSWORD=myPassword

TOKEN=$(curl --insecure -s -X POST ${SERVER}/api/auth -H 'Content-Type: application/json' -d '{"password": "'${PASSWORD}'"}' | jq -r ".token")
echo "Obtained token '${TOKEN}'"

echo -n "Authorization: Bearer " >headers.txt
echo $TOKEN >>headers.txt
```

<details>
<summary>zFCP operations</summary>

### Probing
```bash
$ curl -s -k -X POST ${SERVER}/api/storage/zfcp/probe -H @headers.txt | jq
null
```
### Check if zFCP is supported
```bash
$curl -s -k ${SERVER}/api/storage/zfcp/supported -H @headers.txt | jq
true
```
### Check whether allow_lun_scan is globally enabled or not
```bash
$curl -s -k ${SERVER}/api/storage/zfcp/config -H @headers.txt | jq
{
  "allowLunScan": false
}
```
### Get Controllers with complete information like WWPNs and LUNs
```bash
$ curl -s -k ${SERVER}/api/storage/zfcp/controllers -H @headers.txt | jq
[
  {
    "id": "1",
    "channel": "0.0.fa00",
    "lunScan": false,
    "active": true,
    "lunsMap": {
      "0x500507680d7e284a": [],
      "0x500507680d0e284a": [],
      "0x500507630b181216": [
        "0x4020404900000000"
      ]
    }
  },
  {
    "id": "2",
    "channel": "0.0.fc00",
    "lunScan": false,
    "active": true,
    "lunsMap": {
      "0x500507680d0e284b": [],
      "0x500507630b101216": [
        "0x0000000000000000",
        "0x0001000000000000"
      ],
      "0x500507680d7e284b": []
    }
  }
]
```
### Get Active Disks
```bash
$ curl -s -k ${SERVER}/api/storage/zfcp/disks -H @headers.txt | jq
[
  {
    "name": "/dev/sda",
    "channel": "0.0.fa00",
    "wwpn": "0x500507630b181216",
    "lun": "0x4020404900000000"
  },
  {
    "name": "/dev/sdb",
    "channel": "0.0.fc00",
    "wwpn": "0x500507630b101216",
    "lun": "0x0000000000000000"
  }
]
```
### Disk activation
```bash
CONTROLLERID=1
WWPN=0x500507630b181216
LUN=0x4020404900000000
curl -s -k -v -X POST ${SERVER}/api/storage/zfcp/controllers/${CONTROLLERID}/wwpns/${WWPN}/luns/${LUN}/activate_disk -H @headers.txt | jq
null
```
</details>

## Web UI Screenshots

The web has been adapted according to the recommended components to be used for Pages and content as well the Disk Activation Form has been moved from a Popup to its own route and Page.

Screenshot from 2024-09-17 23-47-40
Screenshot from 2024-09-17 23-45-24
Screenshot from 2024-09-17 23-45-48
Screenshot from 2024-09-17 23-45-58

## Additional notes

The PR also includes minor internal changes related to Storage DASD components, some of them related to page actions.


rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,138 @@
+//! Implements a client to access Agama's D-Bus API related to DASD management.
```
imobachgs, line 1:
```suggestion
//! Implements a client to access Agama's D-Bus API related to zFCP management.
```

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+use axum::{
+    extract::{Path, State},
+    routing::{get, post},
+    Json, Router,
+};
+
+use crate::{error::Error, web::common::EventStreams};
+
+/// Returns the stream of ZFCP-related events.
```
imobachgs, line 19:
```suggestion
/// Returns the stream of zFCP-related events.
```

imobachgs: Please, check other ocurrences of `ZFCP` and replace them with `zFCP`.

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,168 @@
...
+#[derive(Clone)]
+struct ZFCPState<'a> {
+    client: ZFCPClient<'a>,
+}
+
+pub async fn dasd_service<T>(dbus: &zbus::Connection) -> Result<Router<T>, ServiceError> {
+    let client = ZFCPClient::new(dbus.clone()).await?;
+    let state = ZFCPState { client };
+    let router = Router::new()
+        .route("/supported", get(supported))
```
imobachgs, line 42: np: I wonder whether we should have a single end-point under `/api/storage` that reports all the supported technologies.

imobachgs (commented):
The API looks good. Just some documentation here and there is missing (especially about the client), but I know it is a WIP.

rust/agama-server/src/storage/web/zfcp.rs
jreidinger: maybe something like list of supported flags

rust/agama-server/src/web/docs.rs
```diff
@@ -49,6 +49,20 @@ use utoipa::OpenApi;
         crate::storage::web::iscsi::nodes,
         crate::storage::web::iscsi::update_initiator,
         crate::storage::web::iscsi::update_node,
+        crate::storage::web::dasd::probe,
+        crate::storage::web::dasd::supported,
+        crate::storage::web::dasd::devices,
+        crate::storage::web::dasd::format,
+        crate::storage::web::dasd::enable,
+        crate::storage::web::dasd::disable,
+        crate::storage::web::dasd::set_diag,
```
jreidinger, line 58: as we obviously format about dasd in previous PR maybe it is time to consider something like https://github.com/rxdiscovery/utoipa_auto_discovery ?

web/src/types/zfcp.ts
```diff
@@ -0,0 +1,41 @@
...
+    channel: string;
+    WWPN: string;
+    LUN: string;
+  }
+  
+  type ZFCPOptions = {
+    allowLUNScan: boolean;
+  }
+  
+  export type { ZFCPController, ZFCPDisk, ZFCPOptions };
```
dgdavid, line 40: NP: Weird indentation

web/src/queries/zfcp.ts
```diff
@@ -0,0 +1,179 @@
...
+ */
+const ZFCPDisksQuery = () => ({
+  queryKey: ["zfcp", "disks"],
+  queryFn: fetchZFCPDisks,
+});
+
+/**
+ * Returns a query for retrieving the zFCP config
+ */
+const ZFCPConfigQuery = () => ({
```
imobachgs, line 48: I would write it as `zFCPConfigQuery`.

web/src/queries/zfcp.ts
```diff
@@ -0,0 +1,179 @@
...
+          const device: ZFCPController = event.device;
+          queryClient.setQueryData(["zfcp", "controllers"], (prev: ZFCPController[]) => {
+            return [...prev, device];
+          });
+          break;
+        }
+        case "ZFCPControllerRemoved": {
+          const device: ZFCPController = event.device;
+          const { id } = device;
+          queryClient.setQueryData(["zfcp", "controllers"], (prev: ZFCPController[]) => {
```
imobachgs, line 99: If you write the code in a slightly different way, you can make it simpler. Given that `setQueryData` receives a function, why not writing the code the other way around?

```javascript
queryClient.setQueryData(ZFCPConfigQuery.queryKey, (prev: ZFCPController[] | undefined) => {
  if (prev === undefined) return;

  // Put the switch here
})
```

Please be aware that `prev` might be `undefined`. Returning `undefined` from the updater function would make the query to not be updated. Similar to iSCSI [useNodesChanges](https://github.com/openSUSE/agama/pull/1598/files#diff-238cd70d57b6ea7dc5fa5411500ee14efca3f66dd775be6693f64e5dfb735126R96).

web/src/queries/zfcp.ts
teclator: Then all the queries and naming should be renamed in the same way

imobachgs: Yes, I would use camelcase for all of them.

web/src/api/zfcp.ts
```diff
@@ -51,42 +51,42 @@ const probeZFCP = () => post("/api/storage/zfcp/probe");
 /**
  * Activates given controller
  * @param controllerId id of existing controller
- * @returns
```
imobachgs: Returns what?

web/src/api/zfcp.ts
```diff
@@ -51,42 +51,42 @@ const probeZFCP = () => post("/api/storage/zfcp/probe");
...
  * Returns list of WWPNs for given controller
  * @param controllerId id of existing controller
- * @returns 
+ * @returns
  */
 const fetchWWPNs = (controllerId: string): Promise<string[]> => get(`/api/storage/zfcp/controllers/${controllerId}/wwpns`);
 
 /**
  * Returns list of LUNs for give controller and WWPN
- * @param controllerId
```
imobachgs: IMHO, having the name of the arguments without a description is not useful.

web/src/queries/zfcp.ts
jreidinger: agreed, this is mainly my attempt to have it idempotent and your code do it in easier way.

web/src/api/zfcp.ts
imobachgs: Then it would be good to write something there :-)

web/src/components/storage/ZFCPPage.tsx
```diff
@@ -0,0 +1,202 @@
...
+      <Page.Header>
+        <h2>{_("ZFCP")}</h2>
+      </Page.Header>
+
+      <Page.Content>
+        <Content />
+      </Page.Content>
+
+      <Page.Actions>
+        <Page.Back />
```
dgdavid, line 198: As you can see, I used Page.Back as it is here.

web/src/components/storage/ZFCPPage.tsx
```diff
@@ -47,125 +46,134 @@ import { PATHS } from "~/routes/storage";
...
-      <>
-        <LUNScanInfo />
-        <ZFCPControllersTable />
-      </>
-    );
-  };
+const NoDisksFound = () => {
+  const navigate = useNavigate();
+  const controllers = useZFCPControllers();
+  const activeDisks = !controllers.some((c) => c.active);
```
jreidinger, line 77: I do not like name as it emphasis that it is boolean if there is activeDisks...but it is in fact if there is any active controller.

dgdavid: Ok, so please improve naming if you don't mind. I did my best looking at the code I adapted 🥲

dgdavid: Nevermind, I've used `activeController` instead. You're right and now it is more meaningful and readable :)

web/src/components/storage/DASDPage.tsx
```diff
@@ -43,7 +43,7 @@ export default function DASDPage() {
       </Page.Content>
 
       <Page.Actions>
-        <Page.Back>{_("Close")}</Page.Back>
+        <Page.Cancel navigateTo={PATHS.targetDevice}>{_("Back")}</Page.Cancel>
```
dgdavid, line 46: Do you really navigate to a specific path or it's enough going Page.Bavk?

Just a question

web/src/components/storage/DASDPage.tsx
```diff
@@ -40,6 +41,10 @@ export default function DASDPage() {
         <DASDTable />
         <DASDFormatProgress />
       </Page.Content>
+
+      <Page.Actions>
+        <Page.Back>{_("Close")}</Page.Back>
```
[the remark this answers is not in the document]
teclator: I agree, but navigating back is also "dangerous", currently it goes to the form dialog if you press back after trying to activate a disk which is not the expected behavior, therefore, I prefered to specify the path and force to navigate to the entry point of the route..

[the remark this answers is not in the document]
imobachgs: What do you mean by "dangerous"?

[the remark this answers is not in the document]
dgdavid: I think it's not dangerous at all. Just a bit messy. In any case, I had a chat with @teclator and understood the _problem_ with the Page.Back action. A _problem_ that still there because the user can use the browser back. 

So, we agreed to keep it as a navigation action and address the back/cancel _problems_ in other PR.

BTW, the _problem_ could be easily solve by [replacing](https://reactrouter.com/en/main/hooks/use-navigate#optionsreplace) the route when navigating back from or submitting a form since, maybe, we want the user reaching a form **explicitly**

web/src/api/zfcp.ts
teclator: I think that with the defined types and parameter names it is obvious and do not need extra documentation but can add some if really needed.

web/src/components/storage/dasd/DASDPage.tsx
```diff
@@ -20,9 +20,8 @@
  */
 
 import React from "react";
-import DASDTable from "~/components/storage/DASDTable";
+import { DASDTable, DASDFormatProgress } from "~/components/storage/dasd";
```
dgdavid, line 23: These components are in the same scope that the one importing them. Please, use relative route and do not export them if not used outside.

web/src/components/storage/dasd/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export { default as DASDPage } from "./DASDPage";
+export { default as DASDTable } from "./DASDTable";
+export { default as DASDFormatProgress } from "./DASDFormatProgress";
```
dgdavid, line 24: As said in another component, looks like these components are not used outside its own storage/dasd _scope_. Please, do not export them.

web/src/components/storage/zfcp/index.js
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+export { default as ZFCPPage } from "./ZFCPPage";
+export { default as ZFCPDiskActivationPage } from "./ZFCPDiskActivationPage";
+export { default as ZFCPDiskForm } from "./ZFCPDiskForm";
```
dgdavid, line 24: Is it needed to export ZFCPDiskForm. Is it used outside the storage/zfcp _scope_.

dgdavid (approved):

web/src/queries/zfcp.ts
```diff
@@ -26,67 +26,56 @@ import { useInstallerClient } from "~/context/installer";
 import React from "react";
 import { ZFCPConfig, ZFCPController, ZFCPDisk } from "~/types/zfcp";
 
-/**
- * Returns a query for retrieving the zFCP controllers
- */
-const zfcpControllersQuery = () => ({
+const zfcpControllersQuery = {
```
dgdavid, line 28: BTW, maybe I miss something at some point but the only suggestion I see is about the naming: https://github.com/openSUSE/agama/pull/1570#pullrequestreview-2302148845

But not about the full structure.  I've no strong opinion about writing them as function or object. But I though we were writing them always as function for consistence since they could receive arg for building the query. As far as I can see, @imobachgs has wrote some as function and some as object in the storage part too. We can go ahead with such approach if you think it is better, but we should be consistent in all places.

PLEASE, do not take this comment as a suggestion for continue sending changes against this PR. Whatever we decide to choose should be addressed in a change unifying all the queries across the code. Otherwise we will be just adding more and more noise here.

teclator: This change was based on this another suggestion https://github.com/openSUSE/agama/pull/1570#discussion_r1758205470, in that case objects were used, so, I also does not have an strong opinion and agree on unifying.

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+                        active: get_property(properties, "Active")?,
+                        luns_map: self.get_luns_map(path_s.as_str()).await?,
+                    },
+                ))
+            }
+        }
+        Ok(devices)
+    }
+
+    /// Gets controller id that can be used in other controller methods from its dbus path
```
imobachgs, line 102:
```suggestion
    /// Gets controller id that can be used in other controller methods from its D-Bus path
```

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    pub async fn get_controllers(
+        &self,
+    ) -> Result<Vec<(OwnedObjectPath, ZFCPController)>, ServiceError> {
+        let managed_objects = self.object_manager_proxy.get_managed_objects().await?;
+
+        let mut devices: Vec<(OwnedObjectPath, ZFCPController)> = vec![];
+        for (path, ifaces) in managed_objects {
+            if let Some(properties) = ifaces.get("org.opensuse.Agama.Storage1.ZFCP.Controller") {
+                // TODO: maybe move it to model? but it needs also additional calls to dbus
```
imobachgs, line 85:
```suggestion
                // TODO: maybe move it to model? but it needs also additional calls to D-Bus
```

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// ```
+    ///     let path = zbus::zvariant::OwnedObjectPath::try_from("/test/controllers/2").unwrap();
+    ///     let id = agama_lib::storage::client::zfcp::ZFCPClient::controller_id_from_path(&path);
+    ///     assert_eq!(id, "2".to_string());
+    /// ```
+    ///
+    /// edge case with malformed path that will panic in Owned path already
+    /// ```should_panic
+    ///     let path = zbus::zvariant::OwnedObjectPath::try_from("mangled").unwrap();
+    ///     let id = agama_lib::storage::client::zfcp::ZFCPClient::controller_id_from_path(&path);
```
imobachgs, line 114: According to this comment, the call to `OwnedPath::try_from` will fail, right? So what is the point of this code?

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///     assert_eq!(id, "mangled".to_string());
+    /// ```
+    pub fn controller_id_from_path(path: &OwnedObjectPath) -> String {
```
imobachgs, line 117: We already had a helper to get the ID from a path. See https://github.com/openSUSE/agama/blob/bc057f474f9783717f355af970440e7215b2b705/rust/agama-lib/src/dbus.rs#L73.

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+        controller_id: &str,
+    ) -> Result<ZFCPControllerProxy, ServiceError> {
+        let dbus = ZFCPControllerProxy::builder(&self.connection)
+            .path(ZFCP_CONTROLLER_PREFIX.to_string() + "/" + controller_id)?
+            .build()
+            .await?;
+        Ok(dbus)
+    }
+
+    // TODO: does all those method with controller id as parameter deserve to be moved to controller model?
```
imobachgs, line 135: np: Do you plan to do something with this comment? Otherwise, I would remove the `TODO` prefix.

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+        &self,
+        controller_id: &str,
+        wwpn: &str,
+    ) -> Result<Vec<String>, ServiceError> {
+        let controller = self.get_controller_proxy(controller_id).await?;
+        let result = controller.get_luns(wwpn).await?;
+        Ok(result)
+    }
+
+    pub async fn get_luns_map(
```
imobachgs, line 159: This method deserves a some documentation.

rust/agama-lib/src/storage/model/zfcp.rs
```diff
@@ -0,0 +1,68 @@
+//! Implements a data model for DASD devices management.
```
imobachgs, line 1:
```suggestion
//! Implements a data model for zFCP devices management.
```

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+            "/controllers/:controller_id/activate",
+            post(activate_controller),
+        )
+        .route("/controllers/:controller_id/wwpns", get(get_wwpns))
+        .route(
+            "/controllers/:controller_id/wwpns/:wwpn_id/luns",
+            get(get_luns),
+        )
+        .route(
+            "/controllers/:controller_id/wwpns/:wwpn_id/luns/:lun_id/activate_disk",
```
imobachgs, line 65: I am not sure about those URLs that only support POST actions. I wonder whether the arguments should be the regular payload of the POST method. Let's keep it as it is by now.

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+    Ok(router)
+}
+
+/// Returns whether zFCP technology is supported or not
+#[utoipa::path(
+    get,
+    path="/supported",
+    context_path="/api/storage/zfcp",
+    responses(
+        (status = OK, description = "Returns whether ZFCP technology is supported")
```
imobachgs, line 85:
```suggestion
        (status = OK, description = "Returns whether zFCP technology is supported")
```

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+    Ok(Json(state.client.supported().await?))
+}
+
+/// Returns global zFCP configuration
+#[utoipa::path(
+    get,
+    path="/config",
+    context_path="/api/storage/zfcp",
+    responses(
+        (status = OK, description = "Returns global ZFCP configuration", body=ZFCPOptions)
```
imobachgs, line 98:
```suggestion
        (status = OK, description = "Returns global zFCP configuration", body=ZFCPOptions)
```

rust/agama-lib/src/storage/model/zfcp.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// flag whenever channel is active
+    pub active: bool,
+    /// map of associated WWPNs and its LUNs
+    pub luns_map: HashMap<String, Vec<String>>,
+}
+
+/// Represents a zFCP global options (specific to s390x systems).
+#[derive(Clone, Debug, Serialize, Default, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ZFCPOptions {
```
imobachgs, line 55: I would say this struct does not belong to the `agama-lib` model:

* The `TryFrom` is not used, so no need to implement it.
* It is only used in the context of the web. So, please, move it there.

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+    }))
+}
+
+/// Returns the list of known zFCP disks.
+#[utoipa::path(
+    get,
+    path="/disks",
+    context_path="/api/storage/zfcp",
+    responses(
+        (status = OK, description = "List of ZFCP disks", body = Vec<ZFCPDisk>)
```
imobachgs, line 113:
```suggestion
        (status = OK, description = "List of zFCP disks", body = Vec<ZFCPDisk>)
```

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+    Ok(Json(devices))
+}
+
+/// Returns the list of known zFCP controllers.
+#[utoipa::path(
+    get,
+    path="/controllers",
+    context_path="/api/storage/zfcp",
+    responses(
+        (status = OK, description = "List of ZFCP controllers", body = Vec<ZFCPController>)
```
imobachgs, line 133:
```suggestion
        (status = OK, description = "List of zFCP controllers", body = Vec<ZFCPController>)
```

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+    Ok(Json(()))
+}
+
+/// List wwpns for given controller.
+#[utoipa::path(
+    post,
+    path="/controllers/:controller_id/wwpns",
+    context_path="/api/storage/zfcp",
+    responses(
+        (status = OK, description = "list of wwpns", body=Vec<String>)
```
imobachgs, line 175:
```suggestion
        (status = OK, description = "List of WWPNs", body=Vec<String>)
```

It goes to the OpenAPI docs, so please use *WWPNs* or *World Wide Port Names*. The same for the rest of endpoints.

rust/agama-server/src/storage/web/zfcp.rs
```diff
@@ -0,0 +1,254 @@
...
+)]
+async fn get_wwpns(
+    State(state): State<ZFCPState<'_>>,
+    Path(controller_id): Path<String>,
+) -> Result<Json<Vec<String>>, Error> {
+    let result = state.client.get_wwpns(controller_id.as_str()).await?;
+    Ok(Json(result))
+}
+
+/// List luns for given controller and wwpn.
```
imobachgs, line 186:
```suggestion
/// List LUNs for given controller and wwpn.
```

imobachgs: The same for the rest of *luns*.

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///
+    /// edge case with malformed path that will panic in Owned path already
+    /// ```should_panic
+    ///     let path = zbus::zvariant::OwnedObjectPath::try_from("mangled").unwrap();
+    ///     let id = agama_lib::storage::client::zfcp::ZFCPClient::controller_id_from_path(&path);
+    ///     assert_eq!(id, "mangled".to_string());
+    /// ```
+    pub fn controller_id_from_path(path: &OwnedObjectPath) -> String {
+        let mut path_s = path.to_string();
+        let slash_pos = path_s.rfind("/").unwrap_or(0);
```
imobachgs, line 119:
```suggestion
        let slash_pos = path_s.rfind('/').unwrap_or(0);
```

web/src/api/zfcp.ts
```diff
@@ -0,0 +1,91 @@
+/*
```
imobachgs, line 1: IMHO, `zftp` should be `storage/zfcp` (as the `iscsi` case). The same for `dasd` (although I do not expect `dasd` to be moved in this PR).

web/src/queries/zfcp.ts
```diff
@@ -0,0 +1,193 @@
...
+const useZFCPDisksChanges = () => {
+  const client = useInstallerClient();
+  const queryClient = useQueryClient();
+
+  React.useEffect(() => {
+    if (!client) return;
+
+    return client.ws().onEvent((event) => {
+      switch (event.type) {
+        case "ZFCPDiskAdded": {
```
imobachgs, line 150: This code looks too complex to me. You are calling `setQueryData` three times, passing different functions. I would do it the other way around. Please, see [the iSCSI example](https://github.com/openSUSE/agama/blob/bc057f474f9783717f355af970440e7215b2b705/web/src/queries/storage/iscsi.ts#L109).

web/src/utils/zfcp.test.ts
```diff
@@ -0,0 +1,88 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+// @ts-check
```
imobachgs, line 22: Please, remove this line.

imobachgs (changes requested):

web/src/components/storage/zfcp/ZFCPDiskForm.tsx
```diff
@@ -107,7 +107,7 @@ export default function ZFCPDiskForm({
   if (!formData.channel && getChannels().length > 0) select();
 
   return (
-    <Page.Section>
+    <Page.Section aria-label="zfcp-activation-form">
```
dgdavid, line 110: This is not a valid aria-label.nit looks more like an id.  aria-label should be a human readable text. Don't forget to mar it for translation.

rust/agama-lib/src/storage/client/zfcp.rs
teclator: @jreidinger I guess it is a question you wanted to raise and discuss, so, will remove the comment but maybe this thread could be used for having that discussion... by now it looks good to me to leave the definition where it is, @imobachgs, what do you think?

imobachgs: It is fine. I do not like that we use two different approaches, but let's move on.

web/src/api/zfcp.ts
teclator: Already did for the WEB components, so it is ok for moving also in this case.

teclator: Already did for the components, so, it is fine doing it for both cases at the same time.

web/src/queries/zfcp.ts
imobachgs: > This change was based on this another suggestion [#1570 (comment)](https://github.com/openSUSE/agama/pull/1570#discussion_r1758205470), in that case objects were used, so, I also does not have an strong opinion and agree on unifying.

Well, the point of my comment was not the `queryKey` thingie, but the overall approach of the whole function. I insisted in https://github.com/openSUSE/agama/pull/1570#discussion_r1764431153.

rust/agama-lib/src/storage/client/zfcp.rs
teclator: Removed the method, so not needed suggestion ;)

imobachgs (approved):
LGTM. Congrats!

PR #1573, opened 2024-09-03
fix(live): Firefox: disable the initial configuration workflow

lslezak:
## Problem

In SLFO the Firefox shows the welcome screen at the first start instead of the Agama page.

Agama-Firefox-SLE

## Solution

- Disable the initial welcome screen in the default Firefox profile

## Testing

- With this option the welcome screen is skipped


imobachgs (approved):

lslezak:
More details about the welcome screen: https://firefox-source-docs.mozilla.org/browser/components/asrouter/docs/about-welcome.html

PR #1575, opened 2024-09-04
feat(storage): config support to reuse file systems

joseivanlopez:
Allow using *reuseIfPossible* for a *filesystem* section in the storage JSON config. The [auto_storage](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md) document proposes to have a *reuse* property, but *reuseIfPossible* was used instead, see https://github.com/openSUSE/agama/pull/1575#pullrequestreview-2282229940:

If *reuseIfPossible* is true, then Agama will try to reuse the file system, if any. In that case, some other properties from the *filesystem* section would be ignored.

Example:

~~~json
{
  "search": "/dev/vda1",
  "filesystem": {
    "reuseIfPossible": true,
    "path": "/logs",
    "type": "xfs",
    "mkfsOptions": ["rw"]
  }
}
~~~

The properties *type* and *mkfsOptions* would be considered by Agama only if the file system is not finally reused.

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -870,9 +870,9 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "reuse": {
+        "reuseIfPossible": {
           "title": "Reuse file system",
-          "description": "Whether to reuse the existing file system (if any).",
+          "description": "Try to reuse the existing file system. In some cases the file system could not be reused, for example, if the device is re-encrypted.",
           "type": "boolean",
           "default": false
         },
@@ -886,7 +886,7 @@
         "path": {
           "title": "Mount path",
           "type": "string",
-          "examples": ["/dev/vda"]
+          "examples": ["/var/log"]
         },
         "mountBy": {
           "title": "How to mount the device",
--- a/service/test/y2storage/agama_proposal_test.rb
+++ b/service/test/y2storage/agama_proposal_test.rb
@@ -24,6 +24,51 @@ require "agama/config"
 require "agama/storage/config"
 require "y2storage/agama_proposal"
 
+# @param config [Agama::Storage::Configs::Drive, Agama::Storage::Configs::Partition]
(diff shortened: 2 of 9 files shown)


rust/agama-lib/share/profile.schema.json
```diff
@@ -869,9 +869,9 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "reuse": {
+        "reuseIfPossible": {
```
joseivanlopez, line 872: The document [auto_storage](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md) proposes to have a [reuse](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md#keeping-an-existing-file-system-or-encryption-layer) property, but *reuseIfPossible* was used instead because the following reasons:

* A file system cannot be reused if the partition is going to be created (*search* is not specified).
* A file system cannot be reused if the device is going to be encrypted (the encryption layer is not reused). 
* A file system cannot be reused if the device is not currently formatted.

In order to validate JSON config to the usage of *reuse* in the schema would make the schema more complex with very little benefit.  

TODO

service/test/y2storage/agama_proposal_test.rb
```diff
@@ -24,6 +24,51 @@
 require "agama/storage/config"
 require "y2storage/agama_proposal"
 
+# @param config [Agama::Storage::Configs::Drive, Agama::Storage::Configs::Partition]
```
ancorgs, line 27: These methods look like a hint that we need (or will shortly need) shared helpers or factories for creating config objects and similar stuff.

ancorgs (commented):
There are conflicts at the changelog. Other than that it LGTM. Fix the conflict and I will approve.

service/test/y2storage/agama_proposal_test.rb
joseivanlopez: Totally agree.

ancorgs (approved):
LGTM

PR #1577, opened 2024-09-05
feat(web): partially adapt storage UI to queries

imobachgs:
Trello: https://trello.com/c/nV7mjrf9/

This is the first step towards adapting the storage code to use queries. As this part of the code is rather complex, we decided to move as many components as possible to TypeScript to have a kind of security net.

## What is included

* Add a set of types from from our OpenAPI docs (see #1564) in `api/storage/types`.
* Add a set of storage UI types in `types/storage`. We should unify the types, but let's do it later.
* Define a new `api/storage` module moving big chunks of the `StorageClient` code.
* Define a set of queries under `queries/storage`.
* Adapt the `ProposalPage` component to use queries. It retrieves a lot of information that injects to nested components.
* Convert `ProposalPage` and `InstallationDeviceField` to TypeScript.

## What is missing

* As you may see, the front end is not building. That problems (and other subtle bugs) will be fixed as we adapt the rest of storage components to TypeScript. So a bunch of additional pull requests will follow.
* Additionally, we are removing all those skeletons from the UI. We will use a unified mechanism.


web/src/api/storage.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) [2022-2023] SUSE LLC
```
dgdavid, line 2: NP: looks like a new file, wrong copy header?

dgdavid (approved):
As I guess this is just one of many incoming changes against the `storage-typescript` feature branch, please go ahead. It's looking good and, honestly, I prefer to make a more in depth review when the code is closer to a final state. Hope you don't mind.

PR #1583, opened 2024-09-05
 refactor(web): convert some storage components to TypeScript (round 2)

imobachgs:
Continuation of #1577. It covers the conversion of the following components/modules:

* `BootConfigField`
* `device-utils`
* `SpaceActionsTable`
* `DeviceSelectorTable`
* `EncryptionSettingsDialog`
* `ProposalActionDialog`
* `ProposalActionSummary`
* `StoragePage`
* `ProposalResultSection`
* `ProposalTransactionalInfo`
* `SpaceActionsTable`


web/src/components/storage/BootConfigField.tsx
```diff
@@ -21,59 +21,51 @@
 
 // @ts-check
 
-import React from "react";
+import React, { ReactNode } from "react";
```
dgdavid, line 24: NP: I use to not import React types but use a `React.ReactType` instead. No strong opinion, but would be nice if we reach an agreement about it.

web/src/components/storage/BootConfigField.test.tsx
```diff
@@ -54,48 +47,18 @@ const sda = {
...
- */
-const openBootConfigDialog = async () => {
-  const { user } = plainRender(<BootConfigField {...props} />);
-  const button = screen.getByRole("button");
-  await user.click(button);
-  const dialog = screen.getByRole("dialog", { name: "Partitions for booting" });
-
-  return { user, dialog };
-};
-
```
dgdavid, line 83: NP: Should we take the opportunity to drop the below `.skip`?

web/src/components/storage/StoragePage.jsx
```diff
@@ -1,34 +0,0 @@
...
- * find current contact information at www.suse.com.
- */
-
-import React from "react";
-import { Outlet } from "react-router-dom";
-import { Page } from "~/components/core";
-import { _ } from "~/i18n";
-import { navigation } from "./routes";
-
-export default function StoragePage() {
```
dgdavid, line 28: Great catch! :)

dgdavid (approved):

web/src/components/storage/BootConfigField.test.tsx
imobachgs: I tried, but the `Link` we use in the component gave me some problems. So I decided to postpone it.

dgdavid (approved):

PR #1586, opened 2024-09-06
doc(web): update changes file

imobachgs:
Adds the changes related to the use of queries to the changelog.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Fri Sep  6 08:06:41 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Adopt TanStack Query and Axios for state management and
+  interacting with Agama's HTTP API (gh#openSUSE/agama#1439,
+  gh#openSUSE/agama#1452, gh#openSUSE/agama#1483, gh#openSUSE/agama#1492,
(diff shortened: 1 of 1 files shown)


web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Fri Sep  6 08:06:41 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Adopt TanStack Query and Axios for state management and
+  interacting with Agama's HTTP API (gh#openSUSE/agama#1439,
+  gh#openSUSE/agama#1452, gh#openSUSE/agama#1483, gh#openSUSE/agama#1492,
+  gh#openSUSE/agama#1503, gh#openSUSE/agama#1504, gh#openSUSE/agama#1514,
+  gh#openSUSE/agama#1516, gh#openSUSE/agama#1519, gh#openSUSE/agama#1577,
+  gh#openSUSE/agama#1580).
```
dgdavid, line 9: Great! But we're gonna left behind DASD and zFCP parts that are being migrated too. Anyway, I think it is ok as it is.

dgdavid (approved):

web/package/agama-web-ui.changes
imobachgs: DASD and zFCP are different cases: we are rewriting them with the new architecture, not only changing the queries. So they deserve their own entry in the changes log.

We are only missing iSCSI, but I am not sure if the migration will happen before the next release.

PR #1587, opened 2024-09-06
Ensure that product icons are packaged

lkocman:
Hello folks

Here's a fix that ensures installation of missing product icons but I'm not really happy about two things.

It seems that we're somehow trying to webpack this (I don't really have experience with this) I know that we're doing some npm installation in spec so I'm not sure if we should fix the webpack instead.

```
{_builddir}/agama/src/assets/products/*.svg
web/webpack.config.js:  { from: "./src/assets/products/*.svg", to: "assets/logos/[name][ext]" }

```

Why do we have it in src/assets/products/*.svg and install it into assets/logos/*.svg ?
Let's just use same path in both cases. 

Cheers

image

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri Sep  6 15:04:58 UTC 2024 - Lubos Kocman <lubos.kocman@suse.com>
+
+- Ensure that product icons are packaged 
+
+-------------------------------------------------------------------
 Fri Sep  6 08:06:41 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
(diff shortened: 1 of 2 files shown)


imobachgs:
I overlooked this pull request and had created an identical patch to fix the issue :-) Thank you!

imobachgs:
We need to think a little bit about this, because IMHO the logos could be included in the products package (where they are mentioned). But, by now, it is good enough.

dgdavid:
> Why do we have it in src/assets/products/_.svg and install it into assets/logos/_.svg ?

My fault @lkocman I don't know why I chosen the latest while working on #1550

> We need to think a little bit about this, because IMHO the logos could be included in the products package (where they are mentioned)

Agree. The product package should be responsible of placing its assets in the _web\_ui_ assets directory instead of adding them one by one to the web client build. We should work on defining it.

PR #1588, opened 2024-09-06
fix(web): package: do not try to install non-existant directories

g7:
The logos/ directory is already inside web_ui/.

## Problem

Latest agama-web-ui [is not building](https://build.opensuse.org/package/live_build_log/systemsmanagement:Agama:Devel/agama-web-ui/openSUSE_Tumbleweed/x86_64) after #1587 .

## Solution

Since the logos are already installed in `web_ui/logos`, we can drop the two (non-existent)? directories
listed in the %files section:

```
%{_datadir}/agama/assets
%{_datadir}/agama/products
```

## Testing

- *Tested manually*


imobachgs (approved):

imobachgs:
Thanks!

PR #1593, opened 2024-09-09
fix(live): firefox: set browser.startup.homepage_override.mstone to ignore

g7:
Observed on Firefox ESR 128.1 in SLFO (with MozillaFirefox-branding-SLE). Without this option, Agama live would boot to the Firefox "new tab" window.

Only after closing and reopening Firefox multiple times it would then open the correct homepage as specified in the user.js settings.

Setting browser.startup.homepage_override.mstone to ignore helps in these cases.

MozillaZine reference: https://kb.mozillazine.org/Browser.startup.homepage_override.mstone

## Screenshots

Screenshot from 2024-09-09 17-40-00


imobachgs (approved):

imobachgs:
Thanks!

PR #1598, opened 2024-09-12
refactor(web): use queries to handle iSCSI configuration

imobachgs:
Trello: https://trello.com/c/MsGbqiwA/

This is a follow-up of https://github.com/openSUSE/agama/pull/1585 which adapts the iSCSI-related code to use queries instead of the regular Agama client. Additionally, it converts a good share of the code to TypeScript and adds some iSCSI tests, that were not included in the first version.


web/src/api/storage/iscsi.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param node - ISCSI node
+ */
+const logout = async (node: ISCSINode) => post(nodePath(node) + "/logout");
+
+/**
+ * Sets the startup status of the connection
+ *
+ * @param node - ISCSI node
+ * @param startup - startup status
```
jreidinger, line 107: hmm, startup is string, so it can accept any value? or is it kind of enum with limited statuses?

web/src/components/storage/iscsi/InitiatorPresenter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    await user.click(editButton);
+
+    const dialog = await screen.findByRole("dialog");
+    const nameInput = await within(dialog).findByLabelText(/Name/);
+    await user.clear(nameInput);
+    await user.type(nameInput, "my-initiator");
+    const confirmButton = screen.getByRole("button", { name: "Confirm" });
+    await user.click(confirmButton);
+
+    expect(mockInitiatorMutation.mutateAsync).toHaveBeenCalledWith({ name: "my-initiator" });
```
jreidinger, line 65: :+1:  to finally see test with mutation

web/src/components/storage/iscsi/TargetsSection.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  startup: "onboot",
+};
+
+jest.mock("~/queries/storage/iscsi", () => ({
+  ...jest.requireActual("~/queries/storage/iscsi"),
+  useNodes: () => mockNodes,
+}));
+
+jest.mock("~/api/storage/iscsi", () => ({
+  ...jest.requireActual("~/queries/storage/iscsi"),
```
jreidinger, line 58: is there reason to load rest of queries? Cannot it do some backend communication? I am asking this in all those mock that include also requireActual

web/src/components/storage/iscsi/TargetsSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Section } from "~/components/core";
+import { NodesPresenter, DiscoverForm } from "~/components/storage/iscsi";
+import { _ } from "~/i18n";
+import { useNodes, useNodesChanges } from "~/queries/storage/iscsi";
+import { discover } from "~/api/storage/iscsi";
+
+export default function TargetsSection() {
+  const [isDiscoverFormOpen, setIsDiscoverFormOpen] = useState<boolean>(false);
+  const nodes = useNodes();
+  useNodesChanges();
```
jreidinger, line 33: Is there some changes done in `changes` queries? Previously it returns query. So it can be written like `const nodes = useNodesChanges();` . And it is still done this way e.g. in DASD, but in others where it was before I do not see it anymore. ( e.g. https://github.com/openSUSE/agama/blob/master/web/src/queries/dasd.ts#L121 )

web/src/queries/storage/iscsi.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { useMutation, useQueryClient, useSuspenseQuery } from "@tanstack/react-query";
+import React from "react";
```
dgdavid, line 23: Silly NP :see_no_evil: 

```suggestion
import React from "react";
import { useMutation, useQueryClient, useSuspenseQuery } from "@tanstack/react-query";
```

web/src/api/storage/iscsi.ts
imobachgs: It is a generic string. I do not like it that much, but I decided to keep it as it was, considering it was out of this PR's scope. But sure we should improve how we handle the staus.

web/src/components/storage/iscsi/TargetsSection.test.tsx
imobachgs: For no particular reason. I like to keep the original mode and mock only those functions that the test will use.

web/src/components/storage/iscsi/TargetsSection.tsx
imobachgs: Well, that's an interesting topic. We usually have separate hooks for retrieving the query and subscribing to changes. DASD handling is the only place where is done differently. DASD hooks 1) retrieve the value and 2) subscribe to the changes. So, if you use the `useDASDFormatJobsChanges` in more than one place, you are subscribing twice. In this particular case, the `JobAdded` event will not be handled correctly (that's easy to fix).

Let's consider that our DASD formats progress is not blocking (we discussed that during the review), and you want to add a counter to the pending jobs. Then you might want to access that query from your counter, too. And then, you will find the problem mentioned above.

Having two different hooks (`useIssues` and `useIssuesChanges`) might be a problem too. You need to be sure that the changes hooks are used at least at the same level as the query hook. That's why most `changes` hooks are called from the `App` component.

Finally, something we need to fix, is adapting all those *changes* queries to not update the query if it is not defined. I did that correctly (I think) in this pull request, but we might need to fix others.

imobachgs: Perhaps I should move the call to `useNodesChanges` to the `App` level.

web/src/queries/storage/iscsi.ts
```diff
@@ -0,0 +1,115 @@
...
+const initiatorQuery = {
+  queryKey: ["storage", "iscsi", "initiator"],
+  queryFn: async (): Promise<ISCSIInitiator> => {
+    const initiator = await fetchInitiator();
+    // FIXME: what is the offloadCard?
+    return { ...initiator, offloadCard: "" };
+  },
+};
+
+const useInitiator = (): ISCSIInitiator => {
```
imobachgs, line 38: Note to myself: let's document these hooks.

dgdavid (approved):
:+1:

web/src/components/storage/iscsi/TargetsSection.test.tsx
jreidinger: The part that worry me is that it can cause accidental communication with backend...if it is not there, it will immediatelly fail in test, not?

imobachgs: Yes, the test will fail in either case.

jreidinger (approved):

PR #1600, opened 2024-09-13
agama config load/store for "storage" uses the HTTP API

mvidner:
## Problem

- https://trello.com/c/hvPtBtMD

To recap: 
> When moving to the new HTTP-based architecture, we took some shortcuts. One of them was not using HTTP clients in the command-line interface. 

## Solution

- Add `StorageHTTPClient`
- This :stop_sign: ends D-Bus :stop_sign: usage from `SettingsStore` :heavy_check_mark: 

## Testing

- Added a new unit test
- Tested manually: note that it is easy to pass invalid Storage config in this way which will confuse the UI to the point of not rendering the storage parts at all.

## Screenshots

No

### DIFF
--- a/rust/agama-server/src/storage/web.rs
+++ b/rust/agama-server/src/storage/web.rs
@@ -125,11 +125,7 @@ pub async fn storage_service(dbus: zbus::Connection) -> Result<Router, ServiceEr
 )]
 async fn get_config(State(state): State<StorageState<'_>>) -> Result<Json<StorageSettings>, Error> {
     // StorageSettings is just a wrapper over serde_json::value::RawValue
-    let settings = state
-        .client
-        .get_config()
-        .await
-        .map_err(|e| Error::Service(e))?;
+    let settings = state.client.get_config().await.map_err(Error::Service)?;
     Ok(Json(settings))
 }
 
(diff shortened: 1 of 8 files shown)


mvidner:
Hmm, after `load` worked on the first try I wondered how `store` would work. And it does! ... by not validating anything? Is that right?

```console
# echo '{"storage":{"bogus": "validate much?"}}' | agama config load
# agama config show
{
  "user": {
    "...": "...".
  },
  "storage": {
    "bogus": "validate much?"
  }
}
```

rust/agama-server/src/storage/web.rs
```diff
@@ -109,6 +110,55 @@ pub async fn storage_service(dbus: zbus::Connection) -> Result<Router, ServiceEr
...
+)]
+async fn set_config(
+    State(state): State<StorageState<'_>>,
+    Json(settings): Json<StorageSettings>,
+) -> Result<Json<()>, Error> {
+    let _status: u32 = state
+        .client
+        .set_config(settings)
+        .await
+        .map_err(|e| Error::Service(e))?;
```
imobachgs, line 158: You can write this as:

```suggestion
        .map_err(Error::Service)?;
```

imobachgs (approved):
Congrats! I have added just a minor comment but, otherwise, it LGTM.

PR #1602, opened 2024-09-13
Add additional wireless settings

jcronenberg:
## Problem

Network model was missing some required fields for the wicked migration.

## Solution

Added the required fields to both the model and settings. I've also added some fields present in the model to be available via the settings.

## Testing

- *Extended existing unit tests*
- *Tested manually*
```bash
sudo agama auth login && AGAMA_TOKEN=`sudo agama auth show`
curl http://localhost/api/network/connections \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $AGAMA_TOKEN" \
    -d '{ "id": "wifi", "method4": "auto", "method6": "auto", "ignoreAutoDns": false, "wireless": { "security": "wpa-psk", "ssid": "test-ssid", "mode": "infrastructure", "groupAlgorithms": [ "wep40", "wep104" ], "pairwiseAlgorithms": [ "tkip", "ccmp" ], "wpaProtocolVersions": [ "wpa" ], "pmf": 2, "hidden": true, "band": "a", "channel": 100, "bssid": "12:34:56:78:9A:BC" }, "status": "down" }'
curl -X POST http://localhost/api/network/system/apply \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $AGAMA_TOKEN"
nmcli con show wifi
# See relevant fields in 802-11-wireless.* and 802-11-wireless-security.* set.
```

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -186,7 +186,7 @@
                     "title": "Only allow connection to this mac address",
                     "type": "string"
                   },
-                  "group_algorithms": {
+                  "groupAlgorithms": {
                     "type": "array",
                     "items": {
                       "title": "A list of group/broadcast encryption algorithms",
@@ -199,7 +199,7 @@
                       ]
                     },
                   },
-                  "pairwise_algorithms": {
+                  "pairwiseAlgorithms": {
                     "type": "array",
                     "items": {
                       "title": "A list of pairwise encryption algorithms",
@@ -210,7 +210,7 @@
                       ]
                     }
                   },
-                  "wpa_protocol_versions": {
+                  "wpaProtocolVersions": {
                     "type": "array",
                     "items": {
                       "title": "A list of allowed WPA protocol versions",
@@ -312,7 +312,7 @@
                       ]
                     }
                   },
-                  "phase2_auth": {
+                  "phase2Auth": {
                     "title": "Phase 2 inner auth method",
                     "type": "string",
                     "enum": [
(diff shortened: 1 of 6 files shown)


rust/agama-lib/share/profile.schema.json
```diff
@@ -164,6 +164,66 @@
                       "mesh",
                       "ap"
                     ]
+                  },
+                  "hidden": {
+                    "title": "Indicates that the wifi network is not broadcasting it's SSID",
```
imobachgs, line 169: Should it be "wifi", "Wi-Fi", wireless... In this schema we have been using "wireless" but we can change it. Just for consistency.

imobachgs, line 169:
```suggestion
                    "title": "Indicates that the wifi network is not broadcasting its SSID",
```

rust/agama-server/src/network/model.rs
```diff
@@ -1105,6 +1186,108 @@ impl TryFrom<&str> for SecurityProtocol {
...
+    Wep104,
+    Tkip,
+    Ccmp,
+}
+
+#[derive(Debug, Error)]
+#[error("Invalid group algorithm: {0}")]
+pub struct InvalidGroupAlgorithm(String);
+
+impl FromStr for GroupAlgorithm {
```
imobachgs, line 1201: For some time, I have been thinking whether we should use [strum](https://docs.rs/strum/latest/strum/) (see the [EnumString](https://docs.rs/strum/latest/strum/derive.EnumString.html) example) to avoid all those trivial conversions.

I am not sure about how error handle would be, but I think it is worth having a closer look.

In any case, it is out of the scope of this PR.

rust/agama-server/src/network/nm/dbus.rs
```diff
@@ -1656,6 +1751,51 @@ mod test {
         assert_eq!(wep_key0, "5b73215e232f4c577c5073455d");
         let wep_key1: &str = security.get("wep-key1").unwrap().downcast_ref().unwrap();
         assert_eq!(wep_key1, "hello");
+
+        let group_algorithms: &zvariant::Array =
+            security.get("group").unwrap().downcast_ref().unwrap();
+        let group_algorithms: Vec<GroupAlgorithm> = group_algorithms
+            .get()
+            .iter()
+            .map(|x| x.downcast_ref::<str>().unwrap())
```
imobachgs, line 1760: Those `unwraps` are dangerous because is something goes wrong, the whole service would go down. Perhaps you might want to split this chain into smaller functions.

imobachgs (commented):
In general it looks good, just a few comments. Thanks!

rust/agama-server/src/network/nm/dbus.rs
jcronenberg: This is the testing code

rust/agama-server/src/network/model.rs
jcronenberg: Totally agree, strum would make this a lot simpler :slightly_smiling_face:

rust/agama-lib/share/profile.schema.json
jcronenberg: I think `wireless` is fine. I changed the descriptions accordingly.

rust/agama-server/src/network/nm/dbus.rs
imobachgs: Ouch, Friday afternoon, you know 😅

imobachgs (approved):

rust/agama-lib/share/profile.schema.json
imobachgs: OK, we will plan for that.

Issue #1609, opened 2024-09-16
Boot local option in Agama doesn't work on UEFI


lkocman:
Hello folks

(link to bug in case it is not an expected behavior https://bugzilla.suse.com/show_bug.cgi?id=1230595)

all of our **test cases where the boot priority has install media as the first are blocked**. Selecting Boot from local in Agama fails to boot system.

If you change boot order to hardrive first, then installed system boots fine . This confirms that system was properly installed by agama. So it's literally just invalid boot menu entry "Boot from local" in agama-installer live.

Boot instructions for local boot from agama-installer
![image](https://github.com/user-attachments/assets/40953893-5ad1-41d8-b54e-2b6bf8cd6459)

Legacy
![image](https://github.com/user-attachments/assets/294b997c-3c0c-4de2-b849-9b4ab20b8f60)

UEFI
![image](https://github.com/user-attachments/assets/217ea09b-ef94-4460-ac2f-e5d39e77b65e)

Overview can be seen here
https://openqa.opensuse.org/tests/overview?distri=opensuse&version=16.0&build=36.48&groupid=129

lkocman:
Here I see that there are some thumbs down on the "exit" solution for the second boot device too
https://superuser.com/questions/1680733/how-to-boot-from-local-disk-using-grub2-during-a-pxe-boot

But I guess the second boot device could be for example not a HDD.

lkocman:
This is the grub2 output after selecting Boot from disk
![image](https://github.com/user-attachments/assets/90f9c469-4ddb-4cb8-934f-e878be5e3b94)

lkocman:
Here's a recording with vm with already installed Leap 16.0 from agama, using efi and agama-iinstaller cd as first boot device. I'll very likely add the efi workaround (for efi variants) to openQA for now.

[video.webm](https://github.com/user-attachments/assets/e231b8bb-4663-4efe-a919-57ae444d8843)

lkocman:
@ben-brunner found what's the issue

```TW-installer grub.cfg:
menuentry "Boot from Hard Disk" --class opensuse --class gnu-linux --class gnu --class os {
  if search --no-floppy --file /efi/boot/fallback.efi --set ; then
    for os in opensuse sles caasp ; do
      if [ -f /efi/$os/grub.efi ] ; then
        chainloader /efi/$os/grub.efi
        boot
      fi
    done
  fi
  exit
}
```

vs Agama-installer:
```
[...]

PR #1613, opened 2024-09-16
Script generated boot menu for all arches

lkocman:
Fixes Issue #1609 

- Newly added x86_64 and aarch64 write stock copies from kiwi generated grub.cfg

- Intel uefi boot fix from "Boot from disk item" (gh#openSUSE/agama#1609)

- Keep -x on purpose for future trubleshooting (advice from Marcus)

- Use $kiwi_iname instead of static agama-live reference in all of config-cdroom/* with exception of s390x, where for some reason we use SUSE Linux... string instead

- copy .profile to $dst so we can source it from fixboot scripts. Somehow xorriso looses access to ENV from parent which has access to kiwi_iname and other kwiwi variables

image


lkocman:
Test build https://build.opensuse.org/projects/home:lkocman:branches:openSUSE:Leap:16.0/packages/agama-installer-Leap:openSUSE/repositories/images/binaries

live/README.md
```diff
@@ -121,6 +122,15 @@ The main Kiwi source files are located in the [src](src) subdirectory:
 
 The Live ISO is configured to allow using some features and allow running Agama there.
 
+## GRUB2 Menu
+
+grub.cfg, defining boot menu items of the Agama image, is generated by scripts stored in
+[config-cdroot](https://github.com/openSUSE/agama/tree/master/live/config-cdroot).
+
+Both x86_64 and aarch64 grub.cfg are basically copies of KIWI autogenerated grub.cfg.
+The x86_64 grub.cfg contains EUFI fix for Booting from disk (Issue #1609).
```
imobachgs, line 131:
```suggestion
The x86_64 grub.cfg contains UEFI fix for Booting from disk (Issue #1609).
```

imobachgs (commented):
@jreidinger is our expert on the bootloader, so I would appreciate his review. From my POV the changes make sense.

live/config-cdroot/fix_bootconfig
```diff
@@ -0,0 +1,95 @@
+#! /bin/bash
```
lkocman, line 1: This is incorrectly placed, I suppose that we need to have only arch specific versions here. I'll move it to the right directory.

live/src/fix_bootconfig
```diff
@@ -13,12 +13,18 @@
...
 #    it zIPL bootable on s390x.
 #
+set -x
 
-# KIWI config
-test -f .profile && . .profile
 
 dst=${1#iso:}
 
+# KIWI config
```
lkocman, line 21: It's important to do this after setting $dst. I was not able to get acccess to $kiwi_* variables (set in /.profile) fix_bootconfig.$arch . So I simply copied the .profile into the $dst from where the xorriso -> fix_bootconfig$ arch was executed.

This was working (vars inherited, without copying the file) if the fix_bootconfig was called outside xorriso.

live/src/fix_bootconfig
```diff
@@ -13,12 +13,18 @@
...
+# Copy it to $dst from where it can be later read by xorriso
+if [ -f .profile ]; then
+    source .profile
+    cp .profile "$dst"
+fi
+
 arch=`uname -m`
 
 # removing live root setting from command line - it's in /etc/cmdline.d instead
```
lkocman, line 30: I think this is useless now since we have "static" bootconfig (with expansion of $kiwi_iname)

live/config-cdroot/fix_bootconfig.ppc64le
```diff
@@ -74,22 +79,22 @@ chmod 644 $boot_dir/{initrd,linux}
...
-menuentry 'Agama Live' --class os --unrestricted {
+menuentry "$kiwi_iname" --class os --unrestricted {
   echo 'Loading kernel...'
   linux /boot/ppc64le/linux
   echo 'Loading initrd...'
   initrd /boot/ppc64le/initrd
 }
 
-menuentry 'Check Installation Media -- Agama Live' --class os --unrestricted {
+menuentry "Check Installation Media -- $kiwi_iname" --class os --unrestricted {
```
lkocman, line 97: I felt that it would be nice to cosistently use $kiwi_iname across all arches. I did not dare to touch s390x.

live/config-cdroot/fix_bootconfig.x86_64
```diff
@@ -0,0 +1,110 @@
...
+menuentry "Mediacheck" --class os --unrestricted {
+    set gfxpayload=keep
+    echo Loading kernel...
+    linux (\$root)/boot/x86_64/loader/linux mediacheck=1 plymouth.enable=0 \${isoboot}
+    echo Loading initrd...
+    initrd (\$root)/boot/x86_64/loader/initrd
+}
+menuentry "Boot from Hard Disk" --class opensuse --class gnu-linux --class gnu --class os {
+  if search --no-floppy --file /efi/boot/fallback.efi --set ; then
+    for os in opensuse sles leap; do
```
lkocman, line 91:
```suggestion
for os in opensuse sles leap boot; do
```

lkocman, line 91: Seems like somebody exchanged opensuse for leap in the past ... so now [tumbleweed based image](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-9.0.0-openSUSE-Build19.22.iso   has leap too :-)

Screenshot from 2024-09-17 09-00-00

live/src/fix_bootconfig
lkocman, line 30:
```suggestion
```

lkocman, line 30:
```suggestion
```

lkocman, line 30:
```suggestion


```

lkocman, line 30: I don't think that this block is needed since we have now static grub.cfg

live/config-cdroot/fix_bootconfig.aarch64
```diff
@@ -0,0 +1,102 @@
...
+}
+menuentry "Mediacheck" --class os --unrestricted {
+    set gfxpayload=keep
+    echo Loading kernel...
+    linux (\$root)/boot/aarch64/loader/linux mediacheck=1 plymouth.enable=0 \${isoboot}
+    echo Loading initrd...
+    initrd (\$root)/boot/aarch64/loader/initrd
+}
+menuentry "Boot from Hard Disk" --class opensuse --class gnu-linux --class gnu --class os {
+  exit
```
jreidinger, line 90: do you test if that exit works properly on arm? does it move to the next entry in boot menu?

lkocman: Hello Josef, this was a copypaste from the Arm agama-installer without any modifications.

live/src/fix_bootconfig
```diff
@@ -13,12 +13,18 @@
 #    This script (a) adjusts the boot config, (b) runs xorriso, and (c) makes
 #    it zIPL bootable on s390x.
 #
+set -x
```
jreidinger, line 16: maybe we can be even more strict and use `ex` ?

lkocman: Works with me.

live/config-cdroot/fix_bootconfig.aarch64
lkocman: but nope we currently only test on intel in openQA and I have no arm device at hand.

lkocman:
the USB based legacy boot (with usb being the first boot device) still has struggle but that was also before this PR.
https://openqa.opensuse.org/tests/4485755#step/grub_test/5

live/src/fix_bootconfig
lkocman, line 16:
```suggestion
set -ex

jreidinger (approved):

PR #1617, opened 2024-09-18
feature(storage): config support for current size

joseivanlopez:
Add support to the storage config for resizing a partition using its current size as minimum or maximim limit, see [auto_storage.md](https://github.com/openSUSE/agama/blob/master/doc/auto_storage.md#specifying-the-size-of-a-device) document.

For growing a partition as much as possible:

~~~json
{
  "partitions": [
    {
      "search": "/dev/vda1",
      "size": {
        "min": "current"
      }
    }
  ]
}
~~~ 

For shrinking a partition as much as possible:

~~~json
{
  "partitions": [
    {
      "search": "/dev/vda1",
      "size": {
        "min": 0,
        "max": "current"
      }
    }
  ]
}
~~~ 

For omittied size, the default size indicated by the product is used for a new partition. If the partition already exists (has a *search*), then the current device size is used as default (i.e., the device is not resized at all).


ancorgs (approved):
LGTM

PR #1635, opened 2024-09-24
Update the size information at auto_storage.md

ancorgs:
https://github.com/openSUSE/agama/pull/1617 introduced some changes on the acceptable values for sizes and how they are managed. To be precise, it affects the special values like "default" (that is not longer considered an option) and "current".

It also changes the behavior when the size is omitted for a partition or logical volume.

The current PR updates the documentation to reflect the changes introduced at the mentioned pull request.


joseivanlopez (approved):
LGTM

PR #1639, opened 2024-09-25
refactor(web): drop the old HTTPClient

imobachgs:
We recently moved to [TanStack Query](https://tanstack.com/query/latest), completely reimplementing the communication with the back-end. Therefore, the old `HTTPClient` is no longer needed.

This pull request is a big clean-up, adapting the residual use of the `HTTPClient` and removing the unused code.


teclator (approved):
LGTM

PR #1640, opened 2024-09-25
Rename wpa-eap-suite-b192 to wpa-eap-suite-b-192

cfconrad:
## Problem

* `NetworkManger` complain if we use `wpa-eap-suite-b192` as `802-11-wireless-security.key-mgmt` [¹](https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html) with:
```
Could not update the network configuration: D-Bus service error: org.freedesktop.NetworkManager.Settings.Connection.InvalidProperty: 802-11-wireless-security.key-mgmt: 'wpa-eap-suite-b192' is not a valid value for the property
```

I hope it's not to late to change it everywhere. If it is to late, we could only handle `WPA3Only` in `rust/agama-server/src/network/nm/model.rs` properly.

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -152,7 +152,7 @@
                       "wpa-psk",
                       "sae",
                       "wpa-eap",
-                      "wpa-eap-suite-b192"
+                      "wpa-eap-suite-b-192"
                     ]
                   },
                   "ssid": {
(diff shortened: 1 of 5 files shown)


imobachgs (commented):
Please, add an entry in the `changes` file (adding a reference to `gh#agama-project/agama#1640`). Otherwise, it LGTM.

imobachgs (approved):

imobachgs (approved):

Issue #1629, opened 2024-09-23
Still missing Swedish 23th September 2024


bittin:
testing now with todays 23th Septembers iso: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-openSUSE.iso , more Swedish now at least, but its not fully complete yet but maybe there is some strings thats not on Weblate yet ?

imobachgs:
Please, put some example so we can have a look.

imobachgs:
Oh, and the link you posted is generic; it does not tell us which build you are using.

bittin:
@imobachgs https://imgur.com/a/cIZXZDH here is a couple of examples that still have text in english, sorry still sick so was a bit lazy and that is the build from today so Agama 10 25.3

imobachgs:
The untranslated strings are coming from the backend and they are translated in weblate.

imobachgs:
#1642, #1643 and #1644 should fix those problems. There is only one glitch while displaying the language change progress, but that's a different topic we might need to address with some serious refactoring.

imobachgs:
Oh, I almost forgot: the pattern descriptions in the software section are not translated because the repository does not contain any translation to Swedish.

bittin:
This is still not fixed, just tried the Leap 16 Pre Alpha image

alright where to translate the pattern descriptions?

PR #1642, opened 2024-09-27
fix: handle language change in users and storage

imobachgs:
## Problem

Some parts of the UI are not properly translated (see #1629).

## Solution

Properly refresh the issues in the users and storage services.

## Testing

- Tested manually

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri Sep 27 09:23:40 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Handle language change in users and storage (gh#agama-project/agama#1642).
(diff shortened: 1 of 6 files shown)


service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Sep 27 09:23:40 UTC 2024 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Handle language change in users and storage (gh#agama-project/agama#1642).
```
mchf, line 4: NP: well ... instead of "language change" i'd use something like "switching UI language" ... to make obvious that the fix addresses switching of the language the UI uses instead of default EN

mchf (approved):
except one NP it LGTM

mchf (approved):
LeBTM (= looks even better to me ;-)

PR #1643, opened 2024-09-27
fix(rust): localize the keymaps descriptions

imobachgs:
## Problem

Agama's UI always displays the keymaps descriptions in English.

## Solution

Translate keymaps descriptions at API level.

## Testing

- Tested manually


mchf (approved):
LGTM

PR #1644, opened 2024-09-27
fix: translate storage component when UI language changes

imobachgs:
## Problem

Some parts of the UI are not properly translated when the language of the UI changes (see #1629).

## Solution

This is a fix for storage components which consist on not calling `_` at module level.

## Testing

- Tested manually


teclator (approved):
LGTM

PR #1645, opened 2024-09-27
fix(web): fix localization of overview headers

imobachgs:
## Problem

Related to #1629.

Overview headers are not translated.

## Solution

Do not call `_()` at module level (see #1643).

## Testing

- *Tested manually*


joseivanlopez (approved):
LGTM

Issue #1335, opened 2024-06-14
Wrong UTC offset in the list of timezones


imobachgs:
Reported by @jreidinger:

> many countries has UTC even when their timezone is like UTC+2, so probably something is wrong at
> source data or during processing
>
> but I see when search e.g. for UTC-7

imobachgs:
The offset is calculated at the client side: https://github.com/openSUSE/agama/blob/f9c0c0d24390825cbe36cb54950a93113828d616/web/src/utils.js#L355.

mvidner:
@jreidinger I cannot see the problem now. Can you guide me to find it?
![agama-tzofs](https://github.com/openSUSE/agama/assets/102056/4b189bcb-db07-4d50-b9ee-8ea52e6c7481)

(BTW note how it shows the _current_ UTC offset, applying the daylight savings time.)

imobachgs:
It happens only in some cases (I do not know how many). But some timezones are reported as UTC when they are not. Two examples: Atlantic/Canary (UTC+1) and Americas/Adak (UTC-9).

dgdavid:
Just in case it might help, would worth having a look at https://github.com/date-fns/tz

Not only for usin it (I'd do if it make our code work as expected) but also for learning from its source code and tests, 

https://github.com/date-fns/tz/tree/main/src/tzOffset

imobachgs:
The problem is fixed in #1646.

PR #1646, opened 2024-09-30
fix(web): properly display the UTC offset

imobachgs:
## Problem

The UTC offset is not properly calculated, as shown in #1335.

## Solution

* Reimplement the UTC calculation.
* Rely on [@date-fns/tz](https://github.com/date-fns/tz).

## Screenshot

Captura desde 2024-09-30 12-25-24

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -107,6 +107,7 @@
     "webpack-dev-server": "^5.0.4"
   },
   "dependencies": {
+    "@date-fns/tz": "^1.1.2",
     "@icons-pack/react-simple-icons": "^10.0.0",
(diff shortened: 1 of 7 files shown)


web/package.json
```diff
@@ -107,6 +107,7 @@
     "webpack-dev-server": "^5.0.4"
   },
   "dependencies": {
+    "@date-fns/tz": "^1.1.2",
```
lslezak, line 110: Ah, it uses the Intl browser API so no hardcoded database included, nice! :+1:

lslezak (approved):

PR #1650, opened 2024-09-30
feat(rust): add tasks using the xtask pattern

imobachgs:
While working on [Agama's website](https://agama-project.github.io/) we wanted to include some documentation about the CLI. Fortunately, [clap](https://crates.io/crates/clap) includes support for generating markdown documentation, man pages, shell completions, etc.

After some research, we found out that the Rust community is adopting the [xtask pattern](https://github.com/matklad/cargo-xtask), where you write those tasks using Rust code. And it plays really well with clap, as we can generate those artifcats from the sources.

However, we needed to do some refactoring to expose the CLI as a `lib` crate so we can use it in our tasks.

This pull request includes the following tasks: `manpages`, `completions` and `markdown`. The first step should be to take advantage of them and generate the artifacts at runtime. It is open for discussion whether we should do the same with the OpenAPI docs.

<details>
<summary>Generating the man pages</summary>

```shell
$ cargo xtask manpages                                                                                                                                                              at 15:05:46 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/xtask manpages`
Generate manpages documentation at out/man

$ ls out/man/                                                                                                                                                                       agama-auth-login.1
agama-auth-logout.1
agama-auth-show.1
agama-auth.1
agama-config-edit.1
agama-config-load.1
agama-config-show.1
agama-config.1
agama-download.1
agama-install.1
agama-logs-list.1
agama-logs-store.1
agama-logs.1
agama-probe.1
agama-profile-autoyast.1
agama-profile-evaluate.1
agama-profile-import.1
agama-profile-validate.1
agama-profile.1
agama-questions-answers.1
agama-questions-ask.1
agama-questions-list.1
agama-questions-mode.1
agama-questions.1
agama.1
```
</details>

## Todo

- [x] Document usage
- [x] Add a shell-completion package


rust/xtask/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+    use crate::create_output_dir;
+
+    /// Generate auto-completion code for common shells.
+    pub fn generate_complete() -> std::io::Result<()> {
+        let out_dir = create_output_dir("shell")?;
+
+        let mut cmd = Cli::command();
+        clap_complete::generate_to(aot::Bash, &mut cmd, "agama", &out_dir)?;
+        clap_complete::generate_to(aot::Fish, &mut cmd, "agama", &out_dir)?;
+        clap_complete::generate_to(aot::Zsh, &mut cmd, "agama", &out_dir)?;
```
jreidinger, line 20: we should probably package each completion package separately with own supplements so on live cd it will install only shell ones.

jreidinger (approved):

PR #1651, opened 2024-09-30
Update links to the agama-weblate repository

lslezak:
## Problem

- The `agama-weblate` repository has been also moved

## Solution

- Update the references in the GitHub Actions to use generic `github.repository_owner` (the actions use `!github.event.repository.fork` condition so they should never run outside of the `project-agama` , but this is more flexible)
- Update the links in the documentation

## Notes

- I was thinking about using relative URLs in the documentation, but then it would not work in forks (unless you fork the `agama-weblate` repository as well). Absolute URLs are probably better here, I do not expect we move the location anytime soon. :smiley:

### DIFF
--- a/.github/workflows/weblate-merge-po.yml
+++ b/.github/workflows/weblate-merge-po.yml
@@ -51,7 +51,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: agama-weblate
-          repository: openSUSE/agama-weblate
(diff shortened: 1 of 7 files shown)


imobachgs (changes requested):
It is `agama-project`, not `project-agama`.

imobachgs (approved):

PR #1652, opened 2024-10-01
feature(storage): config support for generating physical volumes

joseivanlopez:
Add support to the storage config for using a *generate* section for physical volumes.

~~~json
"physicalVolumes": [
  { "generate": ["first-disk", "second-disk"] }
]
~~~

~~~json
"physicalVolumes": [
  {
    "generate": {
      "targetDevices": ["first-disk"],
      "encryption": {
        "luks2": { "password": "12345" }
      }
    }
  }
]
~~~

The *generate* section allows to indicate the target devices in which to automatically create physical volumes, if needed. Agama will try to create physical volumes to allocate the defined logical volumes. If an *encryption* section is provided by the *generate* section, then the new physical volumes will be encrypted.

Notes:

* For now, *physicalVolumes* can contain either device aliases or a *generate*. In the future it will allow both.
* If there is more than one *generate* in the list of physical volumes, then only the first one will be considered (the rest ones are ignored). Note that the schema does not complain.

### DIFF
--- a/rust/agama-lib/share/examples/storage/generate_pvs.json
+++ b/rust/agama-lib/share/examples/storage/generate_pvs.json
@@ -1,0 +1,48 @@
+{
+  "storage": {
+    "drives": [
+      {
+        "alias": "first-disk"
+      },
+      {
+        "partitions": [
+          {
+            "alias": "pv1",
+            "id": "lvm",
+            "size": { "min": "10 GiB" }
+          }
+        ]
+      }
+    ],
+    "volumeGroups": [
+      {
+        "name": "system",
+        "physicalVolumes": ["pv1"],
+        "logicalVolumes": [
+          {
+            "filesystem": { "path": "/" }
+          }
+        ]
+      },
+      {
+        "name": "logs",
+        "physicalVolumes": [
+          { "generate": ["first-disk"] }
+        ]
+      },
+      {
+        "name": "data",
+        "physicalVolumes": [
+          {
+            "generate": {
+              "targetDevices": ["first-disk"],
+              "encryption": {
+                "luks2": { "password": "12345" }
+              }
+            }
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/service/lib/agama/storage/config_conversions/from_json_conversions/volume_group.rb
+++ b/service/lib/agama/storage/config_conversions/from_json_conversions/volume_group.rb
@@ -20,6 +20,7 @@
 # find current contact information at www.suse.com.
 
 require "agama/storage/config_conversions/from_json_conversions/base"
+require "agama/storage/config_conversions/from_json_conversions/encryption"
 require "agama/storage/config_conversions/from_json_conversions/logical_volume"
 require "agama/storage/configs/volume_group"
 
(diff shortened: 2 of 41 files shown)


rust/agama-lib/share/examples/storage/generate_pvs.json
```diff
@@ -0,0 +1,39 @@
...
+            "size": { "min": "10 GiB" }
+          }
+        ]
+      }
+    ],
+    "volumeGroups": [
+      {
+        "name": "system",
+        "physicalVolumes": [
+          "pv1",
```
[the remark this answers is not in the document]
ancorgs: I suspect it would be easy, but I have to double check.

service/lib/agama/storage/config_filesystem_solver.rb
```diff
@@ -0,0 +1,106 @@
...
+      # @return [Configs::Filesystem]
+      def default_filesystem(path = nil)
+        config_builder.default_filesystem(path)
+      end
+
+      # Default btrfs config defined by the product.
+      #
+      # @param path [String, nil]
+      # @return [Configs::Btrfs]
+      def default_btrfs(path = nil)
```
ancorgs, line 96: I didn't check much, so just asking at first glance... cannot this return nil if the default filesystem for the given path is not btrfs?

service/lib/agama/storage/config_conversions/from_json_conversions/volume_group.rb
```diff
@@ -70,6 +61,40 @@ def convert_extent_size
             Y2Storage::DiskSize.new(value)
           end
 
+          # @return [Array<String>, nil]
+          def convert_physical_volumes_devices
+            generate_json = physical_volume_generate_json
+            return unless generate_json
+
+            generate_json.dig(:generate, :targetDevices)
```
ancorgs, line 69: Didn't we agree to also implement a short version like this for those cases in which we don't need encryption settings?

```json
"volumeGroups": [
  {
    "name": "system",
    "physicalVolumes": [
      { "generate": ["vda"] }
    ],
  }
]
```

Kind of similar to the short versions of the `generate` for partitions and LVs.

rust/agama-lib/share/examples/storage/generate_pvs.json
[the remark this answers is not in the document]
ancorgs: I checked and it would be possible by modifying the yast2-storage-ng algorithm a bit. But we decided to leave it for later, as part of a separate task internally tracked at https://trello.com/c/XAaoYdCL/578-storage-profile-improve-support-for-generating-lvm-physical-volumes

service/lib/agama/storage/config_filesystem_solver.rb
joseivanlopez: If the product does not define Btrfs properties for that volume, then it generates a default Btrfs config (i.e., snapshots to `false`, etc). Theoretically, it would be possible to select Btrfs for any path that admits it.

joseivanlopez: But ... I just realized that we are not checking if the selected filesystem type is suitable for that path.

rust/agama-lib/share/examples/storage/generate_pvs.json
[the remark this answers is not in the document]
joseivanlopez: I have adapted the schema to allow either aliases or 'generate'.

ancorgs (approved):
LGTM

PR #1655, opened 2024-10-03
Automatic generation of physical volumes

ancorgs:
#1652 adds support for configuring the automatic generation of LVM physical volumes. Read there the details about how the feature works.

This pull request implements a first version of the functionality described there, including a couple of unit tests.

Depends on https://github.com/yast/yast-storage-ng/pull/1392


service/lib/y2storage/agama_proposal.rb
```diff
@@ -206,23 +206,28 @@ def drives_with_empty_partition_table(devicegraph)
...
-    def partitions_for_clean
-      # The current logic is quite trivial, but this is implemented as a separate method because
-      # some extra logic is expected in the future (eg. considering partitions on pre-existing
-      # RAIDs and more stuff). See the equivalent method at DevicegraphGenerator.
-      planned_devices.partitions
+    # @return [Array<String>] names of partitionable devices
+    def disks_for_clean
+      return drives_names if config.boot_device.nil? || drives_names.include?(config.boot_device)
+
+      drives_names + [config.boot_device]
```
joseivanlopez, line 215: NP: maybe more direct with `[drive_names, config.boot_device].flatten.compact.uniq` ?

service/lib/y2storage/proposal/agama_vg_planner.rb
```diff
@@ -28,26 +28,57 @@ module Proposal
...
+      #
+      # @param planned [Planned::LvmVg]
+      # @param config [Agama::Storage::Configs::VolumeGroup]
+      def configure_pvs_encryption(planned, config)
+        enc = config.physical_volumes_encryption
+        return unless enc
+
+        planned.pvs_encryption_method = enc.method
+        planned.pvs_encryption_password = enc.password
+        planned.pvs_encryption_pbkdf = enc.pbkd_function
```
joseivanlopez, line 79: Are the rest of encryption attributes ignored (`#label`, `#cipher` and `#key_size`)?

ancorgs: I'm not sure whether label makes sense. We are going to generate several PVs and labels should be unique.

For the other two, I overlooked because `PlannedVg` didn't support them. So it would need changes both here and at yast2-storage-ng. What about making a separate card/PBI out of it?

service/lib/y2storage/agama_proposal.rb
ancorgs: I changed into something similar to your suggestion

service/lib/y2storage/proposal/agama_vg_planner.rb
ancorgs: Added as another point to https://trello.com/c/XAaoYdCL/578-storage-profile-improve-support-for-generating-lvm-physical-volumes

joseivanlopez (approved):
LGTM

PR #1659, opened 2024-10-08
try to optimize getting all zfcp controllers

jreidinger:
## Problem

Getting zfcp controllers is very slow call.

## Solution

Try thread approach without dbus calls for this specific http call.

## Testing

original time on our testing s390

```
real	0m10.851s
user	0m0.005s
sys	0m0.005s
```

### DIFF
--- a/rust/agama-lib/src/storage/client/zfcp.rs
+++ b/rust/agama-lib/src/storage/client/zfcp.rs
@@ -71,9 +71,9 @@ pub async fn get_luns(controller_id: String, wwpn: String) -> Result<Vec<String>
 ///
 /// `controller_id`: controller id
 pub async fn get_luns_map(
-    controller_id: &str,
+    controller_id: String,
 ) -> Result<HashMap<String, Vec<String>>, ServiceError> {
-    let wwpns = get_wwpns(controller_id).await?;
+    let wwpns = get_wwpns(controller_id.as_str()).await?;
     let mut tasks = vec![];
     for wwpn in wwpns {
         tasks.push((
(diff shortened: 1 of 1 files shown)


rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -34,6 +34,60 @@ use crate::{
...
+/// `controller_id`: controller id
+pub async fn get_luns_map(
+    controller_id: &str,
+) -> Result<HashMap<String, Vec<String>>, ServiceError> {
+    let wwpns = get_wwpns(controller_id).await?;
+    let mut tasks = vec![];
+    for wwpn in wwpns {
+        tasks.push((
+            wwpn.clone(),
+            tokio::spawn(get_luns(controller_id.to_string(), wwpn.clone())),
```
imobachgs, line 81: Instead of spawning all those processes, I would suggest giving [join_all](https://docs.rs/futures/latest/futures/future/fn.join_all.html) a try.

imobachgs: Tokio has [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html), which might be interested too.

jreidinger: thanks for suggestion. I plan to have it as experiment first and this tokio spawn is suggested way in that linkedin rust training, so it is the first that I would like to give a try.

jreidinger: and testing shows that my change somehow does not work, but as it is just learning/research project I will put it a bit on hold and focus on planned stuff and return to it probably next week.

jreidinger: ok, I cannot sleep without debugging it...so now it is fixed and time is improved to 6.3 second..so already it goes down from 10.8 to 6.3 and I think it can be even better if I parallel also getting wwpns for controller. And then I can start playing with join_all and JoinSet

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -153,25 +162,40 @@ impl<'a> ZFCPClient<'a> {
...
-                        luns_map: luns_map, // try to use optimized thred version here
                     },
                 ))
             }
         }
-        Ok(devices)
+        let mut tasks = vec![];
+        for (_path, partial_controller) in &devices {
+            tasks.push(
+                tokio::spawn(get_luns_map(partial_controller.channel.clone()))
```
jreidinger, line 184: with this additional paralellization time is reduced to 5.3 seconds. So basically for two controllers with 3 LUNS in total it is gets to half of time. I expect that for bigger systems it will be even more significant.

PR #1662, opened 2024-10-10
PoC for the online demo

lslezak:
# Prototype for the online demo

This is a prototype for building a completely static Agama demo.

We have some screenshots on the home page and in the documentation but it would be better to have some real site with working Agama so people can try it live.

The problem is that Agama needs a real system for probing hardware and so on. But we could simply run it on some real system, dump the collected REST API responses and then just replay them on the demo site.

That means the demo would be read-only and all changes done by user would be ignored. But I think that for the demo purposes it would be good enough.

## Deployed demo

You can try the deployed code at https://agama-online-demo.surge.sh.

There are still some small glitches (changing the installed product gets stuck, you need to reload the page manually) but in general it works fine.

## How to reproduce the build

- First compile with AGAMA_DEMO=record
- Login to Agama, click all sections, go to all subpages to trigger all HTTP requests
- From development console save the `window.agama_demo` value to `src/api/demo.json` file
- Compile with AGAMA_DEMO=replay to build a static site with the recorded data

## Improvements

- Handle better some corner cases (starting installation, product selection...)
- It would be nice to collect the REST API responses automatically, clicking around in a real browser is error prone.
- It would be nice to define the target testing system or even have a script which creates the same virtual machine. If you collect the data on a different machine you will get different proposal (very likely with different partitioning). This should be a reproducible process.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -20,7 +20,8 @@
  * find current contact information at www.suse.com.
  */
 
-import React from "react";
+import React, { useState } from "react";
+import { Alert, AlertActionCloseButton } from "@patternfly/react-core";
 import { Navigate, Outlet, useLocation } from "react-router-dom";
 import { ServerError } from "~/components/core";
 import { Loading, PlainLayout } from "~/components/layout";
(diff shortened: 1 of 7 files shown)


lslezak:
Closing, this is not relevant right now. But I'll look at it again later...

PR #1666, opened 2024-10-11
fix(web): fix the 'Cancel' button in the products selection

imobachgs:
Workaround the "404 Not Found" error. In production, and in that scenario, calling `navigate(-1)` adds a '-1' to the URL. We might need to do further research, but this patch should fix the issue.


ancorgs (approved):
LGTM

ancorgs (approved):
Still LGTM after the merge with master.

PR #1667, opened 2024-10-14
Move to json

jreidinger:
Unify formats to use JSON with only exception beeing products files as it is intended to be PM oriented and rest is user oriented. Also we need translations and multiline content from yaml in products.


imobachgs (commented):
Please add an entry to the changes file. Otherwise, LGTM.

imobachgs (approved):

PR #1670, opened 2024-10-14
feature(storage): config conversion to JSON

joseivanlopez:
Add conversion of storage config to JSON format.

This will be needed for reporting the "solved" config once the *guided proposal* is replaced by the *agama proposal*.

### DIFF
--- a/service/lib/agama/storage/config_conversions/to_json_conversions/encryption.rb
+++ b/service/lib/agama/storage/config_conversions/to_json_conversions/encryption.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2024] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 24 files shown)


service/lib/agama/storage/config_conversions/to_json_conversions/encryption.rb
```diff
@@ -0,0 +1,87 @@
...
+
+            super || convert_swap_encryption
+          end
+
+        private
+
+          # @see Base#conversions
+          def conversions
+            method = config.method
+
```
ancorgs, line 52: Don't we have support for the method `tpm_fde`?

ancorgs (commented):
It generally looks good. But I miss one encryption method that I think should be supported (it is supported at the current storage UI).

service/lib/agama/storage/config_conversions/to_json_conversions/encryption.rb
joseivanlopez: Not yet. It is not supported in the schema neither. The *read_only* property for Btrfs is another missing property. If you don't mind, I will add them in a separate PR, because they have nothing to do with the JSON conversion.

ancorgs: Ok. Let's approve this then to move on.

ancorgs (approved):
LGTM

PR #1673, opened 2024-10-16
feat: add support to run pre/post scripts

imobachgs:
This is the first implementation of pre/post-installation scripts. They are not as capable as their AutoYaST counterparts, but the idea is to evolve them in the future.

## Specifying a script

> [!WARNING]
> JSON does not support multiline values (a.k.a. blocks) but Jsonnet does it. Perhaps we should consider using Jsonnet in the `agama config edit` command.

```jsonnet
{
  scripts: {
    pre: [
      {
         name: "say-hi",
         body: |||
           #!/bin/bash
           echo "before starting the installation"
         |||
      }
    ]
  }
}
```

> [!NOTE]
> What about having a plain list where each script has a `type` (or `group`) like `pre`, `post`, etc.?

## When scripts get executed

* Pre-scripts: when loading a profile. We plan to do [some improvements](https://trello.com/c/PiOGzXSn/584-agama-running-pre-scripts-in-the-right-place) to support the single-product case properly.
* Post-scripts: after the installation.

## Implementation details

The scripting support is written in Rust and does not use the YaST scripts code anymore. We will rewrite the Manager in Rust (at least the Agama-specific part) in the future, and the scripting support will be part of that.

## Future

Agama scripts will gain additional features soon:

* Running in a `chroot`.
* Better error handling if the shebang line is not present.
* Running after the first boot.

## Pending tasks

- [x] Extend the JSON schema
- [x] Copy the resulting files to the installed system
- [x] Add some logging
- [x] Clean-up the old scripts when reloading a profile

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -54,9 +54,7 @@
           "items": {
             "type": "object",
             "additionalProperties": false,
-            "required": [
-              "id"
-            ],
+            "required": ["id"],
             "properties": {
               "id": {
                 "title": "Connection ID",
@@ -79,22 +77,12 @@
               "method4": {
                 "title": "IPv4 configuration method",
                 "type": "string",
-                "enum": [
-                  "auto",
-                  "manual",
-                  "link-local",
-                  "disabled"
-                ]
+                "enum": ["auto", "manual", "link-local", "disabled"]
               },
               "method6": {
                 "title": "IPv6 configuration method",
                 "type": "string",
-                "enum": [
-                  "auto",
-                  "manual",
-                  "link-local",
-                  "disabled"
-                ]
+                "enum": ["auto", "manual", "link-local", "disabled"]
               },
               "gateway4": {
                 "title": "Connection gateway address",
@@ -162,12 +150,7 @@
                   "mode": {
                     "title": "Wireless network mode",
                     "type": "string",
-                    "enum": [
-                      "infrastructure",
-                      "adhoc",
-                      "mesh",
-                      "ap"
-                    ]
+                    "enum": ["infrastructure", "adhoc", "mesh", "ap"]
                   },
                   "hidden": {
                     "title": "Indicates that the wireless network is not broadcasting its SSID",
@@ -176,15 +159,12 @@
                   "band": {
                     "title": "Frequency band of the wireless network",
                     "type": "string",
-                    "enum": [
-                      "a",
-                      "bg"
-                    ]
+                    "enum": ["a", "bg"]
                   },
                   "channel": {
                     "title": "Wireless channel of the wireless network",
                     "type": "integer",
-                    "minimum" : 0
+                    "minimum": 0
                   },
                   "bssid": {
                     "title": "Only allow connection to this mac address",
@@ -195,12 +175,7 @@
                     "items": {
                       "title": "A list of group/broadcast encryption algorithms",
                       "type": "string",
-                      "enum": [
-                        "wep40",
-                        "wep104",
-                        "tkip",
-                        "ccmp"
-                      ]
+                      "enum": ["wep40", "wep104", "tkip", "ccmp"]
                     }
                   },
                   "pairwiseAlgorithms": {
@@ -208,10 +183,7 @@
                     "items": {
                       "title": "A list of pairwise encryption algorithms",
                       "type": "string",
-                      "enum": [
-                        "tkip",
-                        "ccmp"
-                      ]
+                      "enum": ["tkip", "ccmp"]
                     }
                   },
                   "wpaProtocolVersions": {
@@ -219,10 +191,7 @@
                     "items": {
                       "title": "A list of allowed WPA protocol versions",
                       "type": "string",
-                      "enum": [
-                        "wpa",
-                        "rsn"
-                      ]
+                      "enum": ["wpa", "rsn"]
                     }
                   },
                   "pmf": {
@@ -369,10 +338,7 @@
                   "peapVersion": {
                     "title": "Which PEAP version is used when PEAP is set as the EAP method in the 'eap' property",
                     "type": "string",
-                    "enum": [
-                      "0",
-                      "1"
-                    ]
+                    "enum": ["0", "1"]
                   },
                   "peapLabel": {
                     "title": "Force the use of the new PEAP label during key derivation",
@@ -406,11 +372,7 @@
           "examples": ["nots3cr3t"]
         }
       },
-      "required": [
-        "fullName",
-        "userName",
-        "password"
-      ]
+      "required": ["fullName", "userName", "password"]
     },
     "root": {
       "title": "Root authentication settings",
@@ -906,7 +868,7 @@
                             "description": "Name of a disk device.",
                             "type": "string",
                             "examples": ["/dev/vda"]
-                            }
+                          }
                         }
                       },
                       {
@@ -967,6 +929,30 @@
       "items": {
         "type": "object"
       }
+    },
+    "scripts": {
+      "title": "User-defined installation scripts",
+      "description": "User-defined scripts to run at different points of the installation",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "pre": {
+          "title": "Pre-installation scripts",
+          "description": "User-defined scripts to run before the installation starts",
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/script"
+          }
+        },
+        "post": {
+          "title": "Post-installation scripts",
+          "description": "User-defined scripts to run after the installation finishes",
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/script"
+          }
+        }
+      }
     }
   },
   "$defs": {
@@ -1013,7 +999,12 @@
           },
           "minItems": 1,
           "maxItems": 2,
-          "examples": [[1024, "current"], ["1 GiB", "5 GiB"], [1024, "2 GiB"], ["2 GiB"]]
+          "examples": [
+            [1024, "current"],
+            ["1 GiB", "5 GiB"],
+            [1024, "2 GiB"],
+            ["2 GiB"]
+          ]
         },
         {
           "title": "Size range",
@@ -1206,8 +1197,22 @@
         {
           "title": "File system type",
           "enum": [
-            "bcachefs", "btrfs", "exfat", "ext2", "ext3", "ext4", "f2fs", "jfs",
-            "nfs", "nilfs2", "ntfs", "reiserfs", "swap", "tmpfs", "vfat", "xfs"
+            "bcachefs",
+            "btrfs",
+            "exfat",
+            "ext2",
+            "ext3",
+            "ext4",
+            "f2fs",
+            "jfs",
+            "nfs",
+            "nilfs2",
+            "ntfs",
+            "reiserfs",
+            "swap",
+            "tmpfs",
+            "vfat",
+            "xfs"
           ]
         },
         {
@@ -1330,7 +1335,15 @@
               },
               "id": {
                 "title": "Partition ID",
-                "enum": ["linux", "swap", "lvm", "raid", "esp", "prep", "bios_boot"]
+                "enum": [
+                  "linux",
+                  "swap",
+                  "lvm",
+                  "raid",
+                  "esp",
+                  "prep",
+                  "bios_boot"
+                ]
               },
               "size": {
                 "title": "Partition size",
@@ -1394,6 +1407,28 @@
     "lvStripeSize": {
       "title": "Stripe size",
       "$ref": "#/$defs/sizeValue"
+    },
+    "script": {
+      "title": "User-defined installation script",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "description": "Script name, to be used as file name",
+          "type": "string"
+        },
+        "body": {
+          "title": "Script content",
+          "description": "Script content, starting with the shebang",
+          "type": "string"
+        },
+        "url": {
+          "title": "Script URL",
+          "description": "URL to fetch the script from"
+        }
+      },
+      "required": ["name"],
+      "oneOf": [{ "required": ["body"] }, { "required": ["url"] }]
     }
   }
 }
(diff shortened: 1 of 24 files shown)


service/lib/agama/storage/finisher.rb
```diff
@@ -208,12 +209,40 @@ def run
...
+      private
+
+        def copy_agama_scripts
+          return unless Dir.exist?(SCRIPTS_DIR)
+
+          Yast.import "Installation"
+          require "fileutils"
+          logs_dir = File.join(Yast::Installation.destdir, "var", "log", "agama")
+          FileUtils.mkdir_p(logs_dir)
+          FileUtils.cp_r(SCRIPTS_DIR, logs_dir)
```
lslezak, line 232: We should document that the user scripts are copied to the logs in the installed system. The user scripts might contain sensitive data (passwords, tokens, keys, certificates,...) and the users should be aware of this when attaching the logs to bugzilla or sending via email.

rust/agama-lib/share/profile.schema.json
```diff
@@ -967,6 +929,30 @@
...
+    },
+    "scripts": {
+      "title": "User-defined installation scripts",
+      "description": "User-defined scripts to run at different points of the installation",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "pre": {
+          "title": "Pre-installation scripts",
+          "description": "User-defined scripts to run before the installation starts",
```
lslezak, line 941: Um, when exactly? Nothing done yet?

I mean later we might add a ` post-partitioning` script so it should be clean whether at this point something is already done by Agama or the system has not been touched yet.

service/test/agama/storage/manager_test.rb
```diff
@@ -315,17 +330,30 @@
...
 
+    context "when scripts artifacts exist" do
+      before do
+        FileUtils.mkdir_p(scripts_dir)
+        FileUtils.touch(File.join(scripts_dir, "test.sh"))
+      end
+
+      it "copies the artifacts to the installed system" do
+        storage.finish
+        expect(File).to exist(File.join(tmp_dir, "mnt", "var", "log", "agama", "scripts"))
```
lslezak, line 353: I'm not sure if the test should really touch the testing system (create and copy files). I do not see much advantage over just to expect to call `FileUtils.cp_r`.

lslezak (approved):
Just some minor comments, looks good to me.

rust/agama-lib/src/scripts/client.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /// Runs user-defined scripts of the given group.
+    ///
+    /// * `group`: group of the scripts to run
+    pub async fn run_scripts(&self, group: ScriptsGroup) -> Result<(), ServiceError> {
+        self.client.post_void("/scripts/run", &group).await
+    }
+
+    pub async fn scripts(&self) -> Result<Vec<Script>, ServiceError> {
```
jreidinger, line 49: documentation missing

rust/agama-lib/src/scripts/client.rs
```diff
@@ -0,0 +1,58 @@
...
+        let scripts = self.client.get("/scripts").await?;
+        Ok(scripts)
```
jreidinger, line 51: cannot it be simple oneliner `self.client.get("/scripts").await` ?

rust/agama-lib/src/scripts/model.rs
```diff
@@ -0,0 +1,215 @@
...
+};
+
+use serde::{Deserialize, Serialize};
+
+use crate::transfer::Transfer;
+
+use super::ScriptError;
+
+#[derive(Debug, Clone, Copy, PartialEq, strum::Display, Serialize, Deserialize)]
+#[strum(serialize_all = "camelCase")]
```
jreidinger, line 36: I will need to read a bit more about strum. First time I met it.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// Represents a script to run as part of the installation process.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Script {
+    /// Script's name.
+    pub name: String,
+    /// Script's body. Either the body or the URL must be specified.
+    pub body: Option<String>,
+    /// URL to get the script from. Either the body or the URL must be specified.
+    pub url: Option<String>,
```
jreidinger, line 51: this looks like bad representation in rust. I kind of expect here enum with either body or url content (which are just marked string). I am just not sure how hard it will be to express it in serde.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -0,0 +1,215 @@
...
+            .truncate(true)
+            .mode(0o500)
+            .open(&path)?;
+
+        if let Some(url) = &self.url {
+            Transfer::get(url, file)?;
+        } else if let Some(body) = &self.body {
+            write!(file, "{}", &body)?;
+        }
+        // FIXME: else: invalid script definition
```
jreidinger, line 96: as said above, should be already catched by json serializer and not here in code where you can rely on valid code.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Removes all the scripts from the repository.
+    pub fn clear(&mut self) {
+        self.scripts.clear();
+    }
+
+    /// Runs the scripts in the given group.
+    ///
+    /// They run in the order they were added to the repository.
```
jreidinger, line 135: I would document here that it does not return error if script failed. and in this case just log error.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -0,0 +1,215 @@
...
+        let path = &tmp_dir.path().join("pre").join("test.log");
+        let body: Vec<u8> = std::fs::read(&path).unwrap();
+        let body = String::from_utf8(body).unwrap();
+        assert_eq!("hello\n", body);
+
+        let path = &tmp_dir.path().join("pre").join("test.err");
+        let body: Vec<u8> = std::fs::read(&path).unwrap();
+        let body = String::from_utf8(body).unwrap();
+        assert_eq!("error\n", body);
+    }
```
jreidinger, line 214: :+1: for tests

rust/agama-lib/src/scripts/settings.rs
```diff
@@ -0,0 +1,45 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct ScriptConfig {
+    /// Script's name.
+    pub name: String,
+    /// Script's body. Either the body or the URL must be specified.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub body: Option<String>,
+    /// URL to get the script from. Either the body or the URL must be specified.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub url: Option<String>,
```
jreidinger, line 44: I still hope that serde can handle if we have it as enum...especially as it is really exclusive as having  both body and url does not make sense.

rust/agama-lib/src/scripts/store.rs
```diff
@@ -0,0 +1,94 @@
...
+            client: ScriptsClient::new(client),
+        }
+    }
+
+    pub fn new_with_client(client: ScriptsClient) -> Result<Self, ServiceError> {
+        Ok(Self { client })
+    }
+
+    pub async fn load(&self) -> Result<ScriptsConfig, ServiceError> {
+        // TODO: Ok(self.client.scripts().await.unwrap())
```
jreidinger, line 41: ? I do not get this todo

rust/agama-lib/src/scripts/store.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl ScriptsStore {
+    pub fn new(client: BaseHTTPClient) -> Self {
+        Self {
+            client: ScriptsClient::new(client),
+        }
+    }
+
+    pub fn new_with_client(client: ScriptsClient) -> Result<Self, ServiceError> {
+        Ok(Self { client })
```
jreidinger, line 37: how it can fail? this looks kind of strange.

rust/agama-lib/src/scripts/store.rs
```diff
@@ -0,0 +1,94 @@
...
+                .await?;
+        }
+
+        for post in &settings.post {
+            self.client
+                .add_script(&Self::to_script(post, ScriptsGroup::Post))
+                .await?;
+        }
+
+        self.client.run_scripts(ScriptsGroup::Pre).await?;
```
jreidinger, line 65: Hmm, I am not sure I like it. Separation of load and store is understandable, but why it runs pre script when store? This looks kind of unexpected.

rust/agama-lib/src/scripts/store.rs
```diff
@@ -0,0 +1,94 @@
...
+            .collect()
+    }
+
+    fn to_script_config(script: &Script) -> ScriptConfig {
+        ScriptConfig {
+            name: script.name.clone(),
+            url: script.url.clone(),
+            body: script.body.clone(),
+        }
+    }
```
jreidinger, line 93: I have to say that this `to_something` looks kind of strange in rust. What is reason to not implement trait Into?

rust/agama-server/src/scripts/web.rs
```diff
@@ -0,0 +1,139 @@
...
+        (status = 200, description = "The scripts were successfully executed.")
+    )
+)]
+async fn run_scripts(
+    state: State<ScriptsState>,
+    Json(group): Json<ScriptsGroup>,
+) -> Result<(), ScriptServiceError> {
+    let scripts = state.scripts.write().await;
+    if let Err(error) = scripts.run(group).await {
+        tracing::error!("Could not run user-defined scripts: {error}");
```
jreidinger, line 136: so it means that it will return 200 when coulld not run scripts? I do not feel it is good behavior.

service/lib/agama/storage/finisher.rb
imobachgs: We could move them to a different place (in AutoYaST they are copied to `/var/adm/autoinstall`.

rust/agama-lib/share/profile.schema.json
imobachgs: Before the post-partitioning. I can improve the description.

service/test/agama/storage/manager_test.rb
imobachgs: I prefer this (that's why I use a temporary directory) instead of mocking all the calls, to be honest.

rust/agama-lib/src/scripts/client.rs
imobachgs: Yes, like the rest of the calls :-) Done.

rust/agama-lib/src/scripts/model.rs
imobachgs: I discussed about it with the network team. It will allow us to remove a lot of boilerplate code in the networking part.

rust/agama-lib/src/scripts/model.rs
imobachgs: Yes, good point. It should be doable by playing with enums and "flatten".

rust/agama-lib/src/scripts/store.rs
imobachgs: Just a leftover.

rust/agama-lib/src/scripts/store.rs
imobachgs: Well, it is not alien in Rust, for instance you have `to_vec`. Sometimes I am not sure whether we are abusing those traits.

rust/agama-server/src/scripts/web.rs
imobachgs: I think it is OK if one of the scripts fails. You run, log the error, and continue with the next one.

rust/agama-lib/src/scripts/store.rs
imobachgs: Well, that's a good question. When would you run the pre-script? In the future we should be able to run the script when loading the profile (to modify it in-place). But we can postpone it until you click the "install" button. I guess we will need to do more adjustments in this area.

jreidinger: I would do it one level up, when you load profile, then run any pre script it will find....but e.g. not before you validate that profile.

rust/agama-lib/src/scripts/store.rs
imobachgs: Yes, you are right. It behaves in this way for consistency with other clients. I already told to @mchf on IRC that we should change this behavior for all the clients but as a follow-up PR.

rust/agama-lib/src/scripts/store.rs
imobachgs: No, no, question was not where in the code, but when. Given that the pre scripts does not support changing the profile yet, I am not sure there is a use case for them right now.

Perhaps you want to enable something, but then pre scripts might need to run the probing, but probing happens only once in Agama by now (and for single product scenarios it happens too early).

rust/agama-lib/share/profile.schema.json
imobachgs: They were expected to be used to modify the profile, but I am not sure whether we should implement that behavior. Or perhaps to activate some hardware.

service/test/agama/storage/manager_test.rb
lslezak: Fine for me, it's just a matter of preferences.

rust/agama-lib/src/scripts/store.rs
imobachgs: I used the `to_*` approach because you cannot create a `Script` from a `ScriptConfig` (you need the `ScriptGroup`). So I kept the same approach for the `to_script_config`. But implementing `From<&Script> for ScriptConfig`.

rust/agama-lib/src/scripts/store.rs
imobachgs: Actually, it is not even needed 😅. We need to check the other stores too.

imobachgs: Actually, in this case it is not even needed at all (perhaps when we extend the tests suite).

rust/agama-server/src/scripts/web.rs
lslezak: Um, I'm not sure. Maybe some script could depend on the result of the previous one.

Maybe it should be configurable whether to continue on error or not. :thinking:

service/test/agama/storage/manager_test.rb
imobachgs: I have changed the code to store those assets under `/var/log/agama-installation`.

jreidinger (approved):

PR #1674, opened 2024-10-17
Create slowroll.yaml

WesfunOfficial:
Slowroll.yaml is almost indentical to leap_160.yaml and tumbleweed.yaml but will eventually give users the option to install slowroll through agama. was unable to provide translations since i only speak english and i am unsure if there other arch's for slowroll besides x86_64

Slowroll is currently the only option not available to install via agama that (i believe) is not planning on using something different (like Aeon, Kalpa and Leap Micro)

### DIFF
--- a/products.d/slowroll.yaml
+++ b/products.d/slowroll.yaml
@@ -7,7 +7,7 @@ description: 'A slightly slower rolling release of openSUSE designed to update
   less often than Tumbleweed but more often than Leap without forcing users to
   choose between "stable" and newer packages'
 icon: Slowroll.svg
-# Again, currently no translations at this time.
+# Do not manually change any translations! See README.md for more details.
 translations:
   description:
     ca:
--- a/products.d/slowroll.yaml
+++ b/products.d/slowroll.yaml
@@ -1,7 +1,9 @@
 id: Slowroll
 name: Slowroll
 # ------------------------------------------------------------------------------
-# WARNING: There are currently no translations at this time!
+# WARNING: When changing the product description delete the translations located
+# at the at translations/description key below to avoid using obsolete
+# translations!!
 # ------------------------------------------------------------------------------
 description: 'A slightly slower rolling release of openSUSE designed to update
   less often than Tumbleweed but more often than Leap without forcing users to
@@ -9,29 +11,15 @@ description: 'A slightly slower rolling release of openSUSE designed to update
 icon: Slowroll.svg
 # Do not manually change any translations! See README.md for more details.
 translations:
-  description:
-    ca:
-    cs:
-    de:
-    es:
-    fr:
-    id:
-    ja:
-    nb_NO:
-    pt_BR:
-    ru:
-    sv:
-    tr:
-    zh_Hans:
 software:
-  installation_repositories: # was only able to find x86_64 at the time of writing
+  installation_repositories:
     - url: https://download.opensuse.org/slowroll/repo/oss/
       archs: x86_64
     - url: https://download.opensuse.org/slowroll/repo/non-oss/
       archs: x86_64
 
   mandatory_patterns:
-    - enhanced_base # seems to be a consistant pattern between all openSUSE versions
+    - enhanced_base
   optional_patterns: null
   user_patterns:
     - basic-desktop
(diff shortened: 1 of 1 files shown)


products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
+id: Slowroll
+name: Slowroll
+# ------------------------------------------------------------------------------
+# WARNING: There are currently no translations at this time!
```
imobachgs, line 4: Do not worry about that, translations are automatically added by Weblate.

products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
...
+# ------------------------------------------------------------------------------
+description: 'A slightly slower rolling release of openSUSE designed to update
+  less often than Tumbleweed but more often than Leap without forcing users to
+  choose between "stable" and newer packages'
+icon: Slowroll.svg
```
imobachgs, line 9: You might need to add the icon to `web/src/assets/products`. In the future, the icon will be placed next to this definition, but we are not there yet.

products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
...
+# Again, currently no translations at this time.
```
imobachgs, line 10:
```suggestion
# Do not manually change any translations! See README.md for more details.
```

Translations are automatically added through weblate, so it is unnecessary to do it by hand.

products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
...
+translations:
+  description:
```
imobachgs, line 12: Please, remove this `description:` key (and its content, of course).

products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
...
+    zh_Hans:
+software:
+  installation_repositories: # was only able to find x86_64 at the time of writing
+    - url: https://download.opensuse.org/slowroll/repo/oss/
+      archs: x86_64
+    - url: https://download.opensuse.org/slowroll/repo/non-oss/
+      archs: x86_64
+
+  mandatory_patterns:
+    - enhanced_base # seems to be a consistant pattern between all openSUSE versions
```
imobachgs, line 34: I would remove the comment.

products.d/slowroll.yaml
```diff
@@ -0,0 +1,143 @@
...
+        subvolumes:
+          - path: home
+          - path: opt
+          - path: root
+          - path: srv
+          - path: usr/local
+          # Unified var subvolume - https://lists.opensuse.org/opensuse-packaging/2017-11/msg00017.html
+          - path: var
+            copy_on_write: false
+          # Architecture specific subvolumes (didnt see anything other than x86_64)
```
imobachgs, line 85: As you took inspiration from the Leap and Tumbleweed ones, I would keep the same list of architecture-specific subvolumes to keep them as similar as possible. So, in the future, we won't need to update the list if Slowroll runs on a different architecture.

imobachgs (changes requested):
Hi @WesfunOfficial,

Thanks a lot for your contribution. I have a couple of questions, but in general you are heading in the right direction :-)

Please, add an entry in the [changes file](https://github.com/agama-project/agama/blob/master/products.d/agama-products.changes) too (referencing either this PR as `gh#agama-project/agama#1674`).

Regards,
Imo

products.d/slowroll.yaml
WesfunOfficial: i tried to do that but i will double check

WesfunOfficial: sorry i see what you mean and so ill add them :)

products.d/slowroll.yaml
WesfunOfficial: will do :)

products.d/slowroll.yaml
WesfunOfficial: gotcha :)

products.d/slowroll.yaml
WesfunOfficial: @imobachgs looks like I'm not allowed to add images only code, but it would be the file from https://en.opensuse.org/File:Slowroll_logo_by_pprmint_text.svg since it was uploaded by @bmwiedemann and appears to be official

bmwiedemann: Maybe the color of the logo will change later to create some distinction from the other openSUSE distributions. We can update images when that happens.

imobachgs (approved):
Thanks you, @WesfunOfficial! I tried the definition and it works.

JFYI, you can add all the changes in a single pull request (instead of opening #1682 and #1683). Anyway, I will review them and add the logo on a separate change.

PR #1675, opened 2024-10-17
fix(rust): fix BaseHTTPClient doc tests

imobachgs:
## Problem

Agama is failing to build in OBS because the tests are failing.

## Solution

After the latest changes in the `BaseHTTPClient` API, the doctests were not adapted. This PR fixes
the problem.


teclator (approved):
LGTM

PR #1677, opened 2024-10-17
Separate common files from service rubygem

mvidner:
## Problem

split out files independent of Ruby version from rubygem-agama-yast

- https://trello.com/c/yAFVjbv6/3824-3-improve-rubygem-agama-yast-packaging

## Solution

Make `rubygem-agama-yast` in OBS a *multibuild* package where `agama-yast.gem` is a source for two spec files:
- `rubygem-agama-yast.spec` via `gem2rpm.yml`, containing the ruby+gem-versioned files, as usual with Rubygems
- `agama-yast.spec` containing the common files for D-Bus and translations

https://build.opensuse.org/package/show/home:mvidner:branches:systemsmanagement:Agama:Devel/rubygem-agama-yast

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

N/A

### DIFF
--- a/service/package/gem2rpm.yml
+++ b/service/package/gem2rpm.yml
@@ -9,16 +9,6 @@
   # "msgfmt" tool
   BuildRequires:  gettext-runtime
   Requires:       dbus-1-common
-:post_install: |-
-  install -D -m 0644 %{buildroot}%{gem_base}/gems/%{mod_full_name}/share/dbus.conf %{buildroot}%{_datadir}/dbus-1/agama.conf
-  install --directory %{buildroot}%{_datadir}/dbus-1/agama-services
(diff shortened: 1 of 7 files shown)


service/package/gem2rpm.yml
```diff
@@ -9,16 +9,6 @@
   # "msgfmt" tool
   BuildRequires:  gettext-runtime
   Requires:       dbus-1-common
```
jreidinger, line 11: is it still required here for gem or only needed for new package with files?

mvidner: Good catch! Also, I will need to make the new package require the older one

jreidinger (approved):

Issue #1685, opened 2024-10-18
`setup.sh` not working on a fresh installation


dgdavid:
When running ./setup-sh on a fresh openSUSE Tumbleweed installation, I'm facing two problems, 

* `./setup-services.sh: line 119: CI: unbound variable`, related to change introduced in https://github.com/agama-project/agama/commit/fac8bb6e7afc16feb63a2feb4b27f5dc1a9edfd0. Commenting lines added in linked commit leads to following problem
* `An error occurred while installing ruby-augeas (0.5.0), and Bundler cannot continue.` Which is quite similar error that was solved at https://github.com/agama-project/agama/pull/1565 for CI.

imobachgs:
The `unbound variable` problem is because of the `-u` in `set -eu.` If we want to keep the `-u`, we need to set a default value for `CI`.

dgdavid:
To reproduce the problem, please install a fresh openSUSE Tumbleweed and try to run the ./setup.sh

lslezak:
You need to install rubygem-ruby-augeas from RPM before running the script. I'll update the script to install it automatically...

lslezak:
Ah, the script prefers the RPM version only when running in CI, you need to run it with `CI=true`...

imobachgs:
Then it will try to install the gems to `/usr,` which is not what I want on my system. And I will need to run the command as `root`.

lslezak:
Yes, we probably do not want to do this outside CI... :worried:

imobachgs:
Perhaps we could use distrobox to set up the development system: https://github.com/agama-project/agama/pull/1687.

jreidinger:
problem with ruby augeas should be fixed with https://github.com/agama-project/agama/pull/1705 .

imobachgs:
It should be fixed with #1687.

PR #1687, opened 2024-10-21
fix: adapt setup scripts to run properly on distrobox

imobachgs:
Due to a problem with [ruby-augeas](https://rubygems.org/gems/augeas) and Ruby 3.3, the setup script does not work anymore. In our continuous integration we decided to use the system libraries, which causes bundler to install software into `/usr/lib`. Of course, we do not want that behavior everywhere.

Alternatively, this PR adapts the script to work using distrobox. The setup is rather simple:

```
distrobox create --root --init --image tumbleweed --name agama
distrobox enter --root agama
./setup.sh
```

The script detects that it is running on distrobox and it workarounds the augeas problem like in our continuous integration.

Please, see #1685 for more context.

### DIFF
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -149,7 +149,7 @@ $SUDO $ZYPPER install \
   cd $MYDIR/rust
   cargo build
 
-  ln -st /usr/bin $MYDIR/rust/target/debug/agama{,*server}
+  ln -sft /usr/bin $MYDIR/rust/target/debug/agama{,*server}
 )
 
 # - D-Bus configuration
--- a/setup-services.sh
+++ b/setup-services.sh
@@ -117,14 +117,15 @@ fi
       sed -e '/gemspec/a gem "ruby-dbus", path: "/checkout-ruby-dbus"' -i Gemfile
   fi
 
-  if [ -n "$CI" ]; then
+  if [ -n "${CI:-}" ] || [ -n "${DISTROBOX_ENTER_PATH:-}" ]; then
     # in CI reuse the pre-installed system gems from RPMs
     bundle config set --local disable_shared_gems 0
+    sudo bundle install
   else
     bundle config set --local path 'vendor/bundle'
+    bundle install
   fi
 
-  bundle install
 )
 
 # Rust service, CLI and auto-installation.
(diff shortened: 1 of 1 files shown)


dgdavid:
As said in IRC, it looks good to me. What is more, I was able to run the project in a similar way by using distrobox (thanks for the help). But I guess this distrobox thingie should be document as well since there could be people like me that have no idea about such a tool.

> Please, see https://github.com/agama-project/agama/pull/1687 for more context.

Do you mean https://github.com/agama-project/agama/issues/1685?

imobachgs:
JFYI, I still fix a problem with NetworkManager.

imobachgs:
I have adapted the script to use the host's D-Bus system socket when running on Distrobox.

setup-services.sh
```diff
@@ -116,14 +117,15 @@ fi
       sed -e '/gemspec/a gem "ruby-dbus", path: "/checkout-ruby-dbus"' -i Gemfile
   fi
 
-  if [ -n "$CI" ]; then
+  if [ -n "${CI:-}" ]; then
     # in CI reuse the pre-installed system gems from RPMs
     bundle config set --local disable_shared_gems 0
+    sudo bundle install
```
jreidinger, line 123: it should use `$SUDO` see https://github.com/agama-project/agama/pull/1687/files#diff-327a0d405c7c05ecb8395935997d54a646315e7da5e275748c565467fc3b7f61R17

setup-services.sh
```diff
@@ -147,7 +150,7 @@ $SUDO $ZYPPER install \
   cd $MYDIR/rust
   cargo build
 
-  ln -st /usr/bin $MYDIR/rust/target/debug/agama{,*server}
+  sudo ln -sft /usr/bin $MYDIR/rust/target/debug/agama{,*server}
```
jreidinger, line 153: and same here `https://github.com/agama-project/agama/pull/1687/files#diff-327a0d405c7c05ecb8395935997d54a646315e7da5e275748c565467fc3b7f61R17` to allow to use different way to have root permission

jreidinger (approved):

PR #1691, opened 2024-10-23
Resolve search sections that match several devices

ancorgs:
## Problem

The search functionality described at `auto_storage.md` is pretty powerful. But the current implementation only allows to:

- search by name
- return 0 or 1 devices per search

## Solution

This implements several parts of the missing functionality, like:

- Ability to match more than one device per `search` section
- Limit the number of matches by using `max`.

This pull request does not add the ability to use other conditions or matching order beyond `name`. So it's only possible to search either for a device with a given name, either for all the devices sorted by name. The syntax for the latter would be this (a search with no conditions):

```
{ "search": {} }
```

For readability, the following alternative syntax is also introduced at this pull request:

```
{ "search": "*" }
```

The new default for drive entries (if omitted) becomes:

```
{ "search": { "max": 1 } }
```

If several devices match with a given `search` section, the corresponding configuration containing the `search` is replicated as many times as needed. Eg. in a system with two disks, the following config...

```
{
  "storage": {
    "drives": [
      { "search": "*" }
    ]
  }
}
```

...is expanded into:

```
{
  "storage": {
    "drives": [
      { "search": "*" },
      { "search": "*" }
    ]
  }
}
```

This also makes sure everything keeps working when an `alias` is used together with a `search` that matches several devices (so all the resulting config objects share the same alias).

## Testing

- Adapted existing unit tests and added a new ones regarding the `search` functionality
- Added a new unit test to check that resolving a config into several ones (potentially with a shared alias) works as expected.

### DIFF
--- a/service/lib/agama/storage/config_conversions/from_json_conversions/search.rb
+++ b/service/lib/agama/storage/config_conversions/from_json_conversions/search.rb
@@ -41,27 +41,20 @@ module Agama
           # @see Base#conversions
           # @return [Hash]
           def conversions
+            return convert_string if search_json.is_a?(String)
+
             {
-              name:         convert_name,
-              if_not_found: convert_not_found
+              name:         search_json.dig(:condition, :name),
+              max:          search_json[:max],
+              if_not_found: search_json[:ifNotFound]&.to_sym
             }
           end
 
-          # @return [String, nil]
-          def convert_name
-            return search_json if search_json.is_a?(String)
+          # @return [String]
+          def convert_string
+            return { if_not_found: :skip } if search_json == SEARCH_ANYTHING_STRING
 
-            search_json.dig(:condition, :name)
-          end
-
-          # @return [Symbol, nil]
-          def convert_not_found
-            return if search_json.is_a?(String)
-
-            value = search_json[:ifNotFound]
-            return unless value
-
-            value.to_sym
+            { name: search_json }
           end
         end
       end
--- a/service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
+++ b/service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
@@ -33,13 +33,22 @@ module Agama
             Configs::Search
           end
 
+          # @see Base#convert
+          def convert
+            converted = super
+            return SEARCH_ANYTHING_STRING if converted.is_a?(Hash) && anything?(converted)
+
+            converted
+          end
+
         private
 
           # @see Base#conversions
           def conversions
             {
               condition:  convert_condition,
-              ifNotFound: config.if_not_found.to_s
+              ifNotFound: config.if_not_found.to_s,
+              max:        config.max
             }
           end
 
@@ -49,6 +58,14 @@ module Agama
             return unless name
 
             { name: name }
+          end
+
+          # Whether the search can be aliased with the special "*" string
+          def anything?(search_hash)
+            [
+              { ifNotFound: "skip" },
+              { condition: {}, ifNotFound: "skip" }
+            ].include?(search_hash.compact)
           end
         end
       end
--- a/service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
+++ b/service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
@@ -35,10 +35,9 @@ module Agama
 
           # @see Base#convert
           def convert
-            converted = super
-            return SEARCH_ANYTHING_STRING if converted.is_a?(Hash) && anything?(converted)
+            return SEARCH_ANYTHING_STRING if config.all_if_any?
 
-            converted
+            super
           end
 
         private
@@ -58,14 +57,6 @@ module Agama
             return unless name
 
             { name: name }
-          end
-
-          # Whether the search can be aliased with the special "*" string
-          def anything?(search_hash)
-            [
-              { ifNotFound: "skip" },
-              { condition: {}, ifNotFound: "skip" }
-            ].include?(search_hash.compact)
           end
         end
       end
--- a/service/lib/agama/storage/config_search_solver.rb
+++ b/service/lib/agama/storage/config_search_solver.rb
@@ -36,22 +36,7 @@ module Agama
       # @param config [Agama::Storage::Config]
       def solve(config)
         @sids = []
-        config.drives.each do |drive_config|
-          device = find_drive(drive_config.search)
-          drive_config.search.solve(device)
-
-          add_found(drive_config)
-
-          next unless drive_config.found_device && drive_config.partitions?
-
-          drive_config.partitions.each do |partition_config|
-            next unless partition_config.search
-
-            partition = find_partition(partition_config.search, drive_config.found_device)
-            partition_config.search.solve(partition)
-            add_found(partition_config)
-          end
-        end
+        config.drives = config.drives.flat_map { |d| solve_drive(d) }
       end
 
     private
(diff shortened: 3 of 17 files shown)


service/lib/agama/storage/configs/search.rb
```diff
@@ -37,6 +37,10 @@ class Search
         # @return [:create, :skip, :error]
         attr_accessor :if_not_found
 
+        # Optional max number of devices to match
+        # return [Integer, nil]
+        attr_accessor :max
```
joseivanlopez, line 42: NP: I would explicitly mention that nil means it finds all unassigned devices that match the condition.

joseivanlopez (commented):
Some notes we already talked about:

* Adapt schema to include `max` in the search section.
* Adapt schema to include and document `"*"` const as possible value for search.
* Extend conversion to JSON to include `max` in search.

service/lib/agama/storage/configs/encryption.rb
```diff
@@ -65,6 +65,17 @@ def missing_password?
 
           password.nil? || password.empty?
         end
+
+        # Redefined in order to allow marshaling
+        def marshal_dump
```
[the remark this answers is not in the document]
ancorgs: Yes. I found out the culprit was RSpec mocking. Already solved.

service/lib/agama/storage/configs/search.rb
ancorgs: Done via `push -f`

service/lib/agama/storage/config_search_solver.rb
```diff
@@ -62,24 +47,87 @@ def solve(config)
...
+        drive_config
+      end
+
+      # see #solve_drive
+      #
+      # @note The given partition object can be modified
+      #
+      # @param original_partition [Configs::Partition]
+      # @param drive_device [Y2Storage::Partitionable]
+      # @return [Configs::Partition, Array<Configs::Partitions>]
```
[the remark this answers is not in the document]
ancorgs: Fixed via `push -f`

ancorgs:
@joseivanlopez I updated the PR (using force push, sorry) to address all concerns. Please re-review

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
```diff
@@ -50,6 +59,14 @@ def convert_condition
 
             { name: name }
           end
+
+          # Whether the search can be aliased with the special "*" string
```
joseivanlopez, line 63: NP
```suggestion
          # Whether the search can be aliased with the special "*" string
          # @return [Boolean]
```

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
```diff
@@ -33,13 +33,22 @@ def self.config_type
             Configs::Search
           end
 
+          # @see Base#convert
+          def convert
+            converted = super
+            return SEARCH_ANYTHING_STRING if converted.is_a?(Hash) && anything?(converted)
```
joseivanlopez, line 39: I am not sure yet if we should generate shortcuts. I think this is the only case in which we do it. Let's keep it for now.

joseivanlopez, line 39: Is there any case in which `super` does not return a hash?

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
```diff
@@ -50,6 +59,14 @@ def convert_condition
 
             { name: name }
           end
+
+          # Whether the search can be aliased with the special "*" string
+          def anything?(search_hash)
```
joseivanlopez, line 64: What about adding `Configs::Search#anything?` instead of comparing the hashes?

service/lib/agama/storage/config_conversions/from_json_conversions/search.rb
```diff
@@ -41,27 +41,20 @@ def convert
...
-            search_json.dig(:condition, :name)
-          end
-
-          # @return [Symbol, nil]
-          def convert_not_found
-            return if search_json.is_a?(String)
-
-            value = search_json[:ifNotFound]
-            return unless value
+          # @return [String]
```
joseivanlopez, line 53: NP: I think we can follow the same implementation pattern as the rest of conversions:

```ruby
          # @see Base#conversions
          # @return [Hash]
          def conversions
            {
              name:         convert_name,
              max:          search_json[:max],
              if_not_found: convert_not_found
            }
          end

          # @return [String, nil]
          def convert_name
            return if search_json == SEARCH_ANYTHING_STRING
            return search_json if search_json.is_a?(String)

            search_json.dig(:condition, :name)
          end

          # @return [Symbol, nil]
          def convert_not_found
            return :skip if search_json == SEARCH_ANYTHING_STRING
            return if search_json.is_a?(String)
            
            search_json[:ifNotFound]&.to_sym          
          end
```

joseivanlopez (commented):
Some comments. Feel free to ignore the NPs.

Changelog entry is missing.

service/lib/agama/storage/config_conversions/from_json_conversions/search.rb
ancorgs: I find that much harder to follow.

We typically handle a hash except an special case in which a string can be used as shortcut.

With the implementation in this PR, that special case is handled at its own method.

With the implementation you suggest, the management is spread over several methods, all of them with its one exception for the string case. And the more attributes we add, the more methods we have with a dedicated `if` at the beginning. I don't see the point.

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
ancorgs: Not really. Good point.

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
ancorgs: Done (with another method name)

service/lib/agama/storage/config_conversions/to_json_conversions/search.rb
joseivanlopez: I would say that adding the device name is kind of expected for a solved config. I mean, independently of the initial  conditions, I would expect the solved config should be a search by name. For example:

```json
{ 
  search: { 
    condition: { size: { bigger: "10 GiB" } }
  }
}
```

Resolves to:

```json
{ 
  search: "/dev/vdc" 
}
```

But of course, we can reconsider it.

joseivanlopez (approved):
LGTM

service/test/y2storage/agama_proposal_search_test.rb
```diff
@@ -0,0 +1,221 @@
...
+        context "if there are no partitions in the disk" do
+          let(:disk_name) { "/dev/vdc" }
+
+          it "processes the proposal" do
+            devicegraph = proposal.propose
+            disk = devicegraph.find_by_name(disk_name)
+            expect(disk.partitions.size).to eq 1
+          end
+
+          it "register a warning about non-existent partitions" do
```
ancorgs, line 210: BTW, I think we should reconsider this behavior. I don't think it's worth a warning. It can generate noise in many situations.

ancorgs: For the record: we agreed it makes no sense and it will be changed in a subsequent pull request.

PR #1693, opened 2024-10-23
fix: do not freeze when downloading the logs

imobachgs:
## Problem

If you try to download the logs using the web UI, `agama-web-server` gets unresponsive.

## Solution

- Use `tokio::process::Command` to download the logs.
- Rename the endpoint to reflect the file format.

## To do

- The UI is not properly adapted yet but @dgdavid is working on it.


dgdavid (approved):
LGTM, thank you!

PR #1690, opened 2024-10-22
refactor(web): improve access to "Install", "Download logs", and "Installer options" actions

dgdavid:
## Problem

The `Install` button is kind of hidden in the `Overview` page, where users have to come back after fixing installation issues (if any) for start the installation. Twice effort for users: to guess where such a button is and to come back to hit it once the installer is happy and ready to proceed.

Additionally, the `Download logs` and `Installer options` actions are mounted in the _application sidebar_, which is not always available, making impossible to access them in screens such as _Product Selection_ or _Installation Finished_.

## Solution

Move all these actions to the top bar, which is always mounted, visible, and accessible. To do so, many changes were needed here and there to achieve visible changes listed below. See the PR commit by commit to know more.

* [x] Move the “Install” button to the top bar
* [x] Drop the “Install” button from the Overview Page
* [x] Drop the cog icon from the top bar and use a [PF/Drowdown](https://www.patternfly.org/components/menus/dropdown) instead
* [x] Move “Installer options" to the top bar dropdown
* [x] Move “Download logs” to the top bar dropdown
* [x] Keep the “Change product” in the sidebar until we find a better place for it, but with better look&feel.

Apart from the above and the needed changes to make it possible, the `About` component has been deleted too since it was kind of useless at this moment.

## Notes

* Since it removed the _Instalaltion.jsx_ component, commit https://github.com/agama-project/agama/pull/1690/commits/f10153bd56e96c1ec6d3c64a1247d52999bd2bad might fix https://github.com/agama-project/agama/issues/1616
* All _.js_ and _.jsx_ touched files were migrated to TypeScript, which made needed to add a _custom.d.ts_ file to avoid TypeScript complaints at the time to import icons from `@material-symbols/svg-400` package. See https://github.com/agama-project/agama/pull/1690/commits/e02c3a2c36c8ab8278e8b35daec986d330bb9087 and https://webpack.js.org/guides/typescript/#importing-other-assets for more details
* All copyright dates of touched files were properly updated.
* As said in the [commit](https://github.com/agama-project/agama/pull/1690/commits/d14ff337e85e821fb6c81ad70700eee621330741), **the warning issues link is kind of temporary until it gets improved too**
## Screenshots

| Screen | Before | After |
|-|-|-|
| Product selection |before_product_selection|download_logs_and_installer_option_from_product_selection |
| Overview with issues |before_overview_with_issues|overview_page_issues_link|
| Overview w/o issues |before_overview_wo_issues | overview_page_install_button |
[...]

PR #1694, opened 2024-10-23
fix(web): make download logs actions work again

dgdavid:
## Problem 

After changes introduced in https://github.com/agama-project/agama/pull/1693, the `LogsButton` component has become obsolete and does not work as expected.

## Solution

Use an old plain HTML link pointing to the expected URL and using the [download](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) attribute.

## Additional note

Please, bear in mind that this is kind of intermediate change to make the link work again. It will be superseded (and probably improved) by the  WIP https://github.com/agama-project/agama/pull/1690

## Testing

- Tested manually


imobachgs (approved):
Much better now :-)

PR #1696, opened 2024-10-24
Storage: Do not register issues when skipping entries at the config

ancorgs:
## Problem

At the storage configuration, it's possible to specify a section describing a device (drive, partition, volume group, logical volume, etc.) with a `search` entry to indicate such a  description corresponds to one or several previously existing device(s) that should be found in the system.

That `search` entry can contain `ifNotFound: "skip"` to indicate the description should be ignored if the corresponding devices are not found.

If that happens, a warning is added to the list of found issues with a message like "_No device found for an optional drive/partition/whatever_".

That warning looks overkill. Especially having into account the default value for `ifNotFound` is "error", which means profiles using "skip" really mean it.

## Solution

Reduce the noise by not longer generating the issue for missing optional devices.

### DIFF
--- a/service/test/y2storage/agama_proposal_search_test.rb
+++ b/service/test/y2storage/agama_proposal_search_test.rb
@@ -207,7 +207,7 @@ describe Y2Storage::AgamaProposal do
             expect(disk.partitions.size).to eq 1
           end
 
-          it "register a no issues about non-existent partitions" do
+          it "does not include any issue about non-existent partitions" do
             proposal.propose
             expect(proposal.issues_list).to be_empty
           end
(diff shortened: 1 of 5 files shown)


service/test/y2storage/agama_proposal_search_test.rb
```diff
@@ -207,12 +207,9 @@
             expect(disk.partitions.size).to eq 1
           end
 
-          it "register a warning about non-existent partitions" do
+          it "register a no issues about non-existent partitions" do
```
joseivanlopez, line 210:
```suggestion
          it "does not include any issue about non-existent partitions" do
```

joseivanlopez (approved):
LGTM, only a suggestions.

PR #1699, opened 2024-10-25
Use AgamaProposal for the initial storage proposal

ancorgs:
We decided to stop relying on the YaST proposal and always use the new so-called `AgamaProposal` both for the auto-installation and for the interactive one.

This makes that change effective by generating an initial storage configuration that fulfills the product settings and executing `AgamaProposal` for it during the storage probing phase.

## Bonus

This includes a merge of `master` into the feature branch. To keep things in sync and, even more important, to get some fixes from there.


joseivanlopez (changes requested):

joseivanlopez (approved):

PR #1705, opened 2024-10-28
Update ruby-augeas to fix problem at CI

ancorgs:
## Problem

The checks at GitHub were failing for new pull requests because it was not possible to install ruby-augeas 0.5.0 at the containers.

Apparently that was triggered by a recent update of the version of the package at openSUSE Tumbleweed (it moved to 0.6.0).

## Solution

This updates `Gemfile.lock` to use the same version that is now available on Tumbleweed. Apparently that fixes CI for the new pull requests.


joseivanlopez (approved):
LGTM

PR #1712, opened 2024-10-29
Temporary fixes to UI

ancorgs:
Now that we don't have a Proposal object exported at the API, this should prevent the UI from crashing right away.

Temporarily the storage page looks like this:

storage

### DIFF
--- a/rust/agama-lib/src/storage/client.rs
+++ b/rust/agama-lib/src/storage/client.rs
@@ -25,7 +25,7 @@ use super::model::{
     Md, Multipath, Partition, PartitionTable, ProposalSettings, ProposalSettingsPatch, Raid,
     Volume,
 };
-use super::proxies::{ProposalCalculatorProxy, ProposalProxy, Storage1Proxy};
+use super::proxies::{DevicesProxy, ProposalCalculatorProxy, ProposalProxy, Storage1Proxy};
 use super::StorageSettings;
 use crate::dbus::get_property;
 use crate::error::ServiceError;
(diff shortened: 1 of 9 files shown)


ancorgs:
We agreed to merge this to the feature/playground branch even with some broken tests. Anyways, the branch will be heavily polished before even considering merge to master.

rust/agama-lib/src/storage/client.rs
```diff
@@ -69,6 +70,11 @@ impl<'a> StorageClient<'a> {
                 .cache_properties(zbus::CacheProperties::No)
                 .build()
                 .await?,
+            // Same than above, actions are reexported with every call to recalculate
+            devices_proxy: DevicesProxy::builder(&connection)
+                .cache_properties(zbus::CacheProperties::No)
```
joseivanlopez, line 75: The *Storage1.Devices* interface is not reexported. This should be similar to the *storage_proxy* and *calculator_proxy*. I would say caching it is ok here. Anyway, it is not that important because we are caching at client level.

joseivanlopez: At some point I would like to reevalute the *Devices* interface. In think it would be better to add a new method *Storage1#devices* instead of having a separate interface. That method would report a JSON according to a schema, for example:

```
{
  system: {},
  target: {},
  actions: {}
}
```

joseivanlopez (approved):

rust/agama-lib/src/storage/client.rs
ancorgs: Yes. Definitely all the proxies need to be reconsidered... afterwards.

PR #1718, opened 2024-10-31
feat(web): storage config model

joseivanlopez:
The UI components should be agnostic to the structure and complexity of the storage config.

Add a Config model to deal with the storage config. The model generates objects (e.g., `Drive`, `Partition`, etc) from the config data. This makes easier for the UI components to read and modify the config data.

### DIFF
--- a/web/src/api/storage/types/checks.ts
+++ b/web/src/api/storage/types/checks.ts
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) [2024] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import * as config from "./config";
+
+// Type guards.
+
+export function isFormattedDrive(drive: config.DriveElement): drive is config.FormattedDrive {
+  return "filesystem" in drive;
+}
+
+export function isPartitionedDrive(drive: config.DriveElement): drive is config.PartitionedDrive {
+  return !("filesystem" in drive);
+}
+
+export function isSimpleSearchAll(search: config.SearchElement): search is config.SimpleSearchAll {
+  return search === "*";
+}
+
+export function isSimpleSearchByName(
+  search: config.SearchElement,
+): search is config.SimpleSearchByName {
+  return !isSimpleSearchAll(search) && typeof search === "string";
+}
+
+export function isAdvancedSearch(search: config.SearchElement): search is config.AdvancedSearch {
+  return !isSimpleSearchAll(search) && !isSimpleSearchByName(search);
+}
+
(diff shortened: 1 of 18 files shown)


web/src/api/storage/types/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function isFormattedDrive(drive: config.DriveElement): drive is config.FormattedDrive {
+  return "filesystem" in drive;
+}
+
+export function isPartitionedDrive(drive: config.DriveElement): drive is config.PartitionedDrive {
+  return !("filesystem" in drive);
+}
+
+export function isSearchAll(search: config.Search): search is config.SearchAll {
```
ancorgs, line 35: The name of this function can be confusing. There are many ways to search for all.

But this functions checks whether a particular way (the asterisk which is literally a shortcut for `{"search": { "ifNotFound": "skip" }}`) is used.

I would prefer a name that reflects that. I don't have a good suggestion, though. Maybe `isSearchAllShortcut` or something along that line.

web/src/api/storage/types/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+  return search === "*";
+}
+
+export function isSearchByName(search: config.Search): search is config.SearchByName {
```
ancorgs, line 39: Same than above, there are several ways to search for name. This only checks if the shortcut has been used and that's not obvious from the function name.

web/src/api/storage/types/checks.ts
ancorgs: I know it corresponds to the name of the type `searchByName` but I still find it kind of confusing. Maybe a better name for the type, maybe a better name for this function... I don't know.

web/src/api/storage/types/checks.ts
ancorgs: Again same than above, I know it corresponds to the name of the type. But still I see to much room for confusion.

web/src/api/storage/types/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+  return "delete" in partition;
+}
+
+export function isPartitionToDeleteIfNeeded(
+  partition: config.PartitionElement,
+): partition is config.PartitionToDeleteIfNeeded {
+  return "deleteIfNeeded" in partition;
+}
+
+export function isPartition(partition: config.PartitionElement): partition is config.Partition {
```
ancorgs, line 59: Similar to the previous cases (and again, understanding it's a consequence of the type name) this name is confusing. According to the definition, this is false for `isPartition`:

`{"search": "/dev/vda1", "size": { "min": 0, "max": "current" }, "deleteIfNeeded": true }`

While this is true for `isPartition`:

`{"search": "/dev/vda1", "size": { "min": 0, "max": "current" } }`

Sure it makes sense when you compare the three names all along each other and in the right order (`partitionToDelete`, `partitionToDeleteIfNeeded` and `partition`). But not much when you find the last one in isolation.

web/src/storage/model/config/size.ts
```diff
@@ -0,0 +1,81 @@
...
+    const size: Size = { min: this.bytes(sizeRange.min) };
+    if (sizeRange.max) size.max = this.bytes(sizeRange.max);
+
+    return size;
+  }
+
+  private bytes(value: config.SizeValueWithCurrent): number | undefined {
+    if (checks.isSizeCurrent(value)) return;
+    // TODO: bytes from string.
+    if (checks.isSizeString(value)) return;
```
ancorgs, line 74: Is this omission recorded somewhere? Sounds like something we should implement very soon.

web/src/api/storage/types/checks.ts
joseivanlopez: Yes, it is tricky to find good names for that types. If we change it, I would change the type name. IMHO, changing the type guard name is not a good idea.

web/src/storage/model/config/size.ts
joseivanlopez: Yes, there is a follow-up item in the trello card.

ancorgs (approved):
LGTM

PR #1720, opened 2024-11-01
feat(rust): Redesigned agama logs command to have functionality in the backend and accessible via HTTP API

mchf:
## Problem

Agama's CLI logs command used to be implemented with idea CLI running on same machine as real installation (the backend). As this is not true anymore and we have installer's backend and frontend separated with HTTP API, which makes remote access possible, the logs command had to be updated to support this idea.

[Its trello card](https://trello.com/c/3A7vCwfS/3867-agama-cli-move-collecting-logs-into-backend-design-http-api-and-put-it-to-use-in-the-cli)

## Solution

- the particular code which is in control of  collecting logs was moved into agama-lib
- agama's web server has contains http(s)://<server_ip>/api/logs piece in its public API
- agama's CLI uses the above API for downloading the logs

## Testing

- *Tested manually*


rust/agama-server/src/web/http.rs
```diff
@@ -34,6 +34,19 @@ use pam::Client;
...
+pub struct LogsResponse {
+    /// Logs archive file name
+    logs: String,
+}
+
+#[utoipa::path(get, path = "/logs", responses(
+    (status = 200, description = "Compressed Agama logs", body = LogsResponse)
+))]
+pub async fn logs() -> Json<LogsResponse> {
+    Json(LogsResponse { logs: "/path/to/file".to_string() })
```
imobachgs: There is already [a way to get the logs](https://github.com/agama-project/agama/blob/master/rust/agama-server/src/manager/web.rs#L234).

rust/agama-lib/src/base_http_client.rs
```diff
@@ -226,6 +226,16 @@ impl BaseHTTPClient {
         self.unit_or_error(response?).await
     }
 
+    /// Returns raw reqwest::Response. Use e.g. in case when response content is not
+    /// JSON body but e.g. binary data
+    pub async fn get_binary(&self, path: &str) -> Result<Response, ServiceError> {
```
imobachgs, line 231: Does it only work with "binary" data? What if it is a text?

rust/agama-lib/src/logs/http_client.rs
```diff
@@ -0,0 +1,63 @@
...
+        Ok(Self { client })
+    }
+
+    /// Downloads package of logs from the backend
+    ///
+    /// For now the path is path to a destination file without an extension. Extension
+    /// will be added according to the compression type found in the response
+    ///
+    /// Returns path to logs
+    pub async fn store(&self, path: &Path) -> Result<String, ServiceError> {
```
imobachgs: I would most probably put this logic under the manager client, not having an specific client at all.

rust/agama-cli/src/logs.rs
```diff
@@ -58,17 +74,14 @@ pub async fn run(subcommand: LogsCommands) -> anyhow::Result<()> {
...
-            let options = LogOptions {
-                verbose,
-                destination,
-                ..Default::default()
-            };
 
-            Ok(store(options)?)
+            // TODO: deal with return values
+            let destination = client.store(destination.as_path()).await.unwrap();
+            showln(verbose, format!("{:?}", destination).as_str());
```
imobachgs: I do not think a "verbose" flag, in this case, is that useful in this context. I would always write the path, and that's it.

imobachgs (changes requested):
Thanks! I have a few comments because it looks like there is some duplication.

rust/agama-server/src/web/http.rs
mchf: well, i was not aware of this piece.  I knew, we have some logs downloading in the web UI but remembered only an old "hotfix" implementation

But in my POV the code you reference to needs to be redesigned because it exposes a HTTP API which runs CLI "agama logs store" at the backend. Purpose of this PR was to separate frontend (in this case CLI) from actual generating of logs at the backend (to make possible to connect the CLI  to a remote backend). So, the code you pointed doesn't need to use "agama logs store" anymore, but can use new HTTP API call "http://.../api/logs/store" directly.

rust/agama-lib/src/base_http_client.rs
mchf: binary here means that it doesn't try to "deserialize" the content into a particular format ... e.g. JSON like in case of the (most of) other calls.

Naming is not my strongest skill. May be it should be something like "get_raw_response" or so.

rust/agama-cli/src/logs.rs
mchf: no problem ... I'll remove the verbose flag.

rust/agama-server/src/web/http.rs
mchf: I'm not sure whether you want the change I described above in this PR or separate one.

rust/agama-lib/src/base_http_client.rs
imobachgs: Yes, *raw* sounds good to me.

rust/agama-server/src/web/http.rs
imobachgs: My point is that we have `/api/manager/logs` and `/api/logs`. My expectation was to fix the first one to do not use the command instead of creating another one.

mchf: yeah. As said, I overlooked that /api/manager/logs, just rememebered an old implementation. So, I'll add fix for this one and try to merge  manager's http_client any here added logs' http_client ... should not be a big deal, just moving the code around.

rust/agama-lib/src/base_http_client.rs
mchf: fine, so we have a winner

rust/agama-server/src/manager/web.rs
```diff
@@ -226,36 +228,53 @@ async fn installer_status(
...
 
-    match ServeFile::new(path)
-        .try_call(Request::new(axum::body::Body::empty()))
-        .await
-    {
-        Ok(res) => res.into_response(),
-        Err(_) => (StatusCode::INTERNAL_SERVER_ERROR).into_response(),
-    }
-}
+    match storeLogs(LogOptions::default()) {
```
imobachgs, line 245: Always passing `LogOptions::default` makes me think whether the arguments are useful in this case.

rust/agama-lib/src/logs.rs
```diff
@@ -0,0 +1,336 @@
...
+const DEFAULT_RESULT: &str = "/run/agama/agama-logs";
+// what compression is used by default:
+// (<compression as distinguished by tar>, <an extension for resulting archive>)
+pub const DEFAULT_COMPRESSION: (&str, &str) = ("gzip", "tar.gz");
+const TMP_DIR_PREFIX: &str = "agama-logs.";
+
+/// Configurable parameters of the "agama logs" which can be
+/// set by user when calling a (sub)command
+pub struct LogOptions {
+    paths: Vec<String>,
```
imobachgs, line 70: I wonder if the list of paths and commands is part of the options. IMHO, that should not be the caller's responsibility.

imobachgs (changes requested):
It looks better now, but I still have a few questions/comments.

rust/agama-lib/src/logs.rs
mchf: Well original idea was to prepare a way how to make possible to add a reference to a file as parameter. This doesn't look as promising idea anymore. I'll change that.

rust/agama-server/src/manager/web.rs
```diff
@@ -226,36 +228,53 @@ async fn installer_status(
...
+        Ok(path) => {
+            let file = tokio::fs::File::open(path.clone()).await.unwrap();
```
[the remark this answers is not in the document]
mchf: ah that one escaped my hunt for unwraps when i was polishing the prototype, thanks

web/src/api/manager.ts
```diff
@@ -42,6 +42,6 @@ const finishInstallation = () => post("/api/manager/finish");
 /**
  * Returns the binary content of the YaST logs file.
  */
-const fetchLogs = () => get("/api/manager/logs.tar.gz");
+const fetchLogs = () => get("/api/manager/logs/store");
```
imobachgs, line 45: Have you tried this change? Removing the `.tar.gz` will result in a `.gz` file that will not open when clicked on.

imobachgs (changes requested):
It looks better know, but there is still an issue with the web UI client. Please, check.

web/src/api/manager.ts
mchf: i used wrong content type header ... should be fixed now

rust/agama-server/src/manager/web.rs
```diff
@@ -226,36 +226,57 @@ async fn installer_status(
...
+                let _ = std::fs::remove_file(path.clone());
 
-    Command::new("agama")
-        .args(["logs", "store", "-d", path.as_str()])
-        .status()
-        .await
-        .map_err(|e| ServiceError::CannotGenerateLogs(e.to_string()))?;
+                // See RFC2046, RFC2616 and
+                // https://www.iana.org/assignments/media-types/media-types.xhtml
+                // or /etc/mime.types
```
imobachgs, line 255: I do not now how "standard" is the mime.types (outside of Linux), but we can find it out later.

imobachgs (approved):
It looks almost good. Please, update the `agama-web-ui.changes` file and that's it. Thanks!

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Nov  6 06:06:51 UTC 2024 - Michal Filka <mfilka@suse.com>
+
+- url for downloading agama logs adapted to use new HTTP API
```
imobachgs, line 4:
```suggestion
- URL for downloading Agama logs adapted to use new HTTP API
```

imobachgs (approved):

web/package/agama-web-ui.changes
mchf: i tried but your eye is too sharp even in the morning ;-)

imobachgs (approved):

PR #1727, opened 2024-11-04
fix(web): display the "finished" page

imobachgs:
## Problem

The "Installation Finished" page is not shown when the installation finishes. It seems to be related
to latest changes regarding the application layout. Now, the `App` component is not always on top.

## Solution

Listen for status changes in the installation progress page.

## Testing

- Tested manually

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Nov 04 20:32:29 UTC 2024 - David Diaz <dgonzalez@suse.com>
+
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
Thanks a lot for catching and fixing this @imobachgs

I hope you don't mind I have added a few tests in an attempt to avoid unsubscribing the component again in the future by mistake. I know we should have better tests, but I've meanwhile learn how to really mock a subscription let's at least to have something.

Would be nice to have these "global" subscription at the router level, but we are not there yet because, if I remember correctly, it is not possible yet. I will try to write a bit more about this soon, when I find time for refreshing a bit my mind about both, the react-router-dom and the react-query packages.

PR #1728, opened 2024-11-05
fix(web): proper handle of the network prefix

imobachgs:
## Problem

The web UI crashes when you try to edit a network connection containing an additional IP address
because the prefix is not handled correctly. The `as string` construction tells TypeScript to
consider the variable as a string, but it does not do a runtime conversion.

## Solution

Cast the variable to a string at runtime.


ancorgs (approved):
LGTM

PR #1729, opened 2024-11-05
live: config.sh: ignore non-existent translations if already missing

g7:
## Problem

Some packages (e.g. gettext-runtime) might not be installed, and if missing the glob would not be evaluated and rm would fail.

## Solution

Let's use `rm -f` instead.

## Testing

- *Tested manually*


lslezak (approved):

lslezak:
The list of installed packages might change because of changed dependencies. This is more robust, thank you!

PR #1730, opened 2024-11-05

ancorgs:
Completely temporal, just in case it's useful for someone.

cabrear-david-feo

There must be better options than a raw ExpandableSection.

cabrear-david-feo-ext

### DIFF
--- a/rust/agama-lib/share/storage.schema.json
+++ b/rust/agama-lib/share/storage.schema.json
@@ -10,7 +10,7 @@
     "drives": {
       "description": "Drives (disks, BIOS RAIDs and multipath devices).",
       "type": "array",
-      "items": { "$ref": "#/$defs/drive" }
+      "items": { "$ref": "#/$defs/driveElement" }
     },
     "volumeGroups": {
       "description": "LVM volume groups.",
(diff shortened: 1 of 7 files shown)


dgdavid:
We have been trying to shape this new page for a couple of days now. Having other things to do in the middle, of course. Despite all the things I have said "offline", I still having doubts on how to proceed. I have slightly written the rationale at https://trello.com/c/nOUYvYIY/3873-8-first-prototype-of-the-new-ui#comment-672c738f86a51a5abbd1c842 (private link). Summarizing it a bit for a broader audience: it's so challenging to come up with a proposal that put a11y on the front mainly because of time. Somehow I feel we're in a hurry because a lot of things has changed in the backend that now needs a front-end for the users interact with.

That said, a proposal I had in mind still not convincing me, although I like it more than the @ancorgs versions above. It basically needs a lot of more tests and work to ensure it worth it in terms of a11y, reason while I fear we will go an step backward and work in the mode we were working: a11y later. I guess it is a matter of gathering more and more knowledge, working on it.

Screen Shot 2024-11-07 at 07 49 24
Screen Shot 2024-11-07 at 08 09 28

Screen Shot 2024-11-07 at 08 09 25
Screen Shot 2024-11-07 at 08 12 16


If the menu shown in the screenshots above were accessible, even the "Change device" button could be moved to the "More actions", renaming the latest to something more general.

But most probably we will go for the @ancorgs proposal for the initial version.

PR #1734, opened 2024-11-07
Add Leap,Leap-PXE flavors

lkocman:
## Problem

*openSUSE section is not really applicable to Leap*
Leap is rather a mix of SLE and openSUSE profiles (e.g. ruby version), with a custom branding.
As of now I have to do a bunch manual edits on our agama-insatller-Leap every time we want to rebase . With this we might just need to edit multibuild flavors, and override version.

https://build.opensuse.org/package/show/home:lkocman:branches:systemsmanagement:Agama:Devel/agama-installer

### DIFF
--- a/live/src/_multibuild
+++ b/live/src/_multibuild
@@ -1,6 +1,4 @@
 <multibuild>
-  <flavor>Leap</flavor>
-  <flavor>Leap-PXE</flavor>
   <flavor>openSUSE</flavor>
   <flavor>openSUSE-PXE</flavor>
 </multibuild>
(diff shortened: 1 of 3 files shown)


live/src/_multibuild
```diff
@@ -1,4 +1,6 @@
 <multibuild>
+  <flavor>Leap</flavor>
+  <flavor>Leap-PXE</flavor>
```
lkocman, line 3: suggests
```suggestion
```

imobachgs (approved):

Issue #1700, opened 2024-10-25
Incomplete OpenAPI data


lslezak:
## Use case

I created an online Agama demo at https://agama-online-demo.surge.sh/. I collected the testing data by recording the HTTP responses while clicking around in the browser. But that's inconvenient and fragile (you need to go through all Agama settings, if you miss something then the data is incomplete).

I'd like to automate that, the idea is to get the list of endpoints from the OpenAPI data and query all paths by a script.

But when comparing the current OpenAPI data and the data I got from the browser I found some inconsistencies which block this approach.

*Note: In this use case I'm interested only in the GET requests, the same problem or more problems might be with the POST/PUT/... requests.*

## Missing paths

These paths are missing in the OpenAPI:

- /api/manager/logs
- /api/network/connections/:id
- /api/network/state
- /api/network/wifi
- /api/software/issues/product
- /api/software/issues/software
- /api/storage/dasd/supported
- /api/storage/issues
- /api/storage/zfcp/supported
- /api/users/issues

## Missing /api prefix

Some OpenAPI paths miss the `/api` prefix:

- /l10n/locales
- /questions
- /users/first
- /users/root

imobachgs:
Unfortunately, the OpenAPI has never been working at all. We recently did a ton of fixes (see #1564), but it is still a work in progress.

mvidner:
The source for the demo is https://github.com/agama-project/agama/pull/1662

imobachgs:
Most of the problems are already fixed by #1564. The `issues` endpoints are still missing.

PR #1737, opened 2024-11-08
fix(rust): add missing paths to the OpenAPI spec

imobachgs:
The OpenAPI spec is not complete, as reported in #1700. This PR aims to add the missing pieces to the spec. As part of the work, I updated the utoipa dependency to version 5.2 which is much better when it comes to detect omissions and problems.

## Tasks

- [x] Add missing `/api` prefix.
- [x] Add missing paths.
- [x] Add missing components.
- [x] Migrate to utoipa 5.
- [ ] Add the `issues` spec (postponed).


lslezak (approved):
I'm not a Rust or OpenAPI expert, but the change looks reasonable to me. :smile:

rust/agama-lib/src/openapi.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn ip_inet_ref() -> schema::Ref {
+        schema::Ref::from_schema_name("IpInet")
+    }
+
+    pub fn ip_inet_array() -> schema::Array {
+        schema::Array::new(ip_inet_ref())
+    }
+
+    pub fn mac_addr6() -> Object {
+        ObjectBuilder::new().description(Some("macaddr")).build()
```
jreidinger, line 63: well, here I kind of expect better explanation for it...especially that it is mac address in ip V6 which is known format.

rust/agama-server/src/manager/web.rs
```diff
@@ -27,7 +27,7 @@
 
 use agama_lib::{
     error::ServiceError,
-    logs::{list as list_logs, store as store_logs, LogsLists, DEFAULT_COMPRESSION},
```
jreidinger, line 30: :+1:

rust/agama-lib/src/openapi.rs
imobachgs: Yes, you are right. And using those functions in the network OpenAPI part is missing too.

rust/agama-server/src/network/model.rs
```diff
@@ -737,6 +738,7 @@ pub struct InvalidMacAddress(String);
 
 #[derive(Debug, Default, Clone, PartialEq, Serialize, utoipa::ToSchema)]
 pub enum MacAddress {
+    #[schema(value_type = String, format = "MAC address in EUI-48 format")]
```
jreidinger, line 741: now I am confused. it should have descriptuon from that MacAddr6 schema and also its value type should be read from it, not?

imobachgs: It should, but it does not work in that way. To make it work, I would need to convert the variant into a struct-based one:

```
pub enum MacAddress {
  MacAddress {
    address: macaddr::MacAddr6
  },
  Preserve,
  // etc.
}
```

Otherwise, the macro did not work (at least for me).

mvidner:
Thanks! But :)
I want to test its output, but `agama-web-server docs` no longer works, and this feature is poorly discoverable: no "openapi" mentioned in any `*.md` file (and there are too many hits in the remaining files)
Please add a mention to  `rust/xtask/README.md`  I guess

mvidner:
Yay, now I can see the paths that are wrongly missing the `/api` prefix:
```sh
cd rust
cargo xtask openapi
jq '.paths | keys' out/openapi/*json | grep -v /api | grep \"
```

gives me 
```
  "/manager/logs/list",
  "/manager/logs/store"
  "/ping"
  "/network/connections/:id"
```

jreidinger (approved):

imobachgs:
> Yay, now I can see the paths that are wrongly missing the `/api` prefix:
> 
> ```shell
> cd rust
> cargo xtask openapi
> jq '.paths | keys' out/openapi/*json | grep -v /api | grep \"
> ```
> 
> gives me
> 
> ```
>   "/manager/logs/list",
>   "/manager/logs/store"
>   "/ping"
>   "/network/connections/:id"
> ```

Good catch! I have fixed them.

jreidinger (approved):

PR #1741, opened 2024-11-08
Shorten Leap 16.0 installation_labels

lkocman:
* Labels should not exceed 16 characters
 
* Neither yast2 partitioner, nor gnome disks
  allowed me to use openSUSE-Leap-DVD-x86_64

* Use Leap-DVD-$arch to meet the length requirement

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Nov  8 13:20:10 UTC 2024 - Lubos Kocman <lubos.kocman@suse.com>
+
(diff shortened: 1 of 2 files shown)


lslezak:
> Labels are not mandatory for a linux volume. Each can be a maximum of 16 characters long.

https://tldp.org/HOWTO/Partition/labels.html

Interestingly for ISO images the volume ID (used as the label) can be much longer:

```console
> isoinfo -d -j UTF-8 -i *.iso | grep "Volume id:"
Volume id: openSUSE-Leap-15.6-DVD-x86_64710
```

The "710" is the build number, it is appended without any separator so it looks strange.

PR #1742, opened 2024-11-08

ancorgs:
Absolutely temporal. To be used as starting point to code the real solution.

Replaces #1730

datos1

datos2

### DIFF
--- a/rust/agama-lib/share/storage.schema.json
+++ b/rust/agama-lib/share/storage.schema.json
@@ -10,7 +10,7 @@
     "drives": {
       "description": "Drives (disks, BIOS RAIDs and multipath devices).",
       "type": "array",
-      "items": { "$ref": "#/$defs/drive" }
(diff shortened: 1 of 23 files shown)


ancorgs:
Just to clarify. This is not a proposal about the information that should be displayed and how. It's just an attempt to throw all possible information in the UI so @dgdavid has real data to do his magic.

In fact, I would like each device entry to be way more compact. Let's play with Gimp to show what I mean!

datos-compactos

ancorgs:
Notes to myself about the compact design:
- Header is different for drives, volume_groups, md_raids, etc.
- For all entries: option to set alias.
- For drives: option "do not use" for those in which makes sense.
- For drives: info about the current device (model, current content, etc.) displayed on demand.

ancorgs:
Replaced by #1747 

I tried to re-target this PR but seems it's impossible if the original base branch has been deleted.

Issue #1724, opened 2024-11-04
Translations broken in the 3rd November 2024 iso


bittin:
Hey!

testing with https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-10.0.0-openSUSE-Build25.171.iso today as i promised to do some more testing today

if you change to another language it does not get applied and you still get English only

dgdavid:
Are you able to collect the logs? Asking because... it works for me :roll_eyes:

dgdavid:
> it works for me 🙄

At least changing from English to Español.

bittin:
> Are you able to collect the logs? Asking because... it works for me 🙄

here comes the logs: https://webb.marisa.nu/agama-logs.tar.gz

imobachgs:
I can reproduce the problem in the product selection page. If you reload, the UI will be already translated, but it should happen automatically.

bittin:
alright, this is still a problem in the 5th November iso: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-10.0.0-openSUSE-Build26.4.iso

imobachgs:
> alright, this is still a problem in the 5th November iso: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-10.0.0-openSUSE-Build26.4.iso

Yes, we can reproduce the problem, but we have yet to work on it.

bittin:
> > alright, this is still a problem in the 5th November iso: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-10.0.0-openSUSE-Build26.4.iso
> 
> Yes, we can reproduce the problem, but we have yet to work on it.

Alright i see, will test again on Friday or next week when i have time 

sick with migrane, fever and cold at the moment

imobachgs:
Take care! We will update the bug report when the problem is fixed, and we do not expect it to happen this week.

bittin:
> Take care! We will update the bug report when the problem is fixed, and we do not expect it to happen this week.

Alright then i know

bittin:
Still not fixed with the 12th November iso: https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/iso/agama-installer.x86_64-10.0.0-openSUSE-Build28.5.iso

[...]

Issue #860, opened 2023-11-13
Switching display language via the URL parameter to `es_AR` does not fallback to `es_ES`, it uses English instead


lslezak:
## Problem

- The UI language can be switched by the URL parameter `?lang=es_ES` to Spanish, that works fine
- But when the requested language is not supported it falls back to English, not to a similar language
- Example: `?lang=es_AR` (Spanish in Argentina) is not directly supported, but it should fallback to `es_ES` as it is probably very close, definitely much closer than English

dgdavid:
@nagu165, 

Just out of curiosity. Did you have a look to this? Your answer at https://github.com/openSUSE/agama/issues/932#issuecomment-1912421668 made me though you will try it.

nagu165:
> @nagu165,
> 
> Just out of curiosity. Did you have a look to this? Your answer at [#932 (comment)](https://github.com/openSUSE/agama/issues/932#issuecomment-1912421668) made me though you will try it.

yes sir, I did.
I also tried to solve the issue but couldn't solve it properly. I had exams going on so kept it on hold. I will try this again.
Thank you.

imobachgs:
> Hello sir, I am really sorry for no response. I have tried to solve the issue but i couldn't. so i think i need to learn a lot. I'll start contributing again once i feel confident. Thank you so much for helping me a lot.

That's no problem. Feel free to try it again in the future. Thanks for your interest in the project. Good luck!

imobachgs:
I will fix this while working on [bsc#1233160](https://bugzilla.suse.com/show_bug.cgi?id=1233160).

imobachgs:
Fixed with #1746.

PR #1746, opened 2024-11-11
fix: properly handle installer language changes

imobachgs:
## Problem

Problems:

- [x] It is not possible to change the language. See [bsc#1233160](https://bugzilla.suse.com/show_bug.cgi?id=1233160).
- [x] Changing to Portuguese does not work (it might be related to the code, `pt-BR`).
- [x] The message while reloading the software translations is in the old language.
- [x] #1724
- [x] #860

## Solution

Fix the D-Bus path of the LocaleProxy to the Manager's `org.opensuse.Agama1.Locale` interface.

## Testing

- Tested manually


dgdavid:
@imobachgs,

According to the PR's description, looks to me you're working on https://trello.com/c/isE3194l/620-agama-translations-not-working-consitently too. Please, have a look.

Thanks for addressing these translation issues.

rust/agama-lib/src/proxies/locale.rs
```diff
@@ -20,7 +20,8 @@
 //! [D-Bus standard interfaces]: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces,
 use zbus::proxy;
 #[proxy(
-    default_service = "org.opensuse.Agama1",
+    default_service = "org.opensuse.Agama.Manager1",
+    default_path = "/org/opensuse/Agama/Manager1",
```
mvidner, line 24: Good catch... except it does not work: :rofl: :cry: 

When I run this code and change the UI langue in the browser, `journalctl` has:

> Nov 12 13:10:23 c75fbede8655 agama-web-server[28278]: Sending message: Msg { type: MethodCall, serial: 157, sender: UniqueName(":1.7"), path: ObjectPath("/org/freedesktop/Locale"), iface: InterfaceName("org.freedesktop.DBus.Properties"), member: MemberName("Set"), body: Structure(Dynamic { fields: [Str, Str, Variant] }), fds: [] }
...
Nov 12 13:10:23 c75fbede8655 agama-web-server[28278]: 5:185m5:185mCould not synchronize settings in the localization D-Bus service: D-Bus service error: org.freedesktop.DBus.Error.UnknownObject: Unknown object '/org/freedesktop/Locale'

Looking up "Could not synchronize settings" in the source, I arrive at a `update_dbus`  fn, which uses a `set_locales`  fn, defined in... https://github.com/agama-project/agama/blob/65939ce64a7bb117ed66f37ebae814b61376f871/rust/agama-lib/src/localization/proxies.rs#L62 

Which is a similar, yet different file from this diff
:shrug:

imobachgs: Well, it works for me (TM) 😉 Actually, I would not expect the `set_locales` method to be involved because that's responsible for changing the locales to install, not the installer's language. Are you changing the language through the "Installer Options" menu? Did you restart the `agama-dbus-server` process too (restarting the `agama.service` should be enough)?

Having two interfaces with the same name (`org.opensuse.Agama1.Locale`) is unfortunate, I know, but they are different things:

* [localization/proxies](https://github.com/agama-project/agama/blob/65939ce64a7bb117ed66f37ebae814b61376f871/rust/agama-lib/src/localization/proxies.rs#L62) refers to the Locale API (implemented in `agama-dbus-server`). We kept a D-Bus interface because it was needed for internal communication. But the idea is to get rid of it and rely only on HTTP.
* [proxies/locale](https://github.com/agama-project/agama/blob/65939ce64a7bb117ed66f37ebae814b61376f871/rust/agama-lib/src/proxies/locale.rs) refers to an interface that is implemented by all Ruby-based services which allows to change the language.

I wonder if we should add `default_path = "/org/opensuse/Agama1/Locale",` to the 1st proxy, but in my testing system it seems to work (unless I am missing something).

In any case, I will build an image that includes all the changes so we can perform more tests.

imobachgs: I have fixed both proxies and now I see no error in the logs.

mvidner: > Having two interfaces with the same name (org.opensuse.Agama1.Locale) is unfortunate, I know, but they are different things:

It's not unfortunate, it is **wrong**  :smile: but it happens to work now. I hope I can add a simple fix, otherwise postpone it for another PR

imobachgs: Yes, it is wrong 😅 The point is that we plan to drop one of them, so it won't be duplicated anymore.

mvidner (approved):
My manual language switching test works now
(after installing missing packages to my container, will fix that in another PR)

I can't really review the JS/TS code... it looks well formed to me  :+1:

web/src/context/installerL10n.tsx
```diff
@@ -105,65 +91,56 @@ function storeAgamaLanguage(language) {
...
 }
 
 /**
  * Converts a RFC 5646 language tag to a locale.
  *
  * It forces the encoding to "UTF-8".
  *
- * @param {string} language
- * @return {string}
+ * @param language as a TFC 5646 language tag (e.g., "en-US")
```
dgdavid, line 124: Do you mean RFC?

web/src/context/installerL10n.tsx
```diff
@@ -210,71 +187,61 @@ function reload(newLanguage) {
...
-      const config = await cancellablePromise(fetchConfig());
-      const currentLanguage = languageFromLocale(config.uiLocale);
-
-      if (currentLanguage !== newLanguage) {
-        // FIXME: fallback to en-US if the language is not supported.
-        await cancellablePromise(updateConfig({ uiLocale: languageToLocale(newLanguage) }));
-        return true;
-      }
+  const syncBackendLanguage = useCallback(async () => {
+    const config = await cancellablePromise(fetchConfig());
```
dgdavid, line 202: Question, do we still needing cancellablePromise?

dgdavid (approved):

web/src/context/installerL10n.tsx
imobachgs: Good point. No, I think we do not need it.

dgdavid (approved):

PR #1747, opened 2024-11-12

ancorgs:
This replaces #1742 (it was impossible to resurrect and/or re-target that PR since its base branch was deleted).

Currently this looks like this:

datos1

datos2

But that's just temporal. This is not a proposal about the information that should be displayed and how. It's just an attempt to throw all possible information in the UI so @dgdavid has real data to do his magic.

In fact, I would like each device entry to be way more compact. Let's play with Gimp to show what I mean!

datos-compactos

### DIFF
--- a/web/src/agama.js
+++ b/web/src/agama.js
@@ -100,6 +100,20 @@ agama.ngettext = function ngettext(str1, strN, n) {
   return n === 1 ? str1 : strN;
 };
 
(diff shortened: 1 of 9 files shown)


ancorgs:
Closed in favor of #1789 https://github.com/agama-project/agama/pull/1767

PR #1748, opened 2024-11-12
feat(storage): add index to schema

joseivanlopez:
Changes:

* Calculate initial proposal from a JSON config generated from the product data.
* Add index to the storage schema: it allows to detect  which unsolved config correspond to each drive and/or partition. Changing a config (e.g., the size of a partition) will be easier thanks to the index too.
* Improve model to add information about auto sizes.
* Improve model to add information about resized partitions.


ancorgs:
I'm just starting, but I admit I didn't expect to see the indexes on the schema. Feels like leaking a internal implementation detail.

joseivanlopez:
After some discussions, we decided to move the config model to the backend. It is easier to implement without the need of adding any artificial properties to the config schema (like the index). Moreover, other clients (e.g., a TUI) could get benefit of the model too.

Closing in favor of https://github.com/agama-project/agama/pull/1763.

PR #1751, opened 2024-11-12
Storage model: parse sizes given as strings

ancorgs:
## Problem

The storage model cannot parse sizes when they are specified as strings like "1024", "1 KiB", etc.

## Solution

This pull request introduces proper parsing based on xbytes (which is already an Agama dependency).


dgdavid (commented):
Looks good, but please run `npx prettier -w`

ancorgs:
> Looks good, but please run `npx prettier -w`

Done. And PR updated via `commit --amend` to reduce noise.

joseivanlopez (approved):

PR #1757, opened 2024-11-13
web: Adapt the storage overview to stop using ProposalResult

ancorgs:
## Problem

As a first step to start using `AgamaProposal` instead of `MinGuidedProposal` at the web UI, we disabled the storage summary at the overview page (commenting out the usage of the component `Overview/StorageSection`.

## Solution

This enables the component back, but now reading from the correct source of information.

## Screenshots

The "storage" entry is back home. See following screenshot.

storage_section

### DIFF
--- a/web/src/components/overview/StorageSection.tsx
+++ b/web/src/components/overview/StorageSection.tsx
@@ -38,7 +38,7 @@ const Content = ({ children }) => (
 /**
  * Text explaining the storage proposal
  *
- * TODO: The current implementation asumes there are only drives and no other kind of devices like
+ * TODO: The current implementation assumes there are only drives and no other kind of devices like
  * LVM volume groups or MD raids. Support for more cases (like LVM installation) will be added as
  * the rest of the interface is also adapted.
  */
--- a/web/src/components/overview/StorageSection.tsx
+++ b/web/src/components/overview/StorageSection.tsx
@@ -25,75 +25,8 @@ import { Text, TextContent, TextVariants } from "@patternfly/react-core";
 import { deviceLabel } from "~/components/storage/utils";
 import { Em } from "~/components/core";
 import { _ } from "~/i18n";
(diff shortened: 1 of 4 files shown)


web/src/components/overview/StorageSection.tsx
```diff
@@ -104,81 +37,71 @@ const Content = ({ children }) => (
...
- *
- * @param {object} props
- * @param {Proposal} props.proposal
  */
 export default function StorageSection() {
-  const availableDevices = useAvailableDevices();
-  const result = useProposalResult();
-
-  if (result === undefined) return;
+  const drives = useConfigDevices().filter((d) => d.name);
```
joseivanlopez, line 42: NP: all drives should have a name.

joseivanlopez: BTW,  at this moment `useConfigDevices` returns drives only, but the model will be extended with LVM, etc. Should we filter drives here to be sure everything is a drive?

ancorgs: Yes. This only adapts the summary to the very same cases that are handled by the storage section. We are not working on LVM support yet. If LVM is used, both this part and the storage configuration page would be equally broken.

We could of course filter, just to be safe. But then the information would be wrong (well, not fully precise) when we start adding LVM. I think I prefer to add a TODO to make it obvious and let it explode so it's not overlooked.

web/src/components/overview/StorageSection.tsx
```diff
@@ -37,9 +37,13 @@ const Content = ({ children }) => (
 
 /**
  * Text explaining the storage proposal
+ *
+ * TODO: The current implementation asumes there are only drives and no other kind of devices like
```
joseivanlopez, line 41: Typo: assumes.

joseivanlopez (approved):

PR #1763, opened 2024-11-14
feat(storage): initial support for a config model

joseivanlopez:
A config model is an abstraction of the storage config to make easier for clients to deal with the storage config. Clients (e.g., the web UI) do not need to understand the complexity of the storage config. Instead of that, clients directly work with the config model for both reading the config and applying changes.

Example of storage config:

~~~
{ 
  "drives": [
    {
      "search": {
        "condition": { "name": "/dev/vda" },
        "max": 1,
        "ifNotFound": "error"
      },
      "partitions": [
        { "search": "*", "delete": true },
        { "generate": "default" }
      ]
    }
  ]
}
~~~

And its equivalence in the config model:

~~~
{
  "drives": [
    {
      "name": "/dev/vda",
      "spacePolicy": "delete",
      "partitions": [
        { "name": "/dev/vda1", "delete": true },
        { "name": "/dev/vda2", "delete": true },
        {
          "mountPath": "/",
          "filesystem": {
            "default": true,
            "type": "btrfs",
            "snapshots": false
          },
          "size": {
            "default": true,
            "min": 2048
          }
        }
      ]
    }
  ]
}
~~~

Note: this PR replaces https://github.com/agama-project/agama/pull/1748 in which the model was implemented in the web side.


ancorgs (approved):
LGTM. I have been playing with the result during Hack Week and it seems to work.

PR #1765, opened 2024-11-14
AgamaProposal: Honor DiskAnalyzer candidate devices when searching drives

ancorgs:
## Problem

When searching drives, the initial implementation of `AgamaProposal` uses all disk (or disk-like devices, to be precise) in the system.

That is ok as a first attempt, but we always knew we might need to refine that a bit.

Now that we are working on the new interface for storage configuration, fully based on `AgamaProposal`, the fact that `AgamaProposal` can choose a disk that is not considered by `DiskAnalyzer` to be an installation candidate feels a bit odd.

## Solution

For the time being (and with minimal code changes), this adapts the `AgamaProposal` algorithm to filter out disks that are skipped by `DiskAnalyzer`. That typically means the disks from which the installer is running, disks that contain installation repositories, etc.

We may need to refine that in the future again. But for now that's the most consistent behavior we can adopt without taking big decisions.

## Testing

Added a new unit test

Manual testing done by @joseivanlopez.

### DIFF
--- a/service/lib/agama/storage/config_search_solver.rb
+++ b/service/lib/agama/storage/config_search_solver.rb
@@ -23,10 +23,12 @@ module Agama
   module Storage
     # Solver for the search configs.
     class ConfigSearchSolver
-      # @param devicegraph [Devicegraph] used to find the corresponding devices that will get
-      #   associated to each search element.
-      def initialize(devicegraph)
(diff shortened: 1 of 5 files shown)


dgdavid (approved):
LGTM, thanks

PR #1771, opened 2024-11-15
Support for encrypted passwords

lslezak:
## Problem

- The Agama autoinstallation and CLI accept the first user and the root passwords only in plain text
- That's insecure, everybody who can access the installation profile knows the root password

## Solution

- Support passing an already encrypted (hashed) password in the profile
- Similar to AutoYaST, an additional `encryptedPassword` boolean flag is used to determine whether the specified password is encrypted (`true` value) or plain text (`false` value or missing in the profile)

## Notes

- The web UI allows specifying only plain text passwords
- Encrypted passwords are long and hard to type and they need to be encrypted externally

## Features

- Adapted schema definition
- Adapted the AutoYaST conversion tool
- When an encrypted password is set from Agama CLI then web UI resets the flag back to plain text (it supports only plain text passwords)

## Testing

- Tested manually (both root user and first user), tested the AutoYaST profile conversion
- Updated unit tests

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Nov 15 16:48:44 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
+
+- Allow using encrypted passord for root and the first user
+  (gh#agama-project/agama#1771)
+
+-------------------------------------------------------------------
 Thu Nov 14 15:34:17 UTC 2024 - Ancor Gonzalez Sosa <ancor@suse.com>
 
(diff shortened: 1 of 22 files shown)


service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Nov 15 16:48:44 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
+
+- Allow using encrypted passord for root and the first user
```
imobachgs, line 4:
```suggestion
- Allow using encrypted password for root and the first user
```

imobachgs (approved):

PR #1772, opened 2024-11-15
fix(openapi): add issues, progress and status routes

### LINKED ISSUE #1700: Incomplete OpenAPI data

imobachgs:
## Problem

The OpenAPI specification does not describe the common HTTP APIs (e.g., 
The specification of the common (shared) HTTP APIs the OpenAPI description.

## Solution

Add the description of *issues*, *service status*  and *progress* HTTP APIs. Unless there are more omissions, it should close #1700.

## Testing

- *Tested manually*


teclator (approved):
LGTM

PR #1778, opened 2024-11-21
feat(web): move installation issues to a drawer and keep Install button always visible

dgdavid:
## Problem

[A month ago](https://github.com/agama-project/agama/pull/1690) , the 'Install' button achieved more visibility after being moved to the sticky header, allowing users to access it from any page/section/screen.

But that was just an intermediate step towards improving the 'Install' action and the discoverability of 'Installation setup problems,' which sadly introduced some drawbacks:

* The Install button is displayed dynamically, which still _forces_ users to figure out where it will appear when it’s not visible from the start.
* When installation is not possible, users have to navigate to the Overview page to see what is making their setup invalid.

## Solution

* Keep the install button always enabled and visible, triggering the installation when the setup is valid or displaying found _issues_ when not, and
* Allow users to check found problems always, not only  in the overview 

## A discarded alternative

At commit https://github.com/agama-project/agama/pull/1778/commits/6ecdcebd68bb210d507c5071af9a9a825c051611, this set of changes was in a bit different shape, where the interface rendered a disabled "Install" button and an enabled "Warning" button to display the setup problems. Additionally, it featured a _tooltip_ to explain to users why the "Install" button was disabled and where to click to view the reasons in detail.

<details>
<summary>Click to show/hide a few screenshots</summary>

---

| Install button disabled | Install button tooltip | 
|-|-|
| Screen Shot 2024-11-22 at 08 06 09| Screen Shot 2024-11-22 at 08 06 42  |

| Preflight checks visible | Install button enabled |
|-|-|
| Screen Shot 2024-11-22 at 08 06 53 | Screen Shot 2024-11-22 at 08 08 31|
</details>

However, this approach created unnecessary complexity and moving pieces, not to mention the challenge of writing a clear, concise, yet complete explanation for the _tooltip_ and keeping it up-to-date, including translations

What's more, there is no reason to avoid keeping the "Install" button **always active but behaving slightly different** depending on the setup status:

* Displaying found problems when the installation setup is invalid
* Displaying the confirmation dialog when the setup is valid

This way, the user doesn't need to think. They can simply click the "Install" button. If it's not possible, they can immediately start fixing their setup. If it is possible, they just have to confirm they want to begin the installation.

## Testing

- Added a new unit test
- Tested manually

## Note for reviewers

Please, do a manual test too.

### Valid installation setup

| Install button | onClick: Confirmation dialog |
|-|-|
| Screen Shot 2024-11-22 at 14 21 52|Screen Shot 2024-11-22 at 14 21 55  |

### Not valid installation setup

| Install button | onClick: Preflight checks |
|-|-|
| Screen Shot 2024-11-22 at 14 21 29 | Screen Shot 2024-11-22 at 14 21 22 |

---

Related to not yet planned Trello card, https://trello.com/c/DmUOQN1Y (internal link)


web/src/components/core/InstallButton.tsx
```diff
@@ -81,26 +80,44 @@ according to the provided installation settings.",
...
   const close = () => setIsOpen(false);
   const onAccept = () => {
     close();
     startInstallation();
   };
 
+  // TRANSLATORS: The install button label
+  const buttonText = _("Install");
+  // TRANSLATORS: Accessible text included with the install button when there are issues
+  const withIssuesAriaLabel = _("Not possible with current setup. Click to know more.");
```
dgdavid, line 104: Reviewer, please not that this is part of the install button. Thus, an screen reader will announce "Install not possible with..."

[Kooha-2024-11-25-08-23-04.webm](https://github.com/user-attachments/assets/fdb7c1b1-a59d-4c08-ace5-bd1b113ab9b5)

web/src/components/core/IssuesDrawer.tsx
```diff
@@ -0,0 +1,104 @@
...
+    users: _("Users"),
+    storage: _("Storage"),
+    software: _("Software"),
+  };
+
+  if (issues.isEmpty || phase === InstallationPhase.Install) return;
+
+  return (
+    <NotificationDrawer ref={ref}>
+      <NotificationDrawerHeader title={_("Preflight checks")} onClose={onClose} />
```
dgdavid, line 58: Reviewer, please note the use of "Preflight checks" to avoid the "Issues" technicality. A first attempt was to use "Installation issues", but these are not actually "Installation issues" since the installation does not even started. It is more an "Installation settings issues", but sounds difficult to explain. Looks like "Preflight checks" could be a more self-explanatory _term_. As usual, better ideas are more than welcome.

web/src/components/core/IssuesDrawer.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <NotificationDrawerBody className="agama-issues-drawer-body">
+        <Stack hasGutter>
+          <p>
+            {_(
+              "Current installation settings does not meet the necessary requirements for installing the selected product. Click on each section name to access the relevant page and adjust the settings as needed.",
```
dgdavid, line 63: Reviewers, feel free to improve this text as much as you want. But please resist the temptation to remove it. UI explanations are an aid, not a problem.

imobachgs (commented):
It looks much better now! I just have a few suggestions about some texts.

web/src/components/layout/Layout.tsx
```diff
@@ -35,18 +36,34 @@ type LayoutProps = React.PropsWithChildren<{
...
+ *
+ * FIXME: move the focus to the notification drawer when it is open
  */
 const Layout = ({
   mountHeader = true,
   mountSidebar = true,
   headerOptions = {},
   children,
 }: LayoutProps) => {
+  const [issuesDrawerVisible, setIssuesDrawerVisible] = useState<boolean>(false);
```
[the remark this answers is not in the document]
dgdavid: Yes, you're right. When using default values we could let TS to do its work. Thanks!

web/src/components/core/IssuesDrawer.tsx
shundhammer: "Preflight checks" means something to me since I used flight sims a lot 1995-2005, but I am not sure if a normal user can make sense of this; much less one with limited English knowledge.

I don't even dare to think about translations and how much they make sense to users.

Please let's not get overly creative with this; keep it simple.

dgdavid: @shundhammer 

What do you propose?

web/src/components/core/IssuesDrawer.tsx
shundhammer: Regardless of the wording, please fix the broken English:

"Current installation settings **don't** meet the necessary requirements..."

("settings": Plural -> s/"does not"/"do not"/  which nobody ever says (or writes outside of legal documents), so "don't"

shundhammer: How about something along the lines of

"You need to make some decisions about some of the settings.
Click on each section name to check and maybe change the setting."

1. Don't make it sound like it's a problem of the installer; it's not.
2. Address the user directly to clarify that it needs user interaction.
3. Keep it simple. Keep not-so-fluent English speakers in mind.
4. Don't fall into the "let's resort to legalese BS" trap:
  First of all, the message needs to be easy to understand, not so much to cover all eventualities.

web/src/components/core/IssuesDrawer.tsx
shundhammer: How about "consistency checks"?

Try to avoid lingo from a completely different area of technology like aviation.

web/src/components/core/IssuesDrawer.tsx
dgdavid: As you said above, I'd like keep some consistency between the user action and the message shown. I mean, 

User clicks on "Install"
The installer shows something that does not install
But the installer also tell the user "Before install, X"

So, user can be sure that this _dialog_ has been for sure shown because as a side effect of clicking on the Install action, not an unexpected message.

web/src/components/core/IssuesDrawer.tsx
dgdavid: Not sure about the word "consistency". Sounds weird to me in this context. We could go for "Installation checks" or "Settings checks", but both are too generic IMHO

web/src/components/core/IssuesDrawer.tsx
shundhammer: "Before install" is broken English. It's either "Before installing" or "Before the installation" (or maybe "Before starting the installation".

dgdavid: > "Before install" is broken English. It's either "Before installing"

I mean "Before installing", of course. I do not want to break English more that it is already :P

imobachgs (approved):

web/src/components/core/IssuesDrawer.tsx
dgdavid: Finally, I'll go for "Before installing, you have to make some decisions. Click on each section to review the settings."

web/src/components/core/IssuesDrawer.tsx
dgdavid: Let's use "Pre-installation checks"

PR #1787, opened 2024-11-27
feat(web): request a root authentication method after selecting a product

jreidinger:
## Problem

To ensure privileges in the installed systems, users must define at least one authentication method for the `root` user. This can easily be accomplished by visiting the "Users" page. However, since it is not fully mandatory right now, users might overlook it and end up with an unmanageable system.

## Solution

~~As soon as the installer detects an issue within the users' scope, it will display a new page with a form to set at least one `root` authentication method. It should help to mitigate the risk of ending up with a system lacking access to the administrative user.~~

~~Although not perfect, as users could later bypass the installer by adding a regular user and then removing all `root` authentication methods to avoid detection of an issue related to the users, this is a good step towards a future where the first regular user could also have administrative privileges (a decision to be made).~~

After discarding the above _workaround_ in commit https://github.com/agama-project/agama/pull/1787/commits/fe28059f2a8a5b7ad81f15aac1432b18f14aecae, Agama web interface makes a root authentication method mandatory by displaying the form for setting at least one of them as soon as it detects there is none. I.e., at least once after selecting a product for the first time.

Users can still change their chosen method or value later, but they always must provide at least one before installation. This approach is more robust and consistent than the previous implementation. It's also better aligned with the purpose of introducing this additional step: to ensure the installed system has an administrative user properly set.

## Notes for reviewers

Reviewers, in addition to testing the feature manually, please keep the following aspects in mind.

* [UX] Text on the screen: suggest as many improvements as you can, but keep in mind the need to provide all relevant context and instructions for users who will come across this page right after selecting a product.
* [UX] File Upload: As you can see, the file upload component has been tweaked slightly to improve the keyboard experience. It will certainly require further internal improvements, but we believe it's a good starting point.
* [DX] Naming: If you have better suggestions for file and route names, please share them. I'm (@dgdavid) not happy with the current ones either.
* **There are two commits for using "encryptedPassword" in every place instead of "encryptedPassword" in some of them and "passwordEncrypted" in others.** See https://github.com/agama-project/agama/pull/1787/commits/00da7b3860a93d5dcc32d1de3019fcd531c82e5e and https://github.com/agama-project/agama/commit/bf1099295deb711edb7830e0c265914726d26813

## Testing

- Tested manually
- Added unit tests

## Screenshots

| Not filled form | Partially filled form |
|-|-|
|Empty root methods form | Root auth methods form with filled password |

---

Related to https://trello.com/c/YdZIRHoo (internal link)


web/src/queries/users.ts
```diff
@@ -83,12 +83,12 @@ const useFirstUserChanges = () => {
...
     return client.onEvent((event) => {
       if (event.type === "FirstUserChanged") {
-        const { fullName, userName, password, passwordEncrypted, autologin, data } = event;
+        const { fullName, userName, password, encryptedPassword, autologin, data } = event;
         queryClient.setQueryData(["users", "firstUser"], {
           fullName,
           userName,
           password,
-          passwordEncrypted,
+          encryptedPassword,
```
dgdavid, line 91: I'm realizing that this might require a change in the backend side too :/

dgdavid: I can simply storage "passwordEncrypted" from the backend as a "encryptedPassword" in the query cache. But would be nice... some consistency with this kind of things.

imobachgs: Not only in the backend, but in the schema too. However, it should be pretty straightforward.

lslezak: The backend already uses `encryptedPassword`, that should be fine.

dgdavid: > The backend already uses `encryptedPassword`, that should be fine.

Thus, previous code was broken, see 

https://github.com/agama-project/agama/blob/0094c4675e75a34ee49c80d9efe9c181a81a7d5b/web/src/queries/users.ts#L84-L94

Before the renaming I did, it was _extracting_ `passwordEncrypted`  from the event.

dgdavid: Anyway, I have sent another commit changing all occurrences I've found. See https://github.com/agama-project/agama/pull/1787/commits/bf1099295deb711edb7830e0c265914726d26813

web/src/App.tsx
```diff
@@ -84,6 +85,16 @@ function App() {
       return <Navigate to={PRODUCT.progress} />;
     }
 
+    if (
+      phase === InstallationPhase.Config &&
+      !isBusy &&
+      any_users_issues.length > 0 &&
+      location.pathname !== USER.rootUser.edit
+    ) {
+      // TODO: replace with new page
```
jreidinger, line 94: is TODO still valid?

web/src/components/core/Page.tsx
```diff
@@ -61,7 +61,7 @@ type SectionProps = {
   /** Elements to be rendered in the section footer */
   actions?: React.ReactNode;
   /** As short as possible yet as much as needed text for describing what the section is about, if needed */
-  description?: string;
+  description?: React.ReactNode;
```
jreidinger, line 64: I do not get why this change is needed. Even common message does not help

> small core/Page internal improvements

web/src/components/users/RootAuthMethodsPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    await setRootUser.mutateAsync(data);
+
+    navigate(PATHS.root);
+  };
+
+  // TRANSLATORS: %s will be replaced by a link with the text "upload".
+  const [sshKeyStartHelperText, sshKeyEndHelperText] = _(
+    "Write, paste, drop, or %s a SSH public key file in the above textarea.",
+  ).split("%s");
```
jreidinger, line 84: will this work also for right to left locales?

web/src/components/core/Page.tsx
dgdavid: Funny! The time I decided to go for short commit messages you catch me! :sweat_smile: 

But you're right, what a useless commit message. This change is needed to make possible passing not only a string but a JSX node. Is you have a look, I needed to pass a paragraph (`<p>`) as a section description in order to apply some styles (font-size, if I remember correctly).

Hope you got it now.

In any case, the core/Page component deserves more love too. But after the storage rewrite. With a bit of luck, we can drop the "value" param, only used in the storage world.

web/src/components/users/RootAuthMethodsPage.tsx
dgdavid: Great question! Honestly, I don't know. But for now, I’m just using the interpolation we've documented at https://agama-project.github.io/docs/devel/i18n#formatting-texts.

That said, I really appreciate you raising concerns about RTL languages. While I did try to approach things with RTL in mind from the start, I’m afraid we're still far of having a fully RTL-compatible interface. There are so many parts involved, and even the CSS we're using is key.

Don't get me wrong. I’m not suggesting we stop thinking about RTL just because we don’t have everything in place yet. However, I'd keep things as they are for now and take notes on how to improve these translations to make them RTL-compatible, and tackle them all at once when we can.

Thanks so much for your input!

web/src/components/core/Page.tsx
jreidinger: ok, now it makes more sense. Maybe also it deserve a note in jsdoc above element.

dgdavid: I'd improve the prop description mentioning that a React node is accepted, althoug looking at the type should be enough for someone familiar with React types, which I think it's not your case yet :)

Thanks!

web/src/App.test.tsx
```diff
@@ -156,14 +160,34 @@ describe("App", () => {
         mockClientStatus.isBusy = false;
       });
 
-      it("renders the application content", async () => {
-        installerRender(<App />, { withL10n: true });
-        await screen.findByText(/Outlet Content/);
+      describe("if there is issues with users", () => {
```
imobachgs, line 163:
```suggestion
      describe("if there are issues with users", () => {
```

web/src/App.tsx
```diff
@@ -44,6 +44,7 @@ function App() {
   const { connected, error } = useInstallerClientStatus();
   const { selectedProduct, products } = useProduct();
   const { language } = useInstallerL10n();
+  const any_users_issues = useIssues("users");
```
imobachgs, line 47: Please, use camelCase.

web/src/components/users/RootAuthMethodsPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+    <Page>
+      <Page.Content>
+        <Center>
+          <Page.Section
+            headerLevel="h2"
+            title={_("Setup root user authentication")}
+            description={
+              <p className={textStyles.fontSizeXl}>
+                {_(
+                  "To continue with the installation settings, define at least one authentication method for the root user. You can still edit them anytime before installation.",
```
imobachgs, line 96:
```suggestion
                  "You must define at least one authentication method for the root user. You can still edit them anytime before the installation.",
```

imobachgs (changes requested):
Just some minor things. Otherwise, LGTM.

web/src/components/users/RootAuthMethodsPage.tsx
mvidner: !olleH
RTL expert here :-)

This `split("%s")` technique is fine with respect to bidirectional languages: there are just letters, no numbers or punctuation. Refer to https://github.com/mvidner/bidi-test for examples of failures related to that.

If done incorrectly (`"Write, paste, drop, or" + link + "a SSH public key file in the above textarea."`) it would be a problem because of _grammar_, not text direction:
In Japanese, the translation will come out like this:
> "In the above area, a SSH key write, paste, drop or %s".

Our implementation will handle that fine.

web/src/components/users/RootAuthMethodsPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    onReadStarted={startUploading}
+                    onReadFinished={stopUploading}
+                    onClearClick={clearKey}
+                  />
+                  <FormHelperText>
+                    <HelperText>
+                      <HelperTextItem variant="indeterminate">
+                        {sshKeyStartHelperText}
+                        <Button variant="link" isInline onClick={uploadFile}>
+                          {_("upload")}
```
mvidner, line 138: Here a TRANSLATORS comment is needed, or they will not know whether to use a noun or verb, in which form
("*sube* el archivo" vs "*carga* de archivo completada")

I'm not sure of the correct syntax for the comment, please adapt and insert:

// TRANSLATORS: this "upload" is a commanding verb, in the %s place of
// "Write, paste, drop, or %s a SSH public key file in the above textarea."

dgdavid: Thank you!

web/src/components/users/RootAuthMethodsPage.tsx
dgdavid: Thanks a lot for the explanation @mvidner!

web/src/components/users/RootAuthMethodsPage.test.tsx
```diff
@@ -84,7 +84,39 @@ describe("RootAuthMethodsPage", () => {
...
+      mockLocationFn.mockReturnValue({ state: { from: "/users" } });
+    });
+
+    it("navigates to it after submittin the form", async () => {
+      const { user } = installerRender(<RootAuthMethodsPage />);
+      const passwordInput = screen.getByLabelText("Password");
+      const acceptButton = screen.getByRole("button", { name: "Accept" });
+      await user.type(passwordInput, "linux");
+      await user.click(acceptButton);
+      // After submitting the data, it must navigates
```
imobachgs, line 102:
```suggestion
      // After submitting the data, it must navigate
```

web/src/components/users/RootAuthMethodsPage.test.tsx
```diff
@@ -84,7 +84,39 @@ describe("RootAuthMethodsPage", () => {
...
+      mockLocationFn.mockReturnValue({ state: {} });
+    });
+
+    it("navigates to root path", async () => {
+      const { user } = installerRender(<RootAuthMethodsPage />);
+      const passwordInput = screen.getByLabelText("Password");
+      const acceptButton = screen.getByRole("button", { name: "Accept" });
+      await user.type(passwordInput, "linux");
+      await user.click(acceptButton);
+      // After submitting the data, it must navigates
```
imobachgs, line 118:
```suggestion
      // After submitting the data, it must navigate
```

imobachgs (dismissed):
LGTM. Is it worth updating the `agama.changes` file?

web/src/components/users/RootAuthMethodsPage.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    await user.clear(sshKeyTextarea);
+    await user.type(passwordInput, "t0ps3cr3t");
+    await user.click(acceptButton);
+    expect(mockRootUserMutation.mutateAsync).toHaveBeenCalledWith({
+      password: "t0ps3cr3t",
+      encryptedPassword: false,
+    });
+
+    // After submitting the data, it must navigates
```
dgdavid, line 87:
```suggestion
    // After submitting the data, it must navigate
```

dgdavid:
> Is it worth updating the `agama.changes` file?

IMHO, no, it doesn't. Changes in the Rust and Ruby part were just a variables renamed for consistency.

imobachgs:
> > Is it worth updating the `agama.changes` file?
> 
> IMHO, no, it doesn't. Changes in the Rust and Ruby part were just a variables renamed for consistency.

Well, the changes in the Rust part include changes in 1) the profile and 2) the HTTP API. It is a minor one, though, but visible from the outside.

imobachgs (approved):

dgdavid:
> > > Is it worth updating the `agama.changes` file?
> > 
> > 
> > IMHO, no, it doesn't. Changes in the Rust and Ruby part were just a variables renamed for consistency.
> 
> Well, the changes in the Rust part include changes in 1) the profile and 2) the HTTP API. It is a minor one, though, but visible from the outside.

Done.

Now, the changes file is there, the conflict with `master` resolved and the PR description updated for properly describing the implementation that was actually implemented.

Have a look when you have time. 

Thanks for your patient with this kind of _improvised_ change.

imobachgs (approved):

web/src/components/users/RootAuthMethodsPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+    <Page>
+      <Page.Content>
+        <Center>
+          <Page.Section
+            headerLevel="h2"
+            title={_("Setup root user authentication")}
+            description={
+              <p className={textStyles.fontSizeXl}>
+                {_(
+                  "You must define at least one authentication method for the root user. You can still edit them anytime before the installation.",
```
mvidner, line 107: > will the new root password screen be localizable?

@bittin yes of course, as you can see here.
If by any chance it does not work then we may have broken things in the recent refactoring, #1777.

bittin: @mvidner ah alright thanks, have not seen anything pop up at: https://l10n.opensuse.org/projects/agama/-/sv/ yet

bittin:
@mvidner @imobachgs translated to Swedish now :) this morning

bittin:
The screen is not fully translated

dgdavid:
> The screen is not fully translated

What is missing? Can you provide an screenshot or so?

bittin:
@dgdavid 

bild

Still a lot of English here

mvidner:
@bittin what build of Agama are you using?

bittin:
@mvidner 35.12

imobachgs:
Please, take into account that translations are merged once a week.

bittin:
> Please, take into account that translations are merged once a week.

ah alright did not know that, thank you for the information

bittin:
> @mvidner 35.12

this is now fixed in 36.3 thanks @imobachgs @dgdavid and @mvidner :)

PR #1788, opened 2024-11-27
feat(rust): add support for auto-installation "init" scripts

imobachgs:
Add support for running custom-defined scripts during the first boot of the installed system. The scripts are executed by a special `agama-scripts` service, which is included in a package that is installed on the target system *only* if an init-script was included in the profile.

Agama will not complain if the package is available yet. We should enable this behavior when we 1) submit the package and 2) improve software error reporting.

### DIFF
--- a/rust/agama-lib/src/scripts/settings.rs
+++ b/rust/agama-lib/src/scripts/settings.rs
@@ -26,14 +26,14 @@ use super::{Script, ScriptSource};
 #[serde(rename_all = "camelCase")]
 pub struct ScriptsConfig {
     /// User-defined pre-installation scripts
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub pre: Vec<ScriptConfig>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pre: Option<Vec<ScriptConfig>>,
     /// User-defined post-installation scripts
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub post: Vec<ScriptConfig>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub post: Option<Vec<ScriptConfig>>,
     /// User-defined init scripts
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub init: Vec<ScriptConfig>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub init: Option<Vec<ScriptConfig>>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/rust/agama-lib/src/scripts/settings.rs
+++ b/rust/agama-lib/src/scripts/settings.rs
@@ -31,6 +31,9 @@ pub struct ScriptsConfig {
     /// User-defined post-installation scripts
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub post: Vec<ScriptConfig>,
+    /// User-defined init scripts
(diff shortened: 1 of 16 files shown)


rust/agama-lib/src/scripts/settings.rs
```diff
@@ -26,11 +26,14 @@ use super::{Script, ScriptSource};
 #[serde(rename_all = "camelCase")]
 pub struct ScriptsConfig {
     /// User-defined pre-installation scripts
-    #[serde(skip_serializing_if = "Vec::is_empty")]
-    pub pre: Vec<ScriptConfig>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
jreidinger, line 29: :+1:

rust/agama-lib/src/store.rs
```diff
@@ -94,10 +94,10 @@ impl Store {
     pub async fn store(&self, settings: &InstallSettings) -> Result<(), ServiceError> {
         if let Some(scripts) = &settings.scripts {
             self.scripts.store(scripts).await?;
-        }
 
-        if settings.scripts.as_ref().is_some_and(|s| !s.pre.is_empty()) {
-            self.run_pre_scripts().await?;
+            if scripts.pre.as_ref().is_some_and(|s| !s.is_empty()) {
```
jreidinger, line 98: here it is kind of strange...can it happen at all to be some and also empty? From code it looks to me like it cannot happen together.

rust/agama-lib/src/software/model.rs
```diff
@@ -80,3 +80,23 @@ impl TryFrom<u32> for RegistrationRequirement {
...
     }
 }
+
+/// Software resolvable type (package or pattern).
+#[derive(Deserialize, Serialize, strum::Display, utoipa::ToSchema)]
+#[strum(serialize_all = "camelCase")]
+#[serde(rename_all = "camelCase")]
+pub enum ResolvableType {
+    Package = 0,
+    Pattern = 1,
```
jreidinger, line 90: btw future proof I think we will also need product resolvable type which can be used to set different product. But not related to this PR so just NP.

rust/agama-server/src/software/web.rs
```diff
@@ -463,3 +464,27 @@ async fn probe(State(state): State<SoftwareState<'_>>) -> Result<Json<()>, Error
...
+        (status = 400, description = "The D-Bus service could not perform the action
+")
+    )
+)]
+async fn set_resolvables(
+    State(state): State<SoftwareState<'_>>,
+    Path(id): Path<String>,
+    Json(params): Json<ResolvableParams>,
+) -> Result<Json<()>, Error> {
+    let names: Vec<_> = params.names.iter().map(|n| n.as_str()).collect();
```
jreidinger, line 484: why type is needed here? I kind of expect it can infer that it is Vector from using collect.

rust/agama-lib/src/store.rs
imobachgs: I would say that, in JSON, it can be "pre: []", so it is `Some` but empty.

rust/agama-server/src/software/web.rs
imobachgs: Yes, it is needed because the type cannot be inferred from the usage (it is used as a `[&str]` later). Bear in mind that `collect` can be used to build different kind of collections, even a `Result<T, E>`.

jreidinger (approved):

PR #1789, opened 2024-11-28
First incomplete prototype of the new UI

ancorgs:
First version of the agreed new UI. It's read-only, none of the selectors really work (except navigating to iSCSI configuration).

nuevo1

nuevo2

nuevo3

nuevo4

There is a lot of repetition in the code (eg. there is no common widget for the embedded menus) and a lot of room for improvement. But it should be enough to build from there.

This also brings back the warning about transactional systems and removes components that would not be longer used (there are many more components to be removed, but it's a start).

transactional

Queries still need fine-tuning, but that will be done when we introduce the ability to modify the values.


joseivanlopez (approved):
LGTM

PR #1791, opened 2024-11-28
feat(storage): write config model

joseivanlopez:
TODO


web/src/queries/storage/config-model.ts
```diff
@@ -0,0 +1,162 @@
...
+  return drives.find((d) => d.name === driveName);
+}
+
+// TODO: add a second drive if needed (e.g., reusing a partition).
+function changeDrive(model: configModel.Config, driveName: string, newDriveName: string) {
+  const drive = findDrive(model, driveName);
+  if (drive === undefined) return;
+
+  drive.name = newDriveName;
+  if (drive.spacePolicy === "custom") drive.spacePolicy = "keep";
```
ancorgs, line 40: Why "keep" here without even a TODO? I would have expected to use the default product policy.

web/src/components/storage/DriveEditor.tsx
```diff
@@ -88,11 +88,16 @@ const SpacePolicySelector = ({ drive, driveDevice }: DriveEditorProps) => {
...
   const [isOpen, setIsOpen] = useState(false);
+  const navigate = useNavigate();
   const setSpacePolicy = useSetSpacePolicy();
   const onToggle = () => setIsOpen(!isOpen);
-  const onSpacePolicyChange = (spacePolicy: "keep" | "delete" | "resize") => {
-    setSpacePolicy(drive.name, spacePolicy);
-    setIsOpen(false);
+  const onSpacePolicyChange = (spacePolicy: "keep" | "delete" | "resize" | "custom") => {
+    if (spacePolicy === "custom") {
+      return navigate("/storage/space-policy/" + baseName(drive.name));
```
ancorgs: This reminds me we had some conversations about what to use as id (basenames vs aliases vs something else). Is there something documented on that regard?

doc/storage_proposal_from_profile.md
```diff
@@ -0,0 +1,210 @@
...
+of a config object. As result, the config object is totally complete and ready to be used by the agama
+proposal.
+
+### How sizes are solved
+
+A volume size in the profile:
+
+* Can be totally omitted.
+* Can omit the max size.
+* Can use "custom" as value for min and/or max.
```
ancorgs, line 98: I get you mean "current".

ancorgs: You use "custom" instead of "current" all along the document.

doc/storage_proposal_from_profile.md
```diff
@@ -0,0 +1,210 @@
...
+  }
+]
+```
+
+The config conversion generates:
+
+```ruby
+partition.size.default? #=> false
+partition.size.min      #=> nil
+partition.size.max      #=> Y2Storage::DiskSize.Unlimited
```
ancorgs, line 164: Unlimited or nil?

ancorgs: Never mind. The example is correct.

web/src/components/storage/DriveEditor.tsx
joseivanlopez: It is still undecided. But I guess name/basename cannot be used for devices that are going to be created.

web/src/queries/storage/config-model.ts
joseivanlopez: As discussed in a chat, the are going to follow this approach:

* Use "keep" for a device that is added as effect of using it as boot device.
* Honor the default policy defined by the product in other cases.

@ancorgs Applying this change would be easier to do in the follow-up PR https://github.com/agama-project/agama/pull/1808. This part of the code has changed a bit. Do you mind if we keep this as it is with a TODO?

ancorgs: No problem.

ancorgs (approved):
I could not review every single change because the PR is huge. But I looked at it overall and I tested it myself.

So far, LGTM.

PR #1792, opened 2024-11-29
feat(rust): run postscripts in a chroot

imobachgs:
## Problem

The current scripting support does not allow using AutoYaST chroot-like scripts. For instance, if you want to enable a system at the end of the installation, you need to deal with the chroot yourself.

## Solution

This change extends the post-scripts to run on a chroot on the target system. A definition like the following will enable the sshd daemon at the end of the installation:

```jsonnet
scripts: {
  post: [
    { name: 'test', body: '#!/bin/bash\nsystemctl enable sshd', chroot: true },
  ],
}
```

Now, chroot-ed postscripts are the default. Since you most likely want to run postscripts on a chroot, we are supporting `chroot: false` for backward compatibility reasons.

## Refactoring

The PR contains a heavy refactoring of the scripts model. Now, we are using an enum-based approach that makes it easier to cope with the potential differences between script types, given that not all are equal.

By now, the only difference is the `chroot` attribute. However, in the future, we expect that prescripts and postscripts will get support for additional features, like progress reporting, that will not be available for init scripts.

## Testing

- *Tested manually*


rust/agama-lib/src/scripts/model.rs
```diff
@@ -172,12 +312,53 @@ impl Default for ScriptsRepository {
...
+    chroot: bool,
+}
+
+impl ScriptRunner {
+    fn new() -> Self {
+        Default::default()
+    }
+
+    fn with_chroot(self, chroot: bool) -> Self {
+        Self { chroot }
```
jreidinger, line 329: if we plan to have more args in future I kind of expect code like `self.chroot = chroot; self` that will just modify self so it does not loose other params.

jreidinger:
Overall change looks good, but I expect to see somewhere documentation for it including example of usage.

jreidinger (approved):

Issue #1739, opened 2024-11-08
Unattended installation ended with error


serock:
I tried an unattended installation of Leap 16.0 using the `agama-installer.x86_64-10.0.0-openSUSE-Build27.7.iso` in VirtualBox 7.0.18 on Leap 15.6. At the end of the installation, an error was displayed:
![Screenshot from 2024-11-08 03-10-39](https://github.com/user-attachments/assets/ac3a579d-32ed-4eff-8442-dd6c6e5b79c2)

The effective [profile.json](https://github.com/user-attachments/files/17677228/profile.json) is attached with the user and password info omitted.

The end of the `agama.out.log` had:

```
Nov 08 09:08:01 agama agamactl[3007]: [INFO]: software: Commit result [688, [], [], [], []]
Nov 08 09:08:01 agama agamactl[2991]: [INFO]: manager: Configure the system (3/3)
Nov 08 09:08:02 agama agamactl[39077]: /mnt: /etc/locale.conf written.
Nov 08 09:08:02 agama agamactl[39077]: /mnt: /etc/vconsole.conf written.
Nov 08 09:08:02 agama agamactl[39077]: /mnt: /etc/localtime written
Nov 08 09:08:15 agama agamactl[2989]: Loading mocked module /usr/lib64/ruby/gems/3.3.0/gems/agama-yast-10.devel407/lib/agama/dbus/y2dir/modules/InstFunctions.rb
Nov 08 09:08:20 agama agamactl[2991]: [INFO]: manager: Install phase done
```

The installation of Leap 16.0 seemed to be successful; I was able to reboot the virtual machine and login to Leap 16.0.

imobachgs:
Thanks for the report. The installation was finished successfully, but the web UI crashed. We will try to reproduce the problem.

imobachgs:
I was able to reproduce the problem using [this profile](
[bug.json](https://github.com/user-attachments/files/17703254/bug.json). It seems to related to the storage layer, as I see two failing requests to `api/storage/proposal/settings` and `api/storage/proposal/actions`.

I am attaching some debugging information:

* [Agama logs](https://github.com/user-attachments/files/17703277/agama-logs.tar.gz)
* [Browser console logs](https://github.com/user-attachments/files/17703278/console-export-2024-11-11_14-44-20.txt)

imobachgs:
[...]

PR #1793, opened 2024-11-29
fix(web): Try to not crash in unattended installation because of the storage section (#1739).

teclator:
## Problem

The **UI** is not ready for the new storage configuration loaded from the profile.

- https://github.com/agama-project/agama/issues/1739
- https://trello.com/c/3BVQ3zl5

## Solution

Adapted the installation finished page using the storage config instead of the proposal settings as it could not exist.

As the problem is reported just for unattended installation and the finished page we will provide a fix only for the reporeted issue as the whole UI should be ready when the new storage features [branch](https://github.com/agama-project/agama/tree/storage-config-ui) is merged 

## Testing

- *Tested manually*


web/src/api/storage/types/config.ts
```diff
@@ -0,0 +1,369 @@
+/* eslint-disable */
+/**
+ * This file was automatically generated by json-schema-to-typescript.
```
imobachgs, line 3: We should decide on a tool to generate those files, but that's out of the scope of this PR.

web/src/queries/storage.ts
```diff
@@ -123,6 +124,17 @@ const useDevices = (
...
 };
 
+const configQuery = {
+  queryKey: ["storage", "config"],
+  queryFn: fetchConfig,
+};
+
+const useConfig = () => {
+  const { data } = useSuspenseQuery(configQuery);
+  console.log(data);
```
imobachgs, line 134: left-over

imobachgs (changes requested):
It looks good. Please, update the changes file.

web/src/api/storage/types/config.ts
teclator: Sure, BTW this types are fetched from the storage-config-ui branch (https://github.com/agama-project/agama/blob/storage-config-ui/web/src/api/storage/types/config.ts)

imobachgs (approved):

dgdavid (approved):

PR #1797, opened 2024-12-02
fix(rust): upgrade rustc dependency to 1.81

imobachgs:
zbus 5 requires rustc 1.81 or higher. This PR updates bumps the rustc dependency from 1.75 o 1.81 to
meet this requirement.


ancorgs (approved):
LGTM

PR #1803, opened 2024-12-03
fix using agama.install_url with equal in param

jreidinger:
## Problem

We get report that using agama with `agama.install_url=cd:/?devices=/dev/sr1` to read repository from second dvd does not work.

## Solution

Fix parsing cmdline arguments which do too much splitting for each argument.

## Testing

- *Added a new unit test*
- *Tested manually* without VPN and also completely without network


service/test/agama/cmdline_args_test.rb
```diff
@@ -29,8 +29,27 @@
...
+
+    it "converts 'true' and  'false' values into booleans" do
+      args = described_class.read_from(File.join(workdir, "/proc/cmdline"))
+      expect(args.data["web"]).to eql({ "ssl" => true })
+    end
+
+    it "converts keys that contain another dots after 'agama.' to hash value and using first value as key" do
+      args = described_class.read_from(File.join(workdir, "/proc/cmdline"))
+      expect(args.data["web"]).to eq({ "ssl" => true })
+    end
```
dgdavid: These examples has identical expectations. Is it ok?

service/test/agama/cmdline_args_test.rb
```diff
@@ -29,8 +29,27 @@
...
+
+    it "properly parse values that contain '='" do
+      args = described_class.read_from(File.join(workdir, "/proc/cmdline"))
+      expect(args.data["install_url"]).to eq("cd:/?devices=/dev/sr1")
+    end
```
dgdavid, line 54: Is there a missing example for testing the "auto" value?

jreidinger: no, for auto it is plain option and it is in top level one at https://github.com/agama-project/agama/pull/1803/files#diff-68c3c88c1cfcfdc461f9d92425b028aa4ed3ca2a8c6239f2325b01fc4b0c8567R32
here it is intentionally using real cd schema that user complain on slack

dgdavid (approved):

service/test/agama/cmdline_args_test.rb
jreidinger: yes, as both test uses original fixture, so I would like to here emphasize that it do "hashisize" of key and also normalize of value. My plan was to use separate examples with documented ones from https://agama-project.github.io/docs/user/boot_options sadly noone of those options are documented, so I just keep it for testing regressions.

PR #1804, opened 2024-12-03
Disable adjust_by_ram for swap at all products

ancorgs:
## Problem

As part of  #1111 we enabled `adjust_by_ram` for the swap. Only in the case of Tumbleweed and to (quoting from that PR description) "raise awareness and get feedback".

But somehow that setting ended up spreading over all products, even those in which linking the swap size to the RAM size by default doesn't seem to make much sense.

Very few people is installing the distributions in a setup where suspending to RAM is needed. On the other hand, many testers want a default set of settings that work out-of-the-box since they are not really interested in the storage setup.

## Solution

The original goal of getting feedback is achieved. So let's switch the feature back off to ease testing while we decide what the next steps will be regarding `adjust_by_ram` and its default value.

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Dec  3 11:22:09 UTC 2024 - Ancor Gonzalez Sosa <ancor@suse.com>
+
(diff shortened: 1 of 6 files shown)


teclator (approved):
LGTM

PR #1807, opened 2024-12-04
quick POC to fix registration:

jreidinger:
not tested!


imobachgs:
Definitely, a step in the right direction. Thanks!

PR #1808, opened 2024-12-04
feat(storage): configure boot

joseivanlopez:
* Add support in the config model for configuring boot. 
* The *Configure boot* option in the *More options* menu goes to the page for configuring boot.
* Adapt the page for configuring boot to use the new config model hooks.

localhost_8080_ (20)


service/lib/agama/storage/config_checkers/boot.rb
```diff
@@ -0,0 +1,78 @@
...
+          return unless configure? && device_alias.nil?
+
+          error(_("There is no boot device alias"))
+        end
+
+        # @return [Issue, nil]
+        def invalid_alias_issue
+          return unless configure? && device_alias && !valid_alias?
+
+          # TRANSLATORS: %s is the replaced by a device alias (e.g., "boot").
```
ancorgs, line 65: "is the replaced"

service/lib/agama/storage/configs/with_alias.rb
```diff
@@ -33,6 +33,11 @@ module WithAlias
         def alias?(value)
           self.alias == value
         end
+
+        # Ensures the config has a value for alias.
+        def ensure_alias
+          self.alias ||= "device#{rand(10**9)}"
+        end
```
ancorgs, line 40: In yast-storage-ng we use SecureRandom, which feels more... well... secure. ;-)

https://github.com/yast/yast-storage-ng/blob/00c076d27ef8186398cd47cccde95588a90601e5/src/lib/y2storage/planned/device.rb#L61

web/src/queries/storage/config-model.ts
```diff
@@ -26,39 +26,129 @@ import { configModel } from "~/api/storage/types";
...
+}
+
+function isExplicitBoot(model: configModel.Config, driveName: string): boolean {
+  return !model.boot?.device?.default && driveName === model.boot?.device?.name;
+}
+
+function setBoot(originalModel: configModel.Config, boot: configModel.Boot) {
+  const model = copyModel(originalModel);
+  const name = model.boot?.device?.name;
+  const remove = name !== undefined && isExplicitBoot(model, name) && !isUsedDrive(model, name);
```
ancorgs, line 76: Just to check I understood it correctly (I'm not against it). When a drive is not longer necessary after a change at the boot device, that drive is removed here. On the other hand, when a new drive needs to get added because a change at the boot device, that drive is added at the backend (during the conversion from the JSON model to config).

Right?

web/src/components/storage/BootSelection.tsx
```diff
@@ -126,20 +132,20 @@ partitions in the appropriate disk.",
...
-  const setBootDevice = (v) => {
+  const changeBootDevice = (v) => {
     setState({ ...state, bootDevice: v });
   };
 
   return (
     <Page>
       <Page.Header>
-        <h2>{_("Select booting partition")}</h2>
+        <h2>{_("Boot options")}</h2>
```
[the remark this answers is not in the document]
joseivanlopez: Good point!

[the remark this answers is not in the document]
ancorgs: Anyway, this is not a blocker for merging. Let's do a demo at the next review meeting that discuss those details afterwards.

ancorgs (approved):
LGTM

PR #1809, opened 2024-12-05
feat: implement product registration support

imobachgs:
## Solution

This pull request implements support for product registration. It includes changes at several levels:

- Adapt the Ruby part to handle the registration using `/run/agama/zypp` as target directory.
- Add the information about whether a product should be registered or not in the product (and not through a separate D-Bus attribute). Propagate those changes to the HTTP API too.
- Implement a registration page to the web user-interface.

## Testing

You can use the web UI which sends the proper request to the backend or, if you prefer, you can play around with cURL:

```
TOKEN=$(curl -k --silent $AGAMA_URL/api/auth -d '{"password": "linux"}' \
	-H "Content-Type: application/json" | jq .token | tr -d '"')

echo "Content-Type: application/json" >headers.txt
echo -n "Authorization: Bearer " >>headers.txt
echo $TOKEN >>headers.txt

curl -k -H @headers.txt -X PUT \
	$AGAMA_URL/api/software/config \
	-d "{\"product\": \"SLES\"}"

curl -k -H @headers.txt -X POST \
	$AGAMA_URL/api/software/registration \
	-d "{\"key\": \"YOUR-REGISTRATION-CODE\", \"email\": \"user@example.net\"}"
```

Set `$AGAMA_URL` to the Agama HTTP API (e.g., `https://192.168.122.10`).

## Screenshots

reg2
reg1


web/src/components/product/ProductSelectionPage.test.tsx
```diff
@@ -48,7 +49,6 @@ const microOs: Product = {
   registration: "no",
 };
 
-const mockConfigMutation = jest.fn();
```
jreidinger, line 44: thanks for fix. It was caused by my bad manual solve of conflict.

jreidinger (approved):

PR #1810, opened 2024-12-05
fix reading global credentials file

jreidinger:
Work-around issue with suse-connect cannot read credentials file.


lslezak (approved):
OK as a workaround, I do not know the details...

Issue #1756, opened 2024-11-13
Missing wifi drivers (T14s) Qualcomm


lkocman:
Simiar to #1752 we seem to be lacking WiFi drivers that were previously present in installation-images. 

https://github.com/openSUSE/installation-images/blob/fd8629fc4b4b5810b668881fa430c30cc81056f3/obs/installation-images.spec#L520

lkocman:
Interestingly no issues on T495, and our Tuxedo.

lkocman:
I suppose we need to add firmware for Qualcomm Atheros

lkocman@localhost:~> rpm -qa | grep firmware | grep ath
kernel-firmware-ath12k-20240728-150600.3.6.1.noarch
kernel-firmware-atheros-20240728-150600.3.6.1.noarch
kernel-firmware-ath11k-20240728-150600.3.6.1.noarch
kernel-firmware-ath10k-20240728-150600.3.6.1.noarch

lkocman:
But I don't see any of these here https://github.com/openSUSE/installation-images/blob/fd8629fc4b4b5810b668881fa430c30cc81056f3/obs/installation-images.spec#L586

lslezak:
No, that's not a problem. We use `kernel-firmware` which is a bundle of all firmware packages. The `kernel-firmware-ath11k` and others are the split versions, they actually conflict with the `kernel-firmware` we use.

The real problem is that that the kernel modules in this case use wildcards to specify the firmware files, e.g. `ath11k/WCN6855/hw2.1/*`. Unfortunately the script which cleans up the unused firmware uses that as a raw file name without expansion and that does not match.

Another problem is that some firmware uses symlinks, again the script does not check it and it can potentially delete the symlink target if it is not referenced elsewhere and the result is dangling symlink.

I'll fix both problems....

lslezak:
Fixed in #1812

PR #1812, opened 2024-12-05
Fixed firmware cleanup script

lslezak:
## Problem

- See #1756
- Some firmware files were missing in the Live ISO

## Solution

There were basically two problems:

1. Few kernel modules use wildcards to specify the firmware files, e.g. `ath11k/WCN6855/hw2.1/*`. That was not supported by the script, it expected only individual file names referenced by drivers.
2. Some firmware files use symlinks, e.g. the driver refers to `foo.bin` file but in some cases it is actually a symlink pointing to `foo-1.0.bin`. The script then deleted the `foo-1.0.bin` file as it looked like not used by drivers. The result was a dangling symlink and a missing firmware file.

Both issues were fixed. The glob patterns are evaluated using the `Dir.glob` function and if the referenced file is a symlink then it also keeps the target file where the symlink points to.

## Testing

- Tested manually, the firmware for Atheros cards is now present
- I compared the result with installation-images and now we basically include the same firmware as installation-images do in the classic Tumbleweed installation
- The ISO size increased just by ~13MB, fortunately not that much...

### DIFF
--- a/live/root/tmp/fw_cleanup.rb
+++ b/live/root/tmp/fw_cleanup.rb
@@ -20,7 +20,7 @@ puts "Scanning kernel modules in #{dir}..."
 
 # find all symlinks until the final file is found, if a file is passed then the
 # file only is returned
-# @return Array<String> list of files/symlinks
+# @return [Array<String>] list of files/symlinks
 def symlinks(path, fw_dir)
   ret = [path]
 
--- a/live/root/tmp/fw_cleanup.rb
+++ b/live/root/tmp/fw_cleanup.rb
@@ -8,30 +8,82 @@
 require "find"
 require "shellwords"
 
+# really delete or just do a smoke test?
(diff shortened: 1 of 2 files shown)


live/root/tmp/fw_cleanup.rb
```diff
@@ -8,30 +8,82 @@
...
 fw_dir = "/lib/firmware/"
 
+# in the Live ISO there should be just one kernel installed
 dir = Dir["/lib/modules/*"].first
 puts "Scanning kernel modules in #{dir}..."
 
-# list of referenced firmware names from the kernel modules
+# find all symlinks until the final file is found, if a file is passed then the
+# file only is returned
+# @return Array<String> list of files/symlinks
```
imobachgs, line 23:
```suggestion
# @return [Array<String>] list of files/symlinks
```

imobachgs (approved):

PR #1815, opened 2024-12-05
Add Translators section to main README

mvidner:
So that it can be linked from https://l10n.opensuse.org/projects/agama/#information

Motivation: a translator wondered why their work has not showed up in a new image build. It turns out we only sync translations on Mondays. The docs will help answer such questions faster


bittin:
good to know thanks :)

lslezak (approved):
Thanks for improving the documentation!

PR #1821, opened 2024-12-09
refactor(web): simplify page to enforce root password

dgdavid:
## Problem

During a review meeting on December 9th, it was argue that the [new screen for enforcing at least one root authentication method](https://github.com/agama-project/agama/pull/1787) could be intimidating for some users, mainly due to the inclusion of the [SSH public key](https://www.ssh.com/academy/ssh/public-key-authentication) input field.

## Solution

In the same meeting, it was suggested to hide the SSH public key input and display it only upon user request, taking the opportunity for also providing a brief explanation of what an SSH public key is.

While this was a good idea, after testing a couple of approaches it became apparent that there were several corner cases, all of them leading to add too much user interaction to an already intrusive screen that should be kept as minimalist as possible.

Thus, the input has been completely removed instead. Now, the page ask for the root password and provides  information to let users aware about the possibility to change the root authentication method at any time before installation from the _Users_ section.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

| Before | After |
|-|-|
| Screen Shot 2024-12-09 at 14 08 54 | Screen Shot 2024-12-09 at 14 08 43 |

---

Related to https://trello.com/c/GQytAUEU (internal link)

### DIFF
--- a/web/src/assets/styles/app.scss
+++ b/web/src/assets/styles/app.scss
@@ -97,9 +97,3 @@ button.remove-link:hover {
     --pf-v5-c-notification-drawer__list-item--before--BackgroundColor: none;
   }
 }
-
-form#rootAuthMethods {
-  .pf-v5-c-file-upload__file-select {
-    display: none;
-  }
-}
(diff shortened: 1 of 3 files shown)


lslezak (approved):
Yes, looks much better now! Thanks! :clap:

Issue #1818, opened 2024-12-06
s390x installation can't be finished


mmaslano:
I tried to install agama-installer.s390x-10.0.0-SLE-Alpha2.iso, which failed to boot after the installation. But it was possible to pick dasd disk, activate, format and storage management was happy. I didn't setup partitions and used default.

With agama-installer.s390x-10.0.0-SLE-Build38.1.iso I'm unable to finish the installation because of the storage. I activated dasd, formatted, but I still see "Storage proposal not possible". I disabled btrfs snapshots, sadly I was unable to change partitions fs from btrfs to xfs for /. It looks like it's not possible to delete / and start over and edit doesn't help.

I'm attaching logs, but I doubt it's helpful. Maybe the right button is there, but I can't find it, which will be also a problem for customers. 

[agama-logs-1733517061.tar.gz](https://github.com/user-attachments/files/18039492/agama-logs-1733517061.tar.gz)

kobliha:
Maybe related to #1758

ancorgs:
The problem has nothing to do with Btrfs vs XFS or with enabling/disabling snapshots.

Storage cannot make a proposal because there is no disk considered as a valid candidate. I will investigate why.

ancorgs:
Reading the logs...

After activating the DASD, a so-called storage reprobing is performed and dasda is considered as an acceptable disk:

> y2storage/disk_analyzer.rb(candidate_disks):142 Found candidate disks: [<Dasd /dev/dasda 21600000 KiB (20.60 GiB)>]

But then Agama makes a new proposal and seems to ignore that disk for some reason I would need to investigate:

> y2storage/guided_proposal.rb(try_with_target):179 Trying to make a proposal
> using the following settings:
>    root_device: 
>    candidate_devices: [""]

ancorgs:
@mmaslano Gave me access to the machine and I reproduced the bug. The problem is that, after performing the DASD operations, Agama tries to update the list of installation issues. But that process is failing because `storage_manager.proposal` is nil when this line is executed:

https://github.com/agama-project/agama/blob/master/service/lib/agama/storage/proposal_strategies/guided.rb#L65

Everything goes downhill from there.

I will commit a fix.

ancorgs:
#1822 implements a workaround. The root of the issue is not fixed, but the failing code path will disappear in a couple of months when the new storage proposal is finished.

I will keep the issue open to make sure we test this scenario with the new storage proposal.

ancorgs:
[...]

PR #1822, opened 2024-12-09
service: Workaround for error on s390x

ancorgs:
## Problem

As described at #1818, there seems to be a problem when installing at s390x and starting with no available disks. See [this comment](https://github.com/agama-project/agama/issues/1818#issuecomment-2527658289) at the original issue.

After visiting the DASD section and activating+formatting a device, the backend fails and the UI becomes barely usable.

The error is produced when checking `storage_manager.proposal` after re-probing the system and trying an (unsuccessful) proposal.

## Solution

This pull request does not implement a real fix, but a workaround that has proven to be useful in that case. It allowed the installation to continue successfully at the reported scenario.

I tried to fix the real root of the issue, but I was unable to find out why `storage_manager.proposal` was nil at that point.

But since we are rewriting the approach to the storage proposal, the whole `ProposalStrategies::Guided` class is going to disappear so it makes little sense to invest more time.

I will keep the original issue open and will create a Trello card to track it, so we check whether the new implementation of the storage proposal (still under development) is reliable in that situation.

## Testing

Verified with manual testing. Going further would be wasting resources right now.

### DIFF
--- a/service/lib/agama/storage/proposal_strategies/guided.rb
+++ b/service/lib/agama/storage/proposal_strategies/guided.rb
@@ -62,6 +62,9 @@ module Agama
 
         # @see Base#issues
         def issues
+          # Returning [] in case of a missing proposal is a workaround (the scenario should
+          # not happen). But this class is not expected to live long.
+          return [] unless storage_manager.proposal
           return [] unless storage_manager.proposal.failed?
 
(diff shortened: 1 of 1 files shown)


dgdavid (approved):
:+1:

jreidinger (approved):

PR #1826, opened 2024-12-10
fix(web): close issues drawer when users click on a section

dgdavid:
## Problem

During the presentation of the [new location for the pre-installation checks](https://github.com/agama-project/agama/pull/1778) at a team meeting on December 9th, it was discovered that the drawer doesn't close when users click on the section they are currently viewing.

## Solution

 As a quick fix, the onClose callback is triggered when users click on section names. This addresses the issue in a simple way until having time to consider other behavior options, such as not rendering the current section as a link in the drawer but using different elements to hint users they are already on that section.

## Testing

- Added a new unit test
- Tested manually


dgdavid (commented):
Reviewers, I think this does not reserve an entry in the changes file. Please, let me know if you prefer adding one instead.

PR #1828, opened 2024-12-11
fix(web): do not render Questions component at login path

dgdavid:
## Problem

In #1825 questions were placed at `Layout` component level for ensuring they are always rendered across app. However, it makes the login page crash because of  _No QueryClient set, use QueryClientProvider to set one_

localhost_8080_ (13)

## Solution

Stop mounting the `Questions` component when installer is at login path.

localhost_8080_ (14)

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/web/src/components/core/LoginPage.test.tsx
+++ b/web/src/components/core/LoginPage.test.tsx
@@ -22,14 +22,31 @@
 
 import React from "react";
 import { screen, within } from "@testing-library/react";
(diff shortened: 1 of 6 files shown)


dgdavid (commented):
Reviewers, I consider this as part of #1825. Let me know if you think it desires an entry in the changes file anyway.

imobachgs (approved):

PR #1831, opened 2024-12-12
fix(rust): better profile URL handling

imobachgs:
Fix for [bsc#1234362](https://bugzilla.suse.com/show_bug.cgi?id=1234362).

It fixes the behavior when importing an auto-installation profile:

* Follow redirections.
* Determine the file format from the content instead of the extension. It does not apply to AutoYaST profiles, where it still uses the extension in the URL for backward compatibility.

## Follow-up

This feature was not updated after adding the `--api`. The processing of the profile should be done at the server's side.

### DIFF
--- a/rust/agama-cli/src/profile.rs
+++ b/rust/agama-cli/src/profile.rs
@@ -22,7 +22,8 @@ use crate::show_progress;
 use agama_lib::{
     base_http_client::BaseHTTPClient,
     install_settings::InstallSettings,
-    profile::{AutoyastProfile, ProfileEvaluator, ProfileValidator, ValidationResult},
+    profile::{AutoyastProfileImporter, ProfileEvaluator, ProfileValidator, ValidationResult},
+    utils::FileFormat,
     utils::Transfer,
     Store as SettingsStore,
 };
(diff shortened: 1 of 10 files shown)


rust/agama-lib/src/utils/file_format.rs
```diff
@@ -0,0 +1,135 @@
...
+        drop(stdin);
+
+        child.wait().is_ok_and(|s| s.success())
+    }
+
+    /// Whether is is a script.
+    ///
+    /// It returns `true` if the content starts with a shebang.
+    fn is_script(content: &str) -> bool {
+        content.starts_with("#!")
```
jreidinger, line 101: be aware that shebang is not mandatory for shell scripts and also that after shebang can be something like python :)
In general as this means that shell scripts without shebang ends with "unknown file" it think it would be nice to combine both approaches...suffix and content. I would say ideally if there is suffix base decision on it as it is fast...and then check content if suffix is not there or unknown.

imobachgs: Yes, it is optional, but you cannot trust the suffix (as the script comes from a URL, you might not even have a suffix). I think we can educate our users to include a shebang line :-)

Actually, I do not know if using something different from bash would work at all. Check [how we build the command](https://github.com/agama-project/agama/blob/master/rust/agama-cli/src/profile.rs).

rust/agama-cli/src/profile.rs
```diff
@@ -114,48 +115,59 @@ async fn import(url_string: String, dir: Option<PathBuf>) -> anyhow::Result<()>
...
+    match FileFormat::from_file(&tmp_profile_path)? {
+        FileFormat::Jsonnet => {
+            let file = File::create(path)?;
+            let evaluator = ProfileEvaluator {};
+            evaluator
+                .evaluate(&tmp_profile_path, file)
+                .context("Could not evaluate the profile".to_string())?;
+        }
+        FileFormat::Script => {
+            let err = Command::new("bash").args([&tmp_profile_path]).exec();
```
mvidner, line 161: As Josef said, we are checking for *any* script, even Python, but then run it with `bash`.
1. Either check for `#!/bin/bash`,
2.  or make it executable and `Command::new(tmp_profile_path)`

imobachgs: Yes, that's why I pointed to how we run those scripts after reading @jreidinger's comment . For the scripting support, I followed the 2nd approach (make it executable).

However, when handling the agama.auto we have never considered anything different from bash.

So we can keep the current check and make it executable; or just check for bash.

imobachgs: Just to clarify, this line of code [was already there](https://github.com/agama-project/agama/blob/6284cbfebeb4be8b7808120ef8a7d72d91b91569/rust/agama-cli/src/profile.rs#L140). It was just moved in this PR.

rust/agama-lib/src/utils/file_format.rs
```diff
@@ -98,7 +98,10 @@ impl FileFormat {
     ///
     /// It returns `true` if the content starts with a shebang.
     fn is_script(content: &str) -> bool {
-        content.starts_with("#!")
+        let Some(first_line) = content.lines().next() else {
+            return false;
+        };
+        first_line.starts_with("#!") && !first_line.contains("jsonnet")
```
jreidinger, line 104: we should add small docu line explaining why we cannot handle jsonnet as others ( we need injection and also it produce json unlike other scripts which is standalone and using agama CLI )

jreidinger (approved):

rust/agama-lib/src/utils/file_format.rs
```diff
@@ -96,7 +96,9 @@ impl FileFormat {
 
     /// Whether is is a script.
     ///
-    /// It returns `true` if the content starts with a shebang.
+    /// It returns `true` if the content starts with a shebang. However, it excludes the
+    /// case of a "jsonnet" script because it needs special handling: injecting the hardware
+    /// information and processing its output.
```
jreidinger, line 101: :+1:

jreidinger (approved):

PR #1833, opened 2024-12-13
Rename "encrypted password" to "hashed password"

lslezak:
- As discussed on the review meeting, technically it is not an encryption (because the decryption part is missing) but hashing. That's the correct term.

## Testing

- Updated unit tests
- Tested manually

```console
# agama config show | jq .user
{
  "fullName": "Agama",
  "userName": "agama",
  "password": "$5$t3gRfqthxk7ZsJw.$gpduJASf12Xk8aPtIllniJAwhbhlYvLcEUH0.4P9MWA",
  "hashedPassword": true,
  "autologin": false
}
```


dgdavid (approved):
LGTM, thanks

Issue #1846, opened 2024-12-20
web: not possible to add a connection address using a netmask


dgdavid:
While trying to add an IP address to a connection using a netmask instead of a prefix length the request fails silently.

It is a problem from two point of view:

* The form says user can users a netmask, but actually they cannot
* The form is sent as successful request, but actually it failed

The root of the problem could be in the [`formatIp` network utility](https://github.com/agama-project/agama/blob/1477848fb691fd5aebecc7ccafdd536131d1fc6e/web/src/utils/network.ts#L131-L137),  which does not take care to convert the netmask to a prefix length. It is called by [`connection#toApi`](https://github.com/agama-project/agama/blob/1477848fb691fd5aebecc7ccafdd536131d1fc6e/web/src/types/network.ts#L317-L330).

Or maybe the backend should allow netmask too.

Below the error from axios

```json
{
    "message": "Request failed with status code 422",
    "name": "AxiosError",
    "stack": "AxiosError: Request failed with status code 422\n    at settle (http://localhost:8080/index.js:129393:12)\n    at XMLHttpRequest.onloadend (http://localhost:8080/index.js:127842:66)\n    at Axios.request (http://localhost:8080/index.js:128382:41)",
    "config": {
        "transitional": {
            "silentJSONParsing": true,
            "forcedJSONParsing": true,
            "clarifyTimeoutError": false
        },
        "adapter": [
            "xhr",
            "http",
            "fetch"
        ],
        "transformRequest": [
            null
        ],
        "transformResponse": [
            null
        ],
        "timeout": 0,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1,
        "maxBodyLength": -1,
        "env": {},
        "headers": {
            "Accept": "application/json, text/plain, */*",
            "Content-Type": "application/json"
        },
        "responseType": "json",
        "method": "put",
        "url": "/api/network/connections/Wired connection 1",
        "data": "{\"id\":\"Wired connection 1\",\"status\":\"up\",\"nameservers\":[],\"method4\":\"manual\",\"method6\":\"auto\",\"interface\":\"enp1s0\",\"addresses\":[\"192.168.122.195/24\",\"192.168.122.196/22\",\"192.168.122.197/255.255.255.0\"]}"
    },
    "code": "ERR_BAD_REQUEST",
    "status": 422
}
```

dgdavid:
[...]

PR #1847, opened 2024-12-20
fix(web): proper netmask handling

dgdavid:
## Problem

Backend expects a prefix length instead of netmask, but unfortunately current code base is not ensuring of making the conversion before sending a request, making the connection form fail silently.

## Solution

Adapt the `formatIp` util method for relying in `ipPrefixFor` to make such a conversion.

## Testing

- Updated unit test
- Tested manually

---

Fixes https://github.com/agama-project/agama/issues/1846

Related to https://github.com/agama-project/agama/pull/530


ancorgs (approved):
LGTM

PR #1848, opened 2024-12-20
fix(storage): recover config if needed

joseivanlopez:
## Problem

In autoinstallation, the storage config could be set before probing storage. If so, the probing action overwrites the config that was set from the autoinstall profile.

https://bugzilla.suse.com/show_bug.cgi?id=1234711

## Solution

Hotfix for ensuring the storage config is preserved when doing probing for first time. A better solution should be provided for a correct management of the services and precedences.

## Testing

Manually tested.

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -83,30 +83,19 @@ module Agama
       service_status.idle
     end
 
-    def locale=(locale)
-      service_status.busy
-      change_process_locale(locale)
-      users.update_issues
-      start_progress_with_descriptions(
-        _("Load software translations"),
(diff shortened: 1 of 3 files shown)


service/lib/agama/manager.rb
```diff
@@ -159,6 +148,21 @@ def install_phase
     end
     # rubocop:enable Metrics/AbcSize
 
+    def locale=(locale)
```
imobachgs, line 151: Thanks for moving this. I always think that it is misplaced, but I never fix it.

imobachgs (approved):
Thanks for such a quick fix!

PR #1840, opened 2024-12-16
Bootloader force user input

jreidinger:
## Problem

In order to support the QA use-case, we need to make it possible to configure the boot loader to stop at the menu without a timeout.

trello: https://trello.com/c/dKYmc0K7/4018-5-agama-ped-10810-new-section-to-configure-bootloader
jira:
- [PED-10810](https://jira.suse.com/browse/PED-10810)
- Epic: https://jira.suse.com/browse/AGM-54
- Story: https://jira.suse.com/browse/AGM-59
- Tasks: https://jira.suse.com/browse/AGM-67 and https://jira.suse.com/browse/AGM-68

## Solution

Add to profile section `Bootloader` with key `stopOnBootMenu` and boolean value. This allows QA to use `jq -n ".bootloader.stopOnBootMenu = true" | agama config load` after selecting product and before start of installation to force stop without any timeout.

## Testing

- *Added a new unit test*
- *Tested manually*

PR #1854, opened 2024-12-23
Fix storage config

jreidinger:
## Problem

There is collision between https://github.com/agama-project/agama/pull/1848 and https://github.com/agama-project/agama/pull/1840 where bootloader adds another interface with Get/SetConfig on Storage service. This result that generic access no longer works and needs access via specific interface.

reported on slack https://suse.slack.com/archives/C02TLF25571/p1734924853266999

## Solution

Change code to avoid ambitious code.

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/dbus/clients/storage.rb
+++ b/service/lib/agama/dbus/clients/storage.rb
@@ -67,7 +67,8 @@ module Agama
         #
         # @return [Hash]
         def config
-          serialized_config = dbus_object.GetConfig
+          # Use storage iface to avoid collision with bootloader iface
+          serialized_config = dbus_object[STORAGE_IFACE].GetConfig
           JSON.parse(serialized_config, symbolize_names: true)
         end
 
@@ -76,7 +77,8 @@ module Agama
         # @param config [Hash]
         def config=(config)
           serialized_config = JSON.pretty_generate(config)
-          dbus_object.SetConfig(serialized_config)
+          # Use storage iface to avoid collision with bootloader iface
+          dbus_object[STORAGE_IFACE].SetConfig(serialized_config)
         end
 
       private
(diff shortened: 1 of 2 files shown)


service/lib/agama/dbus/clients/storage.rb
```diff
@@ -67,7 +67,8 @@ def finish
         #
         # @return [Hash]
         def config
-          serialized_config = dbus_object.GetConfig
+          # Use storage iface to avoid collision with bootloader iface
```
dgdavid, line 70: Interesting. I wonder if it worth to always use the corresponding iface to avoid more errors like the one this commit fixes and get rid of specific comment. 

Something to be discussed, not a blocker for this PR.

dgdavid (approved):
Although DBus stuff are not my thing, rationale and code looks good to me. If you tested it and it works, I guess you can go ahead.

Nevertheless, I have left a comment to be discussed in a near future, if needed.

service/lib/agama/dbus/clients/storage.rb
jreidinger: yeah, it is zbus way to always use combination of service, interface and object. On other hand ruby-dbus use different way and allow access throught interface, but for simplicity if method in unique, it is added directly to object. My original fix change dbus_object to always return interface, but then mixins start failing as it uses some methods that is not in interface. (during test e.g. this mixins failing https://github.com/agama-project/agama/blob/master/service/lib/agama/dbus/clients/with_service_status.rb#L43 ).
Maybe it will be better to have beside dbus_object also dbus_interface method on which we should call all needed storage methods ( that is idea that I get after thinking a bit more about it after I have working solution and not enough time to test different one ).

PR #1858, opened 2025-01-03
migrate answers example from yaml to json

jreidinger:
## Problem

When changing answers format from yaml to json example file was not converted.

## Solution

Convert it and also use example which provides password.


mvidner (approved):
May I suggest a reformatted version (via `python -m json.tool`)

```json
{
    "answers": [
        {
            "answer": "decrypt",
            "class": "storage.luks_activation",
            "password": "my_password"
        }
    ]
}
```

mvidner (approved):
Still LGTM :)

PR #1866, opened 2025-01-07
Initial implementation for storing Agama configuration provided by kernel cmdline

teclator:
## Problem

As part of the linuxrc compatibilty Jira entry, we want to make sure several boot arguments are translated to environment variables that indeed affect the execution of Agama.

That includes the following patterns: LIBSTORAGE_*, YAST_*, Y2*, ZYPP_FULLLOG.

Some examples (useful for verifying everything actually works): LIBSTORAGE_MULTIPATH_AUTOSTART, LIBSTORAGE_MDPART, YAST_ACTIVATE_LUKS, YAST_REUSE_LVM, Y2DEBUG, ZYPP_FULLLOG

- https://trello.com/c/WDefGcX1

## Solution

The configuration that is provided by the kernel cmdline will be stored to /etc/agama.d/cmdline.conf and the agama services will use it as an EnvironmentFile. It is an initial implementation an probably would need some more work.

## Testing

- *Tested manually*


live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh
```diff
@@ -0,0 +1,32 @@
...
+	for _i in $CMDLINE; do
+		case $_i in
+		LIBSTORAGE_* | YAST_* | agama* | Y2* | ZYPP_*)
+			_found=1
+			;;
+		esac
+
+		if [ -n "$_found" ]; then
+			printf "Agama variable found ($_i)"
+			printf $_i
```
joseivanlopez, line 19: NP: wasn't `$_i` already printed in the line before?

teclator:
```suggestion
```

joseivanlopez:
```suggestion
```

joseivanlopez (approved):

PR #1868, opened 2025-01-07
Switch from X / IceWM to Wayland / Gnome shell

mchf:
## Problem

No problem, just willing to use Wayland and Gnome shell

## Solution

- removed IceWM package(s) and configuration
- removed (some - see later) X related packages
- installed gnome-shell, gnome-kiosk and other stuff
- provided configuration for gnome-shell / gnome-kiosk

## Caveats
1) during development it turned out that Wayland / Gnome is unusable in VirtualBox. Mutter - a wayland display server used by Gnome (or some underlying piece of stack like drivers or so) terribly crashes (core dump or freeze) in 9.5 cases out of 10. Rest of the development was done in libvirtd / virt-manager and QEMU. A discussion with a SUSE's Graphical software engineer turned there were some issues with Mesa drivers. That's where those 0.5 of working cases gets from. After Christmas vacation it started working even in VirtualBox. so probably something got fixed in the meantime.

2) It seems that "Wayland with Gnome" in TW still needs pieces of X. At least one service / script checks for existence of ```/usr/bin/X```. If not present, service crashes with status=6/NOTCONFIGURED. I also found some discussion in Arch Linux forum which kind of confirms that. So far it is not clear whether it can be workarounded by creating the above path manually or there is really a need of X server pieces to be present. X server however is not in dependencies of wayland packages, it has to be added manually. The workaround is in TBD state ;-)

3) ```firefox --kiosk``` doesn't work in wayland - shows black screen . Confirmed on various distributions and desktop environments (not only Gnome). It was confirmed even on bare metal not only in virtualization. It could be related to https://bugzilla.mozilla.org/show_bug.cgi?id=1894338 and according to the discussion it could be fixed by a newer release of firefox. Without the ```--kiosk``` option everything works. So, check of the fix is another TBD.

4) last TBD is to check whether whole Gnome pattern is neeaded or if it is possible to pick just some packages. However, this is going to be a bit time consuming.

## Testing

- *Tested manually*

## Screenshots

Agama installer in virt-manager running in Wayland / Gnome shell:
agama-installer-gnome-shell


live/src/agama-installer.kiwi
```diff
@@ -147,11 +147,9 @@
...
         <package name="rubygem(byebug)"/>
         <package name="microos-tools"/>
-        <package name="icewm-lite"/>
-        <package name="xinit"/>
-        <package name="xorg-x11-server"/>
-        <package name="xorg-x11-fonts-core"/>
-        <package name="xf86-input-libinput" />
+        <package name="patterns-base-basesystem"/>
+        <package name="patterns-gnome-gnome"/>
+        <package name="gnome-kiosk"/>
```
joantolo, line 152: Just a note, gnome-kiosk could be started without the need of GDM, that way those patterns wouldn't be needed, and the iso might be smaller.

Ref: https://github.com/joantolo/agama/commit/a73d83238c238ed9d645399db65c887364386d00#diff-3de5b8b40cc886f42c9cf046debcaa27bcfe84d2a98f98229e882dce2d3dfe0eR148

Conan-Kudo:
What about using Weston's kiosk mode instead?

live/src/agama-installer.kiwi
mchf: for the record. it somehow works now and we are in hurry ... so i suggest further improvements for a different patch

mchf:
With latest OS Tumbleweed packages Firefox works even in ```--kiosk``` mode. Firefox 136.

live/root/etc/sysconfig/displaymanager
```diff
@@ -0,0 +1,69 @@
...
+## Default:     no
+#
+# Display a combobox for Active Directory domains.
+#
+DISPLAYMANAGER_AD_INTEGRATION="no"
+## Path:        Desktop/Display manager
+## Type:        string(Xorg)
+## Default:     "Xorg"
+#
+DISPLAYMANAGER_XSERVER="Xorg"
```
imobachgs, line 69: Is this file still needed? I asking because I sat this `Xorg` reference here.

live/root/root/.local/bin/gnome-kiosk-script
```diff
@@ -0,0 +1,19 @@
...
+# This script is invoked as a part of gnome-kiosk startup procedure. See /usr/bin/gnome-kiosk-script
+# which is at the end of the chain and invokes this script.
+
+TOKEN_FILE=/run/agama/token
+TOKEN=$(cat $TOKEN_FILE)
+PREFS=$HOME/.mozilla/firefox/profile/user.js
+
+sed -e "s/__HOMEPAGE__/http:\/\/localhost\/login?token=$TOKEN/" $PREFS.template > $PREFS
+firefox --kiosk --profile $HOME/.mozilla/firefox/profile
+# Here, we would like to see
```
imobachgs, line 12: I guess this comment is outdated, right?

live/root/usr/bin/X/info.txt
```diff
@@ -0,0 +1,13 @@
...
+The directory is used to trick display-manager.service (script) to think that
+xorg-x11-server is installed. If the directory is not existing the above
+script/service ends with error reporting status=6/NOTCONFIGURED. We want to
+use Wayland (only), there is no sense in installing X11 server too. As it is
+not mentioned as direct dependency in any gnome and wayland related packages
+and kind of dependency happens only due to that check, we trick it by creating
+the directory.
+
+This is status at time of initial Wayland / GNOME Shell deployment in Agama's
+live iso. If the check is removed in the future feel free to drop whole
```
imobachgs, line 12:
```suggestion
live ISO. If the check is removed in the future feel free to drop whole
```

live/src/agama-installer.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Mar 21 10:02:14 UTC 2025 - Michal Filka <mfilka@suse.com>
+
+- Switched live iso to use wayland (gh#agama-project/agama#1781)
```
imobachgs, line 4:
```suggestion
- Switched live ISO to use Wayland (gh#agama-project/agama#1781)
```

live/src/agama-installer.kiwi
imobachgs: Please, add an issue (or a Trello card) to address this.

imobachgs (commented):
It looks good, let's see if we can finally make the switch. I just added a few comments.

live/root/root/.local/bin/gnome-kiosk-script
mchf: hopefully ;-)

live/root/etc/sysconfig/displaymanager
imobachgs: Please, add a comment at the top of the file so we do not forget about it.

live/src/agama-installer.kiwi
mchf: done

mchf: {gnome|base} pattern not required anymore

live/src/agama-installer.kiwi
```diff
@@ -146,7 +146,7 @@
         <package name="agama-auto"/>
         <package name="rubygem(agama-yast)"/>
         <package name="rubygem(byebug)"/>
-        <package name="microos-tools"/>
+        <!-- package name="microos-tools"/ -->
```
imobachgs: Please, remove this line. We were using `microos-tools` just because it included a `systemd-proxy-env` service. However, that service was moved to `aaa_base`, so it should not be needed anymore.

mchf: done

PR #1871, opened 2025-01-08
fix(web): restore cancel behavior at product selection page

dgdavid:
## Problem

The `Cancel` action at product selection page should be only rendered if there is a product already selected. By mistake, such a behavior was drop in https://github.com/agama-project/agama/pull/1769, when sticking actions to the bottom of the page.

## Solution

Restore the conditional rendering for the mentioned action and add a unit test to be aware if come change breaks it again.

## Testing

- Added a new unit test
- Tested manually


joseivanlopez (approved):
LGTM

PR #1872, opened 2025-01-08
fix(live): use gems for the default Ruby version

imobachgs:
Use `rubygem(foo)` instead of `ruby3.X-rubygem-foo` to use the gems for the project's default Ruby version.

> [!WARNING]
> Not tested yet!

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri Jan 10 09:03:06 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Depend on Ruby's default version (gh#agama-project/agama#1872).
+
+-------------------------------------------------------------------
 Tue Dec 10 12:46:06 UTC 2024 - Michal Filka <mfilka@suse.com>
 
 - Updated config.sh to enable agama-dbus-monitor service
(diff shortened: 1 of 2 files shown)


mvidner:
IIRC, we used to do this in YaST, just using whatever Ruby-version of gems that is available, but then people were reporting bugs downstream in the build process, when we got the dependencies seemingly satisfied while in fact they were mismatched.

Ideally we are pinning the version dynamically like this, but in a KIWI XML file that may not be possible
```sh
RUBY_VERSION=$(ruby -e "puts RbConfig::CONFIG['ruby_version']")
zypper --non-interactive install ... "rubygem(ruby:$RUBY_VERSION:yast-rake)" ...
```
https://github.com/agama-project/agama/blob/83ef39c67418aa12348aae3180ffa9b6d953c3f1/.github/workflows/weblate-update-pot.yml#L31-L32

imobachgs:
According to @mchf findings, it seems to work.

mchf:
I did something similar to what is in this PR here: https://github.com/agama-project/agama/pull/1868/commits/75b271525b4b273a23bea2ebc46b46dbc2fe0cd7 and was able to build the iso locally. So, this one LGTM as far as deleting those references with ruby3.2-rubygem in  some sub-products without further replacement is fine.

mchf (approved):
As commented. LGTM

PR #1874, opened 2025-01-09
PoC: Use CSS modules for the Agama components

lslezak:
## CSS modules

When working with Docusaurus I learned about the [CSS modules](https://github.com/css-modules/css-modules).

With the current implementation it is very hard to find the CSS related to the component code. For example the `FirstUserForm` component defines its CSS style in `assets/styles/app.scss` file somewhere in the middle. That's quite difficult to find.

## Pros and cons

CSS modules have several advantages:

- It allows using simple and short CSS names. To avoid conflicts we use `agama-` prefix or long names like `first-username-dropdown`. That's not nice.
- It has local scope, it avoids conflicts in CSS rules. You can decrease the chance for a conflict using a prefix but you cannot completely avoid it.  Webpack generates unique names for the CSS modules which guarantees that a conflict can never happen.
- The CSS and the JS code live next to each other, it is easy to find the related CSS. If you delete a component it is easy to delete the related CSS.

Of course, there also some disadvantages

- When inspecting the page elements you see CSS classes like `Iv3X764_hcFfxZFNQLEm` from which is not obvious what was the original name. Fortunately CSS modules support source mapping properly so you can get the original name and the exact location in the source file quite easily in the browser devtools.
- That might be problematic for the future Agama plugins. On the other hand I expect that the CSS modules will mostly contain layout fixes. Color theming, etc... will be still defined in the global CSS with known names. So I guess you should not need to change the base CSS modules from plugins.

## Notes

- The HMR (Hot Module Replacement) works fine as with normal CSS or JS files. After changing the CSS module file in editor the webpage is automatically updated when using the Webpack development server.
- CSS modules should be used only for the components. The global styles, PatternFly overrides, theming, etc... should be done as it is now.

## Proof of concept

In this example I moved the the CSS for the `FirstUserForm` component to separate `FirstUserForm.module.scss` file. We could do the same for the other components as well.

### DIFF
--- a/web/src/assets/styles/app.scss
+++ b/web/src/assets/styles/app.scss
@@ -26,16 +26,6 @@ button.remove-link:hover {
   color: var(--pf-v5-c-button--m-danger--BackgroundColor);
 }
 
-.first-username-dropdown {
-  position: absolute;
-  width: 100%;
-}
-
-.first-username-wrapper {
-  position: relative;
-  width: 100%;
-}
-
 #productSelectionForm {
   input[type="radio"] {
     align-self: center;
--- a/web/src/components/users/FirstUserForm.module.scss
+++ b/web/src/components/users/FirstUserForm.module.scss
@@ -1,0 +1,9 @@
+.dropdown {
+  position: absolute;
+  width: 100%;
+}
+
+.wrapper {
+  position: relative;
+  width: 100%;
+}
--- a/web/src/components/users/FirstUserForm.tsx
+++ b/web/src/components/users/FirstUserForm.tsx
@@ -43,6 +43,7 @@ import { _ } from "~/i18n";
 import { suggestUsernames } from "~/components/users/utils";
 import { useFirstUser, useFirstUserMutation } from "~/queries/users";
 import { FirstUser } from "~/types/users";
+import * as styles from "./FirstUserForm.module.scss";
 
 const UsernameSuggestions = ({
   isOpen = false,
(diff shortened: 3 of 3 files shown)


lslezak:
Oh, the unit tests fail, it probably misses the CSS loader...

dgdavid:
Honestly, I don not have a clear view on this yet. Sorry for that, but IMHO we have to think a bit more about it before moving ahead. But, of course, great someone else invest some time on this PoC :tada: 

That said, it is completely true that the way we are dropping CSS rules right now it is FAR from idea. And we now it, but because of priorities we are always postponing the "HTML & CSS" infrastructure. Admittedly, it is difficult to find the right one. 

About priorities, I have one after we decided to avoid creating yet another UI library: to reduce custom CSS as much as possible.

Also, think about small utils css rules that are valid for more than one component. Of course, they could live in a "utils module", but I'd like to have it into account.

I'm now migrating to PF v6, which has started to use tokens for the design and CSS, which I expect help us a lot with theming thing.

So, If you don't mind, we can still postponing this decision a bit more in order to have time to finish the migration and explore other techniques :pray: 

BTW, **thanks a ton for working in this proposal**

lslezak:
I'm fine with postponing the CSS refactoring. Let's see what PF6 brings to the table. This was rather an idea or suggestion how to solve some of the current problems.

Reducing custom CSS is a good idea, but I'm a bit afraid that you cannot reduce that to zero as you probably always need to fine tune or tweak something here or there... For that few tweaks we could use CSS modules.

For sharing/extending CSS between more components you can use the `composes` keyword: https://css-tricks.com/css-modules-part-1-need/#aa-the-composes-keyword

lslezak:
Closing as this is not relevant right now. But we should still look at later...

PR #1875, opened 2025-01-09
Activate multipath in case of forced by the user and fix environment file typo

teclator:
## Problem

- The agama service have a typo in the EnvironmentFile variable name.
- Agama is not checking if the user is forcing multipath through a environment variable.

## Solution

Fix the typo and check if multipath is forced.

## Testing

- *Tested manually*

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,6 @@
 root = true
 
 # 2 space indentation
-[{*.sh}]
+[*.sh]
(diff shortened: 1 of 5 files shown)


joseivanlopez (commented):

teclator:
Ups, already added but not commited :P

ancorgs (approved):
LGTM

PR #1876, opened 2025-01-09
Increase constraint

jreidinger:
## Problem

ppc fails to build due to run out of space

## Solution

increase disk constrain to ensure there is enough space. Probably caused by that new xtask compilation task that eats another space.

## Testing

Testing in my IBS branch https://build.suse.de/package/show/home:jreidinger:branches:Devel:YaST:Agama:Head/agama

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Jan  9 12:52:05 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
 
-- Increase disk size in _constrain to fix build on ppc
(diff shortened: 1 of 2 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Jan  9 12:52:05 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
+
+- Increase disk size in _constrain to fix build on ppc
```
mvidner, line 4:
```suggestion
- Increase disk size in _constraints to fix build on ppc
```

PR #1878, opened 2025-01-09
fix(web): map product issues to registration path

dgdavid:
Ensure proper navigation to the _Registration_ section when user clicks on a registration issues from the drawer by restoring code that was deleted by mistake when merging master into registration feature branch.


ancorgs (approved):
LGTM. Unless a changelog entry is required.

imobachgs (approved):

jreidinger (approved):
thanks for fixing it.

PR #1880, opened 2025-01-10
feat(rust): disable HTTP caching

imobachgs:
## Problem

When connecting to the same machine (e.g. a server) but using a different version of Agama (e.g., you are using a new build), the browser cache could display the old UI.

## Solution

Disable the cache by setting the `no-store` value. In the future, we could prefer using a different mechanism, like the `Etag` header.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-server/Cargo.toml
+++ b/rust/agama-server/Cargo.toml
@@ -24,7 +24,12 @@ macaddr = { version = "1.0", features = ["serde_std"] }
 async-trait = "0.1.83"
 axum = { version = "0.7.7", features = ["ws"] }
 serde_json = "1.0.128"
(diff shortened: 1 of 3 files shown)


ancorgs:
To me, using `etag` (eg. based on the machine-id) sounds like the right approach.

imobachgs:
I do not expect any performance problem to be honest. You only need to load the code once (the HTML and a big JavaScript file).

Reloading the app is the only case but, for Agama 11, I would say it is good enough.

imobachgs:
And it is not enough with setting the etag value only. So for Agama 11 I prefer the "conservative" solution. However, I will do some tests.

ancorgs (approved):
LGTM, despite the weird error on CI, which doesn't look related to this PR.

PR #1881, opened 2025-01-10
fix(web): warn the user when using the new storage settings

imobachgs:
## Problem

Agama's web UI does not support the new storage settings yet. Instead of crashing,
it should offer a meaningful message to the user.

## Solution

Detect when the settings are not available and display an error instead of crashing.

## Testing

- *Tested manually*

## Screenshots

<details>
<summary>Overview page</summary>

Captura desde 2025-01-10 12-17-25
</details>

<details>
<summary>Storage page</summary>

Captura desde 2025-01-10 12-17-31
</details>

### DIFF
--- a/web/src/components/overview/StorageSection.tsx
+++ b/web/src/components/overview/StorageSection.tsx
@@ -118,7 +118,7 @@ export default function StorageSection() {
   if (result === undefined) {
     return (
       <Content>
-        <Text>{_("Install using an advanced method.")}</Text>
+        <Text>{_("Install using an advanced configuration.")}</Text>
       </Content>
     );
   }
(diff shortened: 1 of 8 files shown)


web/src/components/overview/StorageSection.tsx
```diff
@@ -115,7 +115,13 @@ export default function StorageSection() {
   const availableDevices = useAvailableDevices();
   const result = useProposalResult();
 
-  if (result === undefined) return;
+  if (result === undefined) {
+    return (
+      <Content>
+        <Text>{_("Install using an advanced method.")}</Text>
```
ancorgs, line 121: What about using "configuration" instead of "method"? I think that's easier to relate to what really happened.

ancorgs (approved):
Now with the changelog it looks even better.

PR #1882, opened 2025-01-10
refactor(web): do not allow unregistering a product

dgdavid:
The action for revoking a product registration is actually not needed. Let's drop it until there is a real need in order to keep the UI more clean.


ancorgs (approved):
LGTM

Issue #1891, opened 2025-01-13
Product selection stuck after registration


jreidinger:
If you register product and go back to product selection it will stuck. Ideally this should be addresses two issues we have:

1. do not allow to switch product after selection as it cause more troubles and benefit is small
2. Properly report errors when you select product and get error from backend.

dgdavid:
> If you register product and go back to product selection it will stuck. Ideally this should be addresses two issues we have:
> 
> 1. do not allow to switch product after selection as it cause more troubles and benefit is small

This should be fixed with #1892 

> 2. Properly report errors when you select product and get error from backend.

I'd create a different issue for this one

dgdavid:
I'm closing this because the first point was already fixed and the second one [transferred to a new issue](https://github.com/agama-project/agama/issues/1904) that later was [converted to a discussion](https://github.com/agama-project/agama/discussions/1905).

PR #1892, opened 2025-01-13
fix(web): do not allow changing product after registering one

dgdavid:
## Problem

Although it is not possible to change a product after registering one, web interface still rendering the _Change product_ action and does not redirect to the root path when users try to navigates to the product selection.

## Solution

Check if there is a registration code set for both, 

* Do not render the change product action
* Navigates to root path in case the user reach the selection product path.

## Testing

- Added a new unit test
- Manually tested by @jreidinger
---

Related to 

  * https://github.com/agama-project/agama/issues/1891
  * https://github.com/agama-project/agama/pull/1882


jreidinger:
manual test works for me as expected.
after_reg

jreidinger (approved):

PR #1893, opened 2025-01-13
live: Drop patterns-yast-yast2_basis requirement

g7:
## Problem

patterns-yast is not available in SLFO/SLE16.

## Solution

~~Added a requirement for the pattern only when using the openSUSE, openSUSE-PXE, Leap and Leap-PXE profiles.~~

As suggested by Josef below, dropped it entirely on openSUSE as well as the whole pattern shouldn't be installed at all.

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jan 15 16:53:28 UTC 2025 - Eugenio Paolantonio <eugenio.paolantonio@suse.com>
+
+- Drop patterns-yast-yast2_basis requirement
+  yast packages should be installed per package and not as
+  a whole pattern.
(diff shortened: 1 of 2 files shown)


jreidinger:
@g7 Thanks for patch, but as it is just about media content and also yast packages should be installed per package and not as whole pattern, I would remove it also from opensuse medium ( so please update PR ).
And one more question I have is why SLE does not have PXE? I get few question and I have no clue. We should be able to easily add it same as for opensuse one.

g7:
Thanks for the review!

v2:
* As suggested, drop the pattern requirement everywhere, not only in SLES.

>  And one more question I have is why SLE does not have PXE? I get few question and I have no clue. We should be able to easily add it same as for opensuse one.

I've passed this question on, it's something that I would expect too to be honest, but let's see.

Thanks!

jreidinger (approved):

PR #1894, opened 2025-01-14
feat(software): Reload repositories after failure

lslezak:
## Problem

agama-repo-failed

- When repository refresh fails there is no way how to retry the operation
- The only workaround is to select a different product and then select back the original product
- This obviously won't work if the installer contains only one product

## Solution

- Detect that a repository failed and offer a reload action

## Recording

[SLES4SAP-screen0.webm](https://github.com/user-attachments/assets/ad9a5798-5f42-4333-8a4a-9419953a85ee)

- There is a new error section with "retry" link
- Unfortunately the VirtualBox does not record the mouse pointer so it not obvious that the "Try again" link was clicked by mouse
- For easier testing I used a local copy of the SLE repository to easily simulate unreachable repository and make it working for the retry attempt. That's why you see a different (local) repository in the recording.

## Details

- It turned out that it is not simple to detect a repository failure in the software page.
- It reports an issue but it might contain a translated text or there might be a different issue like missing product or pattern. That means we cannot use the issues texts.
- We will very likely need a repository management later anyway it makes sense to add API for fetching the current repository setup with the load status (succeeded or failed).

## Tasks

- [x] DBus interface for listing the current repository setup (added the `ListRepositories` DBus method)
- [x] Add the HTTP API endpoint `/api/software/repositories` using the DBus backend above (@jreidinger thanks!)
- [x] Adapt the frontend code to show a failure
- [x] Adapt the frontend to retry after clicking a button


web/src/components/software/SoftwarePage.tsx
```diff
@@ -111,6 +152,7 @@ function SoftwarePage(): React.ReactNode {
           <GridItem sm={12}>
             <IssuesHint issues={issues} />
           </GridItem>
+          <GridItem sm={12}>{reload}</GridItem>
```
dgdavid, line 155: NP: I would render the GridItem only if needed. I.e., I'd move such a wrap to line 142 above. Or goes for something slightly more readable like 

```js
const showReposAlert = repos.some((r) => !r.loaded);

...

{ showReposAlert && <GridItem sm={12}><ReloadSection/><GridItem>}

**Review by dgdavid — COMMENTED** [review]
Web part looks good to me, just few NP comments.

File: web/src/api/storage/types/config.ts [hunk 2]
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable */
```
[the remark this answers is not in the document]
joseivanlopez: Yes, but I have removed it too in our feature branch. I have to check whether that line can be omitted somehow when generating the file.

service/lib/agama/dbus/software/manager.rb
```diff
@@ -67,6 +67,24 @@ def issues
         private_constant :SOFTWARE_INTERFACE
 
         dbus_interface SOFTWARE_INTERFACE do
+          # array of repository properties: pkg-bindings ID, alias, name, URL, product dir, enabled
+          # and loaded flag
+          dbus_method :ListRepositories, "out Result:a(issssbb)" do
```
joseivanlopez, line 72: In storage we are tending to other kind of interfaces. The storage API returns JSON data instead of D-Bus structs, dicts, etc. At the end, we are using D-Bus as a transport layer, and probably we will replace it in the future by a HTTP API. Having JSON data has 2 advandanges in our case:

* Replacing D-Bus by HTTP will be almost transparent.
* The rust code is only a proxy that defines an endpoint (it does not need to interpret the data coming from D-Bus).

I guess it is fine to continue with this approach using structs in software. Anyway, we want to rewrite this service in rust.

joseivanlopez (approved):
LGTM

imobachgs (approved):
I trust @dgdavid when it comes to the UI. The rest looks good to me.

web/src/components/software/SoftwarePage.tsx
```diff
@@ -121,25 +127,23 @@ function SoftwarePage(): React.ReactNode {
...
-          {/* TRANSLATORS: link for retrying failed repository load */}
-          {_("Try again")}
-        </Button>
+        <>
+          {errorMsg}{" "}
+          <Button variant="link" isInline onClick={startProbing}>
+            {/* TRANSLATORS: link for retrying failed repository load */}
+            {_("Try again")}
+          </Button>
+        </>
```
dgdavid, line 141: Sorry, I forgot to mention in yestereday's review. Please, move this component outside of the component is using it. As far as I can see yo just need to give it an `onReload` prop 

```jsx
<ReloadSection onReload={startProbing} />
```

Reason for moving it outside: https://react.dev/learn/preserving-and-resetting-state.

We have had issues with that, for example https://github.com/agama-project/agama/pull/1550#issuecomment-2329451247

lslezak: Thanks a lot about this, I didn't know it!

imobachgs (approved):

PR #1895, opened 2025-01-14
feat(web): Migration to PatternFly 6, round 1

dgdavid:
## Problem

On October 31th, 2024 PatternFly released it's 6th version, which comes with interesting changes. It's look&feel has changed a lot and it has started to make use of [tokens](https://www.patternfly.org/tokens/about-tokens), which can be an opportunity for _sanitizing_ the Agama CSS overrides and to start a good foundation for allowing themes in the Agama web interface. Especially, having in mind that there is a [SUSE branding](https://brand.suse.com/) refinement too.

## Solution

Start migrating to PatternFly 6.

This first round of changes covers several code adaptations as well as some Agama refinements like the deletion of dead code.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

This first round does not take too much care about the final result, which makes useless taking screenshots.


dgdavid:
Closing this in favor of #1900, which is almost the same but against a branch including the new storage interface.

PR #1901, opened 2025-01-15
fix(web): drop components no longer needed

dgdavid:
The rewrite of the storage interface obsoletes a lot of components in _storage_ and _core_ namespaces. Let's drop them by now and just rescue needed parts, if any, when proceed.

As a side-effect, it reduces the list of skipped unit test from 54 to 16.


imobachgs (approved):
+0/-5.132... awesome! 😎

PR #1911, opened 2025-01-16
fix(web): storage reprobe

joseivanlopez:
## Problem

If the system is set as deprecated (e.g., after activating iSCSI devices), then the web client automatically reprobes the storage devices and recalculates the proposal. That is done by 3 different calls (fetch config, probe, and set config). This makes the process very unreliable if more than one reprobing is done at the same time. As side effect, the current storage config could be lost, getting the initial storage proposal instead.

## Solution

Add a new D-Bus method for reprobing in an atomic way. The system is set as "no deprecated" at the end of the action. Now clients can simultaneously call to reprobe without any risk.

### DIFF
--- a/web/src/components/storage/ConfigEditor.tsx
+++ b/web/src/components/storage/ConfigEditor.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2024] SUSE LLC
+ * Copyright (c) [2024-2025] SUSE LLC
  *
  * All Rights Reserved.
  *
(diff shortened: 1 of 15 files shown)


web/src/components/storage/ConfigEditor.tsx
```diff
@@ -34,6 +34,12 @@ export default function ConfigEditor() {
       {model.drives.map((drive, i) => {
         const device = devices.find((d) => d.name === drive.name);
 
+        /**
+         * @fixme Make DriveEditor to work when the device is not found (e.g., after disabling
+         * a iSCSI device).
+         */
+        if (device === undefined) return null;
```
ancorgs, line 41: Please create a Trello card for this.

PR #1915, opened 2025-01-17
WIP just log onClick the Delete icon

mvidner:
## Problem

Making the trash-can icon work

- https://trello.com/c/vPPaRAIg

deleteme_C3_A7

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots


web/src/queries/storage/config-model.ts
```diff
@@ -252,6 +278,23 @@ export function useBoot(): BootHook {
...
+// driveName, like "/dev/sda"
+// mountPath, like "/" or "swap"
+export function usePartition(driveName: string, mountPath: string): PartitionHook | undefined {
+  const model = useConfigModel();
+  const { mutate } = useConfigModelMutation();
+
+  if (findPartition(model, driveName, mountPath) === undefined) return;
+
+  return {
+    onDelete: () => mutate(deletePartition(model, driveName, mountPath)),
```
dgdavid, line 294: NP: I'd call the mutation just _delete_ and keep the _onDelete_ name for naming the event/callback on components.

web/src/queries/storage/config-model.ts
```diff
@@ -104,6 +116,20 @@ function disableBoot(originalModel: configModel.Config): configModel.Config {
...
+  originalModel: configModel.Config,
+  driveName: string,
+  mountPath: string,
+): configModel.Config {
+  const model = copyModel(originalModel);
+  const drive = findDrive(model, driveName);
+  if (drive === undefined) return;
+
+  const partitions = (drive.partitions || []).filter((p) => p.mountPath !== mountPath);
+  drive.partitions = partitions;
```
dgdavid, line 129: NP: I know the PR still in draft mode, but just in case you overlooked it this looks like a good place for using the `findPartition` method you added above.

web/src/queries/storage/config-model.ts
mvidner: I tried that but had a problem with that, apparently `delete` is a reserved word somewhere. I will retry with quoting :)

web/src/queries/storage/config-model.ts
mvidner: Huh, I don't think I can use that method... the lambdas differ in `===` and `!==`... or did you mean something else?

dgdavid: Actually I didn't realize about such a difference :) Go ahead.

mvidner:
Sure, my bad :)

mvidner:
> I haven't reviewed all the code again, but if you don't mind would be nice if you add a unit test for ensuring the interface behaves as we expect.

BTW I noticed that the ci-web.yml job does have a `npm test -- --coverage --silent` step, but we also used to upload the result to coveralls.io which is now commented out. Perhaps it is time to bring it back?

dgdavid:
> > I haven't reviewed all the code again, but if you don't mind would be nice if you add a unit test for ensuring the interface behaves as we expect.
> 
> BTW I noticed that the ci-web.yml job does have a `npm test -- --coverage --silent` step, but we also used to upload the result to coveralls.io which is now commented out. Perhaps it is time to bring it back?

I think so. In fact, I asked for it back in September 12th at internal #team-yast Slack channel :wink:

dgdavid:
@mvidner I have updated the code and the test. Find some random comments below.

About your doubts written as comments in the test, you were right. There are a lot of missing pieces in the interface concerning a11y. That’s somewhat intentional for the first round of this new proposal. It’s not the ideal way to work, since a11y should be another success criterion for a good MVP, but…

Fortunately, RTL helps a lot in spotting an interface with markup issues due to its user interaction-based testing approach. Please, have a look at the changes I made and let’s discuss them in a meeting if needed. I’ve only done a few basic things to ensure we can at least test it easily, but there’s more work ahead (as follow-up tasks, I mean).

As an aside note, think of these `aria-label`s as the missing labels for these selectors. We’ve used sentences to provide context for sighted people, but we need to think of a good way to provide similar information with the same interaction fluency for screen readers.

Another thing, not related to your PR, is the number of components inside the `DriveEditor` file. They should be exported in order to test them in isolation, for unit testing. Also, we need to review the entire file to reduce component nesting as much as possible, which might result in splitting it into a couple of files to group the pieces that are part of the DriveEditor thingy.

Finally, it would be great if you could rebase the `storage-config-ui` branch before proceeding with the merge, as quite some things have changed. Don’t forget to run `npm install` after doing so.

web/src/queries/storage/config-model.ts
```diff
@@ -252,6 +278,23 @@ export function useBoot(): BootHook {
   };
 }
 
+export type PartitionHook = {
+  delete: () => void;
+};
+
+// driveName, like "/dev/sda"
+// mountPath, like "/" or "swap"
```
ancorgs, line 286: It makes sense to follow here the same syntax (jsdoc?) we are using all along the rest of the code.

ancorgs: So this could be the first properly documented function in this file.

web/src/components/storage/DriveEditor.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const mockDrive: ConfigModel.Drive = {
+  name: "/dev/sda",
+  spacePolicy: "delete",
+  partitions: [
+    {
+      mountPath: "swap",
+      size: {
+        min: 2_000_000_000,
+        default: false, // WTF does default mean??
```
ancorgs, line 60: If the size was omitted and, as a result, that min of 2_000_000_000 was calculated by Agama itself, then "default" is true.

On the other hand, if that value of 2_000_000_000 was provided by the user, "default" is false.

ancorgs (commented):
Generally looks good. I added a comment about documentation and a clarification about a question raised in a comment.

Other than that, I only miss the changelog entry.

web/src/components/storage/DriveEditor.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const mockDrive: ConfigModel.Drive = {
+  name: "/dev/sda",
+  spacePolicy: "delete",
+  partitions: [
+    {
+      mountPath: "swap",
+      size: {
+        min: 2_000_000_000,
+        default: false, // false: user provided, true: calculated
```
mvidner, line 60: @ancorgs thanks for the explanation... 
now I know that the comment should actually go to storage.model.schema.json, but I haven't found a Makefile-like declaration to update the generated config-model.ts so I am postponing this yak shave

ancorgs (approved):
LGTM

PR #1917, opened 2025-01-17
relax versions of rubygems and comment out ostruct with reason

jreidinger:
## Problem

SLFO using ruby3.2 and agama gemspec specifies versions from ruby3.4 which are too high.

## Solution

Relax versions as agama does not have problem to run with ruby3.2.
Tricky part is ostruct as it is not provided by ruby3.2 as gem, but as it will be removed after 3.4 lets just comment it out and solve it when need arise ( SLFO should update soon ruby version ).


teclator (approved):
LGTM

Issue #1918, opened 2025-01-17
Unexpected application error


tonyj-suse:
I suspect an issue with the JSON (which was convered by agama-autoyast) but what exactly is unclear.

![Image](https://github.com/user-attachments/assets/9e7575f1-94ac-44e7-aa9c-3a4a3e467a17)

[agama-logs-1737155520.tar.gz](https://github.com/user-attachments/files/18461628/agama-logs-1737155520.tar.gz)

Also this was a serial (PXE install).     And at this point I'm seeing the "Welcome to Agama installer version 11! (ttyS1)" prompt on the serial console.

I booted with "linemode=1 Textmode=1" which are obviously linuxrc and likely irrelevant (sorry I know nothing about agama internals) but had I not opened the HTML5 console on the machines service-processor I would never have seen the above error.

How do you propose to handle unattended serial console installs that fail?

tonyj-suse:
I pared the json down (which was originally translated from XML by agama) to the following but issue still occurs.

If legacyAutoyastStorage is the issue can someone give me the translation into the new structure?  I read the schema but lack the free cycles to really dig into it.

{
  "localization": {
    "keyboard": "us",
    "languages": [
      "en_US.UTF-8",
      "en_US de_DE.UTF-8"
    ],
    "timezone": "Europe/Berlin"
  },
  "product": {
    "id": "SLES",
    "registrationCode": "[redacted]"
  },
  "root": {
    "password": "[redacted]"
    "hashedPassword": true
  },
  "software": {
    "patterns": [
      "base",
      "base_traditional",
      "fips",
      "hardware",
      "minimal_base",
      "selinux",
      "sssd_ldap"
    ]
  },
  "legacyAutoyastStorage": [
    {
      "disklabel": "gpt",
      "initialize": true,
      "partitions": [
        {
          "filesystem": "vfat",
          "format": true,
          "label": "efi",
          "mount": "/boot/efi",
          "partition_nr": 1,
          "size": "512MB"
        },
        {
          "mount": "swap",
          "partition_nr": 2,
          "size": "auto"
        },
        {
          "filesystem": "btrfs",
          "format": true,
          "label": "SLES",
          "mount": "/",
          "partition_nr": 3,
          "size": "60%"
        },
        {
          "filesystem": "ext4",
          "format": true,
          "label": "ABUILD",
          "mount": "/abuild",
          "partition_nr": 4,
          "size": "max"
        }
      ],
      "use": "all"
    }
  ]
}

[...]

PR #1927, opened 2025-01-21
fix(web): fix TPM check when using the legacy AutoYaST mode

imobachgs:
## Problem

When using the legacy AutoYaST mode, the web UI crashes at the end of the installation when trying
to find our whether it is using TPM or not.

- https://github.com/agama-project/agama/issues/1918

## Solution

Properly handle that case:

* `fetchConfig` now returns `null` instead of `undefined`. According to [TanStack Query
documentation](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery), `queryFn`
*Must return a promise that will either resolve data or throw an error. The data cannot be
undefined*.
* `usingTpm` function now handles the case where the configuration is `null`.

## Testing

- *Tested manually*

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jan 21 09:44:08 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

teclator:
LGTM

PR #1930, opened 2025-01-21
fix(service): fix the return value of ListRepositories

imobachgs:
## Problem

`ListRepositories` D-Bus method just crashes. It is expected to return an array with contains the
return value as its first element. However, when there are no repositories, it just returns an empty
array. It causes Ruby D-Bus to fail because it cannot serialize a `nil` value.

## Solution

Fix the return value. For comparison, you can check the following `ListPatterns` method.

## Testing

- _Tested manually_


ancorgs (approved):
LGTM

PR #1931, opened 2025-01-22
Fixed icewm configuration using setup from yast

mchf:
## Problem

In running Agama installer user was able to open IceWM's main menu using Win key

- [bsc#1235478](https://bugzilla.suse.com/show_bug.cgi?id=1235478)

## Solution

Taken [preferences.yast2](https://github.com/yast/yast-theme/blob/master/theme/SLE/wmconfig/preferences.yast2) from yast-theme package

Note: the Makefile patch was taken from @lslezak https://github.com/agama-project/agama/compare/efi_locale?expand=1#diff-3814de00178ccbebe225da81ee5f56663518b579103d005782925a8b014491ca

## Notes

- The default IceWM shortcut for starting xterm is `Ctrl+Alt+T`
- Additionally fixed crashing xterm when displaying the xterm menu (via Ctrl + click)
  agama-xterm
- I (@lslezak) have enabled some IceWM features in the config file. Enable displaying the window title, the maximize/minimize buttons and the window actions menu in the title bar.

  YaST uses popup windows where these features are not wanted, but Agama displays popups in the main browser window (so technically they are not popups and we do not use any `alert()` or similar functions). And these features might be useful when running xterm debug window. See the screenshot above.

## Testing

- Tested manually


imobachgs (approved):
Please, do not merge yet until beta 1 is released.

ancorgs (approved):
LGTM, now with the added fonts

PR #1932, opened 2025-01-22
feat: implement some libzypp error callbacks

imobachgs:
## Problem

Agama only implements a few libzypp error callbacks, so many times it does not report problems
properly.

## Solution

Implement more libzypp error callbacks, reporting more types of problems.

- [x] DoneProvide
- ~~SourceProbeError~~ - not used
- ~~SourceCreateError~~ - not used
- [x] ScriptProblem
- [x] DonePackage

for not used see https://github.com/yast/yast-pkg-bindings/pull/191

## Testing

- Tested manually, testing the package installation failure was a bit tricky. I used `dd if=/dev/zero of=/mnt/eat-space` to eat all available space. But you need to leave enough space for download and not enough to install the package.

## Future

- AcceptUnknownGpgKey
- AcceptVerificationFailed
- AcceptFileWithoutChecksum
- AcceptWrongDigest
- AcceptUnknownDigest

## Screenshots

agama-package-callbacks


service/lib/agama/software/callbacks/provide.rb
```diff
@@ -0,0 +1,101 @@
...
+        # Register the callbacks
+        def setup
+          Yast::Pkg.CallbackDoneProvide(
+            Yast::FunRef.new(method(:done_provide), "string (integer, string, string)")
+          )
+        end
+
+        # DoneProvide callback
+        #
+        # @return [String] "I" for ignore, "R" for retry and "C" for abort (not implemented)
```
jreidinger, line 58: NP: I think it would be better to use constants like `IGNORE` and `RETRY` in code instead of C or I.

jreidinger (approved):

PR #1935, opened 2025-01-22
feat(web): UI for adding a partition

joseivanlopez:
UI for adding new partitions to a drive config or for reusing an existing partition. Allows defining the mount point, the target (a new partition or use an existing one), the file system type (or reuse the existing file system) and a size range.

The size can be autocalculated, informing to the user about the size range used by the installer.


joseivanlopez:
Screenshot From 2025-02-10 14-42-36

Screenshot From 2025-02-10 14-42-26

ancorgs (approved):
LGTM

PR #1936, opened 2025-01-22
Storage more devices

ancorgs:
## Problem

The result box at the Storage section displays all the devices affected by the installation proposal. But for doing that, it relies on the actions to be done on the disks. That means that a disk is not displayed there if is not going to be modified. In principle, that's fine but...

There is a corner case if a disk is chosen only for booting and no action is needed for that (eg. legacy boot with a disk that already includes the bios_boot partition). It's certainly not going to be affected, but the disk is displayed at the configuration section and omitted at the result. That's confusing.

## Solution

Display at the result all disks that are mentioned at the configuration, even if no actions are needed to prepare the disks.

## Testing

- Tested manually
- Pending to extend the unit tests


web/src/components/storage/DevicesManager.js
```diff
@@ -138,11 +138,15 @@ export default class DevicesManager {
    * Disk devices and LVM volume groups used for the installation.
    * @method
    *
-   * @note The used devices are extracted from the actions.
+   * @note The used devices are extracted from the actions, but the optional argument
```
dgdavid, line 141: Note: DeviceManager is one of a few files that still pending to be migrated to TypeScript. Just saying :wink:

dgdavid (commented):
So far looks good, but as you commented in the PR description tests need to be adapted and extended. Thanks

web/src/components/storage/DevicesManager.js
ancorgs: Look! It's a three-headed monkey!

dgdavid (approved):

PR #1937, opened 2025-01-23
add installation labels to sle products

jreidinger:
## Problem

SLES and SLES4SAP product does not have defined installation labels, but it is needed for full medium.

## Solution

Define them as agreed.

### DIFF
--- a/products.d/sles_160.yaml
+++ b/products.d/sles_160.yaml
@@ -54,6 +54,15 @@ translations:
       bulutta ve uçta iş açısından kritik iş yüklerini çalıştırır.
 software:
   installation_repositories: []
+  installation_labels:
+    - label: SLES160-x86_64
+      archs: x86_64
+    - label: SLES160-arch64
+      archs: aarch64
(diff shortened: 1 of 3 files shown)


products.d/sles_160.yaml
```diff
@@ -54,6 +54,15 @@ translations:
       bulutta ve uçta iş açısından kritik iş yüklerini çalıştırır.
 software:
   installation_repositories: []
+  installation_labels:
+    - label: SLES160-x86_64
+      archs: x86_64
+    - label: SLES160-arch64
```
imobachgs, line 60: I checked the requirements and it looks correct, but that missing `a` is confusing to me.

imobachgs (approved):
LGTM, at least according to the requirements.

products.d/sles_160.yaml
jreidinger: I think it is to be consistent with S4SAP and there it reach limit of 15 chars...good part is that users won't see this

PR #1938, opened 2025-01-23
fix: better registration workflow

imobachgs:
## Problem

The registration workflow does not allow installing without registering the system, even if the repositories are offline or are specified through the `install_url` option. Thus, we need to make some changes to the current implementation.

## Solution

Now Agama will complain if the following conditions are met:

* The product can be registered (`registration: true`).
* The base product is missing from the repositories.

If the product can be registered, but the base product is already in the repositories, Agama will not complain. However, you can register the product if you want to.

## The new approach

Instead of exposing information about whether the system should be registered at the HTTP API level, we have decided to rely on installation issues. If the conditions are not met, Agama will set an issue.

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2021-2024] SUSE LLC
+# Copyright (c) [2021-2025] SUSE LLC
 #
 # All Rights Reserved.
 #
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -600,9 +600,7 @@ module Agama
       def missing_registration?
         return false unless product
 
-        registration.reg_code.nil? &&
-          product.registration &&
-          missing_base_product?
+        product.registration && missing_base_product?
       end
 
       # Whether the base product is missing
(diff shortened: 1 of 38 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -597,8 +598,19 @@ def missing_registration_issue
       #
       # @return [Boolean]
       def missing_registration?
+        return false unless product
+
         registration.reg_code.nil? &&
-          registration.requirement == Agama::Registration::Requirement::MANDATORY
+          product.registration &&
+          missing_base_product?
```
jreidinger, line 605: just to be sure, do we update issue when product is change or when repository is added?

imobachgs: Every time we reevaluate the software layer (which includes registration, repositories probing, etc.). But it might happen what we miss some case, we need to do more testing.

service/lib/agama/software/manager.rb
```diff
@@ -589,6 +589,7 @@ def missing_product_issue
       # @return [Agama::Issue]
       def missing_registration_issue
         Issue.new(_("Product must be registered"),
+          kind:     :missing_registration,
```
jreidinger, line 592: Interesting that we need similar way in questions. At first it was just string and then we need to add "class" to quickly check what kind it is. :+1:

PR #1939, opened 2025-01-23
fix(build): The number of build jobs should depend on RAM size

lslezak:
## Problem

- Package build fails in OBS with OOM killer
- https://build.suse.de/package/live_build_log/Devel:YaST:Agama:Head/agama/SLES-16.0/s390x

## Details

Sometimes the build might fail in OBS because Rust compilation requires huge amount of RAM. The problem happens when running many parallel jobs on a machine with not enough RAM. The build on S390 failed when running 8 jobs with 8GB RAM.

Originally I wanted to increase the requirement for RAM in the `_constraints` file from 8GB to 16GB. But that would decrease the number of available workers significantly, esp. on exotic archs like s390. And in most cases the workers run 4 or 8 jobs so requiring 16GB would be an overkill.

## Solution

So to decrease the needed amount of RAM decrease the maximum number of parallel jobs.

My experimental run `/usr/bin/time -v cargo build -j1` showed maximum used memory 1.1GB on x86_64. The compilation on S390 failed with 1GB per job. To be on the safe side let's require 1.3GB per job, different architectures might require more and we need to also leave something for the system and other services.

That means with 8GB RAM it should run at most 6 parallel jobs. That should hopefully avoid triggering the OOM killer.

## Notes

- We can adjust the RAM per job constant later if needed, this initial value is rather an experimental value, let's see how it will work.
- If some architecture needs quite different amount of RAM we can make the setting arch dependent using the `%ifarch` macro.

### DIFF
--- a/rust/package/agama.spec
+++ b/rust/package/agama.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package agama
 #
-# Copyright (c) 2023-2024 SUSE LLC
+# Copyright (c) 2023-2025 SUSE LLC
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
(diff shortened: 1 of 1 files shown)


imobachgs (commented):
It looks like black magic to me :-) No, seriously, it looks good. I tried the logic on my own machine.

Thanks!

imobachgs (approved):

jreidinger (approved):

PR #1941, opened 2025-01-24
feat(web): show a welcome screen

dgdavid:
## Problem

There is not a way for users for easily changing their preferred language even when they know nothing about English.

https://trello.com/c/FS1zHTuL/4061-3-agama-create-a-welcome-screen

## Solution

To start with a quite simple welcome screen displaying a list of available languages.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2022-2024] SUSE LLC
+ * Copyright (c) [2022-2025] SUSE LLC
  *
  * All Rights Reserved.
  *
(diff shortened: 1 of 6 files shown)


dgdavid (commented):
This still in progress, since there are some drawbacks with the current approach of refreshing the whole interface after requesting the selected language. 

* Welcome screen UI cannot react to the selected language for translating the text in the screen and, more important, the main action. Something Agama can live with for the first iteration, though.
* It leads to see too much loader flash (at least four, IIRC) plus an additional flash of the welcome screen, which has to wait until configuration is applied for _navigating ahead_
* There is also something wrong that makes the installer get stuck in a loader instead of reaching the product selection page.
* From the welcome screen it is (or it shouldn't be) possible to setup the language for the selected product, because there is not a selected product yet. An idea from @imobachgs is to set the language for the product according to the installer settings after selecting the product. However, I think would be nice to be able to send some "default settings" for the product when selecting it, not waiting after the product selection is finished for requesting something else automatically.

Issue #1616, opened 2024-09-17
Agama "flashes" congratulations screen before installation begins this confuses openQA!


lkocman:
Hello team

there seems to be a glitch where congratulations screen flashes (like for a second or so) before the installation progress bar is displayed. Which means right after clicking the install button.

This of course confuses openQA if we're waiting for congratulations screen which signalizes end of installation :-)

I can workaround it to ensure that we match installation progress first and then match congratulations.
So definitely low prio, but people might notice this glitch.

(see the timestamp and attached video)
![image](https://github.com/user-attachments/assets/205ac895-1070-4a7c-90e9-1e1c388bcc1f) 
[congratulations.webm](https://github.com/user-attachments/assets/ae81f5f0-e419-4749-a26d-d4bb6723da9c)

dgdavid:
Ouch! Great catch! Thanks for reporting it :smiley: 

>  but people might notice this glitch.

Sure, something to be fixed sooner than later.


JFTR, I suspect the problem is mix between 

https://github.com/openSUSE/agama/blob/bc057f474f9783717f355af970440e7215b2b705/web/src/App.jsx#L60-L62

and

https://github.com/openSUSE/agama/blob/bc057f474f9783717f355af970440e7215b2b705/web/src/components/core/Installation.jsx#L26

For some reason, the Installation.jsx component is reached when is not busy first :man_shrugging: 

First solution that comes to my mind is to drop the Installation.jsx component which is acting as a "router" and move its logic to the App.jsx comonponent which, guess what, is a "router" too. We are aware that App.jsx deserves improvements too, but maybe such a movement could avoid that ugly and unexpected glitch.

dgdavid:
> to drop the Installation.jsx component which is acting as a "router

Which was initially a less dummy component (see https://github.com/openSUSE/agama/commit/11cca006d63ebe535c27a7ae71a6322a4c7040e4#diff-5e04ed7cc51a6f8b3424160cdfeb3aa05b11a5143431092843446a676c914211) but it has become in a dispensable component.

_Funny_ thing is that it was introduced for [*reducing blinking*](https://github.com/openSUSE/agama/pull/420) :laughing: but that was so long ago with a completely different frontend architecture.

lkocman:
Thank you @dgdavid  for looking into it! We have a workaround in place so no rush.

dgdavid:
@lkocman, 

[...]

PR #1942, opened 2025-01-24
fix: prevent reaching the congratulations screen during installation

imobachgs:
## Problem

Agama determines in which point of the installation it is by the combination of two values: the
installation phase (startup, config and install) and the isBusy flag. This method has proven to be
rather fragile.

One of the typical issues is jumping into the "congratulations" screen
when the installation is not even finished. See #1616.

Although we mitigated this problem, it recently appear in our QA testing.

## Solution

Let's introduce a new phase, "finish", which indicates that the installation is finished. As simple
as that. So we do not need a combination of two values anymore at that point. We plan to extend this
approach to other parts.

## Testing

- _Added a new unit test_
- _Tested manually_

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -144,6 +144,7 @@ module Agama
       logger.error "Installation error: #{e.inspect}. Backtrace: #{e.backtrace}"
     ensure
       service_status.idle
+      installation_phase.finish
       finish_progress
     end
     # rubocop:enable Metrics/AbcSize
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -55,11 +55,11 @@ function App() {
   const Content = () => {
     if (error) return <ServerError />;
 
-    if (phase === InstallationPhase.Install && isBusy) {
+    if (phase === InstallationPhase.Install) {
       return <Navigate to={ROOT.installationProgress} />;
     }
 
-    if (phase === InstallationPhase.Install && !isBusy) {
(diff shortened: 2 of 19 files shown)


web/src/App.tsx
```diff
@@ -55,11 +55,11 @@ function App() {
   const Content = () => {
     if (error) return <ServerError />;
 
-    if (phase === InstallationPhase.Install && isBusy) {
+    if (phase === InstallationPhase.Install) {
       return <Navigate to={ROOT.installationProgress} />;
     }
 
-    if (phase === InstallationPhase.Install && !isBusy) {
+    if (phase === InstallationPhase.Finish) {
```
dgdavid, line 62: Nice!

service/lib/agama/manager.rb
```diff
@@ -144,6 +144,7 @@ def install_phase
       logger.error "Installation error: #{e.inspect}. Backtrace: #{e.backtrace}"
     ensure
       service_status.idle
+      installation_phase.finish
```
jreidinger, line 147: as it is in ensure it means that even if installation failed, it will be in finish phase. Is it correct?

imobachgs: Yes. It does not change the behavior at all.

imobachgs: It would be great to have some kind of error summary if something goes wrong.

jreidinger (approved):

PR #1946, opened 2025-01-24
add test that generated json from autoyast xml is valid

jreidinger:
## Problem

autoinst.xml profile conversion can lead to non valid profile.json

https://github.com/agama-project/agama/issues/1960
https://github.com/agama-project/agama/issues/1907
https://github.com/agama-project/agama/issues/1903

## Solution

Add ci test to validate cloned autoyast profile. Fix issue with localization. Fix other autoyast conversion related issues.

Note: sadly I need to add python3-jsonschema to install as rubygem-json-schema does not support anything newer then draft5 json schema and we use newer.

## Testing

- *Added a new unit test*

## Output

before pretty generate

```json
[...]

PR #1951, opened 2025-01-27
fix(service): better handling of script names

imobachgs:
## Problem

There are a few problems when importing an AutoYaST script:

- Agama uses `file_name` instead of `filename`.
- If `file_name` is missing, it does not propose another name, generating an invalid profile.

## Solution

Fix both problems by:

- Using the right attribute name (`filename` instead of `file_name`).
- Falling back to the basename from the location. If no location is given, propose `script-X`
  where `X` is a number.

## Testing

- _Added a new unit test_
- _Tested manually_

### DIFF
--- a/service/lib/agama/autoyast/scripts_reader.rb
+++ b/service/lib/agama/autoyast/scripts_reader.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2024] SUSE LLC
+# Copyright (c) [2024-2025] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -20,6 +20,7 @@
 # find current contact information at www.suse.com.
 
 require "yast"
+Yast.import "URL"
 
 # :nodoc:
 module Agama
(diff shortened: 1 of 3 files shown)


dgdavid (approved):

jreidinger:
@imobachgs we really mess up with our amount of delayed PR as exactly this fixes I have in my PR https://github.com/agama-project/agama/pull/1946

jreidinger:
to be precise I solve it slightly differently, but I have those fixes and also test for it :)

imobachgs:
> to be precise I solve it slightly differently, but I have those fixes and also test for it :)

Yes, I saw it later. Well, the main difference is that in this case I am trying to infer the name of the script from the URL if it is not given. But feel free to close it if you prefer the other way.

The only think I do not like about your PR is the "anonymous-" prefix. I would rather prefer something like "script-" or similar.

imobachgs:
Well, to reduce the duplication, I am closing this PR. I will review yours.

jreidinger:
thanks, change of fallback name is really trivial in mine...lets discuss it there.

PR #1952, opened 2025-01-28
fix(rust): use "dashed" form for systemd-firstboot keymap

imobachgs:
## Problem

When writing the keymap to be used by the target system, Agama does not uses the "dashed" form:
"es(ast)" instead of es-ast.

- [bsc#1236174](https://bugzilla.suse.com/show_bug.cgi?id=1236174)

## Solution

Use the "dashed" form in that case so systemd-firstboot is able to handle the keyboard.


dgdavid (approved):
I miss a unit test, but not sure how you're doing that in Rust.

Code LGTM anyway.

imobachgs:
In this case what I really miss is an integration test to be sure that everything is set up as expected after the first boot.

PR #1956, opened 2025-01-28
move sles and sles4sap to own packages

jreidinger:
***Do not merge yet***

## Problem

If Offline medium should be one per base product it is not possible now.

## Solution

Split sle products into own packages and for backward compatibility keep previous name of both.

## Testing

- *testing IBS build with new sources at https://build.suse.de/package/show/home:jreidinger:branches:Devel:YaST:Agama:Head/agama-products


jreidinger:
not needed in the end

PR #1957, opened 2025-01-28
Add support for translating ifcfg kernel cmdline argument to ip

teclator:
## Problem

As there is no linuxrc in Agama we want to explore the possibility of translating linuxrc network-related boot arguments into the dracut equivalent (which is the supported and recommended method for the new installation media).

- https://jira.suse.com/browse/AGM-38
- https://trello.com/c/iCICenNT

## Solution

A basic translation of the **ifcfg** boot argument has being implemented as a prof of concept but as the nm-initrd-generator parses de cmdline as a cmdline hook it is too early for doing some kind of checks about the devices present in the system and trying to match against the device name and the mac address therefore any kind of pattern except the **'*'** wildcard.

**Supported examples**

- ifcfg=*=dhcp
  ip=dhcp

- ifcfg=eth0=dhcp
  ip=eth0:dhcp

- ifcfg=eth0.10=192.168.0.100/24,192.168.0.1
  vlan=eth0.10:eth0 ip=192.168.0.100::192.168.0.1:24::eth0.10

- ifcfg="eth0=192.168.0.33/24 10.0.0.100/24,192.168.0.1,192.168.0.1 10.0.0.1,suse.de"
  ip=192.168.0.33::192.168.0.1:24::eth0 nameserver=192.168.0.1 nameserver=10.0.0.1 ip=10.0.0.100:::24::eth0

The parser is also added as a cmdline hook, we explored the possibility to call it as a **initqueue/settle** hook, in that case we could try to call the **nm_generate_connections** function directly passing it the **getcmdline** result. At that point we can check the devices present in **/sys/class/net/** doing some filtering based in the name or mac address but for example the try function requires to run the configuration in order to check if the interface was configured or not for continuing trying... for that kind of support we might need to add it to NetworkManager and the nm-initrd-generator.

## Testing

- *Tested manually*

### DIFF
--- a/live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
+++ b/live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
@@ -1,0 +1,178 @@
+#!/bin/sh
+
+[ -e /dracut-state.sh ] && . /dracut-state.sh
+
+. /lib/dracut-lib.sh
+. /lib/net-lib.sh
+
+ifcfg_to_ip() {
+  local ip
+  local v="${2}",
+  local interface="$1"
+  local mac="$3"
+  local conf_path="/etc/cmdline.d/40-agama-network.conf"
+  set --
+  while [ -n "$v" ]; do
+    set -- "$@" "${v%%,*}"
+    v=${v#*,}
+  done
+
+  if [[ $# -eq 0 ]]; then
+    echo "IFCFG 0 options given, must be wrong"
+    return 1
+  fi
+
+  ### See https://en.opensuse.org/SDB:Linuxrc#Network_Config
+  # ifcfg=<interface_spec>=[try,]dhcp*,[rfc2132,]OPTION1=value1,OPTION2=value2...
+  if str_starts "$1" "dhcp"; then
+    autoconf="$1"
+    if [ "$autoconf" = "dhcp4" ]; then
+      echo "AUTOCONF"
+      autoconf="dhcp"
+    fi
+    case $autoconf in
+    "dhcp" | "dhcp6")
+      if [ "$interface" = "*" ]; then
+        echo "ip=${1}" >>"/etc/cmdline.d/40-agama-network.conf"
+      else
+        echo "ip=${interface}:${1}" >>$conf_path
+      fi
+      ;;
+    *)
+      echo "No supported option ${1}"
+      ;;
+    esac
+
+    return 0
+  fi
+
+  # ifcifg=<interface_spec>=ip,gateway,nameserver,domain
+  if strglob "$1" "*.*.*.*/*"; then
+    [[ -n "$2" ]] && gateway=$2
+    [[ -n "$3" ]] && nameserver=$3
+    [[ -n "$4" ]] && domain=$4
+
+    ip="$1 "
+    set --
+    while [ -n "$ip" ]; do
+      set -- "$@" "${ip%% *}"
+      ip="${ip#* }"
+    done
+
+    ## TODO: IP is a LIST_IP
+    ip="$1"
+    mask=${ip##*/}
+    ip=${ip%%/*}
+    shift
+
+    ## Configure the first interface, the gateway must belong to the same network
+    echo "ip=${ip}::${gateway}:$mask::${interface}" >>$conf_path
+
+    ## Configure multiple addresses for the same interface
+    while [[ $# -gt 0 ]]; do
+      ip="$1"
+      mask=${ip##*/}
+      ip=${ip%%/*}
+      echo "ip=${ip}:::$mask::${interface}" >>$conf_path
+      shift
+    done
+
+    ## Configure nameservers
+    if [[ -n $nameserver ]]; then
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
I'm approving it because reading the code it makes sense, but be aware I do not know too much bash scripting.

Also, I trust in the fact that it has been manually tested. Which make me thing that in a no far future we could evaluate the introduction of a shell testing tool like https://github.com/shellspec/shellspec or similar. Just saying :roll_eyes:

dgdavid (commented):
Question: Does it deserves an entry in some changelog file?

teclator:
> Question: Does it deserves an entry in some changelog file?

Probably yes, I will add it, we probably should add an unique entry as this PR is against Jozef branch.

dgdavid:
> > Question: Does it deserves an entry in some changelog file?
> 
> Probably yes, I will add it, we probably should add an unique entry as this PR is against Jozef branch.

Ah, right. I overlooked that small detail :) Thanks!

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
```diff
@@ -0,0 +1,128 @@
...
+  local interface="$1"
+  local conf_path="/etc/cmdline.d/40-agama-network.conf"
+  set --
+  while [ -n "$v" ]; do
+    set -- "$@" "${v%%,*}"
+    v=${v#*,}
+  done
+
+  if [[ $# -eq 0 ]]; then
+    echo "IFCFG 0 options given, must be wrong"
```
jreidinger, line 20: the text looks a bit strange..I would write something like "no parameters given to ifcfg option. Skipping" ( or what action happen after return 1

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
```diff
@@ -0,0 +1,128 @@
...
+    return 1
+  fi
+
+  ### See https://en.opensuse.org/SDB:Linuxrc#Network_Config
+  # ifcfg=<interface_spec>=[try,]dhcp*,[rfc2132,]OPTION1=value1,OPTION2=value2...
+  if str_starts "$1" "dhcp"; then
+    autoconf="$1"
+    if [ "$autoconf" = "dhcp4" ]; then
+      echo "AUTOCONF"
```
jreidinger, line 29: this looks like debug output, not?

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
```diff
@@ -0,0 +1,128 @@
...
+      autoconf="dhcp"
+    fi
+    case $autoconf in
+    "dhcp" | "dhcp6")
+      if [ "$interface" = "*" ]; then
+        echo "ip=${1}" >>"/etc/cmdline.d/40-agama-network.conf"
```
jreidinger, line 35: why not $conf_path here? it should be same, not?

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
```diff
@@ -0,0 +1,128 @@
...
+      else
+        echo "ip=${interface}:${1}" >>$conf_path
+      fi
+      ;;
+    *)
+      echo "No supported option ${1}"
```
jreidinger, line 41: I would write "Unsupported option ${1}"

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
```diff
@@ -0,0 +1,128 @@
...
+    done
+
+    ## Configure nameservers
+    if [[ -n $nameserver ]]; then
+      nameserver="$nameserver "
+      while [ -n "$nameserver" ]; do
+        echo "nameserver=${nameserver%% *}" >>$conf_path
+        nameserver="${nameserver#* }"
+      done
+    fi
```
jreidinger, line 85: I have to say that without automatic testing I would be very scared to touch this file to avoid breaking anything.

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
teclator: yep, look like a leftover

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
teclator: was adde mainly for debugging we could even skip it.. or modify as suggested

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-network-compat.sh
teclator: should be there ... probably missing when modified...

jreidinger (approved):

PR #1959, opened 2025-01-29
Install dir

jreidinger:
## Problem

For full offline medium support we need to check also if there is repositories in `/install` on medium.

## Solution

Add support to detect it and use it if found.
Note: live `/` is not the one from iso, so it needs to check mounted initramfs

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -470,6 +470,27 @@ module Agama
       end
 
       def add_base_repos
+        return if add_repos_by_label
+        return if add_repos_by_dir
+
+        # local repositories not found, use the online repositories
+        product.repositories.each { |url| repositories.add(url) }
+      end
+
+      def add_repos_by_dir
+        # path to cdrom which can contain installation repositories
+        dir_path = "/run/initramfs/live/install"
+
+        if File.exist?(dir_path)
+          logger.info "/install found on source cd"
+          repositories.add("dir://" + dir_path)
+          return true
+        end
+
+        false
+      end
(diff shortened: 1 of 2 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -470,6 +470,27 @@ def import_gpg_keys
...
+
+        # local repositories not found, use the online repositories
+        product.repositories.each { |url| repositories.add(url) }
+      end
+
+      def add_repos_by_dir
+        # path to cdrom which can contain installation repositories
+        dir_path = "/run/initramfs/live/install"
+
+        if File.exist?(dir_path)
```
lslezak, line 484: Just nit-picking, I'd reverse the logic:

```ruby
return false unless File.exist?(dir_path)

logger.info "/install found on installation medium"
repositories.add("dir://" + dir_path)
true
```

This avoids using an indented block.

lslezak (approved):
LGTM

service/lib/agama/software/manager.rb
kobliha: Yes, searching for a dir should be faster.

jreidinger: @kobliha I think you mean change of order of checking for dir versus label?

lslezak (approved):

service/lib/agama/software/manager.rb
jreidinger: I do not think it will make difference and lets merge this one for now to have something working

Issue #1962, opened 2025-01-31
Software checkboxes do not react after deactivate network link


jreidinger:
Way how to reproduce it:
go to software section and change patterns. Now deactive "active" link in VM and as result it clicks on checkboxes are ignored. Search field still works.

dgdavid:
Unless I'm missing something, I'd say this an expected behavior, since these checkboxes are actually performing a `PUT` request against the `/api/software/config`. It's the only _selection screen_ where Agama behaves in that way by the nature of the software selection design: every time user selects a pattern it has to calculate the dependencies and reflect the _auto selected_ ones.

Something I'd like to change, but don't know either, how nor when. I believe it'd require changes at the bakend.

What we could do meanwhile is to disable the selection when Agama detects there is not network connection, and show an alert to hint the user Agama has lost connection. Indeed, would be nice to always alert user when connection has lost, unless you're using an offline medium.

jreidinger:
Forgot to mention that it is on full offline medium with repos in /install. I also do not see any PUT request in network tab in dev tools...it looks like all network traffic is stopped

dgdavid:
Weird.

Not sure if related, but @imobachgs adjusted the TanStack Query network mode in https://github.com/agama-project/agama/pull/1909

imobachgs:
In the client options, there is an additional key `mutations` that works like the `queries` one. So according to a quick test, I would say that something like this in `app.tsx` would fix it:

```typescript
const queryClient = new QueryClient({
  defaultOptions: {
    mutations: {
      networkMode: networkMode(),
    },
    queries: {
      networkMode: networkMode(),
    },
  },
});
```

@dgdavid thanks for pointing to the network mode stuff 😄

dgdavid:
Thanks to you @imobachgs, much appreciated. 

I was debugging the problem and I could see that selected software pattern is not fulfilling below condition

https://github.com/agama-project/agama/blob/54897640480ca1dc455a81c26a21341df75497d5/web/src/components/software/SoftwarePatternsSelection.tsx#L116

But I do not have clear the big picture of the patterns selections right now. Will continue investigating it later.

Unless the issue has to be with the network mode stuff.

PS: sure, I though that your fix at #1909 improves the situation for the described context, but I wanted to see if there was something we were missing related to the recently introduced "always" network mode.

imobachgs:
The fix has been merged.

PR #1963, opened 2025-01-31
fix(web): set mutations network mode to "always"

imobachgs:
## Problem

The web UI does not perform any mutation when the network gets disconnected.

## Solution

Set the network mode to `always` for mutations, as they have a separate set of default options from queries.

## Testing

- Tested manually


imobachgs:
It should fix #1963.

jreidinger (approved):

dgdavid:
> It should fix #1963.

You mean #1962, right?

imobachgs:
> > It should fix #1963.
> 
> You mean #1962, right?

Ouch, yes, you are right.

PR #1964, opened 2025-01-31
Fixed GitHub pages build and deployment

lslezak:
## Problem

- The GitHub pages deployment fails for about a month: https://github.com/agama-project/agama/deployments

## Details

There are several problems:

- The old artifact actions are deprecated and not supported anymore : https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
- The `gdbus-codegen` is not available, probably because of newer Ubuntu version or the they changed the default image content

## Testing

- Tested manually in my fork, it builds and deploys fine to https://lslezak.github.io/agama/

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install DocBook tooling
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


imobachgs (approved):
I am not sure whether we are using those pages anyway. However, let's fix the building process, and decide later what to do with them.

Thanks!

PR #1969, opened 2025-02-03
live: fix_bootconfig.s390x: use initrd.ofs for the initrd offset filename

g7:
## Problem

mksusecd expects the offset to be written in initrd.ofs.

## Solution

Change the filename from `initrd.off` to `initrd.ofs` to match what mksuse expects.

This fixes building of "combined" Full offline images on s390x.

This pull request also adds a commit that ensures that the arch-specific `fix_bootconfig` scripts exit at the first failure.
`xxd` is not available in the build environment (in SLFO:Main at least), so `initrd.off` (now `.ofs`) and `initrd.siz` were never being created in the first place. Exiting as soon as the first failure will hopefully help us catch this kind of bugs earlier :smile: 

**NOTE:** due to this, s390x builds might fail if xxd is not present. This OBS prjconf snippet should ensure that it is getting installed on image builds:

```
Substitute: kiwi-image:iso    kiwi-systemdeps-iso-media qemu-tools xxd
```

## Testing

- *Tested manually*

### DIFF
--- a/live/config-cdroot/fix_bootconfig.aarch64
+++ b/live/config-cdroot/fix_bootconfig.aarch64
@@ -5,6 +5,8 @@
 # https://github.com/openSUSE/agama/issues/1609
 # KIWI config
 
+set -e
+
 dst="$1"
 
 if [ ! -d "$dst" ] ; then
--- a/live/config-cdroot/fix_bootconfig.ppc64le
+++ b/live/config-cdroot/fix_bootconfig.ppc64le
@@ -37,6 +37,8 @@
 #     └── bootinfo.txt
 #
 
+set -e
+
 dst="$1"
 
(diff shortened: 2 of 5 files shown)


imobachgs (approved):

wfeldt:
Why do you rename that file? It has been `initrd.off` on all SUSE media for ages. This will force tooling scripts
to adapt and probably also some s390x documentation.

wfeldt:
Also, the problem is not

> mksusecd expects the offset to be written in initrd.ofs.

but mksusecd expects all files referenced in suse.ins to exist.

g7:
oh! Misunderstood the error message (since it is coming from suse.ins, now it makes sense to me). I will prepare a follow up PR that reverts this change

wfeldt:
Thanks!

PR #1970, opened 2025-02-04
Reboot

teclator:
## Problem

After an unattended installation the user needs to reboot manually which is not very unattended...

- https://trello.com/c/wXYDjhyj

## Solution

- Agama reboot command added. Only does something if Agama is in a final state
- Unattended installation reboots when installation is finished if everything went right

## Testing

- *Tested manually (PENDING)*


rust/agama-cli/src/commands.rs
```diff
@@ -105,4 +105,9 @@ pub enum Commands {
         /// URL pointing to file for download
         url: String,
     },
+    /// Reboot the system.
+    ///
+    /// This command reboots the system which is mainly done after the installation is completed
+    /// successfuly
```
imobachgs, line 111:
```suggestion
    /// successfully
```

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Feb  5 13:14:08 UTC 2025 - Knut Anderssen <kanderssen@suse.com>
+
+- (gh#agama-project/agama#1930).
+  - In an unattended installation the system will try to reboot
+    the system automatically.
+  - The finish installation will not reboot the system if it is not
```
imobachgs, line 7: *- The D-Bus method to finish the installation...*

imobachgs (changes requested):
In general, it looks good. Just some minor things.

rust/agama-cli/src/lib.rs
```diff
@@ -129,6 +130,26 @@ async fn install(manager: &ManagerClient<'_>, max_attempts: u8) -> anyhow::Resul
...
+///
+/// * `manager`: the manager client.
+async fn reboot(manager: &ManagerClient<'_>) -> anyhow::Result<()> {
+    if manager.is_busy().await {
+        println!("Agama's manager is busy. Waiting until it is ready...");
+    }
+
+    // Make sure that the manager is ready
+    manager.wait().await?;
+    let finished = manager.finish().await?;
```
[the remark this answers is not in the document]
teclator: yep, that call is a leftover

live/root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
```diff
@@ -6,6 +6,6 @@
 
 if [ -e /etc/cmdline.d/99-agama-cmdline.conf ]; then
   echo "Creating agama conf"
-  mkdir -p "$NEWROOT/etc/agama.d"
-  cp /etc/cmdline.d/99-agama-cmdline.conf "$NEWROOT/etc/agama.d/cmdline.conf"
+  mkdir -p "$NEWROOT/run/agama/cmdline.d"
+  cp /etc/cmdline.d/99-agama-cmdline.conf "$NEWROOT/run/agama/cmdline.d/agama"
```
imobachgs, line 10: np: I do not know if it should be `agama.conf`.

rust/agama-cli/src/commands.rs
```diff
@@ -105,4 +106,9 @@ pub enum Commands {
         /// URL pointing to file for download
         url: String,
     },
+    /// Finish the installation (reboots the system by default)
```
imobachgs, line 109: "reboots" is in 3rd person, but "finish" is not.

```suggestion
    /// Finish the installation (reboot the system by default)
```

rust/agama-cli/src/commands.rs
```diff
@@ -105,4 +106,9 @@ pub enum Commands {
...
+    ///
+    /// This command finish the installation performing a set of tasks and rebooting the system by
```
imobachgs, line 111:
```suggestion
    /// This command finishes the installation by performing a set of tasks and rebooting the system by
```

imobachgs: Which tasks?

autoinstallation/package/agama-auto.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Feb 13 12:10:11 UTC 2025 - Knut Anderssen <kanderssen@suse.com>
+
+- Adapted agama-auto script to reboot by default once the unattended
+  installation is finished allowing to stop avoid it through the
```
imobachgs, line 5: This long phrase is confusing. I guess you wanted to write "stop" or "avoid," not both.

autoinstallation/package/agama-auto.changes
```diff
@@ -1,3 +1,11 @@
...
+  'agama.finish' kernel cmdline argument (gh#openSUSE/agama#1970).
```
imobachgs, line 6: Please, use `agama-project/agama` instead of the old `openSUSE/agama`.

rust/agama-cli/src/lib.rs
```diff
@@ -211,6 +232,10 @@ pub async fn run_command(cli: Cli) -> Result<(), ServiceError> {
             let manager = build_manager().await?;
             install(&manager, 3).await?
         }
+        Commands::Finish { method } => {
+            let manager = build_manager().await?;
+            finish(&manager, method).await?;
```
imobachgs, line 237: You can do the `method` argument as mandatory and then use the `unwrap_or_default` function.

rust/agama-lib/src/manager.rs
```diff
@@ -85,6 +88,72 @@ impl TryFrom<u32> for InstallationPhase {
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy, utoipa::ToSchema)]
+pub enum FinishMethod {
+    #[serde(rename = "halt")]
+    // Halt the system
+    Halt,
+    #[serde(rename = "reboot")]
+    // Reboots the system
+    Reboot,
+    #[serde(rename = "stop")]
+    Stop,
```
imobachgs, line 101: What is the difference between `stop` and `halt`?

rust/agama-lib/src/manager.rs
```diff
@@ -85,6 +88,72 @@ impl TryFrom<u32> for InstallationPhase {
...
+    Poweroff,
+}
+
+impl Default for FinishMethod {
+    fn default() -> Self {
+        Self::Reboot
+    }
+}
+
+impl fmt::Display for FinishMethod {
```
imobachgs, line 113: We use the [strum](https://crates.io/crates/strum) crate for these conversions. You can check [the](https://github.com/agama-project/agama/blob/ca747cf59f5447fbc0cea0650400c3c626f28a41/rust/agama-lib/src/software/model/packages.rs#L35) [examples](https://github.com/agama-project/agama/blob/ca747cf59f5447fbc0cea0650400c3c626f28a41/rust/agama-lib/src/scripts/model.rs#L38).

rust/agama-lib/src/manager.rs
```diff
@@ -116,9 +185,13 @@ impl<'a> ManagerClient<'a> {
         Ok(self.manager_proxy.commit().await?)
     }
 
-    /// Executes the after installation tasks.
-    pub async fn finish(&self) -> Result<(), ServiceError> {
-        Ok(self.manager_proxy.finish().await?)
+    /// Executes the after installation tasks finishing with the method given or rebooting the
+    /// system by default.
+    pub async fn finish(&self, method: Option<FinishMethod>) -> Result<bool, ServiceError> {
```
imobachgs, line 190: I would avoid using an `Option` here.

rust/agama-server/src/manager/web.rs
```diff
@@ -173,15 +173,21 @@ async fn install_action(State(state): State<ManagerState<'_>>) -> Result<(), Err
...
     )
 )]
-async fn finish_action(State(state): State<ManagerState<'_>>) -> Result<(), Error> {
-    state.manager.finish().await?;
-    Ok(())
+async fn finish_action(
+    State(state): State<ManagerState<'_>>,
+    method: Option<Json<FinishMethod>>,
+) -> Result<Json<bool>, Error> {
+    if let Some(Json(method)) = method {
```
imobachgs, line 186: Just a suggestion:

```rust
let method = method.unwrap_or_default();
state.manager.finish(method).await?; // you might need to adapt the `finish` function.
```

service/lib/agama/manager.rb
```diff
@@ -276,6 +284,22 @@ def iguana?
 
   private
 
+    SHUTDOWN_OPT = { "reboot" => "-r", "halt" => "-H", "poweroff" => "-P" }.freeze
```
imobachgs, line 287: Do we really need `reboot`, `halt` and `poweroff`?

imobachgs (commented):
The approach looks but I would avoid using optional arguments unless required.

live/root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
teclator: I agree, but basically this is mainly copied from #1896

rust/agama-server/src/manager/web.rs
teclator: I was not sure where to set it as mandatory. Thanks ;)

service/lib/agama/manager.rb
teclator: Well, reboot is the default, and all of them are options supported by the shutdown command although them were not requested by now..

**Note**: rd.emergency option also support the same options except stop that is the special one for not doing anything at the end (see https://github.com/dracutdevs/dracut/blob/master/man/dracut.cmdline.7.asc)

> **rd.emergency=[reboot|poweroff|halt]**
specify, what action to execute in case of a critical failure. rd.shell=0 must also be specified.

We also support halt and reboot in AutoYaST (https://documentation.suse.com/sles/15-SP6/single-html/SLES-autoyast/index.html#CreateProfile-General-mode)

rust/agama-lib/src/manager.rs
teclator: Will improve the documentation, stop was requested to do nothing, that is, stop at the end of the installation while halt .... halt the machine at the end of the installation.

autoinstallation/package/agama-auto.changes
teclator: When you rewrite two or three time a sentence and it is clear in your mind but you do not check the resultant one...

rust/agama-server/src/manager/web.rs
imobachgs: As discussed offline, the problem is that `Json` and `match` could be the best option:

```rust
    let method = match method {
        Some(Json(method)) => method,
        _ => FinishMethod::default(),
    };

    Ok(Json(state.manager.finish(method).await?))
```

imobachgs (changes requested):
It looks good. However, there is a conflict in the changes file and the generation of the OpenAPI docs is not working. Please, check.

service/lib/agama/manager.rb
```diff
@@ -251,20 +251,28 @@ def collect_logs(path: nil)
     end
 
     # Whatever has to be done at the end of installation
-    def finish_installation
+    #
+    # @param method [String, nil]
```
imobachgs, line 255: It would be good to document the possible values.

service/lib/agama/dbus/manager.rb
```diff
@@ -65,7 +65,7 @@ def initialize(backend, logger)
         dbus_method(:Commit, "") { install_phase }
         dbus_method(:CanInstall, "out result:b") { can_install? }
         dbus_method(:CollectLogs, "out tarball_filesystem_path:s") { collect_logs }
-        dbus_method(:Finish, "") { finish_phase }
+        dbus_method(:Finish, "in method:s, out result:b") { |m| finish_phase(m) }
```
imobachgs, line 68: You should update the [D-Bus documentation](https://github.com/agama-project/agama/tree/master/doc/dbus), too. However, I am fine with you merging this PR and updating the docs later.

Remember: if you update a proxy, you should update the docs too.

PS: I do not think that documentation is that useful.

imobachgs (approved):
Just some small documentation issues. Nothing critical.

Thanks!

autoinstallation/bin/agama-auto
```diff
@@ -4,23 +4,30 @@ set -ex
...
 
 case "$url" in
-* )
-    /usr/bin/agama profile import "$url"
-    /usr/bin/agama install;;
+*)
+	/usr/bin/agama profile import "$url"
+	/usr/bin/agama install
+	case "$method" in
+	"stop") ;;
```
jreidinger, line 27: maybe you can log it with simple echo call.

jreidinger (approved):

autoinstallation/bin/agama-auto
teclator: agama finish stop is already supported. so adapted it

imobachgs (approved):

PR #1972, opened 2025-02-04
feat(storage): show proper layout accoding to the scenario

joseivanlopez:
TODO


web/src/components/storage/ProposalPage.tsx
```diff
@@ -21,24 +21,176 @@
...
+              "The current storage config cannot be edited. Do you want to reset to the default config?",
+            )}
+            <Button variant="plain" isInline onClick={reset}>
+              {_("Reset")}
+            </Button>
+          </>
+        </Alert>
+      )}
+      {isEditable && (
+        <GridItem sm={12} xl={8}>
```
ancorgs, line 162: This `xl={8}` makes no sense now that the encryption item is not longer there.

service/lib/agama/storage/proposal.rb
```diff
@@ -93,10 +94,16 @@ def storage_json
...
+      # are supported by the model.
+      #
+      # @return [Hash, nil] nil if the config model cannot be generated.
       def model_json
         config = config(solved: true)
-        return unless config
+        return unless config && model_supported?(config)
+
+        issues = Agama::Storage::ConfigChecker.new(config, product_config).issues
+        return if issues.any?(&:error?)
```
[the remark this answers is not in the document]
ancorgs: Just for completeness, this are all the possible strings returned by `ConfigChecker`.

- There is no boot device alias (_fun enough, this one is returned when there is no root partition_)
- There is no boot device with alias '%s'
- No passphrase provided (required for using the method '%{crypt_method}').
- Encryption method '%{crypt_method}' is not available in this system.
- '%{crypt_method}' is not a suitable method to encrypt the device.
- Missing file system type for '%s'
- The file system type '%{filesystem}' is not suitable for '%{path}'
- There is no LVM thin pool volume with alias '%s'
- '%{method}' is not a suitable method to encrypt the physical volumes.
- Mandatory device %s not found
- Mandatory %s not found
- There is no LVM physical volume with alias '%s'
- There is no target device for LVM physical volumes with alias '%s'
- The device '%s' is used several times as target device for physical volumes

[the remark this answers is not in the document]
ancorgs: Regarding "correct and supported by the UI are different things" I think we should distinguish the following cases:

- A There is a valid proposal:
  - A.1 The configuration is really not editable by the UI -> Use the proposed "Unknown storage config" alert (although we need to reword it a lot).
  - A.2 The configuration is editable by the UI -> no alerts
- B There is no valid proposal:
  - B.1 The configuration is editable by the UI...
    - B.1.a and contains no `ConfigChecker` errors. ->  Use the current ProposalFailedInfo.
    - B.1.b but contains `ConfigChecker` errors. -> Use an alert displaying the `ConfigChecker` errors with an introductory sentence[*].
  - B.2 The configuration is not editable by the UI -> Use the current EmptyState (although we need to reword it a lot).

I don't know if we currently can distinguish all those cases from the existing API, but I think it is important to make that difference.

[*] We need to improve the current error for missing root. That "there is no boot device alias" makes no sense.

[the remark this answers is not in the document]
joseivanlopez: About the boot error message:

The error "There is no boot device alias" means that the boot config does not point to any device (it has not a device alias). This could happen if the boot device was not explicitly set and the solver did not find any suitable device for booting. Right now, the solver only looks for the root device in order to automatically assign a boot device. So, with the current implementation, "There is no boot device alias" can only happen if the boot device is implicit and there is not a root device (i.e., root does not exist). Perhaps the solver could try more things like assigning the first device if root is not found. I think we can adapt the error message if we keep the currrent behavior and the checker always assumes that there is no boot device alias because there is no root.

About mandatory volumes

I have realized there are not checks for the presence of the mandatory volumes. We should generate an error if a mandatory volume is missing. Doing that we also relax the problem with the "boot device alias" error message. On the other hand, I see this kind of checks as something related to the product instead of the config. A config could be valid but it does not satisfy all the product requirements.

About model and config with errors:

I think that working with a model generated from an incorrect config makes everything more complex in the UI. I would prefer to limit the UI in order to avoid generating an invalid config.

What to improve in this PR

Based on your comments here and in trello, what about avoiding to remove the root partition (or any mandatory partition)? Doing that, the model will always generate a valid config. So, model will not be generated only in these cases: 

* An incorrect config is manually loaded.
* A config becomes incorrect because any device disappears from the system.

But I am not sure how to show errors if there is no model generated. Let's discuss it.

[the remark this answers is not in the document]
ancorgs: > About the boot error message:
> 
> The error "There is no boot device alias" means that the boot config does not point to any device (it has not a device alias). [...]

I understand how we get to that error. That does not change the fact that the message is close to useless. :-)

> About mandatory volumes
> 
> [...] On the other hand, I see this kind of checks as something related to the product instead of the config. A config could be valid but it does not satisfy all the product requirements.

So you suggest to have two levels of checks? One that would prevent the proposal to be even tried (the current `ConfigChecker`) and another one that would execute the proposal but would then generate issues that simply prevent the installation?

Not sure whether it's worth it, let's discuss it.

In any case, not having a "/" is an error we should likely control in the first level. It's not only that it should be declared as mandatory for all products. It may have more implications regarding the algorithm (like, at least, the already commented problem detecting the boot).

> About model and config with errors:
> 
> I think that working with a model generated from an incorrect config makes everything more complex in the UI. I would prefer to limit the UI in order to avoid generating an invalid config.

That was my original prototype - not allowing to delete "/" and not allowing to delete the disk containing "/". The latter stays as is, but for the former we decided to allow it in order to allow to re-create "/" at a different location, since we don't have a "move" functionality.

So both have pros and cons. Preventing to create an invalid configuration makes some things easier but may complicate others (eg. moving only "/" to a different disk). On the other hand, making the UI able to handle those cases makes it more useful for those who start with a profile-based configuration (that may not be perfect) and fix it with the UI.

> What to improve in this PR

I'm not sure, let's discuss it.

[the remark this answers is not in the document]
joseivanlopez: We agreed:

* API: export model even if there are config errors.
* API: isses source is config.
* Show config issues in the empty state.
* Do not allow deleting root.
* Improve boot issue message.

service/lib/agama/storage/config_checkers/boot.rb
```diff
@@ -55,7 +55,12 @@ def device_alias
         def missing_alias_issue
           return unless configure? && device_alias.nil?
 
-          error(_("There is no boot device alias"))
+          # Currently this situation only happens because the config solver was not able to find
+          # a device config containing a root volume. The message could become inaccurate if the
+          # solver logic changes.
+          error(
+            _("The boot device cannot be automatically selected because there is not root volume")
```
ancorgs, line 62: I'm not sure whether "volume" means something in this context. If anything, it makes me think on LVM. I would better say "there is no root (/) file system"

web/src/components/storage/ProposalPage.tsx
```diff
@@ -21,24 +21,215 @@
...
+        },
+      ],
+    });
+}
+
+type ErrorsEmptyStateProps = {
+  errors: Issue[];
+};
+
+function ErrorsEmptyState({ errors }: ErrorsEmptyStateProps) {
```
ancorgs, line 79: I think we have gone too far accumulating functionality on this file (just look at the size of the "import" section). Can't we extract the empty pattern component(s) to separate files? Similar to what is done for `ProposalTransactionalInfo` and `ProposalFailedInfo`?

service/lib/agama/storage/config_checkers/boot.rb
joseivanlopez: I have the very same doubt.

web/src/components/storage/ProposalPage.tsx
```diff
@@ -47,63 +238,33 @@ export default function ProposalPage() {
...
+        {!isDeprecated && isDevicesEmpty && <DevicesEmptyState />}
+        {!isDeprecated && !isDevicesEmpty && !isEditable && !isValid && !model && (
+          <ConfigEmptyState />
+        )}
+        {!isDeprecated && !isDevicesEmpty && !isEditable && !isValid && model && (
+          <ErrorsEmptyState errors={configErrors} />
+        )}
+        {!isDeprecated && !isDevicesEmpty && (isEditable || isValid) && (
+          <ProposalSections isEditable={isEditable} isValid={isValid} />
+        )}
```
ancorgs, line 267: I find this code organization a bit weird. The fact that `ConfigEmptyState` is not managed at the same level than `ProposalFailedInfo` is completely counterintuitive for me. Do you get what I mean?

web/src/components/storage/ProposalPage.tsx
joseivanlopez: I don't know if extracting single-usage-components like these, which don't have any kind of logic, is good or not.

IMHO, in cases like this, I prefer to have such components defined directly there in the only file where they are needed. Otherwise, there is a lot chance of forgetting about them when the main component is refactored (that has already happened). Of course, there could be exceptions, like having complex components with too much logic.

But I can extract them if you guys prefer to go with that approach.

web/src/components/storage/DriveEditor.tsx
```diff
@@ -552,6 +552,8 @@ const PartitionsNoContentSelector = ({ drive, toggleAriaLabel }) => {
 
 const PartitionMenuItem = ({ driveName, mountPath, description }) => {
   const { delete: deletePartition } = usePartition(driveName, mountPath);
+  const volume = useVolume(mountPath);
+  const isRequired = volume.outline?.required || false;
```
ancorgs, line 556: I'm not against this change, but I was aiming for a simpler work-around only for root ("/") and not for all required file systems.

Using `required` has an interesting implication - all mandatory volumes need to stay together in the same drive. 

This is just a note for the future because with the current product definitions, required is equivalent to only root in all cases.

joseivanlopez: Good point. So, what is better now? I would keep it as it is, and maybe solve the problem in the future with some kind of "move" action. What do you think?

web/src/components/storage/ProposalPage.tsx
joseivanlopez: Sorry, I did not get you. `ProposalFailedInfo` is only shown when the layout contains the sections (config editor and/or result).

web/src/components/storage/DriveEditor.tsx
ancorgs: Yes. Keep it as it is for now.

web/src/components/storage/ProposalPage.tsx
joseivanlopez: We decided to keep that components directly there, but in the future we will extract that components to separate files. Each component will have the logic to decide whether to render any content. The proposal page will simply mount all those components. Moreover, each components will be tested separated.

dgdavid: I didn't expect that after our conversation 2 or 3 hours ago. Why did you decide to postpone such an agreed movement?

imobachgs: In a few words: "time pressure" 😓

ancorgs (approved):
This LGTM.

I'm not merging because the last commit was added by me (together with all the screenshots at the description). Somebody would need to review those texts. Once the texts are reviewed, we can merge.

web/src/components/storage/ProposalPage.tsx
```diff
@@ -78,9 +82,13 @@ function InvalidConfigEmptyState(): React.ReactNode {
         </List>
       </EmptyStateBody>
       <EmptyStateFooter>
-        <Content component="p">{_("Do you want to reset to the default settings?")}</Content>
+        <Content component="p">
+          {_(
+            "You may want to discard those settings and start from scratch with a simple configuration.",
```
imobachgs, line 87: those or these?

web/src/components/storage/ProposalPage.tsx
```diff
@@ -78,9 +82,13 @@ function InvalidConfigEmptyState(): React.ReactNode {
...
+          )}
+        </Content>
         <Button variant="secondary" onClick={() => reset()}>
-          {_("Reset")}
+          {_("Reset to default configuration")}
```
imobachgs, line 91: "Reset to the default configuration"

web/src/components/storage/ProposalPage.tsx
```diff
@@ -92,17 +100,23 @@ function UnknowConfigEmptyState(): React.ReactNode {
...
       <EmptyStateFooter>
-        <Content component="p">{_("Do you want to reset to the default settings?")}</Content>
+        <Content component="p">
+          {_(
+            "You may want to discard the current settings and start from scratch with a simple configuration.",
+          )}
+        </Content>
         <Button variant="secondary" onClick={() => reset()}>
-          {_("Reset")}
+          {_("Reset to default configuration")}
```
imobachgs, line 119: "Reset to the default configuration"

web/src/components/storage/UnsupportedModelInfo.tsx
```diff
@@ -36,18 +36,23 @@ export default function UnsupportedModelInfo(): React.ReactNode {
...
+          <Content component="p">
+            {_(
+              "You can proceed to install with the current settings or you may want to discard the configuration and start from scratch with a simple one.",
             )}
           </Content>
         </StackItem>
         <StackItem>
           <Button variant="secondary" onClick={() => reset()}>
-            {_("Reset")}
+            {_("Reset to default configuration")}
```
imobachgs, line 55: "Reset to the default configuration"

imobachgs (commented):
Just some suggestions.

web/src/components/storage/ProposalPage.tsx
ancorgs: I would say "those" because the settings are not really displayed. And the reference to them is at a previous paragraph. Both sound like "distance" to me.

imobachgs (approved):

web/src/components/storage/ProposalPage.tsx
joseivanlopez: > Why did you guys decide to postpone something agreed less than 3 hours ago? Just curious because it is, again, a kick to the front-end code.

JFTR: because just after talking with you, the PO decided to go with the current code if it works. As @imobachgs said, time pressure. We agreed a) to add a FIXME, b) follow the agreed approach from now on, c) redo this code (and many other things) once time permits.

> I start having the feeling that does not payoff to invest time in meetings.

I disagree. We agreement is there and we will follow it from now on.

PR #1974, opened 2025-02-05
live: fix_bootconfig.s390x: restore initrd.off naming for the initrd offset

g7:
See discussion in bsc#1236781, and #1969.

This essentially reverts c3c8fc4.

## Problem

`initrd.ofs` was a red herring, and I'm not sure how we got to that, most likely it might have slipped in while manually debugging the media.

So the actual bug was the offset file not being created at all, and now it's sorted out too by adding the `xxd` requirement (and thanks to snwint, latest `mksusecd` can re-generate it as well).

## Solution

Restore the previous `initrd.off` naming which is widely used. The other commit from #1969 is not reverted (that introduces `set -e` in the fix_bootconfig scripts) so builds will still fail if xxd is not present in the build environment. 

## Testing

- *Tested manually*

Apologies for the noise.


imobachgs (approved):
Thank you, @g7!

PR #1975, opened 2025-02-05
Disable the local media in the installed system (bsc#1236813)

lslezak:
## Problem

- When installing SLE16 from the offline medium the initial installation repository is present in the installed system
- However it cannot be reached with URL `dir:///run/initramfs/live/install`, that directory does not exist in installed system
- See https://bugzilla.suse.com/show_bug.cgi?id=1236813

## Solution

- Simply remove the repository, 

## Testing

- Added a new unit test
- Tested manually, when using the offline medium there is no repository defined in the installed system

## Notes

- Some unrelated Rspec test fails, but that happened even before these changes :worried:

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -201,10 +201,16 @@ module Agama
 
       # Writes the repositories information to the installed system
       def finish
+        # remove the dir:///run/initramfs/live/install repository and similar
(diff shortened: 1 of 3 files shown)


ancorgs (commented):
LGTM

mvidner (approved):
LGTM

PR #1976, opened 2025-02-05
feat: report unsupported AutoYaST elements

imobachgs:
## Problem

Agama does not offer information about which elements are supported when importing an AutoYaST profile and which ones are just ignored.

https://trello.com/c/0u2wwMN2

## Solution

Agama checks the profile for unsupported elements. If it finds them, it displays a question listing them and asking the user whether to continue.

## Generating the documentation

The mechanism to detect unsupported elements includes a list of all AutoYaST elements and whether they are supported: "yes", "planned", "partial" or "no". The file also contains hints about replacing some elements (e.g., using a post-script).

Although it was out of scope, this PR introduces a new `doc:autoyast_compat` which generates a reference to be included in the [AutoYaST support](https://agama-project.github.io/docs/user/autoyast) documentation. Not all elements are included, but we can create a follow-up PBI to work on that.

## Disabling the check

It is possible to disable the check specifying `agama.ay_check=0` on kernel's command line. Actually, the check is done, but it does not report found problems.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

unsupported-autoyast-popup

Below is an example of the AutoYaST compatibility documentation automatically generated by the new `doc:autoyast_compat` task.

agama-autoyast-compatibility


mvidner (commented):
Comments based so far on the description only, not on code.

1. Agama checks the profile for unsupported elements.

    When does this happen? Only when I try to use the profile, or can I trigger it beforehand, on a dry run?

2. "Not supported AY elements" popup

    I think it would be nice to also include this text.

    > This popup can be disabled by passing agama.ay_check=0 on the kernel command line

web/src/components/questions/UnsupportedAutoYaST.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const actionCallback = (option: string) => {
+    question.answer = option;
+    answerCallback(question);
+  };
+
+  const planned = question.data.planned?.split(",") || [];
+  const unsupported = question.data.unsupported?.split(",") || [];
+
+  return (
+    <Popup isOpen title={_("Not supported AutoYaST elements")}>
```
mvidner, line 75:
```suggestion
    <Popup isOpen title={_("Unsupported AutoYaST elements")}>
```

"not supported" works well in a predicate position like "Foo is not supported" and below,
 but not so well in an adjective position like above.

mvidner:
>     2. "Not supported AY elements" popup

Also IIUC the popup displays the "planned" and "no" categories. What about "partial"?

service/lib/agama/autoyast/profile_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "agama/autoyast/profile_description"
+
+module Agama
+  module AutoYaST
+    # This class checks an AutoYaST profile and determines which unsupported elements are used.
+    #
+    # It does not report unknown elements.
```
mvidner, line 28: So what happens with unknown elements? Are they silently ignored now or does a different part of Agama report them?

imobachgs: I decided to ignore them by now and report only the status of the known elements. The idea is to extend the `compat` file to cover all the AutoYaST elements, stating whether it is unsupported, supported or planned.

imobachgs:
> Comments based so far on the description only, not on code.
> 
>     1. Agama checks the profile for unsupported elements.
>        When does this happen? Only when I try to use the profile, or can I trigger it beforehand, on a dry run?

Well, it happens automatically when using an AutoYaST profile (`agama.auto=`) in kernel's command line. However, I use the following command:

```
agama-autoyast http://192.168.122.10/autoyast/wrong.xml autoinst
```

IMHO, `agama-autoyast` should be "hidden" (probably placed on `/usr/libexec`) and we should extend the CLI to allow this operation. What do you think?

> 
>     2. "Not supported AY elements" popup
>        I think it would be nice to also include this text.
>        > This popup can be disabled by passing agama.ay_check=0 on the kernel command line

Good point.

dgdavid (approved):
Interface Code LGTM

dgdavid:
> 2\. "Not supported AY elements" popup
>     I think it would be nice to also include this text.
>     > This popup can be disabled by passing agama.ay_check=0 on the kernel command line

+1

service/share/autoyast-compat.json
```diff
@@ -68,6 +68,7 @@
           { "key": "description", "support": "no" },
           {
             "key": "bootproto",
+            "support": "no",
```
mvidner, line 71: @imobachgs Creating the schema did help finding the missing `support` here. Yay! Is `no` right?

dgdavid:
> This popup can be disabled by passing agama.ay_check=0 on the kernel command line

@imobachgs 

In the review meeting you said this is actually not disabling the check, but just not informing the user. I wonder if it deserves another name.

imobachgs:
> > This popup can be disabled by passing agama.ay_check=0 on the kernel command line
> 
> @imobachgs
> 
> In the review meeting you said this is actually not disabling the check, but just not informing the user. I wonder if it deserves another name.

I do not think so. It would be pretty long to explain that in a name.

dgdavid:
> > > This popup can be disabled by passing agama.ay_check=0 on the kernel command line
> > 
> > 
> > @imobachgs
> > In the review meeting you said this is actually not disabling the check, but just not informing the user. I wonder if it deserves another name.
> 
> I do not think so. It would be pretty long to explain that in a name.

I see. I was thinking on ay_check_alert=0. Not that much longer, but longer.

imobachgs:
> In today's review you mentioned the problem of keeping the JSON in sync with the code. I know that automating the synchronization would be hard... can we at least detect automatically that the JSON and the implementation have gone out of sync?

I do not think so. Perhaps we can create a PBI to work on that but not now.

mvidner:
> > can we at least detect automatically that the JSON and the implementation have gone out of sync?
> 
> I do not think so. Perhaps we can create a PBI to work on that but not now.

OK, would comments like this make sense at some places?

```rb
# Documentation:
# When updating this code, remember to update service/share/autoyast-compat.json
# to tell users the correct extent of AutoYaST compatibility.
```

imobachgs:
> > > can we at least detect automatically that the JSON and the implementation have gone out of sync?
> > 
> > 
> > I do not think so. Perhaps we can create a PBI to work on that but not now.
> 
> OK, would comments like this make sense at some places?
> 
> ```ruby
> # Documentation:
> # When updating this code, remember to update service/share/autoyast-compat.json
> # to tell users the correct extent of AutoYaST compatibility.
> ```
> 
> IIUC it applies to this whole tree [service/lib/agama/autoyast](https://github.com/agama-project/agama/tree/dbd4927c6b5abc4fb45d6dc9d1b7dc596f6bc982/service/lib/agama/autoyast)

Yes, it applies to the whole tree. I am not sure where is the best place to put such advice.

mvidner:
> > IIUC it applies to this whole tree [service/lib/agama/autoyast](https://github.com/agama-project/agama/tree/dbd4927c6b5abc4fb45d6dc9d1b7dc596f6bc982/service/lib/agama/autoyast)
> 
> Yes, it applies to the whole tree. I am not sure where is the best place to put such advice.

_Dream mode on_: ideally I would tell Github: when touching files that match this glob, enable this reminder, part of the PR template:
>  Modification of `Agama::AutoYaST` code detected. Please remember to adjust the relevant docs: `doc/...` and the compatibility support data: `service/share/autoyast-compat.json`

Hmm, we could even implement it as a CI check that triggers via [/on/push/paths](https://github.com/agama-project/agama/blob/dbd4927c6b5abc4fb45d6dc9d1b7dc596f6bc982/.github/workflows/ci-rubocop.yml#L5) and fails if the doc files are not modified.
 
OK, but am probably investing too much complexity into a minor problem

service/lib/agama/autoyast/profile_description.rb
```diff
@@ -0,0 +1,155 @@
...
+      # Returns the support level.
+      #
+      # If it was not specified when building the object, it infers it from its children. If the
+      # element has no children, it returns :no.
+      def support
+        return @support if @support
+
+        nested = children.map(&:support).uniq
+        return nested.first if nested.size == 1
+        return :partial if nested.include?(:yes) || nested.include?(:partial)
```
mvidner, line 80: Ah, so `partial` support is never declared in the "compat file" (for which we need a better name!) but computed, meaning "at least one of descendants" is declared as `support: yes`

service/lib/agama/autoyast/profile_description.rb
```diff
@@ -0,0 +1,155 @@
...
+      def short_key
+        key.split(ProfileDescription::SEPARATOR).last
+      end
+    end
+
+    # Describes the AutoYaST profile format.
+    class ProfileDescription
+      SEPARATOR = "/"
+
+      attr_reader :elements
```
mvidner, line 105:
```suggestion
      # @return [Array<ProfileElement>]
      attr_reader :elements
```

service/lib/agama/autoyast/profile_fetcher.rb
```diff
@@ -0,0 +1,102 @@
...
+    #
+    # It generates an AutoYaST profile that can be converted into an Agama configuration using the
+    # {Agama::AutoYaST::Converter} class.
+    class ProfileFetcher
+      # @param profile_url [String] Profile URL
+      def initialize(profile_url)
+        @profile_url = profile_url
+      end
+
+      # Converts the profile into a set of files that Agama can process.
```
mvidner, line 39: Perhaps nitpicking, but the description of this class (and of its public methods) doesn't tell me how it collaborates with other classes.
... ah, this `@return` seems to be the key thing I was missing

```suggestion
      # Converts the profile into a set of files that Agama can process.
      # @return [Hash,nil] an evaluated AutoYaST profile
```

mvidner:
I should just cherry pick the single-commit https://github.com/agama-project/agama/pull/1978 into this one

mvidner (approved):
LGTM, thanks!

PR #1977, opened 2025-02-05
Button to add a drive to the storage proposal

ancorgs:
Adds the following button to the new storage UI.

addDrive

If there are no extra devices to add, the button is not displayed.

### Testing

- Tested manually (see screenshot)
- Added unit test for the new menu component

### DIFF
--- a/web/src/components/core/MenuHeader.tsx
+++ b/web/src/components/core/MenuHeader.tsx
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) [2023-2025] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Flex, Content } from "@patternfly/react-core";
+import spacingStyles from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+
+export type MenuHeaderProps = { title: string; description: React.ReactNode };
+
+/**
+ * Renders the content to be used at a menu entry describing subsequent options.
+ * @component
+ *
+ * @param title - Short sentence describing the functionality.
+ * @param descriptions - Extra details rendered with a smaller font.
+ */
+export default function MenuHeader({ title, description }: MenuHeaderProps) {
+  return (
+    <Flex
+      direction={{ default: "column" }}
+      gap={{ default: "gapXs" }}
+      className={[spacingStyles.pxMd, spacingStyles.pyXs].join(" ")}
+    >
+      <Content component="h4">{title}</Content>
+      {description && <small>{description}</small>}
+    </Flex>
+  );
+}
--- a/web/src/components/storage/AddExistingDeviceMenu.tsx
+++ b/web/src/components/storage/AddExistingDeviceMenu.tsx
@@ -80,6 +80,7 @@ export default function AddExistingDeviceMenu() {
       )}
     >
       <DropdownList>
+        {/* @ts-ignore:next-line See https://github.com/patternfly/patternfly/issues/7327 */}
         <DropdownGroup label={<Header drives={model.drives} />}>
           <Divider />
           {devices.map((device) => (
--- a/web/src/components/storage/AddExistingDeviceMenu.tsx
+++ b/web/src/components/storage/AddExistingDeviceMenu.tsx
@@ -80,7 +80,7 @@ export default function AddExistingDeviceMenu() {
       )}
     >
       <DropdownList>
-        {/* @ts-ignore:next-line See https://github.com/patternfly/patternfly/issues/7327 */}
+        {/* @ts-expect-error See https://github.com/patternfly/patternfly/issues/7327 */}
         <DropdownGroup label={<Header drives={model.drives} />}>
           <Divider />
           {devices.map((device) => (
--- a/web/src/components/storage/AddExistingDeviceMenu.tsx
+++ b/web/src/components/storage/AddExistingDeviceMenu.tsx
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) [2025] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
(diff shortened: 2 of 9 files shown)


ancorgs:
@dgdavid The tests are failing due to this PatternFly type definitions. Label only accepts a string for `SelectGroup` and for `DropDownGroup`. Although it accepts a ReactNode for `MenuGroup`.

In fact it works just fine, so the type restriction looks kind of arbitrary to me.

Is there a way to fix it without having to migrate the whole component to use a menu instead of a dropdown?

web/src/components/core/MenuHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+import spacingStyles from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+
+/**
+ * Renders the content to be used at a menu entry describing subsequent options.
+ * @component
+ *
+ * @param title - Short sentence describing the functionality.
+ * @param descriptions - Extra details rendered with a smaller font.
+ */
+export default function MenuHeader({ title, description }): React.ReactNode {
```
joseivanlopez, line 34: Please, define a `MenuHeaderProps` type.

web/src/components/storage/MenuDeviceDescription.tsx
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) [2023-2025] SUSE LLC
```
joseivanlopez, line 2: NP: should it be [2025] only?

web/src/components/storage/MenuDeviceDescription.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Stack, Split, Label } from "@patternfly/react-core";
+import { typeDescription, contentDescription } from "~/components/storage/utils/device";
+
+/**
+ * Renders the content to be used at a menu entry describing a device.
+ * @component
+ *
+ * @param device - Device to represent
+ */
+export default function MenuDeviceDescription({ device }): React.ReactNode {
```
joseivanlopez, line 33: Please, define a `MenuDeviceDescriptionProps` type.

web/src/queries/storage/config-model.ts
```diff
@@ -298,6 +307,22 @@ export function usePartition(driveName: string, mountPath: string): PartitionHoo
...
 }
 
+/**
+ * Hook for operating on the collection of drives from the model.
+ */
+export type DrivesHook = {
+  add: (driveName: string) => void;
+};
+
+export function useDrives(): DrivesHook {
```
joseivanlopez, line 317: In the PR#1935 we added the `useModel` hook to query at model level, see https://github.com/agama-project/agama/pull/1935/commits/cb28ed6688f4d6b2c8716cd544d351dea36ca4a0#diff-c74d91a4c994f0b9fc0d945b55da1bb5d35a78503f1dfeb65c2b1b308a85dfc0R393. Maybe we can add `#addDrive` to that hook.

web/src/components/storage/MenuDeviceDescription.tsx
ancorgs: No. It's code moved from another file. The original code is not new.

web/src/queries/storage/config-model.ts
ancorgs: I was looking for something like that, but I didn't find it (because is still not merged).

web/src/components/core/MenuHeader.tsx
dgdavid: Or simply add the types

```suggestion
export default function MenuHeader({ title, description }: { title: string, description: string) {

```

dgdavid: NP: BTW, as said in another review, until now we have been omitting types that TypeScript is able to infer.

dgdavid:
> @dgdavid The tests are failing due to PatternFly type definitions. The property `label` only accepts a string for `SelectGroup` and for `DropDownGroup`. Although it accepts a ReactNode for `MenuGroup`.
> 
> In fact it works just fine, so the type restriction looks kind of arbitrary to me.
> 
> Is there a way to fix it without having to migrate the whole component to use a menu instead of a dropdown?

Looking at their code base and git history, look to me as an out-dated prop. 
It's is true that internally [SelectGroup is using MenuGroup](https://github.com/patternfly/patternfly-react/blob/deee54dcf05d7237e8ac32248868673ea1916cc5/packages/react-core/src/components/Select/SelectGroup.tsx#L23). But hard to say whether it's label type is intentional or not. 

What I can see is that MenuGroup#label was an string until two years ago, https://github.com/patternfly/patternfly-react/pull/8219 and the component itself was added 5 years ago https://github.com/patternfly/patternfly-react/commit/e867f093c5eef175b44cbb88e3ad229376de123f While the SelectGroup#label types dates from 6 years ago.

So, maybe a _bug report_ for PatternFly :)

Meanwhile, I'd migrate it to Menu if it does not hurt or, if you prefer, add a `/** @ts-expect-error: reason text */` comment.

web/src/components/storage/UseDeviceMenu.tsx
```diff
@@ -0,0 +1,98 @@
...
+  MenuToggleElement,
+  MenuToggle,
+  Divider,
+} from "@patternfly/react-core";
+import { MenuHeader } from "~/components/core";
+import MenuDeviceDescription from "./MenuDeviceDescription";
+import { useAvailableDevices, useConfigModel, useDrives } from "~/queries/storage";
+import { deviceLabel } from "~/components/storage/utils";
+
+export default function UseDeviceMenu() {
```
dgdavid, line 40: I know naming is hard as hell, but I'd avoid `UseDeviceMenu`. In general, I'd avoid naming components with verbs as prefix, but specially `Use` in the React world is too couple to hooks IMHO.

dgdavid (commented):
Would be nice if you add these missing tests :smiling_imp:

web/src/components/storage/UseDeviceMenu.tsx
dgdavid: What about `DevicesMenu`? More consistent whith the rest of name until now which looks like Thing plus Group when needed, as `ProposalPage`, `ConfigEditorMenu`, etc.

web/src/components/storage/AddExistingDeviceMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+          onClick={toggle}
+          aria-label={_("Use additional disk toggle")}
+          isExpanded={isOpen}
+        >
+          {_("Use additional disk")}
+        </MenuToggle>
+      )}
+    >
+      <DropdownList>
+        {/* @ts-ignore:next-line See https://github.com/patternfly/patternfly/issues/7327 */}
```
dgdavid, line 83: Please, use @ts-expect-error instead. It will let us know when there is no an error anymore if PatternFly finally ends up changing the label prop type. In such case, we can drop the directive instead of leaving there forever, which might hide another typing error in the future.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

dgdavid (approved):
Except the minor comment I put and the fact that I think we should improve the hooks approach (our of the scope of this PR), LGTM :+1:

web/src/components/storage/AddExistingDeviceMenu.tsx
dgdavid: And the same with @ts-ignore

> Hey-o 👋 I might seem to be helpful, but you should really avoid using me in the code!

PR #1978, opened 2025-02-07
add error context: Failed to run agama-autoyast

mvidner:
## Problem

it says "No such file or directory" when I pass an existing profile. It turns out it cannot find a helper _program_

Encountered when trying to run something from a git checkout, it probably works fine on our ISO

## Solution

Add context to the error

## Testing

- Tested manually

## Screenshots

No


imobachgs:
Added to #1976.

PR #1980, opened 2025-02-07
First version of an improved Proposal error alert

ancorgs:
## Problem

The current error displayed when the proposal failed is not good enough, we just moved the content from the old "result" box (that is not displayed anymore if the proposal is not possible) to an alert at the top of the screen. But we never adapted the texts.

## Solution

We want a better error that gives the user some hint of what's going on. This is a first version.

failed_proposal_step1

## Future

The is room for improvements, like adding texts explaining the impact of disabling Btrfs snapshots and so on. But we need to move fast, so we agreed to not invest too much for now.

https://github.com/ancorgs/agama/pull/2 is a draft pull request on top of this one that shows all the information that could be displayed. But it needs more work regarding code organization, look&feel and wording.

### DIFF
--- a/web/src/components/storage/ProposalFailedInfo.tsx
+++ b/web/src/components/storage/ProposalFailedInfo.tsx
@@ -21,7 +21,7 @@
  */
 
 import React from "react";
-import { Alert } from "@patternfly/react-core";
+import { Alert, Content } from "@patternfly/react-core";
 import { _, n_, formatList } from "~/i18n";
 import { useIssues } from "~/queries/issues";
 import { useConfigModel } from "~/queries/storage";
@@ -34,11 +34,11 @@ function Description({ partitions }) {
 
   if (!newPartitions.length) {
     return (
-      <p>
+      <Content>
         {_(
           "It is not possible to install the system with the current configuration. Adjust the settings below.",
         )}
-      </p>
+      </Content>
     );
   }
 
@@ -57,15 +57,16 @@ function Description({ partitions }) {
 
   return (
     <>
-      <p>{msg1}</p>
-      <p>{_("Adjust the settings below to make the new system fit into the available space.")}</p>
+      <Content>{msg1}</Content>
+      <Content>
+        {_("Adjust the settings below to make the new system fit into the available space.")}
+      </Content>
     </>
   );
 }
 
 /**
  * Information about a failed storage proposal
- * @component
  *
  */
 export default function ProposalFailedInfo() {
--- a/web/src/components/storage/ProposalFailedInfo.tsx
+++ b/web/src/components/storage/ProposalFailedInfo.tsx
@@ -27,6 +27,7 @@ import { useIssues } from "~/queries/issues";
 import { useConfigModel } from "~/queries/storage";
 import { IssueSeverity } from "~/types/issues";
 import * as partitionUtils from "~/components/storage/utils/partition";
+import { sprintf } from "sprintf-js";
 
 function Description({ partitions }) {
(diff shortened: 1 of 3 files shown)


web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -0,0 +1,83 @@
...
+      mountPaths.length,
+    ),
+    formatList(mountPaths),
+  );
+
+  return (
+    <>
+      <p>{msg1}</p>
+      <p>{_("Adjust the settings below to make the new system fit into the available space.")}</p>
+    </>
```
dgdavid, line 62: You're already on top of a branch using PF6, so you can use `<Content>` or `<Content component="p">` here if you wish for letting the design system / components framework to do its styling work.

Not sure if maybe you already now it but still using plain `<p>` because the lack of spacing between them due to the CSS reset. If so, let me clarify two things

* I also prefer plain HTML when possible, but the reality if that they are not directly styled with the default styles of the tool. We should avoid taking shortcuts to by-pass default styles, but understand and adapt them instead.
* This reminds me that we should adjust a little bit more the header size used by these alerts, since it is too big IMHO.

web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -0,0 +1,84 @@
...
+  );
+}
+
+/**
+ * Information about a failed storage proposal
+ * @component
```
dgdavid, line 68: NP: I'm not adding that JSDoc directive. Not sure how useful it is right now, and we also commented somewhere that public documentation of Agama components does not make sense at this point of development. I think that was useful for the tool generating public documentation, don't know. 

I should have commented this in a previous PR, but wanted to reduce the level of NP. Anyway, this is just my view and what I'm doing. Feel free to keep using them.

web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -0,0 +1,84 @@
...
+ *
+ */
+export default function ProposalFailedInfo() {
```
dgdavid, line 71: NP: you're using a warning alert but naming the component as Info :smiling_face_with_tear:

dgdavid (approved):
:heavy_check_mark:

web/src/components/storage/ProposalFailedInfo.tsx
ancorgs: Just because there was already a `ProposalTransactionalInfo` also used to display a similar `<Alert>` at the top of the page. On the other hand, I think that using WhateverInfo scales nicely for future changes in severity (ie. a different variant) or if we change the component we use to present those boxes of information. "Info" is more aseptic than "alert", "warning", "error" or any of those things that may change.

web/src/components/storage/ProposalFailedInfo.tsx
ancorgs: Fine with me, I was just following what I saw in others.

web/src/components/storage/ProposalFailedInfo.tsx
dgdavid: > "Info" is more aseptic than "alert", "warning", "error" or any of those things that may change.

You’re right. I didn’t give it much thought when I wrote the comment above, but actually I fully agree with quoted statement.

dgdavid:
Arg! I realize now that component still missing a test. I overlooked it while doing the review. I guess it's related to the fact that this was just the "step1" of the work.

PR #1984, opened 2025-02-10
Make the "lang" URL query optional

lslezak:
## Problem

- Agama fails starting locally on non-UEFI systems

agama-missing-lang

## Details

- The `lang` URL parameter is used only in UEFI systems, on other systems it is not defined/used.
- It is related to #1924

## Solution

- Make the `lang` URL query parameter optional

## Testing

- Tested manually in bot UEFI and non-UEFI (standard BIOS) systems


imobachgs (approved):

bittin:
This is fixed now, thanks for taking care of it

PR #1986, opened 2025-02-11
feat(web): add option to reset storage config

joseivanlopez:
Add *Reset to defaults* option into the *More options* menu.

Screenshot on an s390x system with several disks:

s390

Screenshot on a non-s390x system with several disks:

other

### DIFF
--- a/doc/dbus/bus/org.opensuse.Agama.Storage1.bus.xml
+++ b/doc/dbus/bus/org.opensuse.Agama.Storage1.bus.xml
@@ -51,6 +51,9 @@
       <arg name="serialized_config" direction="in" type="s"/>
       <arg name="result" direction="out" type="u"/>
     </method>
+    <method name="ResetConfig">
+      <arg name="result" direction="out" type="u"/>
+    </method>
     <method name="GetConfig">
       <arg name="serialized_config" direction="out" type="s"/>
     </method>
--- a/doc/dbus/org.opensuse.Agama.Storage1.doc.xml
+++ b/doc/dbus/org.opensuse.Agama.Storage1.doc.xml
@@ -40,6 +40,17 @@
       <arg name="result" direction="out" type="u"/>
     </method>
     <!--
+      Resets the default storage config defined by the selected product.
+    -->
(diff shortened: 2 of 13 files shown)


joseivanlopez:
Screenshot From 2025-02-11 12-55-57

ancorgs:
I would take this as an opportunity to clean-up/reorganize "more options". Let's say:

- Reset to default configuration
- Change boot options
- [Separator]
- iSCSI
- s390 stuff

dgdavid:
> I would take this as an opportunity to clean-up/reorganize "more options". Let's say:
> 
> * Change boot options
> * Reset to default configuration
> * [Separator]
> * Configure iSCSI
> * [s390 stuff (with their current titles)]

Excuse my ignorance, but what about LVM and RAID options? Part of "Use additional disk"?

dgdavid:
> > Excuse my ignorance, but what about LVM and RAID options? Part of "Use additional disk"?
> 
> No. They will be here... but we have not implemented any UI for them yet. :-(

Ah, right. So, the idea is to remove them by now and decide later their place, isn't it?

dgdavid:
> Just a note. At #1972 we agreed on the following wording for the buttons/links displayed on the empty state screens (sure is not the same context, so we don't have to reproduce them here):
> 
>     * Reset to the default configuration
> 
>     * Connect to iSCSI targets
> 
>     * Activate zFCP disks
> 
>     * Manage DASD devices

That was a good improvement, indeed.

> 
> 
> Not sure what is better here. Maybe those are too long and even confusing for a drop-down menu. Or maybe they could be used as descriptions. Something like:
> 
>     * Change boot options - Select the disk to configure partitions for booting
> 
>     * Reset to defaults - Start from scratch with the default configuration
> 
>     * [Separator]
> 
>     * Configure iSCSI - Discover and connect to iSCSI targets
> 
>     * Configure zFCP - Activate zFCP disks
> 
>     * Configure DASD - Activate and format DASD devices

I'd go for something like this. Items with descriptions.

ancorgs (approved):
LGTM

PR #1990, opened 2025-02-12
Live ISO optimizations: faster build and smaller ISO

lslezak:
All these changes make the build one minute faster (on my local workstation) and the resulting ISO is 34MB smaller (on x86_64, might be different elsewhere).

## Improvements

- Do not print details about removed kernel drivers and firmware files during build, this speeds up the build significantly
 (~1 minute faster build) and avoids huge build log.
- Uncompress the kernel drivers, no need to compress them twice (they are compressed by the squashfs as well). Compressing all drivers together in the image is more effective than compressing several thousands individual files. This makes the image about 33MB smaller (on x86_64).
- Hardlink the duplicate licenses, makes the ISO ~1MB smaller

### DIFF
--- a/live/root/tmp/driver_cleanup.rb
+++ b/live/root/tmp/driver_cleanup.rb
@@ -127,25 +127,9 @@ loop do
   to_delete.reject!{|a| referenced.any?{|d| d.path == a.path}}
 end
 
-# total size counter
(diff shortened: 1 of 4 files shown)


mvidner (approved):
Nice!

PR #1992, opened 2025-02-13
Change the default lsm to SELinux for both Leap and Tumblweed jsc#PED-12021

lkocman:
- Change the default lsm to SELinux enforcing for both Leap and Tumblweed
  keep AppArmor around as an option jsc#PED-12021

Related article  https://news.opensuse.org/2025/02/13/tw-plans-to-adopt-selinux-as-default/


DimStar77:
Matching the PR done in the yast2-based installer: https://github.com/yast/skelcd-control-openSUSE/pull/287

PR #1993, opened 2025-02-13
Replace agama cmdline args prefix with inst

mchf:
## Problem

We want to use generic prefix for kernel command line arguments prefix.

## Solution

replaced agama. with inst.

## Testing

- *Tested manually*

### DIFF
--- a/live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh
+++ b/live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh
@@ -15,7 +15,7 @@ get_agama_args() {
       _found=1
       _env=1
       ;;
-    agama*)
+    inst* | agama*)
       _found=1
       ;;
     esac
--- a/service/lib/agama/cmdline_args.rb
+++ b/service/lib/agama/cmdline_args.rb
@@ -74,6 +74,8 @@ module Agama
     # command line options into a standard shape
     def self.standardize(option)
       # agama. is now obsolete original kernel argument prefix
+      return if !option.start_with?("agama.")
+
       option.sub("agama.", CMDLINE_PREFIX)
     end
 
--- a/service/lib/agama/cmdline_args.rb
+++ b/service/lib/agama/cmdline_args.rb
@@ -73,9 +73,8 @@ module Agama
     # Despite Agama is young it already contains some relicts. This method should purge them and put
     # command line options into a standard shape
     def self.standardize(option)
-      obsolete_cmdline_prefix = "agama."
-
-      option.sub(obsolete_cmdline_prefix, CMDLINE_PREFIX)
+      # agama. is now obsolete original kernel argument prefix
+      option.sub("agama.", CMDLINE_PREFIX)
     end
 
     # Convenience method to normalize the given value by now it just convert "true" and "false"
--- a/service/lib/agama/cmdline_args.rb
+++ b/service/lib/agama/cmdline_args.rb
@@ -22,7 +22,8 @@
 module Agama
   # This class is responsible for reading Agama kernel cmdline options
   class CmdlineArgs
-    CMDLINE_PREFIX = "agama."
+    CMDLINE_PATH = "/proc/cmdline"
+    CMDLINE_PREFIX = "inst."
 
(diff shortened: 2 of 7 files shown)


service/lib/agama/cmdline_args.rb
```diff
@@ -63,6 +64,14 @@ def self.read_from(path)
       args
     end
 
+    # Despite Agama is young it already contains some relicts. This method should purge them and put
+    # command line options into a standard shape
+    def self.standardize(option)
+      obsolete_cmdline_prefix = "agama."
```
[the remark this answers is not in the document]
mchf: I didn't want to mess "global" scope with a constant with very specific / limited usage. So, I'll go for second option.

mchf:
if this: https://github.com/agama-project/agama/pull/1970 is merged, then some more love for this patch will be needed.

service/lib/agama/cmdline_args.rb
```diff
@@ -41,6 +41,7 @@ def self.read_from(path)
       args = new({})
 
       options.split.each do |option|
+        option = standardize(option)
```
imobachgs, line 44: It is OK, but I guess you can even do the substitution on the whole line.

imobachgs (commented):
It looks good. But, please, update the changes file.

service/lib/agama/cmdline_args.rb
imobachgs: Yes, you are right.

mchf:
Rebased on top of patches made by @teclator and @jreidinger

mchf:
I'm currently retesting it with all those new patches merged ....

jreidinger (changes requested):
It is missing adaptation at https://github.com/agama-project/agama/blob/ba00a180da7d0feaf3c1b525ac10f5bbc98083e6/live/root/usr/bin/kernel-cmdline-conf.sh#L19 and https://github.com/agama-project/agama/blob/ba00a180da7d0feaf3c1b525ac10f5bbc98083e6/live/root/usr/bin/info-cmdline-conf.sh#L12 if you want to support inst.info

service/lib/agama/cmdline_args.rb
```diff
@@ -68,6 +70,13 @@ def self.read_from(path)
       args
     end
 
+    # Despite Agama is young it already contains some relicts. This method should purge them and put
+    # command line options into a standard shape
+    def self.standardize(option)
+      # agama. is now obsolete original kernel argument prefix
+      option.sub("agama.", CMDLINE_PREFIX)
```
jreidinger, line 77: sadly this does not work as you intend as it replace also inside of values. So it is quite fragile:
```
jreidinger@localhost:~/prace> irb
irb(main):001:0> "my_arg=agama.place".sub("agama.", "inst.")
=> "my_arg=inst.place"
```

jreidinger: similarly if you use e.g. `inst.info=http://agama.host.com/my.info` it will be broken.

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh
```diff
@@ -15,7 +15,7 @@ get_agama_args() {
       _found=1
       _env=1
       ;;
-    agama*)
+    inst* | agama*)
```
teclator, line 18: Maybe we should inform that the agama*whatever parameter is deprecated in favor of the inst equivalent... like the getarg -d option...

service/lib/agama/cmdline_args.rb
mchf: it works ... because next line filters such key, value pairs out (it requires option to start_with CMDLINE_PREFIX)

but i agree it is not good practise and will polish it

live/root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh
mchf: ```agama``` being obsolete is covered in documentation update ...

question is whether will someone be able to recognize such output ?

jreidinger (approved):

Issue #1960, opened 2025-01-30
Fix wrong L10n key in autoyast converter output


jreidinger:
This sub-issue is focused only on fixing output of autoyast converter.

imobachgs:
Fixed in #1995.

Issue #1907, opened 2025-01-16
Output pretty json from autoyast converter


jreidinger:
It is mainly about changing to_json to pretty_generate at https://github.com/agama-project/agama/blob/master/service/lib/agama/autoyast/converter.rb#L66

imobachgs:
Fixed by #1995.

PR #1995, opened 2025-02-13
fix(ruby): fix localization and scripts conversion

imobachgs:
## Problem

The conversion of the localization and scripts sections has several problems:

- #1903
- #1907
- #1960

This PR include some changes from #1951 and #1946.

## Solution

- Fix the handling of the scripts file name and location.
- Fix the name of the localization setting ("localization" instead of "l10n").

Additionally, use "pretty" JSON when exporting the converted profile.

## Testing

- _Added a new unit test_

### DIFF
--- a/autoinstallation/README.md
+++ b/autoinstallation/README.md
@@ -32,8 +32,8 @@ repository to get an impression of how a profile looks like.
 
 ```json
 {
-  "l10n": {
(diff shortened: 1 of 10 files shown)


joseivanlopez (approved):
LGTM

Issue #1985, opened 2025-02-11
UI/UX: Improve libzypp error callbacks


lslezak:
## Problem

- The current UI for the libzypp callbacks is inconsistent and unintuitive. The user actions are not described so the user can be confused which option to choose and what will happen.
- Related to
  - #1909 
  - #1932

## Screenshots

Package download
![Error message from libzypp](https://github.com/user-attachments/assets/66808bc4-0b0d-4b3d-b64d-7d642ba9af9e)

Package installation
![agama-package-callbacks](https://github.com/user-attachments/assets/ceb9bbe5-72a1-437c-8061-9dc199f151d9)

## Details

- Different button order ("Retry" first vs last)
- Different default action ("Retry" vs "Skip")
- Different button labels "Skip" and "Ignore" although the action is basically the same: do not care about the problem.
- Unclear result, skipping/ignoring causes inconsistent system with broken dependencies. If an important package is missing the system later might not work properly, they might be crashes, security problems or the system might not even boot. User should be aware of these consequences.
- It is not clear that the "Cancel" button aborts the installation completely.

Note: As discussed at the review, in the future we might reduce the number of offered actions to just two: Retry or abort the installation. After skipping a package the same download or installation problem will very likely happen again with the next package. So in the end the user has to abort the installation anyway. And as written above, skipping results in a broken system so we should discourage from that.

## Proposal

- Use "Skip" button in both cases
- Probably use "Abort" label instead of "Cancel", that has a stronger meaning. Maybe use "Abort installation" to make it clear that it aborts the installation completely.
- The default should be "Retry", it is a safe action which can potentially fix the problem, all others will result in an incomplete/inconsistent system and are not safe. (Of course, if retry fails again and again then in the end the user has to abort anyway, but that's a different story...)
- Describe the consequences of the "Skip" and "Abort" actions in additional text below the error message.

lslezak:
@dgdavid any comments or proposals?

dgdavid:
First of all,  I miss a dialog title. We need a suitable title that quickly sets the context for such an intrusive dialog. Options like "Installation Issue," "Missing package," or "Package not found" could work. The key is to ensure it immediately conveys the problem.

[...]

PR #1996, opened 2025-02-14
fix(ux): Improve libzypp callbacks

lslezak:
## Problem

- https://github.com/agama-project/agama/issues/1985
- The current UI is not much intuitive :worried: 
  Error message from libzypp

## Solution

- Better button labels (now even translated!)
- "Try again" is now the default
- Display a title to have a context
- Describe the consequences of skipping package installation
- @dgdavid proposed even more enhancements (see [his comment](https://github.com/agama-project/agama/issues/1985#issuecomment-2650697282)), but that would require non trivial enhancements

## Notes

- This is not a perfect solution, the current Query API is quite limited
- I'll create a new card for enhancing the API to  allow some more features (and remove the current workarounds)

## Testing

- Tested manually

## Screenshots

agama-error-package-download
agama-error-package-integrity
agama-error-package-installation2


service/lib/agama/software/callbacks/base.rb
```diff
@@ -0,0 +1,75 @@
...
+        def abort_label
+          # TRANSLATORS: button label, abort the installation completely after an error
+          _("Abort installation")
+        end
+
+        # dialog title
+        def title
+          # TRANSLATORS: Dialog title
+          _("Package installation failed")
+        end
```
dgdavid, line 63: I actually meant a title of the dialog. Here you have added the title to the question, which is ok, but then you have had to include it as part of the question text and use CSS workarounds for having separated lines.

web/src/components/questions/GenericQuestion.tsx
```diff
@@ -47,7 +47,9 @@ export default function GenericQuestion({
 
   return (
     <Popup isOpen aria-label={_("Question")}>
```
dgdavid, line 49: As said, I expected a title={something}  here. 

```suggestion
    <Popup isOpen title={_("Package installation failed")}>
```

I know this is the GenericQuestion component, but if I'm not wrong we could create an specific component for "RetryPackageInstallationQuestion" question and the backend give a type for it. Do not remember the details form the top of my head.

lslezak: Yes, but I'd like to avoid a specific PackageQuestion. I think having a title make sense in many cases (all?) so I'd like to support optional title in the generic implementation.

Because extending the Questions API is not trivial (do not forget that questions can be handled using command line interface) I'll create a separate card for that.

dgdavid: But the problem is that title is not exposed, but included in the text. It's not optimal IMHO and lead us to start adding workarounds not truly needed. 

Now we do not have actually a title, but just a short sentence in the full text. Try to "getByRole("heading", { name: "Package installation failed" }) to see what I mean. Or simply add a the prop title and test how the dialog looks.

service/lib/agama/software/callbacks/media.rb
```diff
@@ -75,13 +80,13 @@ def media_change(error_code, error, url, product, current, current_label, wanted
 
           question = Agama::Question.new(
             qclass:         "software.medium_error",
-            text:           error,
-            options:        [:Retry, :Skip],
-            default_option: :Skip,
+            text:           title + separator + broken_system_warning(error),
```
imobachgs, line 83: We need to evolve the question API, and that's true. However, I find this composition a bit tricky (especially because you rely later on CSS to make it look good).

With the current API, you could:

1. Add the missing information to the `data:` hash.
2. Build a specialized component on the React side that understands such dialogs.

web/src/components/questions/GenericQuestion.tsx
dgdavid: >  having a title make sense in many cases (all?)

Yes, all dialoga should have good titles

> Because extending the Questions API is not trivial (do not forget that questions can be handled using command line interface) I'll create a separate card for that.

So, let's do the temporary solution more robust by adding a specialized question component. Or goes por the implementation proposed by @imobachgs and add missing data as part of the mentioned hash data instead of gluing all in the text field.

dgdavid (commented):
Taking a look to the screenshots in the description

* ~~I'd not use icon as part of the title~~ But using it does not hurt and it is ok. Moreover, you have used the one used in the installation button when there are issues, which is great in terms of consistence. 
* I'd try to keep the buttons order. I.e, not adding the abort installation in the middle of the other two, which changes the order of buttons.
* I'd use a "tertirary" or "plain" variant for the "Abort installation" action, if possible. Also, I'd add the "isDanger" prop.

web/src/components/questions/PackageErrorQuestion.test.tsx
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) [2024] SUSE LLC
```
imobachgs, line 2:
```suggestion
 * Copyright (c) [2025] SUSE LLC
```

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Feb 14 09:46:02 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Render newlines in question messages
```
imobachgs, line 4: Is this the right description of the changes?

web/src/components/questions/PackageErrorQuestion.tsx
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (c) [2022-2024] SUSE LLC
```
imobachgs, line 2:
```suggestion
 * Copyright (c) [2022-2025] SUSE LLC
```

web/src/components/questions/PackageErrorQuestion.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <Popup
+      isOpen
+      title={_("Package installation failed")}
+      titleIconVariant={() => <Icon name="error" size="s" />}
+    >
+      <Text>{question.text}</Text>
+      {/* non-breaking space for extra empty line between texts */}
+      <Text>&nbsp;</Text>
```
imobachgs, line 67: Is this the best way to add some space there?

@dgdavid, what do you think?

web/src/components/questions/Questions.tsx
```diff
@@ -58,5 +59,10 @@ export default function Questions(): React.ReactNode {
     QuestionComponent = UnsupportedAutoYaST;
   }
 
+  // special popup for package errors (libzypp callbacks)
+  if (currentQuestion.class?.startsWith("software.package_error.")) {
```
imobachgs, line 63: :+1:

imobachgs (commented):
Just some minor things. Otherwise, it looks much better now.

web/src/components/questions/PackageErrorQuestion.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Text>{warning}</Text>
```
dgdavid, line 68: We shouldn't take car of extra lines. Use a layout component for laying out the content. I'd go for an Stack with gutters here.

```suggestion
      <Stack hasGutter>
        <Text>{question.text}</Text>
        <Text>{warning}</Text>
      </Stack>
```

web/src/components/questions/PackageErrorQuestion.tsx
dgdavid: See [comment](https://github.com/agama-project/agama/pull/1996#pullrequestreview-2623254712) below.

web/src/components/questions/PackageErrorQuestion.tsx
dgdavid: NP: it's actually a new file, so  

```suggestion
 * Copyright (c) [2025] SUSE LLC
```

dgdavid (approved):
Let's merge it, thank you

PR #2005, opened 2025-02-17
feat: expose passwords and drop autologin

imobachgs:
## Problem

The users API does not expose the passwords. Initially, we implemented like that for security
reasons, but giving that the password is send over an authenticated HTTPS connection, we do not
think it is relevant. After all, you are sending/receiving the password using the same channel.

Moreover, the D-Bus API is kind of inconsistent. The "first user" is exposed as a single "object",
while the root data are separate D-Bus properties.

## Solution

- Make user/root passwords visible.
- Drop the autologin property, as it is not wanted anymore.

About the D-Bus API, this PR puts all the root information together using a `RootUser` property.
Ideally, we should have a `SetRootUser` method too, but it will have to wait a bit.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/doc/dbus/org.opensuse.Agama.Users1.doc.xml
+++ b/doc/dbus/org.opensuse.Agama.Users1.doc.xml
@@ -29,6 +29,6 @@
       <arg name="result" direction="out" type="u"/>
     </method>
     <property type="(sbs)" name="RootUser" access="read"/>
-    <property type="(sssbba{sv})" name="FirstUser" access="read"/>
+    <property type="(sssba{sv})" name="FirstUser" access="read"/>
   </interface>
 </node>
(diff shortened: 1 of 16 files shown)


doc/dbus/org.opensuse.Agama.Users1.doc.xml
```diff
@@ -54,25 +28,7 @@
...
-    <property type="s" name="RootSSHKey" access="read"/>
-
-    <!--
-        FirstUser:
-        struct( string FullName, string UserName, string Password, boolean AutoLogin, map AdditionalData)
-        Info about first user to set. if Username is empty, it means not set and other values can be ignored
-    -->
-    <property type="(sssba{sv})" name="FirstUser" access="read"/>
+    <property type="(sbs)" name="RootUser" access="read"/>
+    <property type="(sssbba{sv})" name="FirstUser" access="read"/>
```
ancorgs, line 32: What's the second boolean?
name(s), username(s), password(s), hashedPassword(b), ?????(b), data(a{sv})

PR #2007, opened 2025-02-18
feat: initial SCC proxy support

imobachgs:
## Problem

Our QA team needs basic support to register against SCC proxy servers.

## Solution

* Add a new `inst.register_url` to define the URL for registering the system.
* Take the opportunity to clean-up unused registration code.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>
<summary>Error using a different server</summary>

Captura desde 2025-02-18 10-55-30
</details>

### DIFF
--- a/service/lib/agama/cmdline_args.rb
+++ b/service/lib/agama/cmdline_args.rb
@@ -35,6 +35,10 @@ module Agama
       @data = data
     end
 
+    def self.read
+      read_from("/proc/cmdline")
+    end
+
     # Reads the kernel command line options
     def self.read_from(path)
       options = File.read(path)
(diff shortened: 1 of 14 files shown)


service/lib/agama/cmdline_args.rb
```diff
@@ -35,6 +35,10 @@ def initialize(data = {})
       @data = data
     end
 
+    def self.read
+      read_from("/proc/cmdline")
```
jreidinger, line 39: sadly this collides with https://github.com/agama-project/agama/pull/1896 In general agama should not read /proc/cmdline but generated one...and that brings me to topic maybe for retro, why we have after beta1 so many open pull requests?

imobachgs: Having the `read` method is good, so you do not need to refer to `/proc/cmdline` everywhere. And we can adapt it whenever is needed.

lslezak (approved):
LGTM, I'd solve the collision with #1896 there..

PR #1999, opened 2025-02-15
feat(web): rework users interface

dgdavid:
Apart from **eliminating the step introduced in #1821 to force users to enter a password for root**, this PR significantly revamps the interface of the Authentication page (formerly the Users page) to reduce its complexity and make interactions more intuitive.

Summarizing a lot, the key changes include improving the first user form and moving the root user authentication methods to a new form. This allows users to easily set or clear authentication methods without having to navigate complex actions. I.e., tables and most dynamic actions have been eliminated, making the page much easier to interact with and enabling users to quickly find the options they need.

| Authentication page | Root user form |
|-|-|
|localhost_8080_ (21) | localhost_8080_ (22) |

However, there are still areas for improvement:

* Enhance validations and the underlying code that supports them.
* Decouple the input with suggestions from the FirstUserForm (move it to the core namespace).
* Use a simpler widget/input for adding public SSH keys. While the current one works, it is somewhat complex and not well-suited for these form’s use cases.
* Enable the ability to add more than one public SSH key.

## Note for reviewers

> [!WARNING]  
> After commit https://github.com/agama-project/agama/pull/1999/commits/c7c424c62fe844b9fcb1f7a3885a5e6f40832d09 this PR needs changes done at https://github.com/agama-project/agama/pull/2005 (not merged at the time of writing).

Please jump into the branch and  use the interface as part of the review process.

If interested in the details, check commits.

---

Related to https://trello.com/c/7ewFvR0X (internal link)

PR #2015, opened 2025-02-18
fix(web): make username suggestions work again

dgdavid:
These suggestions stop working at #1999 (specifically in commit afb1a54c4eda) which started using a form internal state for keeping input values instead of relying on the native Form API.

Sadly, there wasn't a test for the username suggestions and it was broken silently.


ancorgs (approved):
LGTM

ancorgs (approved):
LBTM (looks better to me)

PR #2017, opened 2025-02-18
web: navigation CSS adjustments

dgdavid:
A workaround for using less padding to avoid the menu overlapping the content. Better solution would be more than welcome.


ancorgs (approved):
LGTM

PR #2027, opened 2025-02-19
web: Prevent resizing or deleting re-used partitions

ancorgs:
## Problem

When using a custom space policy, it is possible to mark a reused partition to be shrunk or deleted.

## Solution

Prevent that at UI level

canttouchme

## Testing

- Tested manually
- Extended the component unit tests.

### DIFF
--- a/web/src/components/storage/SpaceActionsTable.tsx
+++ b/web/src/components/storage/SpaceActionsTable.tsx
@@ -34,7 +34,7 @@ import {
 import { sprintf } from "sprintf-js";
 
 import { _ } from "~/i18n";
-import { deviceSize } from "~/components/storage/utils";
+import { deviceSize, formattedPath } from "~/components/storage/utils";
 import {
   DeviceName,
   DeviceDetails,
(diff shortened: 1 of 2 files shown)


web/src/components/storage/SpaceActionsTable.tsx
```diff
@@ -43,8 +43,25 @@ import {
 } from "~/components/storage/device-utils";
 import { Icon } from "~/components/layout";
 import { PartitionSlot, SpacePolicyAction, StorageDevice } from "~/types/storage";
+import { configModel } from "~/api/storage/types";
 import { TreeTableColumn } from "~/components/core/TreeTable";
 import { Table, Td, Th, Tr, Thead, Tbody } from "@patternfly/react-table";
+import { useConfigModel } from "~/queries/storage/config-model";
+
+function isUsedPartition(partition: configModel.Partition): boolean {
```
dgdavid, line 51: Just a note because I do not remember if I already commented it in a PR from you: until now we were using the fat arrow sintax 🤷‍♂️

web/src/components/storage/SpaceActionsTable.tsx
```diff
@@ -43,8 +43,25 @@ import {
...
+  return partition.filesystem !== undefined || partition.alias !== undefined;
+}
+
+// FIXME: there is too much logic here. This is one of those cases that should be considered
```
dgdavid, line 55: 👍

dgdavid (approved):

web/src/components/storage/SpaceActionsTable.tsx
```diff
@@ -43,8 +43,25 @@ import {
 } from "~/components/storage/device-utils";
 import { Icon } from "~/components/layout";
 import { PartitionSlot, SpacePolicyAction, StorageDevice } from "~/types/storage";
+import { configModel } from "~/api/storage/types";
 import { TreeTableColumn } from "~/components/core/TreeTable";
 import { Table, Td, Th, Tr, Thead, Tbody } from "@patternfly/react-table";
+import { useConfigModel } from "~/queries/storage/config-model";
+
+const isUsedPartition = (partition: configModel.Partition): boolean => {
```
dgdavid, line 51: NP: these are the things I still envision in a storage utils module. But it's hard to convince you guys.

dgdavid (approved):

web/src/components/storage/SpaceActionsTable.tsx
ancorgs: Is not that we don't agree. There is also a comment right below that like saying tht we need to reconsider all that. It currently feels like dealing with the old YaST huge storage "maps" (pre storage-ng).

PR #2029, opened 2025-02-20
Fixed syntax error in 99agama-cmdline dracut module

lslezak:
## Problem

- The Live ISO does not boot

agama-boot-syntax-error

## Solution

- Fix the syntax error

## Testing

- Tested manually, the ISO boots fine


teclator (approved):
LGTM

imobachgs (approved):

PR #2031, opened 2025-02-20
feat(web): re-introduce the storage encryption settings

dgdavid:
First iteraration to bring the encryption section back.

The implementation at backend side is very simplistic and it does not properly work with several scenarios, for example:

* If some partitions are encrypted and others don't.
* If different encryptions are used (e.g., a password for root and other for home). 

In those cases, the model considers that all the new partitions are encrypted, and the encryption config is taken from either the root partition or any other encrypted partition.

Note that these scenarios can only happen if a custom config is loaded. They cannot be accomplished by using the UI.


web/src/components/storage/EncryptionSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+the device, including data, programs, and system files.",
+      )}
+      pfCardBodyProps={{ isFilled: true }}
+      actions={<Link to={STORAGE.encryption}>{_("Edit")}</Link>}
+    >
+      <Card isCompact isPlain>
+        <CardBody>
+          <DescriptionList isHorizontal isFluid displaySize="lg" isCompact>
+            <DescriptionListGroup>
+              <DescriptionListTerm>{_("Mode")}</DescriptionListTerm>
```
dgdavid, line 69: Obviously, a better _label_ would be needed here. But I did not get enough inspiration when writing this, as you can check.

web/src/components/storage/EncryptionSettingsPage.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React, { useState, useRef } from "react";
+import { useNavigate } from "react-router-dom";
+import { ActionGroup, Alert, Checkbox, Content, Form, Stack, Switch } from "@patternfly/react-core";
+import { Page, PasswordAndConfirmationInput } from "~/components/core";
+import { EncryptionMethods } from "~/types/storage";
+import sizingStyles from "@patternfly/react-styles/css/utilities/Sizing/sizing";
```
dgdavid, line 28: I'd like to drop this and its usage, but need a bit more of time for testing how to limit the width used by the Checkbox description without affecting to the rest of the form and allowing `Form#isWidthLimited` to do its work properly.

Will invest some time on Monday, if time permits.

ancorgs:
I just wanted to mention that the length of the form feels a big inconsistent depending on whether TPM is possible or not.

Without TPM

short

With TPM (it doesn't feel ideal and fully polished, but it's certainly better)

longer

ancorgs:
> I just wanted to mention that the length of the form feels a big inconsistent depending on whether TPM is possible or not.

I will fixed with a hack just for today.

dgdavid:
> I just wanted to mention that the length of the form feels a big inconsistent depending on whether TPM is possible or not.

I know. It is what I meant at https://github.com/agama-project/agama/pull/2031#discussion_r1963796377

I think we can go ahead wit it and fix it later since the "two versions" of the form will be not displayed at same time.

I'm going to make a suggestion, thought. Yo can check and use it or just provide a fix if you find one. I need more time for reaching a good solution.

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
```diff
@@ -0,0 +1,66 @@
...
+          end
+
+          # @return [Y2Storage::EncryptionMethod::Base]
+          def convert_method
+            method_conversions = {
+              "luks1"          => Y2Storage::EncryptionMethod::LUKS1,
+              "luks2"          => Y2Storage::EncryptionMethod::LUKS2,
+              "pervasiveLuks2" => Y2Storage::EncryptionMethod::PERVASIVE_LUKS2,
+              "tpmFde"         => Y2Storage::EncryptionMethod::TPM_FDE,
+              "randomSwap"     => Y2Storage::EncryptionMethod::RANDOM_SWAP,
```
ancorgs, line 55: NP: It does not make much sense to support the XXX_SWAP methods. They simply cannot be used as a global encryption method. Even the pervasive one would be debatable.

service/lib/agama/storage/config_conversions/to_model_conversions/encryption.rb
```diff
@@ -0,0 +1,65 @@
...
+          # @return [string]
+          def convert_method
+            method_conversions = {
+              Y2Storage::EncryptionMethod::LUKS1.id           => "luks1",
+              Y2Storage::EncryptionMethod::LUKS2.id           => "luks2",
+              Y2Storage::EncryptionMethod::PERVASIVE_LUKS2.id => "pervasiveLuks2",
+              Y2Storage::EncryptionMethod::TPM_FDE.id         => "tpmFde",
+              Y2Storage::EncryptionMethod::RANDOM_SWAP.id     => "randomSwap",
+              Y2Storage::EncryptionMethod::PROTECTED_SWAP.id  => "protectedSwap",
+              Y2Storage::EncryptionMethod::SECURE_SWAP.id     => "secureSwap"
```
ancorgs, line 56: NP: same than above, swap-specific methods are out of the question.

web/src/api/storage/types/config-model.ts
```diff
@@ -4,6 +4,14 @@
  * and run json-schema-to-typescript to regenerate this file.
  */
 
+export type EncryptionMethod =
+  | "luks1"
+  | "luks2"
+  | "pervasiveLuks2"
+  | "tpmFde"
+  | "protectedSwap"
```
ancorgs, line 12: NP: more of the same. Several methods make little sense there. Not crucial since all this is temporary.

web/src/components/storage/EncryptionSettingsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+                <p key={`error_${i}`}>{e}</p>
+              ))}
+            </Alert>
+          )}
+          <Switch
+            label={_("Encrypt the system")}
+            isChecked={isEnabled}
+            onChange={() => setIsEnabled(!isEnabled)}
+          />
+          <Stack className={isTpmAvailable && sizingStyles.w_50OnLg} hasGutter>
```
dgdavid, line 129: 👍

web/src/components/storage/EncryptionSettingsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+              initialValue={encryptionConfig?.password}
+              value={password}
+              onChange={changePassword}
+              isDisabled={!isEnabled}
+              showErrors={false}
+            />
+          </Stack>
+          {isTpmAvailable && (
+            <Checkbox
+              className={sizingStyles.w_50OnLg}
```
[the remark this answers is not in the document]
ancorgs: I tried that 75 and I didn't love the result. So went for the conditional class instead. Good enough for today.

ancorgs (approved):
LGTM. This is, of course, far from perfection. But it exactly what we need (and can afford) now.

PR #2033, opened 2025-02-20
New storage UI and migration to patternfly 6

ancorgs:
## Problem

The migration to Patternfly 6, the new storage UI and some changes at user management had been kept in a separate feature branch for too long.

## Solution

Finally merge all that into master.

Almost all commits were already reviewed during its integration into the `storage-config-ui` branch. Refer to the corresponding pull requests for details.

Compared to that, the only real addition in this pull request is the modification of the different changelog files.


imobachgs:
Once those little conflicts are fixed, we can merge.

PR #2041, opened 2025-02-20
doc: update AutoYaST compatibility document

imobachgs:
- docs(ruby): put unsupported sections at the end
- docs(ruby): add kdump to the AutoYaST compatibility document

Used to generate [the reference](https://agama-project.github.io/docs/user/autoyast/reference).

### DIFF
--- a/service/lib/tasks/autoyast.rb
+++ b/service/lib/tasks/autoyast.rb
@@ -39,12 +39,16 @@ module Agama
         lines = ["# AutoYaST compatibility reference"]
 
         top_level = description.elements.select(&:top_level?)
+          .sort_by(&:short_key)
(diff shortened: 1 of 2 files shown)


service/lib/tasks/autoyast.rb
```diff
@@ -86,6 +90,20 @@ def elements_table(elements)
         lines
       end
 
+      def unsupported_elements(elements)
+        lines = []
+        lines << "## Unsupported sections"
+        lines << ""
+        lines << "The following sections are not supported and we do not plan to " \
```
jreidinger, line 97: I would probably write it inpersonated like `and there are no plans to`

jreidinger (approved):

PR #2046, opened 2025-02-21
feat(live): ISO size reduction

lslezak:
## Live ISO improvements

Reduce the initrd and ISO size.

### Delete not needed packages

- Remove Mesa packages, we do not need OpenGL support
- Remove libyui-qt and Qt libs
- Remove python (installed just because of two simple scripts which we do not use)
  - To avoid removing it accidentally when some new dependency appears there is a dependency checl

## Additional improvements

Some more improvements were copied from the openSUSE Tumbleweed Live ISO:

- Use a better compression for initrd (xz instead of zstd with the best compression level)
- Removed some AMD GPU firmware (likely need only for OpenCL or GPU computing)
- Copied Lenovo X13 dracut configuration, might be usable also on other ARM64 machines

## Results (on x86_64)

- The total ISO size decreased from 692MB to 634MB (-58MB, -8%)
- The initrd size decreased from 87MB to 65MB (-22MB, -25%)

On different archs the numbers might be slightly different.

Note: XZ compression is more effective but requires more RAM and CPU for decompression, let's see how that will work on some PPC64 or ARM64 machines...

### DIFF
--- a/live/src/config.sh
+++ b/live/src/config.sh
@@ -164,6 +164,34 @@ rm -rf /usr/share/doc/packages/*
 du -h -s /usr/share/man
 rm -rf /usr/share/man/*
 
+# python is installed just because of few simple scripts /usr/sbin/bcache-status (bcache-tools package)
+# and /usr/sbin/xfs_protofile (xfsprogs package)
+# both are not used by libtorage-ng, yast2-storage-ng nor agama
+python=$(rpm -q --whatprovides python3-base || true)
+if [ -n "$python" ]; then
+  echo "Python package: $python"
+  python_deps=$(rpm -e "$python" 2>&1 || true)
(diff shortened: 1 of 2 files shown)


live/src/config.sh
```diff
@@ -164,6 +164,34 @@ rm -rf /usr/share/doc/packages/*
...
+    echo "Warning: Extra Python dependency detected:"
+    echo "$python_deps"
+    echo "Keeping the python packages installed"
+  fi
+fi
+
+# remove OpenGL support
+rpm -qa | grep ^Mesa | xargs rpm -e --nodeps
+
+# uninstall libyui-qt and libqt (pulled in by the YaST dependencies)
```
jreidinger, line 191: I think agama web ui should simply provide libyui backend, so it is not installed

jreidinger (approved):

PR #2053, opened 2025-02-23
service: Adjust management of default encryption settings

ancorgs:
## Problem

When using TPM-based encryption, Agama is proposing a separate /boot partition.

The root of the problem is that the default PBKDF (pbkdf2) was not being initialized for the planned root partition. That is a problem because BootRequirementsChecker relies on `Planned::Device#encryption_pbkdf` to decide whether a separate "/boot" is needed.

## Solution

Relax the criteria to ignore the default value of `encryption_pbkdf`. In fact, the whole area would deserve an important overhaul. But discriminating based on the encryption type instead of the encryption method is good enough for now.

## Testing

Manually tested that this fixes the TPM case.

### DIFF
--- a/service/lib/agama/storage/config_solvers/encryption.rb
+++ b/service/lib/agama/storage/config_solvers/encryption.rb
@@ -53,10 +53,7 @@ module Agama
 
           encryption = config.encryption
           encryption.method ||= default_encryption.method
-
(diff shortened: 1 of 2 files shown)


joseivanlopez (commented):
The fix looks ok. I am only missing a changelog entry and maybe some unit test?

joseivanlopez (approved):

PR #2054, opened 2025-02-24
Two minor visual tweaks at the partitions form

ancorgs:


ancorgs:
Maybe the size was not originally added because this sentence was ambiguous: "Add new partition of /dev/sda (60 GiB)". It could be the size of the disk or the size of the partition.

But I hope it will not be ambiguous in the context of the whole form.

whole

joseivanlopez (approved):

PR #2061, opened 2025-02-24
web: Only severe issues prevent installation

ancorgs:
## Problem

Agama distinguish two kind of issues: errors (which prevent the installation) and warnings.

The UI was making no difference between both. As a consequence, non-critical issues detected during storage probing made it impossible to install using the UI, although it was possible to progress using the CLI (`agama install).

Initial screen with both real errors and warnings.

all

Blocking situation after fixing the real errors.

warnings

## Solution

To follow the same criteria than the rest of Agama, now the UI only consider issues with severity Error.


imobachgs (approved):
LGTM

Issue #2047, opened 2025-02-22
black screen during installation


gcomes:
After I start agama, if I leave the main screen unattended for 10 minutes, the screen becomes black and there is no way to make it work again. Pressing any key, moving the mouse, even hitting ctrl-alt-f1 does not have any effect. Agama can still be accessed remotely, if I know the root password, but locally no interaction is possible anymore.
I have tested in several scenarios:
1) start agama and when you have to select the product, do nothing. After 10 min -> black screen.
2) start agama and make the installation process last longer by selecting all the installable patterns. If the machine is old enough (and therefore slower), the installation process takes more than 10 min. After 10 min from the last screen interaction, while still installing, the screen goes black.
3) start agama and perform the installation. When the installation completes, and you arrive at the reboot screen, if you don't hit the restart button immediately and wait, the screen goes black (always after 10 min from the last interaction).

PR #2063, opened 2025-02-24
fix(live): disable screen saver

joseivanlopez:
## Problem

Using the live iso, the screen becomes black after ten 10 minutes and there is no way to make it work again, see https://github.com/agama-project/agama/issues/2047. 

## Solution

Configure icewm and X11 to disable dpms (Display Power Management Signaling) and screen saver.

### DIFF
--- a/live/root/.xinitrc
+++ b/live/root/.xinitrc
@@ -1,1 +1,7 @@
+# disable Display Power Management Signaling
+xset -dpms
+
+# disable screen saver
+xset s off
+
 icewm-session -c /etc/icewm/preferences.yast2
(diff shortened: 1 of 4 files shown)


ancorgs (approved):
LGTM

lslezak:
The `xset` command is missing on the Live ISO, please add it in the `.kiwi` file (package "xset").

live/root/.xinitrc
```diff
@@ -1 +1,7 @@
 icewm-session -c /etc/icewm/preferences.yast2
```
lslezak, line 1: This `icewm-session` call is actually blocking, the `xset` calls are not executed. Only when exiting the X session...

After moving the `icewm-session` to the end it works fine, `xset q` reports "DPM is Disabled".

PR #2069, opened 2025-02-25
fix(live): Run shellcheck in CI to find errors in the scripts

lslezak:
## Problem

- There might be typos  (like in #2029) or mistakes in the Live scripts.

## Solution

- Run `shellcheck` in the CI and fail if any error is reported
- Because shellcheck reports lots of warning and some of them are false positives let's run it in two stages:
  - The first run checks only for errors, if an error is found then the CI fails
  - If no error is found it runs it in full mode with warnings enabled. But in this case the CI always passes.
- Later we should fix the warnings and silence the false positives and make the check to fail also with warnings.

## Testing

- Tested manually and in GitHub Actions
- The first run [found an error](https://github.com/agama-project/agama/actions/runs/13516640787/job/37766570369#step:6:26) and failed
- After fixing the errors (there is actually one more), then it [reports several warnings](https://github.com/agama-project/agama/actions/runs/13516732009/job/37766805780#step:6:31) but the CI succeeds

### DIFF
--- a/live/config-cdroot/fix_bootconfig.aarch64
+++ b/live/config-cdroot/fix_bootconfig.aarch64
@@ -27,7 +27,7 @@ if [ -d "$dst/boot/grub2/themes/SLE" ]; then
   theme="SLE"
 else
   theme="openSUSE"
-end
+fi
 
 #
 # Create grub.cfg
--- a/live/src/config.sh
+++ b/live/src/config.sh
@@ -93,10 +93,14 @@ echo 'add_dracutmodules+=" agama-logging "' > /etc/dracut.conf.d/10-agama-loggin
 # FIXME: remove when the module is included in the default driver list in
 # in /usr/lib/dracut/modules.d/90kernel-modules/module-setup.sh, see
 # https://github.com/openSUSE/dracut/blob/7559201e7480a65b0da050263d96a1cd8f15f50d/modules.d/90kernel-modules/module-setup.sh#L42-L46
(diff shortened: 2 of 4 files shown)


live/src/config.sh
```diff
@@ -93,10 +93,14 @@ echo 'add_dracutmodules+=" agama-logging "' > /etc/dracut.conf.d/10-agama-loggin
 # FIXME: remove when the module is included in the default driver list in
 # in /usr/lib/dracut/modules.d/90kernel-modules/module-setup.sh, see
 # https://github.com/openSUSE/dracut/blob/7559201e7480a65b0da050263d96a1cd8f15f50d/modules.d/90kernel-modules/module-setup.sh#L42-L46
-if [ -f /lib/modules/*/kernel/drivers/usb/host/xhci-pci-renesas.ko* ]; then
```
lslezak, line 96: See https://www.shellcheck.net/wiki/SC2144 for more details about the problem.

live/config-cdroot/fix_bootconfig.aarch64
```diff
@@ -27,7 +27,7 @@ if [ -d "$dst/boot/grub2/themes/SLE" ]; then
   theme="SLE"
 else
   theme="openSUSE"
-end
+fi
```
lslezak, line 30: Switching between ruby and shell too often is not good... :worried:

imobachgs (approved):

teclator (approved):
LGTM

PR #2070, opened 2025-02-25
feat(web): show filesystem labels

joseivanlopez:
The labels of the file-systems were not visible in the parts of the UI used to select a disk and/or a partition. This adds them

| Before | After |
|-|-|
| uno-old | uno-new |
| dos-old | dos-new |
| tres-old | tres-new |

### DIFF
--- a/web/src/components/storage/AddExistingDeviceMenu.tsx
+++ b/web/src/components/storage/AddExistingDeviceMenu.tsx
@@ -31,6 +31,9 @@ import {
   MenuToggleElement,
   MenuToggle,
   Divider,
+  Split,
+  Flex,
+  Label,
 } from "@patternfly/react-core";
 import { MenuHeader } from "~/components/core";
 import MenuDeviceDescription from "./MenuDeviceDescription";
@@ -90,7 +93,16 @@ export default function AddExistingDeviceMenu() {
               description={<MenuDeviceDescription device={device} />}
               onClick={() => modelHook.addDrive(device.name)}
             >
-              {deviceLabel(device)}
+              <Split hasGutter>
+                {deviceLabel(device)}
+                <Flex columnGap={{ default: "columnGapXs" }}>
+                  {device.systems.map((s, i) => (
+                    <Label key={i} isCompact>
+                      {s}
+                    </Label>
+                  ))}
+                </Flex>
+              </Split>
             </DropdownItem>
           ))}
         </DropdownGroup>
--- a/web/src/components/storage/MenuDeviceDescription.tsx
+++ b/web/src/components/storage/MenuDeviceDescription.tsx
@@ -21,8 +21,12 @@
  */
 
 import React from "react";
-import { Stack, Split, Label } from "@patternfly/react-core";
-import { typeDescription, contentDescription } from "~/components/storage/utils/device";
+import { Stack, Flex, Split, Label } from "@patternfly/react-core";
+import {
+  typeDescription,
+  contentDescription,
+  filesystemLabels,
+} from "~/components/storage/utils/device";
 import { StorageDevice } from "~/types/storage";
 
 /**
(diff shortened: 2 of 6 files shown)


web/src/components/storage/PartitionPage.tsx
```diff
@@ -491,10 +493,17 @@ type PartitionDescriptionProps = {
...
+
   return (
     <Split hasGutter>
-      <span>{partition.description}</span>
-      <span>{deviceSize(partition.size)}</span>
+      <SplitItem>{deviceSize(partition.size)}</SplitItem>
+      <SplitItem>{partition.description}</SplitItem>
+      {label && (
+        <SplitItem>
+          <Label isCompact>{label}</Label>
```
dgdavid, line 504: As discussed at G1 room, we should consider to use [outlined labels](https://www.patternfly.org/components/label#outlined-labels) instead. Reason: avoid confusion among widgets. At the time of writing, storage selectors and gray labels are quite similar in appearance.

web/src/components/storage/AddExistingDeviceMenu.tsx
```diff
@@ -90,7 +93,16 @@ export default function AddExistingDeviceMenu() {
               description={<MenuDeviceDescription device={device} />}
               onClick={() => modelHook.addDrive(device.name)}
             >
-              {deviceLabel(device)}
+              <Split hasGutter>
```
dgdavid, line 96: I'd try removing this Split and placing the deviceLabel inside the Flex unless the spacing using there is too small

web/src/components/storage/MenuDeviceDescription.tsx
```diff
@@ -38,13 +42,13 @@ export default function MenuDeviceDescription({ device }: { device: StorageDevic
...
         <span>{contentDescription(device)}</span>
       </Split>
-      <Split hasGutter>
-        {device.systems.map((s, i) => (
-          <Label key={i} isCompact>
+      <Flex columnGap={{ default: "columnGapXs" }}>
+        {filesystemLabels(device).map((s, i) => (
+          <Label key={i} variant="outline" isCompact>
             {s}
           </Label>
```
dgdavid, line 49: If I'm not wrong I have seen this code before. Something to extract if we finally took some time for refactoring DriveEditor and friends.

web/src/components/storage/PartitionPage.tsx
```diff
@@ -491,10 +499,18 @@ type PartitionDescriptionProps = {
...
     <Split hasGutter>
-      <span>{partition.description}</span>
-      <span>{deviceSize(partition.size)}</span>
+      <SplitItem>{partition.description}</SplitItem>
+      {label && (
+        <SplitItem>
+          <Label isCompact variant="outline">
+            {label}
+          </Label>
+        </SplitItem>
```
dgdavid, line 512: What about still using the Flex with custom size approach?

web/src/components/storage/AddExistingDeviceMenu.tsx
ancorgs: I wanted some extra space between the first part and the list of systems.

web/src/components/storage/MenuDeviceDescription.tsx
ancorgs: Not now, but definitely in the mid-term that's absolutely needed.

web/src/components/storage/PartitionPage.tsx
ancorgs: Well, I like the separation given by Split here.

web/src/components/storage/AddExistingDeviceMenu.tsx
dgdavid: Ok

web/src/components/storage/PartitionPage.tsx
dgdavid: So, go ahead. 

As commented in IRC, it should not be a question of taste but consistence and correct abstractions to improve maintenance and those things that we use to make so important in the backend side :wink:

PR #2072, opened 2025-02-25
web: Synchronous probing when registering or choosing product

ancorgs:
## Problem

There is a race condition that leads to the interface reporting that there are no available devices after changing the product or registering it.

Moreover, even if that race condition is not hit, the interface is displaying the previous storage configuration. That's wrong because changing the product or registering actually resets the storage configuration at the backend (the registration case is debatable, but that's what happens now).

## Solution

Do a synchronous probing and invalidate all the storage queries when probing has finished. With that, the UI contains an up-to-date representation of the backend reality.

## Testing

Tested manually


joseivanlopez (approved):
LGTM

PR #2073, opened 2025-02-25
fix(web): adjust product title

dgdavid:
It uses the [CSS `clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp) function to set the font size, allowing it to adjust fluidly based on the viewport size. This is a workaround that requires further testing and refinement. It should also be applied to the other headings.

## Screenshots

|Width | Before | After |
|-|-|-|
|1024  | Screen Shot 2025-02-25 at 13 41 40 | Screen Shot 2025-02-25 at 13 41 46 |
|1280 | Screen Shot 2025-02-25 at 13 40 48 | Screen Shot 2025-02-25 at 13 40 56 |
|1600|Screen Shot 2025-02-25 at 13 41 30 | Screen Shot 2025-02-25 at 13 41 24 |

### DIFF
--- a/web/src/assets/styles/index.scss
+++ b/web/src/assets/styles/index.scss
@@ -191,6 +191,10 @@
 .pf-v6-c-masthead {
   --pf-v6-c-masthead__logo--Width: fit-content;
   --pf-v6-c-masthead--PaddingBlock: var(--pf-t--global--spacer--sm);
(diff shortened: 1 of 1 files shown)


ancorgs (approved):
LGTM

PR #2074, opened 2025-02-25
feat(web): edit file system label

joseivanlopez:
Allow setting the file system label for the new partitions.

Note: the label field only appears if the file system is going to be created (i.e., an existing file system is not reused).

* https://bugzilla.suse.com/show_bug.cgi?id=1237165
* https://jira.suse.com/browse/AGM-122

Screenshot From 2025-02-26 06-38-22

### DIFF
--- a/web/src/components/storage/PartitionPage.tsx
+++ b/web/src/components/storage/PartitionPage.tsx
@@ -80,6 +80,7 @@ type FormValue = {
   mountPoint: string;
   target: string;
   filesystem: string;
+  filesystemLabel: string;
   sizeOption: SizeOptionValue;
   minSize: string;
   maxSize: string;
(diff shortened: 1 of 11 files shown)


web/src/components/storage/PartitionPage.tsx
```diff
@@ -625,6 +632,26 @@ function FilesystemSelect({
...
+function FilesystemLabel({ id, value, onChange }: FilesystemLabelProps): React.ReactNode {
+  const change = (v: string) => /^[\w-_.]*$/.test(v) && onChange(v);
+
+  return (
+    <TextInput
+      id={id}
+      aria-label={_("File system label")}
+      value={value}
+      onChange={(_, v) => change(v)}
+      className={sizingStyles.w_100OnLg}
```
dgdavid, line 650: I don't think this is necessary. While we do use it occasionally, the goal is to keep its use to a minimum. In fact, we actually removed a few today.

From my testing, adding the w_100 here doesn't seem to make any difference.

imobachgs (approved):

Issue #2077, opened 2025-02-25
init scripts not executed


gcomes:
I'm using a profile where I put pre scripts and init scripts.
The pre scripts do get executed, but the init scripts do not.
After I complete the installation and reboot, in /var/log/agama-installation/scripts/pre
I can see the pre scripts plus the .log, .err and .out files.
In /var/log/agama-installation/scripts/init I can only see the init scripts
but not the .log, .err and .out files. Also the action the init scripts were supposed to
do, did not happen, which make me think that the init scripts were not executed.
The package agama-scripts is installed but I'm not sure agama-scripts.service gets enabled for the first boot. 
I don't see it listed in systemd-presets-common-SUSE.
If, after the installation is complete and before I reboot the machine, I manually
enable the agama-scripts service by creating the link in /etc/systemd/system/default.target.wants,
on reboot the agama-scripts service get started but it fails.
It turns out that in /usr/libexec/agama-scripts.sh the she-bang '#!/bin/bash' is missing.
So, if before the reboot after the installation, I fix agama-scritps.sh and enable the agama-script service,
the init scripts are executed.

imobachgs:
Partially fixed with #2080. We need to add `agama-scripts` to the SUSE preset or, alternatively, add our [own preset](https://github.com/agama-project/agama/pull/2081).

imobachgs:
And fixed with #2083.

PR #2081, opened 2025-02-25
fix(rust): create a systemd preset file

imobachgs:
Proposal to add a systemd presets file. See https://github.com/agama-project/agama/issues/2077.


dgdavid (approved):

imobachgs:
Replaced with #2077.

PR #2083, opened 2025-02-26
fix(ruby): enable the agama-scripts service if installed

imobachgs:
Ensure the `agama-scripts` service is enabled so that "init" scripts run on the first boot. This one is the second part of #2080.

It enables this service in the same way AutoYaST enables `autoyast-initscripts`.

## Tests

* Manually tested.
* It looks like a better alternative to #2081.

### DIFF
--- a/service/lib/agama/storage/finisher.rb
+++ b/service/lib/agama/storage/finisher.rb
@@ -260,8 +260,8 @@ module Agama
         # So this method just tries to enable the service.
         def enable_init_scripts
           # systemctl will return 1 if the service does not exist.
-          Yast::Execute.locally!(
-            "systemctl", "--root", "/mnt", "enable", "agama-scripts",
+          Yast::Execute.on_target!(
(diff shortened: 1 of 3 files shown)


service/lib/agama/storage/finisher.rb
```diff
@@ -241,10 +241,30 @@ def label
...
 
         def run
+          run_post_scripts
+          enable_init_scripts
+        end
+
+      private
+
+        # Run the post scripts
+        def run_post_scripts
```
joseivanlopez, line 251: Not related to this PR, but I think all these things about scripts are misplaced in a storage finisher.

joseivanlopez (approved):
LGTM

service/lib/agama/storage/finisher.rb
imobachgs: Yes, there are quite some things that do not belong here. We will find better places as we continue moving stuff to Rust.

PR #2089, opened 2025-02-27
feat(ui): show volume groups

joseivanlopez:
Show the volume groups used for the installation and their logical volumes.

Note: The volume groups and logical volumes cannot be edited yet. 

Screenshot From 2025-03-07 11-35-51

Screenshot From 2025-03-06 16-05-56

### DIFF
--- a/web/src/components/storage/DeviceHeader.tsx
+++ b/web/src/components/storage/DeviceHeader.tsx
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) [2025] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Content } from "@patternfly/react-core";
+
+export type DeviceHeaderProps = {
+  title: string;
+  children: React.ReactNode;
+};
+
+export default function DeviceHeader({ title, children }: DeviceHeaderProps) {
+  const [txt1, txt2] = title.split("%s");
+
+  return (
+    <Content component="h4">
+      {txt1} {children} {txt2}
+    </Content>
+  );
+}
--- a/web/src/components/storage/utils/configEditor.tsx
+++ b/web/src/components/storage/utils/configEditor.tsx
@@ -23,11 +23,16 @@
 // @ts-check
 
 import React, { useId, useRef, useState } from "react";
+import { useNavigate } from "react-router-dom";
+import { useVolume } from "~/queries/storage";
+import * as partitionUtils from "~/components/storage/utils/partition";
 import { Icon } from "../../layout";
 import {
   Menu,
   MenuContainer,
   MenuContent,
+  MenuItem,
+  MenuItemAction,
   MenuToggle,
   MenuToggleProps,
   MenuToggleElement,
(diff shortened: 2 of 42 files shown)


ancorgs:
Just for the record. We agreed to proceed with this model, but we will need to revisit it in the future since it has some consistency problems.

For example, drives have a property `name` that contains the device name of the real device referenced by the drive. On the other hand, volume groups also have a `name` property but it refers to the name of the volume group. That's not only confusing, but it will be conflictive in the future when we introduce the possibility to reuse an existing volume group.

web/src/components/storage/utils/volumeGroup.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+// @ts-check
```
dgdavid, line 23: No longer needed, it was a directive for checking types written in JSDoc syntax

ancorgs:
@dgdavid Pointed that is inconsistent to have all those components grouped at `utils/configEditor.tsx`. They should all live in their own separate file at the same level than the other components.

@joseivanlopez Can you fix that? Just to make sure to not set a bad precedent.

web/src/types/storage/model.ts
```diff
@@ -0,0 +1,38 @@
...
+ * find current contact information at www.suse.com.
+ */
+
+import * as apiModel from "~/api/storage/types/config-model";
+
+type Drive = apiModel.Drive;
+
+type LogicalVolume = apiModel.LogicalVolume;
+
+interface VolumeGroup extends Omit<apiModel.VolumeGroup, "targetDevices" | "logicalVolumes"> {
```
dgdavid, line 29: Why interface?

web/src/components/storage/utils/configEditor.tsx
```diff
@@ -0,0 +1,144 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+// @ts-check
```
dgdavid, line 23: No longer needed.

dgdavid (commented):
I didn't review it in depth, especially the components since I expect more changes on them and I need more time to think on them. However, I think we can go ahead.

web/src/types/storage/model.ts
dgdavid: I see. I used to do something like

```ts
type VolumeGroup = Omit<apiModel.VolumeGroup, "targetDevices" | "logicalVolumes">  & {
  targetDevices: Drive[];
  logicalVolumes: LogicalVolume[];
}
```

joseivanlopez: Do we want to use type intersections for extending a type? AFAIS, extending interfaces is preferred:

joseivanlopez: JFTR: This article is also interesting https://blog.logrocket.com/types-vs-interfaces-typescript/.

About the video, it defends type over interface to avoid the interface declaration merging and to be able to use unions. IMHO, the discussion is not about using either interfaces or types, but when to use one or the other.

For exteding types, all the resources I have read recommend interfaces. It seems that React also uses the very same approach as the code proposed in this PR (i.e., extend on Omit). Of course, type alias makes a lot of sense for other use cases like unions, etc. IMHO, I would use interface when we need to extend a type and type alias for the rest.

dgdavid: The video was just a summary, but fine with me to use both if there are reasons to do so.

dgdavid: JFTR, due the little advantage of using both I was most all for using just one and keep things simple, but don't want to invest too much time in _code style_ discussions since they are not too much useful right now.

web/src/components/storage/DeviceHeader.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default function DeviceHeader({ title, children }: DeviceHeaderProps) {
+  const [txt1, txt2] = title.split("%s");
+
+  return (
+    <h4>
+      <span>{txt1}</span>
+      {children}
+      <span>{txt2}</span>
+    </h4>
```
dgdavid, line 38: A comment for my self: check if going for a more PatternFly way does not break the layout here. I mean, 

```tsx
<Content component="h4">{txt} {children} {txt2}</Content>
```

web/src/types/storage/model.ts
joseivanlopez: As discussed in a meeting, we found useful to use interface for extending types. We will add a new section in the development resources of the agama documentation, https://agama-project.github.io/docs/devel

dgdavid (approved):
Go ahead, thanks.

PR #2090, opened 2025-02-27
Make the sles_enhanced_base pattern mandatory for SLES

teclator:
## Problem

Firewalld should be installed in SLES product but the package is provided by the enhanced base pattern and it is not visible to the user as well it is not mandatory.

- https://trello.com/c/Ksvh989B/4288-sle16-verify-firewall-status

## Solution

Make the sles_enhanced_pattern mandatory for the SLES product.

## Testing

- *Tested manually*

### DIFF
--- a/products.d/sles_160.yaml
+++ b/products.d/sles_160.yaml
@@ -66,11 +66,11 @@ software:
 
   mandatory_patterns:
     - base_traditional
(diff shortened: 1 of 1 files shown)


teclator:
It has been replaced by https://github.com/agama-project/agama/pull/2116

PR #2097, opened 2025-02-28
Set BLS env for storage-ng according to the product selection

mchf:
## Problem

Folowup of https://github.com/yast/yast-storage-ng/pull/1396/files#diff-d0c8ea3d63135679325c6c61b13d11f85edc06bbb141d983037c55846699e125R143

The above proposal intoduced BLS boot proposal. Only limitation was target architecture, but it doesn't fit for conservative products like SLE for now.

## Solution

Added product based condition for seting nev variable controling BLS selection in storage-ng. BLS is proposed only for tumbleweed (including Slowroll)

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/service/lib/agama/config.rb
+++ b/service/lib/agama/config.rb
@@ -217,6 +217,13 @@ module Agama
       data.dig("storage", "lvm") || false
     end
 
+    # Boot strategy for the product
+    #
+    # @return [String, nil]
+    def boot_strategy
+      data.dig("storage", "boot_strategy") || nil
+    end
+
   private
 
     def mandatory_path?(path)
--- a/service/lib/agama/config.rb
+++ b/service/lib/agama/config.rb
@@ -221,7 +221,7 @@ module Agama
     #
     # @return [String, nil]
     def boot_strategy
-      data.dig("storage", "boot_strategy") || nil
+      data.dig("storage", "boot_strategy")
     end
 
   private
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -125,6 +125,8 @@ module Agama
 
         @product = new_product
 
+        ENV["ENV_NO_BLS_BOOT"] = "yes" if use_bls?(id)
+
         update_issues
         true
       end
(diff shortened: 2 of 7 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -152,6 +154,8 @@ def probe
       def initialize_target
         # create the zypp lock also in the target directory
         ENV["ZYPP_LOCKFILE_ROOT"] = TARGET_DIR
+        # controls if grub-bls should be used. We enable it only for some products.
+        ENV["ENV_NO_BLS_BOOT"] = "no"
```
ancorgs, line 158: I like the approach of injecting the env variable. But I wonder why this is at software manager instead of the storage one.

service/lib/agama/software/manager.rb
```diff
@@ -443,6 +447,14 @@ def locale=(locale)
       # @return [Logger]
       attr_reader :logger
 
+      # Decides whether grub-bls should be used for given product
+      def use_bls?(id)
+        return true if id =~ /Tumbleweed/
+        return true if id =~ /Slowroll/
```
ancorgs, line 453: Cool. But I would make this a setting at the Agama product file (you know, those under `product.d`), instead of hardcoding product names in the code.

mchf: sure ...  this is in proof of concept only now.

no problem when we're on same continent this time it is not  problem to move between countries ;-)

service/lib/agama/config.rb
```diff
@@ -217,6 +217,13 @@ def lvm?
       data.dig("storage", "lvm") || false
     end
 
+    # Boot strategy for the product
+    #
+    # @return [String, nil]
+    def boot_strategy
+      data.dig("storage", "boot_strategy") || nil
```
ancorgs, line 224: NP: that `|| nil` is redundant. Hash#dig already returns nil if not found.

ancorgs (commented):
This looks good. Especially if it's already tested manually. Please:

- Add a changelog
- Re-target it to the branch "after-release-beta2"

mchf:
> This looks good. Especially if it's already tested manually. 

Just for record. The manual test was:
i had no system suitable for UEFI boot, so I:
- ran patched agama installer, picked Slowroll (or non Tumbleweed product later on)
- custom patched yast-storage-ng, added log message to see whether it sees updated YAST_NO_BLS_BOOT env variable - nothing for Slowroll or "1" for non Tumbleweed products - in boot strategy proposal method

ancorgs (approved):
LGTM

Issue #2092, opened 2025-02-27
Allow unsigned repodata for local repository / offline media


lkocman:
Hello team

product-composer currently can't produce offline agama image with signed repodata.
So pretty much all installations will hit following issue. And we'll have to workaround it in openQA.

Can we please make repo signature for local repository  optional or disabled for the time being?
We definitely want to keep the signature verification for remote repos https/http.

![Image](https://github.com/user-attachments/assets/d4f95868-1de5-464d-bacb-1cf9788e5a65)

Thank you

lkocman:
I hope that we can resolve the signing part until later Milestones.

imobachgs:
We could handle that at the callback level, discarding those messages for local repositories. Do you have an ISO we could try, please?

kobliha:
Or we could have a cmdline option to ignore such stuff and build the media with that option? Such option could be reused for other cases in the future.

imobachgs:
Yes, an option could work to enable such a behavior, but the implementation would be at callback level. Anyway, I assume some kind of time pressure here, right?

imobachgs:
We should not check for signatures in the repository included in the media. However, we should keep the check for other local media, right? I am talking, for instance, about the packages DVD.

lkocman:
Our image is buid here https://build.opensuse.org/package/show/openSUSE:Leap:16.0/000productcompose.all

g7:
I think https://github.com/openSUSE/product-composer/pull/18 should fix the issue that caused this bug, at least I get a signed repomd now (but from my home project, so agama still rightfully complains)

imobachgs:
We have merged the changes, so if you use the version in [systemsmanagement:Agama:Devel](https://build.opensuse.org/project/show/systemsmanagement:Agama:Devel), everything should be fine. Please, let me know if it still complains.

PR #2098, opened 2025-02-28
feat(ruby): disable signature checking for dir:// repositories

imobachgs:
Temporarily disable signature checking for dir:// because with the new product composer it is not possible to sign the repositories yet. It will be fixed in the future and we should remove this code then.

See #2092 for further information.

## Testing

* Add a unit test.
* Manually tested.

### DIFF
--- a/service/lib/agama/software/callbacks/signature.rb
+++ b/service/lib/agama/software/callbacks/signature.rb
@@ -56,6 +56,11 @@ module Agama
         # @param repo_id [Integer] Repository ID. It might be -1 if there is not an associated repo.
         def accept_unsigned_file(filename, repo_id)
           repo = Yast::Pkg.SourceGeneralData(repo_id)
+
+          # Temporarily disable signature checking for local repositories.
+          # https://github.com/agama-project/agama/issues/2092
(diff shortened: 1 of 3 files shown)


service/lib/agama/software/callbacks/signature.rb
```diff
@@ -56,6 +56,11 @@ def setup
         # @param repo_id [Integer] Repository ID. It might be -1 if there is not an associated repo.
         def accept_unsigned_file(filename, repo_id)
           repo = Yast::Pkg.SourceGeneralData(repo_id)
+
+          # Temporarily disable signature checking for local repositories.
+          # https://github.com/agama-project/agama/issues/2092
+          return true if repo && repo["url"].start_with?("dir://")
```
lslezak, line 62: Libzypp allows `dir:/directory` URL (with a single slash), I'd make the test less strict.

```suggestion
          return true if repo && repo["url"].start_with?("dir:/")
```

imobachgs: Yes, I tested it locally. I have changed it to `dir:`, which is the scheme, and that's it.

lslezak (approved):

PR #2102, opened 2025-03-03
fix(live): Include the "nvme-cli" package (bsc#1238038)

lslezak:
## Problem

- The NVMF devices are not activated
- https://bugzilla.suse.com/show_bug.cgi?id=1238038

## Solution

- Include the `nvme-cli` package in the Live ISO, it contains a systemd service which automatically connects such devices.


imobachgs (approved):

lslezak:
Closing, the package is actually pulled in by the [service dependencies](https://github.com/agama-project/agama/blob/400d4bc76952cfcc090659c102cb4bf54881bc6d/service/package/gem2rpm.yml#L53).

Issue #430, opened 2023-02-16
Generate D-Bus API documentation


joseivanlopez:
Right now, documentation for D-Bus API is manually written in a markdown document. Documentation should be automatically generated from comments in the code.

@mvidner already started some experiment by following the PackageKit approach: https://github.com/yast/d-installer/pull/415.

PR #2103, opened 2025-03-03
Add `--api URL`  support to `agama profile`

mvidner:
## Problem

`agama profile` still executes its actions directly from the CLI client, not using the web service. Fix it to make it work like `agama --api $URL config ...` already does.

- https://trello.com/c/fqPGL2qe

`agama profile` command summary... it is a diverse bunch! (related, unplanned stub: https://trello.com/c/3aUupc2F/461-unify-config-and-profile-commands-in-the-cli )
```
profile_evaluate(jsonnet_path) -> json stdout
profile_validate(json_path) -> print whether valid, what problems
profile_autoyast(autoyast_url) -> json stdout
profile_import(url) -> performs config_load
  # oh fun!
  if url ends with .xml OR .erb OR /
    profile_autoyast
  else download the file and analyze content
    if is_script # starts with hash-bang and the program is not "jsonnet"
      EXECVE it # which means replace current process. OMG WTF inconsistent UX
    else if is_jsonnet
      profile_evaluate
    else if is_json
      cat
    else
      return Error
  # at this point we have a JSON file
  profile_validate # NOTE that invalid profile is just a message and we continue
  agama_config_load
```

## Solution

Added these 3 new web API paths corresponding 1-1 to the CLI commands:
- POST /api/profile/validate ?url= ?path= (or body)
- POST /api/profile/evaluate ?url= ?path= (or body)
- POST /api/profile/autoyast ?url=

But `agama profile import` is notable that it effectively includes an `agama
config load` which is implemented as multiple backend calls, so `import` stays at the front end and the back end only gets what's needed to make it work,
which is running a script:
- POST /api/profile/execute_script ?url= ?path= (or body)

(TODO explain the rest)

### Backend errors now include their causes 

Before:
> Anyhow(Backend call failed with status 400 and text '{"error":"Error: Could not evaluate the profile"}')
    
After:
> Anyhow(Backend call failed with status 400 and text '{"error":"Error: Could not evaluate the profile: Failed to read system's hardware information: Failed to read agama.libsonnet: No such file or directory (os error 2)"}')

### JSON validation errors are more readable now

<details>
<summary>
We used to include a debug representation of the validation error but reading the manual(!) shows there is a way to just point to the offending piece of JSON
</summary>

```console
$ cat ...
{  "product": {    "ID": "Tumbleweed"  }}
$ agama profile validate rust/agama-lib/share/examples/profile_tw_invalid.json
The profile is not valid. Please, check the following errors:
```

Before:
> * Additional properties are not allowed ('ID' was unexpected). ValidationError { instance: Object {"ID": String("Tumbleweed")}, kind: AdditionalProperties { unexpected: ["ID"] }, instance_path: JSONPointer([Property("product")]), schema_path: JSONPointer([Keyword("properties"), Property("product"), Keyword("additionalProperties")]) }
> * "id" is a required property. ValidationError { instance: Object {"ID": String("Tumbleweed")}, kind: Required { property: String("id") }, instance_path: JSONPointer([Property("product")]), schema_path: JSONPointer([Keyword("properties"), Property("product"), Keyword("required")]) }

After:
> * Additional properties are not allowed ('ID' was unexpected). /product
> * "id" is a required property. /product
</details>

## Testing

Unit tests: no

### Manual tests

...

### Added integration tests expecting a running web service.

Run like (part of .github/workflows/ci-integration-tests.yml)
```sh
(cd service/; bundle exec rspec --pattern 'test/integration/**_itest.rb')
```
- [x] agama profile validate
- [x] agama profile evaluate
- [x] agama profile autoyast ... turns out hard to test with the AY-special URLs as **none** of them work in my container
- [x] agama profile import
- [ ] agama --api ... profile ... hitting an opaque SSL handshake error :sob: 

## Screenshots

(textual screenshots coming up)

## Documentation

Documented the extended argument handling:

- autoinstallation/README.md adjusted
- https://github.com/agama-project/agama-project.github.io/pull/70

But we are actually still missing documentation for the `--api` option itself, except for [a mention in the blog](https://agama-project.github.io/blog/2025/01/21/Agama-11#remotecli). IMHO we should add it as part of the `config`/`profile` cleanup.


mvidner:
First commit, client code is missing, server implements only the validation part.

```sh
# save auth token
agama auth login
# try what surely works
curl -k -v -H "Authorization: Bearer $(cat $HOME/.local/agama/token)" https://localhost/api/l10n/config
# try out new code
curl -k -H "Authorization: Bearer $(cat $HOME/.local/agama/token)" 'https://localhost/api/profile/validate?path=/etc/os-release'
```

rust/agama-cli/Cargo.toml
```diff
@@ -18,10 +18,11 @@ tempfile = "3.13.0"
 nix = { version = "0.27.1", features = ["user"] }
 tokio = { version = "1.40.0", features = ["macros", "rt-multi-thread"] }
 async-trait = "0.1.83"
-reqwest = { version = "0.11", features = ["json"] }
+reqwest = { version = "0.12.8", features = ["json"] }
```
mvidner, line 21: BTW this was the only case where we required 2 different versions of a library.
I wonder if there is an automated check for such things

rust/agama-cli/src/lib.rs
```diff
@@ -153,14 +154,15 @@ async fn finish(manager: &ManagerClient<'_>, method: FinishMethod) -> anyhow::Re
     Ok(())
 }
 
+/// What does this do??
```
jreidinger, line 157: this code should show progress in case you call CLI when backend is busy...so it writes you that it is busy and show progress. https://github.com/agama-project/agama/pull/2103/files#diff-f6b6409f33ceb3754c4e1f79a4b62d89ee044f73cbf6ca4d29987a4021197fa5R175

rust/agama-cli/src/profile.rs
```diff
@@ -53,17 +46,18 @@ pub enum ProfileCommands {
...
     /// Evaluate a profile, injecting the hardware information from D-Bus
     ///
     /// For an example of Jsonnet-based profile, see
     /// https://github.com/openSUSE/agama/blob/master/rust/agama-lib/share/examples/profile.jsonnet
     Evaluate {
-        /// Path to jsonnet file.
-        path: PathBuf,
+        /// Jsonnet file, URL or path or `-` for standard input
+        // TODO can I declare CliInput here and have it shown nicely?
+        url_or_path: String,
```
jreidinger, line 60: as you properly have in TODO, it should be CliInput here.

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +70,245 @@ pub enum ProfileCommands {
...
-        .context(format!("Could not validate the profile {:?}", path))?;
-    match result {
-        ValidationResult::Valid => {
-            println!("{} The profile is valid.", style("\u{2713}").bold().green(),);
+/// TODO better name
+// Represents the ways user can specify the input on the command line
+// and passes appropriate representations to the web API
+enum CliInput {
+    Url(String),
+    Path(String),
```
jreidinger, line 83: I kind of expect PathBuf here.

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +70,245 @@ pub enum ProfileCommands {
...
-        .validate_file(path)
-        .context(format!("Could not validate the profile {:?}", path))?;
-    match result {
-        ValidationResult::Valid => {
-            println!("{} The profile is valid.", style("\u{2713}").bold().green(),);
+/// TODO better name
+// Represents the ways user can specify the input on the command line
+// and passes appropriate representations to the web API
+enum CliInput {
+    Url(String),
```
jreidinger, line 82: and same here with `url::Url` as type instead of plain string. String can be used for passing stuff around, but I kind of expect proper data type here.

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +70,245 @@ pub enum ProfileCommands {
...
+    Path(String),
+    Stdin,
+    // the full text as string
+    Full(String),
+}
+
+impl CliInput {
+    fn new(url_or_path: String) -> Self {
```
jreidinger, line 90: convention is to call such conversion methods `from_string` https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#general-conventions-[rfc-#430]

rust/agama-server/src/web/docs/profile.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl ApiDocBuilder for ProfileApiDocBuilder {
+    fn title(&self) -> String {
+        "Profile HTTP API".to_string()
+    }
+    fn paths(&self) -> Paths {
+        PathsBuilder::new()
+            .path_from::<crate::profile::web::__path_autoyast>()
+            .path_from::<crate::profile::web::__path_evaluate>()
+            .path_from::<crate::profile::web::__path_validate>()
```
mvidner, line 35: execute_script is missing

rust/agama-cli/src/profile.rs
mvidner: This turns out to be tricky. The **clap** CLI library does not seem to like it when argument parsing can fail.

imobachgs (commented):
I have only reviewed the client side, but it is looking really promising. Great work!

rust/agama-cli/src/profile.rs
jreidinger: interesting. I kind of expect that clap can do nice error report when you pass invalid URL.

rust/agama-cli/Cargo.toml
mvidner: Thanks, I did not know `cargo tree`. It does not do exactly what I want but it is a good start

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +69,244 @@ pub enum ProfileCommands {
         /// AutoYaST specific ones. Supported files are json, jsonnet, sh for Agama profiles and ERB, XML, and rules/classes directories
         /// for AutoYaST support.
         url: String,
-        /// Specific directory where all processing happens. By default it uses a temporary directory
+        /// Ignored, formerly the directory for intermadiate files.
```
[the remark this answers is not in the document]
mvidner: Right, now I have checked that the extra argument is in fact nowhere used or documented :+1:

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +69,244 @@ pub enum ProfileCommands {
...
-            let mut perms = std::fs::metadata(&tmp_profile_path)?.permissions();
-            perms.set_mode(0o750);
-            std::fs::set_permissions(&tmp_profile_path, perms)?;
-            let err = Command::new(&tmp_profile_path).exec();
-            eprintln!("Exec failed: {}", err);
-        }
-        FileFormat::Json => {
-            std::fs::rename(&tmp_profile_path, path.as_ref())?;
+            let api_url = format!("/profile/execute_script?url={}", url_string);
+            // FIXME: try out
```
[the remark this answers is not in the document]
mvidner: "It compiles but I still have to test that it actually works" :fearful:

rust/agama-cli/src/profile.rs
mvidner: Ah, here's an example of how to do value errors, https://docs.rs/clap/latest/clap/_cookbook/typed_derive/index.html

rust/agama-lib/src/error.rs
```diff
@@ -81,6 +84,6 @@ pub enum ProfileError {
     InputOutputError(#[from] io::Error),
     #[error("The profile is not a valid JSON file")]
     FormatError(#[from] serde_json::Error),
-    #[error("Error: {0}")]
+    #[error("Error: {0:#}")]
```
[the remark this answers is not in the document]
mvidner: added comment:

    // `#` is std::fmt "Alternate form", anyhow::Error interprets as "include causes"

products.d/install.sh
```diff
@@ -0,0 +1,19 @@
...
+set -eu
+# After building this part of Agama, install it so that it is ready for run time
+# This is used by agama-products.spec and testing-in-container.sh
+
+if [ "${1-}" = --system ]; then
+    SRCDIR=.
+    DESTDIR=""
+    datadir=/usr/share
+fi
+# DESTDIR=%{buildroot}
```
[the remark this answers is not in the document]
mvidner: No, it means to document the calling convention, I have improved the comment

rust/agama-server/src/profile/web.rs
```diff
@@ -0,0 +1,270 @@
...
+    let request_has_body = !script.is_empty() && script != "null";
+    query.check(request_has_body)?;
+    let script_string = match query.retrieve_profile()? {
+        Some(retrieved) => retrieved,
+        None => script,
+    };
+
+    // write to a temporary path and spawn in background and return
+    // May be long running
+    // FIXME: can agama+axum wait for completion and return output?
```
imobachgs, line 251: Yes, because this action blocks the whole server. You could give [tokio::process::Command](https://docs.rs/tokio/latest/tokio/process/struct.Command.html) a try. Alternatively, if it does not work, we could use a separate Tokio task.

rust/agama-cli/src/profile.rs
```diff
@@ -76,123 +69,256 @@ pub enum ProfileCommands {
...
+    let output = evaluate_client(client, url_or_path).await?;
+    println!("{}", output);
     Ok(())
 }
 
-async fn import(url_string: String, dir: Option<PathBuf>) -> anyhow::Result<()> {
+async fn import(client: BaseHTTPClient, url_string: String) -> anyhow::Result<()> {
+    // useful for store_settings
     tokio::spawn(async move {
         show_progress().await.unwrap();
```
imobachgs, line 236: This code will not work anymore because the D-Bus server is not accessible from here. We might need to rely on the HTTP API instead.

imobachgs (commented):
I have reviewed the server part now and we are almost there. IMHO the blocking issues are:

* As you mentioned, the `execute_script` blocks the whole server.
* The progress reporting will not work anymore when using the CLI from a remote system (as it relied on D-Bus). 

If you want, we could work on the second problem on a separate PR because it might imply some work.

rust/agama-cli/src/profile.rs
mvidner: Is that a recent change? Which PR?

imobachgs:
JFTR, it could be a good idea to offer a way to upload a local profile from the CLI:

```
agama profile import --api http://server.lan/api --local import my-profile.jsonnet
```

But. of course, that's totally out of scope for this PR.

rust/agama-cli/src/profile.rs
mvidner: Duh, you mean when working with the `--api` option, right

rust/agama-lib/src/profile.rs
```diff
@@ -30,11 +30,11 @@ use std::{
     process::Command,
 };
 use tempfile::{tempdir, TempDir};
-use url::Url;
+pub use url::Url;
```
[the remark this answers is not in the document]
mvidner: It was a hack to avoid adding `url` to agama-server/Cargo.toml, but it's fine to do just that

rust/agama-server/Cargo.toml
```diff
@@ -22,7 +22,7 @@ gettext-rs = { version = "0.7.1", features = ["gettext-system"] }
 regex = "1.11.0"
 macaddr = { version = "1.0", features = ["serde_std"] }
 async-trait = "0.1.83"
-axum = { version = "0.7.7", features = ["ws"] }
```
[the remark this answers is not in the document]
mvidner: Debugging leftover, deleting it.
`#[axum::debug_handler]` on a web handler function provides better compilation errors

rust/agama-cli/src/profile.rs
imobachgs: Yes, I meant from a remote system (`--api`). You cannot connect to the D-Bus server of the system you are installing.

rust/agama-server/src/profile/web.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+#[utoipa::path(
+    post,
+    path = "/validate",
+    context_path = "/api/profile",
+    params(ProfileQuery),
+    responses(
+        (status = 200, description = "Validation result", body = ValidationOutcome),
+        (status = 400, description = "FIXME some error has happened")
```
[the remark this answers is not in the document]
mvidner: Well, it turned out to be, minus the "FIXME", because I can't think of anything more specific

rust/agama-server/src/web/docs/profile.rs
mvidner: ... added

rust/agama-server/src/profile/web.rs
```diff
@@ -0,0 +1,291 @@
...
+        .context("Writing script text")?;
+    // close the file otherwise exec fails with ETXTBSY
+    let path = named_tempfile.into_temp_path();
+    let path = path.keep().context("Persisting script file")?;
+
+    let mut child = std::process::Command::new(&path)
+        .spawn()
+        .context(format!("Spawning script {:?}", path))?;
+
+    // Do not child.wait() for the script to finish,
```
imobachgs, line 283: As discussed in IRC, I am not sure about this behavior. I would expect to have always the same answer, no matter whether the call takes 50ms or 60ms. I would go for:

1) Waiting until it finishes but without blocking the server. That's why I mentioned using `tokio::process::Command` instead (but I have not tried if it works as expected). We could even move the script to a separate Tokio task.
2) Always return a 200 and move the script executation to a separate Tokio task.

We even discussed about having an API to run jobs. You would create a job by calling `POST /jobs`, which would return a job ID. So you can check the current status with `GET /jobs/1`. Of course, we could integrate this API with the events system too. But I would say it is out of scope.

imobachgs (approved):

PR #2113, opened 2025-03-05
feat(ruby): enable signature checking for `dir://` repositories

imobachgs:
Once the problem with the produce composer is fixed (see #2092), it is time to enable the signature
checking again for `dir://` repositories.


teclator (approved):
LGTM

jreidinger (approved):

PR #2115, opened 2025-03-05
fix(ruby): properly run postPartitioning scripts

imobachgs:
- **fix(ruby): use the correct scripts type**
- **docs(ruby): update changes file**


ancorgs (approved):
LGTM

PR #2118, opened 2025-03-05
feat: add support for some YaST-like URLs

imobachgs:
## Problem

Agama [Transfer's API](https://github.com/agama-project/agama/blob/master/rust/agama-lib/src/utils/transfer.rs) does not understand [YaST-specific URLs](https://github.com/yast/yast-installation/blob/master/doc/url.md).

It is impossible to handle URLs like `usb://OEMDRV/autoinst.xml`, `device:///autoinst.xml`, etc.

> [!WARNING]
> This PR is still a work in progress, but many cases seem to work fine.

## Solution

This PR aims to add partial support for these URLs. The following schemes are supported: `device:`, `usb:`, `label:`, `HD:`, `DVD:` and `cd:`. Support for well-known URLs (e.g., `file:`, `http:`, `https:`, `ftp:`, `nfs:`, etc.) is still implemented using CURL.

## Testing

- Added a new unit test
- Tested manually


rust/agama-cli/src/lib.rs
```diff
@@ -198,6 +202,22 @@ async fn allowed_insecure_api(use_insecure: bool, api_url: String) -> Result<boo
     }
 }
 
+pub fn download_file(url: &str, path: &PathBuf) -> Result<(), ServiceError> {
+    let mut file = fs::OpenOptions::new()
+        .create(true)
+        .truncate(true)
```
jreidinger, line 208: create already did truncatenation https://doc.rust-lang.org/std/fs/struct.File.html#method.create

rust/agama-cli/src/lib.rs
```diff
@@ -198,6 +202,22 @@ async fn allowed_insecure_api(use_insecure: bool, api_url: String) -> Result<boo
...
+        .write(true)
+        .mode(0o400)
+        .open(path)
+        .context(format!("Cannot write the file '{}'", path.display()))?;
+
+    match Transfer::get(&url, &mut file) {
+        Ok(()) => println!("File saved to {}", path.display()),
```
jreidinger, line 215: we probably should introduce in future some quiet and verbose options.

rust/agama-lib/src/utils/transfer/file_finder.rs
```diff
@@ -0,0 +1,121 @@
...
+        file_system: &FileSystem,
+        file_name: &str,
+        writer: &mut impl Write,
+    ) -> TransferResult<()> {
+        const DEFAULT_MOUNT_PATH: &str = "/run/agama/mount";
+        let default_mount_point = PathBuf::from(DEFAULT_MOUNT_PATH);
+        let mount_point = file_system
+            .mount_point
+            .clone()
+            .unwrap_or(default_mount_point);
```
jreidinger, line 72: note that here it construct unnecessary PathBuf. It will be more efficient to use here `unwrap_or_else` to create in only if needed.

rust/agama-lib/src/utils/transfer/file_finder.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        println!("Searching {} in {}", &file_name, &file_system.block_device);
```
jreidinger, line 74: is it debug print or intentional output? I think this can cause trouble if it is used e.g. from scripts or files url.

rust/agama-lib/src/utils/transfer/file_finder.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        if !file_system.is_mounted() {
+            Self::mount_file_system(&file_system, &mount_point)?;
+        }
+
+        let file_name = file_name.strip_prefix("/").unwrap_or(file_name);
+        let source = mount_point.join(&file_name);
+        let result = Self::copy_file(source, writer);
+
+        if !file_system.is_mounted() {
```
jreidinger, line 84: this basically means that even if previously file system was already mounted, it will be unmounted here, which can be surprising and buggy. What about using closure like guarantee for mount that will handle it like
```rust
let file_name = file_name.strip_prefix("/").unwrap_or(file_name);
let source = mount_point.join(&file_name);
let result = file_system.ensure_mounted(&mount_point, || Self::copy_file(source, writer))
```

rust/agama-lib/src/utils/transfer/file_finder.rs
```diff
@@ -0,0 +1,121 @@
...
+            Self::umount_file_system(&mount_point)?;
+        }
+
+        result
+    }
+
+    /// Mounts the file system on a given point.
+    ///
+    /// TODO: move this logic to the FileSystem struct.
```
jreidinger, line 93: yeah, my point :)

rust/agama-lib/src/utils/transfer/file_finder.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn mount_file_system(file_system: &FileSystem, mount_point: &PathBuf) -> TransferResult<()> {
+        std::fs::create_dir_all(mount_point)?;
+        let output = Command::new("mount")
+            .args([file_system.device(), mount_point.display().to_string()])
```
jreidinger, line 97: what about using by default `ro` to avoid accidental overwrite? and maybe for some devices you will need `-o loop`

rust/agama-lib/src/utils/transfer/file_finder.rs
imobachgs: No, it will not umount the file system. The `is_mounted()` retains the original value, so there is no danger.

rust/agama-lib/src/utils/transfer/file_systems.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    pub fn to_vec(&self) -> Vec<FileSystem> {
+        self.file_systems.clone()
+    }
+
+    /// Returns the file system with the given block device name.
+    ///
+    /// * `name`: block device name.
+    pub fn by_name(&self, name: &str) -> Option<&FileSystem> {
```
jreidinger, line 91: regarding naming I would probably use `with_` prefix instead of `_by` as by invokes to me sorted "by"...and with invokes for me objects "with".
Also what I do not like is inconsistent return type. So I would probably use
```
find_name
with_label
with_transport
```

rust/agama-lib/src/utils/transfer/file_systems.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub fn read_from_system() -> Vec<FileSystem> {
+        let lsblk = Command::new("lsblk")
+            .args([
+                "--output",
+                "KNAME,FSTYPE,MOUNTPOINTS,TRAN,LABEL",
+                "--pairs",
+                "--path",
+            ])
+            .output()
+            .unwrap();
```
jreidinger, line 143: this deserve at least todo for failures to report it more nicely.

rust/agama-cli/src/lib.rs
imobachgs: Yes, in the future we should think about that. The CLI needs more love.

rust/agama-lib/src/utils/transfer/file_systems.rs
```diff
@@ -0,0 +1,284 @@
...
+        Self::read_from_string(&output)
+    }
+
+    /// Turns the output of lsblk into a list of file systems.
+    pub fn read_from_string(lsblk_string: &str) -> Vec<FileSystem> {
+        let mut file_systems = vec![];
+        let mut parent_transport: Option<String> = None;
+        let re =
+            Regex::new(r#"KNAME="(.+)" FSTYPE="(.*)" MOUNTPOINTS="(.*)" TRAN="(.*)" LABEL="(.*)""#)
+                .unwrap();
```
jreidinger, line 154: yeah, that unwrap for regexp is what I hate on regexp API. Ideally it should be checked during compilation for static strings and avoid those runtime unwraps.

rust/agama-lib/src/utils/transfer/file_finder.rs
imobachgs: It is intentional. No, it is not a problem because the command now expects a destination file. Otherwise, it only tells you that it failed without further explanation.

rust/agama-lib/src/utils/transfer/handlers/yast.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct HdHandler {}
+
+impl HdHandler {
+    pub fn get(&self, url: Url, out_fd: &mut impl Write) -> TransferResult<()> {
+        let device = url.query_pairs().find(|(key, _value)| key == "devices");
+
+        let Some((_, device_name)) = device else {
+            return Err(TransferError::MissingDevice(url));
+        };
+        let device_name = device_name.strip_prefix("/dev/").unwrap();
```
jreidinger, line 43: what about allowing to just pass hd name instead of full "/dev/hda1" ? Code wise it will be easy:
```diff
        let device_name = device_name.strip_prefix("/dev/").unwrap_or(device_name);
```

rust/agama-lib/src/utils/transfer/handlers/yast.rs
```diff
@@ -0,0 +1,155 @@
...
+        let file_name = url.path();
+        if file_name.is_empty() {
+            return Err(TransferError::MissingPath(url));
+        }
+
+        let Some(label) = url.host_str() else {
+            return Err(TransferError::MissingLabel(url));
+        };
+
+        let file_systems = FileSystemsList::from_system().by_label(label);
```
jreidinger, line 66: btw with new name it will look here like
```suggestion
        let file_systems = FileSystemsList::from_system().with_label(label);
```

jreidinger (approved):

PR #2125, opened 2025-03-07
Implemented conversion of files section from AutoYaST xml format to Agama JSON

mchf:
## Problem

We want to (partially) support files section when converting old AY profile into Agama's new json based

## Solution

implemented files reader for Agama's AutoYaST converter.
For now only static file sources are supported no dynamic files allowed (files embedded into the profile)

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/autoyast/files_reader.rb
+++ b/service/lib/agama/autoyast/files_reader.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2024] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+
+# :nodoc:
+module Agama
+  module AutoYaST
+    # Builds the Agama "files" section from an AutoYaST profile.
+    class FilesReader
+      # @param profile [ProfileHash] AutoYaST profile
+      def initialize(profile)
+        @profile = profile
+      end
+
+      # Returns a hash corresponding to Agama "files" section.
+      #
+      # If there is no files-related information, it returns an empty hash.
+      #
+      # @return [Hash] Agama "files" section
+      def read
+        return {} if profile["files"].nil?
+
+        # files section contains list of file nodes.
+        # supported file sub-elements:
+        # - file_location (a file source)
+        # - file_path (this one seems to be the only mandatory,
+        #             a file destination or directory to be created if ends with '/')
+        # - file_owner
+        # - file_permissions
+        #
+        # unsupported for now:
+        # - file_content
+        # - file_script
+        return {}
+      end
+
+    private
+
+      attr_reader :profile
+
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


service/lib/agama/autoyast/files_reader.rb
```diff
@@ -0,0 +1,74 @@
...
+        # - file_permissions (permissions)
+        #
+        # unsupported for now:
+        # - file_content
+        # - file_script
+        files = profile.fetch_as_array("files")
+
+        files_json = files.reduce([]) do |res, file|
+          # Currently we support only url to file source
+          file["source"] = file.delete("file_location")
```
jreidinger, line 58: I think you get rust part wrong. There are two enum options:
- url ( here file location )
- content ( the file_content )

so in the end only file_script should not be supported.

service/lib/agama/autoyast/files_reader.rb
```diff
@@ -0,0 +1,74 @@
...
+          file["destination"] = file.delete("file_path")
+          file["permissions"] = file.delete("file_permissions")
+          file["owner"] = file.delete("file_owner")
```
jreidinger, line 61: same here, how it will look like? rust part now uses pair which I am not sure how it is serialized, but I can write own serializer that will use something like "root:root" string for serialization.

service/lib/agama/autoyast/files_reader.rb
```diff
@@ -0,0 +1,74 @@
...
+        # unsupported for now:
+        # - file_content
+        # - file_script
+        files = profile.fetch_as_array("files")
+
+        files_json = files.reduce([]) do |res, file|
+          # Currently we support only url to file source
+          file["source"] = file.delete("file_location")
+          file["destination"] = file.delete("file_path")
+          file["permissions"] = file.delete("file_permissions")
```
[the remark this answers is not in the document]
mchf: well in documentation / example was 444 ...

service/lib/agama/autoyast/files_reader.rb
mchf: in documentation is "tux.users"  so I can either create a tuple, array, hash,  or keep it as a string ....

service/lib/agama/autoyast/files_reader.rb
mchf: ok

mchf:
new screenshot for untagged enums:
files json

service/lib/agama/autoyast/files_reader.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    private
+
+      attr_reader :profile
+
+      def file_source(file)
+        return {} if file.nil? || file.empty?
+
+        if file.key?("file_location")
+          { "source": { "url": file["file_location"] } }
```
jreidinger, line 77: this is still not correct. "source should not be there. We use same format as scripts: https://agama-project.github.io/docs/user/unattended/scripts#scripts-definition

service/lib/agama/autoyast/files_reader.rb
```diff
@@ -0,0 +1,86 @@
...
+        # unsupported for now:
+        # - file_script
+        files = profile.fetch_as_array("files")
+
+        files_json = files.reduce([]) do |res, f|
+          file = file_source(f)
+
+          file["destination"] = f["file_path"] if f["file_path"]
+          file["permissions"] = f["file_permissions"] if f["file_permissions"]
+          file["owner"] = f["file_owner"] if f["file_owner"]
```
jreidinger, line 61: as mentioned we will need to translate dot into colon as agama will use "root:root" and autoyast according to docu I found uses "root.root" see https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-configuration-installation-options.html#id-1.7.5.2.38.11

jreidinger: as discussed with imo files won't have owner field and instead there will be user and group field.

jreidinger: see https://github.com/agama-project/agama/pull/2121/commits/939ed07e7392c97d0d03494c9617e4174c4a204d

mchf: so instead of ```{ "owner": "user:group" }``` add two json members ```{ "user": "user", "group": "group" }```? Ok, will update accordingly

mchf:
So after the last update it looks like:
files json

jreidinger:
I will merge it now and fine tune  and test it in my branch together.

PR #2129, opened 2025-03-07
Allow to delete mandatory filesystems

ancorgs:
## Problem

Agama does not allow to delete the "/" partition nor the drive containing it.

But that was only implemented to hide some limitations of the current UI. In fact, deleting "/" or any other mandatory partition should be allowed. Of course, a proposal without mandatory mount paths is not valid and cannot be computed. But that doesn’t mean the user cannot delete them temporarily in order to create it elsewhere.

## Solution

- Detect and properly report when mandatory partitions are missing.
- Adapt the UI to not panic (ie. resort to an empty state) when there are configuration errors that can actually be fixed by using the UI.
- Allow to delete all partitions (including "/").
- Adapt the "do not use" button of the drives to the new situation

## Screenshots

Now the trash-can icon is there for root as well...

avion1

But nothing dramatic happens if the user clicks on it.

avion2

## Testing

- Tested manually
- Unit tests adapted

### DIFF
--- a/web/src/queries/storage/config-model.ts
+++ b/web/src/queries/storage/config-model.ts
@@ -324,6 +324,27 @@ function unusedMountPaths(model: configModel.Config, volumes: Volume[]): string[
   return volPaths.filter((p) => !assigned.includes(p));
 }
 
+/*
+ * Pretty artificial logic used to decide whether the UI should display buttons to remove
+ * some drives. The logic is tricky and misplaced, but it is the lesser evil taking into
+ * account the current code organization.
+ *
+ * TODO: Revisit when LVM support is added to the UI.
+ */
+function hasAdditionalDrives(model: configModel.Config): boolean {
+  if (model.drives.length <= 1) return false;
+  if (model.drives.length > 2) return true;
+
+  // If there are only two drives, the following logic avoids the corner case in which first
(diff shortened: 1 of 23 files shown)


web/src/queries/storage/config-model.ts
```diff
@@ -91,6 +91,21 @@ function allMountPaths(drive: configModel.Drive): string[] {
   return drive.partitions.map((p) => p.mountPath).filter((m) => m);
 }
 
+// FIXME this logic does not belong here, but it is the lesser evil taking into account the
+// current code organization
+function mustStay(model: configModel.Config): boolean {
```
dgdavid, line 96: BIG NP: Setting aside that this is marked as a hack, for final code, I’d suggest using a method that expresses the opposite of "stay", such as isDeletable (https://www.collinsdictionary.com/dictionary/english/deletable), canBeDeleted, or something similar.

While mustStay is shorter and straightforward too, it doesn’t quite fit with other attributes like isBoot, isExplicitBoot, etc. It feels a bit less natural as an attribute name in this context, IMHO.

dgdavid: On the other hand, `if (mustStay) return` is definitely much more readable than `if (!canBeDeleted) return` :stuck_out_tongue_winking_eye:

PR #2139, opened 2025-03-10
Do only one automatic download retry

lslezak:
## Problem

- Related to https://github.com/agama-project/agama/pull/2117
- As discussed on the review, there should be just one automatic retry to report a failure quicker to the user

## Solution

- Change the number
- Additionally the logic has been changed so the retry count really means the retry count. Before it was actually the total amount of attempts, not just the retries.


dgdavid (approved):
Since changes are against "after-release-beta2" branch I guess the changelog entry for this is the same than linked PR in the description (and already merged against same branch).

Thus, let's go ahead with merge.

Thanks!

lslezak:
Yes, that's why I did add new changelog entry.

PR #2142, opened 2025-03-11
Hostname handling

teclator:
## Problem

Currently the backend and the UI does not allow to set the system hostname, and the kernel cmdline argument only allow to set the transient hostname.

- https://trello.com/c/V2GgJrWo/4349-5-agama-add-support-for-setting-the-hostname

## Solution

- Added a hostname specific model and API for getting/setting the transient and persistent hostname.
- Set the persistent hostname in dracut if the hostname option is given as linuxrc used to do.
- Allow to disable the manage of the hostname only in case the kernel cmdline SetHostname option is given
- By now we will not expose the hostname-mode or the via_dhcp option through the profile or config.. see below some notes and clarifications about it.

> I decided to not expose the by_dhcp option  by now until discussed, and maybe renaming it to `via_dhcp` would be better, or just use `mode` allowing to set it to {default|dhcp|none}.
> 
> From the doc page ... https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/NetworkManager.conf.html
> 
> hostname-mode
> 
> Set the management mode of the hostname. This parameter will affect only the transient hostname. If a valid static hostname is set, NetworkManager will skip the update of the hostname despite the value of this option. An hostname empty or equal to 'localhost', 'localhost6', 'localhost.localdomain' or 'localhost6.localdomain' is considered invalid.
> 
> default: NetworkManager will update the hostname with the one provided via DHCP or reverse DNS lookup of the IP address on the connection with the default route or on any connection with the property hostname.only-from-default set to 'false'. Connections are considered in order of increasing value of the hostname.priority property. In case multiple connections have the same priority, connections activated earlier are considered first. If no hostname can be determined in such way, the hostname will be updated to the last one set outside NetworkManager or to 'localhost.localdomain'.
> 
> dhcp: this is similar to 'default', with the difference that after trying to get the DHCP hostname, reverse DNS lookup is not done. Note that selecting this option is equivalent to setting the property 'hostname.from-dns-lookup' to 'false' globally for all connections in NetworkManager.conf.
> 
> none: NetworkManager will not manage the transient hostname and will never set it.
> 
> So, basically, if a static hostname is set then it will not be updated by any DHCP Ack. Therefore, I'm even considering to not expose it.
> 
> In linurxc we had the SetHostname option for disabling set the hostname via DHCP. Currently the kernel cmdline argument hostname=test only sets the transient hostname, but this PR add the linuxrc behavior setting the persistent hostname which prevents the hostname to be modified via DHCP, but it also allows to pass the SetHostname=0 to set the hostname-mode to none disabling the manage of the transient hostname by NetworkManager. This option is what we wanted to offer in the interface, but as setting the persistent hostname is enough I think we can omit it by now.
> 
> With linuxrc was different, the persistent hostname can be modified via DHCP if not disable the DHCP hostname.

## Testing

- *Tested manually*


rust/agama-lib/src/hostname/http_client.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) [2024] SUSE LLC
```
dgdavid, line 1: NP: 2025

rust/agama-lib/src/hostname/model.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Represents a Hostname
+#[derive(Clone, Debug, Serialize, Deserialize, Default, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct HostnameSettings {
+    #[serde(rename = "transient", skip_serializing_if = "String::is_empty")]
+    pub hostname: String,
+    #[serde(rename = "static")]
+    pub static_hostname: String,
+    pub by_dhcp: bool,
```
dgdavid: _Semantic question_: is it "by" or "enable"?

I mean. What setting this to true means? Using only the hostname set by DHCP or use the static hostname unless a _new one_ is set by DHCP? Or does it mean that the static hostname value was set by DHCP?

rust/agama-lib/src/hostname/store.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) [2024] SUSE LLC
```
dgdavid, line 1: NP 2025

rust/agama-lib/src/hostname/model.rs
```diff
@@ -26,9 +26,8 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, Serialize, Deserialize, Default, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct HostnameSettings {
-    #[serde(rename = "transient", skip_serializing_if = "String::is_empty")]
+    #[serde(rename = "transient")]
     pub hostname: String,
-    #[serde(rename = "static")]
+    #[serde(rename = "static", skip_serializing_if = "String::is_empty")]
     pub static_hostname: String,
-    pub by_dhcp: bool,
```
dgdavid: Most probably I miss the reason for this commit

rust/agama-lib/src/hostname/model.rs
teclator: I decided to not expose it by now until discussed, but maybe via DHCP would be better, or just use mode allowing to set it to {default|dhcp|none}.

From the doc page ... https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/NetworkManager.conf.html

> **hostname-mode**
> 
> Set the management mode of the hostname. This parameter will affect only the transient hostname. **If a valid static hostname is set, NetworkManager will skip the update of the hostname despite the value of this option**. An hostname empty or equal to 'localhost', 'localhost6', 'localhost.localdomain' or 'localhost6.localdomain' is considered invalid.
> 
> default: NetworkManager will update the hostname with the one provided via DHCP or reverse DNS lookup of the IP address on the connection with the default route or on any connection with the property hostname.only-from-default set to 'false'. Connections are considered in order of increasing value of the hostname.priority property. In case multiple connections have the same priority, connections activated earlier are considered first. If no hostname can be determined in such way, the hostname will be updated to the last one set outside NetworkManager or to 'localhost.localdomain'.
> 
> dhcp: this is similar to 'default', with the difference that after trying to get the DHCP hostname, reverse DNS lookup is not done. Note that selecting this option is equivalent to setting the property 'hostname.from-dns-lookup' to 'false' globally for all connections in NetworkManager.conf.
> 
> none: NetworkManager will not manage the transient hostname and will never set it.

So, basically, if a static hostname is set then it will not be updated by any DHCP Ack. Therefore, I'm even considering to not expose it.

In linurxc we had the SetHostname option for disabling set the hostname via DHCP. Currently the kernel cmdline argument **hostname=test** only sets the transient hostname, but this PR add the linuxrc behavior setting the persistent hostname which prevents the hostname to be modified via DHCP, but it also allows to pass the SetHostname=0 to set the hostname-mode to **none** disabling the manage of the transient hostname by NetworkManager. This option is what we wanted to offer in the interface, but as setting the persistent hostname is enough I think we can omit it by now. 

With linuxrc was different, the persistent hostname can be modified via  DHCP if not disable the DHCP hostname.

dgdavid: 😵

rust/agama-lib/src/hostname/model.rs
teclator: What commented above and what I want to discuss

rust/agama-lib/src/hostname/model.rs
```diff
@@ -0,0 +1,33 @@
...
+// find current contact information at www.suse.com.
+
+//! Implements a data model for Hostname configuration.
+
+use serde::{Deserialize, Serialize};
+
+/// Represents a Hostname
+#[derive(Clone, Debug, Serialize, Deserialize, Default, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct HostnameSettings {
```
jreidinger, line 28: as this is part of utoipa schema, I think it would be useful to have documentation for its elements including special values like what means empty static hostname

rust/agama-lib/src/hostname/store.rs
```diff
@@ -0,0 +1,49 @@
...
+            hostname_client: HostnameHTTPClient::new(client),
+        })
+    }
+
+    pub async fn load(&self) -> Result<HostnameSettings, ServiceError> {
+        self.hostname_client.get_config().await
+    }
+
+    pub async fn store(&self, settings: &HostnameSettings) -> Result<(), ServiceError> {
+        self.hostname_client.set_config(settings).await?;
```
jreidinger, line 46: because hostname client already returns `Result<(), ServiceError>` then simple `self.hostname_client.set_config(settings).await` without next line is enough

rust/agama-lib/src/store.rs
```diff
@@ -71,6 +74,7 @@ impl Store {
     pub async fn load(&self) -> Result<InstallSettings, ServiceError> {
         let mut settings = InstallSettings {
             bootloader: Some(self.bootloader.load().await?),
+            hostname: Some(self.hostname.load().await?),
```
jreidinger, line 77: store part is missing...that means if you do `agama config load` with hostname, then `agama config show` will miss it ( I know it as I did identical mistake :)

rust/agama-server/src/hostname/web.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Clone)]
+struct HostnameState<'a> {
+    client: HostnameClient<'a>,
+}
+
+/// Sets up and returns the axum service for the hostname module.
+pub async fn hostname_service() -> Result<Router, ServiceError> {
+    let dbus = zbus::Connection::system().await?;
+    let client = HostnameClient::new(dbus).await?;
```
jreidinger, line 45: does client even need dbus parameter when it always use system bus? I think it at least deserve here explanation why it always use system bus instead of agama one passed as parameter like other services did

rust/agama-lib/src/hostname/client.rs
```diff
@@ -0,0 +1,69 @@
...
+// To contact SUSE LLC about this file by physical or electronic mail, you may
+// find current contact information at www.suse.com.
+
+//! Implements a client to access Hostnamed D-Bus API related to hostname management.
+
+use zbus::Connection;
+
+use crate::{error::ServiceError, hostname::model::HostnameSettings, proxies::Hostname1Proxy};
+
+/// Client to connect to Agama's D-Bus API for Hostname management.
```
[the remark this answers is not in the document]
teclator: yep, the comment is wrong, I didn't updated the comment in this case, and mainly copied bootloader client as skeleton, so, a leftover

rust/agama-server/src/hostname/web.rs
teclator: we can ommit it, just modified the pass of the dbus parameter of the hostname_service but not the hostname client and as commented above used bootloader as skeleton, but as you commented in this case we do not need to pass the dbus zbus connnection.

rust/agama-lib/src/hostname/client.rs
```diff
@@ -51,16 +51,23 @@ impl<'a> HostnameClient<'a> {
...
     pub async fn set_config(&self, config: &HostnameSettings) -> Result<(), ServiceError> {
         let settings = self.get_config().await?;
-        if settings.hostname != config.hostname {
-            self.hostname_proxy
-                .set_hostname(config.hostname.as_str(), false)
-                .await?;
+
+        // order is important as otherwise the transient hostname could not be set in case the
+        // static one is not empty
+        if let Some(config_static_hostname) = config.static_hostname.clone() {
```
jreidinger, line 57: this clone looks quite strange in let condition. maybe it should be cloned later if needed?

jreidinger (approved):

jreidinger (approved):

Issue #2140, opened 2025-03-10
/var/log/agama-installation permission


gcomes:
The pre/post scripts may contain sensible information, it would be appropriate to make by default /var/log/agama-installation not world readable

imobachgs:
Hi! It should be fixed in #2148.

Thanks!

PR #2148, opened 2025-03-12
feat(ruby): collect the logs at the end of the installation

imobachgs:
## Problem

Agama does not copy the logs to the installed system, so they are lost. Additionally, the /var/log/agama-installation permissions are way too open.

## Solution

* Save the logs into the installed system.
* Set the proper permissions (700) to the /var/log/agama-installation folder (#2140).

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -258,10 +258,6 @@ module Agama
     # @param method [HALT, POWEROFF, STOP, REBOOT]
     # @return [Boolean]
     def finish_installation(method)
-      logs = collect_logs(path: "/tmp/var/logs/")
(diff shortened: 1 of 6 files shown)


dgdavid (commented):
There is an orphan test failing at manager now.

https://github.com/agama-project/agama/blob/5b8933e24fd02bb6d5a14a70ca6f73e0dc00a4ed/service/test/agama/manager_test.rb#L237-L240

dgdavid (approved):

dgdavid (approved):
Thanks for catching the wrong target branch. Not it looks better :)

PR #2153, opened 2025-03-13
feat: allow selecting individual packages

imobachgs:
This pull request implements support for selecting individual packages during the installation.

```json
{
  "software": {
    "patterns": ["gnome_desktop"],
    "packages": ["vim"],
  }
}
```

## Links

* Trello: https://trello.com/c/Ondb9itc/

## To improve

When a package does not exist, the problem is logged but not reported. We could handle this problem in a separate PR (or even a separate PBI) because we might need a way to handle a) errors in the `PackagesProposal`.


imobachgs:
@lslezak I need your help here :wink: I wonder whether it is better:

* Add the packages to the proposal (using the `SetResolvables` method). The solver does not run after that, but we can fix that. Deselecting packages is not supported this way, but I'd like to know whether we need/want that.
* Follow a similar approach to what we do with patterns: have a SetUserPatterns method at D-Bus level. This one is more complex, but it won't be hard to implement it.

We can start with a) and move to b) later. However, I would like to have the point of view of a software management expert.

lslezak:
I do not know the details but I'd probably use the same mechanism as for selecting the mandatory packages from the product configuration. To be consistent.

imobachgs:
> I do not know the details but I'd probably use the same mechanism as for selecting the mandatory packages from the product configuration. To be consistent.

Good point. Then I will continue with the current implementation that tuses the proposal's SetResolvables.

Thanks!

rust/agama-lib/src/software/client.rs
```diff
@@ -189,6 +191,28 @@ impl<'a> SoftwareClient<'a> {
...
+            USER_RESOLVABLES_LIST,
+            ResolvableType::Package,
+            names.as_slice(),
+            true,
+        )
+        .await?;
+        Ok(())
+    }
+
+    pub async fn user_selected_packages(&self) -> Result<Vec<String>, ServiceError> {
```
teclator, line 211: why not just selected_packages?

imobachgs: Because they are selected by the user, not by the product definition itself. And it is consistent with `user_selected_patterns`.

teclator: Ok, there is already user_selected_patterns and selected_patterns methods and this is the equivalent for the first one.

teclator (approved):
LGTM

PR #2160, opened 2025-03-14
Allow to delete mandatory filesystems

ancorgs:
This is a re-incarnation of #2129, that was automatically closed by Github.

## Problem

Agama does not allow to delete the "/" partition nor the drive containing it.

But that was only implemented to hide some limitations of the current UI. In fact, deleting "/" or any other mandatory partition should be allowed. Of course, a proposal without mandatory mount paths is not valid and cannot be computed. But that doesn’t mean the user cannot delete them temporarily in order to create it elsewhere.

## Solution

- Detect and properly report when mandatory partitions are missing.
- Adapt the UI to not panic (ie. resort to an empty state) when there are configuration errors that can actually be fixed by using the UI.
- Allow to delete all partitions (including "/").
- Adapt the "do not use" button of the drives to the new situation

## Screenshots

Now the trash-can icon is there for root as well...

avion1

But nothing dramatic happens if the user clicks on it.

avion2

## Testing

- Tested manually
- Unit tests adapted

### DIFF
--- a/service/lib/agama/storage/config_checkers/filesystems.rb
+++ b/service/lib/agama/storage/config_checkers/filesystems.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2025] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "agama/storage/config_checkers/base"
+require "agama/storage/proposal_settings_reader"
+require "yast/i18n"
+
+module Agama
+  module Storage
+    module ConfigCheckers
+      # Class for checking the boot config.
+      class Filesystems < Base
+        include Yast::I18n
+
+        # Boot config issues.
+        #
+        # @return [Array<Issue>]
+        def issues
+          [missing_paths_issue].compact
+        end
+
+      private
+
+        # @return [Issue, nil]
+        def missing_paths_issue
+          missing_paths = required_paths.reject { |p| configured_path?(p) }
+          return if missing_paths.empty?
+
+          error(
+            format(
+              # TRANSLATORS: %s is a path like "/" or a list of paths separated by commas
+              n_(
+                "A separate file system for %s is required.",
+                "Separate file systems are required for the following paths: %s",
+                missing_paths.size
+              ),
+              missing_paths.join(", ")
+            ),
+            kind: :required_filesystems
+          )
+        end
+
+        # @return [Boolean]
+        def configured_path?(path)
+          filesystems.any? { |fs| fs.path?(path) }
+        end
+
+        # @return [Array<Configs::Filesystem>]
+        def filesystems
+          @filesystems ||= storage_config.filesystems
+        end
+
+        # return [Array<String>]
+        def required_paths
+          volumes.select { |v| v.outline.required }.map(&:mount_path)
+        end
+
+        # @return [Array<Agama::Storage::Volume>]
+        def volumes
+          @volumes ||= VolumeTemplatesBuilder.new_from_config(product_config).all
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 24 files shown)


web/src/components/storage/DriveEditor.tsx
```diff
@@ -380,14 +380,15 @@ const SearchSelector = ({ drive, selected, onChange }) => {
 
 const RemoveDriveOption = ({ drive }) => {
   const driveModel = useDrive(drive.name);
+  const { hasAdditionalDrives } = useModel();
 
   if (!driveModel) return;
 
   const { isExplicitBoot, delete: deleteDrive } = driveModel;
 
+  if (!hasAdditionalDrives) return;
```
ancorgs, line 389: When `hasAdditionalDrives` no button is displayed. That's correct IMHO.

But the cases of `hasPv` and `isExplicitBoot` should actually display a disabled button with the corresponding explanation. I forgot to add a comment to the code explaining all that.

service/lib/agama/storage/config_checkers/filesystems.rb
```diff
@@ -0,0 +1,83 @@
...
+        def configured_path?(path)
+          filesystems.any? { |fs| fs.path?(path) }
+        end
+
+        # @return [Array<Configs::Filesystem>]
+        def filesystems
+          @filesystems ||= storage_config.filesystems
+        end
+
+        # return [Array<String>]
```
dgdavid, line 71:
```suggestion
        # @return [Array<String>]
```

web/src/components/storage/DriveEditor.test.tsx
```diff
@@ -204,7 +210,8 @@ describe("PartitionMenuItem", () => {
     const deleteRootButton = within(partitionsMenu).queryByRole("menuitem", {
```
dgdavid, line 210: NP: I'd go for a `getByRole` here now that you don't have to check that such a thing is not in the document.

web/src/components/storage/FixableConfigInfo.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { Alert, List, ListItem } from "@patternfly/react-core";
+import { n_ } from "~/i18n";
+import { useConfigErrors } from "~/queries/issues";
+
+const Description = ({ errors }) => {
+  return (
```
dgdavid, line 29: NP: I'm not sure if it should be a list when it is just one issue. But please, do not invest time on it now. This is mostly a comment for taking notes for future rounds of small improvements.

dgdavid (approved):
Despite adding a few NP comments, I consider this ready to merge if you'd like to do so

service/lib/agama/storage/config_checkers/filesystems.rb
```diff
@@ -0,0 +1,83 @@
...
+# find current contact information at www.suse.com.
+
+require "agama/storage/config_checkers/base"
+require "agama/storage/proposal_settings_reader"
+require "yast/i18n"
+
+module Agama
+  module Storage
+    module ConfigCheckers
+      # Class for checking the boot config.
```
joseivanlopez, line 29: Please, fix/adapt the documentation.

service/lib/agama/storage/config_checkers/filesystems.rb
```diff
@@ -0,0 +1,83 @@
...
+      class Filesystems < Base
+        include Yast::I18n
+
+        # Boot config issues.
```
joseivanlopez, line 33: Please, fix/adapt the documentation.

service/lib/agama/storage/config_checkers/filesystems.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+
+        # @return [Array<Configs::Filesystem>]
+        def filesystems
+          @filesystems ||= storage_config.filesystems
+        end
+
+        # return [Array<String>]
+        def required_paths
+          volumes.select { |v| v.outline.required }.map(&:mount_path)
```
joseivanlopez, line 73: There is a `#mandatory_paths` method https://github.com/agama-project/agama/blob/lvm-form/service/lib/agama/config.rb#L202.

dgdavid (changes requested):
I think it worth addressing @joseivanlopez comments at least.

web/src/components/storage/DriveEditor.tsx
ancorgs: Comments added

service/lib/agama/storage/config_checkers/filesystems.rb
ancorgs: Fixed

web/src/components/storage/DriveEditor.test.tsx
ancorgs: Done.

service/lib/agama/storage/config_checkers/filesystems.rb
ancorgs: Done.

service/lib/agama/storage/config_checkers/filesystems.rb
ancorgs: Fixed.

service/lib/agama/storage/config_checkers/filesystems.rb
ancorgs: Done.

ancorgs:
Updated to cover all important comments, I hope.

dgdavid (approved):

Issue #2144, opened 2025-03-11
agama-scripts.sh should directly execute scripts


jreidinger:
quote from slack: `Why does /usr/libexec/agama-scripts.sh use a dot instead of just doing execute of the scripts? The docs state somewhere that each script requires a shebang. Each one is marked as executable by agama.`

imobachgs:
Fixed with #2161. Thanks!

PR #2161, opened 2025-03-14
fix: improve handling of init scripts

imobachgs:
## Problem

Init scripts are executed properly. See #2144.

## Solution

- Run init scripts properly
- Make agama-scripts.sh idempotent
- Allow setting the scripts path with the SCRIPTS_DIR environment variable.
- Do not exit with an error if there are not scripts

## Testing

- _Tested manually_


teclator (approved):
LGTM, just miss a changelog entry

teclator:
Even better now!, thanks (LGTM)!

PR #2168, opened 2025-03-18
refactor(web): use checkbox instead of switch

dgdavid:
## Problem

During the revamp of certain web interface parts, some switches were incorrectly used where checkboxes should have been. This may have been due to the previous behavior, where these elements triggered changes or requests immediately. However, with the shift to a more traditional approach this is no longer the case.

## Solution

Replace the switches with checkboxes in the appropriate places.

Additionally, since the it had to be updated, the Encryption form has been restructured to match the layout used by the other forms. Now, dependent widgets are displayed only when needed, rather than changing their enabled state.

## Note for reviewers

Due to the restructuring of the Encryption form, the wording may still need improvements. Please pay close attention to the text and suggest any changes when reviewing the code.

## Testing

- Adapted unit tests
- Tested manually

## Screenshots

<em>Screenshots might be outdated due to review suggestions. They will not be updated unless the suggestions significantly change the final result.</em>

| Before | After |
|-|-|
| localhost_8080_ (17) | localhost_8080_ (16) |
| localhost_8080_ (18) | localhost_8080_ (14) |
| localhost_8080_ (19) |  localhost_8080_ (15) |


imobachgs (approved):

ancorgs (changes requested):
I agree with turning those switches into checkboxes, but not with the other changes introduced at the encryption form. I will elaborate in subsequent comments, but I wanted to block merging ASAP.

ancorgs (commented):
I don't get why the wording was slightly changed. I think the original was perfectly fine also after the change (I will try to add an screenshot later).

And, more than anything else, I don't understand why TPM is now so prominent and displayed even in systems without the appropriate hardware. I don't see the point and it would be the equivalent to displaying all the DASD and zFCP options at non-s390x system. What's the point if the user can do nothing to enable those hardware-dependent options?

ancorgs (commented):
This is how things would look like keeping the original wording and behavior.

When encryption is disabled in any system. The original wording mentioning FDE is kept and still works IMO. No unconditional mention to TPM (I don't see why it became so prominent on this PR).

nothing

When encryption is enabled in a system not supporting TPM-based unlocking.

dgdavid:
> I don't get why the wording was slightly changed. I think the original was perfectly fine also after the change (I will try to add an screenshot later).
> 
> And, more than anything else, I don't understand why TPM is now so prominent and displayed even in systems without the appropriate hardware. I don't see the point and it would be the equivalent to displaying all the DASD and zFCP options at non-s390x system. What's the point if the user can do nothing to enable those hardware-dependent options?

IF I understood how TPM works, it's not the same case. I mentioned TPM before displaying it to hint uses that such an option will be there, if applicable, as part of the encryption options.

Based on the, probably my misunderstanding, that you could have the hardware but not see the option until selecting a _compatible product_

In any case, feel free to undo the change

dgdavid:
> This is how things would look like keeping the original wording and behavior.
> 

Not the behaviour ;) Before the change these inputs where disabled instead of hidden ;)


> When encryption is disabled in any system. The original wording mentioning FDE is kept and still works IMO. No unconditional mention to TPM (I don't see why it became so prominent on this PR).
> 

Already explained. I'll revert the change as soon as I have the laptop at hand. Don't worry.

> nothing
> 
> When encryption is enabled in a system not supporting TPM-based unlocking.
> 
> something
> 
> When encryption is enabled in a system with the appropriate hardware and conditions to support TPM-based unlocking.
> 
> everything
> 
> That's basically how things worked before changing the switch into a checkbox... and I don't get why that change should imply all the others.

As said, take it easy. I'll revert changes as soon as possible.

ancorgs:
> Based on the, probably my misunderstanding, that you could have the hardware but not see the option until selecting a _compatible product_

Yes and no. The conditions to use TPM unlocking are (all must be met):

- The system is EFI (hardware related)
- The system has a compatible TPMv2 chip (hardware related)
- The package `fde-tools` is available (ie. whether the full repo is already accessible)

IMO, if we could make a difference between all those cases, it would make sense to display the hint if the first two conditions are met but the second fails. But since we do not know the exact details (we just have a boolean for the whole thing), mentioning TPM in all cases would imply a lot of cases in which it is just noise. Just to have an extra warning in the corner case of an online installation that is still not registered.

dgdavid:
Except for the mentioned extra warning, suggestion of reverting some of the changes at encryption settings can be found at #2209

PR #2123, opened 2025-03-06
feat(web): allow creation of LVM volume group

dgdavid:
Add page for creating a volume group.

localhost_8080__

PR #2177, opened 2025-03-19
feat(web): edit volume group

joseivanlopez:
Allow editing a volume group.


web/src/components/storage/LvmPage.tsx
```diff
@@ -37,47 +37,59 @@ import {
...
-  }, [model]);
+    if (volumeGroup) {
+      setName(volumeGroup.vgName);
+      const targetNames = volumeGroup.getTargetDevices().map((d) => d.name);
+      const targetDevices = allDevices.filter((d) => targetNames.includes(d.name));
+      setSelectedDevices(targetDevices);
+    } else if (model && !model.volumeGroups.length) {
+      setName("system");
+    }
+  }, [model, volumeGroup, allDevices]);
```
dgdavid, line 89: I'm having troubles with this useEffect while trying to write a basic unit test. No matter the solution for making the test work, I would like to find another way to set these default values in forms. 

As the project evolves, I find myself questioning the best approach more often. In this case, I’d prefer to wait until all the necessary information is available before rendering. I thought this might be achievable with Suspense queries, or perhaps with a feature from the Router (I can’t recall the name right now—I'll look it up later).

web/src/components/storage/LvmPage.tsx
```diff
@@ -37,47 +37,59 @@ import {
...
 
-function checkErrors(model: model.Model, vgName: string, targetDevices: StorageDevice[]): string[] {
-  const vgNameError = (): string | undefined => {
-    if (!vgName.length) return sprintf(_("Name is empty"), vgName);
+function vgNameError(
+  vgName: string,
+  model: model.Model,
+  volumeGroup?: model.VolumeGroup,
+): string | undefined {
+  if (!vgName.length) return sprintf(_("Name is empty"), vgName);
```
dgdavid, line 54: Locally I've removed the `sprintf` here, which is actually not needed. The change will be sent along the unit test.

That said, let me add something regarding these errors that I forgot to mention in the previous PR.

Although is true that we have agreed on going ahead and unify later how to render the form errors, these message errors can be improved right now bearing in mind that all rendered fields are considered mandatory.

So, I would go for something like 

```
Form has errors and cannot be sent

  * Provide a name for the volume group.
  * Select at least one disk for placing the volume group.
```

```
Form has errors and cannot be sent


* There is a volume group using "system" name already. Please, enter a different name
````

Or something like that. 

"Name is empty" and "No disk is selected"  are somehow useless and obvious from a helpful error message point of view.

Also, when we can provide even useful warning errors to users in the future, we could go for something like

```
Form has errors and cannot be sent


* There is a volume group using "system" name already. Please choose a different one or going to *edit system* instead.
````

Where *edit system* would be a link for switching to the edition form

File: web/src/components/storage/LvmPage.tsx (hunk 1 above)
Comment (dgdavid) [reply web/src/components/storage/LvmPage.tsx:89]: I suspect the useEffect is triggering more than needed because of model or allDevices, which are objects. I saw an useMemo somewhere, but still having problems when using the RTL `user.clear()`: it does work but immediately the name is updated to "system" again because of this useEffect.

In anycase, since it works manually I'll put a todo in the test in order to keep going.

File: web/src/components/storage/LvmPage.tsx (hunk 2 above)
Comment (joseivanlopez) [reply web/src/components/storage/LvmPage.tsx:54]: Sure, feel free to adapt it. I simply got inspired by the errors in the encryption form.

File: web/src/components/storage/LvmPage.test.tsx [hunk 3]
```diff
@@ -0,0 +1,300 @@
...
+  ...jest.requireActual("~/queries/storage"),
+  useAvailableDevices: () => mockUseAllDevices,
+  useDevices: () => [sda],
+}));
+
+jest.mock("~/hooks/storage/model", () => ({
+  ...jest.requireActual("~/hooks/storage/model"),
+  __esModule: true,
+  useVolumeGroup: (id: string) => (id ? mockRootVolumeGroup : null),
+  default: () => mockUseModel,
```
dgdavid, line 129: A _side effect_ of [_exporting these hooks as default_](https://github.com/agama-project/agama/pull/2123#discussion_r2006544612) :man_shrugging:

web/src/components/storage/LvmPage.test.tsx
```diff
@@ -0,0 +1,300 @@
...
+jest.mock("~/queries/storage", () => ({
+  ...jest.requireActual("~/queries/storage"),
+  useAvailableDevices: () => mockUseAllDevices,
+  useDevices: () => [sda],
+}));
+
+jest.mock("~/hooks/storage/model", () => ({
+  ...jest.requireActual("~/hooks/storage/model"),
+  __esModule: true,
+  useVolumeGroup: (id: string) => (id ? mockRootVolumeGroup : null),
```
dgdavid, line 128: Thought :thought_balloon: Mocking is becoming each time more ugly

web/src/hooks/storage/api-model.ts
```diff
@@ -20,23 +20,14 @@
...
  */
 
-import { useMemo } from "react";
 import { useQuery, useSuspenseQuery } from "@tanstack/react-query";
 import { configModelQuery } from "~/queries/storage/config-model";
 import { apiModel } from "~/api/storage/types";
 import { QueryHookOptions } from "~/types/queries";
 
-function useApiModel(options?: QueryHookOptions): apiModel.Config | null {
+export default function useApiModel(options?: QueryHookOptions): apiModel.Config | null {
```
dgdavid, line 28: No matter where default is placed. I still feeling is weird exporting hooks in that way

web/src/components/storage/LvmPage.tsx
dgdavid: I finally found the problem in testing: the lack of mocking the [_useVolumeGroup_](https://github.com/agama-project/agama/blob/85795832db34eb9fdb5f1018eb85f303505a5c20/web/src/hooks/storage/model.ts#L105-L111) hook despite of _useModel_ being mocked. Since it is exported and mocked as default, the _useVolumeGroup_ was using (or trying to use, actually don't know) the real implementation.

In any case, still thinking we should revisit, when possible, the approach for initializing the forms when it comes to edit, trying  to avoid these useEffects when possible since they make things a bit confusing and put the form at risk to be re-rendered more than needed. Specially when dependencies are objects.

web/src/components/storage/LvmPage.tsx
dgdavid: I have changed them.

> I simply got inspired by the errors in the encryption form.

As I said in the commit message, I'm aware of

> it remains inconsistent with the rest of the interface until the tone, style, and error handling can be unified.

dgdavid:
I've added a simple unit test file that has A LOT OF room for improvements. I really sorry for that. But as said in the commit message 

> The test is limited and could be improved, especially in terms of
mocking. It also highlights potential issues with form initialization
and the use of useEffect. However, it ensures the form works as expected
at a basic level.

Hopefully, we will have time for improve at some point in the future. Ideally after having time to rework forms and unify how they behaves on all aspects: initialization, validations, tone, etc.

web/src/hooks/storage/api-model.ts
joseivanlopez: I wrote `export default function` because  somehow I am following "this rule": if there is only one thing exported and it is exported by default (i.e., that cases in which the file name reveals what it is exporting) then I use `export default function` directly. Otherwise I export things at the end.

web/src/components/storage/LvmPage.test.tsx
joseivanlopez: I feel we should mock the queries instead of the hooks, but I don't know if it is possible.

joseivanlopez: And going further, I would like to simply mock the result of a http call (we have already talked about it).

web/src/components/storage/LvmPage.test.tsx
```diff
@@ -0,0 +1,300 @@
...
+    partitions: [sda1],
+    unpartitionedSize: 0,
+    unusedSlots: [{ start: 3, size: gib(2) }],
+  },
+  udevIds: ["ata-Micron_1100_SATA_512GB_12563", "scsi-0ATA_Micron_1100_SATA_512GB"],
+  udevPaths: ["pci-0000:00-12", "pci-0000:00-12-ata"],
+  description: "",
+};
+
+const mockSdaDrive: Drive = {
```
joseivanlopez, line 70: NP: I would use `model.Drive`. Otherwise I have to go back to the imports to know whether it means `apiModel.Drive` or `model.Drive`.

web/src/components/storage/LvmPage.test.tsx
```diff
@@ -0,0 +1,300 @@
...
+  ...jest.requireActual("~/hooks/storage/add-volume-group"),
+  __esModule: true,
+  default: () => mockAddVolumeGroup,
+}));
+
+jest.mock("~/hooks/storage/edit-volume-group", () => ({
+  ...jest.requireActual("~/hooks/storage/edit-volume-group"),
+  __esModule: true,
+  default: () => mockEditVolumeGroup,
+}));
```
joseivanlopez, line 146: In other tests we do this for mocking a default:

~~~
jest.mock("./ConfigEditorMenu", () => () => <div>config editor menu</div>);
~~~

Would the same work for these hooks?

~~~
jest.mock("~/hooks/storage/edit-volume-group", () => mockEditVolumeGroup);
~~~

web/src/components/storage/LvmPage.test.tsx
```diff
@@ -0,0 +1,300 @@
...
+      const moveMountPointsCheckbox = screen.getByRole("checkbox", {
+        name: /Move the mount points currently configured at the selected disks to logical volumes/,
+      });
+      const acceptButton = screen.getByRole("button", { name: "Accept" });
+
+      // Clear default value for name
+      await user.clear(name);
+      await user.type(name, "root-vg");
+      await user.click(sdaCheckbox);
+      // By default move move mount points should be checked
```
joseivanlopez, line 164: Repeated "move"

web/src/components/storage/LvmPage.tsx
```diff
@@ -132,7 +133,7 @@ export default function LvmPage() {
       <Page.Content>
         <Form id="lvmForm" onSubmit={onSubmit}>
           {errors.length > 0 && (
-            <Alert variant="warning" isInline title={_("Something went wrong")}>
+            <Alert variant="warning" isInline title={_("Check the following before continuing")}>
```
joseivanlopez, line 136: NP: Should it end with colons?

web/src/components/storage/LvmPage.test.tsx
dgdavid: I see. I didn't do it on purpose, but thinking about it I see more benefits importing it as it is now that name spacing. If I'm not wrong, If for some reason we end up with dead code in the file  holding "model", tree shaking will not be able to ignore it when making the build, since we are importing the whole moduleninstead of the needed parts. So, maybe it can be imported `as ModelDrive` or we could even choose only one of these module for importing it completely and the other in parts. I.e, whatever not prefixed comes always from the same module.

Of course, that approach is a bit harder to follow, and we’re already importing these modules fully in here and there.

web/src/components/storage/LvmPage.tsx
dgdavid: I don't think so. It's acting as a title.

> Avoid using full stops in titles and headings unless they are complete sentences. https://freelingual.com/full-stop-punctuation

Also having a look to the [SUSE documentation style guide](https://documentation.suse.com/style/current/html/docu_styleguide/art-styleguide.html) I've noticed they do not use it in titles neither

web/src/components/storage/LvmPage.test.tsx
dgdavid: I tried and didn't work, but maybe I was doing something wrong at some point. I'll try again later, but still being a pain in the ass mocking some modules in a way and others in another.

I can live we such a pain, though. But don't know why still adding more complexity to simple things.

dgdavid: Yes, it works. Like you in your above example, I was missing a function in the middle in my previous attempts :shrug: 

```diff
diff --git a/web/src/components/storage/LvmPage.test.tsx b/web/src/components/storage/LvmPage.test.tsx
index d3014b167..145166b95 100644
--- a/web/src/components/storage/LvmPage.test.tsx
+++ b/web/src/components/storage/LvmPage.test.tsx
@@ -139,11 +139,7 @@ jest.mock("~/hooks/storage/add-volume-group", () => ({
   default: () => mockAddVolumeGroup,
 }));
 
-jest.mock("~/hooks/storage/edit-volume-group", () => ({
-  ...jest.requireActual("~/hooks/storage/edit-volume-group"),
-  __esModule: true,
-  default: () => mockEditVolumeGroup,
-}));
+jest.mock("~/hooks/storage/edit-volume-group", () => () => mockEditVolumeGroup);
 
 describe("LvmPage", () => {
   describe("when creating a new volume group", () => {
```

However, I will keep mocking as it is now because, leaving aside the fact that the whole topic about these default exports is not clear, we have in a rush with other topics.

Feel free to change it before merging if you consider it really important.

web/src/components/storage/LvmPage.test.tsx
dgdavid: BTW, types/storage/model exports everything and then "model" and not "default". Definitely I'm fully lost with it comes to storage area and the way we do things there.

joseivanlopez (commented):
For me, everything looks good ;)

dgdavid (approved):
Approving and merging based on Iván's [comment](https://github.com/agama-project/agama/pull/2177#pullrequestreview-2710194278)

PR #2178, opened 2025-03-20
fix(live): use "agama download" to fetch the info file

### LINKED ISSUE #2176: inst.info should use `agama download` instead of curl
At https://github.com/agama-project/agama/blob/master/live/root/usr/bin/info-cmdline-conf.sh#L19 it downloads the info file with curl.

As we already have `agama download` supporting all those nice URIs, it would be great to use the same also for the info file that can be, e.g., in an OEMDRIVE-labeled partition or at a USB, etc.

imobachgs:
## Problem

YaST-like URLs are not supported for info files because we use CURL to retrieve the file.

Fixes #2176.

## Solution

Use "agama download" to get the info file.

## Testing

- Updated the integration test.

> [!WARNING]
> The tests fail because the `agama download` command tries to connect to the Agama HTTP server even if unnecessary. See https://github.com/agama-project/agama/pull/2192 for a fix.

### DIFF
--- a/.github/workflows/ci-live.yml
+++ b/.github/workflows/ci-live.yml
@@ -51,8 +51,8 @@ jobs:
       # disable unused repositories to have faster refresh
       run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && zypper ref
 
-    - name: Install Ruby development files
-      run: zypper --non-interactive install ShellCheck make
+    - name: Install development files
+      run: zypper --non-interactive install ShellCheck make agama-cli
 
     - name: Run shellcheck
       run: make shellcheck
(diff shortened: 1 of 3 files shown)


dgdavid (approved):

imobachgs:
@dgdavid Sorry, I would need another review. There were some strange conflicts in the translation files (I am not sure why) so I decided to rebase.

dgdavid (approved):

dgdavid:
> @dgdavid Sorry, I would need another review. There were some strange conflicts in the translation files (I am not sure why) so I decided to rebase.

No problem at all. Second review already done. It looks ready to be merged.

PR #2180, opened 2025-03-20
feat(auto): try to load profiles from predefined locations

imobachgs:
## Problem

Linuxrc & AutoYaST used to load profiles from predefined locations automatically. Some customers and partners depend on such behavior, so Agama should behave in the same way.

## Solution

If no URL was given, Agama tries to load the profile from the following locations:

- An `autoinst.jsonnet`, `autoinst.json` and `autoinst.xml` files located on a file system labeled as `OEMDRV`.
- An `autoinst.jsonnet`, `autoinst.json` and `autoinst.xml` files in the root file system.

## Testing

- *Tested manually*

## To do

- [x] Do not report an error if any of the fallback profiles do not work (see https://github.com/yast/yast-autoinstallation/pull/882).
- [x] Read profiles from the predefined locations.

### DIFF
--- a/autoinstallation/bin/agama-auto
+++ b/autoinstallation/bin/agama-auto
@@ -20,7 +20,7 @@ import_profile() {
     echo "Using the profile located at $url"
     agama profile import "$url" && return 0
   else
-    urls=("label://OEMDRV/autoinst.xml" "label://OEMDRV/autoinst.jsonnet" "label://OEMDRV/autoinst.json" "file:///autoinst.xml" "file:///autoinst.jsonnet" "file:///autoinst.json")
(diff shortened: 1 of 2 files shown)


autoinstallation/bin/agama-auto
```diff
@@ -1,35 +1,58 @@
...
 
-if [ -z "$url" ]; then
-  echo "no autoinstallation profile"
-  exit 0
-fi
+  if [ -n "$url" ]; then
+    echo "Using the profile located at $url"
+    agama profile import "$url" && return 0
+  else
+    urls=("label://OEMDRV/autoinst.xml" "label://OEMDRV/autoinst.jsonnet" "label://OEMDRV/autoinst.json" "file:///autoinst.xml" "file:///autoinst.jsonnet" "file:///autoinst.json")
```
kobliha, line 23: Nitpicking: I'd probably change the order and search for Agama-specific "profiles" first, then AutoYaST.

kobliha (approved):
I don't see any obvious errors or omissions.
Of course, except that the changes file entry is missing.

dgdavid (approved):

PR #2182, opened 2025-03-20
feat(live): Initial inst.dud=URI functionality

kobliha:
## Problem

*Customers and partners need a way to patch the Live Media with PTF or similar mechanism*

- https://jira.suse.com/browse/PED-3670
- https://jira.suse.com/browse/AGM-65

## Solution

There is a new agama-dud systemd service that runs after the agama-self-update so it overwrites anything that the self-update could have changed automatically. This service calls agama-dud script that goes over all inst.dud=... in /run/agama/cmdline.d/agama.conf, downloads those packages and force-installs/upgrades them with rpm.

The solution is "inspired" by the self-update functionality and other scripts and services from the project ;)

Things to solve later

- Everything is an RPM, no other type of DUD is supported yet
- There are no checks, no validation, etc.
- Logging should be added too

## Testing

Tested only manually by adding links to RPMs into /run/agama/cmdline.d/agama.conf and running the new script

## Example of the output

> Downloading DUD from https://download.opensuse.org/repositories/artwork:/wallpapers/openSUSE_Tumbleweed/noarch/propaganda00-0-2.73.noarch.rpm
> File saved to /tmp/agama-WT3vi0ryww.rpm
> warning: /tmp/agama-WT3vi0ryww.rpm: Header V3 DSA/SHA1 Signature, key ID 5b8ee893: NOKEY
> Preparing...                          ################################# [100%]
> Updating / installing...
>    1:propaganda00-0-2.73              ################################# [100%]

### DIFF
--- a/live/root/etc/systemd/system/agama-dud.service
+++ b/live/root/etc/systemd/system/agama-dud.service
@@ -11,7 +11,6 @@ After=agama-self-update.service
 # before starting the Agama servers so they use the new packages
 Before=agama-web-server.service
 Before=agama.service
-Before=x11-autologin.service
 # before interactive password services
 Before=live-password-dialog.service
 Before=live-password-systemd.service
@@ -19,9 +18,7 @@ Before=live-password-systemd.service
 [Service]
 Type=oneshot
 Environment=TERM=linux
-ExecStartPre=dmesg --console-off
 ExecStart=agama-dud
-ExecStartPost=dmesg --console-on
 TTYReset=yes
 TTYVHangup=yes
 StandardInput=tty
--- a/live/root/etc/systemd/system/agama-dud.service
+++ b/live/root/etc/systemd/system/agama-dud.service
@@ -1,0 +1,31 @@
+[Unit]
+Description=Agama DUD
+
+After=network-online.target
+
+# and after we process agama params like info which can contain password
+After=agama-cmdline-process.service
+# the DUD should replace wrong self_update
+After=agama-self-update.service
+
+# before starting the Agama servers so they use the new packages
+Before=agama-web-server.service
+Before=agama.service
+Before=x11-autologin.service
+# before interactive password services
+Before=live-password-dialog.service
+Before=live-password-systemd.service
+
+[Service]
+Type=oneshot
+Environment=TERM=linux
+ExecStartPre=dmesg --console-off
+ExecStart=agama-dud
+ExecStartPost=dmesg --console-on
+TTYReset=yes
+TTYVHangup=yes
+StandardInput=tty
+TimeoutStartSec=infinity
+
+[Install]
+WantedBy=default.target
--- a/live/root/usr/bin/agama-dud
+++ b/live/root/usr/bin/agama-dud
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# Experimental DUD funtionality for Agama
+#
+# Usage: inst.dud=URI [inst.dud=URI]
+#
+# This is a very simple version of the DUD functionality
+# - There can be multiple inst.dud=URI entries, but...
+# - All of them are considered to be RPMs
+# - The script does not check if downloading succeeded
+# - Or for any deps of the package
+# - Of whether installation/upgrade succeeded
+# - Does not restart itself if the script is updated
+
+# check if there is any DUD
+if ! grep -q "\b\(inst\|agama\)\.dud=.\+\b" /run/agama/cmdline.d/agama.conf; then
+  echo "No DUD present"
+  exit 0
+fi
+
+# Make sure that it does not refresh
+zypper modifyrepo --disable --all
+
+for URI in `sed -n 's/\(.*[[:space:]]\|^\)\(inst\|agama\)\.dud=\([^[:space:]]\+\).*/\3/p' /run/agama/cmdline.d/agama.conf`; do
+  echo "Downloading DUD from" $URI
+  TMPDUD=`mktemp /tmp/agama-XXXXXXXXXX.rpm`
+  agama download $URI $TMPDUD;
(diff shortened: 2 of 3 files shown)


live/root/usr/bin/agama-dud
```diff
@@ -0,0 +1,33 @@
...
+# check if there is any DUD
+if ! grep -q "\b\(inst\|agama\)\.dud=.\+\b" /run/agama/cmdline.d/agama.conf; then
+  echo "No DUD present"
+  exit 0
+fi
+
+# Make sure that it does not refresh
+zypper modifyrepo --disable --all
+
+for URI in `sed -n 's/\(.*[[:space:]]\|^\)\(inst\|agama\)\.dud=\([^[:space:]]\+\).*/\3/p' /run/agama/cmdline.d/agama.conf`; do
```
mchf, line 24: in a PBI about renaming install parameters we discussed that backward compatibility with "old" ```agama.``` prefix should be just temporal. So, there is no reason to introduce it here for new parameter.

It reminds me that this is approximately 27th place where is this pattern matching done. May be that we should think of some central processing of kernel parameters and generate e.g. a json for other componets.

live/root/usr/bin/agama-dud
```diff
@@ -0,0 +1,33 @@
...
+  echo "Downloading DUD from" $URI
+  TMPDUD=`mktemp /tmp/agama-XXXXXXXXXX.rpm`
+  agama download $URI $TMPDUD;
+  rpm --upgrade --force --verbose --hash --noverify --nodeps --excludedocs $TMPDUD
+  zypper --non-interactive install $TMPDUD
```
mchf, line 29: ```rpm --install --force``` should do it too and you don't need disable zypper repos

live/root/etc/systemd/system/agama-dud.service
```diff
@@ -0,0 +1,31 @@
...
+[Service]
+Type=oneshot
+Environment=TERM=linux
+ExecStartPre=dmesg --console-off
+ExecStart=agama-dud
+ExecStartPost=dmesg --console-on
+TTYReset=yes
+TTYVHangup=yes
+StandardInput=tty
+TimeoutStartSec=infinity
```
mchf, line 28: this seems to be a bit risky ... if this service blocks in some network call then starting whole installer can be blocked too due to that After / Before dependencies above

live/root/usr/bin/agama-dud
kobliha: Yes, that's what I basically mentioned at the daily today. But I simply wanted to stay compatible with the other code for now. Till someone says let's remove it (and replace it with a shared functionality like a lib).

live/root/etc/systemd/system/agama-dud.service
kobliha: This is copied from the self-update service written by @lslezak, so it's consistent.

It might be understood as risky, but we really need this to finish whatever it takes otherwise you'd have the system in an inconsistent way. And if some URL does not work / takes too long, then you don't want to continue anyway (and you should remove it from the command-line).

kobliha: A bit more info: DUD is a functionality for updating/extending the Live Media. It's also used for updating the Agama itself and of course other services that need to be updated. Blocking them till the update is done is a reasonable approach - better than restarting them later.

live/root/usr/bin/agama-dud
mchf: well i thought that conclusion was, that we will keep it only for those params which already were implemented .

Anyway it is not blocker for me.

live/root/etc/systemd/system/agama-dud.service
mchf: if we resign on best effort approach in running installer then fine with me.

live/root/etc/systemd/system/agama-dud.service
```diff
@@ -0,0 +1,31 @@
...
+
+# and after we process agama params like info which can contain password
+After=agama-cmdline-process.service
+# the DUD should replace wrong self_update
+After=agama-self-update.service
+
+# before starting the Agama servers so they use the new packages
+Before=agama-web-server.service
+Before=agama.service
+Before=x11-autologin.service
```
mchf, line 14: this dependency will have to be removed once https://github.com/agama-project/agama/pull/1868 get merged

live/root/usr/bin/agama-dud
```diff
@@ -0,0 +1,27 @@
...
+# - There can be multiple inst.dud=URI entries, but...
+# - All of them are considered to be RPMs
+# - The script does not check if downloading succeeded
+# - Or for any deps of the package
+# - Of whether installation/upgrade succeeded
+# - Does not restart itself if the script is updated
+
+# check if there is any DUD
+if ! grep -q "\b\(inst\|agama\)\.dud=.\+\b" /run/agama/cmdline.d/agama.conf; then
+  echo "No DUD present"
```
mchf, line 17: useless due to the ```dmesg --console-off``` in the service file.

live/root/etc/systemd/system/agama-dud.service
kobliha: This is my best effort :)

mchf (approved):

live/root/etc/systemd/system/agama-dud.service
kobliha: Addressed in the last commit

live/root/usr/bin/agama-dud
kobliha: Addressed in the last commit, dmesg --console-off is now gone
Let's make it simple

PR #2184, opened 2025-03-21
Add the aaa_base-extras package

imobachgs:
## Problem

The Agama ISO is not building anymore because it is missing `setup-systemd-proxy-env.path`. This path was part of `microos-tools`, but it as been [moved to aaa_base](https://github.com/openSUSE/microos-tools/blob/fa1a910b31274675516b29443923738d3738bba2/NEWS#L2).

## Solution

Add `aaa_base-extras`, which contains the path.

## Testing

- *Tested manually*


teclator (approved):
LGTM

PR #2186, opened 2025-03-21
feat(web): delete volume group

joseivanlopez:
* Allow deleting a volume group.
* The target devices of the deleted volume group are deleted too as side effect if they are not used.
* The config editor shows an alert if the config has not devices.

### DIFF
--- a/web/src/components/storage/ConfigEditor.tsx
+++ b/web/src/components/storage/ConfigEditor.tsx
@@ -21,15 +21,23 @@
  */
 
 import React from "react";
+import { _ } from "~/i18n";
 import { useDevices } from "~/queries/storage";
 import { useConfigModel } from "~/queries/storage/config-model";
 import DriveEditor from "~/components/storage/DriveEditor";
 import VolumeGroupEditor from "~/components/storage/VolumeGroupEditor";
-import { List, ListItem } from "@patternfly/react-core";
+import { List, ListItem, EmptyState } from "@patternfly/react-core";
 
 export default function ConfigEditor() {
   const model = useConfigModel({ suspense: true });
   const devices = useDevices("system", { suspense: true });
+
+  const drives = model.drives || [];
+  const volumeGroups = model.volumeGroups || [];
+
+  if (!drives.length && !volumeGroups.length) {
+    return <EmptyState titleText={_("Select a device for installing the system")} status="info" />;
+  }
 
   return (
     <List isPlain>
--- a/web/src/components/storage/ConfigEditor.tsx
+++ b/web/src/components/storage/ConfigEditor.tsx
@@ -22,11 +22,36 @@
 
 import React from "react";
 import { _ } from "~/i18n";
-import { useDevices } from "~/queries/storage";
+import { useDevices, useResetConfigMutation } from "~/queries/storage";
 import { useConfigModel } from "~/queries/storage/config-model";
 import DriveEditor from "~/components/storage/DriveEditor";
 import VolumeGroupEditor from "~/components/storage/VolumeGroupEditor";
-import { Alert, List, ListItem } from "@patternfly/react-core";
+import { Alert, Button, List, ListItem } from "@patternfly/react-core";
+
+const NoDevicesConfiguredAlert = () => {
+  const { mutate: reset } = useResetConfigMutation();
+  const title = _("No devices configured yet");
+  // TRANSLATORS: %s will be replaced by a "resets to default" button
+  const body = _(
+    "Use actions below to set up your devices or click %s to start from scratch with the default configuration.",
+  );
+  const [bodyStart, bodyEnd] = body.split("%s");
+
+  return (
+    <Alert title={title} variant="custom" isInline>
+      {bodyStart}{" "}
+      <Button variant="link" onClick={() => reset()} isInline>
+        <b>
+          {
+            // TRANSLATORS: label for a link
+            _("resets to default")
+          }
+        </b>
+      </Button>{" "}
+      {bodyEnd}
+    </Alert>
+  );
+};
 
 export default function ConfigEditor() {
   const model = useConfigModel({ suspense: true });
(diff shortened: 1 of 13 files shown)


web/src/hooks/storage/edit-volume-group.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import useApiModel from "~/hooks/storage/api-model";
+import useUpdateApiModel from "~/hooks/storage/update-api-model";
+import { editVolumeGroup } from "~/hooks/storage/helpers/volume-group";
+import { QueryHookOptions } from "~/types/queries";
+
+export type EditVolumeGroupFn = (
+  odlVgName: string,
+  VgName: string,
```
dgdavid, line 30: NP:

It should be "vgName" instead of "VgName".

In any case, I wonder if we should give the "id" and the "nextState" only instead of having to add a new argument as form evolves. I mean,` updateVg(vgName, `nextState`)` where `nextState` (or "changes" or whatever other name fits better) would be something like `{ vgName, targetDevices, whatEverInTheFuture }`. Maybe something similar or inspired by _reducers_.

dgdavid:
Please ignore comments above until the LV actions are completed. They are intended to serve as considerations for later.

web/src/hooks/storage/helpers/volume-group.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { apiModel } from "~/api/storage/types";
+import copyApiModel from "~/hooks/storage/helpers/copy-api-model";
```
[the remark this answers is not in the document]
joseivanlopez: Discussed. It will be improved in the next PR.

web/src/components/storage/ConfigEditor.tsx
```diff
@@ -22,11 +22,36 @@
...
+  const [bodyStart, bodyEnd] = body.split("%s");
+
+  return (
+    <Alert title={title} variant="custom" isInline>
+      {bodyStart}{" "}
+      <Button variant="link" onClick={() => reset()} isInline>
+        <b>
+          {
+            // TRANSLATORS: label for a link
+            _("resets to default")
```
joseivanlopez, line 47: NP: The rest of the UI uses imperative mood, that is, *reset to default*.

joseivanlopez (commented):
LGTM

web/src/components/storage/ConfigureDeviceMenu.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const enabledDescription = drivesCount
+    ? sprintf(
+        n_(
+          "Extends the installation beyond the currently selected disk",
+          "Extends the installation beyond the current %d disks",
+          drivesCount,
+        ),
+        drivesCount,
+      )
+    : _("Extends the installation using a disk");
```
joseivanlopez, line 76: All those *Extends* should be *Extend*.

dgdavid: As discussed int the IRC channel, I do not agree. It's true that for actions we use imperative form, but this is a description of such an action: "**Select** whatever" -> "It **does** that thing"

It's also true that there is an inconsistency with the same description for the "Add LVM" action. But the problem is there, not here.

What I mean is that, **IMHO** we should look at the entire Agama before taking isolated decisions.

Last but not least, I actually think that these descriptions deserves a better review beyond the consistency on the verb form and I fully trust in the fact that we could improve them a bit.

web/src/components/storage/ConfigureDeviceMenu.tsx
```diff
@@ -0,0 +1,241 @@
...
+  onDeviceClick,
+}: DisksDrillDownMenuItemProps) => {
+  const isDisabled = !devices.length;
+
+  const disabledDescription = _("Already using all availalbe disks.");
+  const enabledDescription = drivesCount
+    ? sprintf(
+        n_(
+          "Extends the installation beyond the currently selected disk",
+          "Extends the installation beyond the current %d disks",
```
dgdavid, line 71: @ancorgs

A quick note on wording.

Just to clarify, I’m bringing this up now because it prompted  a thought, not because it needs immediate attention. This is more about _starting the discussion_.

As I’ve mentioned to you on several occasions in our private conversations, I really like a lot of the resources out there, and one I find particularly helpful is the [GOV.UK design system](https://design-system.service.gov.uk/).

Sadly, not everything from it directly applies to Agama, at least not at the moment. Or at least I do not have it clear. But after spending some time with their [Writing for Interfaces guide](https://www.gov.uk/service-manual/design/writing-for-user-interfaces), I think we could borrow some ideas from it.

One piece of advice they offer could be helpful not just for these two lines but for various texts across the code:

> “Don’t follow strict grammar conventions if it makes things clearer not to.”

Putting my English skills aside, I believe this principle could help make certain sentences more readable, even at the cost of some grammatical rules. For example, we could go for _Extends installation beyond currently selected disk_ by dropping _the_ which would make the message easier to understand while keeping the meaning intact.

Just sharing my thoughts to help improve the clarity of interface explanations and make them more digestible (and shorter :wink)

web/src/components/storage/ConfigureDeviceMenu.tsx
dgdavid: Looks like @ancorgs finally decided to [go for your proposal](https://github.com/joseivanlopez/agama/pull/7/commits/3d0d411084217731675a61530bb56ea76d431f6e). I'm going to review the rest of the interface to make it consistent with storage.

web/src/components/storage/ConfigureDeviceMenu.tsx
imobachgs: At some point we should create a PBI to go through all the UI and check the texts. Please, bear in mind that the texts are translated, so we should get them right as soon as possible to avoid invalidating translators work.

PR #2187, opened 2025-03-21
Register extensions  (autoinstallation only)

lslezak:
## Problem

- Extensions cannot be registered

## Solution

- Added full support for registering extensions from SCC
- Updated the code to handle repeated registration during profile import, if the addon is already registered it does nothing. This allows running `agama config edit` several times without any errors or side effects.

## Testing

- Updated unit tests
- Tested manually

## Screenshots

After registering the HA extension the HA pattern is displayed in the software selection:

agama-ha-pattern

## TODO

- [x] Changes
- [ ] Select the product to install (will be solved separately, this PR is already quite big)


rust/agama-lib/src/software/model/registration.rs
```diff
@@ -29,6 +29,15 @@ pub struct RegistrationParams {
     pub email: String,
 }
 
+/// Addon registration
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AddonParams {
+    pub id: String,
```
imobachgs, line 36: Please, describe each field. Those descriptions will be included in the OpenAPI specification.

rust/agama-server/src/software/web.rs
```diff
@@ -312,6 +317,55 @@ async fn register(
...
 
+/// returns list of registered addons
+///
+/// * `state`: service state.
+#[utoipa::path(
+    get,
+    path = "/registration/addons/registered",
+    context_path = "/api/software",
+    responses(
+        (status = 200, description = "List of registered addons, used by Agama CLI", body = Vec<AddonParams>),
```
imobachgs, line 328: This description goes to the OpenAPI spec. So I would omit the CLI part.
```suggestion
        (status = 200, description = "List of registered addons", body = Vec<AddonParams>),
```

rust/agama-server/src/software/web.rs
```diff
@@ -213,6 +213,11 @@ pub async fn software_service(dbus: zbus::Connection) -> Result<Router, ServiceE
             "/registration",
             get(get_registration).post(register).delete(deregister),
         )
+        .route("/registration/addons", post(register_addon))
+        .route(
+            "/registration/addons/registered",
```
imobachgs, line 218: I am not sure about this organization of URLs and HTTP verbs. At first glance, I would expect `POST /registration/addons` to create a new addon, and that's not the intention.

I think will need a way to retrieve the list of addons when we work on the UI, right? So I would reserve `/registration/addons` for that.

* `GET /registration/addons` for the list of addons (out of scope).
* `GET /registration/addons/registered` for the list of registered addons. If both endpoints will use the same format, you could use `GET /registration/addons?registered=true` instead. 
* `POST /registration/addons/register` (or perhaps a better naming) for the registration.

Does it make sense?

rust/agama-lib/src/product/store.rs
```diff
@@ -70,8 +77,15 @@ impl ProductStore {
         if let Some(reg_code) = &settings.registration_code {
             let email = settings.registration_email.as_deref().unwrap_or("");
             self.product_client.register(reg_code, email).await?;
+            // TODO: avoid reprobing if the system has been already registered with the same code?
```
imobachgs, line 80: :+1:

rust/agama-server/src/software/web.rs
```diff
@@ -312,6 +317,55 @@ async fn register(
...
+/// Register an addon
+///
+/// * `state`: service state.
+#[utoipa::path(
+    post,
+    path = "/registration/addons",
+    context_path = "/api/software",
+    responses(
+        (status = 204, description = "registration successful"),
+        (status = 422, description = "Registration failed. Details are in body", body = RegistrationError),
```
imobachgs, line 349:
```suggestion
        (status = 422, description = "Registration failed. Details are in the body", body = RegistrationError),
```

imobachgs (commented):
It is looking good, although I have a few comments (nothing critical).

rust/agama-server/src/software/web.rs
lslezak: I already thought about `/registration/addons?registered=true`, but the problem is that for the CLI we only need minimal data (name, version, reg. code), but for the full web UI we will need much more details, like description, if the extension is free or requires a registration code, license (optional), dependencies, whether it should be selected by default or not, if it is mirrored on the server (RMT only)... That's why it uses a different endpoint in the end.

About `POST /registration/addons/register`, yes, that makes more sense, I'll update the path.

imobachgs: You can also have a query argument, which allows you to switch between a full representation and a summary. But we can discuss this later, when we implement the `/registration/addons` for the UI.

imobachgs (approved):

PR #2189, opened 2025-03-21
Add NM dhcp settings to network model

jcronenberg:
## Problem

For the wicked2nm migration we require some additional dhcp settings, which are currently missing.

## Solution

Add all the necessary values to the model. As these are pretty specific and unlikely to really be nedded by agama, I just added these to the network model and not the network settings and also took care that these shouldn't interfere in any way with the "normal" agama operation.

## Testing

- *Extended unit tests*
- *Tested manually*


rust/agama-server/src/network/model.rs
```diff
@@ -820,6 +820,206 @@ pub struct IpConfig {
     pub routes4: Vec<IpRoute>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub routes6: Vec<IpRoute>,
+    pub dhcp4_settings: Option<Dhcp4Settings>,
+    pub dhcp6_settings: Option<Dhcp6Settings>,
+    pub ip6_privacy: Option<i32>,
+}
+
+#[skip_serializing_none]
```
teclator, line 828: Not checked deeply but I guess we should skip serializing it in general if it is the default...
that is, default **send_hostname** and unset **client_id** and **iaid**.

teclator:
It is already quite useful as we probably will need to expose these settings too, we were already planning to add the per connection hostname dhcp settings at least.

rust/agama-server/src/network/model.rs
jcronenberg: I think for the model it isn't as important as for e.g. the settings. I was basing client_id etc. mostly on `MacAddress` which also isn't skipped if unset. But if you prefer it not being serialized, I'm not against it.

rust/agama-server/src/network/nm/dbus.rs
```diff
@@ -347,6 +347,34 @@ fn ip_config_to_ipv4_dbus(ip_config: &IpConfig) -> HashMap<&str, zvariant::Value
     if let Some(gateway) = &ip_config.gateway4 {
         ipv4_dbus.insert("gateway", gateway.to_string().into());
     }
+
+    if let Some(dhcp4_settings) = &ip_config.dhcp4_settings {
+        ipv4_dbus.insert("dhcp-send-hostname", dhcp4_settings.send_hostname.into());
```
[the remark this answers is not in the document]
teclator: Once it is available in both I guess we are fine adopting it.

teclator (approved):
LGTM

PR #2192, opened 2025-03-22
fix(rust): do not try to connect when it is not needed

imobachgs:
## Problem

Agama CLI tries to connect to the HTTP server even when it is unnecessary. For instance, when running the `agama download`, the client will try to connect to the HTTP server.

## Solution

With this change, the Agama CLI will connect to the server only when needed.

## Testing

- Tested manually

## Future work

I decided to make a simple fix because we plan to redesign the commands.

### DIFF
--- a/rust/agama-cli/src/lib.rs
+++ b/rust/agama-cli/src/lib.rs
@@ -183,10 +183,10 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
 
 /// True if use of the remote API is allowed (yes by default when the API is secure, the user is
 /// asked if the API is insecure - e.g. when it uses self-signed certificate)
-async fn allowed_insecure_api(use_insecure: bool, api_url: String) -> Result<bool, ServiceError> {
+async fn allowed_insecure_api(use_insecure: bool, api_url: &str) -> Result<bool, ServiceError> {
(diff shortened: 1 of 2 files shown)


rust/agama-cli/src/lib.rs
```diff
@@ -217,29 +217,39 @@ pub fn download_file(url: &str, path: &PathBuf) -> Result<(), ServiceError> {
...
+    insecure: bool,
+    authenticated: bool,
+) -> Result<BaseHTTPClient, ServiceError> {
     let mut client = BaseHTTPClient::default();
 
-    client.base_url = api_url.clone();
+    client.base_url = api_url.to_string();
 
-    if allowed_insecure_api(cli.opts.insecure, api_url.clone()).await? {
+    if allowed_insecure_api(insecure, api_url.to_string()).await? {
```
dgdavid, line 229: Question: why not using client.base_url here?

dgdavid (approved):

rust/agama-cli/src/lib.rs
imobachgs: For no particular reason. I just moved the code to a function with the minimal changes. I have updated that call.

dgdavid (approved):

PR #2147, opened 2025-03-11
feat(web): allow editing the hostname

dgdavid:
## Problem

Currently, users cannot edit the hostname through the web user interface.

## Solution

Add a page that allows users to change the hostname directly from the web UI. Additionally, warn users at the registration form about hostname implications with a link to the section for changing it.

## Tests

* Tested manually
* Added unit tests

## Notes

* This PR includes some cherry-picks from `lvm-form` branch to make forms looks a bit better. The motivation to do so is: a) LVM PR is taking longer than expected; b) this PR is no longer against a temporary branch but against `master` 
* Since updating the registration form was initially needed, this PR also takes the opportunity to improve the form. That's why Some commits are purely related to the registration form
* A new "system" namespace has been introduced and the code place there. That's because the hostname entry will be moved to a "System" page when time permits. See comment https://github.com/agama-project/agama/pull/2147#issuecomment-2718431759 to know more

## Screenshots

| Not using static hostname  | Using static hostname |
|-|-|
| localhost_8080_ (8) | localhost_8080_ (10) |

---

Related to https://trello.com/c/bDToEZ43 (protected link)

PR #1017, opened 2024-01-26
Document Avahi/mDNS in the latest Live ISO

lslezak:
- Added Avahi/mDNS section
- Update the  default port on Live ISO (default HTTPS, no 9090 anymore)
- Related to https://build.opensuse.org/request/show/1141765

PR #2198, opened 2025-03-24
refactor(web): hostname page improvemnts.

dgdavid:
Follow up to #2147 (and #2183)

This PR includes few improvements for the hostname page/form. Go commit by commit to know more.

Please, add as many changes you want and merge it when it is in a shape you think is acceptable.

To see the final result run it locally and test by yourself.

### DIFF
--- a/web/src/components/system/HostnamePage.tsx
+++ b/web/src/components/system/HostnamePage.tsx
@@ -91,7 +91,9 @@ export default function HostnamePage() {
 
         {hasTransientHostname && (
           <Alert variant="custom" title={transientHostnameAlertTitle}>
-            {_("This hostname is dynamic and may change after a reboot or network update.")}
+            {_(
+              "This hostname is dynamic and may change after a reboot or network update, as configured by the local network administrator.",
+            )}
           </Alert>
         )}
 
--- a/web/src/components/system/HostnamePage.tsx
+++ b/web/src/components/system/HostnamePage.tsx
@@ -43,10 +43,10 @@ export default function HostnamePage() {
   const { selectedProduct: product } = useProduct();
   const { transient: transientHostname, static: staticHostname } = useHostname();
   const { mutateAsync: updateHostname } = useHostnameMutation();
-  const hasStaticHostname = !isEmpty(staticHostname);
+  const hasTransientHostname = isEmpty(staticHostname);
   const [success, setSuccess] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
-  const [settingHostname, setSettingHostname] = useState(hasStaticHostname);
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

BTW, we have to merge today to have it in beta 3. So, if nobody suggests changes in the texts, then please merge it. Thanks!

imobachgs:
@mvidner Please, do you have any suggestion to improve the wording? Otherwise, it LGTM.

web/src/components/system/HostnamePage.tsx
```diff
@@ -93,20 +89,25 @@ export default function HostnamePage() {
...
         )}
 
-        <Alert variant="custom" title={hostnameAlertTitle}>
-          {hasStaticHostname
-            ? _("This hostname is permanent and will not change unless manually updated.")
-            : _("This hostname is dynamic and may change after a reboot or network update.")}
-        </Alert>
+        {hasTransientHostname && (
+          <Alert variant="custom" title={transientHostnameAlertTitle}>
+            {_("This hostname is dynamic and may change after a reboot or network update.")}
```
mvidner, line 94:
```suggestion
            {_("This hostname is dynamic and may change after a reboot or network update, as configured by the local network administrator.")}
```

"Network update" is quite vague for me. IIUC the transient hostname is provided by the DHCP server but I avoided that term and referred to a human.

Also, perhaps off topic for this PR, there are two sources for the hostname
1. this machine's administrator
2. LAN administrator

Both may pass on having an opinion.
If none is provided, we should apply `linux-a1b2c` (like YaST but 5 characters because innovation)
If both are provided, we may want to say
> Your name for this machine is `www` but other machines on the network know it as `dhcp47348.example.com`

teclator: We could apply the naming suggestion but the question is if it should be persistent or not. We are currently not touching the persistent hostname but just modifying the transient hostname in case there is nothing and was mainly done to use Mutlicast DNS using the agama.local hostname resolution.

Related to:
- Original agama-hostname.service PR (have to find it...)
- https://github.com/agama-project/agama/pull/1432
- https://github.com/agama-project/agama/pull/1017

teclator: Having a random one will require having to know the hostname each time, but will register the system with a different name, in that case we maybe should change it to be persistent or otherwise will be lost after rebooting the system.

There could be a problem if the network is configured later with DHCP and receiving a hostname after if it is set persistently which as I explained yesterday will not modify the hsotname as in case of set persistently then the transient hostname will be ignored..

imobachgs:
> BTW, do we need changelog entry for this?

No, I do not think so.

PR #2199, opened 2025-03-24
Fixing DUD functionality to start during boot

kobliha:
## Problem

*The DUD functionality was not enabled and really working in the Live Environment*

- Found during manual testing of the built media

## Solution

- The DUD service is now enable in the kiwi config
- Dmesg does not print anything on the console while DUD runs
- The DUD script has an exec bit

## Testing

- *Tested manually only*


live/src/agama-installer.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Mar 24 10:57:22 UTC 2025 - Lukas Ocilka <locilka@suse.com>
+
+- Fixed DUD issues
```
imobachgs, line 4:
```suggestion
- Fixed DUD issues (gh#agama-project/agama#2199):
```

imobachgs (approved):

PR #2200, opened 2025-03-24
feat(web): manage logical volumes

joseivanlopez:
Allow managing LVM logical volumes (add, edit and delete).

Notes: 

* *LogicalVolumePage* and *PartitionPage* components must be refactored. 
* The validation in both pages should be adaped in order to follow the same approach as the rest of the UI (i.e., validating on accept).

Disclaimer: 

The code in this PR was done in a hurry for including LVM managent in SLE16 Beta3. It does not satisfy our expected quality thresholds and will be refactored as soon as possible.


web/src/components/storage/LvmPage.test.tsx
```diff
@@ -128,20 +128,15 @@ jest.mock("~/queries/storage", () => ({
...
-  useVolumeGroup: (id: string) => (id ? mockRootVolumeGroup : null),
-  default: () => mockUseModel,
+  useModel: () => mockUseModel,
 }));
 
-jest.mock("~/hooks/storage/add-volume-group", () => ({
-  ...jest.requireActual("~/hooks/storage/add-volume-group"),
+jest.mock("~/hooks/storage/volume-group", () => ({
+  ...jest.requireActual("~/hooks/storage/volume-group"),
   __esModule: true,
```
dgdavid, line 136: NP: this can be deleted now, thanks.

ancorgs:
Testing this manually I see the checkboxes for the different disks look weird.

checkboxes

dgdavid:
> Testing this manually I see the checkboxes for the different disks look weird.
> 
> checkboxes

Really weird. Thanks for sharing the screenshot

ancorgs:
> Testing this manually I see the checkboxes for the different disks look weird.

It is fixed by a `align-content: baseline` properly placed (although I'm not sure where is the most correct place).

dgdavid:
> > Testing this manually I see the checkboxes for the different disks look weird.
> 
> It is fixed by a `align-content: baseline` properly placed (although I'm not sure where is the most correct place).

I was having a look. We already are defining how checkboxes should be aligned: centered. But the problem is that when the description is longer, these grids loss the alignment between them because there is less space for the first row in those which longer descriptions. Somehow it grows to the top.

Screenshot From 2025-03-26 17-42-32

ancorgs:
Problem: There is an easy way to create LVs without a name. Which is not properly handled by the UI.

1. Start with a default installation
2. Choose the option to create an LVM on top of that disk, moving the partitions to LVs.
3. Click on "edit" on any of the LVs.

ancorgs:
Suggestion: when creating the first LVM (the one that automagically gets the name "system"), I would also automagically pre-select the disks corresponding to the existing drives.

ancorgs:
I'm still not finished, but I have implemented some changes at https://github.com/joseivanlopez/agama/pull/7

ancorgs (approved):
I have been playing with this branch and I can confirm it works (although I have checked the code only very very briefly).

So taking into account the context, it LGTM.

PR #2202, opened 2025-03-24
append only valid key=value lines for agama.conf

### LINKED ISSUE #2185: inst.info should extract only the usable lines
As of b402cc10acc591cae8d02805b0e2726d66d9354d, [‎live/root/usr/bin/info-cmdline-conf.sh](https://github.com/agama-project/agama/blob/462192e0e202fb346b7fc20ca1de47e6fb049c44/live/root/usr/bin/info-cmdline-conf.sh#L23) uses `cat` to append the retrieved content verbatim. I think the expected input is `key=value`. As such the usage of `cat` should be replaced by `grep =`.

This will make it equivalent to the behavior of linuxrc `info=URL`, which just ignores input that is not valid.

olafhering:
inst.info=URL allows to retrieve agama settings from a remote location. Usually the expected settings are in key=value format.

This change allows to reuse the inst.info=URL feature for the cobbler "nopxe" API, just like it is done in linuxrc. The system needs to notify cobbler that it booted "far enough", so that cobbler can change the boot loader default from "boot from network" to "boot from local disk". The client has to send a HTTP GET request to the specified URL. cobbler sends the four bytes 'True' in return, which can system has to ignore. Since the content is not and can not be a valid Agama key=value pair, consider only lines that contain '=' in the input file.

Fixes #2185

### DIFF
--- a/live/root/usr/bin/info-cmdline-conf.sh
+++ b/live/root/usr/bin/info-cmdline-conf.sh
@@ -20,7 +20,7 @@ expand_info_arg() {
   # remove info param
   sed -in 's/\([[:space:]]\|^\)\(inst\|agama\)\.info=[^[:space:]]\+//' "${TARGET}"
   # and add content of info file
-  cat "${INFO_CONTENT}" >> "${TARGET}"
+  grep -- '=' "${INFO_CONTENT}" >> "${TARGET}"
 
   return 0
 }


kobliha:
There are some options that do not have any value and there might be even more in the future, for example

- live.password_dialog
- live.password_systemd

See https://agama-project.github.io/docs/user/boot_options

olafhering:
The = for password could be easily enforced, and Y2DEBUG could be specified by kernel cmdline, if really required.

jreidinger:
question is if we should abuse info param as in past or better have dedicated support in agama with something like `agama.cobbler=<URL>` that basically do that notification and also verify that it properly return "True" as requested and if not, then report error in agama ( something we plan also for failed of sending info or others ). For me it would be more clear if we do dedicated support with own parameter.

olafhering:
Well, the client can not do anything about whatever it receives. The point is that it needs to send a GET request, and it needs to deal with any content, by just ignoring it.
There has to be some `something=$URL` where the received content and every possible error is ignored. And unfortunately nothing like that exists, the new inst.info= option is close to what is required, and in addition there is no dracut upstream to even discuss a generic `something=$URL` or `something=$URL,$outputfile`.

PR #2152, opened 2025-03-13
adapt patterns name for SLES 16 / SLES4SAP 16

fcrozat:
Some patterns are being renamed in SLES 16/ SLES4SAP 16.

PR #2203, opened 2025-03-25
docs: update agama-products changes file

imobachgs:
Add changes for #2106.

PR #2204, opened 2025-03-25
products: add gnome pattern to the SLES 16.0 product , adjust patterns name for SLES 16 / SLES4SAP 16

g7:
This also includes and supersedes #2152.


g7:
v2: rebased on top of #2203

imobachgs (approved):
Thanks!

PR #2205, opened 2025-03-25
fix(web): hostname page rewording

dgdavid:
To make its texts short and consistent with the rest of pages.


joseivanlopez (approved):
LGTM

Does it need a changelog entry?

PR #2208, opened 2025-03-25
Drop yast2-schema and autoyast2-installation requirements (SR#367433, SR#367417)

g7:
yast2-schema and autoyast2 were dropped from SLFO:Main, so we shouldn't install them anymore.


g7:
v2: also drop autoyast2-installation.

imobachgs:
Yes, sorry, I went too far with removing packages. I will resubmit some of them.

Issue #341, opened 2022-11-29
Consider using cage for gnome/firefox less execution


lkocman:
using something like https://github.com/Hjdskes/cage for d-installer would eliminate the requirement to load gnome and firefox 

Originally raised by Patrick Fitzgerald on the community call.

ancorgs:
We welcome any idea in these fields:

- An improved and more light-weight Live ISO.
- A container that can run Wayland + a browser so we don't even need an ISO (see some experiments [here](https://build.opensuse.org/package/show/YaST:Head:Containers/kiosk-container)

Those are areas in which experts in operating systems, graphical environments, browsers and/or containers could help without needing to code D-Installer or YaST.

If you are reading this, please consider jumping in. :wink:

lkocman:
Communicated back to Patrick, just for tracking you could start with forking this project
https://build.opensuse.org/package/show/YaST:Head:D-Installer/d-installer-live

jreidinger:
Yeah, I play with cage and also sway ( which is in TW as opposed to cage ) and my main issue was that it failing with Leap15.X qemu. So can be tested only on TW host which I sadly do not have at hand. And question is how looks real hardware support ( especially on other archs like aarch64 ).

javierllorente:
Perhaps we could get rid of Firefox completely by using an embedded Web browser (ie: Qt WebEngine)

jotolo2:
Hi guys!
I've been able to configure a live `d-installer` iso that uses `cage` with a `qt-kiosk-browser`:
https://build.opensuse.org/package/show/home:jtorres:branches:YaST:Head:D-Installer/d-installer-live-test

This is a POC and many things should be improved, more specifically on `cage` side.

`Cage` is started as a systemd unit service. At this moment it can display the browser with `cockpit-desktop`, however, it has some issues with the cursor and a lot of lagging, it's not usable. I should research what is making this issue. 

It can be attached to that `cage` instance a `wayvnc` daemon to show its display through VNC protocol. I've set up a systemd unit service for this too.

The `qt-kiosk-browser` is a minimal browser I've set up based on the `Qt` examples. It would make sense to see how to make it more minimal and avoid some `Qt` dependencies it has at the moment.

The next step I will do is drop all Xorg dependencies and see if the iso improves on light-weight.

Lunarequest:
[...]

PR #2211, opened 2025-03-26
Switch firefox from kiosk mode to fullscreen mode

jreidinger:
The change add extension for starting in screen mode as now there is no simple way how to start ff in fullscreen mode. Extension is simple 15 lines and was manually reviewed by me. License is MIT.


lslezak:
Please update the `live/README.md` documentation why we install the plugin. Please add also some links so we know where it was downloaded from and could possibly update it later.

dgdavid:
> Please update the `live/README.md` documentation why we install the plugin. Please add also some links so we know where it was downloaded from and could possibly update it later.

Additionally, what about forking it, if possible, in the agama-project organization?

jreidinger:
@dgdavid is there reason to fork it? We need to have it signed and with metadata, so we would need to push it to get unique id that is used as name of xpi.

jreidinger:
@lslezak updated with requested info

dgdavid:
> @dgdavid is there reason to fork it? We need to have it signed and with metadata, so we would need to push it to get unique id that is used as name of xpi.

It was just an idea to make sure we have the full control of the code and also still having it if for some reason the orinal dissapear. 

But if it's being downloaded just one time I guess it does not worth

jreidinger:
@dgdavid well, the code does not change for 4 years and code is really trivial. You can see it yourself. Basically it sets all open window full screen status at https://github.com/insiderser/AutoFullscreen/blob/master/startup.js#L28

dgdavid:
> @dgdavid well, the code does not change for 4 years and code is really trivial. You can see it yourself. Basically it sets all open window full screen status at https://github.com/insiderser/AutoFullscreen/blob/master/startup.js#L28

I know. I already saw it. I was just commenting an idea that I already commented (somewhere, in a meeting or a comment, don't remember) for other dependencies too. Based on the https://en.wikipedia.org/wiki/Npm_left-pad_incident for which does not matter how many years without changes the project was before.

But it was just a (wild) idea. Let's forget it and go ahead.

jreidinger:
@dgdavid thanks for bringing it up topic, but I see difference to mentioned npm issue. We have copy of that extension at git and we control it. It will be different topic if we do download of extension during build in CI or in OBS, but here we have source under our control. It can make sense for other npm/rust/ruby dependencies, but I think it is solved on SUSE level with having all required sources to build in OBS for e.g. reproducible builds, so such disruption should be prevented.

dgdavid:
@jreidinger, the work here might be closing https://github.com/agama-project/agama/issues/341

jreidinger:
@dgdavid I see more issues that can be closed with it, but lets see how wayland will go as this kind of expect it to be done first.

PR #2214, opened 2025-03-27
test: fix LicenseDialog test

imobachgs:
These tests were broken after the list of supported languages was reduced due to the low coverage of most of the translations.

We should find a way to make it more robust, but by now it is enough.


dgdavid (approved):

dgdavid:
Thanks for this. It could unblock also other PRs

PR #2215, opened 2025-03-27
refactor(web): drop unused icons

dgdavid:
As the project has evolved, many icons are no longer used, but they were still being imported and included in the final bundle.

This PR removes the unused icons and also drops a dependency that is no longer needed after the clean-up.

Since there are no visual changes for users, no entry has been added to the changes file.


joseivanlopez (approved):
LGTM

PR #2224, opened 2025-03-28
fix: properly convert "sources" to "content"

imobachgs:
In #2121 we decided to unify files and scripts to use the same `content` property, instead of `body`. However, we did not update the conversion for AutoYaST scripts.


jreidinger (approved):
thanks

PR #2226, opened 2025-03-28
Copy the /etc/hostname to the target system if exist

teclator:
## Problem

When an static hostname exists it should be copied to the installed system at the end of the installation but it is not.

## Solution

Copy the /etc/hostname when it exists.

## Testing

- *Added a new unit test*


service/lib/agama/network.rb
```diff
@@ -117,5 +120,21 @@ def copy_directory(source, target)
...
+
+    # Copies a file
+    #
+    # This method checks whether the source file exists. It copies the file to the target system if
+    # it exists
+    #
+    # @param source [String] source file
+    # @param target [String,nil] target directory, only needed in case it is different to the
+    # original source path in the target system.
+    def copy(source, target = nil)
```
dgdavid, line 132: NP: I wonder if this deserves to live in a more _generic_ place instead of here. Something to think about.

dgdavid (approved):

service/lib/agama/network.rb
dgdavid: In fact, there is _similar_ code at https://github.com/agama-project/agama/blob/8525cdda29d9e4eca8c2b5d195a60506a941dc49/service/lib/agama/software/manager.rb#L686 that could use your method if extracted somewhere. Unless I'm wrong.

PR #2227, opened 2025-03-28
Bootloader timeout

jreidinger:
## Problem

Feedback from QA was that they need beside stopping on boot menu also for some specific hardware to just set long waiting time to capture boot screen, but have problem to send enter.

## Solution

Add timeout option to bootloader configuration which is mutually exclusive with stopOnBootMenu.

## Testing

- *Added a new unit test*
- *Tested manually*

testing command was `jq -n '.bootloader.timeout = 60' | agama config load` to test minute long boot timeout.

note from testing. If both options are specified in json, then schema error is not much helpful, but sadly I do not know if it is possible to make it better:
`{'timeout': 5, 'stopOnBootMenu': True}: {'timeout': 5, 'stopOnBootMenu': True} is valid under each of {'required': ['timeout']}, {'required': ['stopOnBootMenu']}`


jreidinger:
rust ci failure is not related ssl failure for coveralls.

service/lib/agama/storage/bootloader.rb
```diff
@@ -31,22 +31,41 @@ class Bootloader
...
         attr_accessor :stop_on_boot_menu
+        # bootloader timeout. Only positive numbers are supported and stop_on_boot_menu has
+        # precedence
+        attr_accessor :timeout
+        # as both previous keys are conflicting, remember which one to set or none. It can be empty
+        # and it means export nothing
+        attr_accessor :keys_to_export
 
         def initialize
+          @keys_to_export = []
```
imobachgs, line 42: Do you need to track the keys in a separate variable? I would have expected them to be `nil` if they were not specified, so yo need whether to export them or not depending on that condition.

imobachgs (commented):
It looks good, although I am not sure about the approach to decide which values to export.

service/lib/agama/storage/bootloader.rb
jreidinger: well, in this case I prefer to have it explicit..in general I try to avoid nils as it cause troubles on DBus even in this case API is just string one with json encoded. Also using nil means I would need to nullify other value when the first one is set.

imobachgs (approved):

czerw:
> note from testing. If both options are specified in json, then schema error is not much helpful, but sadly I do not know if it is possible to make it better:
{'timeout': 5, 'stopOnBootMenu': True}: {'timeout': 5, 'stopOnBootMenu': True} is valid under each of {'required': ['timeout']}, {'required': ['stopOnBootMenu']}


Should be easy, just remove `stopOnBootMenu` feature.

czerw:
Will work `jq -n '.bootloader.timeout = -1'` ?

jreidinger:
> Will work `jq -n '.bootloader.timeout = -1'` ?

no, the schema specify that timeout has to be possitive. So either specify stopOnBootMenu or timeout

czerw:
I would just drop `stopOnBootMenu` on just allow bootloader to use `-1`.  Current approach dosn't makes much sense.

ancorgs:
> I would just drop `stopOnBootMenu` on just allow bootloader to use `-1`. Current approach doesn't make much sense.

Allowing `-1` and making that mean "stop forever" only makes sense because we are used to it. It's the (relatively hacky) grub2 way to do things, but not the only one (others boot loaders may propose other syntax) and definitely not the most semantic.

kobliha:
We understand that the solution might not make sense to someone, but it makes sense to us for several reasons.
While technically _timeout = -1_ is the way how it is implemented, it's an implementation detail that we want to be hidden from the user. Agama tries to implement solutions logically from the user's point of view instead of using logic which depends on the used software, which can actually change and the API would have to change as well. The way how it is implemented does not need to change the user-facing config file and the docs if the bootloader of your choice is different.

czerw:
Thanks for explanation. I understand now that you don't  want be compatible with old style logic and you reserved space for the future.

PR #2230, opened 2025-03-31
Register extension from web UI

lslezak:
## Problem

- Extensions can be registered only from command line or in autoinstallation

## Solution

- Implement the web UI

## Screenshots

The first proposal:

agama-extension-ui

After registering the registration code field and the button is replaced by "registered" status text:

image


web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,6 +181,81 @@ const HostnameAlert = () => {
...
+                    </Label>
+                  )}
+                  {ext.free && (
+                    <Label color="green" isCompact>
+                      {_("Free")}
+                    </Label>
+                  )}
+                </div>
+                <div>{ext.description}</div>
+                <Form>
```
dgdavid, line 230: I’m not entirely sure about going in this direction (directly exposing the form) in the future, but as mentioned in IRC, this is good enough for a first iteration until we know more details. However, please consider the following points:

    * Add a unique ID to the form, as there could potentially be more than one form.
    * For the same reason, add an ARIA label to help users of assistive technology distinguish between different forms
    * Do not render the form when there are no fields. In these cases, please use a simple button for triggering the register action. Again, use an ARIA label for these buttons as well, since it might be difficult to distinguish one "Register" button from another when navigating them.

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,6 +181,81 @@ const HostnameAlert = () => {
   );
 };
 
+const Extensions = () => {
```
dgdavid, line 184: Honestly, I think this component deserves to live in it's own file and, if possible, having its own unit test.

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -191,6 +273,8 @@ export default function ProductRegistrationPage() {
       <Page.Content>
         {isUnregistered && <HostnameAlert />}
         {isUnregistered ? <RegistrationFormSection /> : <RegisteredProductSection />}
+        {/* TODO: display only when registered */}
```
dgdavid, line 276: Do not forget about this.

dgdavid (commented):
Thanks, @lslezak!

To be honest, I'm not completely sure how this will look when we have a bunch of extensions, but it definitely seems like a very good starting point to me.

That said, I’ve left a few comments that I think could help refine it a bit more before releasing a first iteration.

Hope it helps.

service/lib/agama/dbus/software/product.rb
```diff
@@ -151,6 +151,27 @@ def registered_addons
           end
         end
 
+        # list of available addons
+        #
+        # @return [Array<Hash<String, Object>>] List of addons
+        def available_addons
+          addons = backend.registration.available_addons || []
+
+          addons.map do |a|
```
lslezak, line 160: Unfortunately this fails with DBus error:  `Specified type has 2 members but value has 9 members; caused by 1 sender=:1.1 -> dest=org.opensuse.Agama.Software1...` 

I guess I need to wrap something into an Array... @mvidner ?

lslezak (commented):
DBus error

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,9 +204,112 @@ const HostnameAlert = () => {
...
+            {_("Beta")}
+          </Label>
+        )}
+        {extension.recommended && (
+          <Label color="orange" isCompact>
+            {_("Recommended")}
+          </Label>
+        )}
+      </Title>
+      <p>&nbsp;</p>
```
lslezak, line 253: @dgdavid any hint how to add some spacing instead of this hack?

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,9 +204,112 @@ const HostnameAlert = () => {
...
+
+const Extensions = () => {
+  const extensions = useAddons();
+
+  // count the extension versions
+  const counts = {};
+  extensions.forEach((e) => (counts[e.id] = counts[e.id] ? counts[e.id] + 1 : 1));
+
+  const extensionComponents = extensions.map((ext, index) => (
+    <Extension key={`extension-${index}`} extension={ext} unique={counts[ext.id] === 1} />
```
dgdavid, line 304: Hard to understand the count thingy. Isn't there another way to know if the extension is unique?

What is more, 

`(counts[e.id] = counts[e.id] ? counts[e.id] + 1 : 1)`

Looks a not valid expression to me. I mean, it will go always for the true branch because there is no a comparison but an assignation.

Although since you have tested the code I suspect I'm missing something. Do you mind explaining this better?

web/src/queries/software.ts
```diff
@@ -166,6 +187,22 @@ const useRegisterMutation = () => {
...
+ * Hook that builds a mutation for registering an addon
+ *
+ */
+const useRegisterAddonMutation = () => {
+  const queryClient = useQueryClient();
+
+  const query = {
+    mutationFn: registerAddon,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ["software/registration/addons/registered"] });
```
dgdavid, line 200: NP: not because of your changes, but something I'd like to do/improve is how we use query keys to avoid mistyping them and having bugs. Here you could go for `queryKey: registeredAddonsQuery().queryKey`. Or we could have constants for queryKeys. I don't know, but we have to found a way. Out of the scope of this PR, but adding the comment for making others aware. @imobachgs, @joseivanlopez friendly ping since you both had worked with queries quite a lot.

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,9 +204,118 @@ const HostnameAlert = () => {
...
+                isLoading={loading}
+              >
+                {_("Register")}
+              </Button>
+            </ActionGroup>
+          </Form>
+        ) : (
+          <Alert title={_("Not available")} variant="warning">
+            {_(
+              "This extension is not available on the server. Please ask the server administrator to mirror the extension.",
```
dgdavid, line 287: **Please** avoid using _please_. The sentence is clear and even more direct without it. I'm working on improving overall wording and following (among other recommendations) the SUSE doc style guide, which also advises against using please, see https://documentation.suse.com/style/current/single-html/docu_styleguide/index.html#sec-vocabulary. The goal is to avoid introducing new instances and to clean up existing ones when possible.

```suggestion
              "This extension is not available on the server. Ask the server administrator to mirror the extension.",
```

Or better if we split it in two.


```jsx
<Alert title={_("Extension not available on the server")} variant="warning">
  {_("Ask the server administrator to mirror the extension.")}
 </Alert>
```

mvidner:
Just a note about the `aa{sv}` versus `a{sv}` confusion referenced [earlier](https://github.com/agama-project/agama/pull/2230#pullrequestreview-2742352923), I have tried to improve ruby-dbus to report a better error message for this, but it is a bit hard to fix because the weird "a Dictionary is actually an Array of Dictionary-Entries" concept is spread through the code base.

https://github.com/mvidner/ruby-dbus/pull/148

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -174,6 +181,81 @@ const HostnameAlert = () => {
...
+  const extensionComponents = extensions.map((ext) => (
+    <DataListItem key={`${ext.identifier}-${ext.version}`}>
+      <DataListItemRow>
+        <DataListItemCells
+          dataListCells={[
+            <DataListCell key="summary">
+              <Stack hasGutter>
+                <div>
+                  {/* remove the "(BETA)" suffix, we display a Beta label instead */}
+                  <b>{ext.friendlyName.replace(/\s*\(beta\)$/i, "")}</b>{" "}
```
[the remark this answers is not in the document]
lslezak: Unfortunately not, there is `shortname` property but that on the other does not include the version number in some cases. But when you show `shortname` + `version` then in the rare cases when the `shortname` already contains a version the version is displayed twice. :man_shrugging: 

Personally I do not like touching the label text but there is no better way. :roll_eyes:

web/src/components/product/ProductRegistrationPage.tsx
lslezak:
```
(counts[e.id] = counts[e.id] ? counts[e.id] + 1 : 1)
                ^^^^^^^^^^^^
```

There is an implicit comparison, the `counts[e.id]` above is evaluated as boolean. If the key does not exist it returns `undefined` which in boolean context is evaluated as `false`. If the key exists the returned number is evaluated as `true`.

dgdavid: Still quite tricky, but if there is no other option just go ahead.

lslezak: I have changed that.

web/src/components/product/RegistrationComponents.tsx
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import React from "react";
+import { TextInputProps } from "@patternfly/react-core";
+
+import { PasswordInput } from "~/components/core";
+import { _ } from "~/i18n";
+
+// TRANSLATORS: input field label
+const KEY_LABEL = _("Registration code");
```
dgdavid, line 30: NP: I see you’re reusing something you found elsewhere, but it is kind of old approach that we finally are not following. My fault, sorry. That said, let’s avoid exporting that key. It's better to just use it directly where it's needed. Even though you’re using the same key in a couple of places now, the forms could end up needing different ones later.

The point is that even when you are "instantiating" the component in two different places using, now, the same key, these forms could use a different one at any point. 

Also, about extracting this to a separate component: I get why you did it, especially with the width on the Password input. Personally, I’m not sure it’s necessary since it's simply using a different value for a simple prop, but I can see the reasoning.

But please, 

* Let’s rename the file to match the exported component name.
* Use a default export for the component.
* And again, let’s not define and export the label.

Thanks a lot!

rust/agama-lib/src/product/client.rs
```diff
@@ -146,6 +146,64 @@ impl<'a> ProductClient<'a> {
...
+            .registration_proxy
+            .available_addons()
+            .await?
+            .into_iter()
+            .map(|hash| AddonProperties {
+                id: hash
+                    .get("id")
+                    .unwrap_or(&zbus::zvariant::Value::new(""))
+                    .try_into()
+                    .unwrap_or_default(),
```
jreidinger, line 161: get optional property is not possible to use? https://github.com/agama-project/agama/blob/master/rust/agama-lib/src/dbus.rs#L51 ... also all those unwrap_or_default looks bad. It means it can hide errors in empty string.

jreidinger: Basically if property is mandatory it should not use empty string if it is not there....and if it is not mandatory then type should be `Option<String>`...regarding the last `unwrap_or_default` it should be simple `?` to propagate error.

jreidinger: example usage is at https://github.com/agama-project/agama/blob/master/rust/agama-server/src/storage/web/iscsi.rs#L106

rust/agama-lib/src/software/model/registration.rs
```diff
@@ -41,6 +41,23 @@ pub struct AddonParams {
     pub registration_code: Option<String>,
 }
 
+/// Addon registration
+#[derive(Clone, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AddonProperties {
+    // Addon identifier
```
jreidinger, line 48: it is part of schema, so `///` should be used to see documentation in schema...also all properties should have user documentation.

lslezak:
I tried to simplify the code in https://github.com/agama-project/agama/pull/2230/commits/b40dea6e8e924f0bf80f788d334b0da9f1353468, but now it fails with error:

```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `std::ops::FromResidual`)
   --> agama-lib/src/product/client.rs:157:46
    |
156 |             .map(|hash| AddonProperties {
    |                  ------ this function should return `Result` or `Option` to accept `?`
157 |                 id: get_property(&hash, "id")?,
    |                                              ^ cannot use the `?` operator in a closure that returns `software::model::registration::AddonProperties`
    |
    = help: the trait `std::ops::FromResidual<std::result::Result<std::convert::Infallible, zbus::zvariant::Error>>` is not implemented for `software::model::registration::AddonProperties`
```

See the full log at https://github.com/agama-project/agama/actions/runs/14464157813/job/40562519285#step:8:619.

No idea how to fix that... :worried:

lslezak:
OK, now it compiles fine.

rust/agama-lib/src/product/client.rs
```diff
@@ -146,6 +146,28 @@ impl<'a> ProductClient<'a> {
...
     }
 
+    // details of available addons
+    pub async fn available_addons(&self) -> Result<Vec<AddonProperties>, ServiceError> {
+        let addons: Vec<AddonProperties> = self
+            .registration_proxy
+            .available_addons()
+            .await?
+            .into_iter()
+            .map(|hash| AddonProperties {
```
imobachgs, line 156: It would be a good idea to move this code to a separate function. But, for now, it is good enough. Thanks!

imobachgs (commented):
The Rust part looks good.

imobachgs (approved):

jreidinger (approved):

PR #2233, opened 2025-03-31
Print also the IPv6 installer URLs in the console (bsc#1240197)

lslezak:
## Problem

- The installer prints the "Network is not available, the Agama installer cannot be used remotely" error message on the console when only an IPv6 connection exists.
- https://bugzilla.suse.com/show_bug.cgi?id=1240197

## Solution

- Query also the network IPv6 addresses and display the URLs for them too
- Ignore the IPv6 link local addresses, they are not supported by browsers (mostly because the URL is not standardized :roll_eyes:)
- QR codes are not generated for IPv6 addresses, only for IPv4
  - I'm not sure about the IPv6 support in mobile phones/mobile browsers
  - More over the WiFi would need to support IPv6
  - And you need working IPv6 connection between WiFi and the network where your installer is running
  - Also the IPv6 address is longer and requires a bigger QR code, the logic for displaying IPv4 and IPv6 QR codes side-by-side would be a lot more complicated

## Testing

- Tested manually

## Screenshots

- For testing purposes I temporarily disabled the link local address filtering to display some IPv6 address in my testing VM, in the final code the IPv6 URL below would not be displayed.

agama-ipv6-issue

### DIFF
--- a/live/root/usr/bin/agama-issue-generator
+++ b/live/root/usr/bin/agama-issue-generator
@@ -172,11 +172,15 @@ function create_qr_codes() {
     if [ -n "$TERM_WIDTH" ] && [ "$TERM_WIDTH" -gt 80 ]; then
       # compute how much QR codes can fit on the screen side-by-side
       QR_NUM=$(( TERM_WIDTH / QR_WIDTH ))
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM

PR #2236, opened 2025-04-01
Fix forget wireless network action.

teclator:
## Problem

When try to forget a wireless network it raises an error because it remains selected when it does not exist anymore.

## Solution

It is deselected and disconnected from the selected network to be forgotten.

## Testing

- *Tested manually*

### DIFF
--- a/web/src/queries/network.ts
+++ b/web/src/queries/network.ts
@@ -96,7 +96,7 @@ const connectionsQuery = () => ({
 });
 
 /**
- * Returns a query for retrieving the list of known access points sortered by
+ * Returns a query for retrieving the list of known access points sorted by
  * the signal strength.
  */
 const accessPointsQuery = () => ({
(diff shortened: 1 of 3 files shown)


web/src/queries/network.ts
```diff
@@ -96,13 +96,14 @@ const connectionsQuery = () => ({
 });
 
 /**
- * Returns a query for retrieving the list of known access points
+ * Returns a query for retrieving the list of known access points sortered by
```
imobachgs, line 99:
```suggestion
 * Returns a query for retrieving the list of known access points sorted by
```

imobachgs (changes requested):
We are working in a new form for handling wireless connections so, please, drop the "forget" to minimize conflicts. WDYT?

imobachgs:
Replaced with #2247.

PR #2238, opened 2025-04-01
do not export empty defaults for user

jreidinger:
## Problem

When user do not configure first user and try to export profile or do `agama config edit` and do not touch user section it failed to load with invalid user that cannot be empty.

## Solution

Skip exporting user if it is not defined.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-lib/src/users/settings.rs
+++ b/rust/agama-lib/src/users/settings.rs
@@ -84,11 +84,8 @@ pub struct RootUserSettings {
     pub ssh_public_key: Option<String>,
 }
 
-
 impl RootUserSettings {
     pub fn skip_export(&self) -> bool {
-        self.password.is_none()
-            && self.hashed_password.is_none()
-            && self.ssh_public_key.is_none()
+        self.password.is_none() && self.hashed_password.is_none() && self.ssh_public_key.is_none()
     }
 }
--- a/rust/agama-lib/src/users/settings.rs
+++ b/rust/agama-lib/src/users/settings.rs
@@ -38,12 +38,16 @@ pub struct UserSettings {
 #[serde(rename_all = "camelCase")]
 pub struct FirstUserSettings {
     /// First user's full name
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub full_name: Option<String>,
     /// First user's username
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub user_name: Option<String>,
     /// First user's password (in clear text)
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub password: Option<String>,
     /// Whether the password is hashed or is plain text
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub hashed_password: Option<bool>,
 }
 
--- a/rust/agama-lib/src/users/settings.rs
+++ b/rust/agama-lib/src/users/settings.rs
@@ -26,8 +26,9 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Default, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct UserSettings {
-    #[serde(rename = "user")]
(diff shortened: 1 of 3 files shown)


rust/agama-lib/src/users/store.rs
```diff
@@ -40,12 +40,22 @@ impl UsersStore {
         })
     }
 
+    // small helper to convert dbus empty string to None
+    fn optional_string(value: String) -> Option<String> {
```
imobachgs, line 44: I would put this function in https://github.com/agama-project/agama/blob/master/rust/agama-lib/src/dbus.rs

rust/agama-lib/src/users/settings.rs
```diff
@@ -38,12 +38,16 @@ pub struct UserSettings {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub full_name: Option<String>,
     /// First user's username
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub user_name: Option<String>,
     /// First user's password (in clear text)
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub password: Option<String>,
     /// Whether the password is hashed or is plain text
+    #[serde(skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
imobachgs: Yes, it is fine with me.

imobachgs (commented):
Is the `user` key still exported?

rust/agama-lib/src/users/settings.rs
```diff
@@ -63,3 +83,9 @@ pub struct RootUserSettings {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub ssh_public_key: Option<String>,
 }
+
+impl RootUserSettings {
+    pub fn skip_export(&self) -> bool {
+        self.password.is_none() && self.hashed_password.is_none() && self.ssh_public_key.is_none()
```
jreidinger, line 89: @imobachgs sadly here we have identical problem to first user with hashed password...but here it is even more tricky as password and hashed_password are always skipped during deserializing, so unless ssh_public_key is set, then it is always empty....but on other hand we want somehow indicate user that they need to set it in root user.
So I am not sure how to proceed here. Of course I can change it to something like `self.hashed_password == Some(false)`

jreidinger: btw if hashed_password is set to true, should we really omit serialization of password to profile?

rust/agama-lib/src/users/store.rs
```diff
@@ -43,21 +44,32 @@ impl UsersStore {
...
-            user_name: Some(first_user.user_name),
-            full_name: Some(first_user.full_name),
-            password: Some(first_user.password),
+            user_name: optional_string(first_user.user_name),
+            full_name: optional_string(first_user.full_name),
+            password: optional_string(first_user.password),
+            // sadly for boolean we do not have on dbus way to set it to not_set
             hashed_password: Some(first_user.hashed_password),
         };
+        let first_user = if first_user.skip_export() {
```
imobachgs, line 53: Perhaps we could extend the D-Bus API with information about whether the user or the root was set instead of checking many conditions here.

dgdavid:
@imobachgs, @jreidinger

Please, excuse me for the noise since this comment is not strictly related to the changes proposed here, but they somehow reminded me a concern about optional fields I found when reworking authentication forms in the web interface: an empty string has to be sent for unset a value, see

https://github.com/agama-project/agama/blob/a4e41691642bb950fbec0dce36ad904090811448/web/src/components/users/RootUserForm.tsx#L120-L122

Just in case this rings a bell now that you're touching the users code. It might be worth creating a card with the details to address this behavior in the future. Sending empty values to represent 'no value' feels a bit awkward.

jreidinger:
@dgdavid yeap, it basically exposing dbus API which use empty string as unset. So it can be also addressed, I am just not sure if it is better to send patch, delete or something else ( I think json allows even sending null ).

imobachgs:
Replaced with #2462.

PR #2239, opened 2025-04-01
feat(autoinstallation): Make the addon version optional

lslezak:
## Problem

- The addon version is required in the autoinstallation profile, i.e. you have to pass
  ```json
  {
    "id": "sle-ha",
    "version": "16.0",
    "registrationCode": "<CODE>"
  }
  ```
- For most addons from SCC the version is bound to the SP release, e.g. in SP1 the version will be "16.1", in SP2 "16.2", etc...
- That means you have to update the profile for each SP release and you cannot use the same profile for multiple SP releases. That's quite annoying.

## Solution

- Make the version attribute optional, allow it to be missing.
- If the version is missing query the available addons and find the version to register automatically from the list.
- If multiple versions are found report an error, in that case the version needs to be specified explicitly in the profile.
- So now you can just write:
  ```json
  {
    "id": "sle-ha",
    "registrationCode": "<CODE>"
  }
  ```
  For free extensions you will only need the `id` value.

## Notes

- The installer remembers whether a specific version was requested or it was found automatically.
- It uses this information when exporting the current configuration. When the version was specified at import it will be specified in the export, if it was missing at import it will be missing in export as well.
- That means at export you should get the same imported data.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/rust/agama-lib/src/product/client.rs
+++ b/rust/agama-lib/src/product/client.rs
@@ -135,7 +135,11 @@ impl<'a> ProductClient<'a> {
             .into_iter()
             .map(|(id, version, code)| AddonParams {
                 id,
-                version,
+                version: if version.is_empty() {
+                    None
+                } else {
+                    Some(version)
+                },
                 registration_code: if code.is_empty() { None } else { Some(code) },
             })
             .collect();
@@ -158,7 +162,7 @@ impl<'a> ProductClient<'a> {
             .registration_proxy
             .register_addon(
                 &addon.id,
-                &addon.version,
+                &addon.version.clone().unwrap_or_default(),
                 &addon.registration_code.clone().unwrap_or_default(),
             )
             .await?)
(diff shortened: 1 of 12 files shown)


rust/agama-lib/src/product/client.rs
```diff
@@ -135,7 +135,11 @@ impl<'a> ProductClient<'a> {
             .into_iter()
             .map(|(id, version, code)| AddonParams {
                 id,
-                version,
+                version: if version.is_empty() {
```
imobachgs, line 138: [This function](https://github.com/agama-project/agama/pull/2238#discussion_r2022666760) could be useful in this case. However, it is not included yet.

rust/agama-lib/src/product/client.rs
```diff
@@ -158,7 +162,7 @@ impl<'a> ProductClient<'a> {
             .registration_proxy
             .register_addon(
                 &addon.id,
-                &addon.version,
+                &addon.version.clone().unwrap_or_default(),
```
imobachgs, line 165: Would an empty version work?

imobachgs (approved):

rust/agama-lib/src/product/client.rs
lslezak: Yes, in the end the backend receives empty version because `nil` cannot be sent over DBus. So omitting the version is the same as using empty string.

And if SCC really uses an empty version then it will work too, just using a slightly more complicated path in the code.

PR #2241, opened 2025-04-02
Reorganize and redefine some storage menus

ancorgs:
## Problem

We recently added support for LVM to the storage section of the web interface. But we lacked time to polish some details like:

- It's too easy to end up with a configuration without drives as a result of deleting a volume group.
- Moving the initial partition-based setup to an LVM is not as obvious as we wanted it to be.

Additionally, the current interface is problematic when the name of a disk is too long (eg. multipath   devices) or when there are too many disks to choose from.

## Solution

This reorganizes the menus a bit to improve the overall situation, although it probably does not fully fix any of the problems in a definitive way.

- First of all, it identifies the disks using their basenames instead of full device names. See more about device names below. In addition, this truncates very long base names to 17 characters at several places (although it uses a wrong approach). These changes do not fully fix the existing problems with long names but mitigates them to some extend.
- This also redefines the action to delete an LVM specification. If possible, the system now will move the LVs as partitions to the underlying disk. If that is not possible (multiple underlying disks) or there are no LVs defined, then some sensible deletion of the underlying drives comes into place. That will never lead to the deletion of drives that are used for other purposes or to a setup with no drives.
- The menu to change the target of a drive was extracted to a new component `DriveDeviceMenu` and its options were reorganized:
  - Now the option to change the disk is isolated into its own submenu. That's more clear and should also help to handle scenarios with many disks in the future.
  - An option to create an LVM in the disk was added. That should ease the creation of LVM-based setups without the user visiting the "more devices" menu.

## Implementation details

- The new `DriveDeviceMenu` and the already existing `ConfigureDeviceMenu` components are now based on new `core/MenuButton` component. More menus could be adapted in the future.

## Pending

- Maybe: Add new unit tests

## About device names

Most disks have a device name like `/dev/sda`  or `/dev/nvme0n1`. But if multipath is activated with its default configuration, every disk will get a name that is based on a world-wide unique [WWID](https://en.wikipedia.org/wiki/World_Wide_Name). See some examples of disks and partition names taken from a real system in which multipath was activated:

```
/dev/mapper/360050768028383d7f000000000000195
/dev/mapper/360050768028383d7f000000000000196
/dev/mapper/360050768028383d7f000000000000197
/dev/mapper/360050768028383d7f000000000000198
/dev/mapper/360050768028383d7f000000000000199
/dev/mapper/360050768028383d7f00000000000019a
/dev/mapper/360050768028383d7f000000000000195-part1
/dev/mapper/360050768028383d7f000000000000195-part2
/dev/mapper/360050768028383d7f000000000000195-part3
/dev/mapper/360050768028383d7f000000000000195-part4
/dev/mapper/360050768028383d7f000000000000196-part1
/dev/mapper/360050768028383d7f000000000000196-part2
```


web/src/helpers/storage/volume-group.ts
```diff
@@ -63,6 +64,29 @@ function addVolumeGroup(
...
+function newVgName(apiModel: apiModel.Config): string {
+  const vgs = (apiModel.volumeGroups || []).filter((vg) => vg.vgName.match(/^system\d*$/));
+
+  if (!vgs.length) return "system";
+
+  const numbers = vgs.map((vg) => parseInt(vg.vgName.substring(6)) || 0);
+  return `system${Math.max(...numbers) + 1}`;
+}
+
+function driveToVolumeGroup(apiModel: apiModel.Config, driveName: string): apiModel.Config {
```
joseivanlopez, line 76: NP: shouldn't this be in a drive helper?

ancorgs:
The part about avoiding the hardcoded "20" for the truncates is just partially done. At the time of writing this comment there are another 4 occurrences that were overlooked (at several components like `DriveDeviceMenu`, `LvmPage`, `PartitionPage` or `VolumeGroupEditor`.

I will try to commit an alternative solution.

web/src/components/storage/ConfigureDeviceMenu.tsx
```diff
@@ -79,41 +68,32 @@ const DisksDrillDownMenuItem = ({
...
-                      {s}
-                    </Label>
-                  ))}
-                </Flex>
-              </Split>
-            </MenuItem>
-          ))}
-        </DrilldownMenu>
-      }
+      upProps={{ label: title }}
```
ancorgs, line 75: I like having the original label as some kind of "header" for the drilled menu. It allows to retain the context. But...

I wonder whether we should also display the description of whether that would be too much.

I'm not sure whether we should still show the word "back" somewhere. Clicking on "select disk" to actually go back from that action may not be optimal.

ancorgs: Restored the old "back" word. It was agreed on a video call.

ancorgs:
> I will try to commit an alternative solution.

Done.

web/src/components/core/MenuButton.tsx
```diff
@@ -0,0 +1,204 @@
...
+    props = { itemId, ...props };
+    return <MenuItem {...props}>{children}</MenuItem>;
+  }
+
+  props = { itemId, ...props, onClick: onClick(props.onClick) };
+
+  return (
+    <MenuItem
+      {...props}
+      direction="down"
```
dgdavid, line 78: Why not destructuring the onClick instead?

```suggestion
export function MenuButtonItem({
  items = [],
  upProps = { label: _("Back") },
  onClick,
  children,
  ...props
}: React.PropsWithChildren<MenuButtonItemProps>): React.ReactNode {
  const itemId = useId();
  const menuId = useMenuId();
  const onItemClick = (event) => {
    event.stopPropagation();
    onClick && onClick(event);
  };

  if (!items.length) {
    return <MenuItem itemId={itemId} {...props}>{children}</MenuItem>;
  }
  return (
    <MenuItem
      itemId={itemId}
      {...props}
      onClick={onItemClick}
      direction="down"
```

I also took the opportunity for not overwriting props and avoid destructuring it twice.

In any case, I have no strong opinion here, although I feel getting ride of the callback thing make it easier to follow. Let me know if I'm missing something.

web/src/components/core/MenuButton.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  menuProps = { popperProps: {}, closeOnClick: true, ...menuProps };
+
+  const resetState = () => {
+    setMenuDrilledIn([]);
+    setDrilldownPath([]);
+    setActiveMenu(rootId);
+    // Reset the height of the root menu. Otherwise, the memorized height could be wrong if the
+    // items of the root menu change.
+    if (!isOpen) setMenuHeights({ ...menuHeights, [rootId]: undefined });
```
dgdavid, line 128: :ok_hand:

dgdavid (approved):
Left just a couple of minor notes, but overall it looks really good.

Thanks for handling the creation of the MenuButton core component. It looks solid and should help a lot with centralizing the code for building PatternFly menus more easily.

web/src/components/core/MenuButton.tsx
joseivanlopez: I like your proposal. Please, apply it ;-).

joseivanlopez (commented):
Changelog is missing.

PR #2245, opened 2025-04-03
use for all services which makes sense in headless mode target multi-user

jreidinger:
## Problem

Almost all agama systemd services uses default.target. This has two drawbacks:

1. it is discouraged by systemd itself - https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#default.target ( `For typical unit files please set "WantedBy=" to a regular target (like multi-user.target or graphical.target), instead of default.target, since such a service will also be run on special boots like on system update, emergency boot…`
2. it means that user wants to run installer in headless mode and use standard `systemd.unit=multi-user.target` it does not start agama and other services as live iso has as default graphical.

## Solution

change services to use multi-user everywhere where it does not need graphical target ( so all of them as only x11 autologin is for graphical )

## Testing

- *Tested manually* - both with default and with headless setup

### DIFF
--- a/autoinstallation/systemd/agama-auto.service
+++ b/autoinstallation/systemd/agama-auto.service
@@ -13,4 +13,4 @@ Type=simple
 User=root
 
 [Install]
-WantedBy=default.target
+WantedBy=multi-user.target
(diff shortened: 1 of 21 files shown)


lslezak (approved):
LGTM

But I think we should document it somewhere how to start headless (remote or CLI only) installation.

PR #2246, opened 2025-04-04
Fix condition for multiple elements in bootloader

jreidinger:
A bit of details. If allOf is used, then it is correct also if just single one is used, but intention is to express that one of given parameters is specified or none.


dgdavid (approved):
Approving it assuming manual testing, although I still thinking would be better to have unit testing to ensure the expected behavior.

PR #2252, opened 2025-04-08
Fix efi boot

jreidinger:
## Problem

Failed to boot from hard disk for EFI boot.

bsc: https://bugzilla.suse.com/show_bug.cgi?id=1240646

## Solution

Add `suse` as EFI boot entry which is used by SLES16.


imobachgs (approved):

kobliha (approved):
LGTM, thanks!

PR #2253, opened 2025-04-08
fix(web): CLI do not crash if it cannot mount a file system

imobachgs:
It might happen that, for some reason, "agama download" cannot mount a file system (e.g., `/dev/loop1`). In that case, it should continue with the next device to probe.


dgdavid (approved):
I guess you have manually tested it. If it's cheaper, would be nice to have a unit test too. In any case, let's go ahead with the merge, thanks.

joseivanlopez (approved):

Issue #2240, opened 2025-04-02
Dancing components in the screen


alexandrevicenzi:
The components keep moving in the screen text display while showing the packages being installed.

For example, here, each stage has one size:

![Image](https://github.com/user-attachments/assets/2d8c5827-30cb-4b59-ac76-329c82226266)

But here, the size is smaller:

![Image](https://github.com/user-attachments/assets/4f881fea-a93c-4e28-9a41-6038a29345ba)

The size should be fixed, and components should not be dancing around the screen while packages are being installed. If the package name is too long, use an ellipsis.

dgdavid:
Thanks for reporting it. It's actually a known issue, but is good to have it tracked.

> If the package name is too long, use an ellipsis.

It's not that easy. In fact, your first screenshot already shows the progress using an ellipsis. In any case, we will actually to fix/rework/rethink the whole approach. Stay tuned :)

alexandrevicenzi:
I was not aware of the Trello board. I did look over older issues in GH, but I did not find any.
I'm aware that it might not be an easy fix, UI is never that easy.
This issue is a small annoyance, nothing major that prevents usage.

dgdavid:
Let's see if #2255 really fixes the problem until we have time for re-work the whole progress screen 😉 

Thanks again for reporting the issue.

PR #2255, opened 2025-04-09
fix(web): mitigate flickering at progress and loaders

dgdavid:
## Problem

The interface experiences flickering when transitioning between loading states or from one substep to another within a progress stepper. This causes a noticeable "dancing" effect, disrupting the user experience.

## Solution

To mitigate the flickering effect, improvements have been made to the involved components, bearing in mind that more significant updates or a full redesign may be necessary in the future.

In brief, the `ProgressStepper` now has a fixed full width and also sets a fixed height for its steps, reducing layout shifts. The loading indicator has been centered more effectively, and the default text has been removed to minimize the perception of movement. This change addresses the issue of three distinct loading states in the application, triggered at different stages (without layout, with layout but no header/sidebar, and within a layout that includes both the header and possibly a sidebar, depending on the available width).

Additionally, removing the default text helps prevent its misuse. For example, the text "Loading installation environment" was displayed even after the environment was set up and the application was actually waiting for installation settings.

## Testing

* Updated affected and existing unit tests.
* Tested manually.

## Screenshots

N/A

It is better to test manually to see changes in action.

## Documentation
N/A

---

Related to https://trello.com/c/nvTb8WwM (private link)
Fixes https://github.com/agama-project/agama/issues/2240


web/src/components/layout/Loading.tsx
```diff
@@ -31,11 +31,37 @@ const Layout = (props: LayoutProps) => (
   <PlainLayout mountHeader={false} mountSidebar={false} {...props} />
 );
 
+type LoadingProps = {
+  /** Text to be rendered alongside the spinner */
+  text?: string;
+  /**
+   * Whether the loading screen should listen for and render any questions
```
joseivanlopez, line 45: NP: listen for or listen to?

joseivanlopez (approved):

joseivanlopez:
Approved, but there is a conflict ;-)

web/src/components/layout/Loading.tsx
dgdavid: You know English isn't really my thing ;) But from what I understand, in this context it’s better to use _listen for_

* https://www.collinsdictionary.com/dictionary/english/listen (entries 2 and 3)
* https://ell.stackexchange.com/questions/118448/listen-to-or-listen-for
* https://hinative.com/questions/4216094

dgdavid:
> Approved, but there is a conflict ;-)

Ah, I see. You were faster :stuck_out_tongue_winking_eye: 

Conflict fixed. Let me know if I can merge it now.

PR #2256, opened 2025-04-10
feat(live): add a kiwi variant which includes only the minimal software to run Agama

imobachgs:
## Problem

The PXE image is too big and it is not possible to use it on a system with 2GiB of RAM. It must include a minimal desktop and a browser, which takes quite some space.

However, in many situations the system can be installed remotely or using the CLI. For those scenario, we should offer an smaller PXE image that does not include the desktop stack.

## Solution

Split the main list of packages in two:

* Mandatory packages (basic system, kernel, Agama, etc.).
* Graphical stack (X.org, IceWM, etc.).

Additionally, the MINI image does not contain:

* GPU kernel drivers.
* Wi-Fi and Bluetooth drivers.

> [!WARNING]
> This PR conflicts with the switch to Wayland (#1868), but it can be adapted later. The same approach still applies.


lslezak:
My changes reduced the mini PXE image from 419MB to 343MB (x86_64), mostly by dropping GPU, Wifi and Bluetooth drivers.

live/src/config.sh
```diff
@@ -61,11 +61,16 @@ systemctl enable checkmedia.service
 systemctl enable qemu-guest-agent.service
 systemctl enable setup-systemd-proxy-env.path
 systemctl enable x11-autologin.service
-systemctl enable spice-vdagentd.service
+test -f  /usr/lib/systemd/system/spice-vdagentd.service && systemctl enable spice-vdagentd.service
```
jreidinger, line 64: I kind of expect that identical thing should happen to qemu-guest-agent..and also x11-autologin

jreidinger (approved):

PR #2257, opened 2025-04-10
Copy files needed for NVMe from instsys to the target

mchf:
## Problem

NVMe devices not properly discovered after installation and reboot to the target

[bsc#1238038](https://bugzilla.suse.com/show_bug.cgi?id=1238038)

## Solution

Copy needed pieces of the configuration to the target

## Testing

- *Tested manually*
- updated testsuite

### DIFF
--- a/service/lib/agama/storage/finisher.rb
+++ b/service/lib/agama/storage/finisher.rb
@@ -130,7 +130,9 @@ module Agama
         FILES = [
           { dir: "/etc/udev/rules.d", file: "40-*" },
           { dir: "/etc/udev/rules.d", file: "41-*" },
-          { dir: "/etc/udev/rules.d", file: "70-persistent-net.rules" }
+          { dir: "/etc/udev/rules.d", file: "70-persistent-net.rules" },
+          { dir: "/etc/nvme", file: "hostnqn" },
+          { dir: "/etc/nvme", file: "hostid" }
         ].freeze
 
         def label
--- a/service/lib/agama/storage/finisher.rb
+++ b/service/lib/agama/storage/finisher.rb
@@ -145,15 +145,22 @@ module Agama
         end
 
         def run
-          target = File.join(Yast::Installation.destdir, UDEV_RULES_DIR)
-          FileUtils.mkdir_p(target)
-          FileUtils.cp(glob_files, target)
+          glob_files.each do |file|
+            target = File.dirname(file).sub(root_dir, dest_dir)
+
+            FileUtils.mkdir_p(target)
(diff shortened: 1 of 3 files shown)


service/lib/agama/storage/finisher.rb
```diff
@@ -152,7 +154,8 @@ def root_dir
         end
 
         def glob_files
-          Dir.glob(FILES.map { |f| File.join(root_dir, f[:dir], f[:file]) })
+          paths = FILES.map { |f| File.join(root_dir, f[:dir], f[:file]) }
+          Dir.glob(paths.reject { |p| Dir.glob(p).empty? })
```
ancorgs, line 158: Is this really needed? I think glob already will filter any not found file.

```ruby
Dir.glob(["/etc", "/not/existing"])  # => ["/etc"]
```

ancorgs (commented):
In general, this looks good. Just some trivial requests on my side:

- Squash all the current commits into one
- Add some pointer somewhere to [bug#1238038](https://bugzilla.suse.com/show_bug.cgi?id=1238038), either in a comment next to the filenames or in the commit message. So we can track why stated to copy these files.
- Add changelog entry and version bump

mchf:
> In general, this looks good. Just some trivial requests on my side:
> 
> * Squash all the current commits into one
> * Add some pointer somewhere to [bug#1238038](https://bugzilla.suse.com/show_bug.cgi?id=1238038), either in a comment next to the filenames or in the commit message. So we can track why we started to copy these files.
> * Add changelog entry and version bump

Updated & rebased. However I'm not sure about raising the version ... you mean in rubygem*.spec 12.devel16 -> 12.devel17 ?

ancorgs (approved):
LGTM

mchf:
There was a bug in destination directory (was hardcoded) ... fixed & updated testsuite to cover that

ancorgs (approved):
LGTM

service/lib/agama/storage/finisher.rb
```diff
@@ -142,9 +145,12 @@ def run?
         end
 
         def run
-          target = File.join(Yast::Installation.destdir, UDEV_RULES_DIR)
-          FileUtils.mkdir_p(target)
-          FileUtils.cp(glob_files, target)
+          glob_files.each do |file|
+            target = File.dirname(file).sub(root_dir, dest_dir)
```
ancorgs, line 149: This relies in the fact that `dest_dir` includes a final slash. That is, is "/mnt/". It would not work if `dest_dir` is "/mnt" (without the trailing slash). But I guess we can live with that.

PR #2259, opened 2025-04-10
fix(rust): do not freeze when trying to use profiles in OEMDRV

imobachgs:
## Problem

The `agama-auto` service does not work after introducing some changes to make it possible to use the `agama profile` command from an remote system (see #2103). The call to the `POST /api/profile/autoyast` endpoint freezes the whole server.

## Solution

* Make the code that takes care of importing the AutoYaST profile async (`AutoyastImporter::read`).
* Temporarily disable AutoYaST fetch errors because now they are not properly handled in the `OEMDRV` feature (it does not pass the `YAST_SKIP_PROFILE_FETCH_ERROR` env-var). Consider this a temporary workaround to have a working development image again.

### DIFF
--- a/rust/agama-lib/src/profile.rs
+++ b/rust/agama-lib/src/profile.rs
@@ -23,12 +23,7 @@ use anyhow::Context;
 use jsonschema::JSONSchema;
 use log::info;
 use serde_json;
-use std::{
-    fs::{self, File},
-    io::Write,
-    path::Path,
-    process::Command,
-};
+use std::{fs, io::Write, path::Path, process::Command};
 use tempfile::{tempdir, TempDir};
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


mvidner:
OK, this means that my testing was not sufficient in #2103.
Please describe how you trigger the problem that this fixed by this.
Hopefully we can add a test into `profile_itest.rb`

imobachgs:
> OK, this means that my testing was not sufficient in #2103. Please describe how you trigger the problem that this fixed by this. Hopefully we can add a test into `profile_itest.rb`

You need to execute the `agama-auto` service to detect the problem. In this case, the easiest way is to just build a new image with the updated code. You can use the [development scripts](https://github.com/agama-project/agama/tree/master/devel) to do that.

mvidner (approved):
I see. We should convert the environment variable into a CLI switch, but this is good for a quick fix.

And yay for the elegance that `async` provides!

imobachgs:
> I see. We should convert the environment variable into a CLI switch, but this is good for a quick fix.

I agree. Adding a command switch would be a good option.

> And yay for the elegance that `async` provides!

😃

mvidner:
I see the environment variable was introduced in https://github.com/agama-project/agama/pull/2180 (set) and https://github.com/yast/yast-autoinstallation/pull/882 (checked)
Well it is a hacky design... and now I see another thing that has stopped working: `export YAST_SKIP_XML_VALIDATION=1`. Bypassing API layers like this is error prone when doing any refactoring

PR #2262, opened 2025-04-11
add bootloader as mandatory pattern for SLES 16.0

fcrozat:
Ensure bootloader pattern is always installed on SLES 16.0 (bsc#1240015).


jreidinger (approved):

jreidinger:
@fcrozat should it be mandatory also for SAP?

fcrozat:
> @fcrozat should it be mandatory also for SAP?

yes, I forgot it :(

PR #2268, opened 2025-04-14
fix(web): make hostname strings fully translatable

dgdavid:
## Problem

@ancorgs detected and reported some not translatable strings at HostnamePage.tsx

>  \<ancorgs\> dgdavid: retrospective code review. These strings are not translatable https://github.com/agama-project/agama/blob/master/web/src/components/system/HostnamePage.tsx#L66

## Solution

Make them translatable and take the opportunity for avoid the usage of "please" word according to SUSE documentation style guide, https://documentation.suse.com/style/current/single-html/docu_styleguide/index.html#sec-vocabulary

## Notes

A new `setRequestError` has been added for simplicity, since it make no sense to render something like 

> **Check the following before continuing**
> 
> Hostname could not be updated.

Even when such an error will likely not happen. So, better to go for

> Hostname could not be updated

The presence or not presence of the dot at the end of the sentence it is also intentional: it's not used in titles

Of course, it is needed a better error reporting for forms to easily distinguish between validation warnings and request errors, but sadly Agama is not there at the time of sending this hotfix to make possible the translation of these texts. Thus, skipping further improvements in this form until such a goal can be achieved.

### DIFF
--- a/web/src/components/system/HostnamePage.tsx
+++ b/web/src/components/system/HostnamePage.tsx
@@ -46,6 +46,7 @@ export default function HostnamePage() {
   const hasTransientHostname = isEmpty(staticHostname);
   const [success, setSuccess] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
+  const [requestError, setRequestError] = useState<string | null>(null);
   const [settingHostname, setSettingHostname] = useState(!hasTransientHostname);
   const [hostname, setHostname] = useState(staticHostname);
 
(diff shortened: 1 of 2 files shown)


web/src/components/system/HostnamePage.tsx
```diff
@@ -99,6 +99,7 @@ export default function HostnamePage() {
 
         <Form id="hostnameForm" onSubmit={submit}>
           {success && <Alert variant="success" isInline title={success} />}
+          {requestError && <Alert variant="warning" isInline title={requestError} />}
```
dgdavid, line 102: Not using `error` here, which would be more precise, because it is not being used in other forms at the time of writing. As said in the PR description, we're aware that better form validation and error reporting is needed but they should be addresses all together once we find a way to do so.

ancorgs (approved):
LGTM

PR #2269, opened 2025-04-14
Fix iBFT detection and rearrange iSCSI initiator page

ancorgs:
## Problem

Although the iBFT configuration was being properly applied if found, the web user interface was always displaying "iBFT: no". See [bsc#1239046](https://bugzilla.suse.com/show_bug.cgi?id=1239046).

Even if we can say that was just a cosmetic issue, it is fair to say the original interface was not helping to make things understandable.

See the following screenshot in which several things are wrong:

fake

- As mentioned, the value of the column "iBFT" of the initiator should be "yes".
- The action to edit the initiator name was always available, even for initiators configured via iBFT.
- The column "Offload card" is obsolete, YaST does not longer use such a field and it has never been possible to configure that via AutoYaST.

Last, but not least, that interface was not consistent with the design principles used on the most modern sections of the Agama interface.

## Solution

First of all, this pull request fixes detection of iBFT for the user interface. So now the interface can actually tell whether iBFT is in use.

On top of that, this pull request reorganizes the interface a bit showing this in case iBFT is in use (slightly outdated screenshot).

ibft

And this in case iBFT is not detected.

ibfnot

That is more aligned with the current interface and capabilities of Agama and should help reducing confusion.

## Testing

- Tested manually by mocking the system (I don't have iBFT devices at hand)
- Adapted unit tests


web/src/components/storage/iscsi/InitiatorSection.tsx
```diff
@@ -24,17 +24,46 @@ import React from "react";
...
+  DescriptionListGroup,
+  DescriptionListTerm,
+  DescriptionListDescription,
+} from "@patternfly/react-core";
+import { InitiatorForm } from "~/components/storage/iscsi";
+import { useInitiator, useInitiatorMutation, useInitiatorChanges } from "~/queries/storage/iscsi";
+
+const InitiatorDescription = ({ initiator }) => {
+  return (
+    <DescriptionList aria-label={_("Initiator details")} isHorizontal>
```
dgdavid, line 38: I wonder how it looks if using "isFluid" property.

web/src/components/storage/iscsi/InitiatorSection.tsx
```diff
@@ -24,17 +24,46 @@ import React from "react";
...
   useInitiatorChanges();
 
+  const submitForm = async ({ name }) => {
+    await updateInitiator({ name });
+  };
+
+  const desc = initiator.ibft
+    ? _("Configuration read from the iSCSI Boot Firmware Table (iBFT).")
+    : _(
+        "No iBFT (iSCSI Boot Firmware) configuration found. The initiator can be configured manually.",
```
dgdavid, line 59: NP: I'd keep the naming equal/stable in both branches. I.e.,

"No **iSCSI Boot Firmware Table (iBFT)** found. The initiator name can be configured manually"

dgdavid (approved):
Thanks for taking care of this beyond the proper iBFT.

I just left two minor comments there, but LGTM.

dgdavid (commented):
This change deserves an entry in changes files,    right?

service/lib/agama/storage/iscsi/adapter.rb
```diff
@@ -67,8 +67,7 @@ def read_initiator
           Initiator.new.tap do |initiator|
             initiator.name = Yast::IscsiClientLib.initiatorname
             initiator.offload_card = Yast::IscsiClientLib.GetOffloadCard()
```
joseivanlopez, line 69: Could the `#offloard_card` be removed from the `Initiator` class?

web/src/components/storage/iscsi/InitiatorSection.tsx
ancorgs: Better, indeed.

service/lib/agama/storage/iscsi/adapter.rb
ancorgs: Removed.

PR #2271, opened 2025-04-15
feat(rust): replace --api option by --host

joseivanlopez:
## Problem

The CLI allows connecting to a different server by using the `--api <URL>` option, but that URL differs from the URL you use in the browser (*https://localhost* vs *https://localhost/api*). Ideally, the same URL should be used for both CLI and browser clients.

## Solution

The `--api` option is replaced by `--host`, which allows indicating a host name or IP address without the */api* suffix. The protocol is optional, using *https* if omitted. The default host is *http://localhost*.

Examples:

~~~
$ agama --host https://my-server.lan config show
$ agama --host my-server.lan config show
$ agama --host 192.168.100.101 config show
~~~

## Documentation

mvidner:
The `--api` option is only once mentioned in a blog post, nothing to adjust IMHO.
Put the examples directly in the --help` output (long only, not for `-h`)

### DIFF
--- a/rust/agama-cli/src/lib.rs
+++ b/rust/agama-cli/src/lib.rs
@@ -58,10 +58,9 @@ use url::Url;
 /// Agama's CLI global options
 #[derive(Args)]
 pub struct GlobalOpts {
-    #[clap(long, default_value = "http://localhost/api")]
-    /// URI pointing to Agama's remote API. If not provided, default https://localhost/api is
-    /// used
-    pub api: String,
+    #[clap(long, default_value = "http://localhost")]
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

mvidner:
This will break my use case of connecting to the same host but different port (forwarded to a container)

mvidner (changes requested):
> This will break my use case of connecting to the same host but different port (forwarded to a container)

OK, false alarm, it will continue to work, but the description did not make it clear.

But let me put the examples where the *user* will see them.

PR #2272, opened 2025-04-16
fix(rust): document finish options

joseivanlopez:
## Problem

The explanation about the possible options of the `agama finish` command was manually added to the Agama CLI documentation, see https://github.com/agama-project/agama-project.github.io/pull/50/files. This implies that:

* `agama finish --help` does not contain such as explanation.
* The documentation has to be manually restored after updating `cli.md` with the `cargo xtask markdown` output.

## Solution

Extend the `agama finish` help.

### DIFF
--- a/rust/agama-cli/src/commands.rs
+++ b/rust/agama-cli/src/commands.rs
@@ -114,7 +114,7 @@ pub enum Commands {
     Finish {
         /// What to do after finishing the installation. Possible values:
         ///
-        /// stop - do not reboot and the system continues running.
+        /// stop - do not reboot and the Agama backend continues running.
         ///
         /// reboot - reboot into the installed system.
         ///
--- a/rust/agama-cli/src/commands.rs
+++ b/rust/agama-cli/src/commands.rs
@@ -110,14 +110,17 @@ pub enum Commands {
         /// File name
         destination: PathBuf,
     },
-    /// Finish the installation rebooting the system by default.
-    ///
-    /// This command finishes the installation by performing a set of tasks and rebooting the system by
-    /// default if the installation is completed successfully.
-    ///
-    /// Optionally an argument can be given for determining whether the system should be rebooted,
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

mvidner (commented):
When I looked at `cli.md` I thought "Oh this must be hard to keep in sync", too bad that there is nothing to help with that"
So now I know of `cargo xtask markdown` but it is only because this PR mentions it. It is not discoverable.
`cli.md` should contain a comment like
> This documentation is generated. Run `cargo xtask markdown` to update it.

rust/xtask/src/main.rs
```diff
@@ -40,6 +43,7 @@ mod tasks {
         let filename = out_dir.join("agama.md");
         let mut file = File::create(&filename)?;
         file.write_all(markdown.as_bytes())?;
+        file.write_all(GENERATED.as_bytes())?;
```
joseivanlopez, line 46: Shouldn't it be placed before the markdown content?

mvidner: I haven't checked how that would look like, before the title. Feel free to move it.

joseivanlopez: It woud not be a valid md, so it is ok now.

rust/agama-cli/src/commands.rs
```diff
@@ -110,14 +110,17 @@ pub enum Commands {
...
     Finish {
+        /// What to do after finishing the installation. Possible values:
+        ///
+        /// stop - do not reboot and the Agama backend continues running.
+        ///
+        /// reboot - reboot into the installed system.
+        ///
+        /// halt - halt the installed machine.
+        ///
+        /// poweroff - poweroff the installed machine.
```
ancorgs, line 123: "power off the installed machine". Poweroff as a single word is not a verb.

ancorgs (approved):
LGTM

Issue #2261, opened 2025-04-11
CLI does not allow logging in to multiple systems


imobachgs:
Agama's CLI authentication mechanism does not allow to stay logged in different systems at the same time. What is worse, if you are authenticated on system A and try to authenticate on system B, you get a cryptic error message like this:

```
$ agama --api 'https://192.168.1.49/api' auth login
> Please enter the root password: ********
Error: Permission denied (os error 13)
```

You need to log out from the first server to make this work, which is not what a user would expect.

imobachgs:
Fixed with #2278.

PR #2278, opened 2025-04-21
feat(rust): allow authenticating into multiple hosts

imobachgs:
## Problems

* #2261.
* Any connection problem is reported as an certificate problem when it is not true. To make things worse, Agama asks the user whether to accept an insecure connection (even if that's not the problem).
* Agama will perform an HTTP request even if it is not needed (e.g., `agama auth show`).

### Storing one token per host

Instead of storing the token in `$HOME/.agama/token`, use a file (`$HOME/.local/share/agama`) containing all tokens (one per host). For instance:

```json
{
  "192.168.122.145": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE3NDUzMTg2MTV9.G36w8NQ-ZCAkCQGGwjajRpd8EfkOM8fFLLfGjzTAqP0"
}
```

### Do not ask for insecure connections

If there is a problem when connecting to the API, just report the error (no matter whether it is an SSL problem or not). The user can use the `--insecure` flag to disable certificate checks.

## Testing

- Added a new unit test
- Tested manually


teclator (approved):
LGTM

ancorgs (approved):
I can confirm the conflict was properly resolved. I didn't check the rest, but I will rely on previous @teclator approval to say that this LGTM.

rust/agama-lib/src/error.rs
```diff
@@ -37,7 +37,7 @@ pub enum ServiceError {
     DBusProtocol(#[from] zbus::fdo::Error),
     #[error("Unexpected type on D-Bus '{0}'")]
     ZVariant(#[from] zvariant::Error),
-    #[error("Failed to communicate with the HTTP backend '{0}'")]
+    #[error(transparent)]
```
mvidner, line 40: Neat!

PR #2286, opened 2025-04-23
feat(storage): initial support for MD RAIDs

joseivanlopez:
Add support for creating and reusing MD RAID devices (UI is not covered yet).


rust/agama-lib/share/storage.schema.json
```diff
@@ -39,7 +44,7 @@
       ]
     },
     "formattedDrive": {
-      "description": "Drive without a partition table (e.g., directly formatted).",
+      "description": "Drive without a partition table (directly formatted).",
```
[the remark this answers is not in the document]
joseivanlopez: Actually, this schema only matches with a formatted drive because "filesystem" property is required. The rest (partitioned, not-partitioned) uses the *partitionedDrive* schema (yes, the name is not 100% accurate).

rust/agama-lib/share/storage.schema.json
```diff
@@ -63,6 +68,89 @@
...
+      "title": "MD parity",
+      "description": "Only applies to raid5, raid6 and raid10",
+      "enum": [
+        "left_asymmetric",
+        "left_symmetric",
+        "right_asymmetric",
+        "right_symmetric",
+        "first",
+        "last",
+        "left_asymmetric_6",
```
ancorgs, line 136: Not sure whether we should support the variants with `_6` for creating new RAIDs.

ancorgs, line 136: Not sure whether we should support the variants with `_6` for creating new RAIDs.

rust/agama-lib/share/storage.schema.json
```diff
@@ -63,6 +68,89 @@
...
+        "raid5",
+        "raid6",
+        "raid10"
+      ]
+    },
+    "mdRaidParity": {
+      "title": "MD parity",
+      "description": "Only applies to raid5, raid6 and raid10",
+      "enum": [
+        "left_asymmetric",
```
ancorgs, line 130: At the man page of `mdadm` the names uses a dash instead of an underscore. Not sure what is best (or whether we should support both).

> The layout of the RAID5  parity  block  can  be  one  of  left-asymmetric,  left-symmetric, right-asymmetric, right-symmetric, la, ra, ls, rs.  The default is left-symmetric.
> 
> It  is also possible to cause RAID5 to use a RAID4-like layout by choosing parity-first, or parity-last.

rust/agama-lib/share/storage.schema.json
ancorgs: According to the man page of `mdadm` and to the documentation of `Y2Storage::MdParity`, those values are there only to transition a RAID5 into RAID6.

service/lib/agama/storage/config_solvers/boot.rb
```diff
@@ -45,74 +45,139 @@ def solve(config)
...
-          device = root_device
+          device = boot_device
           return unless device
 
           device.ensure_alias
           config.boot.device.device_alias = device.alias
         end
 
-        # Config of the drive used for allocating root, directly or inderectly.
+        # Config of the device used for allocating root, directly or inderectly.
```
ancorgs, line 59: Typo: inderectly (yes, I know it was not introduced here, but let's fix it).

rust/agama-lib/share/storage.schema.json
joseivanlopez: In the schema we are using snake_case for the enum values:

~~~
        "id": {
          "title": "Partition id",
          "enum": [
            "linux",
            "swap",
            "lvm",
            "raid",
            "esp",
            "prep",
            "bios_boot"
          ]
~~~

~~~
    "encryptionSwap": {
      "description": "Swap encryptions.",
      "enum": ["protected_swap", "secure_swap", "random_swap"]
    },
~~~

I think it makes sense to use the same format as the rest.

service/lib/agama/storage/config.rb
```diff
@@ -142,6 +140,13 @@ def with_size
         partitions + logical_volumes
       end
 
+      # Configs with configurable partitions.
+      #
+      # @return [#partitions]
+      def with_partitions
```
ancorgs, line 146: I'm a bit hesitant about the methods `with_encryption`, `with_filesystem`, `with_filesystem`, `with_partitions`, etc. I feel they are a loaded gun.

I would feel more comfortable with something like `supporting_xxx` or something similar.

PR #2288, opened 2025-04-23
fix(live): Rename "root" to "live-root" to avoid confusion with "root/root"

lslezak:
## Problem

The Live ISO sources contain `root` directory and `root/root` subdirectory. The top level "root" means `/` in the Live medium, the nested one is `/root`. That is confusing and someone even created `root/.xinitrc` file which is then stored as `/.xinitrc` in the Live system which is wrong. The same file is stored also at the correct location `root/root/.xinitrc` (`/root/.xinitrc` in the Live system).

We should remove the wrong files and rename the top level `root` to avoid mistakes.

## Solution

- Rename the top level `root` to `live-root` so it is more obvious that it is `/` in the Live system and not `/root`.
- Removed duplicate `/root/.mozilla/firefox/profile/user.js.template` and `root/.xinitrc`, the same file is in the correct location at `root/root` (or `live-root/root` now).

## Testing

- Tested manually, the ISO builds fine locally and still boots :smiley:


joseivanlopez (approved):
LGTM

PR #2291, opened 2025-04-24
Added inst.copy_network to kernel cmdline and config server no-auto-default 

teclator:
## Problem

The network configuration is a crucial part of the installation process. In some cases it is even required to start the installation. The problem is that, in some cases, the network configuration used during the installation could differ to the one wanted in the target system.

It will be nice to be able to configure completely the "picture" of the network after the installation as we started to design in YaST but as a primary step at least would be nice to select if the installation network configuration should be kept or not allowing to disable the copy of the connections completely. Currently that is not possible and the running and persistent configuration are always copied to the target system as well as we probe DHCP over all the ethernet devices even when there is some explicit configuration given.

- [bsc#1236885](https://bugzilla.suse.com/show_bug.cgi?id=1236885)
- [bsc#1239777](https://bugzilla.suse.com/show_bug.cgi?id=1239777)
- https://trello.com/c/YqZmKHAs/4639-2-agama-allow-to-switch-between-copy-or-not-the-network-configuration-to-the-target-system
- https://trello.com/c/fyE0VwAQ/4558-3-copy-network-configuration-to-the-installed-system-only-in-some-cases

## Solution

We proposed some changes in the current behavior. 

First of all, only the persistent configuration will be copied to the target system (default), this copy can be omitted using the `inst.copy_network=0` kernel cmdline argument although we are considering to define it as an enum allowing `none, persistent, all and maybe a comma separated name of connections` but just keep it simple by now only supporting 0 or 1.

During the boot process **(dracut)**, if an explicit **NetworkManager** configuration is given through the **'ip='** kernel cmdline option then the running configuration will be copied persistently before switching the rootfs and we will also set the **no-auto-default=*** to not run DHCP over all the ethernet devices when the installer is started.

## Testing

- *Added a new unit test*
- *Tested manually*

Screenshot From 2025-04-24 12-39-08

### DIFF
--- a/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
+++ b/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
@@ -7,3 +7,15 @@
 if [ -e /etc/hostname ]; then
   cp /etc/hostname "$NEWROOT/etc/hostname"
 fi
+
+if [ -e /run/agama/copy_network ]; then
+  if getargbool 1 inst.copy_network; then
+    mkdir -p /run/NetworkManager/conf.d
+    echo '[main]' >/run/NetworkManager/conf.d/00-agama-server.conf
+    echo 'no-auto-default=*' >>/run/NetworkManager/conf.d/00-agama-server.conf
+    echo 'ignore-carrier=*' >>/run/NetworkManager/conf.d/00-agama-server.conf
+
+    mkdir -p "$NEWROOT/etc/NetworkManager/system-connections/"
+    cp /run/NetworkManager/system-connections/* "$NEWROOT/etc/NetworkManager/system-connections/"
+  fi
+fi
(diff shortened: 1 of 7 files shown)


live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
```diff
@@ -7,3 +7,15 @@
 if [ -e /etc/hostname ]; then
   cp /etc/hostname "$NEWROOT/etc/hostname"
 fi
+
+if [ -e /run/agama/copy_network ]; then
+  if getargbool 1 inst.copy_network; then
+    mkdir -p /run/NetworkManager/conf.d
+    echo '[main]' >/run/NetworkManager/conf.d/00-agama-server.conf
```
jreidinger, line 14: change looks good, but I would like to see here documentation *WHY* we create those entries and why we copy all system connections.

jreidinger (approved):

PR #2270, opened 2025-04-15
Self-signed certificate support

jreidinger:
## Problem

Using registration server that uses self-signed certificate is not possible.

## Solution

In this initial pull request implement question for user if certificate should be imported or not.

## Testing

- *Added a new unit test*
- *Tested manually*

json used to invoke popup:
```json
{
  "product": {
    "id": "SLES",
    "registrationCode": "test",
    "registrationUrl": "https://migration-rmt2.qe.nue2.suse.org"
  }
}
```

json to test unattended ( will return product not found on server instead of certificate product as I do not have any rmt with SLES16 support)
```json
{
  "product": {
    "id": "SLES",
    "registrationCode": "test",
    "registrationUrl": "https://migration-rmt2.qe.nue2.suse.org"
  },
  "security": {
	  "sslCertificates": [{
		 "fingerprint": "F6:7A:ED:BB:BC:94:CF:55:9D:B3:BA:74:7A:87:05:EF:67:4E:C2:DB",
		"algorithm": "SHA1"
	  }
	 ]
  }
}

```

## Screenshots

localhost_8080_ (23)

## Documentation

*Remember to look at it from the user's perspective. Yes you have made the compiler happy.*
*But will the **humans** even know about your contribution? Sometimes they cannot miss it,*
*other times they need advertisement and explanation.*

*Look for relevant sections and adjust:*
- <https://agama-project.github.io/> ([source](https://github.com/agama-project/agama-project.github.io/))
- Run: `git ls-files '*.md'`

PR #2292, opened 2025-04-24
refactor(rust): reduce ServiceError (round 1)

imobachgs:
The Rust part of the project was introduced just as a command-line interface. However, it has turned into a full blown HTTP/API server (with a command-line) and we plan to move more code from Ruby to Rust.

We are learning Rust as we go, so some old decisions might need to be revisited. I have started by trying to simplify the [ServiceError](https://github.com/agama-project/agama/blob/70e7d987c9478f1c0ed3d1bf0c851485b9b88d84/rust/agama-lib/src/error.rs#L29-L78) enum, which represents an error but includes variants for many different situation (even unrelated ones). We even have an "anyhow" variant!

This first round includes:

- Defining an error for each HTTP client.
- Defining an error for each store.
- Defining an error for the storage, which includes one variant for each store.
- Remove some `ServiceError` variants.
- Make use of `anyhow` in more places of the binary.

Once we have put everything into a better place, we can decide to simplify the whole thing if needed. Let's go step by step.

## To do

- [x] Test the changes
- [ ] Convert errors with a single variant into structs (?)
- [x] Adapt changes from https://github.com/agama-project/agama/pull/2270.
- [x] Update the changes file

## Follow-up

- There is another whole category errors that are related to D-Bus. Let's move them to their own error and finish the `ServiceError` clean-up.
- The HTTP client constructors do not need to be async.
- The D-Bus clients are the "special cases". Perhaps we should consider moving from DomainHTTPClient to DomainClient (and DomainClient to DomainDBusClient).

> [!WARNING]
> Do not merge before https://github.com/agama-project/agama/pull/2270. We need to fix some conflicts first.


rust/agama-lib/src/bootloader/http_client.rs
```diff
@@ -20,9 +20,16 @@
...
-use crate::ServiceError;
+use crate::{
+    base_http_client::{BaseHTTPClient, BaseHTTPClientError},
+    bootloader::model::BootloaderSettings,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum BootloaderHTTPClientError {
+    #[error(transparent)]
+    HTTP(#[from] BaseHTTPClientError),
```
jreidinger, line 31: using transparent for base looks good to me, I would just probably use Base as enum key instead of HTTP as it is confusing with identical enum vealue from BaseHTTPClientError.

rust/agama-lib/src/bootloader/store.rs
```diff
@@ -20,30 +20,36 @@
...
+use super::{
+    http_client::{BootloaderHTTPClient, BootloaderHTTPClientError},
+    model::BootloaderSettings,
+};
 use crate::base_http_client::BaseHTTPClient;
-use crate::error::ServiceError;
 
-use super::http_client::BootloaderHTTPClient;
-use super::model::BootloaderSettings;
+// FIXME: should we follow this approach more often?
```
jreidinger, line 29: I have to say I quite like this Result shortcut. I would say it is quite idiomatic in rust ecosystem.

rust/agama-lib/src/store.rs
```diff
@@ -21,20 +21,56 @@
...
+    storage::{
+        http_client::{ISCSIHTTPClient, ISCSIHTTPClientError},
+        StorageStore, StorageStoreError,
+    },
+    users::{UsersStore, UsersStoreError},
 };
 
+#[derive(Debug, thiserror::Error)]
+pub enum StoreError {
+    #[error(transparent)]
```
jreidinger, line 45: does it make sense to have it transparent here? I think additional line can help with seeing which part of store failed

rust/agama-lib/src/bootloader/http_client.rs
imobachgs: This variant actually wraps `BaseHTTPClientError` which contains HTTP-related errors (get, post, etc.). The idea is that other errors are more domain-specific.

rust/agama-lib/src/store.rs
imobachgs: Each of the errors already contain a description (e.g., "Error processing storage settings: {0}").

Issue #2280, opened 2025-04-22
The special PXE boot parameters are copied to the installed system


lslezak:
## Problem

When booting the installer via PXE the special PXE boot parameters are copied to the installed system. They should be omitted, they do not make sense there.

## Details

When booting via PXE the `/proc/cmdline` contains these extra options (split into multiple lines for better readability):

```
rd.kiwi.install.pxe
rd.kiwi.install.image=http://example.com/agama-installer.x86_64-14pre.0.0.xz
rd.kiwi.ramdisk
ramdisk_size=2097152
initrd=pxeboot.agama-installer.x86_64-14pre.0.0.initrd
BOOT_IMAGE=pxeboot.agama-installer.x86_64-14pre.0.0.kernel
```

In the installed system the `/boot/grub2/grub.cfg` file contains this line:

```
linux   /boot/vmlinuz-6.12.0-160000.9-default root=UUID=2f34bd48-4862-4776-8831-a0dc274a90ae  ${extra_cmdline} rd.kiwi.install.pxe rd.kiwi.install.image=http://example.com/agama-installer.x86_64-14pre.0.0.xz rd.kiwi.ramdisk ramdisk_size=2097152 initrd=pxeboot.agama-installer.x86_64-14pre.0.0.initrd BOOT_IMAGE=pxeboot.agama-installer.x86_64-14pre.0.0.kernel mitigations=auto quiet security=selinux selinux=1 enforcing=1
```

And all those parameters are visible in `/proc/cmdline` after booting the installed system.

### Notes

- The URL in `rd.kiwi.install.image` depends on the deployment environment and the Kiwi supports several protocols  for fetching the image (like FTP/TFTP/SMB...). Kiwi uses `curl` for downloading the image so basically all protocols supported by curl can be used there.
- The `initrd` and `BOOT_IMAGE` parameters are added by the PXE bootloader itself, they are not part of boot options configured on the PXE server. I'm not sure if the PXE bootloader can be configured to not add these, but at least that's the default behavior and very likely most users will use that configuration.
- The suggested boot parameters are generated by Kiwi and are stored in the `*.append` file included in the built tarball.

jreidinger:
filtering for target system is done at https://github.com/agama-project/agama/blob/master/live/root/usr/bin/kernel-cmdline-conf.sh
and it is tested at https://github.com/agama-project/agama/blob/master/live/test/kernel_cmdline_test.rb

jreidinger:
[...]

PR #2293, opened 2025-04-24
Filter out the special PXE boot parameters

lslezak:
## Problem

- The special PXE boot parameters are passed to the installed system, that's wrong
- https://github.com/agama-project/agama/issues/2280

## Solution

- Enhance the filtering code to ignore the specific PXE boot parameters
- Remove also most of the network settings

## Testing

- The unit test has been updated
- Tested manually

## Screenshots

After booting the installer via PXE the extra parameters can be seen in the `/proc/cmdline` file. But the `/run/agama/cmdline.d/kernel.conf` does not include them and and journal log shows they were ignored:
agama-boot-params

After booting the installed system no PXE boot option is present on the `linux` line:
agama-boot-params-installed

### DIFF
--- a/live/test/fixtures/expected/cmdline
+++ b/live/test/fixtures/expected/cmdline
@@ -1,1 +1,1 @@
- BOOT_IMAGE=/boot/vmlinuz splash=silent mitigations=auto quiet nosimplefb=1
+ splash=silent mitigations=auto quiet nosimplefb=1
--- a/live/live-root/usr/bin/kernel-cmdline-conf.sh
+++ b/live/live-root/usr/bin/kernel-cmdline-conf.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 
 # Script to clean kernel command line from agama specific parameters. Result is later used for bootloader proposal.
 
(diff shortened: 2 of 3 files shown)


live/test/fixtures/expected/cmdline
```diff
@@ -1 +1 @@
- BOOT_IMAGE=/boot/vmlinuz splash=silent mitigations=auto quiet nosimplefb=1
\ No newline at end of file
+ splash=silent mitigations=auto quiet nosimplefb=1
```
jreidinger, line 1: well, to be sure I would enhance test with various parameters we are using like some rd.kiwi params and some oh others which are more tricky to be sure that regexp matching

lslezak:
BTW using spaces in parameters does not work properly. e.g. `initrd="/boot/   /initrd"` is not processed correctly.

We use a simple shell word splitting functionality, for proper parsing we would need to write our full parser and process the input character by character. That's not trivial and then we would probably need to fix all code which uses the processed `/run/agama/cmdline.d/kernel.conf` file.

Using spaces or newlines in file names usually causes troubles. I'd keep it as it is and fix it only when we get a real use case for this scenario.

ancorgs (approved):
LGTM

PR #2295, opened 2025-04-25
feat(live): Configure Firefox devtools to open in fullscreen

lslezak:
## Problem

- The Firefox devtools are displayed at the bottom of the page
- If we need to take a screenshot of the browser console in openQA then some important errors might be hidden (need scrolling to see them)

The original state after pressing F12:
agama-browser-devtools

## Solution

- Open devtools in a separate window
- By default display the console tab
- Display timestamps in the console, that can help to match the browser log with the backend log

The new state after pressing F12:
agama-browser-devtools-fullscreen

### DIFF
--- a/live/live-root/root/.mozilla/firefox/profile/user.js.template
+++ b/live/live-root/root/.mozilla/firefox/profile/user.js.template
@@ -20,6 +20,14 @@ user_pref("browser.startup.homepage_override.mstone", "ignore");
 // TODO: make the XDG portal working again, it has more features (mounting USB flash)
 user_pref("widget.use-xdg-desktop-portal.file-picker", 0);
 
+// configure devtools
+// show devtools in a separate window
+user_pref("devtools.toolbox.host", "window");
+// show the console tab instead of the page inspector after pressing F12
+user_pref("devtools.toolbox.selectedTool", "console");
+// show time stamps in the console
+user_pref("devtools.webconsole.timestampMessages", true);
+
 // start always in the custom homepage
 user_pref("browser.startup.page", 1);
 // custom homepage: the value is expected to be replaced with the login URL by the startup script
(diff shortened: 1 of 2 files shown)


dgdavid:
> So, despite this is a nice job for easing testing/debugging process I still beating for an endpoint in the server for recording the console messages, when possible.

But don't get me wrong. Nice work, thanks!

lslezak:
The problem with console messages is that you cannot reliably catch all messages, you can catch `console.log()` calls from JavaScript, but you cannot catch the errors reported by the browser itself, like that "Too many calls to Location or History APIs within a short timeframe" error. :worried: 

So taking the console screenshot will be still important feature.

live/live-root/root/.mozilla/firefox/profile/user.js.template
```diff
@@ -20,6 +20,14 @@ user_pref("browser.startup.homepage_override.mstone", "ignore");
 // TODO: make the XDG portal working again, it has more features (mounting USB flash)
 user_pref("widget.use-xdg-desktop-portal.file-picker", 0);
 
+// configure devtools
+// show devtools in a separate window
+user_pref("devtools.toolbox.host", "window");
+// show the console tab instead of the page inspector after pressing F12
+user_pref("devtools.toolbox.selectedTool", "console");
+// show time stamps in the console
+user_pref("devtools.webconsole.timestampMessages", true);
```
dgdavid, line 29: For saving more space, I guess you could even show only errors by disabling others

```
devtools.webconsole.filter.info	        false	
devtools.webconsole.filter.warn       false   
```

lslezak: We use logging at very few places in the web UI, most of them are errors or warnings already. We can change the filtering later when really needed., for now I'd keep it as it is.

imobachgs (approved):

PR #2299, opened 2025-04-28
Ensure the NetworkManager persistent connections are copied unless disabled explicitly

teclator:
## Problem

If there is no network configuration specified through the kernel cmdline then the /run/agama/copy_network is not created and that is wrong as the network configuration written persistently during installation will not be copied to the target system.

- related to #2291 

## Solution

Modify the check discarding the copy of the network files if disabled explicitly (it will check if the **/run/agama/not_copy_network** file exists).

## Testing

- *Tested manually*

### DIFF
--- a/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
+++ b/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
@@ -9,7 +9,6 @@ if [ -e /etc/hostname ]; then
 fi
 
 if getargbool 1 inst.copy_network; then
-  : >/run/agama/copy_network
   # If there is some explicit network configuration provided through the ip= kernel cmdline option
   # then we set the config server configuration diabling the DHCP auto configuration for ethernet
   # devices and also copying the configuration persistently (bsc#1241224, bsc#122486, bsc#1239777,
@@ -23,4 +22,6 @@ if getargbool 1 inst.copy_network; then
     mkdir -p "$NEWROOT/etc/NetworkManager/system-connections/"
     cp /run/NetworkManager/system-connections/* "$NEWROOT/etc/NetworkManager/system-connections/"
   fi
+else
+  : >/run/agama/not_copy_network
 fi
--- a/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
+++ b/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
@@ -8,12 +8,13 @@ if [ -e /etc/hostname ]; then
   cp /etc/hostname "$NEWROOT/etc/hostname"
 fi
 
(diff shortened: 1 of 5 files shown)


live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
```diff
@@ -22,4 +22,6 @@ if [ -e /run/agama/copy_network ]; then
     mkdir -p "$NEWROOT/etc/NetworkManager/system-connections/"
     cp /run/NetworkManager/system-connections/* "$NEWROOT/etc/NetworkManager/system-connections/"
   fi
+else
+  : >/run/agama/not_copy_network
```
cfconrad, line 26: Your "Solution" mention creating "`/run/agama/copy_network` always", I guess you need to update your text.

cfconrad (approved):
Looks good to me.

live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
teclator: Yes, it was not updated. Done, and thanks for reporting.

live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
```diff
@@ -8,12 +8,12 @@ if [ -e /etc/hostname ]; then
...
 
-# If there is some explicit network configuration provided through the ip= kernel cmdline option
-# then we set the config server configuration diabling the DHCP auto configuration for ethernet
-# devices and also copying the configuration persistently (bsc#1241224, bsc#122486, bsc#1239777,
-# bsc#1236885).
-if [ -e /run/agama/copy_network ]; then
-  if getargbool 1 inst.copy_network; then
+if getargbool 1 inst.copy_network; then
+  # If there is some explicit network configuration provided through the ip= kernel cmdline option
+  # then we set the config server configuration diabling the DHCP auto configuration for ethernet
```
imobachgs, line 13:
```suggestion
  # then we set the config server configuration disabling the DHCP auto configuration for ethernet
```

live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/save-agama-conf.sh
```diff
@@ -8,12 +8,12 @@ if [ -e /etc/hostname ]; then
...
+  # devices and also copying the configuration persistently (bsc#1241224, bsc#122486, bsc#1239777,
```
imobachgs, line 14: np: do you need `persistently` here?

teclator: we could replace it by `to /etc`..

PR #2302, opened 2025-04-28
Dasd unattended

jreidinger:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`


rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -66,6 +76,149 @@ impl<'a> DASDClient<'a> {
...
+        Ok(())
+    }
+
+    async fn config_activate(&self, config: &DASDConfig) -> Result<(), ServiceError> {
+        let pairs = self.config_pairs(config).await?;
+        let to_activate: Vec<&str> = pairs
+            .iter()
+            .filter(|(system, _config)| system.enabled == false)
+            .filter(|(_system, config)| {
+                config.state.clone().unwrap_or_default() == DASDDeviceState::Active
```
imobachgs, line 100: If you derive `Copy` for `DASDDeviceState`, you can omit the `clone`.

rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -66,6 +76,149 @@ impl<'a> DASDClient<'a> {
...
+            })
+            .map(|(system, _config)| system.id.as_str())
+            .collect();
+        self.enable(&to_activate).await?;
+
+        let pairs = self.config_pairs(config).await?;
```
imobachgs, line 106: Any reason to read the devices again? I wonder whether it was possible to use a single call to `config_pairs`.

rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -66,6 +76,149 @@ impl<'a> DASDClient<'a> {
...
+
+    async fn wait_for_format(&self, job_path: String) -> Result<(), ServiceError> {
+        let mut finished = false;
+        let jobs_client = JobsClient::new(
+            self.connection.clone(),
+            "org.opensuse.Agama.Storage1",
+            "/org/opensuse/Agama/Storage1/jobs",
+        )
+        .await?;
+        while !finished {
```
imobachgs, line 156: I guess this one works just fine, but I wonder whether it could be possible to rely on signals (I do not know if jobs emit signals).

rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -66,6 +76,149 @@ impl<'a> DASDClient<'a> {
...
+            .collect();
+        config
+            .devices
+            .iter()
+            .map(|c| {
+                Ok((
+                    devices_map
+                        .get(c.channel.as_str())
+                        .ok_or(ServiceError::DASDChannelNotFound(c.channel.clone()))?
+                        .clone(),
```
imobachgs, line 215: Is it possible that you find the same channel in more than one device? If it is possible, please, ignore the rest of this message :-)

If it is not possible, you can avoid that `clone` by using `remove` instead of `get` (and making the HashMap mutable, of course).

rust/agama-lib/src/storage/http_client.rs
```diff
@@ -49,29 +51,3 @@ impl StorageHTTPClient {
         Ok(self.client.put_void("/storage/config", config).await?)
     }
 }
-
-#[derive(Debug, thiserror::Error)]
```
imobachgs, line 53: :+1:

imobachgs (changes requested):
In general, it looks good. However, unless I am missing something, the OpenAPI does not include the new API.

rust/agama-lib/src/storage/client/dasd.rs
jreidinger: Ah, I planned to have config_pair produce iterator instead of vector, so in that case it will be consumed. But maybe it will be more efficient to just use original vector.

rust/agama-lib/src/storage/client/dasd.rs
```diff
@@ -66,6 +76,149 @@ impl<'a> DASDClient<'a> {
...
 
+    pub async fn get_config(&self) -> Result<DASDConfig, ServiceError> {
+        // TODO: implement
+        Ok(DASDConfig::default())
+    }
+
+    pub async fn set_config(&self, config: DASDConfig) -> Result<(), ServiceError> {
+        // at first probe to ensure we work on real system info
+        self.probe().await?;
+        self.config_activate(&config).await?;
```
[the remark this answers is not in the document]
jreidinger: well, I just want logically split steps. It needs whole configuration as it is list of configuration for devices and each step apply part of configuration. But splitting config to that parts will be non-trivial code.

PR #2305, opened 2025-04-29
feat: add support for relative URLs

imobachgs:
## Problem

The `relurl` schema is used in several places, especially in AutoYaST profiles. It can be used to refer a script, a file, repositories, etc. They are useful because you do not need to specify the full URL, but it uses a URL which is relative to the profile URL.

The `relurl` is not directly supported by the FileFromUrl API, but it is implemented in different places. The goal of this PR is adding support to read files and scripts from relative locations.

`relurl` is an AutoYaST-specific invention.

## Solution

Instead of porting `relurl`, introduce the concept of **URL reference**, well known from HTML `<a href=...>` and standardized in  [RFC3986][] (Uniform Resource Identifier (URI): Generic Syntax).
A URL reference is either a URL (absolute), or a relative reference (typically just a path), resolved in relation to a base URL.

[RFC3986]: https://datatracker.ietf.org/doc/html/rfc3986.html

```jsonnet
{
  files: [
    {
      destination: '/etc/issue.d/agama.issue',
      url: 'issue-sles',
    },
    {
      destination: '/etc/issue.d/readme.issue',
      url: 'http://192.168.122.1/agama/issue-readme',
    },
  ],
  scripts: {
    post: [
      {
        name: "disable-nginx",
        url: "disable-nginx.sh",
      },
      {
        name: "disable-firewall",
        url: "http://192.168.122.1/agama/disable-firewall.sh",
      },
    ],
  }
}
```

### A new Url

The [Url struct](https://docs.rs/url/latest/url/struct.Url.html) from the [url crate](https://crates.io/crates/url) does not support relative URLs. This PR uses [fluent_uri](https://github.com/yescallop/fluent-uri-rs) but the old `Url` has not been replaced everywhere yet.

### What is the base URL?

The first problem when using relative URLs is to determine which is the base URL. In this case, it is the profile's URL. We could store the URL somewhere (in the manager?) but the true is that it can change with every call to `agama profile import`. And what about `agama config load`?

For this first implementation, I have decided to inject the URL depending on the context:

* `agama profile import`: it uses the URL of the profile.
* `agama config load`: it injects the current directory. It will not work as expected when working remotely, but it should do when calling `agama config load` locally. Another option might be to just report an error about not being able to resolve the URL.

(Upcoming changes: according to #2289, `agama config generate BASE_URL` (the replacement for the most complex part of `agama profile import`) will be responsible for resolving any relative references, according to the explicit base given to it, and replace them with absolute URL in its  JSON output. `agama config load` will not need to care.)

### Other changes

* Fix handling of "file_owner" when importing a files definition.
* Improve test coverage of the Ruby part.

## To do

- [x] Add support for files
- [x] Add support for scripts
- [x] Reduce the amount of cloning (should `resolve_url` modify the receiver?).
- [x] (optional) Implement YaST conversion (out of scope originally).

## Testing

- Added a new unit test
- Tested manually


mvidner:
Thanks! Commenting on the PR description so far, not the code yet:

My reason for wanting to retire `relurl:` is that I consider it a reinvented wheel, where the original wheel is a well understood concept from HTML+HTTP:

RFC 3986: Uniform Resource Identifier (URI): Generic Syntax
Section 5.2.  Relative Resolution
https://datatracker.ietf.org/doc/html/rfc3986.html#section-5.2

We should tell the users: we just follow the standard (and ideally use a library that resolves relative URIs for us so we don't mess up and end up with a slightly incompatible wheel)

> **A new Url**  The [Url struct](https://docs.rs/url/latest/url/struct.Url.html) from the [url crate](https://crates.io/crates/url) does not support relative URLs. This PR introduces our own Url concept, which is just an enum with a wrapper around the original Url (perhaps we should use a different name).

A search for [rust relative uri resolution](https://duckduckgo.com/?t=ftsa&q=rust+relative+uri+resolution&ia=web) points me to `http` crate https://docs.rs/http/latest/http/uri/struct.Uri.html which claims to support relative ones.

imobachgs:
> A search for [rust relative uri resolution](https://duckduckgo.com/?t=ftsa&q=rust+relative+uri+resolution&ia=web) points me to `http` crate https://docs.rs/http/latest/http/uri/struct.Uri.html which claims to support relative ones.

I checked that crate and it does not support URLs like this: `../something/else`. That kind of URLs are rather common and I would like to get them supported.

mvidner:
Another top hit from that search is <https://docs.rs/fluent-uri/latest/fluent_uri/> which seems to target our use case.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -65,20 +70,60 @@ impl BaseScript {
...
+    }
 }
 
+#[serde_as]
 #[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema)]
 #[serde(untagged)]
 pub enum ScriptSource {
     /// Script's content.
     Text { content: String },
     /// URL to get the script from.
```
mvidner, line 95:
```suggestion
    /// URL to get the script from (or a URL reference, relative to FIXME).
```

rust/agama-lib/src/file_source.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// * `file`: where to write the data.
+    pub fn write<P: AsRef<Path>>(&self, path: P, mode: u32) -> Result<(), FileSourceError> {
+        let mut file = OpenOptions::new()
+            .mode(mode)
+            .write(true)
+            .create(true)
+            .open(path)?;
+
+        match &self {
+            FileSource::Text { content } => write!(file, "{}", &content)?,
```
mvidner, line 85: NP: (and untested)
formatted write seems inappropriate here

```suggestion
            FileSource::Text { content } => file.write_all(content.as_bytes())?,
```

rust/agama-lib/src/file_source.rs
```diff
@@ -0,0 +1,186 @@
...
+            FileSource::Remote { url } => Transfer::get(&url.to_string(), &mut file)?,
```
mvidner, line 86: This looks OK but then I realize that `url` can be a relative `UriRef` in which case the user will get an error.
IMHO you forgot this case

rust/agama-lib/src/files/store.rs
```diff
@@ -55,7 +55,25 @@ impl FilesStore {
...
+                .filter_map(|file| {
+                    file.resolve_url(&base)
+                        .inspect_err(|e| {
+                            log::warn!("Error processing file {}: {e}", file.destination)
+                        })
+                        .ok()
+                })
+                .collect()
+        } else {
+            files
```
mvidner, line 75: This will silently leave relative refs unresolved if we happen to have no context. A comment why that is OK please?

rust/agama-lib/src/files/store.rs
```diff
@@ -55,7 +55,25 @@ impl FilesStore {
...
 
     /// stores the list of user files via http API
-    pub async fn store(&self, files: &Vec<UserFile>) -> FilesStoreResult<()> {
-        Ok(self.files_client.set_files(files).await?)
+    pub async fn store(
+        &self,
+        files: &Vec<UserFile>,
+        context: &StoreContext,
+    ) -> FilesStoreResult<()> {
+        let resolved_files = if let Some(base) = &context.source {
```
mvidner, line 63: This should be a separate `fn`

rust/agama-lib/src/files/store.rs
```diff
@@ -55,7 +55,25 @@ impl FilesStore {
...
+            &files
+                .iter()
+                .filter_map(|file| {
+                    file.resolve_url(&base)
+                        .inspect_err(|e| {
+                            log::warn!("Error processing file {}: {e}", file.destination)
```
mvidner, line 69: I don't understand why we only log the error (and omit the erroneous Files from the store). Shouldn't the user see the failure?

rust/agama-lib/src/scripts/model.rs
```diff
@@ -48,39 +45,21 @@ pub enum ScriptsGroup {
...
-            .truncate(true)
-            .mode(0o500)
-            .open(&script_path)?;
-
-        match &self.source {
-            ScriptSource::Text { content } => write!(file, "{}", &content)?,
-            ScriptSource::Remote { url } => Transfer::get(url, &mut file)?,
-        };
-
+        self.source.write(&script_path, 0o500)?;
```
mvidner, line 58: Not writable?
For the `auth login` file we also omitted the write bit and it did bite us.

rust/agama-lib/src/scripts/model.rs
```diff
@@ -454,4 +459,53 @@ mod test {
         _ = repo.clear();
         assert!(!script_path.exists());
     }
+
+    #[test]
+    async fn test_resolve_url_relative() {
+        let base_url = Uri::parse("http://example.lan/sles").unwrap().to_owned();
```
mvidner, line 465:
```suggestion
        let base_url = Uri::parse("http://example.lan/sles.json").unwrap().to_owned();
```

to make it clear that it is not a directory

rust/agama-lib/src/scripts/store.rs
```diff
@@ -98,6 +111,31 @@ impl ScriptsStore {
...
+    /// If it uses a relative URL, it will be resolved against the base URL if given.
+    ///
+    /// * `script`: script definition.
+    /// * `base_url`: base URL to resolve the script URL against.
+    async fn add_script(
+        &self,
+        script: Script,
+        base_url: &Option<Uri<String>>,
+    ) -> ScriptStoreResult<()> {
+        let resolved = match base_url {
```
mvidner, line 125: I have the same (several) comments as for `FilesStore::store`

mvidner (changes requested):
...

rust/agama-lib/src/file_source.rs
imobachgs: I did not forgot, but I simply let `Transfer` to handle the error. In the future I would like to replace `url::Url` with `fluent_uri::UriRef` and have better handling here.

mvidner: Please add a comment about it.

rust/agama-lib/src/file_source.rs
```diff
@@ -0,0 +1,187 @@
...
+                content: content.clone(),
+            },
+            Self::Remote { url } => {
+                let resolved_url = if url.has_scheme() {
+                    url.clone()
+                } else {
+                    let resolved = url
+                        .resolve_against(base)
+                        .map_err(|e| FileSourceError::ResolveUrlError(url.to_string(), e))?;
+                    UriRef::parse(resolved.to_string()).unwrap() // unwrap OK: resolved is already a Uri<String>
```
mvidner, line 67: Aaah, then I have a shorter way :)
```suggestion
                    UriRef::from(resolved)
```

imobachgs: > Aaah, then I have a shorter way :)

Oh, I must be blind :-) I did not see that function. Thanks!

mvidner (changes requested):
The code is good, but seriously we are missing documentation.

Looking at the "Solution" part of the PR description, it seems obvious to me that the minimal docs change should be the JSON schema. Please do :)

imobachgs:
As mentioned in IRC, I decided to change the implementation a bit. The logic to resolve the URLs is now invoked from `InstallSettings` itself (in a single place). BTW I updated the documentation in the JSON schema. I hope it helps.

rust/agama-cli/src/lib.rs
```diff
@@ -185,15 +187,24 @@ async fn build_manager<'a>() -> anyhow::Result<ManagerClient<'a>> {
...
         .open(path)
         .context(format!("Cannot write the file '{}'", path.display()))?;
 
-    match Transfer::get(url, &mut file) {
+    let context = InstallationContext::from_env().unwrap();
+    let uri = UriRef::parse(url).context("Invalid URL")?;
+    let absolute_url = if uri.has_scheme() {
+        uri.to_string()
+    } else {
+        uri.resolve_against(&context.source)?.to_string()
```
mvidner, line 204: Documentation! Again we are adding a secret feature. Please locate the 2 occurrences of `relurl` in Agama source, both are affected by this PR, one by this commit :)

mvidner (changes requested):
The help for `agama download` needs adjusting

mvidner (approved):
LGTM now, thank you!

PR #2314, opened 2025-04-30
Correctly handle the "zh-Hans" language (bsc#1238584)

lslezak:
## Problem

- The storage page crashes after selecting Chinese language ("zh-Hans")
- https://bugzilla.suse.com/show_bug.cgi?id=1238584
- A similar problem exists also with the `pt_BR` language as it also uses a country code. (All other languages use just a simple language code without any separator.) But it is less serious as the `pt_BR` locale is known by glibc so we do not need to do any additional conversions besides using the correct language separator.
- The underlying problem is that Linux and the Web (browsers) use two different standards for identifying locales and languages: POSIX locale identifiers such as `zh_TW.UTF-8` and IETF language tags such as `zh-Hant`. Agama sometimes uses a punctuation-only conversion which only works for some cases.

## Solution

- Change the `zh_Hans` in Weblate to `zh_CN` so we do not need to do any conversions
- Change the web frontend to consistently use the dash (`-`) separator in all places, avoid unnecessary conversions, convert to underscore only when sending to the backend
- Convert the underscore (`_`) separator to dash (`-`) when importing the frontend translations from Weblate
- Handle exceptions in the `Intl.ListFormat` formatting function, in case of exception fallback to a simple formatting function with comma separator (", ")

## Notes

- It turned out that actually only the `C.UTF-8` and `en_US.UTF-8` locales were present in the Live ISO. Unfortunately the filtering regexp was not updated after renaming the PO translation files from `po.<lang>.js` to `po-<lang>.js`. :scream: 
- I have changed the code to read the supported languages from the `languages.json` file. That allows more precise filtering. In the past we kept e.g. all `de_DE`, `de_AT`, `de_CH` for German language. Now we know that after selecting German the `de_DE` locale will be used so we can delete the `de_AT` and `de_CH` locales from the live ISO.

## Testing

- Added a new unit test
- Tested manually
  -  The web frontend does not crash a properly displays the formatted text, the `Intl.ListFormat` function is used in the `New partitions will be created for "/" <and> "swap"` text
  - The backend receives a valid Linux locale (`zh_CN.UTF-8`)
  - The Web UI translation files use the dash separator
  - The Live ISO contains all supported locales (`locale -a`)

agama-zh-CN

image


web/src/context/installerL10n.tsx
```diff
@@ -127,7 +127,13 @@ function languageFromLocale(locale: string): string {
  * @see https://www.rfc-editor.org/info/bcp78
  */
 function languageToLocale(language: string): string {
-  const [lang, country] = language.split("-");
+  let [lang, country] = language.split("-");
+
+  // glibc does not know the "zh_HANS.UTF-8" locale, change it to "zh_CN.UTF-8"
+  if (country === "Hans") {
+    country = "CN";
+  }
```
mvidner, line 135: (syntax?)
```suggestion
  } else if (country === "Hant") {
    country = "TW";
  }
```

web/src/context/installerL10n.tsx
```diff
@@ -127,7 +127,13 @@ function languageFromLocale(locale: string): string {
  * @see https://www.rfc-editor.org/info/bcp78
  */
 function languageToLocale(language: string): string {
```
mvidner, line 129: please note there is also an inverse `languageFromLocale` function above, which also needs fixing, and testing

web/src/agama.ts
```diff
@@ -104,8 +104,15 @@ const agama = {
    * @param options passed to the Intl.ListFormat constructor
    */
   formatList: (list: string[], options: object): string => {
-    const formatter = new Intl.ListFormat(agama.language, options);
-    return formatter.format(list);
+    try {
+      // convert the locale string ("_" separator) to HTTP language code ("-" separator)
```
mvidner, line 108: Sorry, WHAT? Are we adding the same bug here that we are fixing in other parts of this PR??

web/src/agama.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe("agama", () => {
+  describe("formatList", () => {
+    afterEach(() => {
+      // restore the default language
+      agama.language = "en";
+    });
+
+    it("it accepts a locale with underscore", () => {
+      agama.language = "zh_Hans";
```
mvidner, line 33: This is nonsense. `zh_Hans` with an underscore is neither a POSIX locale nor an IETF language tag

Not sure what the authoritative reference on posix locales is, let's take https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02

mvidner: Grr, apparently `zh_Hans` comes from Weblate, and we also use that form in the products.d/*.yaml files. I still think it is incorrect.
Note that agama and opensuse-welcome are the only ones using zh_Hans, other projects use zh_CN: https://l10n.opensuse.org/languages/

live/src/config.sh
```diff
@@ -200,7 +200,7 @@ ls -1 -d /usr/lib/locale/*.utf8 | sed -e "s#/usr/lib/locale/##" -e "s#utf8#UTF-8
 
 # delete translations and unsupported languages (makes ISO about 22MiB smaller)
 # build list of ignore options for "ls" with supported languages like "-I cs* -I de* -I es* ..."
-readarray -t IGNORE_OPTS < <(ls /usr/share/agama/web_ui/po.*.js.gz | sed -e "s#/usr/share/agama/web_ui/po\.\(.*\)\.js\.gz#-I\n\\1*#")
+readarray -t IGNORE_OPTS < <(ls /usr/share/agama/web_ui/po-*.js.gz | sed -e "s#/usr/share/agama/web_ui/po-\(.*\)\.js\.gz#-I\n\\1*#" -e "s/zh_Hans/zh_CN/")
```
mvidner, line 203: adding Taiwan (for democracy), and Newlineland (for readability)
```suggestion
readarray -t IGNORE_OPTS < <( \
    ls /usr/share/agama/web_ui/po-*.js.gz \
     | sed \
       -e "s#/usr/share/agama/web_ui/po-\(.*\)\.js\.gz#-I\n\\1*#" \
       -e "s/zh_Hans/zh_CN/" \
       -e "s/zh_Hant/zh_TW/" \
  )
```

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr 30 11:45:08 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Use the correct locale format in the Intl.ListFormat.format call
+  (bsc#1238584)
+- Set the "zh_CN" locale in backend for the "zh-Hans" language
```
mvidner, line 6:
```suggestion
- Set the zh_CN, zh_TW locales in backend for the zh-Hans, zh-Hant language tags
```

mvidner (changes requested):
1. Some more cleanup in the code you have already touched
2. Unfortunately Weblate is giving us a hybrid of the two language standards, IMHO we should fight back or be forever stuck with a mess

web/src/agama.test.ts
dgdavid: A though from my side: if above comments are right, it's a pity to have hacks in the code instead of aiming for an _upstream_ fix.

imobachgs:
May I suggest to prevent crashes by catching exceptions in the [formatList](https://github.com/agama-project/agama/blob/932f83f64eaec7e2a6ad98108c522185f605d179/web/src/agama.ts#L106) function?

web/src/agama.ts
```diff
@@ -104,8 +104,14 @@ const agama = {
...
    */
   formatList: (list: string[], options: object): string => {
-    const formatter = new Intl.ListFormat(agama.language, options);
-    return formatter.format(list);
+    try {
+      const formatter = new Intl.ListFormat(agama.language, options);
+      return formatter.format(list);
+    } catch (e) {
+      // use a trivial formatting with commas when the locale formatting fails for whatever reason
+      console.warn("Using fallback function for formatting localized array:", e);
```
dgdavid, line 112: NP: would be useful to log the language in use too.

web/src/agama.ts
```diff
@@ -104,8 +104,17 @@ const agama = {
...
+    try {
+      const formatter = new Intl.ListFormat(agama.language, options);
+      return formatter.format(list);
+    } catch (e) {
+      // use a trivial formatting with commas when the locale formatting fails for whatever reason
+      console.warn(
+        `Using fallback list formatting function for language "${agama.language}", details:`,
+        e,
+      );
+      return list.join(", ");
```
imobachgs, line 116: Why not use the `Intl.ListFormat` for `en-us`? The result would be most probably the same, but it still uses the same function.

imobachgs (commented):
In general it looks great. Just a minor comment about the `formatList` fallback.

web/src/agama.ts
imobachgs: OK, fair enough, although I do not think that's a problem at all. It is just a fallback for corner cases.

imobachgs (approved):

web/src/agama.ts
dgdavid: But it will be add ", and" at the end. Not sure if it is what we are looking for.

lslezak: OK, but it might fail for another reason, like if you by mistake pass integers instead of strings. Then it would fail again, that `.join(", ")` handles that fine:

image

Anyway this is a corner case and hopefully we won't get any exceptions now after dropping "zh_Hans" and using the dash separator everywhere. This is just a safety belt preventing the worst (a crash).

imobachgs: If you pass integers, it will fail in any language :-)

imobachgs: OK, I see. Well, let's keep the join.

dgdavid: As commented in IRC, please note that Agama `formatList` only accepts a list of strings :)

live/src/config.sh
```diff
@@ -199,20 +199,25 @@ mkdir -p /etc/agama.d
 ls -1 -d /usr/lib/locale/*.utf8 | sed -e "s#/usr/lib/locale/##" -e "s#utf8#UTF-8#" >/etc/agama.d/locales
 
 # delete translations and unsupported languages (makes ISO about 22MiB smaller)
-# build list of ignore options for "ls" with supported languages like "-I cs* -I de* -I es* ..."
-readarray -t IGNORE_OPTS < <(ls /usr/share/agama/web_ui/po.*.js.gz | sed -e "s#/usr/share/agama/web_ui/po\.\(.*\)\.js\.gz#-I\n\\1*#")
+# build list of ignore options for "ls" with supported languages like "-I cs -I cs_CZ ..."
```
mvidner, line 202: NP: Sorry, `jq` is over the "magic" threshold for me, I need explanatory notes for the implementation:

```suggestion
# build list of ignore options for "ls" with supported languages like "-I cs -I cs_CZ ..."
# languages.json is like: { "ca-ES": "Català", "de-DE": "Deutsch", ...}
# jq prints ca-ES\nde-DE\n...
```

web/package/agama-web-ui.spec
```diff
@@ -45,9 +45,9 @@ local-npm-registry %{_sourcedir} install --with=dev --legacy-peer-deps || ( find
 NODE_ENV="production" npm run build
 
 %install
-install -D -m 0644 --target-directory=%{buildroot}%{_datadir}/agama/web_ui %{_builddir}/agama/dist/*.{gz,html,js,map,svg}
+install -D -m 0644 --target-directory=%{buildroot}%{_datadir}/agama/web_ui %{_builddir}/agama/dist/*.{css,gz,html,js,json,map,svg}
 install -D -m 0644 --target-directory=%{buildroot}%{_datadir}/agama/web_ui/fonts %{_builddir}/agama/dist/fonts/*.ttf
-install -D -m 0644 --target-directory=%{buildroot}%{_datadir}/agama/web_ui/assets/logos %{_builddir}/agama/src/assets/products/*.svg
+install -D -m 0644 --target-directory=%{buildroot}%{_datadir}/agama/web_ui/assets/logos %{_builddir}/agama/dist/assets/logos/*.svg
```
mvidner, line 50: NP: in the future we might want to go with a `web/install.sh` approach like the other parts of Agama do (part of #2103). I forgot to do it because I was not doing any changes in the front end at the time

mvidner (approved):
LGTM, thanks!

Consider applying a `jq` explanation

web/package/agama-web-ui.spec
lslezak: Yes, would be nice to unify this to have a single place for changes.

PR #2317, opened 2025-05-02
chore(rust): update dependencies

imobachgs:
Update a bunch of Rust dependencies. `axum` and `axum-extra` has not been updated because they might need more work to adapt to latest API changes.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri May  2 07:42:21 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Update dependencies (gh#agama-project/agama#2317).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri May  2 07:42:21 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Update dependencies (gh#agama-project/agama#2317).
```
dgdavid, line 4: I wonder why to add this in the changelog. I'm not against, but not sure how helpful it is from a change log perspective if it does not introduce a breaking change or so.

dgdavid (approved):

rust/package/agama.changes
imobachgs: Just in case something goes wrong later, I would expect this to make it easier to track (especially for somebody who has not been involved).

PR #2319, opened 2025-05-03
Services-manager conversion to scripts

mchf:
## Problem

Missing support for `AY profile ```<services-manager>``` section .

## Solution

Another piece to AutoYaST profile conversion. Takes ```<services-manager>``` section and turns it into agama's post install script(s)

## Testing

- TBD: *Tested manually*

- Run: `git ls-files '*.md'`

### DIFF
--- a/service/lib/agama/autoyast/converter.rb
+++ b/service/lib/agama/autoyast/converter.rb
@@ -27,7 +27,6 @@ require "agama/autoyast/product_reader"
 require "agama/autoyast/root_reader"
 require "agama/autoyast/scripts_reader"
 require "agama/autoyast/security_reader"
-require "agama/autoyast/services_manager_reader"
 require "agama/autoyast/software_reader"
 require "agama/autoyast/storage_reader"
 require "agama/autoyast/user_reader"
@@ -60,7 +59,6 @@ module Agama
         "root",
         "scripts",
         "security",
-        "services-manager",
         "software",
         "storage",
         "user"
@@ -73,17 +71,10 @@ module Agama
       # @return [Hash] Agama profile
       def to_agama(profile)
         SECTIONS.reduce({}) do |result, section|
-          name = section.sub("-", "_")
-          require "agama/autoyast/#{name}_reader"
-          klass = "#{name}_reader".split("_").map(&:capitalize).join
+          require "agama/autoyast/#{section}_reader"
+          klass = "#{section}_reader".split("_").map(&:capitalize).join
           reader = Agama::AutoYaST.const_get(klass).new(profile)
-          result.merge(reader.read) do |key, value1, value2|
-            # scripts and services-manager readers both produce a hash with
-            # the "scripts" key. This should currently be the only one such case.
-            raise IOError, "AutoYaST conversion failed" if key != "scripts"
-
-            value1.merge(value2) { |_, sl1, sl2| sl1 + sl2 }
-          end
+          result.merge(reader.read)
         end
       end
     end
--- a/service/lib/agama/autoyast/converter.rb
+++ b/service/lib/agama/autoyast/converter.rb
@@ -27,6 +27,7 @@ require "agama/autoyast/product_reader"
 require "agama/autoyast/root_reader"
 require "agama/autoyast/scripts_reader"
 require "agama/autoyast/security_reader"
+require "agama/autoyast/services-manager_reader"
 require "agama/autoyast/software_reader"
 require "agama/autoyast/storage_reader"
 require "agama/autoyast/user_reader"
@@ -59,6 +60,7 @@ module Agama
         "root",
         "scripts",
         "security",
+        "services-manager",
         "software",
         "storage",
         "user"
@@ -72,7 +74,7 @@ module Agama
       def to_agama(profile)
         SECTIONS.reduce({}) do |result, section|
           require "agama/autoyast/#{section}_reader"
-          klass = "#{section}_reader".split("_").map(&:capitalize).join
+          klass = "#{section}_reader".sub("-", "_").split("_").map(&:capitalize).join
           reader = Agama::AutoYaST.const_get(klass).new(profile)
           result.merge(reader.read)
         end
--- a/service/lib/agama/autoyast/converter.rb
+++ b/service/lib/agama/autoyast/converter.rb
@@ -27,7 +27,7 @@ require "agama/autoyast/product_reader"
 require "agama/autoyast/root_reader"
 require "agama/autoyast/scripts_reader"
 require "agama/autoyast/security_reader"
-require "agama/autoyast/services-manager_reader"
+require "agama/autoyast/services_manager_reader"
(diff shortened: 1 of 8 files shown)


service/lib/agama/autoyast/converter.rb
```diff
@@ -59,6 +60,7 @@ class Converter
         "root",
         "scripts",
         "security",
+        "services-manager",
```
imobachgs, line 63: Sorry, but you are not supposed to add the `services-manager`to this list. These are the Agama sections, not the AutoYaST ones. You need to extend the `ScriptsReader` instead.

service/lib/agama/autoyast/converter.rb
```diff
@@ -71,10 +73,17 @@ class Converter
...
         SECTIONS.reduce({}) do |result, section|
-          require "agama/autoyast/#{section}_reader"
-          klass = "#{section}_reader".split("_").map(&:capitalize).join
+          name = section.sub("-", "_")
+          require "agama/autoyast/#{name}_reader"
+          klass = "#{name}_reader".split("_").map(&:capitalize).join
           reader = Agama::AutoYaST.const_get(klass).new(profile)
-          result.merge(reader.read)
+          result.merge(reader.read) do |key, value1, value2|
+            # scripts and services-manager readers both produce a hash with
```
imobachgs, line 81: Please, use the `ServicesManagerReader` from the `ScriptsReader` instead.

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2024-2025] SUSE LLC
```
imobachgs, line 3:
```suggestion
# Copyright (c) [2025] SUSE LLC
```

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,133 @@
...
+      def read
+        return {} if services_section.empty?
+
+        # 1) create "post" => [... list of hashes defining a scipt ...]"
+        # 2) each script has to contain name ("randomized/indexed" one or e.g. based on service
+        #    name) and chroot option
+        # 3) script body is one or two lines per service, particular command depends on AY's service
+        #    type
+        {
+          "scripts" => {
```
imobachgs, line 52: If you use it from the `ScriptsReader`, there is no need to return a full `scripts` section.

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      def script(body)
+        return nil if body.nil? || body.empty?
+
+        @index += 1
+
+        {
+          "name"   => "agama-services-manager-#{@index}",
+          "chroot" => true,
+          "body"   => "|||
```
imobachgs, line 98: Sorry, but `|||` is Jsonnet code. What is the result of this conversion?

imobachgs (changes requested):
You need to adjust the approach a bit. Please, check the comments. Thanks!

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,133 @@
...
+      def enabled_to_cmd
+        enabled_services.map do |service|
+          "systemctl enable #{service}"
+        end.join("\n")
+      end
+
+      def ondemand_to_cmd
+        ondemand_services.map do |service|
+          "systemctl enable #{service}.socket"
+        end.join("\n")
```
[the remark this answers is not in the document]
mchf: I thought that json doesn't work with:
```
body: "first line
second line"
```

but

```body: "first line\nsecondline"``` should be fine

but you're right i need to double check whether i really produce the second case

File: service/lib/agama/autoyast/scripts_reader.rb [hunk 7]
```diff
@@ -85,6 +86,8 @@ def read_post_scripts
         scripts = scripts_section.fetch("chroot-scripts", []).map do |script|
           read_post_script(script)
         end
+        scripts += ServicesManagerReader.new(profile).read
```
imobachgs, line 89: If you want to change the array in place, you might want to use `concat`. But I do not have a strong opinion.

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,127 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+Yast.import "URL"
+
+# :nodoc:
+module Agama
+  module AutoYaST
+    # Converts AutoYaST's <services-manager> section into post-install script(s). See the
+    # ScriptsReader for a details.
```
imobachgs, line 29:
```suggestion
    # ScriptsReader for further details.
```

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -0,0 +1,127 @@
...
+    #
+    # Conversion is not done in one-to-one manner. Several services can be joined into one
```
imobachgs, line 31: It is not clear: "can be joined" or "are joined"?

imobachgs (commented):
Please, update https://github.com/agama-project/agama/blob/master/service/share/autoyast-compat.json#L288-L289.

Generating four different scripts looks too much to me. Why not combine all of them into a single script? You would not need to worry about indexes and that stuff.

mchf:
> 
> Generating four different scripts looks too much to me. Why not combine all of them into a single script? You would not need to worry about indexes and that stuff.

The main idea was that in theory we might decide to generate a slightly different scripts for various services types in the future.

imobachgs:
> The main idea was that in theory we might decide to generate a slightly different scripts for various services types in the future. So, just "logical" separation.

I think it adds some complexity to the implementation and, more important, to the resulting profile with no benefits. We can split the script in the future if needed.

service/lib/agama/autoyast/services_manager_reader.rb
```diff
@@ -86,10 +80,8 @@ def ondemand_services
       def script(body)
         return nil if body.nil? || body.empty?
 
-        @index += 1
-
         {
-          "name"    => "agama-services-manager-#{@index}",
+          "name"    => "agama-services-manager-scripts",
```
imobachgs, line 84:
```suggestion
          "name"    => "agama-services-manager",
```

imobachgs (approved):
LGTM. Thanks!

service/share/autoyast-compat.json
```diff
@@ -247,8 +247,43 @@
   },
   {
     "key": "services-manager",
-    "support": "planned",
-    "notes": "Meanwhile you can use post-installation scripts to handle these cases."
+    "support": "yes",
```
imobachgs, line 250: Please, add a note explaining that the section is automatically converted to a post-script.

imobachgs (approved):

PR #2323, opened 2025-05-04
feat(web): improve storage/ProposalFailedInfo

dgdavid:
## Problem

When a proposal fails because not able to allocate space for an logical volume, the warning alert shown to users does not provide enough information because it is only aware of such a situation for partitions.

## Solution

To make such an alert aware of logical volumes but avoiding to entering in details and just mentioning affected mount paths.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

|Before|After|
|-|-|
| localhost_8080_ (51) | localhost_8080_ (50) |

---

Related to https://trello.com/c/vE4qvtDl (private link)

### DIFF
--- a/web/src/components/storage/ProposalFailedInfo.tsx
+++ b/web/src/components/storage/ProposalFailedInfo.tsx
@@ -26,92 +26,8 @@ import { IssueSeverity } from "~/types/issues";
 import { useApiModel } from "~/hooks/storage/api-model";
 import { useIssues, useConfigErrors } from "~/queries/issues";
 import * as partitionUtils from "~/components/storage/utils/partition";
-import { _, n_, formatList } from "~/i18n";
+import { _, formatList } from "~/i18n";
 import { sprintf } from "sprintf-js";
-
-const UnallocatablePartitions = ({ partitions, hasBoot }) => {
-  const mountPaths = partitions.map((p) => partitionUtils.pathWithSize(p));
-
-  return hasBoot
-    ? sprintf(
-        // TRANSLATORS: %s is a list of formatted mount points with size like
-        // '"/" (at least 10 GiB), "/var" (20 GiB) and "swap" (2 GiB)' (or a
-        // single mount point in the singular case).
-        n_(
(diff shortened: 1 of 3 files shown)


web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -42,59 +57,43 @@ function Description({ partitions, booting }) {
...
-}
+};
 
 /**
- * Information about a failed storage proposal
+ * Displays information to help users understand why a storage proposal
+ * could not be generated with the current configuration.
  *
+ * Renders nothing if:
+ *   - The proposal could not be generated at all (known by the presense of
```
ancorgs, line 83: typo: presense

web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -42,59 +57,43 @@ function Description({ partitions, booting }) {
...
-
   return (
     <>
-      <Content component="p">{msg1}</Content>
+      <Content component="p">
+        {sprintf(
+          // TRANSLATORS: %s is a list that includes "boot partition" and one or
+          // more formatted mount points with size like: '"/" (at least 10 GiB),
+          // "/var" (20 GiB), and "swap" (2 GiB)'.
+          _("It is not possible to allocate space for %s."),
```
ancorgs, line 67: I'm not against this, but I just wanted to mention that putting something like "boot partition" in the same list that the mountpoints can be a bit challenging for translations (even for the translators to understand how this and the separate string "boot partition" work together).

I'm not sure if something like this would be better for i18n:

"It is not possible to allocate space for the boot partition and for %s."
and
"It is not possible to allocate space for %s."

ancorgs (approved):
Pre-approved (just the typo of "presense" needs to be fixed).

web/src/components/storage/ProposalFailedInfo.tsx
dgdavid: Indeed, your suggestion is better and has a lot of sense. Thanks.

ancorgs (approved):
LGTM

PR #2324, opened 2025-05-05
fix(rust): add iscsi.schema.json to agama-cli package

imobachgs:
## Problem

It is not possible to validate a profile since the `iscsi.schema.json`, introduced recently, is missing from the package.

## Solution

Add the ìscsi.schema.json file to the `agama-cli` package. Now that the validation is performed in the client side, we might consider moving those files to another package. But, by now, let's fix the auto-installation.


teclator (approved):
LGTM

PR #2328, opened 2025-05-06
Fix target file path

teclator:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`

### DIFF
--- a/service/lib/agama/storage/finisher.rb
+++ b/service/lib/agama/storage/finisher.rb
@@ -146,7 +146,8 @@ module Agama
 
         def run
           glob_files.each do |file|
-            target = File.dirname(file).sub(root_dir, dest_dir)
+            relative_path = File.dirname(file).sub(root_dir, "")
+            target = File.join(dest_dir, relative_path)
 
             FileUtils.mkdir_p(target)
             FileUtils.cp(file, target)


service/lib/agama/storage/finisher.rb
```diff
@@ -146,7 +146,8 @@ def run?
 
         def run
           glob_files.each do |file|
-            target = File.dirname(file).sub(root_dir, dest_dir)
```
imobachgs, line 149: Perhaps using `delete_prefix` will communicate better your intention. But as it is already tested, let's keep it as it is.

imobachgs (approved):
It looks good as a quick fix (which is already tested). I wonder why we do not need to update the tests, though.

For the future, I guess that working with `Pathname` objects instead of strings might be a better alternative.

dgdavid (approved):
Would be nice to have a unit test for this, but I've checked the whole file test is missing. Since you have tested it manually, go ahead 👍

dgdavid:
> Would be nice to have a unit test for this, but I've checked the whole file test is missing. Since you have tested it manually, go ahead 👍

Looks like @imobachgs left interesting comments. Please check them before merging.

service/lib/agama/storage/finisher.rb
teclator: Yeah, already thinking on that, I agree, fixed and also added "/" at the end of the rootdir to show the test was failing with previous implementation but will not change the order of the commits to not force a push.

teclator:
> Would be nice to have a unit test for this, but I've checked the whole file test is missing. Since you have tested it manually, go ahead 👍

There is a unit test but unfortunately with the happy path, added end "/" for making it break without the proper fix.

imobachgs (approved):

PR #2336, opened 2025-05-07
fix(rust): validate the product/addons section

imobachgs:
The `product/addons` section is not included in the schema so it is considered as invalid. Hence, a profile like the one below will not validate:

```json
{
  "id": "SLES",
  "registrationCode": "SOME-REG-CODE",
  "addons": [
    {
      "id": "sle-ha",
      "registrationCode": "SOME-REG-CODE"
    }
  ]
}
```

This PR adds this element to the validation. Manually tested.

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -225,6 +225,7 @@
               },
               "version": {
                 "title": "Version of the add-on",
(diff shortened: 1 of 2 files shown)


mvidner:
Thanks for illustrating the need to validate `agama config show` too :smile:

PR #2338, opened 2025-05-07
feat(storage): add search conditions

joseivanlopez:
Extend storage schema with more search conditions:

* Drives and MD RAIDs: by name, by size (equal, greater or less).
* Partitions: by name, by size (equal, greater or less), by partition number.

Requires https://github.com/yast/yast-storage-ng/pull/1410.


service/lib/agama/storage/configs/search.rb
```diff
@@ -32,14 +32,26 @@ def self.new_for_search_all
           new.tap { |c| c.if_not_found = :skip }
         end
 
+        # Search by name.
```
ancorgs, line 35: This design of `Configs::Search` with separate explicit attributes for name, size and partition_number doesn't look too flexible for the future. How will it work in the future to combine conditions? I mean stuff like this (not sure about the syntax yet):

```json
{
  "partitions": [
    {
      "search": {
        "if_not_found": "create",
        "condition": {
          { "or": [
              { "number": 2 },
              { "and": [
                  { "size": { "greater": "1 MiB" } },
                  { "size": { "less": "10 MiB" } }
                ]
              }
            ]
          }
        }
      }
    }
  ] 
}
```

joseivanlopez: I did the minimal changes to `Configs::Search` to support the current needs. It will evolve according to what we want to offer. I think in the fututure we will have somehing like `Configs#Search#condition` and `Configs#SearchCondition`, or something similar.

service/lib/agama/storage/config_solvers/drives_search.rb
```diff
@@ -0,0 +1,78 @@
...
+        # @return [Boolean]
+        def match_condition?(drive_config, drive)
+          match_name?(drive_config, drive) && match_size?(drive_config, drive)
+        end
+
+        # Candidate drives for solving the search of drive configs.
+        #
+        # @return [Array<Y2Storage::Drive, Y2Storage::StrayBlkDevice>]
+        def candidate_drives
+          disk_analyzer = Y2Storage::DiskAnalyzer.new(devicegraph)
```
ancorgs, line 71: One of the main features/goals of DiskAnalyzer is to cache all the intermediate and final results that are expensive (eg. need to lookup into the system, to mount something, etc.).

In other words, disk analyzers are more designed to be shared together with its devicegraph than to be created over and over.

I feel kind of uncomfortable creating instances of `DiskAnalyzer` for every search (or for every kind of search). Have we checked it's not a potential performance problem?

joseivanlopez: I haven't checked the performance. I will pass the disk analyzer to avoid some potential penalty. Anyway, it would be nice to have `DiskAnalyzer#devicegraph` as a public method. So we don't have to pass a devicegraph-analyzer.

ancorgs (approved):
LGTM (except, of course, the lack of a changelog).

ancorgs:
Now that I realize, should we update the dependency on yast2-storage-ng?

joseivanlopez:
> Now that I realize, should we update the dependency on yast2-storage-ng?

Yes, it is updated, see https://github.com/agama-project/agama/pull/2338/files#diff-ceee56f39d11b19309aec78a4c4b31b91f100062d85d7548a3d1cba14972cfa2.

PR #2346, opened 2025-05-08
UI: handle reused MD RAIDs that are already included in the configuration

ancorgs:
## Problem

If a storage configuration includes reusing an existing MD RAID device, such a device is not displayed at the Agama configuration UI. But actually that case is not so different from using a disk.

## Solution

This extends the component `ConfigEditor` to also display reused MD RAIDs, offering the same capabilities that are there for disks (like creating partitions or defining what to do with the existing ones).

md0b

md0a

This does not offer any possibility to remove the RAID from the configuration or to add a new RAID. It only works with RAIDs that are already present at the loaded configuration.

## Implementation notes

Some of the pre-existing components like `DriveEditor`, `SpacePolicySelection` or `PartitionsPage` use the old queries at `config-model.ts` to communicate with the backend. Something that is deprecated in favor of https://gist.github.com/joseivanlopez/a8d1e4269e6f1c8d41837dfd89fb556d?permalink_comment_id=5569616

This pull request gives several steps on the agreed direction, but a few inner components still use the old approach.

## Testing

- Tested manually
- Backend tests reorganized and expanded
- Frontend tests updated


service/lib/agama/storage/config_conversions/from_model_conversions/config.rb
```diff
@@ -53,21 +54,24 @@ def default_config
           # @return [Hash]
           def conversions
             drives = convert_drives
+            raids = convert_raids
+            partitionables = Array(drives) + Array(raids)
 
             {
-              boot:          convert_boot(drives || []),
+              boot:          convert_boot(partitionables),
```
joseivanlopez, line 61: Can we assign a MD as boot device? Right now the solver only considers drives as candidates for boot device.

joseivanlopez: Discussed in a planning.

joseivanlopez:
As discussed in a meeting, we will adapt the storage editor components (e.g., *DriveEditor*) to receive a list and an index instead of a device object. This will allow to:

* Simplify the model types (*list* and *listIndex* properties are not needed).
* All the components works in a similar way (*DriveEditor*, *PartitionPage*, etc). They receive a list and an index and each  component retrieve the device from the model if needed.
* The components always have the needed info for generating an url.
* The partitions and logical volumes can also be referenced by an index, so it opens the door to have partitions and lvs without a  mount path.

These changes will be done once we need to allow creating partitions without mount path (e.g., for manually creating physical volumes).

joseivanlopez (approved):
LGTM

PR #2356, opened 2025-05-13
Send different NM dbus values depending on version

jcronenberg:
## Problem

Agama doesn't detect and handle different NetworkManager versions. I know that currently this isn't an issue for agama itself, however I think it may become one in the future. Also for wicked2nm it represents a problem because when it is run on Leap15, it tries to send values that are not present due to the old NM version. See https://github.com/openSUSE/wicked2nm/issues/24

## Solution

Parse NM version via the semver crate and then act accordingly when populating dbus values. I haven't checked all the values and when they were introduced, but I checked for Leap15 and also I mentioned in #2189 that `dhcp-send-hostname` is deprecated in favor of `dhcp-send-hostname-v2`

## Testing

- *Added a new unit test*
- *Tested manually*
- *Tested with wicked2nm*


imobachgs (commented):
I am not convinced about the approach (relying on version numbers) but perhaps there is no better way. I wonder if we could have a model of which features are supported by the current API and leave the version comparison to the top level.

I will try to include this work in one of our upcoming sprints. Or are we blocking you?

jcronenberg:
@imobachgs Any update from your side wrt my previous comment? I think if I reduce it to just the `dhcp-send-release` changes I really think it is very minimal and should in no way interfere with normal agama operations (especially since these dhcp values aren't used) and if you want to implement a different approach reverting the changes should take next to no effort.

jcronenberg:
Thanks! :slightly_smiling_face:

PR #2358, opened 2025-05-14
Do not set the root option in /boot/grub2/grub.cfg (bsc#1236873)

lslezak:
## Problem

- The root option in `/boot/grub2/grub.cfg` points to a non-existing file, the target name is not constant and is changed
- But the root is already set in the `/boot/grub2/earlyboot.cfg` file, this line is not needed actually
- See https://bugzilla.suse.com/show_bug.cgi?id=1236873

## Solution

- Remove that unnecessary and invalid line

## Testing

- Tested manually, the Live ISO still boots :smiley: 

## Additional fix

- Replace underscores in the boot menu label by spaces
- The separator in the Kiwi profile names was changed from dash to underscore in the past without updating this transformation
- It changes the "Install SUSE_SLE_16 (x86_64)" boot menu label to "Install SUSE SLE 16 (x86_64)" which looks better

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue May 13 16:56:56 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Do not set the root in /boot/grub2/grub.cfg file, it was wrong
+  and the root is already set in the /boot/grub2/earlyboot.cfg
+  file (bsc#1236873)
+- Replace underscores in the boot menu label by spaces
+
+-------------------------------------------------------------------
 Wed May  7 07:07:33 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
 - Fixed locale cleanup to not delete all locales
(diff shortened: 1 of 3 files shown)


dgdavid (approved):

live/src/agama-installer.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue May 13 16:56:56 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Do not set the root in /boot/grub2/grub.cfg file, it was wrong
+  and the root is already set in the /boot/grub2/earlyboot.cfg
+  file (bsc#1236873)
+- Replace underscores in the boot menu label by spaces
```
jreidinger, line 7: would it make sense to replace both? like underscore and also dashes?

lslezak: Dashes actually cannot be used as the Kiwi profile name is then used as a part of the version in OBS and dash is not allowed in RPM version.

jreidinger (approved):

PR #2361, opened 2025-05-14
Coveralls fixes

lslezak:
## Problem

- The Coveralls reporting step in the GitHub Actions got stuck ([example job](https://github.com/agama-project/agama/actions/runs/15013186385/job/42185530706#step:9:170))
- And because the default timeout is 6 hours then the CI was stuck for 6 hours
- The number of the assigned Action runners is limited so after reaching the limit new CI actions were not started and were waiting until some previous job is finished.

## Debugging

- At first I noticed an error in the Coveralls output:
  `⚠️ git log -1 --pretty=format:'%H%n%aN%n%aE%n%cN%n%cE%n%s' (return code 128)`
  see https://github.com/agama-project/agama/actions/runs/15023225072/job/42217441468#step:14:166
- Running this command in a separate step showed that it was failing because of a different file owner:
  `fatal: detected dubious ownership in repository at '/__w/agama/agama'`
  see https://github.com/agama-project/agama/actions/runs/15020310724/job/42207691974#step:9:9
- But even after adding that config command it got still stuck
- After enabling the Coveralls steps in the web frontend again I noticed that it works fine without any problem so I tested moving the Coveralls step outside the container and it worked.
- Very likely there is some incompatibility in the Coveralls action and it does not work well when running inside a container with openSUSE Tumbleweed (Different `git` CLI tool? Different glibc, nodejs or something else?)

## Solution

- Enable back the code coverage reporting in the web frontend
- Configure `git` so a different owner is OK (this was not the root of the problem in the end but this could cause troubles elsewhere later so I'm keeping this change anyway.)
- Run the coverage actions outside container, run them in the Ubuntu host directly
  (That's a bit tricky, if a job runs in a container then all steps run in the container, you cannot run a command outside the container. The workaround is to define a dependent job which runs outside the container and pass the coverage data as an artifact which is uploaded/downloaded.)
- Decrease the timeout, explicitly set 1 hour for CI and 20 minutes for Coveralls to not block for too long if the job gets stuck for whatever reason.
- Limit the permissions as suggested by the GitHub scanner to only read the repository content.

## Testing

- Works fine, see the green marks for this pull request, the Coveralls reports were correctly sent.

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -38,6 +38,8 @@ on:
 
 jobs:
   rust_ci:
+    # the default timeout is 6 hours, that's too much if the job gets stuck
+    timeout-minutes: 60
     runs-on: ubuntu-latest
     env:
       COVERAGE: 1
@@ -59,6 +61,9 @@ jobs:
 
     - name: Git Checkout
       uses: actions/checkout@v4
+
+    - name: Configure git
+      run:  git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
     - name: Configure and refresh repositories
       # disable unused repositories to have faster refresh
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

PR #2362, opened 2025-05-14
fix(storage): correct candidate drives

joseivanlopez:
## Problem

The list of candidate drives for installation includes some specific MD RAIDs that are considered as BIOS RAID by YaST. Moreover, the search of MD RAIDs do not discard the devices with mounted content or with installation repositories. 

## Solution

Fix candidate drives to avoid including any MD RAID.
Fix candidate MD RAIDs to avoid including devices with mounted content or with installation repositories.

### DIFF
--- a/service/lib/agama/storage/proposal_strategies/base.rb
+++ b/service/lib/agama/storage/proposal_strategies/base.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2024] SUSE LLC
+# Copyright (c) [2024-2025] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -71,6 +71,8 @@ module Agama
         end
 
         # Available devices for installation.
+        #
+        # @deprecated Only used by Guided strategy.
         #
         # @return [Array<Y2Storage::Device>]
         def available_devices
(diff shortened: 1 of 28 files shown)


service/lib/agama/storage/system.rb
```diff
@@ -0,0 +1,79 @@
...
+# find current contact information at www.suse.com.
+
+require "y2storage/storage_manager"
+require "y2storage/disk_analyzer"
+
+module Agama
+  module Storage
+    # Helper class for asking about the system devices.
+    class System
+      # @param devicegraph [Y2Storage::Divecegraph, nil] Devicegraph representing the system. Probed
```
ancorgs, line 29: Divecegraph

service/lib/agama/storage/proposal_strategies/base.rb
```diff
@@ -72,6 +72,8 @@ def disk_analyzer
 
         # Available devices for installation.
         #
+        # @deprecated Only used by Guided strategy.
+        #
```
ancorgs, line 76: Why don't we move the method to the Guided class then?

ancorgs: In general. I think it makes sense to change `ProposalStrategies::Base` to work with the new class system, instead of only implementing that change for the Agama strategy.

Or there is anything stopping us from doing so?

joseivanlopez: Yes, I think we can use `Storage::System`.

PR #2364, opened 2025-05-16
Cache the software data to avoid blocked UI during package installation

lslezak:
## Problem

- Agama installation jumps to congratulation page without any middle page shown
- https://bugzilla.suse.com/show_bug.cgi?id=1241208
- This is a race condition, it depends on exact timing. Usually the data is read before the package installation starts, but in some cases not.

## Details

- The problem is that the software service is blocked by the running libzypp installation and does not respond to the requests.
- The `/api/software/config` and `/api/software/products` results are returned after the whole installation is finished
  agama-blocked-services

## Solution

- Store the software data before starting the installation, during installation return the cached data.

## Testing

- Tested manually


rust/agama-server/src/l10n/web.rs
```diff
@@ -161,6 +161,7 @@ async fn set_config(
         state.manager_proxy.set_locale(&locale_string).await?;
         changes.ui_locale = Some(locale_string);
 
+        // TODO: reset/reread the software product cache, it contains translations
```
lslezak, line 164: The code works fine, the only problem is that the product cache needs to be read again after changing the language. Currently after changing the language the product descriptions are still displayed in English.

imobachgs (approved):

PR #2368, opened 2025-05-19
Monitor

imobachgs:
## Problem

The CLI still relies on D-Bus for some important actions like `agama install`, `agama probe`, etc. Additionally, it still relies on D-Bus for progress reporting.

## Solution

This PR introduces some changes regarding the HTTP API:

* Adapts the old D-Bus based actions to use the HTTP/API.
* Adds two new "agama monitor" and "agama events" commands.

To implement these features, it was needed to add a new `WebSocketClient` which allows reading the events using the WebSocket.

## Testing

- *Tested manually*

## Screenshots

## Documentation

A new PR updating the 

## To do

- [ ] Create the PR updating the documentation.
- [ ] Minor refactoring.
- [ ] Create a new PR dropping D-Bus code that it is not needed anymore.
- [ ] Migrate to Ratatui? Separate PR in any case.


rust/agama-cli/src/lib.rs
```diff
@@ -263,33 +272,46 @@ fn find_client_token(api_url: &Url) -> Option<AuthToken> {
...
 
-pub async fn run_command(cli: Cli) -> Result<(), ServiceError> {
-    // somehow check whether we need to ask user for self-signed certificate acceptance
+async fn build_clients(
+    api_url: Url,
+    insecure: bool,
+) -> anyhow::Result<(BaseHTTPClient, MonitorClient)> {
+    let client = build_http_client(api_url.clone(), insecure, true).await?;
+    let ws_client = build_ws_client(api_url, true).await?;
+    let monitor = Monitor::connect(client.clone(), ws_client).await.unwrap();
```
jreidinger, line 281: this unwrap looks wrong. At least comment why it cannot fail.

PR #2377, opened 2025-05-21
Do not include the root= dracut config in PXE images

teclator:
## Problem

The unconditional set of root= setting in the initrd is problematic in case of PXE images.

From the bug report:

> .... while this might work for live ISO's booting via the dracut live (livenet) module it is a wrong and conflicting setting for the kiwi-dump module code which is active if you are not building a live image but an oem image with installpxe="true". 
> 
> The unconditional setting for root=live:... causes the dracut live module code to be called and not the actual expected kiwi-dump code. When mixing the kernel cmdline parameters for both modules there is also a timing issue on the probing of the network such that sometimes one or the other module wins.
> 
> In short this unconditional root= setting causes a lot of issues and should only be done in the live ISO builds of agama and nowhere else.
> 

- https://bugzilla.suse.com/show_bug.cgi?id=1242976

## Solution

Remove the root= configuration for PXE images.

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed May 21 12:42:34 UTC 2025 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not set root= dracut configuration in PXE images
+  (gh#agama-project/agama2377).
+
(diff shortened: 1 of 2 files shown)


mtomaschewski (commented):
Looks fine to me.

imobachgs (approved):

PR #2378, opened 2025-05-21
Fix write of root config except for PXE images

teclator:
Fix for #2377 as it does the contrary, it sets the root= configuration in PXE images.


lslezak (approved):
LGTM :+1:

PR #2379, opened 2025-05-22
feat: non-blocking and centralized issues

imobachgs:
## Problem

It is a known problem that the software service can get blocked at several stages and it has an impact in the UI and the CLI. We mitigated that problem in #2364 but there are still some cases that are not handled properly.

For instance, during software probing, Agama is not able to return the list of issues. we could argue whether the list of issues is still valid, but that's a different topic. At least, Agama should return something.

## Solution

This PR implements a new `IssuesService` which handles and caches the issues for any service. At this point, it is only used by the software service, but the idea is to use it from all the services.

Apart from caching the issues, it offers a centralized way of keeping track of the issues, listening for changes in single place (instead of building and processing an stream of events for each of the services). For instance, the iSCSI API was not emitting the `IssuesChanged` events because we forgot to add the call to `issues_stream`. Now, it is done automatically.

The service uses [Tokio channels](https://tokio.rs/tokio/tutorial/channels) and message passing to [share the state](https://tokio.rs/tokio/tutorial/shared-state).

> [!NOTE]
> The problem is not fully solved because the "can_install" function still relies on D-Bus, which still blocks.

## Testing

- Tested manually

## Tasks

- [x] Implement an IssuesService.
- [x] Emit the events from the service itself.
- [x] Adopt `IssuesService` in other services.

## Follow ups

Most probably, we should implement a mechanism like this for progress tracking too, although given to the amount of progress signals it is rarely a problem.


rust/agama-server/src/web/common/issues.rs
```diff
@@ -0,0 +1,280 @@
...
+
+impl IssuesService {
+    /// Sets up and starts the service as a Tokio task.
+    ///
+    /// Once it is started, the service waits for:
+    ///
+    /// * Commands from a client ([IssuesClient]).
+    /// * Relevant events from D-Bus.
+    pub async fn start(dbus: zbus::Connection, events: EventsSender) -> IssuesClient {
+        let (tx, rx) = mpsc::channel(16);
```
jreidinger, line 94: why buffer size 16? For issues it looks quite big for me...for progress it would be probably too low.

rust/agama-server/src/web/common/issues.rs
```diff
@@ -0,0 +1,280 @@
...
+        Ok(())
+    }
+
+    /// Handles PropertiesChanged events.
+    fn handle_property_changed(&mut self, message: PropertiesChanged) -> IssuesServiceResult<()> {
+        let args = message.args()?;
+        let inner = message.message();
+        let header = inner.header();
+        let Some(path) = header.path() else {
+            return Ok(());
```
jreidinger, line 147: I think I need comment here why it is Ok to not have header path.

rust/agama-server/src/web/common/issues.rs
```diff
@@ -0,0 +1,280 @@
...
+        };
+
+        // Only process Agama issues signals.
+        if args.interface_name.as_str() != "org.opensuse.Agama1.Issues" {
+            return Ok(());
+        }
+
+        let Some(all) = args.changed_properties().get("All") else {
+            return Ok(());
```
jreidinger, line 156: also here it looks like error for me, so please justify why it is ok

rust/agama-server/src/web/common/issues.rs
```diff
@@ -0,0 +1,280 @@
...
+            path: path.to_string(),
+            issues,
+        };
+        self.events.send(event)?;
+        Ok(())
+    }
+
+    /// Gets the issues for a given D-Bus service and path.
+    async fn get(&mut self, service: &str, path: &str) -> IssuesServiceResult<Vec<Issue>> {
+        if let Some(issues) = self.issues.get(path) {
```
jreidinger, line 177: so this is basically that cache implementation. Maybe it should be mentioned in comment?

rust/agama-utils/src/dbus.rs
```diff
@@ -101,6 +102,21 @@ pub fn extract_id_from_path(path: &OwnedObjectPath) -> Result<u32, zvariant::Err
...
+/// It listens for changes in several objects that are related to a network device.
+pub async fn build_properties_changed_stream(
+    connection: &zbus::Connection,
+) -> Result<MessageStream, zbus::Error> {
+    let rule = MatchRule::builder()
+        .msg_type(MessageType::Signal)
+        .interface("org.freedesktop.DBus.Properties")?
+        .member("PropertiesChanged")?
+        .build();
+    let stream = MessageStream::for_match_rule(rule, connection, Some(1)).await?;
```
jreidinger, line 116: the third parameter really needs comment.  I have to google it to found it is some buffer size.

rust/agama-server/src/web/common/issues.rs
imobachgs: It is not an error, it is simply what we are not interested in this message.

rust/agama-server/src/web/common/issues.rs
imobachgs: Yes, you are right, it is too big. I would reduce it to, e.g., 4.

jreidinger (approved):

PR #2382, opened 2025-05-22
Better reg errors

jreidinger:
## Problem

When adding libzypp service as part of registration failed, it provides ugly error
```
thread 'main' panicked at /home/abuild/rpmbuild/BUILD/agama/agama-lib/src/product/http_client.rs:85:81:
called `Result::unwrap()` on an `Err` value: Error("missing field `id`", line: 1, column: 378)
```

## Solution

Properly report error and wrap it into registration response.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

add service failed


dgdavid (commented):
It seem that it is not part of this PR, but looking at the screenshot in the PR description is clear to me that we should take care of the order of these warnings. The last one should be at the top, after the one about registration.

lslezak (approved):

PR #2386, opened 2025-05-23
Allow to specify a RAID as explicit boot device

ancorgs:
This replaces https://github.com/ancorgs/agama/pull/3 (I was not able to re-target it to the proper repository)

## Problem

Right now a configuration error is reported if `storage.boot.device` points to an entry at `mdRaids`. 

A configuration like this:

```json
{
  "storage":  {
    "mdRaids": [
        {
           "search":     { "max": 1 },
           "alias":      "raid",
           "partitions": [
              { "search": "*", "delete": true },
              { "filesystem": { "path": "/" } }
           ]
         }
    ],
    "boot":    { "configure": true, "device": "raid" }
  }
}
```

...resulted in the error "_There is no boot device with alias 'raid'_".

## Solution

Honor the mentioned setting if the mdRaid entry corresponds to a pre-existing RAID device. Since the configuration is explicit, is the user responsibility to choose a RAID that actually boots.

## Testing

- Extended unit tests for the configuration
- Added a new unit test for the proposal that proves it really does the job if the RAID exists beforehand.

## Not working

If the entry at `mdRaids` corresponds to a new RAID that will be created during the proposal, then the configuration will be accepted but the proposal will not use the new RAID to allocate the boot partitions. That's kind of expected given the nature of the proposal.

Tracked at https://trello.com/c/yuLVe1BR/1585-storage-check-whether-it-makes-sense-to-selecti-a-new-raid-as-boot-device

### DIFF
--- a/service/lib/agama/storage/config.rb
+++ b/service/lib/agama/storage/config.rb
@@ -64,7 +64,7 @@ module Agama
       def boot_device
         return unless boot.configure? && boot.device.device_alias
 
-        drives.find { |d| d.alias?(boot.device.device_alias) }
+        supporting_partitions.find { |d| d.alias?(boot.device.device_alias) }
       end
 
       # Drive config containing root.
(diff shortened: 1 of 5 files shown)


joseivanlopez (commented):
I miss a changelog entry. Other than that, it looks good.

joseivanlopez (approved):
LGTM

PR #2388, opened 2025-05-23
Consider bootable RAIDs when resolving an omitted boot device

ancorgs:
This pull request adapts Agama to some historical behavior of YaST, but (hopefully) in a more structured and explicit way.

## Historical background

YaST considers some software RAIDs to be bootable and, thus, it:
  - offers them as an option in the Guided Setup
  - considers them as candidates for the initial proposal

The criteria used by YaST is kind of arbitrary, based on some heuristic imposed by SUSE partners and refined over time. See [this explanation](https://github.com/yast/yast-storage-ng/blob/db1febb341453c8cc4972dfe36affe55c013f435/src/lib/y2storage/disk_analyzer.rb#L131).

To not break the existing use cases, we should keep the ability in Agama to install on those RAIDs. That’s relevant for:
  1. The initial proposal
  2. The web UI
  3. The selection of the implicit boot device (when is omitted in the configuration)

But we also need the option to manipulate other RAID devices that are not considered to be candidates for a normal installation (ie. bootable). For example, to be able to select them at the web UI with any purpose.

## Implementation

This pull request:
- Sets the basis for all scenarios by providing a clear way to distinguish different types of RAIDs (and drives).
- Improves the situation at 3 (ie. when resolving the omitted boot device)

To address the first goal, `System` now offers four methods `available_drives`, `candidate_drives`, `available_md_raids` and `candidate_md_raids`. Check the documentation for the exact meaning.

To address the latter goal, this introduces the ability to automatically select a bootable RAID as boot device if the configuration omits the boot device but specifies that root partition must be located at a bootable (candidate) RAID.

## Dependencies

This PR depends on https://github.com/yast/yast-storage-ng/pull/1411


service/lib/agama/storage/config_solvers/boot.rb
```diff
@@ -65,72 +76,94 @@ def solve_device_alias
         #
         # The boot device is recursively searched until reaching a drive.
         #
-        # @return [Configs::Drive, nil] nil if the boot device cannot be inferred from the config.
+        # @return [Configs::Drive, Configs::MdDrive, nil] nil if the boot device cannot be inferred
```
joseivanlopez, line 79: Wrong type: `MdDrive`

service/lib/agama/storage/config_solvers/boot.rb
```diff
@@ -65,72 +76,94 @@ def solve_device_alias
...
+        end
+
+        # Recursively looks for the first partitioned config from the given MD RAID.
+        #
+        # @param md_raid [Configs::MdRaid]
+        # @return [Configs::Drive, Configs::MdDrive, nil]
+        def partitionable_from_found_md_raid(md_raid)
+          return md_raid if storage_system.candidate?(md_raid.found_device)
+
+          # TODO: find the correct underlying disk devices for the MD RAID (note they may lack
```
joseivanlopez, line 138: Should Agama actually solve this case or simply complain?

joseivanlopez (commented):

service/lib/agama/storage/config_solvers/boot.rb
ancorgs: I think it is doable, so I don't see why we should not try.

service/lib/agama/storage/config_solvers/boot.rb
ancorgs: Copy&paste is evil. Fixed.

PR #2389, opened 2025-05-23
feat: non-blocking and centralized progress reporting

imobachgs:
## Problem

Follow-up of #2379.

It is a known problem that the software service can get blocked at several stages and it has an impact in the UI and the CLI. We handled part of that problem in #2364 and #2379, but there are still some cases that are not handled properly.

For instance, during software probing, Agama is not able to return the current progress.

## Solution

This PR implements a new `ProgressService` which handles and caches the progress for any service.

Apart from caching the progress, it offers a centralized way of keeping track of the progress, listening for changes in single place (instead of building and processing an stream of events for each of the services).

The service uses [Tokio channels](https://tokio.rs/tokio/tutorial/channels) and message passing to [share the state](https://tokio.rs/tokio/tutorial/shared-state).

As part of this change, I have decided to introduce a new `ProgressChanged` signal, which carries the information we need, instead of relying on the generic `PropertiesChanged`.

## Testing

- Tested manually
- An ISO is [temporarily available](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/non-blocking-progress/images/iso/).

### DIFF
--- a/rust/agama-lib/src/monitor.rs
+++ b/rust/agama-lib/src/monitor.rs
@@ -264,8 +264,6 @@ impl MonitorStatusReader {
             "/manager/progress",
         )
         .await?;
-        // FIXME: do not read the software status yet because it might block
-        // the progress. Enable this line when the software service does not block
         self.add_service_progress(
             &mut status,
             SOFTWARE_PROGRESS_OBJECT_PATH,
--- a/rust/agama-lib/src/monitor.rs
+++ b/rust/agama-lib/src/monitor.rs
@@ -61,7 +61,7 @@ use crate::{
 };
 
 const MANAGER_SERVICE: &str = "org.opensuse.Agama.Manager1";
-// const SOFTWARE_SERVICE: &str = "org.opensuse.Agama.Software1";
+const SOFTWARE_SERVICE: &str = "org.opensuse.Agama.Software1";
 
 #[derive(thiserror::Error, Debug)]
 pub enum MonitorError {
@@ -82,7 +82,7 @@ pub struct MonitorStatus {
     ///
     /// FIXME: do not hold the full status (some elements are not updated)
     pub installer_status: InstallerStatus,
-    /// Progress for each service using the name as the key. If the progress is
+    /// Progress for each service using the D-Bus object path as the key. If the progress is
     /// finished, the entry is removed from the map.
     pub progress: HashMap<String, Progress>,
 }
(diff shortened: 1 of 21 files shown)


rust/agama-lib/src/monitor.rs
```diff
@@ -262,8 +262,8 @@ impl MonitorStatusReader {
             .await?;
         // FIXME: do not read the software status yet because it might block
```
jreidinger, line 263: Does FIXME still apply?

rust/agama-server/src/error.rs
```diff
@@ -42,6 +46,8 @@ pub enum Error {
     Locale(#[from] LocaleError),
     #[error("Issues service error: {0}")]
     Issues(#[from] IssuesServiceError),
+    #[error("Issues service error: {0}")]
```
jreidinger, line 49: Typo, it is not issues service.

rust/agama-server/src/web/common/progress.rs
```diff
@@ -0,0 +1,298 @@
...
+        &mut self,
+        message: ProgressChanged,
+    ) -> ProgressServiceResult<()> {
+        let args = message.args()?;
+        let inner = message.message();
+        let header = inner.header();
+
+        // We are not interested on this message
+        let Some(path) = header.path() else {
+            tracing::warn!("Found a ProgressChanged signal without a path");
```
jreidinger, line 156: maybe add to log also more details like all args and message?

rust/agama-server/src/web/common/progress.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub async fn get(&self, service: &str, path: &str) -> ProgressServiceResult<ProgressSequence> {
+        let (tx, rx) = oneshot::channel();
+        self.0
+            .send(ProgressCommand::Get(
+                service.to_string(),
+                path.to_string(),
+                tx,
+            ))
+            .await
+            .map_err(|_| ProgressServiceError::SendCommand)?;
```
jreidinger, line 227: should not we set also content of original error to help with debugging?

rust/agama-server/src/web/common/progress.rs
imobachgs: Actually, it should not happen. But has the `header()` is defined for all messages, it returns an Option. I have added a comment.

PR #2390, opened 2025-05-23
feat(storage): add boot checks

joseivanlopez:
Add checks to the storage config to avoid using wrong devices as target for boot partitions. For example:

* The device is directly formatted.
* The device is a member of a drive.
* The device is a physical volume.


ancorgs (approved):
LGTM

PR #2393, opened 2025-05-23
feat(storage): configurator

joseivanlopez:
Add a storage configurator for automatically configuring storage.

Till now, the first available drive was used as installation device by default. This implies that storage might not be correctly configured if the first device is not suitable for the installation (e.g., the disk is too small).

Now there is a storage configurator (`Agama::Storage::Configurator`) which takes care of automatically configuring storage if no config has been provided yet. It tries several attempts trying to find a successful config.

Each attempt depends on the product config (e.g., LVM is used or not) and the device used as target for the installation. The order of the candidate devices is the following:

* BOSS devices
* Bootable software RAIDs.
* Bootable drives.

The removable devices (e.g., usb) are used as last resort.


ancorgs (approved):
LGTM, although there is a conflict in the changelog now. Approved anyways.

ancorgs (approved):
Approved again, now without the conflict.

PR #2402, opened 2025-05-26
Network persistent connections handling

teclator:
## Problem

- https://trello.com/c/P6oGGmjH/4745-ui-handle-persistent-non-persistent-network-connections

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/rust/agama-network/src/model.rs
+++ b/rust/agama-network/src/model.rs
@@ -48,6 +48,7 @@ pub struct StateConfig {
     pub devices: bool,
     pub connections: bool,
     pub general_state: bool,
+    pub keep_connections: bool,
 }
 
 impl Default for StateConfig {
@@ -57,6 +58,7 @@ impl Default for StateConfig {
             devices: true,
             connections: true,
             general_state: true,
+            keep_connections: true,
         }
     }
 }
@@ -67,6 +69,7 @@ pub struct NetworkState {
     pub access_points: Vec<AccessPoint>,
     pub devices: Vec<Device>,
     pub connections: Vec<Connection>,
+    pub keep_connections: HashMap<String, bool>,
 }
 
 impl NetworkState {
@@ -81,12 +84,14 @@ impl NetworkState {
         access_points: Vec<AccessPoint>,
         devices: Vec<Device>,
         connections: Vec<Connection>,
+        keep_connections: HashMap<String, bool>,
     ) -> Self {
         Self {
             general_state,
             access_points,
             devices,
             connections,
+            keep_connections,
         }
     }
 
@@ -154,6 +159,8 @@ impl NetworkState {
     ///
     /// It uses the `id` to decide whether the connection already exists.
     pub fn add_connection(&mut self, conn: Connection) -> Result<(), NetworkStateError> {
+        self.keep_connections
+            .insert((&conn.id).to_string(), (&conn.keep).to_owned());
         if self.get_connection(&conn.id).is_some() {
             return Err(NetworkStateError::ConnectionExists(conn.id));
         }
@@ -168,6 +175,8 @@ impl NetworkState {
     ///
     /// Additionally, it registers the connection to be removed when the changes are applied.
     pub fn update_connection(&mut self, conn: Connection) -> Result<(), NetworkStateError> {
+        self.keep_connections
+            .insert((&conn.id).to_string(), (&conn.keep).to_owned());
         let Some(old_conn) = self.get_connection_mut(&conn.id) else {
             return Err(NetworkStateError::UnknownConnection(conn.id.clone()));
         };
@@ -180,6 +189,7 @@ impl NetworkState {
     ///
     /// Additionally, it registers the connection to be removed when the changes are applied.
     pub fn remove_connection(&mut self, id: &str) -> Result<(), NetworkStateError> {
+        self.keep_connections.remove(id);
         let Some(conn) = self.get_connection_mut(id) else {
             return Err(NetworkStateError::UnknownConnection(id.to_string()));
         };
@@ -463,6 +473,7 @@ mod tests {
 pub struct GeneralState {
     pub hostname: String,
     pub connectivity: bool,
+    pub copy_network: bool,
     pub wireless_enabled: bool,
     pub networking_enabled: bool, // pub network_state: NMSTATE
 }
@@ -520,6 +531,8 @@ pub struct Connection {
     pub ieee_8021x_config: Option<IEEE8021XConfig>,
     pub autoconnect: bool,
     pub state: ConnectionState,
+    pub keep: bool,
+    pub filename: String,
 }
 
 impl Connection {
@@ -593,6 +606,8 @@ impl Default for Connection {
             ieee_8021x_config: Default::default(),
             autoconnect: true,
             state: Default::default(),
+            keep: true,
+            filename: Default::default(),
         }
     }
 }
@@ -648,6 +663,7 @@ impl TryFrom<NetworkConnection> for Connection {
         connection.interface = conn.interface;
         connection.mtu = conn.mtu;
         connection.autoconnect = conn.autoconnect;
+        connection.keep = conn.keep;
 
         Ok(connection)
     }
@@ -675,6 +691,8 @@ impl TryFrom<Connection> for NetworkConnection {
             .ieee_8021x_config
             .and_then(|x| IEEE8021XSettings::try_from(x).ok());
         let autoconnect = conn.autoconnect;
+        let keep = conn.keep;
+        let filename = conn.filename;
 
         let mut connection = NetworkConnection {
             id,
@@ -692,6 +710,8 @@ impl TryFrom<Connection> for NetworkConnection {
             mtu,
             ieee_8021x,
             autoconnect,
+            keep,
+            filename,
             ..Default::default()
         };
 
--- a/rust/agama-network/src/nm/client.rs
+++ b/rust/agama-network/src/nm/client.rs
@@ -20,6 +20,8 @@
 
 //! NetworkManager client.
 use std::collections::HashMap;
+use std::fs::{self, OpenOptions};
+use std::path::Path;
 
 use super::builder::DeviceFromProxyBuilder;
 use super::dbus::{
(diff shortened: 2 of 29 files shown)


rust/agama-network/src/nm/client.rs
```diff
@@ -71,25 +73,41 @@ impl<'a> NetworkManagerClient<'a> {
...
+        let copy_network = !Path::new("/run/agama/not_copy_network").exists();
 
         if wireless_enabled != state.wireless_enabled {
             self.nm_proxy
                 .set_wireless_enabled(state.wireless_enabled)
                 .await?;
         };
 
+        if copy_network != state.copy_network {
+            let path = Path::new("/run/agama/not_copy_network");
```
imobachgs: Duplicated, not to mention that perhaps we should move `/run/agama/not_copy_network` to a const.

rust/agama-network/src/model.rs
```diff
@@ -552,10 +555,19 @@ impl Connection {
...
+        self.status == Status::Down
+    }
+
     pub fn set_up(&mut self) {
         self.status = Status::Up
     }
 
+    pub fn set_keep(&mut self, keep: bool) {
+        self.keep = keep;
+        self.status = Status::Keep
```
imobachgs, line 569: Why do we have a `Status::Keep` value in addition to the `keep` boolean? I would expect both concepts to be orthogonal.

rust/agama-network/src/nm/client.rs
```diff
@@ -214,11 +232,14 @@ impl<'a> NetworkManagerClient<'a> {
                     }
 
                     Self::add_secrets(&mut connection.config, &proxy).await?;
+                    connection.flags = flags;
+                    connection.keep = if flags != 0 { false } else { true };
```
imobachgs: Would this work?

```suggestion
                    connection.keep = flags == 0;
```

rust/agama-server/src/network/web.rs
```diff
@@ -121,8 +121,10 @@ pub async fn network_service<T: Adapter + Send + Sync + 'static>(
                 .put(update_connection)
                 .get(connection),
         )
-        .route("/connections/:id/connect", patch(connect))
-        .route("/connections/:id/disconnect", patch(disconnect))
+        .route("/connections/:id/connect", post(connect))
+        .route("/connections/:id/disconnect", post(disconnect))
+        .route("/connections/:id/keep", post(keep))
```
imobachgs: To be honest, I am not sure about this API for setting the `keep` value. I have two comments about this:

1. We need to make clear why it cannot be just a regular update (or a regular update + an apply).

2. The path looks wrong to me. In the code below, you act on all connections if `:id == "all"` which looks like a workaround to me. Given that you are issuing a `post` call, I would say you could use the body to specify the list of connections:

```json
{ "connections": ["id1", "id2"] }
```

And in that context, you might consider that not specifying them  is working with all the connections. Moreover, you could even use a single action and use the body to set the `keep` value too.

```json
{ "connections": ["id1", "id2"], "keep": true }
```

rust/agama-server/src/network/web.rs
```diff
@@ -427,6 +425,100 @@ async fn disconnect(
...
+            return Err(NetworkError::UnknownConnection(id));
+        };
+
+        conn.set_keep(true);
+
+        state
+            .network
+            .update_connection(conn)
+            .await
+            .map_err(|_| NetworkError::CannotApplyConfig)?;
```
imobachgs: This error is too generic. We should improve the error hierarchy because the real cause of the problem is missing.

service/lib/agama/http/clients/base.rb
```diff
@@ -43,6 +43,15 @@ def post(path, data)
           @logger.warn "server returned #{response.code} with body: #{response.body}"
         end
 
+        # send GET request with given path.
+        # @param path[String] path relatived to `api`` endpoint.
+        def get(path)
```
imobachgs, line 49: Please, document the `@return` value. I know, the `post` is not documented, but we should do that too.

imobachgs (changes requested):
I have reviewed the backend part. In general, I do not like the API and I think it deserves some work before merging to `master`.

web/src/components/core/Annotation.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import React from "react";
+import { screen } from "@testing-library/react";
+import { plainRender } from "~/test-utils";
+import Annotation from "./Annotation";
+
+describe("Annotation", () => {
+  it('renders the default "emergency" icon when no icon is provided', () => {
+    const { container } = plainRender(<Annotation>Configured for installation only</Annotation>);
```
imobachgs, line 30: For this cases I guess you can use the `render` function from React Testing Library. `plainRender` injects the query client too (although it is not a problem).

web/src/components/core/SwitchEnhanced.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+        isChecked={false}
+        onChange={onChangeMock}
+      />,
+    );
+
+    const switchElement = screen.getByRole("switch");
+
+    await user.click(switchElement);
+
+    expect(onChangeMock).toHaveBeenCalledTimes(1);
```
imobachgs: Is it necessary to check whether the switch has updated its value?

web/src/components/network/NoPersistentConnectionsAlert.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+let mockConnections: Connection[];
+
+jest.mock("~/queries/network", () => ({
+  ...jest.requireActual("~/queries/network"),
+  useConnections: () => mockConnections,
+}));
+
+describe("NoPersistentConnectionsAlert", () => {
+  describe("when there are connections to be kept", () => {
```
imobachgs:
```suggestion
  describe("when there are persistent connections", () => {
```

web/src/components/network/NoPersistentConnectionsAlert.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      ];
+    });
+
+    it("renders nothing", () => {
+      const { container } = plainRender(<NoPersistentConnectionsAlert />);
+      expect(container).toBeEmptyDOMElement();
+    });
+  });
+
+  describe("when there are no connections to be kept", () => {
```
imobachgs:
```suggestion
  describe("when there are no persistent connections", () => {
```

web/src/components/network/WifiConnectionDetails.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    within(section).getByText("192.168.69.201/24");
+    // DNS
+    within(section).getByText("192.168.69.100");
+    // Gateway 4
+    within(section).getByText("192.168.69.4");
+    // Gateway 6
+    within(section).getByText("192.168.69.6");
+  });
+
+  it("renders link for editing connection", () => {
```
imobachgs:
```suggestion
  it("renders link for editing the connection", () => {
```

web/src/components/network/WifiNetworksList.test.tsx
```diff
@@ -142,6 +142,82 @@ describe("WifiNetworksList", () => {
       });
     });
 
+    describe("and the connection is selected to be kept", () => {
```
imobachgs:
```suggestion
    describe("and the connection is persistent", () => {
```

web/src/components/network/WifiNetworksList.test.tsx
```diff
@@ -142,6 +142,82 @@ describe("WifiNetworksList", () => {
...
+      });
+
+      it("does not render any hint", () => {
+        // @ts-expect-error: you need to specify the aria-label
+        installerRender(<WifiNetworksList />);
+        expect(screen.queryByText("Configured for installation only")).toBeNull;
+      });
+    });
+
+    describe("and the connection is not selected to be kept", () => {
```
imobachgs:
```suggestion
    describe("and the connection is not persistent", () => {
```

web/src/components/network/WiredConnectionsList.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    it("does not render any hint", () => {
+      // @ts-expect-error: you need to specify the aria-label
+      installerRender(<WiredConnectionsList />);
+      expect(screen.queryByText("Configured for installation only")).toBeNull;
+    });
+  });
+
+  describe("and the connection is not selected to be kept", () => {
```
imobachgs:
```suggestion
  describe("and the connection is not persistent", () => {
```

web/src/components/network/WiredConnectionsList.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+jest.mock("~/queries/network", () => ({
+  ...jest.requireActual("~/queries/network"),
+  useNetworkChanges: jest.fn(),
+  useNetworkDevices: () => [mockDevice],
+  useConnections: () => mockConnections,
+}));
+
+describe("WiredConnectionsList", () => {
+  describe("and the connection is selected to be kept", () => {
```
imobachgs:
```suggestion
  describe("and the connection is persistent", () => {
```

imobachgs (commented):
I reviewed the TypeScript code and it is looking fine. Just some minor comments.

web/src/components/core/Annotation.test.tsx
dgdavid: Agree. Something to think about, but not in this PR but for when we take the time to improve the overall testing approach.. Thanks.

web/src/components/core/SwitchEnhanced.test.tsx
dgdavid: I think not strictly necessary, but I'll have a look with fresh eyes now. If I remember correctly, there was a problem to do so because it is acting as a controlled component here that never gets its `isChecked` prop updated because of the onChangeMock. Let's see if I manage to improve the test for making it more complete and useful. Thanks.

dgdavid: Have a look to https://github.com/agama-project/agama/pull/2402/commits/f926da287ebbe4167ce2acf577b6a43931acd0cc, does it looks better now?

rust/agama-network/src/model.rs
teclator: The _Status::Keep_ has been introduced to apply changes without activating / deactivating a connection. I currently do not like we call the activation / deactivation in the update_connection call, but while we rethink that approach I wanted to only call _Update2_ method to update the profile.

rust/agama-network/src/nm/client.rs
```diff
@@ -96,14 +98,17 @@ impl<'a> NetworkManagerClient<'a> {
         };
 
         if copy_network != state.copy_network {
-            let path = Path::new("/run/agama/not_copy_network");
             if state.copy_network {
-                if let Err(error) = fs::remove_file(path) {
-                    tracing::error!("Cannot remove /run/agama/not_copy_network file {:?}", error);
+                if let Err(error) = fs::remove_file(not_copy_path) {
+                    tracing::error!("Cannot remove {:?} file {:?}", NOT_COPY_NETWORK_PATH, error);
```
imobachgs:
```suggestion
                    tracing::error!("Cannot remove {} file {:?}", NOT_COPY_NETWORK_PATH, error);
```

rust/agama-network/src/nm/client.rs
```diff
@@ -96,14 +98,17 @@ impl<'a> NetworkManagerClient<'a> {
...
                 }
             } else {
-                if let Err(error) = OpenOptions::new().create(true).write(true).open(path) {
-                    tracing::error!("Cannot write /run/agama/not_copy_network file {:?}", error);
+                if let Err(error) = OpenOptions::new()
+                    .create(true)
+                    .write(true)
+                    .open(not_copy_path)
+                {
+                    tracing::error!("Cannot write {:?} file {:?}", NOT_COPY_NETWORK_PATH, error);
```
imobachgs:
```suggestion
                    tracing::error!("Cannot write {} file {:?}", NOT_COPY_NETWORK_PATH, error);
```

PR #2403, opened 2025-05-26
Simple Rescue System

kobliha:
- Implements a rescue system by booting to the Installer Live Media image with special systemd parameters that disable booting to the graphical mode and starting Agama service
- Some services are kept, such as DUD, Self-Update, root password

## Problem

The current media do not have a Rescue System but users are asking for it

- https://trello.com/c/Kgaovgu8

## Solution

This PR does not implement any special Rescue System. It only reuses what we already have and prevents Agama from starting. 

## Testing

- Parameters were tested manually on a Kernel command line

### DIFF
--- a/live/config-cdroot/fix_bootconfig.aarch64
+++ b/live/config-cdroot/fix_bootconfig.aarch64
@@ -97,6 +97,13 @@ menuentry "Check Installation Medium" --class os --unrestricted {
     echo Loading initrd...
     initrd (\$root)/boot/aarch64/loader/initrd
 }
+menuentry "Rescue System" --class os --unrestricted {
+    set gfxpayload=keep
+    echo Loading kernel...
+    linux (\$root)/boot/aarch64/loader/linux \${extra_cmdline} \${isoboot} $rescue_system_boot_settings
+    echo Loading initrd...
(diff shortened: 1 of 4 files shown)


imobachgs (commented):
It looks fine. Did you try it?

Please, update [the changes file](https://github.com/agama-project/agama/blob/master/live/src/agama-installer.changes).

imobachgs (dismissed):

lslezak (approved):
Looks good as the initial implementation for some rescue system.

However there are still some questions:
- Which services should be actually started in the rescue mode? Should we e.g. configure the network?
- Systemd already provides it's own rescue mode, boot with `systemd.unit=rescue.target live.password=password` (The rescue target requires a configured root password, that's why I updated the `live-password.service` file in this PR to allow setting the password in the rescue mode as well.)
- Implementation detail: Maybe we should define a separate systemd target for this?

kobliha:
Thanks @lslezak, yes I thought about a special target too, but it's a bigger thing and requires a better plan, more impl. time and better testing. I simply went for the simplest solution as we don't have the requirements yet. Anyway, you should open an issue or a discussion here at GH for the topic.

PR #2416, opened 2025-05-28
Update sles_sap_160.yaml

ajaeger:
Add SLES patterns to SLES for SAP

## Problem

SLES for SAP pattern selection only shows the SLES for SAP specific ones but misses the base SLES patterns.

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1243757

## Solution

Copy over patterns from SLES to SLES for SAP

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed May 28 15:04:21 UTC 2025 - Andreas Jaeger <aj@suse.com>
+
+- Add all SLES 16.0 patterns to SLES for SAP 16.0 as well (bsc#1243757).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


angelabriel:
To be honest I'm not really happy that we need to copy the user_pattern definition from sles_160.yaml to sles_sap_160.yaml to get the `Software selection` working on all media (full and online) after register the system to SCC. That means for the future we need to monitor the file sles_160.yaml for changes to duplicate all these changes to SLES for SAP too. This will probably not work fluently and SLES for SAP will always be behind.
Do we have any other way to solve the problem?

angelabriel (approved):
The changes look good to me.
For my general concerns look at my previous comment

imobachgs (approved):

angelabriel:
> Would it work for you?

yes of course and many thanks for the explanation.

PR #2419, opened 2025-05-29
feat(storage): allow selecting software RAIDs for booting

joseivanlopez:
Adapt the storage UI to offer the bootable software RAIDs (a.k.a., candidate RAIDs) as selectable options in the page for configuring the boot device.

The D-Bus/HTTP APIs are extended to distinguish between available/candidate drives and MD RAIDs.


web/src/components/overview/StorageSection.test.tsx
```diff
@@ -160,8 +155,13 @@ describe("when the configuration contains several drives", () => {
 
   describe("but one of them has a different space policy", () => {
     beforeEach(() => {
-      mockUseConfigModelFn.mockReturnValue({
-        drives: [sdaDrive, { ...sdbDrive, spacePolicy: "resize" }],
+      addMockApi({
```
dgdavid, line 158: As mentioned in IRC, I don't have a strong opinion on this approach at this moment. That said, I really appreciate the effort you're putting into simplifying and improving the way we write tests :heart: Thanks!

Yesterday I took another look at [MSW](https://mswjs.io/) and its [comparison](https://mswjs.io/docs/comparison) with [Nock](https://github.com/nock/nock), which is what's used in the TanStack Query examples. See https://tanstack.com/query/latest/docs/framework/react/guides/testing. I think it would be a good tool for helping us mocking the API in a cleaner way while giving us a ton of features that might help in complex tests.

Thus, I think **it would be worth allocating some dedicated time to work on a _testing revamp_** and, as discussed yesterday, start writing tests for hooks as well.

That being said, I'm fine moving forward with your current approach for now if it works for your use case. We can always revisit and refine it later if needed, in the allocated _testing time_ I mentioned above.

One small question: would it be possible to get rid of the `beforeEach`s and instead use something like `mockAPI("endpoint", data)` directly?

web/src/mocks/api.ts
```diff
@@ -0,0 +1,69 @@
...
+  "/api/storage/devices/available_drives"?: Awaited<
+    ReturnType<typeof apiStorage.fetchAvailableDrives>
+  >;
+  "/api/storage/devices/available_md_raids"?: Awaited<
+    ReturnType<typeof apiStorage.fetchAvailableMdRaids>
+  >;
+  "/api/storage/devices/system"?: Device[];
+  "/api/storage/config_model"?: Awaited<ReturnType<typeof apiStorage.fetchConfigModel>>;
+  "/api/storage/issues"?: Awaited<ReturnType<typeof apiIssues.fetchIssues>>;
+};
```
dgdavid, line 41: That was clever! :nerd_face:  As you said, it make us to _be safe of changes in the API_

web/src/components/storage/BootSelection.tsx
```diff
@@ -54,35 +60,45 @@ type BootSelectionState = {
...
     if (state.load) return;
 
+    const boot = model?.boot;
     let selectedOption: string;
 
-    if (!boot.configure) {
+    if (!boot?.configure) {
       selectedOption = BOOT_DISABLED_ID;
-    } else if (boot.isDefault) {
+    } else if (boot?.isDefault) {
```
ancorgs: If I understand this correctly, an omitted `boot` in the model would mean the selector is set to "manual". If such an omitted `boot` in the model is possible? Shouldn't it mean automatic boot?

ancorgs: And if it's impossible to get an omitted boot from the model, why all the defensive code?

joseivanlopez: `boot` cannot be omitted but `model` can be null. I am going to explicity protect in case of a `null` model. Thanks!

web/src/components/overview/StorageSection.test.tsx
joseivanlopez: As we have discussed in other channels, this mocking is a POC. I would not invest more time for now. Thanks!

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Mon Jun  9 06:12:23 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Extend the D-Bus API to get avialable/candidate drives and MD
```
ancorgs: Typo: avialable

service/lib/agama/storage/config_checkers/md_raid.rb
```diff
@@ -146,19 +146,59 @@ def reused_member_issues
...
+        def deleted_reused_member_issue(member_config)
+          return unless storage_config.supporting_delete.include?(member_config)
+          return unless member_config.delete? || member_config.delete_if_needed?
+
+          error(
+            format(
+              _(
+                # TRANSLATORS: %{member} is replaced by a device name (e.g., "/dev/vda") and
+                #   %{md_raid} is replaced by a MD RAID name (e.g., "/dev/md0").
+                "The device '%{member}' cannot be deleted because it is a device member of the " \
```
ancorgs: The expression "...it is a device member of the reused MD RAID..." is probably technically correct, but a bit weird. What about saying simply "...it is part of the RAID..."

service/lib/agama/storage/config_checkers/md_raid.rb
```diff
@@ -146,19 +146,59 @@ def reused_member_issues
...
+          return unless storage_config.supporting_size.include?(member_config)
+          return if member_config.size.default?
+
+          error(
+            format(
+              _(
+                # TRANSLATORS: %{member} is replaced by a device name (e.g., "/dev/vda") and
+                #   %{md_raid} is replaced by a MD RAID name (e.g., "/dev/md0").
+                "The device '%{member}' cannot be resized because it is a device member of the " \
+                "reused MD RAID %{md_raid}"
```
ancorgs: Same than above.

service/lib/agama/storage/config_checkers/md_raid.rb
```diff
@@ -229,6 +269,40 @@ def target_reused_member_issue(member_config)
...
+          parent_config = find_config(device.partitionable)
+          return unless parent_config&.filesystem
+
+          error(
+            format(
+              _(
+                # TRANSLATORS: %{device} and %{member} are replaced by a device name (e.g.,
+                # "/dev/vda") and %{md_raid} is replaced by a MD RAID name (e.g., "/dev/md0").
+                "The device '%{device}' cannot be formatted because its partition '%{member}' is " \
+                "a device member of the reused MD RAID %{md_raid}"
```
ancorgs: Same comment than above.

rust/agama-server/src/storage/web.rs
```diff
@@ -448,22 +451,71 @@ async fn actions(State(state): State<StorageState<'_>>) -> Result<Json<Vec<Actio
...
-/// Gets the SID (Storage ID) of the devices usable for the installation.
-///
-/// Note that not all the existing devices can be selected as target device for the installation.
+/// Gets the SID (Storage ID) of the available drives for the installation.
+#[utoipa::path(
+    get,
+    path = "/devices/available_drives",
+    context_path = "/api/storage",
+    responses(
+        (status = 200, description = "Lis of SIDs", body = Vec<DeviceSid>),
```
kobliha, line 460: What is Lis of SIDs? Is it **List** of SIDs?

rust/agama-server/src/storage/web.rs
```diff
@@ -448,22 +451,71 @@ async fn actions(State(state): State<StorageState<'_>>) -> Result<Json<Vec<Actio
...
+    Ok(Json(sids))
+}
+
+/// Gets the SID (Storage ID) of the candidate drives for the installation.
+#[utoipa::path(
+    get,
+    path = "/devices/candidate_drives",
+    context_path = "/api/storage",
+    responses(
+        (status = 200, description = "Lis of SIDs", body = Vec<DeviceSid>),
```
kobliha, line 477: The same as above, List?

rust/agama-server/src/storage/web.rs
```diff
@@ -448,22 +451,71 @@ async fn actions(State(state): State<StorageState<'_>>) -> Result<Json<Vec<Actio
...
+    Ok(Json(sids))
+}
+
+/// Gets the SID (Storage ID) of the available MD RAIDs for the installation.
+#[utoipa::path(
+    get,
+    path = "/devices/available_md_raids",
+    context_path = "/api/storage",
+    responses(
+        (status = 200, description = "Lis of SIDs", body = Vec<DeviceSid>),
```
kobliha, line 494: +1 ^^^

PR #2424, opened 2025-05-29
Add custom repos

jreidinger:
## Problem

There is not support for additional repositories that can be used for installation.

## Solution

Add support to unattended installation.

## Follow ups

Currently there is no easy way to specify GPG key checksum for custom repositories. As result there was question appeared in web UI. Probably security section should be extended and added to https://github.com/agama-project/agama/blob/master/service/lib/agama/software/callbacks/signature.rb

## Testing

- *Added a new unit test*
- *Tested manually*

Tested with this profile to add Games repo and install software available only in it:

```json
{
  "product": {
    "id": "Tumbleweed"
  },
  "software": {
    "extraRepositories": [
	{
	  "alias": "games",
	   "url": "https://download.opensuse.org/repositories/games/openSUSE_Tumbleweed/"
        }
    ],
    "packages": ["AtomicWorm"]
  }
}
```

Just expect that in web UI question to accept GPG key will be raised.


rust/agama-lib/share/profile.schema.json
```diff
@@ -183,6 +183,40 @@
             "type": "string",
             "examples": ["vim"]
           }
+        },
+        "extraRepositories": {
+          "title": "List of user specified repositories that will be used on top of default ones",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": ["alias", "url"],
```
lslezak, line 192: The alias could be optional, we could generate some random unique ugly name if missing. But I'm not sure how that would be used by users.

Personally I'd like to specify a well known meaningful name instead of some random one.

lslezak (approved):
LGTM just a question whether we really want to have the alias mandatory.

rust/agama-lib/share/profile.schema.json
jreidinger: me too, as I would like to use alias as id if we need to address repo elsewhere in config

PR #2431, opened 2025-06-02
Network persistent connections handling RC1

teclator:
Changes added by #2402 but targeting [rc1.](https://github.com/agama-project/agama/tree/rc1)


rust/agama-network/src/model.rs
```diff
@@ -692,6 +711,8 @@ impl TryFrom<Connection> for NetworkConnection {
             mtu,
             ieee_8021x,
             autoconnect,
+            keep,
+            filename,
```
imobachgs, line 715: for master: is the filename really needed?

rust/agama-network/src/model.rs
```diff
@@ -593,6 +606,9 @@ impl Default for Connection {
             ieee_8021x_config: Default::default(),
             autoconnect: true,
             state: Default::default(),
+            keep: true,
```
imobachgs, line 609: master: Should we consider using `copy`? For me it sounds inconsistent having `keep` and `copy`.

rust/agama-network/src/nm/client.rs
```diff
@@ -231,8 +253,11 @@ impl<'a> NetworkManagerClient<'a> {
         }
 
         for conn in connections.iter_mut() {
-            if controlled_by.contains_key(&conn.uuid) {
-                conn.controller = Some(conn.uuid);
+            // FIXME: Is this OK?
```
imobachgs, line 256: ?

rust/agama-network/src/nm/client.rs
```diff
@@ -266,18 +291,33 @@ impl<'a> NetworkManagerClient<'a> {
         let path = if let Ok(proxy) = self.get_connection_proxy(conn.uuid).await {
             let original = proxy.get_settings().await?;
             let merged = merge_dbus_connections(&original, &new_conn)?;
-            proxy.update(merged).await?;
+            // https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMSettingsConnectionFlags
+            // 0x1 persist to disk, 0x8 memory only
+            let persist = if conn.keep { 0x1 } else { 0x8 };
```
imobachgs, line 296: master: it would be nice to keep to use constants with good names for those flags.

rust/agama-network/src/system.rs
```diff
@@ -427,7 +427,12 @@ impl<T: Adapter> NetworkSystemServer<T> {
     /// Writes the network configuration.
     pub async fn write(&mut self) -> Result<(), NetworkAdapterError> {
         self.adapter.write(&self.state).await?;
-        self.state = self.adapter.read(StateConfig::default()).await?;
+        self.state = self
+            .adapter
+            .read(StateConfig {
+                ..Default::default()
```
imobachgs, line 433: master: Why not `StateConfig::default()`?

rust/agama-server/src/network/web.rs
```diff
@@ -121,8 +121,10 @@ pub async fn network_service<T: Adapter + Send + Sync + 'static>(
                 .put(update_connection)
                 .get(connection),
         )
-        .route("/connections/:id/connect", patch(connect))
-        .route("/connections/:id/disconnect", patch(disconnect))
+        .route("/connections/:id/connect", post(connect))
+        .route("/connections/:id/disconnect", post(disconnect))
+        .route("/connections/:id/keep", post(keep))
```
imobachgs, line 126: master: I am not sure whether it is a separate action. Let's open the discussion later.

rust/agama-server/src/network/web.rs
```diff
@@ -427,6 +426,100 @@ async fn disconnect(
...
+    context_path = "/api/network",
+    responses(
+      (status = 204, description = "Keep the given connection after the installation", body = String)
+    )
+)]
+async fn keep(
+    State(state): State<NetworkServiceState>,
+    Path(id): Path<String>,
+) -> Result<impl IntoResponse, NetworkError> {
+    if id == "all" {
```
imobachgs, line 441: master: IMHO there is too much logic in this function.

imobachgs (approved):
We need these fixes in RC1. But I would like to have a closer look when working on the `master` branch version.

rust/agama-network/src/model.rs
teclator: not anymore

rust/agama-network/src/nm/client.rs
teclator: already fixed :)

rust/agama-network/src/system.rs
teclator: leftover, it should be default again, during a time it passes keep_connections as false, but that is ont the case anymore

rust/agama-server/src/network/web.rs
teclator: probably should be the same with an option to set it as true or false instead and the same to connect and disconnect or it even could be a patch of the connection just with the keep and status fields

rust/agama-server/src/network/web.rs
teclator: I agree

rust/agama-network/src/model.rs
teclator: what about: **persistent** ... at the end it is about persist or not to disk using the **enum NMSettingsAddConnection2Flags** or **enum NMSettingsUpdate2Flags**

teclator: I mean, this is the connection one, so, for me having it a persist or persistent looks better while for the general state it is about copying or not copying the persistent configuration.

PR #2258, opened 2025-04-10
Add bridge support for the HTTP API and CLI

teclator:
## Problem

Although there is a bridge model in the backend, there is no way to configure it through the CLI or through the HTTP api.

- https://trello.com/c/LoDehu4m/4559-1-agama-add-support-for-configuring-a-bridge

## Solution

Added bridge support similar to what was added for bonding in (https://github.com/agama-project/agama/pull/885) and also fix it because looks like we broke the bonding handling when moved to the HTTP Api (https://github.com/agama-project/agama/pull/1064).

An example of the json bridge configuration looks like:

```json
{
  "network": {
    "connections": [
      {
        "id": "Bridge0",
        "method4": "manual",
        "interface": "br0",
        "addresses": ["192.168.1.100/24"],
        "gateway4": "192.168.1.1",
        "nameservers": ["192.168.1.1"],
        "bridge": {
          "ports": ["eth0", "eth1"],
          "stp": false
        }
      }
    ]
  }
}
```

## Testing

- *Added a new unit test*
- *Tested manually*

PR #2433, opened 2025-06-03
Use slave-type instead of port-type for NM<1.46.0

jcronenberg:
## Problem

With #2356 I unfortunately missed the changes from #2258 where `slave-type` was changed to `port-type` which only NM >= 1.46.0 supports.

## Solution

Send `port-type` only for NM >= 1.46.0, otherwise `slave-type` is used.
Also since I was already adding a test for this, I also added two additional test for `connection_to_dbus` with bridge and bond connections.

## Testing

- *Added a new unit test*
- *Tested manually*


rust/agama-network/src/nm/dbus.rs
```diff
@@ -85,7 +89,11 @@ pub fn connection_to_dbus<'a>(
...
         connection_dbus.insert("autoconnect", false.into());
     } else {
-        connection_dbus.insert("port-type", "".into());
+        if VersionReq::parse(">=1.46.0").unwrap().matches(&nm_version) {
+            connection_dbus.insert("port-type", "".into());
+        } else {
+            connection_dbus.insert("slave-type", "".into());
+        }
         connection_dbus.insert("master", "".into());
     }
```
jcronenberg, line 98: Btw, I might be missing something but isn't this nonsensical to have, because it seems to me to get cleaned up later by `cleanup_dbus_connection()` anyway? So should I maybe just remove this part completely?

jcronenberg: Mhm maybe it's relevant for merging connections? Though if the empty value then gets cleaned up before it's sent to NM, NM will just keep the value as is anyway so it will also not be unset/updated.

imobachgs: Yes, I have the same impression. That value will be cleaned-up later. I guess we need to have a closer look to this.

teclator: I found some issue when working on this exactly because of the merge and also when for example moving a port from a bond to a bridge. It was reading the port-type but setting the slave-type or something like that..

jcronenberg: Okay so I tested it a bit, indeed the merging seems to be a problem because for whatever reason NM still seems to send `slave-type` over `port-type` to agama and then the merging gets difficult, so I have now added a bit in `cleanup_dbus_connection()` that prefers the `port-type` of agama so that it takes precedence. And also deleting the section I mentioned here does not work because then the merge wouldn't overwrite the values/`cleanup_dbus_connection()` wouldn't see a change. But also what I mentioned previously
> Though if the empty value then gets cleaned up before it's sent to NM, NM will just keep the value as is anyway so it will also not be unset/updated.

isn't correct in this case, if nothing is sent to NM it will actually unset this property, so `cleanup_dbus_connection()` is working as intended.

imobachgs (approved):

rust/agama-network/src/nm/dbus.rs
imobachgs: Thanks for checking it. Then I guess we can proceed.

PR #2434, opened 2025-06-03
refactor(web): update software query keys to use array form

dgdavid:
## Problem

A caching issue in the `l10n` namespace was [recently fixed](https://github.com/agama-project/agama/pull/2428) by switching from string-based to array-based TanStack query keys. The `software` namespace still uses the string form, which could lead to similar out-of-sync issues.

## Solution
Update query keys in the software namespace to use the array form for consistency and to avoid caching problems.

## Testing

Manually verified that everything still works as expected.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jun  5 15:35:42 UTC 2025 - David Diaz <dgonzalez@suse.com>
+
+- Improve caching in the software area by using array-form
+  query keys (gh#agama-project/agama#2434).
+
+-------------------------------------------------------------------
 Thu Jun  5 11:26:05 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
 - The "InstallationPhaseChanged", "ProductChanged" and


web/src/queries/software.ts
```diff
@@ -166,9 +166,9 @@ const useConfigMutation = () => {
   const query = {
     mutationFn: updateConfig,
     onSuccess: async (_, config: SoftwareConfig) => {
-      queryClient.invalidateQueries({ queryKey: ["software/config"] });
-      queryClient.invalidateQueries({ queryKey: ["software/product"] });
-      queryClient.invalidateQueries({ queryKey: ["software/proposal"] });
+      queryClient.invalidateQueries({ queryKey: ["software", "config"] });
+      queryClient.invalidateQueries({ queryKey: ["software", "product"] });
+      queryClient.invalidateQueries({ queryKey: ["software", "proposal"] });
```
dgdavid, line 171: I wonder if we should invalidate the whole "software" cache instead. Note that this is triggered on a config mutation.

jreidinger: I would say that it does not hurt to invalidate it all...another option is to invalidate only parts that are affected.

dgdavid: > I would say that it does not hurt to invalidate it all...

That is what I think, but asking because I'm not sure if it was in that shape because of a reason


>another option is to invalidate only parts that are affected.

Which is the current behavior :wink:

imobachgs (approved):

web/src/queries/software.ts
dgdavid: I'm going to keep it as it is by now. We can adapt it later. Thanks anyway for the feedback @jreidinger

PR #2442, opened 2025-06-05
fix(rust): fix software-related tests

imobachgs:
## Problem

* Some tests got broken in #2424.
* Additionally, the extra repositories element was not using camel case.

## Solution

Adapt the tests to the new code.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Jun  5 12:57:11 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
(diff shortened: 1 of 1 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Jun  5 12:57:11 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Use camel case for the extra repositories key (#agama-project/agama#2442).
```
dgdavid:
```suggestion
- Use camel case for the extra repositories key (gh#agama-project/agama#2442).
```

PR #2443, opened 2025-06-05
Fixed UI event handling

lslezak:
## Problem

- The web UI is stuck and displays the spinner until the installation is finished.
- https://bugzilla.suse.com/show_bug.cgi?id=1241208

## Details

- When the "status" query data was not obtained yet the "InstallationPhaseChanged", "ProductChanged" and "IssuesChanged" web socket events were ignored.
- This could possibly get the UI stuck when the loading screen was displayed because of the `selectedProduct === undefined && isBusy` condition.
- If the "ProductChanged" event was ignored in that case the loading screen would be displayed until the installer is busy (i.e. until the end of installation) because the selected product would not be refreshed.

## Solution

- Handle those events even when the "status" is not defined.
- Add more logging to see more details if this fix is still not enough.

## Testing

- The bug happens only rarely and requires a very specific timing between the web browser requests and the command line starting automatic installation in background so I can not verify that it really fixes the problem.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun  5 11:26:05 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- The "InstallationPhaseChanged", "ProductChanged" and
+  "IssuesChanged" events were ignored in some situation
+  (possibly causing bug bsc#1241208)
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
Thanks for this @lslezak 

I know these `console.log`s don’t hurt, and I understand the reasoning behind them, especially since you explained it clearly in the PR description. That said, I’m a bit unsure about adding this much logging 🙈

Nitpick aside, I’m fine moving forward with it for now. But, ideally, these logs should at least be conditionally excluded from final images in the long run. Not a big deal, though.

PR #2440, opened 2025-06-04
chore(web): update dependencies 2025-06-04

dgdavid:
Update web dependencies to their latest versions except React and React Router DOM and related ones, since both has published major versions that should be updated in separated PRs with their corresponding migrations after 
PatternFly supports React 19.

---

Following commands has been executed with success, 

* `npm run eslint`
* `npm run stylelint`
* `npm run check-types`
* `npm run server`
* `npm run build`
* `NODE_ENV=production npm run build`

Tested manually too using https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/update-deps-20250603/images/iso/agama-installer.x86_64-openSUSE.iso from https://build.opensuse.org/project/show/systemsmanagement:Agama:branches:update-deps-20250603 and performing below actions

* Select a product
* Download logs
* Change a product
* Change language and keyboard
* Change hostname
* Change timezone
* Edit a network connection
* Edit the storage proposal (using a LVM)
* Edit software selection
* Create first user
* Create password for root
* Start the installation

## Note for reviewers

* To see all updated dependencies, please check the commit history.
* Some unused dependencies were also removed.
* This update was performed using [npm-check-updates](https://www.npmjs.com/package/npm-check-updates).
* It also [prevent dev server from hanging on ESLint error](https://github.com/agama-project/agama/pull/2440/commits/a7f9c45d23c7148197f7eefd960537d68cbe074a), reason why a previous update was [reverted](#2353).

---

Previous updates:

 * https://github.com/agama-project/agama/pull/2353
 * https://github.com/agama-project/agama/pull/2315

PR #2444, opened 2025-06-05
web: fix package build in OBS

lslezak:
## Problem

- The package does not build in OBS
- It fails with error `ERROR in [eslint] Cannot find module 'eslint'`

## Solution

- Disable Eslint checks in RPM builds

## Testing

- Tested manually in a local build


dgdavid (commented):
As said in IRC, I'd keep the linter _alive_ for production too and simply revert https://github.com/agama-project/agama/pull/2440/commits/8686c928ee240edb05a7b8a2a8dae616a5228f53

dgdavid:
>     Tested manually in a local build

I'm curious to know how could you _reproduce it_. Asking because a `NODE_ENV=production npm run build` works for me without this workaround, but then it does not work in OBS

lslezak:
OK, let's revert the removal.

PR #2447, opened 2025-06-07
refactor(web): use query key factories for safer and more consistent queries

dgdavid:
> [!NOTE]
> 
> **PR stil work in progress**. Full description will be added once the PR is more polished.
>
>> Instead of writing query keys manually each time, this commit introduces
the use of query key factories. This helps prevent mistakes like typos
or mismatched keys, ensures consistency, and improves maintainability by
centralizing key definitions, making future changes easier.
>>
>> This follows a recommended pattern from the React Query community:
https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories
>

### DIFF
--- a/web/src/queries/l10n.ts
+++ b/web/src/queries/l10n.ts
@@ -25,12 +25,20 @@ import { useQueryClient, useMutation, useSuspenseQueries } from "@tanstack/react
 import { useInstallerClient } from "~/context/installer";
 import { fetchConfig, fetchKeymaps, fetchLocales, fetchTimezones, updateConfig } from "~/api/l10n";
 
+const l10nKeys = {
+  all: () => ["l10n"] as const,
+  config: () => [...l10nKeys.all(), "config"] as const,
+  locales: () => [...l10nKeys.all(), "locales"] as const,
+  keymaps: () => [...l10nKeys.all(), "keymaps"] as const,
+  timezones: () => [...l10nKeys.all(), "timezones"] as const,
+};
+
 /**
  * Returns a query for retrieving the localization configuration
  */
 const configQuery = () => {
   return {
-    queryKey: ["l10n", "config"],
+    queryKey: l10nKeys.config(),
     queryFn: fetchConfig,
   };
 };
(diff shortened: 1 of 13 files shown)


web/src/queries/l10n.ts
```diff
@@ -25,12 +25,20 @@ import { useQueryClient, useMutation, useSuspenseQueries } from "@tanstack/react
...
 import { fetchConfig, fetchKeymaps, fetchLocales, fetchTimezones, updateConfig } from "~/api/l10n";
 
+const l10nKeys = {
+  all: () => ["l10n"] as const,
+  config: () => [...l10nKeys.all(), "config"] as const,
+  locales: () => [...l10nKeys.all(), "locales"] as const,
+  keymaps: () => [...l10nKeys.all(), "keymaps"] as const,
+  timezones: () => [...l10nKeys.all(), "timezones"] as const,
+};
+
```
dgdavid, line 35: Of course, would be more readable something like 

```ts
const queryKeys = {
  all:       () => ["l10n"] as const,
  config:    () => ["l10n", "config"] as const,
  locales:   () => ["l10n", "locales"] as const,
  keymaps:   () => ["l10n", "keymaps"] as const,
  timezones: () => ["l10n", "timezones"] as const
}
```

or even, 

```ts
const queryKeys = {
  all:       ["l10n"] as const,
  config:    ["l10n", "config"] as const,
  locales:   ["l10n", "locales"] as const,
  keymaps:   ["l10n", "keymaps"] as const,
  timezones: ["l10n", "timezones"] as const
}
```
But I'd go with the proposed approach of always using functions, to stay consistent with the rest of the query keys in the codebase. Some keys take parameters, and defining them all as functions keeps the API uniform and easier to work with.

About using composed keys (`[...queryKeys.all(), "config"]`) instead of hardcoding them (`["l10n", "config"]`), it also follows TanStack Query’s recommended practices. While it may slightly impact readability, it enables a layered composition strategy that simplifies future changes, especially when modifying top-level segments or adding new levels of nesting.

dgdavid: Above comment becomes more evident when looking at more complex query keys such as those in the software, network, or storage domains. There, the benefits of consistent structure and layered composition are much easier to appreciate.

PR #2455, opened 2025-06-09
Web: Allow to use some RAIDs as if they were disks

ancorgs:
## Goal

This allows users of the web UI to install on any of the so-called “candidate” RAIDs just like they where regular disks.

That implies:

  - Offering  the option to create a new mdRaid entry that reuses one of those candidate software RAIDs (other RAIDs are not offered... yet).
  - Offering the option to remove such an mdRaid entry (stop using the RAID).
  - Move the definition of a drive into an mdRaid (pointing to a candidate RAID) and vise versa.
  - Include candidate RAIDs as possible target devices when creating/editing an VG (other RAIDs are not offered... yet).

raids

## Future steps (for upcoming pull requests)

- Allow in the UI to directly format a partitionable device (drive or RAID).
- Allow in the UI to use a whole partitionable device (drive or RAID) as PV, without partitioning it.
- Remove the current limitations, so all RAIDs (and not only the candidate ones) can be added to the configuration or used to allocate PVs.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jun 12 12:51:38 UTC 2025 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Allow selecting bootable MD RAIDS to create partitions and LVM
+  volume groups on them (gh#agama-project/agama#2455).
+
+-------------------------------------------------------------------
 Tue Jun 10 16:51:15 UTC 2025 - David Diaz <dgonzalez@suse.com>
 
 - Improve post-installation user experience by providing feedback


web/src/helpers/storage/search.ts
```diff
@@ -25,22 +25,39 @@ import { model } from "~/types/storage";
...
+  const device = buildModelDevice(apiModel, name, list);
+  if (!device || device.isUsed) return apiModel;
 
-  apiModel.drives.splice(index, 1);
+  apiModel[list].splice(index, 1);
   return apiModel;
 }
 
-export { deleteIfUnused };
+function addSearched(apiModel: apiModel.Config, data: data.SearchedDevice): apiModel.Config {
```
joseivanlopez: I am not sure about the *search* concept and the usage of `list` parameter to distinguish drives and md raids. I feel the `list` concept is exposing something internal about the model. I would prefer to have explicit hooks/helpers for drives and mds. For example:

* */helpers/storage/drive*:
  * `addDrive()`
* */helpers/storage/md-raid*:
  * `addMdRaid()`  
* */hooks/storage/drive*:
  * `useAddDrive()`
* */hooks/storage/md-raid*:
  * `useAddMdRaid()` 
  
I know that *drive* and *mdRaid* are basically the same at this moment, but they will differ at some point. I would share methods (e.g., `buildModelDevice`) to avoid repetition of common parts.

Another thing I found a bit weird about the *search* concept for hooks/helpers is that there are more searchable devices, for example partitions. And in the future we will have volume groups, logical volumes, ...

web/src/components/storage/LvmPage.tsx
```diff
@@ -57,14 +57,18 @@ import { _ } from "~/i18n";
...
+  const candidateDevices = useCandidateDevices();
   const systemDevices = useDevices("system", { suspense: true });
   const model = useModel({ suspense: true });
 
   const targetDevices = useMemo(() => {
-    const raids = (model.mdRaids || []).map((r) => systemDevices.find((d) => d.name === r.name));
-    return [...availableDevices, ...raids];
-  }, [availableDevices, systemDevices, model]);
+    const sids = candidateDevices.map((d) => d.sid);
+    const raids = (model.mdRaids || [])
```
joseivanlopez: `model.mdRaids` cannot be null.

web/src/components/storage/LvmPage.tsx
```diff
@@ -57,14 +57,18 @@ import { _ } from "~/i18n";
  * that are already there.
```
joseivanlopez: The comment is outdated, right?

web/src/components/storage/LvmPage.tsx
joseivanlopez: Are these raids the currently configured raids that are not candidates? Please, add a comment to make clear what raids we are adding.

joseivanlopez: Could not we offer all the available devices?

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -25,27 +25,24 @@ import { Split, Flex, Label } from "@patternfly/react-core";
...
-const driveLabel = (device: StorageDevice): string => deviceLabel(device, true);
+const baseName = (device: StorageDevice): string => deviceBaseName(device, true);
+const label = (device: StorageDevice): string => deviceLabel(device, true);
 
-const UseOnlyOneOption = (drive: model.Drive): boolean => {
-  const driveModel = useDrive(drive.name);
-  if (!driveModel) return false;
+const UseOnlyOneOption = (device: model.Drive | model.MdRaid): boolean => {
+  const hasPv = device.isTargetDevice;
+  if (!driveUtils.hasFilesystem(device) && (hasPv || device.isExplicitBoot)) return true;
```
joseivanlopez: `#hasFilesystem` is something we can move from the utils to the model. BTW, `hasFilesystem` accepts a `apiModel.Drive` but `device` here has other type.

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -25,27 +25,24 @@ import { Split, Flex, Label } from "@patternfly/react-core";
...
 
-  const { isExplicitBoot, hasPv } = driveModel;
-  if (!driveUtils.hasFilesystem(drive) && (hasPv || isExplicitBoot)) return true;
-
-  return driveUtils.hasReuse(drive);
+  return driveUtils.hasReuse(device);
```
joseivanlopez: The same for `#hasReuse`.

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -254,28 +250,45 @@ const DisksDrillDownMenuItem = ({
...
+      items={searchSelectorOptions(modelDevice, devices, selected, onDeviceClick)}
     >
       {text}
     </MenuButtonItem>
   );
 };
 
-type RemoveDriveOptionProps = { drive: model.Drive };
+type RemoveEntryOptionProps = {
+  device: model.Drive | model.MdRaid;
```
joseivanlopez, line 263: NP: I have seen this union (`model.Drive | model.MdRaid`) several times in this file. Maybe we can use a type alias here, something like `ModelDevice`.

web/src/helpers/storage/search.ts
ancorgs: I hope this is better now. Feel free to comment again if that's not the case.

web/src/components/storage/LvmPage.tsx
ancorgs: No. Is not. That function (and the comment) is still needed until we allow to add all kind of RAIDs (and not only the candidates) in the UI. I will write that to the comment.

web/src/components/storage/LvmPage.tsx
ancorgs: Both questions answered in the updated comment and at the description of the PR.

web/src/components/storage/LvmPage.tsx
ancorgs: Comment expanded

web/src/components/storage/LvmPage.tsx
ancorgs: Fixed. Although this is temporary code that I expect to die in less than a month.

web/src/components/storage/SearchedDeviceMenu.tsx
ancorgs: Done. `utils/drive` is getting more and more slim. :-)

web/src/components/storage/SearchedDeviceMenu.tsx
ancorgs: Done.

PR #2456, opened 2025-06-09
chore: update spec files

imobachgs:
Update the spec files, including some copyright notices and URLs.


web/package/agama-web-ui.spec
```diff
@@ -1,7 +1,7 @@
 #
 # spec file for package cockpit-machines
```
lslezak: Please update this comment as well

PR #2458, opened 2025-06-10
User Repositories: Adapt autoyast conversion

jreidinger:
Adapt autoyast conversion to new extraRepositories implemented at https://github.com/agama-project/agama/pull/2424


service/lib/agama/autoyast/software_reader.rb
```diff
@@ -37,12 +37,16 @@ def initialize(profile)
...
         software = {}
 
-        software["patterns"] = profile["software"].fetch_as_array("patterns")
-        software["packages"] = profile["software"].fetch_as_array("packages")
+        if profile["software"]
+          software["patterns"] = profile["software"].fetch_as_array("patterns")
+          software["packages"] = profile["software"].fetch_as_array("packages")
+        end
+        if profile["add-on"]
+          repos = process_repos
```
dgdavid, line 47: NP: I do not know if that logic should live here. A bit disconnected of that part currently. But I'd at least rename it to `fecth_addon_repos` to make it more readable.

service/lib/agama/autoyast/software_reader.rb
```diff
@@ -51,6 +55,21 @@ def read
...
       attr_reader :profile
+
+      def process_repos
+        repos = profile["add-on"].fetch_as_array("add_on_products") +
+          profile["add-on"].fetch_as_array("add_on_others")
+        repos.each_with_index.map do |repo, index|
+          res = {}
+          res["url"] = repo["media_url"]
+          # alias is mandatory to craft one if needed
+          res["alias"] = repo["alias"] || "autoyast_#{index}"
```
dgdavid, line 66: Question: we will sill using "autoyast" naming?

dgdavid (approved):
Implementation-wise, I don't have strong opinions. The code looks solid and appears thoroughly tested. Approving it.

service/lib/agama/autoyast/software_reader.rb
jreidinger: well, it is conversion from autoyast profile. But we can name it whatever we want...I am just not sure about better name

service/lib/agama/autoyast/software_reader.rb
jreidinger: well, idea of method is to translate autoyast entries into agama profile ones...not sure if fetch is the right verb for it

service/test/agama/autoyast/software_reader_test.rb
```diff
@@ -62,5 +76,18 @@
         expect(patterns).to eq(["vim"])
       end
     end
+
+    context "when a list of add-ons are provided" do
+      it "includes the list of repositoriess under 'software.extraRepositories'" do
```
joseivanlopez: Typo: repositoriess

joseivanlopez (approved):
LGTM

PR #2459, opened 2025-06-10
Rethinking storage selectors and other aspects of the storage UI

dgdavid:
## Problem

With RAID devices nearly ready, the current `MenuButton` used for device selection may become less usable than initially expected, especially since Agama does not yet have defined a devices amount threshold for when it should switch to a full-page selector.

Additionally, we have reports of some users having problems to understand how to use the `MenuButton` entries to define the new (or reused) partitions and to define what to do with the existing partitions.

## Solution

This PR introduces a small reorganization of the storage UI, rearranging the elements and using some labels with the hope to make the whole thing more understandable.

This also introduces a preliminary device selector implemented as a table. It can be embedded within a modal dialog to provide a simplified selection experience or rendered on its own dedicated page for the full-featured version. The modal version should include a way for users to "escape" to the full-page view.

The goal is to offer a quick and straightforward device selection interface without overwhelming users in constrained UI spaces.

Key points:

   * Provides a lightweight, modal-based device selector for simple and fast selection.
   * Includes a pathway to a full-page view for advanced use cases.
   * The full-page version supports detailed views, filtering, additional columns, and extended functionality.

This pattern should balance ease of use with power and flexibility depending on user needs and context (think on environments with tons of devices).

## Screenshots

menu

default

selector

raid

> [!CAUTION]
> 
> The following screenshot is just an explication for developers, not available in the current version

Screen Shot 2025-06-10 at 10 38 24


web/src/assets/styles/index.scss
```diff
@@ -510,3 +514,14 @@ button:focus-visible {
...
   outline-offset: 3px;
 }
+
+.agm-strong-icon {
+  font-size: 1.6rem;
+  vertical-align: middle;
+}
+
+// FIXME: make ir more generic, if possible, or even without CSS but not
+// rendering such a label if "storage instructions" are more than one
```
dgdavid, line 529:
```suggestion
// FIXME: make it more generic, if possible, or even without CSS
```

web/src/components/storage/ConfigureDeviceMenu.tsx
```diff
@@ -131,29 +121,43 @@ export default function ConfigureDeviceMenu(): React.ReactNode {
...
-      }}
-      items={[
-        <DisksDrillDownMenuItem
-          key="select-disk-option"
-          usedCount={usedDevicesCount}
+    <>
+      <MenuButton
+        menuProps={{
+          "aria-label": _("Configure device menu"),
+          popperProps: { minWidth: `min(${longestTitle * 0.75}em, 75vw)`, width: "max-content" },
```
dgdavid: Maybe this is not longer needed now that devices are open in a modal dialog. What do you think, @ancorgs ?

web/src/components/storage/DeviceSelectorModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+type DeviceSelectorProps = {
+  devices: StorageDevice[];
+  selectedDevices?: StorageDevice[];
+  onSelectionChange: SelectableDataTableProps<StorageDevice>["onSelectionChange"];
+  selectionMode?: SelectableDataTableProps<StorageDevice>["selectionMode"];
+};
+
+// TODO: document
+const DeviceSelector = ({
```
dgdavid, line 73: Apart from defining what info to display and properly documenting the component, I thought that `DeviceSelector` will likely need to live in its own file since it could be reused outside the modal, e.g., in a standalone page and/or at _its own path_.

That said, maybe it’s worth considering whether we want a single shared selector with many props to adapt it to different use cases or, instead , to have more tailored internal versions per _top-level components_. The latter might avoid making the shared one too bloated or hard to maintain.

Open to thoughts on that!

web/src/components/storage/DeviceSelectorPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+        itemIdKey="sid"
+        itemsSelected={selectedDevices}
+        onSelectionChange={onSelectionChange}
+        selectionMode={selectionMode}
+      />
+    </>
+  );
+};
+
+export default function DeviceSelectorPage(): React.ReactNode {
```
dgdavid, line 68: @ancorgs,

Although I’m still convinced of the value of this approach (providing the feature for the modal allowing navigation to a more complete selector in its own standalone page/path) , I believe it might be better to address that in a separate iteration.

web/src/components/storage/DriveEditor.tsx
```diff
@@ -21,40 +21,47 @@
...
-      <CardHeader>
-        <CardTitle>
-          <DriveHeader drive={drive} driveDevice={driveDevice} />
-        </CardTitle>
-      </CardHeader>
-      <CardBody className={spacingStyles.plLg}>
-        <Flex direction={{ default: "column" }}>
-          <SpacePolicyMenu modelDevice={drive} device={driveDevice} />
+    <ConfigEditorItem
+      header={<PartitionableHeader drive={drive} device={driveDevice} />}
```
dgdavid: This applies to all the "editors" in the storage area. I'm wondering if, now or maybe in a future iteration, it would be possible to simplify their props a bit, or at least make them a bit easier to follow.

To be honest, I’m still not fully confident I grasp the distinctions between things like `drive`, `device`, `selected`, etc., especially when jumping between components. That said, it could just be me still getting familiar with the storage business logic and maybe things are already in their best shape.

Just raising it in case there’s even a tiny bit of room for improvement 😊

web/src/components/storage/SpacePolicyMenu.tsx
```diff
@@ -70,36 +65,24 @@ export default function SpacePolicyMenu({ modelDevice, device }) {
...
-      <MenuGroup label={<SpacePolicySelectorIntro device={device} />}>
-        <MenuList>
-          <Divider />
-          {SPACE_POLICIES.map((policy) => (
-            <PolicyItem key={policy.id} policy={policy} />
-          ))}
-        </MenuList>
-      </MenuGroup>
-    </DeviceMenu>
+    <Flex gap={{ default: "gapXs" }}>
```
dgdavid, line 71: We should sit down at some point and go through all these gaps used in the editors, so we can pick one and make sure all components use the same value wherever it makes sense.

web/src/components/storage/ConfigureDeviceMenu.tsx
ancorgs: Removed.

web/src/components/storage/DeviceSelectorPage.tsx
ancorgs: Sure. This will need to wait for now.

ancorgs (approved):
Both @dgdavid and me have been working together in this pull request, reviewing each other changes in the process. So I'm approving.

PR #2461, opened 2025-06-10
feat(rust): improve user/root exported information

imobachgs:
## Problem

There are a few problems with the user and root settings that "agama config show" generates.

1. It is not possible to get the root password (it is not exposed). See [bsc#1235602](https://bugzilla.suse.com/show_bug.cgi?id=1235602).
2. The `user` section is exported even if a first user does not exist.
3. The `root` section is exported even if no authentication mechanism has been defined.
4. The `hashed_password` key is exported even if the `root` only has an SSH public key.

For instance, here's the output for a just started Agama instance:

```json
{
  "root": {},
  "user": {
    "fullName": "",
    "userName": "",
    "password": "",
    "hashedPassword": false
  }
}
```

## Solution

* Move the password settings to a single struct (`UserPassword`). The `password` and the `hashedPassword` keys are always exported together.
* Do not export the `user` or `root` sections unless it is needed.

I decided to not make the changes at HTTP level because it would imply adapting the web client too and we have more urgent things to work on.

## Testing

- *Updated the unit test*
- *Tested manually*


rust/agama-lib/src/users/store.rs
```diff
@@ -50,23 +50,44 @@ impl UsersStore {
...
+            } else {
+                Some(UserPassword {
+                    password: first_user.password,
+                    hashed_password: first_user.hashed_password,
+                })
+            };
+
+            Some(FirstUserSettings {
+                user_name: Some(first_user.user_name),
+                full_name: Some(first_user.full_name),
```
jreidinger: this means we still export even empty full_name. Is it correct?

rust/agama-lib/src/users/store.rs
```diff
@@ -85,18 +106,24 @@ impl UsersStore {
...
+            password: settings
+                .password
+                .as_ref()
+                .map(|p| p.password.clone())
+                .unwrap_or_default(),
+            hashed_password: settings
+                .password
+                .as_ref()
+                .map(|p| p.hashed_password)
+                .unwrap_or_default(),
```
jreidinger: when I see how tricky it is...maybe it belongs more to FirstUser or FirstUserSettings and having trait like Into?

rust/agama-lib/src/users/store.rs
imobachgs: Yes, the full_name is not mandatory.

rust/agama-lib/src/users/store.rs
imobachgs: OK, I can change that. At first sight, I preferred to keep this logic just here because it is not needed anywhere else and, for consistency, I would have to implement other conversions too (`From<&FirstUser> for FirstUserSettings`, `From<&RootUser> for RootUserSettings` and `From<&RootUserSettings> for RootUser`).

rust/agama-lib/src/users/store.rs
jreidinger: so in exported profile will be `full_name: ""` which looks wrong to me.

PR #2462, opened 2025-06-10
feat(web): add post-installation feedback

dgdavid:
## Problem

Currently, users receive no visual feedback after clicking either "Reboot" or "Finish" on the installation completion page. This lack of confirmation can lead to uncertainty about whether the intended action has taken effect.

Reported in: https://bugzilla.suse.com/show_bug.cgi?id=1237410 (protected link)

## Solution

Display proper feedback to provide clear confirmation that the system has either restarted or powered off. It also explicitly informs users that installer is not longer available and the page can be safely closed.

## Implementation notes

* The new path (`/installation/exit`) does not take Iguana into account, as it has been agreed that Agama will no longer support it.

* Agama assumes the system reboot has been triggered successfully without waiting for a response. Although it’s unlikely, there could be a chance the response might not be received if the system begins rebooting before the server sends it. This edge case still requires testing, but for now, assuming the reboot will take effect is considered sufficient.

* Although initially considered, the reboot action does not explicitly trigger a WebSocket disconnection. The connection will drop automatically after the maximum number of reconnection attempts. Importantly, it won’t fall back to either the `Loading` or `ServerError` screens, since the post-installation views are not under the scope of [internal `App.tsx` routing](https://github.com/agama-project/agama/blob/abd8552989bccbdaaed639c15e74df12ad04dfd8/web/src/App.tsx#L55-L90).

* Unfortunately, it’s not possible to fully clear the browser history stack to prevent users from navigating back to earlier installation steps. Users are encouraged to close the installer upon completion. If they do navigate back, they will likely encounter a `ServerError` screen.

## Testing

* Added small unit tests
* Tested manually by ~~using live image from https://build.opensuse.org/project/show/systemsmanagement:Agama:branches:improve-reboot-feedback (https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/improve-reboot-feedback/images/iso/agama-installer.x86_64-openSUSE.iso)~~ generating the `/web` production build and copying it to `/user/share/agama/web_ui` in a running installation image using the latest openSUSE-based version since the custom image was not working because of 

  > "Agama service error: D-Bus service error: org.freedesktop.DBus.Error.UnknownMethod: Method "ListUserRepositories" on interface "org.opensuse.Agama.Software1

  possibly related to the rubygem not being up-to-date

## Screenshots

localhost_8080_ (8)

---

Related to https://trello.com/c/OJhG0Ovr/ (protected link)

### DIFF
--- a/web/src/components/core/InstallationFinished.tsx
+++ b/web/src/components/core/InstallationFinished.tsx
@@ -37,15 +37,15 @@ import {
   GridItem,
   Stack,
 } from "@patternfly/react-core";
+import { Navigate, useNavigate } from "react-router-dom";
 import { Icon } from "~/components/layout";
-import { _ } from "~/i18n";
 import alignmentStyles from "@patternfly/react-styles/css/utilities/Alignment/alignment";
 import { useInstallerStatus } from "~/queries/status";
 import { useConfig } from "~/queries/storage";
 import { finishInstallation } from "~/api/manager";
 import { InstallationPhase } from "~/types/status";
-import { Navigate } from "react-router-dom";
 import { ROOT as PATHS } from "~/routes/paths";
+import { _ } from "~/i18n";
 
 const TpmHint = () => {
   const [isExpanded, setIsExpanded] = useState(false);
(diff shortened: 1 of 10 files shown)


jreidinger:
hmm, does the iguana usage still makes sense? For me it looks like we should just drop support for it to not waste time with it.

web/src/components/core/InstallationFinished.tsx
```diff
@@ -98,9 +99,54 @@ function usingTpm(config): boolean {
   return root?.encryption?.tpmFde !== undefined;
 }
 
+const Congratulations = ({ useIguana, onFinish }) => {
+  const config = useConfig();
+
+  const finish = () => {
+    finishInstallation();
+    // TODO: Ensure WebSocket is disconnected. The system is initiating a
```
jreidinger: agreed, we basically need to add to https://github.com/agama-project/agama/blob/7cc0f21635bfb510b941e1cbc998cbfcbb52b6b7/web/src/client/ws.ts#L126 method close which will call `this.client.close()` to close socket - https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close

dgdavid:
> hmm, does the iguana usage still makes sense? For me it looks like we should just drop support for it to not waste time with it.

Thanks for bringing this up. I actually had the same thought but forgot to include it in the description. Thus, I have exactly the same feeling than you.

web/src/components/core/InstallationFinished.tsx
dgdavid: As stated in the "Implementation notes" in the PR description, it has been discarded by now.

jreidinger (approved):

mvidner:
@dgdavid as suspected, the Ruby backend in the image that you tested is missing this change, so it explains the error
```
Tue Jun  3 12:22:41 UTC 2025 - Josef Reidinger <jreidinger@suse.com>

- Add support for adding extra repositories as installation sources
  (jsc#AGM-125)
```
https://build.opensuse.org/projects/systemsmanagement:Agama:branches:improve-reboot-feedback/packages/rubygem-agama-yast/files/rubygem-agama-yast.changes?expand=1

dgdavid:
> @dgdavid as suspected, the Ruby backend in the image that you tested is missing this change, so it explains the error
> 
> ```
> Tue Jun  3 12:22:41 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
> 
> - Add support for adding extra repositories as installation sources
>   (jsc#AGM-125)
> ```
> 
> https://build.opensuse.org/projects/systemsmanagement:Agama:branches:improve-reboot-feedback/packages/rubygem-agama-yast/files/rubygem-agama-yast.changes?expand=1

I see, maybe related with a similar problem I faced with outdated npm package.json, since looking at https://build.opensuse.org/package/show/systemsmanagement:Agama:branches:improve-reboot-feedback/rubygem-agama-yast it can be seen that it pulled changes from 8 days ago :/

PR #2463, opened 2025-06-11
feat(storage): AutoYaST conversion for iscsi-client

joseivanlopez:
Add conversion of the `<iscsi-client>` section of the AutoYaST profile.

### DIFF
--- a/service/lib/agama/autoyast/iscsi_reader.rb
+++ b/service/lib/agama/autoyast/iscsi_reader.rb
@@ -70,9 +70,15 @@ module Agama
           "name"    => name
         }
 
-        target["startup"] = startup if startup
+        target["startup"] = startup if valid_startup?(startup)
         target["interface"] = interface if interface
         target
+      end
+
+      # @param value [String, nil]
+      # @return [Boolean]
+      def valid_startup?(value)
+        ["onboot", "manual", "automatic"].include?(value)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


service/share/autoyast-compat.json
```diff
@@ -362,31 +361,28 @@
       },
       {
         "key": "enable",
-        "support": "yes",
         "children": [
           {
             "key": "service",
-	    "support": "yes"
+	          "support": "yes"
```
ancorgs: This (and several more below) looks strange. Some problem with spaces vs tabs?

service/lib/agama/autoyast/iscsi_reader.rb
```diff
@@ -0,0 +1,79 @@
...
+      # @return [Hash]
+      def read_target(profile_target)
+        address, port = profile_target["portal"].split(":")
+        name = profile_target["target"]
+        startup = profile_target["startup"]
+        interface = profile_target["iface"]
+
+        target = {
+          "address" => address,
+          "port"    => port.to_i,
```
ancorgs, line 69: This means a portal that omits the port will end-up with a zero. I'm not saying it is right or wrong, but at least I miss a unit test for that case. On the one hand to somehow "document" the behavior, on the other hand to make sure future changes to the code have that case into account and do not crash in case of an incomplete portal (feel free to add unit tests for other malformed AutoYaST format you can easily think of).

service/share/autoyast-compat.json
joseivanlopez: I don't know.  I open the file and it looks good.

ancorgs (approved):
LGTM

PR #2464, opened 2025-06-11
fix(rust): consider hashedPassword as "false" if not given

imobachgs:
## Problem

When deserializing a password, Agama expected both the `password` and the `hashedPassword` attributes to be set.

Found by @mvidner during [this review](https://github.com/agama-project/agama-project.github.io/pull/81/files#r2139859193). Thanks!

## Solution

Consider `hashedPassword` is `false` if not given.

Addionally, avoid writing the profile to the logs.

## Testing

- The first change was manually tested.

### DIFF
--- a/rust/agama-lib/src/profile.rs
+++ b/rust/agama-lib/src/profile.rs
@@ -184,7 +184,7 @@ impl ProfileEvaluator {
     pub fn evaluate_string(&self, profile: &str) -> anyhow::Result<String> {
         let dir = tempdir()?;
         let working_path = dir.path().join("profile.jsonnet");
-        fs::write(working_path, dbg!(profile))?;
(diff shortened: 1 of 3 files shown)


mvidner:
> When deserializing a password, Agama expected both the password and the hashedPassword attributes to be set.

The non-obvious part is the interaction with its containing struct, which includes it like this:
```rs
    #[serde(flatten)]
    pub password: Option<UserPassword>,
```

Without the `default` on the `bool`, I **guess** that the whole `Option<UserPassword>` ended up as None.

mvidner (approved):
The unit test is not convincing me :)
but I have  tested that `agama config generate -` now keeps `password` even if `hashedPassword` is missing.

PR #2468, opened 2025-06-12
fix(i18n): Configure the console font

lslezak:
## Problem

- The some non-ASCII characters are not displayed properly, a filled square is displayed instead
- E.g. when selecting the Czech locale and keyboard some Czech characters are not displayed properly when typing.

agama-console-font-broken

## Solution

- Read the appropriate console font for the selected locale from the langtable data.
- Unfortunately the `systemd-firstboot` tool does not support configuring the font so we need to write it directly into the `/etc/vconsole.conf` configuration file.
- Set the `eurlatgr` console font also in the Live ISO installer.

## Testing

- Tested manually, now all Czech characters are displayed properly.

agama-console-font-fixed


rust/agama-server/src/l10n/model.rs
```diff
@@ -263,6 +265,23 @@ impl L10n {
         let output = cmd.output()?;
         tracing::info!("{:?}", &output);
 
+        // unfortunately the console font cannot be set via the "systemd-firstboot" tool,
+        // we need to write it directly to the config file
+        let locale = self.locales_db.entries().iter().find(|l| l.id == locale);
```
imobachgs: Perhaps we could add a method to find a locale to `LocalesDatabase`.

rust/agama-server/src/l10n/model/locale.rs
```diff
@@ -110,7 +112,20 @@ impl LocalesDatabase {
                 id: code.clone(),
                 language: language_label,
                 territory: territory_label,
+                consolefont: language
+                    .consolefonts
+                    .consolefont
+                    .first()
+                    .unwrap_or(&RankedConsoleFont {
```
imobachgs: You are using an artificial value (with id `""`) to model the fact that it does not have a console font. IMHO, you should leverage the `Option<T>` type here.

imobachgs (changes requested):
Please, do not rely on `is_empty()` in this case. `Option<T>` is a much better alternative.

rust/agama-server/src/l10n/model/locale.rs
```diff
@@ -106,11 +113,20 @@ impl LocalesDatabase {
                 .or_else(|| names.name_for(DEFAULT_LANG))
                 .unwrap_or(territory.id.to_string());
 
+            let consolefont = match language.consolefonts.consolefont.first() {
```
imobachgs: In those cases where you are not going to do anything with the `None`, I would use a `map`:

```rust
let consolefont = language
    .consolefonts
    .consolefont
    .first()
    .map(|f| f.id.clone());
```

It looks longer just because of the format.

PS: Out of the scope of this PR, we might consider in the future to implement the iterator for the `RankedConsoleFonts.

rust/agama-server/src/l10n/model/locale.rs
```diff
@@ -79,6 +79,11 @@ impl LocalesDatabase {
         &self.locales
     }
 
+    /// Find the locale in the database
```
imobachgs, line 82: Please, document the `locale` param. I know, it is obvious, but still...

imobachgs (approved):

PR #2469, opened 2025-06-12
Rmt registration

jreidinger:
## Problem

The RMT can be configured to not require registration key. So sending empty one should be valid option, but it is not possible with current agama.

## Solution

Change registration API to distinguish between registered and registration key, that can be empty ( used in past as indication of not registered system ).

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/service/lib/agama/dbus/software/product.rb
+++ b/service/lib/agama/dbus/software/product.rb
@@ -88,7 +88,7 @@ module Agama
         def select_product(id)
           if backend.product&.id == id
             [1, "Product is already selected"]
-          elsif backend.registration.reg_code
+          elsif backend.registration.registered
             [2, "Current product must be deregistered first"]
           else
             backend.select_product(id)
@@ -126,6 +126,10 @@ module Agama
 
             [[code, description]]
           end
+        end
+
+        def registered
+          !!backend.registration.registered
         end
 
         def reg_code
@@ -207,13 +211,11 @@ module Agama
         def register(reg_code, email: nil)
           if !backend.product
             [1, "Product not selected yet"]
-          elsif backend.registration.reg_code
-            # report success and do nothing when already registered with the same code
-            if backend.registration.reg_code == reg_code
-              [0, ""]
-            else
-              [2, "Product already registered"]
-            end
+          # report success and do nothing when already registered with the same code
+          elsif backend.registration.reg_code == reg_code
+            [0, ""]
+          elsif backend.registration.registered
+            [2, "Product already registered"]
           elsif !backend.product.registration
             [3, "Product does not require registration"]
           else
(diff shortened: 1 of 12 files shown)


service/lib/agama/dbus/software/product.rb
```diff
@@ -88,7 +88,7 @@ def selected_product
         def select_product(id)
           if backend.product&.id == id
             [1, "Product is already selected"]
-          elsif backend.registration.reg_code
+          elsif backend.registration.registered
```
imobachgs, line 91: 👍🏻

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -200,9 +193,7 @@ const Extensions = () => {
 
 export default function ProductRegistrationPage() {
   const { selectedProduct: product } = useProduct();
-  const { key } = useRegistration();
-  // FIXME: this needs to be fixed for RMT which allows registering with empty key
-  const isRegistered = !isEmpty(key);
+  const { registered } = useRegistration();
```
imobachgs, line 196: You could do:

```javascript
const { registered: isRegistered } = useRegistration();
```

imobachgs (commented):
It looks good. I just miss the changes entries.

web/package/agama-web-ui.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+
+Fri Jun 13 08:24:59 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
+
+- Allow empty registratio key to support RMT that can be configured
```
imobachgs: What about: "Allow registering against an RMT without a registration code."

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jun 13 08:24:39 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
+
+- Split registration key and registered status to allow RMT with
```
```suggestion
- Split registration key and registered status to support RMT with
```

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jun 13 08:23:46 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
+
+- Split registration key and registered status to allow RMT with
```
```suggestion
- Split registration key and registered status to support RMT with
```

web/src/components/core/ChangeProductOption.tsx
```diff
@@ -37,7 +37,7 @@ export default function ChangeProductOption({ children, ...props }: Omit<Dropdow
   const to = useHref(PATHS.changeProduct);
 
   if (products.length <= 1) return null;
-  if (registration.registered) return null;
+  if (registration && registration.registered) return null;
```
imobachgs:
```suggestion
  if (registration?.registered) return null;
```

imobachgs (approved):

PR #2470, opened 2025-06-12
fix(live): use localectl instead of setxkbmap

imobachgs:
## Problem

- `setxkbmap` is not available since the switch to Wayland.
- It caused the language/keymap change to fail.

## Solution

- Use `localectl` instead of `setxkbmap`.
- Drop the old code.

## Testing

- Tested manually


rust/agama-server/src/l10n/error.rs
```diff
@@ -32,6 +32,6 @@ pub enum LocaleError {
     UnknownKeymap(KeymapId),
     #[error("Invalid keymap: {0}")]
     InvalidKeymap(#[from] InvalidKeymap),
-    #[error("Could not apply the changes")]
+    #[error("Could not apply l10n the changes: {0}")]
```
dgdavid: :ok_hand:

rust/agama-server/src/l10n/error.rs
```diff
@@ -32,6 +32,6 @@ pub enum LocaleError {
     UnknownKeymap(KeymapId),
     #[error("Invalid keymap: {0}")]
     InvalidKeymap(#[from] InvalidKeymap),
-    #[error("Could not apply the changes")]
+    #[error("Could not apply the l10n settings: {0}")]
```
jreidinger, line 35: :+1:

rust/agama-server/src/l10n/model.rs
```diff
@@ -266,21 +173,20 @@ impl L10n {
...
-        let output = output.map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?;
-        let output = output.unwrap_or(String::new());
+    fn ui_keymap() -> Result<KeymapId, LocaleError> {
+        let output = Command::new("/usr/bin/localectl")
+            .output()
+            .map_err(LocaleError::Commit)?;
+        let output = String::from_utf8_lossy(&output.stdout);
 
-        let keymap_regexp = Regex::new(r"(?m)^layout: (.+)$").unwrap();
+        let keymap_regexp = Regex::new(r"(?m)VC Kayout: (.+)$").unwrap();
```
jreidinger: that regexp looks strange...on my system I have `VC Keymap: us` and on my testing VM it is identical. So unless there is some change of output in wayland ( which I do not test), I would say that this regexp is broken.

imobachgs: Yes, it looks broken to me.

imobachgs: I have fixed the problem and I did the proper test 🫤 (change the locale, restart the server and check the configuration).

PR #2472, opened 2025-06-13
Add support to bind a connection to an specific mac-address (bsc#1244259)

teclator:
## Problem

In an Agama profile it is possible to bind a connection to an specific device using its interface name but it is not possible to use the mac address, that option is only supported by the kernel cmdline or using nmcli directly. The profile allows to set a mac address but that is the assign-mac-address instead of the mac-address attribute.

At least the Agama profile should support it.

In the UI there is no option to bind a connection to an specific device and this specially problematic with the default.

- https://bugzilla.suse.com/show_bug.cgi?id=1244259
- https://trello.com/c/yt0Xd9ah/1663-1244259-unable-to-bind-a-specific-mac-address-to-a-connection?search_id=a1d25067-f7fc-40ec-b829-0d278f9836b0

## Solution

Modify the schema renaming the mac address assignation attribute to customMacAddress and using the macAddress to bind the connection to an specific mac.

```json
{
  "network": {
    "connections": [
      {
        "id": "eth0",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "interface": "eth0",
        "macAddress": "A0:CE:C8:F2:F7:CA",
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  }
}
```

There was not possible to delete optional attributes from NetworkManager like the **interface-name** or the **macAddress** using Optional values because if set already the merge will use the one which is present, that has been fixed for these two attributes allowing to switch between them although the solution should be extended to other cases.

Example (should unset the **mac-address** as it is not declared in the first example and the **interface-name** in the second):

```json
{
  "network": {
    "connections": [
      {
        "id": "eth0",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "interface": "eth0",
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  }
}

{
  "network": {
    "connections": [
      {
        "id": "eth0",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "macAddress": "A0:CE:C8:F2:F7:CA",
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  }
}
```

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-network/src/nm/dbus.rs
+++ b/rust/agama-network/src/nm/dbus.rs
@@ -261,12 +261,25 @@ pub fn merge_dbus_connections<'a>(
     original: &'a OwnedNestedHash,
     updated: &'a NestedHash,
 ) -> Result<NestedHash<'a>, zvariant::Error> {
+    // FIXME: it should contain all the sections and attributes which can be absent and known by
+    // agama in order to allow the user to remove them otherwise the original value will be kept
+    let handled_by_agama: Vec<&str> = vec!["interface-name", "mac-address"];
+
     let mut merged = HashMap::with_capacity(original.len());
     for (key, orig_section) in original {
         let mut inner: HashMap<&str, zbus::zvariant::Value> =
             HashMap::with_capacity(orig_section.len());
+
         for (inner_key, value) in orig_section {
-            inner.insert(inner_key.as_str(), value.try_into()?);
+            if handled_by_agama.contains(&inner_key.as_str()) {
+                tracing::info!(
+                    "Do not insert '{}' from the original section '{}' as it is handled by agama",
+                    &inner_key,
+                    &key
+                );
+            } else {
+                inner.insert(inner_key.as_str(), value.try_into()?);
+            }
         }
         if let Some(upd_section) = updated.get(key.as_str()) {
             for (inner_key, value) in upd_section {
@@ -314,6 +327,10 @@ pub fn cleanup_dbus_connection(conn: &mut NestedHash) {
     if let Some(connection) = conn.get_mut("connection") {
         if connection.get("interface-name").is_some_and(is_empty_value) {
             connection.remove("interface-name");
+        }
+
+        if connection.get("mac-address").is_some_and(is_empty_value) {
+            connection.remove("mac-address");
         }
 
         if connection.get("master").is_some_and(is_empty_value) {
@@ -842,15 +859,15 @@ fn mac_address6_from_dbus(
             .map(|u| u.downcast_ref::<u8>())
             .collect::<Result<Vec<u8>, _>>()?;
 
-        // FIXME: properly handle the failing case
-        Ok(Some(MacAddr6::new(
-            *mac.first().unwrap(),
-            *mac.get(1).unwrap(),
-            *mac.get(2).unwrap(),
-            *mac.get(3).unwrap(),
-            *mac.get(4).unwrap(),
-            *mac.get(5).unwrap(),
-        )))
+        if mac.len() != 6 {
+            return Err(NmError::InvalidDBUSValue("mac-address".to_string()));
+        }
+
+        let [a, b, c, d, e, f] = mac[0..6] else {
+            return Err(NmError::InvalidDBUSValue("mac-address".to_string()));
+        };
+
+        Ok(Some(MacAddr6::new(a, b, c, d, e, f)))
     } else {
         Ok(None)
     }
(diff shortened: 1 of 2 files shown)


rust/agama-lib/share/profile.schema.json
```diff
@@ -302,11 +302,15 @@
                 "title": "The name of the network interface bound to this connection",
                 "type": "string"
               },
-              "macAddress": {
+              "CustomMacAddress": {
```
cfconrad: maybe we should write the first letter lower? `customMacAddress`

teclator: yep, it should be customMacAddress, human error ;)

rust/agama-lib/share/profile.schema.json
```diff
@@ -302,11 +302,15 @@
                 "title": "The name of the network interface bound to this connection",
                 "type": "string"
               },
-              "macAddress": {
-                "title": "Custom MAC address",
+              "customMacAddress": {
+                "title": "Custom MAC address to be used",
```
```suggestion
                "title": "Custom MAC address to use",
```

rust/agama-network/src/nm/dbus.rs
```diff
@@ -822,7 +833,32 @@ fn base_connection_from_dbus(conn: &OwnedNestedHash) -> Result<Connection, NmErr
...
+fn mac_address6_from_dbus(
+    config: &HashMap<String, OwnedValue>,
+) -> Result<Option<MacAddr6>, NmError> {
+    if let Some(mac) = get_optional_property::<zvariant::Array>(config, "mac-address")? {
+        let mac: Vec<u8> = mac
+            .iter()
+            .map(|u| u.downcast_ref::<u8>())
+            .collect::<Result<Vec<u8>, _>>()?;
+
+        // FIXME: properly handle the failing case
```
imobachgs: We should avoid adding more FIXMEs :-)

imobachgs (commented):
In general it looks good. It would be better if we avoid the `FIXME`.

rust/agama-network/src/nm/dbus.rs
teclator: definitely, inherited from the BSSID handling, but we need to improve that situation and handle better possible errors

rust/agama-lib/share/profile.schema.json
```diff
@@ -347,7 +351,7 @@
                   "type": "string"
                 }
               },
-              "dns_searchlist": {
+              "dnsSearchlist": {
```
jreidinger, line 354: I am not sure if openQA testing it, but we should notify them about those backward incompatible changes.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -125,9 +130,13 @@ pub fn connection_to_dbus<'a>(
                     ("mtu", Value::new(conn.mtu)),
                     (
                         "assigned-mac-address",
-                        Value::new(conn.mac_address.to_string()),
+                        Value::new(conn.custom_mac_address.to_string()),
                     ),
                 ]));
+
+                if let Some(mac) = conn.mac_address {
+                    wireless_dbus_key.insert("mac-address", Value::new(mac.as_bytes()));
```
jreidinger, line 138: NP: the name `wireless_dbus_key` is a bit confusing as I kind of expect something else.

rust/agama-network/src/nm/dbus.rs
jreidinger: well, proper error handling is not hard. Just use slicing with else syntax. See playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0832077953ca0104efc718ddb6f7a3f8

jreidinger (approved):

rust/agama-network/src/nm/dbus.rs
```diff
@@ -261,12 +261,25 @@ pub fn merge_dbus_connections<'a>(
     original: &'a OwnedNestedHash,
     updated: &'a NestedHash,
 ) -> Result<NestedHash<'a>, zvariant::Error> {
+    // FIXME: it should contain all the sections and attributes which can be absent and known by
+    // agama in order to allow the user to remove them otherwise the original value will be kept
```
jreidinger, line 265: is it really intentional to allow user remove attributes via agama that are not managed by agama? I would just keep it outside of agama as long as agama does not support it.

teclator: No, that is not possible. That is why we are getting connection values from DBUS in ordert to not lost the values Agama does not know about. 

But the current problem are optional values which are not present in the updated connection but are in the original connection (obtained from DBUS). For example, changing from **interface-name** to **macAddress** was not allowed because both values are optional in the connection and once written to NM there will be obtained when retrieved.

jreidinger (approved):

rust/agama-network/src/nm/dbus.rs
teclator: if you have some suggestion just share it ;) by now as it was already there I will keep as it is

rust/agama-network/src/nm/dbus.rs
teclator: Added example to the desription

PR #2473, opened 2025-06-13
fix(storage): add missing iSCSI steps

joseivanlopez:
Installing on iSCSI was not working because some iSCSI steps were missing:

* Start iSCSI services in the live media.
* Copy iSCSI configs to the target system.
* Enable iSCSI services in the target system.   

https://bugzilla.suse.com/show_bug.cgi?id=1243973

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jun 19 10:21:13 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Add missing steps for iSCSI installation
+  (gh#agama-project/agama#2473).
+
+-------------------------------------------------------------------
 Wed Jun 18 12:28:14 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
 - Disable local installation repositories instead of removing them


service/lib/agama/storage/finisher.rb
```diff
@@ -332,13 +411,20 @@ def run
...
+        include Yast::I18n
+
         IGUANA_PATH = "/iguana"
         private_constant :IGUANA_PATH
         IGUANA_MOUNTLIST = File.join(IGUANA_PATH, "mountlist").freeze
         private_constant :IGUANA_MOUNTLIST
 
+        def initialize(logger)
+          super
+          textdomain "agama"
```
jreidinger: do we need to have textdomain in each chield? Won't be enough to havve it in Step?

joseivanlopez: I don't know. Would it work if the base class `Step` is moved to a separate file?

service/lib/agama/storage/iscsi/adapter.rb
```diff
@@ -137,6 +139,13 @@ def update_node(node, startup:)
 
       private
 
+        # Starts the iSCSI services.
+        def start_services
+          Yast::Service.start("iscsi")
+          Yast::Service.start("iscsid")
+          Yast::Service.start("iscsiuio")
```
jreidinger, line 146: I would like to have here comment *why* we need all three services

joseivanlopez: I don't know why. We are replicating what YaST does. The three services are always available in the live media.

service/lib/agama/storage/finisher.rb
jreidinger: well, from ruby POV it will...from our script POV probably not...but maybe @lslezak we can improve our script for collecting it to always just use agama textdomain for files in agama git?

joseivanlopez: Anyway, in this case using textdomain in the base class is enough for the script. So let's do it.

PR #2474, opened 2025-06-13
Initial support for sorting storage devices during the search

ancorgs:
## Problem

The results of the `search` section in the description of `drives`, `partitions` and `mdRaids` are always sorted by name. Making it very hard to specify something like "the two biggest disks".

## Solution

This implements support for implementing such searches with the syntax shown in the following examples:

```json
{
  "search": {
    "sort": { "size": "desc" },
    "max": 2
  }
}
```

Several criteria can be specified to resolve ties.

```json
{
  "search": {
    "sort": [
        { "size": "desc" },
        "name"
    ],
    "max": 2
  }
}
```

And, of course, all that can be combined with a condition to filter out devices before sorting.

```json
{
  "search": {
    "condition": { "size": { "less": "1TiB" } },
    "sort": { "size": "desc" },
    "max": 2
  }
}
```

## Testing

- Added unit tests for the new parts
- Tested manually

### DIFF
--- a/service/lib/agama/storage/configs/sort_criteria.rb
+++ b/service/lib/agama/storage/configs/sort_criteria.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2025] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Agama
+  module Storage
+    module Configs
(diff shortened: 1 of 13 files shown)


service/lib/agama/storage/configs/sort_criteria.rb
```diff
@@ -0,0 +1,88 @@
...
+          # Constructor
+          def initialize
+            @asc = true
+          end
+
+          # @return [Boolean] Whether sorting must be in ascending order
+          attr_accessor :asc
+          alias_method :asc?, :asc
+
+          # @return [Boolean] Whether sorting must be in descending order
```
joseivanlopez, line 38: NP: I would simplify the API by removing `#desc?` and `#desc=`, or at least `#desc=`. This could be misleading:

~~~
criterion = Size.new
criterion.asc = true
criterion.desc = true
~~~

service/test/agama/storage/config_solvers/drives_search_test.rb
```diff
@@ -358,5 +398,90 @@
         expect(p3.search.device.name).to eq("/dev/vda3")
       end
     end
+
+    context "if a drive config sorts the search" do
```
joseivanlopez, line 402: Do we need similar tests for MD?

service/lib/agama/storage/configs/sort_criteria.rb
ancorgs: I removed the setter. I still find the getter to be useful and not really confusing.

service/test/agama/storage/config_solvers/drives_search_test.rb
ancorgs: Done.

PR #2479, opened 2025-06-13
fix: make sure queries are sync

imobachgs:
## Problem

Here is a proposal to solve [bsc#1243276](https://bugzilla.suse.com/show_bug.cgi?id=1243276) and other synchronization problems. It needs to be tested more extensively.

We have detected that the web UI can get into the following scenario:

1. The web UI tries to connect to the WebSocket.
2. The web UI asks for the selected product. It caches the response.
3. The product gets selected, for instance, loading a profile.
4. The server emits the signal `ProductSelected`.
5. **The WebSocket connection is finally established.**

As you can see, there is a chance that Agama misses the signal (step 4). If the WebSocket connections happens before (which is the normal case), everything should work.

## Solution

* Do not mount the `App` component until the `WebSocket` is ready.
* If the connection is lost, `App` is umounted and it has the chance to clean-up the queries.

## Tasks

- [x] Fix the tests.
- [x] Build an ISO to test the change. Testing images are [here](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/sync-queries/images/iso).

### DIFF
--- a/web/jest.config.js
+++ b/web/jest.config.js
@@ -23,7 +23,7 @@ module.exports = {
   collectCoverage: true,
 
   // An array of glob patterns indicating a set of files for which coverage information should be collected
-  collectCoverageFrom: ["src/**/*.{js,jsx}", "!src/lib/*.js"],
+  collectCoverageFrom: ["src/**/*.{js,jsx,ts,tsx}"],
 
   // The directory where Jest should output its coverage files
   coverageDirectory: "coverage",
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Tue Jun 17 12:22:46 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Make sure queries data is in sync with the WebSocket messages
+  (bsc#1243276, gh#agama-project/agama#2479).
+- Properly jump to the progress page when the product is selected
+  by a third party (e.g., "agama config load").
+- Log WebSocket messages in the console with "debug" log level.
+
+-------------------------------------------------------------------
 
 Fri Jun 13 08:24:59 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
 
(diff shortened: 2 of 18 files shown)


web/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   collectCoverage: true,
 
   // An array of glob patterns indicating a set of files for which coverage information should be collected
-  collectCoverageFrom: ["src/**/*.{js,jsx}", "!src/lib/*.js"],
+  collectCoverageFrom: ["src/**/*.{js,jsx,ts,tsx}"],
```
dgdavid, line 26: :+1:

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Tue Jun 17 12:22:46 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Make sure queries data is in sync with the WebSocket messages
+  (bsc#1243276, gh#agama-project/agama#2479).
+- Properly jump to the progress page when the product is selected
+  by a third party (e.g., "agama config load").
+- Log WebSocket messages in the console with "debug" log level.
```
dgdavid, line 8: :ok_hand:

web/src/App.tsx
```diff
@@ -20,40 +20,51 @@
...
+  const queryClient = useQueryClient();
+
+  useEffect(() => {
+    // Invalidate the queries when unmounting this component.
+    return () => {
+      queryClient.invalidateQueries();
+    };
+  }, [queryClient]);
 
+  console.log("App component", {
```
dgdavid, line 59: I understand this is _an alternative to_ the deleted "Loading screen: Initialization" and that at some point in the future we're going to drop all the logging from this component.

dgdavid (approved):

web/src/App.tsx
imobachgs: Yes. We just want some additional logging in order to track this problem (if it still persists).

Issue #2395, opened 2025-05-23
L10n: A different font for non-Latin languages


FAlexei:
The default font apparently only supports Latin characters. As a result, when a non-Latin language (Russian, Ukrainian, etc.) is selected, most characters are displayed in a different font, causing punctuation marks and Latin letters to stand out from the rest of the sentence.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/fd1fcfe0-65be-47bc-8e1b-4e717c6174eb" />

dgdavid:
> Maybe a bug report for https://github.com/SUSE/suse-font/

Done, see https://github.com/SUSE/suse-font/issues/11

PR #2488, opened 2025-06-17
fix(web) improve font support for Chinese and Georgian languages

dgdavid:
## Problem

The installation ISO’s embedded browser failed to render some languages correctly (e.g., Chinese, Georgian). This issue did not appear when using a remote browser, which highlighted inconsistencies in font availability.

Inspection using Firefox’s [font tools](https://firefox-source-docs.mozilla.org/devtools-user/page_inspector/how_to/edit_fonts/index.html) revealed the browser was falling back to [DejaVu fonts](https://dejavu-fonts.github.io/) in the remote system due to missing or misconfigured `@font-face` rules. In particular, 

- Chinese was not being served due to a copy-paste error in its declaration.
- Georgian fonts were missing entirely.

## Solution

- Add proper `@font-face` declarations for Georgian using [Noto Sans Georgian](https://fonts.google.com/noto/specimen/Noto+Sans+Georgian)
- Fix the font-family name in the Chinese font-face rule.
- Ensure Noto Sans is used as a first fallback for the SUSE font.

Additionally, a configuration file has been added to the live ISO to enable the use of fonts located at `/usr/share/agama/web_ui/fonts`. This change addresses cases where native HTML elements, such as `<option>`, were rendering using system fonts instead of those served by Agama.

Together, these updates improve multilingual font rendering and ensure a consistent visual appearance, even on systems that lack certain fonts. While it's unlikely, users accessing the interface via a remote browser from a system without these fonts might encounter minor rendering issues. However, in such cases, the fonts relevant to their language should still be available, minimizing the impact.

## Tests

Manually tested using the [generated ISO](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/noto-sans-as-fallback/images/iso/) (note: the link may become unavailable in the future).

## Screenshots

| Before | After |
|-|-|
| fonts_broken | fonts_fixed |
| fonts_broken_2 | fonts_fixed_2 |

---

Related to 

  * https://github.com/agama-project/agama/pull/2467
  * https://github.com/agama-project/agama/issues/2395
  * https://trello.com/c/NpSkfUAD (protected link)
  * https://bugzilla.suse.com/show_bug.cgi?id=1236261 (protected link)


web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Jun 18 08:42:49 UTC 2025 - David Diaz <dgonzalez@suse.com>
+
+- Improve multilingual font support by fixing the font-face rule for
+  Chinese, adding Noto Sans Georgian, and making bundled fonts
+  available for the system via a new config in the live ISO
```
lslezak: This change could be also mentioned in the [live/src/agama-installer.changes](https://github.com/agama-project/agama/blob/master/live/src/agama-installer.changes) file.

dgdavid: Right. I will adapt it

lslezak (approved):

PR #2489, opened 2025-06-18
fix: preserve AutoYaST pre-scripts

imobachgs:
## Problem

When using an AutoYaST profile, the pre-scripts are executed as part of the initial profile validation (`agama config generate`). The problem is that those scripts, and its logs, are missing from the `/run/agama/scripts` directory and hence not copied to the final system.

## Solution

The fix consists on copying those files to `/run/agama/scripts/autoyast`. We are not using the regular `/run/agama/scripts/pre` because they are executed at a different point.

## Testing

- [x] Added a new unit test
- [x] Tested manually

### DIFF
--- a/rust/agama-lib/src/scripts/model.rs
+++ b/rust/agama-lib/src/scripts/model.rs
@@ -25,6 +25,7 @@ use std::{
 };
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


teclator (approved):
LGTM

PR #2491, opened 2025-06-18
Disable local installation repositories (bsc#1243064)

lslezak:
## Problem

- When using a local installation repository (e.g. Full DVD) then this repository is removed from the system
- It should behave like in SLE15 which keeps the installation repository, just disables it. After enabling it manually it can be used for installing additional packages.

## Solution

- Disable local repositories instead of removing them

## Notes

- Because we keep the Full installation repository in the system it should have some meaningful name and alias. And because the packages are fixed the autorefresh should be disabled.
- This required extending the API to allow specifying required data.

## DVD

- For DVDs during installation the device name is included in the URL to ensure the correct repository is used. (In theory there might be multiple DVD drives in the system and the other drive might contain a different version of the product or a completely different product. This ensures libzypp uses the correct drive.
- After installation the device is removed from URL, this allows inserting the installation medium into a different drive without a failure. Libzypp verifies that the correct medium is inserted by checking the `/media.1/media` content with the file saved in the  cache.

## Hard disk or USB flash disk

- For these media it uses the `/dev/disk/by-id` device name. That should be more specific and should avoid problems with using incorrect media. It is important  for USB flash disk for which the device name might change in the installed system. (The installation flash disk might become `/dev/sdc` instead of `/dev/sdb` if you attach some different USB flash disk before it.

## Testing

- Updated unit test
- Tested manually

## Screenshots

- In all cases after enabling the repository manually it was possible to install additional package from the original installation medium.

### Installation from DVD

- When installing from the DVD the repository URL is modified
- Notice the descriptive repository name and alias

agama-installed-dvd-repos

### Installation from second hard disk

- The installation ISO was dumped to a second disk and the system was booted from it (actually I converted the ISO file to QCOW image format using `qemu-img convert` command and used the QCOW image for the second disk).
- The system was installed on the first disk.

agama-installed-sata-repos

### Installation from USB flash disk

- Because VirtualBox cannot boot from USB I used Qemu for testing installation from USB
- To use a virtual flash disk use these additional Qemu parameters: `-drive if=none,id=stick,format=raw,file=full.iso -device nec-usb-xhci,id=xhci -device usb-storage,bus=xhci.0,drive=stick,removable=on -boot menu=on`, at boot press `Esc` and select the second option (the USB disk) in the boot menu.

agama-installed-usb-repos

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -21,6 +21,7 @@
 
 require "fileutils"
 require "json"
+require "shellwords"
 require "yast"
 require "y2packager/product"
 require "y2packager/resolvable"
@@ -217,8 +218,8 @@ module Agama
 
       # Writes the repositories information to the installed system
       def finish
-        # remove the dir:///run/initramfs/live/install repository and similar
-        remove_local_repos
+        # disable local repositories (DVD, USB flash...)
+        disable_local_repos
         Yast::Pkg.SourceSaveAll
         Yast::Pkg.TargetFinish
         # copy the libzypp caches to the target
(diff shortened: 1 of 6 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -589,6 +598,30 @@ def installation_device(label)
...
+      # @param path [String] Local path where the Full repository is mounted
+      # @param url_path [String] Path part of the resulting URL
+      # @return [String,nil] URL or `nil` if the Full repository device was not found
+      def full_repo_url(path, url_path)
+        # find the device which is mounted at the repository location
+        live_device = `findmnt -n -o SOURCE --target #{Shellwords.escape(path)}`.chomp
+        logger.info "Installation device: #{live_device}"
+        return nil unless live_device
+
+        # distinguish between DVD and hard disks/USB flash
```
imobachgs, line 611: I guess this regexp is more than enough. I checked whether `findmnt` offered the same options than `lsblk` (e.g., `TRAN`) but it does not.

service/lib/agama/software/repository.rb
```diff
@@ -31,11 +31,34 @@ module Software
     #
     # @see RepositoriesManager
     class Repository < Y2Packager::Repository
+      Yast.import "Pkg"
```
imobachgs: Why is this inside the class definition?

PR #2493, opened 2025-06-19
Only required

jreidinger:
## Problem

Agama is missing option to specify that only mandatory dependencies should be installed.

- trello: https://trello.com/c/czY7e3c0/4898-3-agm-154-rc2-gehc-would-like-to-choose-not-to-install-recommended-packages-auto
- jira: https://jira.suse.com/browse/AGM-154

## Solution

Implement support for unattended installation to specify if only required dependencies should be used.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`

### DIFF
--- a/doc/dbus/org.opensuse.Agama.Software1.doc.xml
+++ b/doc/dbus/org.opensuse.Agama.Software1.doc.xml
@@ -126,5 +126,12 @@
     </method>
     <property type="a{sy}" name="SelectedPatterns" access="read"/>
     <property type="a(ussa(uss))" name="Conflicts" access="read"/>
+    <!--
+      flag for only required packages
+      0 not set
+      1 false
+      2 true
+    -->
+    <property type="u" name="OnlyRequired" access="readwrite"/>
   </interface>
 </node>
--- a/rust/agama-lib/src/software/settings.rs
+++ b/rust/agama-lib/src/software/settings.rs
@@ -37,11 +37,18 @@ pub struct SoftwareSettings {
     /// List of user specified repositories to use on top of default ones.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub extra_repositories: Option<Vec<RepositoryParams>>,
+    /// List of user specified repositories to use on top of default ones.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub only_required: Option<bool>,
 }
 
 impl SoftwareSettings {
     pub fn to_option(self) -> Option<Self> {
-        if self.patterns.is_none() && self.packages.is_none() && self.extra_repositories.is_none() {
+        if self.patterns.is_none()
+            && self.packages.is_none()
+            && self.extra_repositories.is_none()
+            && self.only_required.is_none()
+        {
             None
         } else {
             Some(self)
--- a/service/lib/agama/dbus/software/manager.rb
+++ b/service/lib/agama/dbus/software/manager.rb
@@ -70,6 +70,11 @@ module Agama
         private_constant :SOFTWARE_INTERFACE
 
         dbus_interface SOFTWARE_INTERFACE do
+          # Flag for proposing required only dependencies
+          # Do not forget to call Propose method after changing flag
+          # value mapping 0 for not set, 1 for false and 2 for true
+          dbus_accessor :only_required, "u"
+
           # array of repository properties: pkg-bindings ID, alias, name, URL, product dir, enabled
           # and loaded flag
           dbus_method :ListRepositories, "out Result:a(issssbb)" do
@@ -182,6 +187,30 @@ module Agama
           busy_while do
             backend.locale = locale
           end
+        end
+
+        def only_required
+          case backend.proposal.only_required
+          when nil then 0
+          when false then 1
+          when true then 2
+          else
+            @logger.warn(
+              "Unexpected value in only_required #{backend.proposal.only_required.inspect}"
+            )
+            0
+          end
+        end
+
+        def only_required=(flag)
+          value = case flag
+          when 0 then nil
+          when 1 then false
+          when 2 then true
+          else
+            @logger.warn "Unexpected value in only_required #{flag.inspect}"
+          end
+          backend.proposal.only_required = value
         end
 
         def ssl_fingerprints
--- a/service/lib/agama/software/proposal.rb
+++ b/service/lib/agama/software/proposal.rb
@@ -65,6 +65,10 @@ module Agama
       # @return [Array<Hash<string, Object>>>] List of conflicts from the last solver run
       attr_reader :conflicts
 
+      # @return [boolean, nil] flag to indicate that solver should add only required packages
+      #   and not recommended. Nil means not set
+      attr_accessor :only_required
+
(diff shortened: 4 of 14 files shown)


service/lib/agama/software/proposal.rb
```diff
@@ -195,7 +200,8 @@ def initialize_target
         Yast::Pkg.SetPackageLocale(preferred || "")
         Yast::Pkg.SetAdditionalLocales(additional)
 
-        Yast::Pkg.SetSolverFlags("ignoreAlreadyRecommended" => false, "onlyRequires" => false)
+        Yast::Pkg.SetSolverFlags("ignoreAlreadyRecommended" => false,
+          "onlyRequires" => @only_required)
```
lslezak: Sending `nil` will cause reset to the solver default, see https://github.com/yast/yast-pkg-bindings/blob/05457dacb4aa44d118c8b0e6de0a1e26ebb6350e/src/Package.cc#L1820-L1829

I'm not sure if that is expected, we might use `null` in JSON for "not set" or "do not change" meaning in JSON profile. I do not know how it is handled elsewhere.

jreidinger: Thanks for notice, I will change it to `!!@only_required` to send always boolean

rust/agama-lib/src/software/settings.rs
```diff
@@ -37,11 +37,18 @@ pub struct SoftwareSettings {
     /// List of user specified repositories to use on top of default ones.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub extra_repositories: Option<Vec<RepositoryParams>>,
+    /// List of user specified repositories to use on top of default ones.
```
lslezak: Fix the comment please.

lslezak (commented):
Looks good so far...

doc/dbus/org.opensuse.Agama.Software1.doc.xml
```diff
@@ -126,5 +126,12 @@
     </method>
     <property type="a{sy}" name="SelectedPatterns" access="read"/>
     <property type="a(ussa(uss))" name="Conflicts" access="read"/>
+    <!--
+      flag for only required packages
```
mvidner, line 130: Thanks for caring about the API docs!
As an API user, my reaction is, WTF, why can't you decide what the default is and just use a boolean?

mvidner: The `/etc/zypp/zypp.conf` default is `false`
```
##
## Whether only required packages are installed.
##
## Recommended packages, will not be regarded.
##
## Valid values: boolean
## Default value: false
##
# solver.onlyRequires = false
```

service/lib/agama/dbus/software/manager.rb
```diff
@@ -66,10 +66,41 @@ def issues
...
+          # propose again after changing solver flag
+          propose
+        end
+
         SOFTWARE_INTERFACE = "org.opensuse.Agama.Software1"
         private_constant :SOFTWARE_INTERFACE
 
         dbus_interface SOFTWARE_INTERFACE do
+          # Flag for proposing required only dependencies
+          # Do not forget to call Propose method after changing flag
```
mvidner:
```suggestion
          # Propose is called automatically whenever the value is assigned.
```

service/lib/agama/software/manager.rb
```diff
@@ -747,6 +750,41 @@ def copy_zypp_to_target
...
+          ::File.read(final_path(path))
+        end
+
+        # Writes file content with respect of changed root in installation.
+        def self.write(path, content)
+          ::File.write(final_path(path), content)
+        end
+
+        def self.final_path(path)
+          ::File.join("/mnt", path)
```
mvidner: Are you sure hardcoding `/mnt` here is fine?

service/lib/agama/software/manager.rb
```diff
@@ -747,6 +750,41 @@ def copy_zypp_to_target
...
+        return unless proposal.only_required
+
+        zypp_conf = Yast::Packager::CFA::ZyppConf.new(file_handler: TargetFile)
+        zypp_conf.load
+        tree = zypp_conf.generic_get("main")
+        if !tree
+          tree = ::CFA::AugeasTree.new
+          zypp_conf.generic_get("main", tree)
+        end
+        zypp_conf.generic_set("solver.onlyRequires", "true", tree)
```
mvidner: Huh? So we are working with an `Option<bool>` all the time, only to always write `true`? Please tell me I am wrong.

doc/dbus/org.opensuse.Agama.Software1.doc.xml
mvidner: On reading the rest of the code, I can see why we don't want to handle the default value, but it is sad that D-Bus does not have an Option type.

jreidinger: yeap, it would be nice to have option in dbus. Basically I need ternary state. Do not care/yes/no.

service/lib/agama/dbus/software/manager.rb
jreidinger: right, I change it in the end to ensure it works always.

service/lib/agama/software/manager.rb
jreidinger: well, original yast one use `scr_root` which sadly is not reliable as we do not have always switched SCR. For agama we so far always use `/mnt`. I am not sure if we have some global constant or definition that can be used ( it will be helpful also for testing purpose ).

service/lib/agama/software/manager.rb
jreidinger: well, the zypp default is false. So what we do is that we run this code only when user wants to have it as false. I do not want to touch it otherwise as e.g. user wants to modify it via files or script in unattended installation, so I am trying to do minimal changes here.

mvidner: >  we run this code only when user wants to have it as false

You mean `true` :)

And I missed the `return` at the top, OK.

mvidner:
```suggestion
        zypp_conf.generic_set("solver.onlyRequires", "#{!!proposal.only_required}", tree)
```

This is only slightly more complex and guards against stupid mistakes when the default changes

service/lib/agama/software/manager.rb
mvidner:
```suggestion
          ::File.join(Yast::Installation.destdir, path)
```

This is what the rest of Agama uses. Please add the necessary imports.

service/lib/agama/software/manager.rb
lslezak: JFYI: The value in default `zypp.conf` is commented out, the real default is hardcoded in libzypp.

jreidinger: yeap, so maybe it may in the end sense to determine that ternary logic also in agama and set true/false when user explicitly set it and skip it when not set.

PR #2487, opened 2025-06-17
Add hyper-v bootstrap package (bsc#1236961) #2486

teclator:
## Problem

The hyper-v.rpm package includes a number of packages which interact with the host. One useful feature is the hv_kvp_daemon, which reports the active IP addresses to the host tooling. It would be nice to have it in Agama.

- https://bugzilla.suse.com/show_bug.cgi?id=1236961

## Solution

Added the package to the list of bootstrap packages.

PR #2494, opened 2025-06-19
Add hyper-v package only on x86_64 and aarch64

teclator:
## Problem

The package is not available for s390x and ppc64le archs (it was added by #2487)

## Solution

Add the package only on x86_64 and aarch64.

### DIFF
--- a/live/src/agama-installer.kiwi
+++ b/live/src/agama-installer.kiwi
@@ -108,7 +108,7 @@
         <package name="systemd"/>
         <package name="procps"/>
         <package name="iputils"/>
-        <package name="hyper-v" arch="x86_64"/>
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

live/src/agama-installer.kiwi
```diff
@@ -108,7 +108,7 @@
         <package name="systemd"/>
         <package name="procps"/>
         <package name="iputils"/>
-        <package name="hyper-v"/>
+        <package name="hyper-v" arch="x86_64"/>
```
lslezak: Hyper-v is available on aarch64, should we include it as well?

teclator: Looks reasonable.

PR #2495, opened 2025-06-19
(feat UI): Allow to bind a connection to an specific interface using its name or MAC address

teclator:
## Problem

In an Agama profile it is possible to bind a connection to an specific device using its interface name but it is not possible to use the mac address, that option is only supported by the kernel cmdline or using nmcli directly. The profile allows to set a mac address but that is the assign-mac-address instead of the mac-address attribute.

This support has been implemented by #2472 and this PR is a branch from it.

In the **UI** there is no option to bind a connection to an specific device and this could be specially problematic with the default connection.

- https://trello.com/c/yt0Xd9ah

## Solution

Added support in the UI for binding a connection to an specific interface usings its interface name or MAC address.

## Testing

- *Added a new unit test* (TODO).
- *Tested manually*

## Screenshots

[TODO]


web/src/components/network/BindingSettingsForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+    const { id, ...connectionOptions } = connection;
+
+    const updatedConnection = new Connection(id, {
+      ...connectionOptions,
+      iface: state.mode === "iface" ? state.iface : undefined,
+      macAddress: state.mode === "mac" ? state.mac : undefined,
+    });
+
+    updateConnection(updatedConnection)
+      .then(() => navigate(-1))
```
jreidinger, line 170: this `-1` looks quite magic. At least document why and what it do.

web/src/components/network/BindingSettingsForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+  const labelAttributes =
+    valueKey === "macAddress" ? ["macAddress", "name"] : ["name", "macAddress"];
+
+  return (
+    <FormSelect value={value} {...formSelectProps}>
+      {devices.map((device, index) => (
+        <FormSelectOption
+          key={index}
+          value={device[valueKey]}
+          label={labelAttributes.map((attr) => device[attr]).join(" - ")}
```
lslezak: I think we should allow to localize also the format string, some languages might use a different separator than `-`.
```suggestion
          // TRANSLATOS: format the MAC address selection list,
          // the first %s is the interface name (e.g. "eth0"), the second %s is it's MAC address ("01:02:...")
          label={sprintf("%s - %s"), attr, device[attr] }
```

web/src/components/network/BindingSettingsForm.tsx
dgdavid: It's not magic. It is just going back, documented at React Router library.

web/src/components/network/BindingSettingsForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+    updateConnection(updatedConnection)
+      .then(() => navigate(-1))
+      .catch(console.error);
+  };
+
+  return (
+    <Page>
+      <Page.Header>
+        <Content component="h2">{sprintf(_("Binding settings for '%s'"), connection.id)}</Content>
```
lslezak: Please add a TRANSLATORS comment what the `%s` stands for.

web/src/components/network/BindingSettingsForm.tsx
dgdavid: I mean, it is navigating one step back in the history stack. It is mentioned in React Router documentation https://reactrouter.com/6.30.1/hooks/use-navigate#usenavigate and also part of the History API documentation https://developer.mozilla.org/en-US/docs/Web/API/History_API

I see no reason for adding a comment for such an expected behavior each time it is used. *

Of course, we could extract this to a custom function helper or so, but not as part of this PR.

lslezak: For better readability we should define a constant for that.

```js
const GO_BACK = -1;
...
navigate(GO_BACK);
```

dgdavid: No in this PR, please.

web/src/components/network/BindingSettingsForm.tsx
dgdavid: Agree on comments for translators, but the comment is not accurate with the dynamic component behavior. So it should be better explained for translators.

dgdavid: BTW, the suggestions looks broken. Let me think if we can please such a different separator request in a different way to preserver the component behavior.

PR #2496, opened 2025-06-20
fix(web): use exact size for partition

joseivanlopez:
## Problem

The size displayed in the partition/lv form sometimes does not correspond exactly with the number of bytes from the config. Just editing and saving withouth touching any value might slightly change the number of bytes.

## Solution

The size field is always filled with an exact size, using until 2 decimals and the bigger unit possible. If the size presented by the UI is an approximate size, then the size field shows a hint with the approx. size (see screenshot).

Moreover, the concept of max size was hidden in the UI. Now, only a fixed size can be entered, with an option for growing if possible. If the size from the config has a max, then the UI allows to switch to a fixed size.

## Screenshots

localhost_8080_ (24)

### DIFF
--- a/web/src/components/storage/AutoSizeText.tsx
+++ b/web/src/components/storage/AutoSizeText.tsx
@@ -164,7 +164,7 @@ function AutoSizeTextDynamic({
         return sprintf(
           // TRANSLATORS: %1$s is a mount point (eg. /) and %2$s is another one (eg. /home)
           _(
-            "The size range for %1$s will be dynamically adjusted based on the amount of RAM in the system, the usage of Btrfs snapshots and the presence of a separate file system for %2$s.",
+            "The size for %1$s will be dynamically adjusted based on the amount of RAM in the system, the usage of Btrfs snapshots and the presence of a separate file system for %2$s.",
           ),
           path,
           otherPaths[0],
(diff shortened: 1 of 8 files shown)


web/src/components/storage/AutoSizeText.tsx
```diff
@@ -175,7 +175,7 @@ function AutoSizeTextDynamic({
         // TRANSLATORS: %1$s is a mount point and %2$s is a list of other paths
         return sprintf(
           _(
-            "The size range for %1$s will be dynamically adjusted based on the amount of RAM in the system, the usage of Btrfs snapshots and the presence of separate file systems for %2$s.",
+            "The size for %1$s will be dynamically adjusted based on the amount of RAM in the system, the usage of Btrfs snapshots and the presence of separate file systems for %2$s.",
```
dgdavid, line 178: NP: Not sure if we should take the opportunity for clarifying translator the subtle difference between this translation and the one in singular form.

web/src/components/storage/SizeModeSelect.tsx
```diff
@@ -0,0 +1,210 @@
...
+}: SizeModeSelectProps): React.ReactNode {
+  const changeSize = (size: SizeRange) => onChange(value, size);
+
+  return (
+    <Stack hasGutter>
+      <StackItem>
+        <Select
+          id={id || "sizeMode"}
+          value={value}
+          label={value === "auto" ? _("Automatic") : _("Custom")}
```
dgdavid, line 189: NP: Not sure which word is better now, _Custom_ or _Manual_ (after discarding _Fixed_ because of the "Allow grow" behavior). But let's @ancorgs to decide about final wording :)

dgdavid (approved):
In general it LGTM. I miss the unit test for the new _SizeModeSelect_, though.

web/src/components/storage/AutoSizeText.tsx
joseivanlopez: I don't know. I simply kept it as it was there.

PR #2501, opened 2025-06-23
fix(storage): reprobe storage if needed

jreidinger:
## Problem

The storage devices are not reprobed after loading a config with a *dasd* and/or *iscsi* section, even if the system was marked as dirty. This makes the proposal to fail because it does not find the activated devices.

https://bugzilla.suse.com/show_bug.cgi?id=1245114

## Solution

Ensure storage is reprobed if needed while loading a config, even if the config is missing storage section.


rust/agama-lib/src/store.rs
```diff
@@ -203,6 +206,12 @@ impl Store {
         if let Some(dasd) = &settings.dasd {
             self.dasd.store(dasd).await?
         }
+
+        // here we need to check if storage has dirty flag and if so, do the reprobe
+        // it is done this way for efficience, so zFCP, DASD and iSCSI do not need after
+        // for themself after each step. So it is done here before storage loads its setting.
```
joseivanlopez: NP, I would reword this a bit:

```suggestion
        // The storage devices are reprobed if the system was marked as deprecated.
        // Note that a reprobing is not directly done by zFCP, DASD or iSCSI services
        // for a matter of efficiency. For now, clients are expected to explicitly repobe.
        // It is important to reprobe here before loading the storage settings. Otherwise,
        // the new storage devices are not used.
```

joseivanlopez:
I have tested it and it fails with the following message:

~~~
> sudo ./target/debug/agama config load < examples/example-iscsi.json 
✓ The profile is valid.

Error processing storage settings: Backend call failed with status 404 and text ''

Caused by:
    Backend call failed with status 404 and text ''
~~~

~~~
> cat examples/example-iscsi.json 
{
  "iscsi": {
    "initiator": "iqn.1996-04.de.suse:01:351e6d6249",
    "targets": [
      {
        "address": "192.168.100.152",
        "port": 3260,
        "name": "iqn.2025-06.com.example:0a5b93b9a9a79ff5db53",
        "interface": "default",
        "startup": "onboot"
      }
    ]
  }
}
~~~

It works with the code in *master*.

PR #2507, opened 2025-06-24
Added "make sles" to easily build the SLES Live ISO sources

lslezak:
## Problem

- Building the SLES Live ISO sources is not trivial, it needs some manual modifications
- See the exact steps in https://trello.com/c/9sl2EdSl

## Solution

- Added `make sles` target to automate building the sources
  - It uses a special `_multibuild` file for SLES
  - Renames the changes and the Kiwi file to `agama-installer-SLES.changes` and `agama-installer-SLES.kiwi`

## Testing

- Tested manually, both openSUSE and SLES sources are built correctly

## TODO

- [ ] Update the Trello card to use the new `make sles` command

### DIFF
--- a/live/Makefile
+++ b/live/Makefile
@@ -21,10 +21,19 @@ OBS_API = "https://api.opensuse.org"
 OBS_TARGET = "images"
 ARCH = $(shell uname -m)
 
-# files to copy from src/
-COPY_FILES = $(patsubst $(SRCDIR)/%,$(DESTDIR)/%,$(wildcard $(SRCDIR)/*))
+# list of target files, ignore the SLES specific files (matching *_sles pattern)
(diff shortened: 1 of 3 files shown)


live/Makefile
```diff
@@ -80,4 +89,4 @@ shellcheck:
 	@find config-cdroot root src -type f -exec grep -l -E "^#! *(/usr/|)/bin/(ba|)sh" \{\} \; \
```
mvidner: Not SLES related, but since we're touching this Makefile:
`make shellcheck` is looking for `root` which is no longer here

mvidner (approved):
LGTM

BTW the makefile is using advanced features so if we ever need to touch those I would recommend to use a simple makefile with a script in a better-known language (Ruby, shell+shellcheck)

PR #2508, opened 2025-06-25
agama-installer.kiwi: Set OBS-Milestone.

gleidi-suse:
The OBS-Milestone allows, after setting the `%current_milestone` in the OBS project configuration, to have correctly versioned artifacts.

Add the following (or equivalent code) to your project configuration:

```
%define current_milestone Snapshot-202506-1
Macros:
%current_milestone Snapshot-202506-1
:Macros
```

## Problem

SLES 16.0 Agama build artifacts don't have the correct milestone name and break OBS releases.

## Solution

Setting `OBS-Milestone` and the `%current_milestone` macro.

### DIFF
--- a/live/src/agama-installer.kiwi
+++ b/live/src/agama-installer.kiwi
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- OBS-ExclusiveArch: aarch64 ppc64le x86_64 s390x -->
 <!-- OBS-Profiles: @BUILD_FLAVOR@ -->
+<!-- OBS-Milestone: %current_milestone -->
 
 <image schemaversion="7.5" name="agama-installer">
     <description type="system">


gleidi-suse:
Hi :smile: 
> Thanks for the PR, but I'd like to understand what is going on before rubberstamping some magic :)

Good point, I'll try to explain the point to the best of my understanding of it. 

> I don't even understand the Problem:
> 
> > SLES 16.0 Agama build artifacts don't have the correct milestone name and break OBS releases.

Check out https://gitlab.suse.de/buildops/release-support/-/issues/62#note_841322 if you have visibility. Anyway we need in the end to publish SLES Agama bases PXE images. To be able to correctly do so we need:

- change baseiso-containment to copy milestone files along with the created rpms (see osc rq 1288232)
- change agama-installer-SLES to have a milestone (current PR)
- https://github.com/openSUSE/open-build-service/pull/18074

> 
>     * something like "in https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/images/ the binaries shall have Snapshot-YYYYMM-N in the name"?

Only if you have `%current_milestone` set in your project configuration.

> 
>     * break the release in what way?

Without all of the above 3 changes SLES Agama based PXE images do not have the correct filename, and release artifacts are manually renamed.

> 
> If there is some documentation for this, please point to it.
> 
> Can you please point to some OBS sources and binaries where this works correctly, as an example?

https://src.suse.de/products/SLES/src/branch/16.0/kiwi-templates-Minimal/Minimal.kiwi#L3

or

https://build.opensuse.org/projects/home:amorgante:branches:SUSE:SLE-15-SP5:Update:Products:Micro55/packages/SLE-Micro/files/SLE-Micro.kiwi?expand=0 at line 3

I hope this helps clarify why we need this, thank you!

gleidi-suse:
@mvidner do you have any feedback on this change?

PR #2512, opened 2025-06-25
allow minimal installed system without libzypp

jreidinger:
## Problem

- follow-up to #2493
- https://trello.com/c/czY7e3c0

For SLES16 the minimal system can be without libzypp. If it is case together with onlyRequires flag, then software service crashes when trying to load zypp.conf.

## Solution

Add check if file exists and if not, just log it and continue.
Also change as agreed modification of zypp.conf to explicitly write onlyRequires if user explicitly mention it in profile to not just use libzypp default.

## Testing

- *Tested manually*

## Documentation

- Run: `git ls-files '*.md'`

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -129,11 +129,15 @@ jobs:
     - name: Install packages
       run: |
         sudo apt-get update
-        # locales are needed for tests
-        sudo apt-get -y install libclang-18-dev libpam0g-dev python3-langtable locales
+        sudo apt-get -y install libclang-18-dev libpam0g-dev python3-langtable
+
+    - name: Prepare for tests
+      run: |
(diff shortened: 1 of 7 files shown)


jreidinger:
@mvidner according to https://suse.slack.com/archives/C082VE1U2F5/p1750842000769769?thread_ts=1750758112.536869&cid=C082VE1U2F5 I would say it is backward compatible with SLE15 where libzypp is not mandatory on target system. Of course there are ways for users how to add it like sw_management pattern or directly in list of packages. So I would say missing libzypp is feature that was already in SLE15 ( just maybe yast cannot do it? not sure )

imobachgs (commented):
The part about skipping the tests in CI looks good.

lslezak (approved):

PR #2513, opened 2025-06-26
feat: warn the user when using a weak password

imobachgs:
## Problem

Agama's web user interface does not warn the user when using a weak password (e.g., "12", "12345678", "secret", etc.).

- [bsc#1237480](https://bugzilla.suse.com/show_bug.cgi?id=1237480)

## Solution

Agama will not prevent the user about doing that. After all, the unattended mode writes just the password included in the profile. However, it will warn the user in those cases.

This feature is based on `libpwquality`, which checks several aspects related to the password strength. In the future, we might introduce a visual indicator of the strenght or try to display all the rules, but this one looks like a good enough first step.

## Testing

- [ ] Added a new unit test
- [ ] Tested manually

## Screenshots


web/src/assets/styles/index.scss
```diff
@@ -490,6 +490,11 @@ label.pf-m-disabled + .pf-v6-c-check__description {
   }
 }
 
+// Nested alerts should kept no header margin
+:is(.pf-v6-c-content--h4, .pf-v6-c-content h4) {
+  margin-block-start: 0;
+}
+
```
dgdavid, line 497: Hey @imobachgs!

I've added this in commit  54abff01e516a04d65272dd581215cc954c45f8f as a temporary solution for a problem caused by the usage of PF/Content in core/NestedContent and core/SubtleContent

| Before | After |
|-|-|
| Screenshot From 2025-06-27 08-11-27 | Screenshot From 2025-06-27 08-11-10 |

As written there, most probably the _right fix_ should be to stop using PF/Content for these internal wrappers, since it seems it was leading a missuse of it when nesting component, breaking a bit the look& feel. But as commented in the commit, it should be done and testing as part of other PR, in order to ensure nothing is broken while fixing this. In any case, such a change could be something in the line of below diff (briefly tested by me, but discarded to do so when having time to check all NestedContent consumers)

```diff
diff --git a/web/src/components/core/NestedContent.tsx b/web/src/components/core/NestedContent.tsx
index 2ee583e22..9d76d7b40 100644
--- a/web/src/components/core/NestedContent.tsx
+++ b/web/src/components/core/NestedContent.tsx
@@ -21,12 +21,12 @@
  */
 
 import React from "react";
-import { Content, ContentProps } from "@patternfly/react-core";
 import spacingStyles from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 
-type NestedContentProps = {
-  margin?: keyof typeof spacingStyles;
-} & ContentProps;
+type NestedContentProps = React.HTMLProps<HTMLDivElement> &
+  React.PropsWithChildren<{
+    margin?: keyof typeof spacingStyles;
+  }>;
 
 /**
  * Wrapper on top of PF/Content to allow visually nesting its content by adding
@@ -41,8 +41,8 @@ export default function NestedContent({
 }: NestedContentProps) {
   const classNames = [className, spacingStyles[margin]].join(" ");
   return (
-    <Content {...props} className={classNames}>
+    <div {...props} className={classNames}>
       {children}
-    </Content>
+    </div>
   );
 }
 ```

File: web/src/api/users.ts [hunk 2]
```diff
@@ -52,4 +53,16 @@ const fetchRoot = (): Promise<RootUser> => get("/api/users/root");
  */
 const updateRoot = (changes: Partial<RootUser>) => patch("/api/users/root", changes);
 
-export { fetchFirstUser, updateFirstUser, removeFirstUser, fetchRoot, updateRoot };
+/**
+ * Adds the password check result.
```
dgdavid: NP: I'd say _Performs a password chec_, _Check strength of given password_, or so.

web/src/components/users/PasswordCheck.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import { Alert } from "@patternfly/react-core";
+import React, { useEffect, useState } from "react";
+import { checkPassword } from "~/api/users";
+import { _ } from "~/i18n";
+
+const MINIMAL_SCORE = 50;
+
+const PasswordCheck = ({ password }: { password: string }) => {
```
dgdavid, line 30: NP: In the future, it could receive a _minimalScore_ prop  falling back to MININAL_SCORE const.

web/src/components/users/PasswordCheck.tsx
```diff
@@ -0,0 +1,56 @@
...
+      } else {
+        setError("");
+      }
+    });
+  }, [password]);
+
+  if (!error) return;
+
+  return (
+    <div aria-live="polite">
```
dgdavid, line 50: NP: I'm not sure how this should work. I mean, the "aria-live" wrapper is not always there because the component is early returning when it has nothing to render. Maybe is the parent which should wrapper the component in an aria-live div. Or maybe we can simply add the aria-live attribute to the Alert itself if it does not matter that the element is being rendering intermittently. Not sure. I have to read a bit more about it.

dgdavid (approved):
In general it looks good to me, thanks

PR #2514, opened 2025-06-26
fix relative URL resolution for /files

mvidner:
## Problem

1. reported by Leo Manfredi: When using a relative reference in `files`, `agama config generate` does not resolve them
```
  files: [{
     destination: '/usr/local/share/dummy.xml',
     url: 'dummy.xml',
     chroot: false,
  }],
```

- openQA link: https://progress.opensuse.org/issues/184001#note-17

2. it is because the config fails schema validation (differently in Leo's case and my test case)
3. Agama outputs a validation failure message, and the invalid config, without resolving references. (Outputting the invalid config is intentional to make debugging easy)
4. It also does not stop processing at that point. That is IMHO the real bug

## Solution

## Testing

- for now just modify a test to show the bug

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`

### DIFF
--- a/service/test/integration/profile_itest.rb
+++ b/service/test/integration/profile_itest.rb
@@ -127,14 +127,12 @@ describe "agama config" do
       let(:profile_body) do
         json = <<~JSON
           {
-            "scripts": {
-              "postPartitioning": [
+            "files": [
                 {
                   "name": "foo",
                   "url": "my-script.sh"
                 }
               ]
-            }
           }
         JSON
         json


imobachgs:
The source of the problem is that the profile is not valid. The behaviour of the `agama config generate` project is pretty confusing, because it prints the profile and the user might think it was processed despite the error. See https://github.com/agama-project/agama/blob/master/rust/agama-cli/src/config.rs#L233.

PR #2517, opened 2025-06-27
fix(web): improve software queries invalidation

imobachgs:
## Problem

The software configuration is handled using two different queries: `software/config` and `software/product`. The invalidation of those queries happen in different places: `useProductChanges` and `useInstallerStatusChanges`.

We found this problem while trying to debug [bsc#1243276](https://bugzilla.suse.com/show_bug.cgi?id=1243276) but we do not think it is going to solve that problem.

## Solution

This PR unifies the handling of the configuration. The `software/product` is removed in favor of `software/config`, and the `useInstallerStatusChanges` does not invalidate the software queries anymore.

Additionally, when the product changes, all software queries are invalidated.

## Testing

- Tested manually*


dgdavid (approved):
LGTM

PR #2484, opened 2025-06-16
Add Kalpa to products

sfalken:
Adds Kalpa Desktop as a product, for the Agama Installer

PR #2520, opened 2025-06-30
fix(products): fix Kalpa product definition

imobachgs:
## Problem

#2484 added Kalpa to the list of installable products. However, the definition is wrong and the installation does not work.

## Solution

Fix the product definition to make Kalpa installable.

## Testing

- Tested manually

### DIFF
--- a/products.d/kalpa.yaml
+++ b/products.d/kalpa.yaml
@@ -5,14 +5,14 @@ name: Kalpa Desktop
 # at the at translations/description key below to avoid using obsolete
 # translations!!
 # ------------------------------------------------------------------------------
-description: 'A rolling release immutable desktop product, using the Plasma
+description: "A rolling release immutable desktop product, using the Plasma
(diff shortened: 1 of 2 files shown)


products.d/kalpa.yaml
```diff
@@ -5,14 +5,14 @@ name: Kalpa Desktop
...
+description: "A rolling release immutable desktop product, using the Plasma
   Desktop, leveraging Flatpak for Application Delivery, a Read-Only base, and
-  automatic and atomic updates of your system'
+  automatic and atomic updates of your system"
 icon: Kalpa.svg
 # Do not manually change any translations! See README.md for more details.
 translations:
   description:
-
+software:
```
mvidner, line 15: Oops, what a mistake in the previous PR :cry:

mvidner (approved):

products.d/kalpa.yaml
imobachgs: Yes, we might need some validation for products file too.

PR #2523, opened 2025-06-30
fix(web): improve device size 

joseivanlopez:
Follow-up of https://github.com/agama-project/agama/pull/2496.


web/src/components/storage/SizeModeSelect.tsx
```diff
@@ -72,15 +72,15 @@ function UnsupportedSize({ value, onClick }: UnsupportedSizeProps): React.ReactN
...
+              "The size is configured as a range between %s and %s, but this interface cannot handle ranges with a given max size.",
             ),
             value.min,
             value.max,
           )}
         </StackItem>
         <StackItem>
           <Button variant="link" isInline onClick={onClick}>
-            {_("Do you want to use a supported size?")}
+            {_("Do you want to redefine the size limits?")}
```
dgdavid: This change is indeed an improvement, but I wonder if using word "limits" can lead to confusion because the mentioned "max size" will be not there in the interface except for the "unlimited max size", aka "Allow growing".

What about something more in the line of 

"Do you want to discard the maximum size and continue with simplified configuration?"

Even better, although we have to check how it was done at user hashed passwords and change it accordingly, maybe we should make the action more direct: "Discard the maximum size and continue with simplified configuration"

dgdavid: And also we can through there all the information.

> The current configuration defines a fixed size range (%s to %s), but this interface only supports a minimum size with an optional “Allow to grow” setting.
> 
> [Discard fixed maximum size and continue with simplified size configuration]

or 

> The current configuration defines a fixed size range (%s to %s), but this interface only supports a minimum size with an optional “Allow to grow” setting.
> 
> [Discard fixed maximum size]

joseivanlopez: I also prefer this last version without a question but an action.

PR #2524, opened 2025-06-30
feat(web): warn the user about weak encryption passwords

imobachgs:
#2513 introduced a warning when using a weak password for the root or the first user. However, Agama days nothing if the encryption password for your disk is weak. This PR introduces the same behavior for the encryption password.


ancorgs (approved):
LGTM. Tested manually to double-check.

PR #2490, opened 2025-06-18
Ipmi support for the Agama installer

mchf:
## Problem

https://jira.suse.com/browse/PED-12285
We were requested to signal Agama's status via IPMI

## Solution

Initial implementation for IPMI statuses. Works if ```/dev/ipmi0``` is found in the system.

## Testing

- *Tested manually* (manual installer run together with ipmi emulator)

PR #2525, opened 2025-06-30
 ipmitool invocation: do not write spurious % percent sign

mvidner:
## Problem

Noticed a suspicious character in `ipmitool` input during team review of #2490.
The tool may be permissive and work just fine, but let's remove it just to be sure.

- https://trello.com/c/F1y5U83s
- https://jira.suse.com/browse/PED-12285

## Solution

- remove `%`
- format the number in hex
- add comments about the magic numbers

## Testing

Not yet :cry: , just ran `irb` to check that `31.to_s(16)` returns `"1f"`.

Looked really hard at the change and hope that it improves the result of calling the command.

Will use an IPMI emulator as hinted in jira

## Screenshots

No

## Documentation

Developer: added specification links
User: no

### DIFF
--- a/service/lib/agama/ipmi.rb
+++ b/service/lib/agama/ipmi.rb
@@ -61,6 +61,7 @@ module Agama
 
   private
 
+    # See the body of #command for specification links
     IPMI_STARTED = 0x7
     IPMI_FINISHED = 0x8
     IPMI_ABORTED = 0x9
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

mvidner:
To have a VM with a `/dev/ipmi0` device, add this to your `qemu-kvm` command:
`-device ipmi-bmc-sim,id=bmc0 -device isa-ipmi-kcs,bmc=bmc0,irq=5`

With it, `ipmitool` runs and reports a problem in the event format:

```console
# cat /tmp/ipmi1bad.txt
0x4 0x1F 0x0 0x6f %9 0x0 0x0 # before fix
# ipmitool event file /tmp/ipmi1bad.txt
Invalid token in file: [%9]
Invalid Event: 041f006f000000
# echo $?
1
```

Replacing `%` with `0x`, a different error appears:
```console
# cat /tmp/ipmi2bad.txt
0x4 0x1F 0x0 0x6f 0x9 0x0 0x0 # after fix
# ipmitool event file /tmp/ipmi2bad.txt
Get Channel Info command failed: Invalid command
Failed to send the platform event via an unknown channel
# echo $?
1
```


That's probably because `ipmi-bmc-sim` is a *basic* simulator:
>  This is a simulation of a hardware management interface processor that normally sits on a system. It provides a watchdog and the ability to reset and power control the system.

PR #2526, opened 2025-07-01
fix(storage): inhibit storage subsystems

joseivanlopez:
## Problem

Agama live media differs from the old YaST installation media. With the new live ISO, systemd is present and there are some storage subsystems that interfer with the libstorage-ng operation (e.g., udisk, systemd mounts, mdadm auto-assembly). 

https://bugzilla.suse.com/show_bug.cgi?id=1245159
https://bugzilla.suse.com/show_bug.cgi?id=1246133

## Solution

Call to the storage inhibitors while starting the storage service, similar to what the YaST Expert Partitioner client does in an installed system.

### DIFF
--- a/service/bin/agamactl
+++ b/service/bin/agamactl
@@ -61,7 +61,6 @@ def start_service(name)
 
   logger = logger_for(name)
   service_runner = Agama::DBus::ServiceRunner.new(name, logger: logger)
-  logger.info "Starting the service"
   service_runner.run
 end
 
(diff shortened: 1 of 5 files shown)


joseivanlopez:
It was tested by QA and it works, see https://bugzilla.suse.com/show_bug.cgi?id=1246133#c17.

ancorgs (approved):
LGTM

PR #2527, opened 2025-07-01
Add Leap Micro 6.2

lkocman:
## Problem

Add Leap Micro 6.2

Micro 6.3 will be simply part of Leap 16.1, however I'd like to market it as a separate entry / product similar to what MicroOS is for Tumblweed. Let's start early with 6.2.


products.d/leap_micro_62.yaml
```diff
@@ -0,0 +1,110 @@
+id: LeapMicro_6.2
+name: openSUSE Leap Micro 6.2 Beta
```
imobachgs, line 2: You might want to add the archs list here: `archs: x86_64,aarch64`.

imobachgs (changes requested):
In general, it looks good. But you need to add the file to the [agama-products.spec](https://github.com/agama-project/agama/blob/master/products.d/agama-products.spec#L59) so it gets included into the `agama-products-opensuse` package.

products.d/leap_micro_62.yaml
lkocman: Yes!

lkocman:
> In general, it looks good. But you need to add the file to the [agama-products.spec](https://github.com/agama-project/agama/blob/master/products.d/agama-products.spec#L59) so it gets included into the `agama-products-opensuse` package.

Added. Thx for catching this!

PR #2528, opened 2025-07-01
Make agama-installer build more reproducible (boo#1245501)

imobachgs:
Synchronize with changes from https://bugzilla.suse.com/show_bug.cgi?id=1245501.


mvidner:
JFYI, @bmwiedemann

PR #2529, opened 2025-07-02
fix(storage): add missing iSCSI discovery

joseivanlopez:
## Problem

The iSCSI discovery is not done while importing an iSCSI config. Without that, the target login fails.

https://bugzilla.suse.com/show_bug.cgi?id=1245171

## Solution

Add the missing discovery step.

Requires https://github.com/yast/yast-iscsi-client/pull/146.

### DIFF
--- a/service/lib/agama/storage/iscsi/adapter.rb
+++ b/service/lib/agama/storage/iscsi/adapter.rb
@@ -63,6 +63,14 @@ module Agama
           Yast::IscsiClientLib.discover(host, port, authentication(credentials), silent: true)
         end
 
+        # Discovers iSCSI targets from a portal.
+        #
+        # @param portal [String]
+        # @param interfaces [Array<String>]
+        def discover_from_portal(portal, interfaces: [])
+          Yast::IscsiClientLib.discover_from_portal(portal, interfaces)
+        end
+
         # Reads the iSCSI initiator config.
         # @return [Initiator]
         def read_initiator
--- a/service/lib/agama/storage/iscsi/manager.rb
+++ b/service/lib/agama/storage/iscsi/manager.rb
@@ -296,9 +296,10 @@ module Agama
         def apply_targets_config(config)
           return unless config.targets
 
-          start_progress_with_size(2)
+          start_progress_with_size(3)
           progress.step(_("Logout iSCSI targets")) { logout_targets }
-          progress.step(_("Login iSCSI targets")) { login_targets(config.targets) }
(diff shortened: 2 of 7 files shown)


service/lib/agama/storage/iscsi/manager.rb
```diff
@@ -296,26 +297,51 @@ def apply_initiator_config(config)
...
+        end
+
+        # Discovers iSCSI targets from a portal.
+        #
+        # @param portal [String]
+        # @param interfaces [Array<String>]
+        def discover_from_portal(portal, interfaces: [])
+          iface_params = interfaces.map { |i| ["-I", i] }.flatten
+
+          Yast::Execute.locally!(
```
ancorgs: I didn't expect a call to `iscsiadm` here. At most, I would expect something like:

- `IscsiClientLib` calls `iscsiadm`
- `Agama::Storage::ISCSI::Adapter` relies on `IscsiClientLib`
- The iSCSI manager relies on  the adapter.

Maybe some shorcuts are needed for this, but calling `iscsiadm` directly from the manager (which is the upper level of the current architecture), looks like quite a leap to me.

service/lib/agama/storage/iscsi/adapter.rb
```diff
@@ -63,6 +63,14 @@ def discover(host, port, credentials: {})
           Yast::IscsiClientLib.discover(host, port, authentication(credentials), silent: true)
         end
 
+        # Discovers iSCSI targets from a portal.
+        #
+        # @param portal [String]
+        # @param interfaces [Array<String>]
+        def discover_from_portal(portal, interfaces: [])
+          Yast::IscsiClientLib.discover_from_portal(portal, interfaces)
```
ancorgs, line 71: If I'm not wrong, this can raise an exception (is using `Execute.locally!`) which is not controlled here at Agama. So this looks like the new code could result in an Agama crash if something goes wrong (eg. the `iscsiadm` binary is not there).

I know I approved the yast-iscsi-client pull request, but thinking it twice... that execution error should likely be controlled there. Other methods of `IscsiClientLib` include timeouts, error handling, etc. and finally simply return nil, false or an empty array if things go wrong. While `IscsiClientLib.discover_from_portal` leaves that responsibility to the caller. I can live with that, but then we need to add that control here at Agama.

joseivanlopez: Fixed at `IscsiClientLib` side, see https://github.com/yast/yast-iscsi-client/pull/147.

PR #2531, opened 2025-07-02
fix(rust): cache products during the startup

imobachgs:
## Problem

Some time ago we, we introduced [a cache](https://github.com/agama-project/agama/pull/2364) to store the software data and avoid blocking the UI. The problem is that the list of products is cached only when a client asks for it. So in the case of an unattended installation, it might happen that the installation starts before the web UI asks for the list of products for the first time.

- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1241208

## Solution

Read the list of products during the web server initialization. It will be updated only when the locale changes.

## Testing

- *Tested manually*


teclator (approved):
LGTM

PR #2532, opened 2025-07-03
fix(storage): keep storage config after registration

joseivanlopez:
## Problem

The storage config is lost after registering a product. 

https://bugzilla.suse.com/show_bug.cgi?id=1245400

## Solution

Reprobe the system instead of a complete probing.

Note: in the future, a better solution will be implemented to avoid delegating the reprobing call to the clients, see https://gist.github.com/joseivanlopez/8d8103ffd4a8e5539046cb6e134be05f.

### DIFF
--- a/rust/agama-lib/src/manager/http_client.rs
+++ b/rust/agama-lib/src/manager/http_client.rs
@@ -1,4 +1,4 @@
-// Copyright (c) [2024] SUSE LLC
+// Copyright (c) [2024-2025] SUSE LLC
 //
 // All Rights Reserved.
 //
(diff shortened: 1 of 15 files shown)


web/src/api/manager.ts
```diff
@@ -32,6 +32,11 @@ const startProbing = () => post("/api/manager/probe");
  */
 const probe = () => post("/api/manager/probe_sync");
 
+/**
+ * Triggers a synchronous reprobing process.
```
dgdavid, line 36: Jus a question, does Agama server provide their async counterparts? Asking because I also see in manager "probe_sync" and "probe".

joseivanlopez: No, an async version is not needed yet.

rust/agama-lib/src/manager/http_client.rs
```diff
@@ -52,6 +52,13 @@ impl ManagerHTTPClient {
         Ok(self.client.post_void("/manager/probe_sync", &()).await?)
     }
 
+    /// Starts a "reprobing".
+    pub async fn reprobe(&self) -> Result<(), ManagerHTTPClientError> {
+        // BaseHTTPClient did not anticipate POST without request body
+        // so we pass () which is rendered as `null`
+        Ok(self.client.post_void("/manager/reprobe_sync", &()).await?)
```
dgdavid, line 59: 🤯

web/src/api/manager.ts
joseivanlopez: Anyway, all this about sync/async probing/reprobing is something to revisit. For now, just following the current approach.

dgdavid: > Anyway, all this about sync/async probing/reprobing is something to revisit.

Great then.

Thank you

PR #2534, opened 2025-07-04
feat: notify the user when the auto-installation cannot be started

imobachgs:
## Problem

Until now, Agama does not notify the user when the configuration for an unattended installation cannot be loaded.

### A partial solution

This is the first step towards having better handling of the auto-installation process. It is not 100% complete, as it does not report exactly the problem found when loading the profile, but we do not want to invest more time into this at this point. We plan to do it better for SLES 16.1.

### Replacing `agama-auto`

To implement this behavior, we decided to replace the old `agama-auto` script with a new `agama-autoinstall` binary. The new binary can use all the goodies from `agama-lib` and, in the future, it will not need to rely on the command-line interface anymore.

### The behavior

If the user specifies an `inst.auto` and the profile could not be loaded, Agama will display a sort of generic message asking the user whether to retry. In the future, we would like to:

* Offer better diagnostics.
* Allow the user to specify a new URL.

If the user decides to not retry, Agama will not start the auto-installation process.

If the user did not specify a `inst.auto`, Agama tries to silently load one from the pre-defined locations.

### A new `inst.install` option

Setting `inst.install` to `0` makes Agama to stop before starting the installation, after loading the configuration. It might be useful for inspecting the result of applying the configuration before starting the installation.

## Testing

- [x] Added a new unit test (partial)
- [x] Tested manually

## Screenshots

The text needs some improvements, but the screenshot is useful to get the idea.

config-load-error

## Follow-up

- Document the new `inst.install` cmdline argument.
- Allow the user to specify a new URL.
- Do not rely on the CLI anymore.
- Extend the CLI to support asnwering questions (like this one).

### DIFF
--- a/rust/agama-autoinstall/src/auto_loader.rs
+++ b/rust/agama-autoinstall/src/auto_loader.rs
@@ -107,7 +107,8 @@ impl ConfigAutoLoader {
     async fn should_retry(&self, url: &str) -> anyhow::Result<bool> {
         let text = format!(
             r#"
-                It was not possible to load the configuration from {url}. Do you want to try again?"
+                It was not possible to load the configuration from {url}.
+                It was unreachable or invalid. Do you want to try again?"
                 "#
         );
         let generic = GenericQuestion {


live/src/agama-installer.kiwi
```diff
@@ -145,7 +145,7 @@
         <package name="agama-web-ui"/>
         <package name="agama-cli"/>
         <package name="agama-cli-bash-completion"/>
-        <package name="agama-auto"/>
+        <package name="agama-autoinstall"/>
```
jreidinger, line 148: can we rename packages in RC phase?

rust/agama-autoinstall/src/loader.rs
```diff
@@ -0,0 +1,135 @@
...
+        Ok(Self {
+            questions: QuestionsHTTPClient::new(http_client)?,
+        })
+    }
+
+    /// Loads the configuration from the given URL.
+    ///
+    /// When running in interactive mode, it report errors and ask the user
+    /// to continue.
+    pub async fn load(&self, url: &str, interactive: bool) -> anyhow::Result<()> {
```
jreidinger: to be honest I am not that happy about mixing that interactive mode with config loader. I kind of see it on different layer..so for me it makes more sense to
1. have class for config loader
2. have class that use config loader and ask for retry via question.

Maybe even more reusable with three classes where the third one is representation for question, so the class 2 can get question and result and return if it should be retried or not.

rust/agama-autoinstall/src/cmdline.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::{collections::HashMap, path::Path};
+
+use anyhow::Context;
+
+/// Implements a mechanism to read the kernel's command-line arguments.
+///
+/// It supports multiple values for a single key.
+pub struct CmdlineArgs(HashMap<String, Vec<String>>);
+
+impl CmdlineArgs {
```
jreidinger: I really dislike name as it is confusing with CLI command line. I would call it `KernelCmdline` or call file `kernel` and then just `args` struct here...so usage will be `kernel::args`

rust/install.sh
```diff
@@ -34,9 +34,10 @@ install6() {
     install -m 0644 "$@"
 }
 
-install -D -t "${DESTDIR}${bindir}" "${SRCDIR}"/target/${RUST_TARGET}/agama
-install -D -t "${DESTDIR}${bindir}" "${SRCDIR}"/target/${RUST_TARGET}/agama-dbus-server
-install -D -t "${DESTDIR}${bindir}" "${SRCDIR}"/target/${RUST_TARGET}/agama-web-server
+install -D -t "${DESTDIR}${bindir}" "${SRCDIR}"/target/"${RUST_TARGET}"/agama
```
jreidinger: the change looks strange, why not simple `"${SRCDIR}/target/${RUST_TARGET}/agama"`

rust/agama-autoinstall/src/loader.rs
imobachgs: Well, we could move the responsability to ask for retry to the `AutoLoader` struct which is the one that orchestrates the whole process.

live/src/agama-installer.kiwi
imobachgs: I hope so. It is mostly an implementation detail and only for the installation media.

jreidinger (approved):

live/src/config.sh
```diff
@@ -59,7 +59,7 @@ systemctl enable avahi-daemon.service
 systemctl enable agama.service
 systemctl enable agama-web-server.service
 systemctl enable agama-dbus-monitor.service
-systemctl enable agama-auto.service
```
mvidner, line 62: Also delete this file and the script it runs?

imobachgs: Yes. It is replaced with `agama-autoinstall`.

PR #2540, opened 2025-07-07
Do not set the LSM resolvables but use the software selection

teclator:
## Problem

Currently **Selinux** is selected based on the product **LSM** definition but it is not possible to deselect it from the Software selection and it is always installed.


service/lib/agama/manager.rb
```diff
@@ -102,7 +102,10 @@ def config_phase
       #   storage config could be applied before probing. In that case, the config has to be
       #   recovered.
       progress.step { first_time ? probe_and_recover_storage : storage.probe }
-      progress.step { software.probe }
+      progress.step do
+        software.probe
+        security.probe
```
jreidinger: I miss comment here we discuss about reasons why security have to be after software, so we do not break it in future ( or can re evaluate if conditions change )

service/lib/agama/security.rb
```diff
@@ -63,26 +63,30 @@ class Security
...
-    end
-
-    def probe
-      selected_lsm = config.data.dig("security", "lsm")
-      lsm_config.select(selected_lsm)
+      return if lsm_patterns.empty?
 
-      patterns = if selected_lsm.nil?
-        []
+      if (lsm_patterns - proposal_patterns).empty?
```
jreidinger: so if I get it, for now it only detects if lsm patterns are unselected and if so, then it won't save anything.

So what is not covered in this pr is
1. switch to apparmor
2. if we use selinux for live medium, than it is not removed from kernel params.

I think both are ok if there are somewhere captured ( I think on last retro we talk about follow ups? )

jreidinger: if we want to address both, then we will need to check what each lsm uses as patterns and if none was selected by user then use none lsm which will remove kernel params.

service/lib/agama/dbus/storage/manager.rb
```diff
@@ -520,6 +520,7 @@ def register_software_callbacks
           backend.software.on_probe_finished do
             # A PropertiesChanged signal is emitted (see ::DBus::Object.dbus_reader_attr_accessor).
             self.encryption_methods = read_encryption_methods
+            backend.security.probe
```
jreidinger, line 525: please coment why it is needed.

teclator: Yes, I was already doing it, please check the commit below.

service/lib/agama/manager.rb
```diff
@@ -201,6 +204,13 @@ def network
       @network ||= Network.new(logger)
     end
 
+    # Security manager
+    #
+    # @return [Security]
+    def security
```
jreidinger: it is not longer needed. So please remove this method and also line 27 as it can confuse others to use it and cause inconsistent data.

teclator: true

service/lib/agama/security.rb
teclator: So, I'm going to implement this proposal.

Therefore, if the default LSM patterns are selected it will be choosen if not we will check the other candidates and the first that match the proposal will be used instead.

jreidinger (approved):

PR #2541, opened 2025-07-08
feat(web): RMT registration

joseivanlopez:
Add support for RMT in the web UI.

* https://jira.suse.com/browse/AGM-156
* https://bugzilla.suse.com/show_bug.cgi?id=1244625

localhost_8080_ (25)

localhost_8080_ (26)

### DIFF
--- a/service/lib/agama/dbus/software/product.rb
+++ b/service/lib/agama/dbus/software/product.rb
@@ -387,6 +387,8 @@ module Agama
           connect_result_from_error(e, first_error_code + 8)
         rescue Agama::Software::ServiceError => e
           connect_result_from_error(e, first_error_code + 9)
+        rescue StandardError => e
+          connect_result_from_error(e, first_error_code + 10)
         end
 
         # Generates a result from a given error.
--- a/service/lib/agama/dbus/software/product.rb
+++ b/service/lib/agama/dbus/software/product.rb
@@ -212,7 +212,7 @@ module Agama
           if !backend.product
             [1, "Product not selected yet"]
           # report success and do nothing when already registered with the same code
-          elsif backend.registration.reg_code == reg_code
+          elsif backend.registration.registered && backend.registration.reg_code == reg_code
             [0, ""]
           elsif backend.registration.registered
             [2, "Product already registered"]
--- a/service/test/agama/registration_test.rb
+++ b/service/test/agama/registration_test.rb
@@ -243,7 +243,7 @@ describe Agama::Registration do
             expect(subject.reg_code).to eq("11112222")
           end
 
-          it "does not set the email" do
+          it "sets the email" do
             expect { subject.register("11112222", email: "test@test.com") }
               .to raise_error(Timeout::Error)
 
--- a/service/test/agama/registration_test.rb
+++ b/service/test/agama/registration_test.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2023] SUSE LLC
+# Copyright (c) [2023-2025] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -95,8 +95,13 @@ describe Agama::Registration do
       context "and the product is not registered yet" do
         it "announces the system" do
           expect(SUSE::Connect::YaST).to receive(:announce_system).with(
-            { language: anything, token: "11112222", email: "test@test.com",
-              verify_callback: anything },
+            {
+              language:        anything,
+              url:             "https://scc.suse.com",
+              token:           "11112222",
+              email:           "test@test.com",
+              verify_callback: anything
+            },
             "test-5-x86_64"
           )
 
@@ -105,8 +110,13 @@ describe Agama::Registration do
 
         it "sets the current language in the request" do
           expect(SUSE::Connect::YaST).to receive(:announce_system).with(
-            { language: "de-de", token: "11112222", email: "test@test.com",
-              verify_callback: anything },
+            {
+              language:        "de-de",
+              url:             anything,
+              token:           "11112222",
+              email:           "test@test.com",
+              verify_callback: anything
+            },
             "test-5-x86_64"
           )
 
@@ -437,8 +447,13 @@ describe Agama::Registration do
 
         it "deactivates the system" do
           expect(SUSE::Connect::YaST).to receive(:deactivate_system).with(
-            { token: "11112222", email: "test@test.com", verify_callback: anything,
-              language: anything }
+            {
+              url:             anything,
+              token:           "11112222",
+              email:           "test@test.com",
+              verify_callback: anything,
+              language:        anything
+            }
           )
 
           subject.deregister
--- a/service/test/agama/registration_test.rb
+++ b/service/test/agama/registration_test.rb
@@ -236,25 +236,25 @@ describe Agama::Registration do
               .to raise_error(Timeout::Error)
           end
 
-          it "does not run the callbacks" do
-            expect(callback).to_not receive(:call)
-
-            expect { subject.register("11112222", email: "test@test.com") }
-              .to raise_error(Timeout::Error)
-          end
-
-          it "does not set the registration code" do
+          it "sets the registration code" do
             expect { subject.register("11112222", email: "test@test.com") }
               .to raise_error(Timeout::Error)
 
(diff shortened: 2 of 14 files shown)


web/src/components/product/RegistrationExtension.tsx
```diff
@@ -136,24 +141,18 @@ export default function RegistrationExtension({
           <RegisteredExtensionStatus registrationCode={registrationData.registrationCode} />
         )}
         {!isRegistered && extension.available && !extension.free && (
-          <Form id={`register-form-${extension.id}-${extension.version}`} onSubmit={submit}>
+          <Form id={formId} onSubmit={submit}>
             {/* // TRANSLATORS: input field label */}
-            <FormGroup label={_("Registration code")}>
+            <FormGroup fieldId={inputId} label={_("Registration code")}>
```
dgdavid, line 146: I've updated the `id` value here to ensure proper association between the label and its corresponding input. This was identified while writing unit tests using [getByLabelText](https://testing-library.com/docs/queries/bylabeltext/), which relies on that association. We're using this method specifically for password fields, since they can't be queried by role because they has no implicit role. Read https://github.com/testing-library/dom-testing-library/issues/567

web/src/components/product/RegistrationExtension.tsx
```diff
@@ -136,24 +141,18 @@ export default function RegistrationExtension({
...
                 isDisabled={loading}
-                id={`input-reg-code-${extension.id}-${extension.version}`}
+                id={inputId}
                 value={regCode}
                 onChange={(_, v) => setRegCode(v)}
               />
             </FormGroup>
             <ActionGroup>
-              <Button
-                id={`register-button-${extension.id}-${extension.version}`}
```
dgdavid, line 151: For consistency with other ids, I moved this one into a variable too. That said, I’m not sure if it’s still necessary since it seems like it was originally added for testing purposes, but I have updated the test to avoid relying on ids. I’ve kept it for now to be safe, but it might be worth revisiting and potentially removing later.

web/src/components/product/ProductRegistrationPage.test.tsx
```diff
@@ -168,87 +316,163 @@ describe("ProductRegistrationPage", () => {
...
+          // only the end of the code is displayed
+          screen.getByText(/\*+ad42/);
+          // not the full code
+          expect(screen.queryByText(registeredAddonInfoMock[0].registrationCode)).toBeNull();
+
+          // after pressing the "Show" button
+          await user.click(visibilityCodeToggler);
+          // the full code is visible
+          screen.getByText(registeredAddonInfoMock[0].registrationCode);
+        });
```
dgdavid, line 475: NP: While updating this, I had the feeling that such a test actually belongs to the missing RegistrationExtensions.test.tsx

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -87,24 +109,221 @@ const RegisteredProductSection = () => {
...
   const { mutate: register } = useRegisterMutation();
+  const [server, setServer] = useState<ServerOption>("default");
+  const [url, setUrl] = useState("");
   const [key, setKey] = useState("");
   const [email, setEmail] = useState("");
+  const [provideKey, setProvideKey] = useState(true);
   const [provideEmail, setProvideEmail] = useState(false);
-  const [error, setError] = useState<string | null>(null);
+  const [requestError, setRequestError] = useState<string | null>(null);
+  const [errors, setErrors] = useState<string[]>([]);
```
dgdavid, line 270: NP: Not a blocking comment but just a suggestion for the future: when a component or form starts managing a large number of state variables, I feel can be cleaner and more maintainable to use a reducer instead.

service/lib/agama/dbus/software/product.rb
```diff
@@ -212,7 +212,7 @@ def register(reg_code, email: nil)
           if !backend.product
             [1, "Product not selected yet"]
           # report success and do nothing when already registered with the same code
-          elsif backend.registration.reg_code == reg_code
+          elsif backend.registration.registered && backend.registration.reg_code == reg_code
```
jreidinger, line 215: good catch :+1:

service/test/agama/registration_test.rb
```diff
@@ -226,25 +236,25 @@
...
-
+          it "sets the registration code" do
             expect { subject.register("11112222", email: "test@test.com") }
               .to raise_error(Timeout::Error)
+
+            expect(subject.reg_code).to eq("11112222")
           end
 
-          it "does not set the registration code" do
+          it "does not set the email" do
```
jreidinger: hmm, title says does not set, but in test below it is set. This looks confusing to me.

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Mon Jul 14 16:02:55 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Allow spcifying the registration server (jsc#AGM-156).
```
jreidinger: typo `spcifying`

web/src/api/http.ts
```diff
@@ -48,7 +48,8 @@ const patch = (url: string, data?: object) => http.patch(url, data);
  * @param url - endpoint URL
  * @param data - request payload
  */
-const put = (url: string, data: object) => http.put(url, data);
+const put = (url: string, data: object | string, config?: AxiosRequestConfig) =>
```
jreidinger, line 52: missing documentation of new parameter `config`

web/src/api/software.ts
```diff
@@ -111,6 +111,12 @@ const probe = () => post("/api/software/probe");
 const register = ({ key, email }: { key: string; email?: string }) =>
   post("/api/software/registration", { key, email });
 
+/**
+ * Updates the URL for the registration
+ */
+const updateRegistrationUrl = (url: string) =>
+  put("/api/software/registration/url", url, { headers: { "Content-Type": "application/json" } });
```
jreidinger, line 120: any reason why in this case the content-type is required where other parts do not need it? I think reason should be ideally part of the code so other readers who have same question will see it

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -87,24 +109,221 @@ const RegisteredProductSection = () => {
...
+  return (
+    <FormGroup fieldId={id} label={SERVER_LABEL}>
+      <Select
+        id={"server"}
+        value={value}
+        label={value === "default" ? SCC_SERVER_LABEL : CUSTOM_SERVER_LABEL}
+        onChange={(v: ServerOption) => onChange(v)}
+      >
+        <SelectList aria-label={_("Server options")}>
+          <SelectOption value="default" description={_("Register to scc.suse.com")}>
```
jreidinger: I am not sure how much users are familiar with scc.suse.com...maybe it can be something like `Register to SUSE server`?

web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -87,24 +109,221 @@ const RegisteredProductSection = () => {
...
+        <Checkbox
+          id={optionalId}
+          label={_("Provide email address")}
+          isChecked={isProvided}
+          onChange={() => onProvidedChange(!isProvided)}
+        />
+      </FormGroup>
+
+      {isProvided && (
+        <NestedContent margin="mxMd" aria-live="polite">
```
jreidinger, line 251: here I have to say that I do not get what that margin value means and AI was surprisingly helpful with explaning what it means. So it is `In essence, margin="mxMd" tells the component to apply a predefined "medium" horizontal margin to itself.`

dgdavid: You just had to look at PatternFly documentation 😉 But since it is a kind of common pattern, AI was able to tell you the right answer 😄

dgdavid: See https://www.patternfly.org/utility-classes/spacing#documentation

That said, it's true we can also improve a little bit the Nested content documentation for at least linking the above link

https://github.com/agama-project/agama/blob/523ef4d65e729e5508ed9cd92666845b0d703e65/web/src/components/core/NestedContent.tsx#L26

jreidinger (approved):

PR #2542, opened 2025-07-08
chore: drop the deprecated agama-auto service

imobachgs:
The auto-installation mechanism was replaced with a Rust-based alternative (see #2534). For this reason, the code below `autoinstall` directory is not needed anymore.


dgdavid (approved):
Hard to put a change entry here, I guess.

LGTM

PR #2543, opened 2025-07-09
Use the DUD repository for installing packages

lslezak:
## Problem

- The DUD repository is not used for installing packages into the target system

## Solution

- Add the local DUD repository into the list of used repositories, use a higher priority so the provided packages are preferred (even when they have a lower version than in the other repositories). 

## Notes

- Libzypp supports "plaindir" repository type, that is basically a local directory with RPM files without any metadata. When no metadata is found in a `dir:/` repository then libzypp automatically switches to the "plaindir" type and scans the available RPMs there.
- That is easier to handle (no need to create metadata) and it saves some RAMdisk space (not creating the metadata files). The most important feature is that the "plaindir" repositories are by definition not GPG signed so libzypp does not complain about using an unsigned repository.
- So we can remove the `createrepo_c` package from the Live medium and do not call it.

## TODO

- [ ] Add an option for ignoring the GPG problems for the DUD packages

## Testing

- Tested manually with a DUD with trivial [hello-world](https://build.opensuse.org/package/show/home:lslezak:dud-test2/hello-world) package
- Note: The testing package uses the `Supplements: aaa_base` dependency to select the package to install, otherwise the package would need to be selected to install by using CLI.

## Screenshots

The testing package is signed by an unknown GPG key, an error is reported during installation:

agama-install-unsigned-package

But after pressing "Continue Anyway" it is installed into the system and works as expected and the temporary DUD repository is not included in the installed system:

agama-dud-package-installed

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -59,7 +59,7 @@ module Agama
     #   * Manages product selection.
     #   * Manages software and product related issues.
     #
-    #   It shoud be splitted in separate and smaller classes.
+    #   It should be splitted in separate and smaller classes.
     class Manager # rubocop:disable Metrics/ClassLength
       include Helpers
       include WithLocale
@@ -69,6 +69,23 @@ module Agama
 
       GPG_KEYS_GLOB = "/usr/lib/rpm/gnupg/keys/gpg-*"
       private_constant :GPG_KEYS_GLOB
+
+      # location of the custom DUD package repository,
+      # see the /usr/lib/dracut/modules.d/99agama-dud/agama-dud-apply.sh script
+      DUD_REPOSITORY_DIR = "/var/lib/agama/dud/repo"
+      private_constant :DUD_REPOSITORY_DIR
+
+      # name for the custom DUD package repository, use some special name to
+      # minimize possible conflicts with user defined repositories, ugly name
+      # does not matter, it is deleted in the end anyway
+      DUD_REPOSITORY_NAME = "AgamaDriverUpdate"
+      private_constant :DUD_REPOSITORY_NAME
+
+      # use a higher priority for the custom DUD package repository,
+      # the default priority is 99, the lower number the higher priority!
+      # the linuxrc default is 50, let's use the same value here as well
+      DUD_REPOSITORY_PRIORITY = 50
+      private_constant :DUD_REPOSITORY_PRIORITY
 
       # Selected product.
       #
@@ -223,6 +240,7 @@ module Agama
       def finish
         # disable local repositories (DVD, USB flash...)
         disable_local_repos
+        remove_dud_repo
         Yast::Pkg.SourceSaveAll
         Yast::Pkg.TargetFinish
         # copy the libzypp caches to the target
@@ -516,6 +534,7 @@ module Agama
       end
 
       def add_base_repos
+        add_dud_repo
         return if add_repos_by_label
         return if add_repos_by_dir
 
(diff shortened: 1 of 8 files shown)


service/lib/agama/cmdline_args.rb
```diff
@@ -51,20 +53,11 @@ def self.read_from(path)
         next unless option.start_with?(CMDLINE_PREFIX)
 
         key, value = option.split("=", 2)
-        key.gsub!(CMDLINE_PREFIX, "")
```
lslezak, line 54: Ouch, this could potentially delete more than expected! :scream: 

```
irb(main):001:0> "inst.rage.against.the.machine".gsub("inst.", "")
=> "rage.agathe.machine"
```

live/live-root/usr/lib/dracut/modules.d/99agama-dud/agama-dud-parser.sh
```diff
@@ -10,6 +10,18 @@ fetch_updates() {
...
     echo "$updates" >/tmp/agamadud.info
+
+    # automatically configure network when a remote DUD is used,
+    # check for all remote URL scheme supported by the "agama download"
+    # (anything supported by the curl backend https://everything.curl.dev/protocols/curl.html),
+    # because it can match everywhere in the text it also matches the tftp:// and sftp://
+    if grep -q -s -E -e "https?://" -e "ftps?://" -e "smbs?://" -e "scp://" /tmp/agamadud.info; then
+      echo "rd.neednet=1" > /etc/cmdline.d/agama-generated.conf
+      # use DHCP if there is no network configuration provided by user
+      if ! getarg "ip="; then
```
imobachgs, line 21: 👍🏻

service/lib/agama/software/callbacks/pkg_gpg_check.rb
```diff
@@ -0,0 +1,98 @@
...
+            Yast::FunRef.new(method(:pkg_gpg_check), "string(map)")
+          )
+        end
+
+        # Should be the DUD packages GPG signatures verified? The GPG errors can
+        # be ignored by using the "inst.dud_packages.gpg=0" boot option
+        #
+        # @return [Boolean] `true` if the GPG errors should be ignored, `false` otherwise
+        def ignore_dud_packages_gpg_errors?
+          return @ignore_dud_packages_gpg_errors if !@ignore_dud_packages_gpg_errors.nil?
```
imobachgs: np: I would go with `unless`

service/lib/agama/software/callbacks/pkg_gpg_check.rb
```diff
@@ -0,0 +1,98 @@
...
+require "agama/software/manager"
+require "agama/software/callbacks/base"
+
+Yast.import "Pkg"
+
+module Agama
+  module Software
+    module Callbacks
+      # Provide callbacks
+      class PkgGpgCheck < Base
```
imobachgs, line 35: Please, add some unit tests as we have for other callbacks.

service/lib/agama/software/manager.rb
```diff
@@ -70,6 +70,23 @@ class Manager # rubocop:disable Metrics/ClassLength
...
       private_constant :GPG_KEYS_GLOB
 
+      # location of the custom DUD package repository,
+      # see the /usr/lib/dracut/modules.d/99agama-dud/agama-dud-apply.sh script
+      DUD_REPOSITORY_DIR = "/var/lib/agama/dud/repo"
+      private_constant :DUD_REPOSITORY_DIR
+
+      # name for the custom DUD package repository, use some special name to
+      # minimize possible conflicts with user defined repositories, ugly name
+      # does not matter, it is deleted in the end anyway
```
imobachgs, line 80: np: I think it is good enough. But we could even consider generating a random identifier (`agama-3a9d0ca` or whatever).

imobachgs (commented):
It looks almost good to me. Thanks!

live/live-root/usr/lib/dracut/modules.d/99agama-dud/agama-dud-parser.sh
lslezak: Ah, actually stolen from [kiwi](https://github.com/OSInside/kiwi/blob/main/dracut/modules.d/90kiwi-dump/parse-kiwi-install.sh#L11)... :stuck_out_tongue:

service/lib/agama/software/manager.rb
lslezak: Pkg-bindings can [already do that](https://github.com/yast/yast-pkg-bindings/blob/853496f527543e6d51730fd7e3126ad94b13c303/src/Source_Create.cc#L412-L419) if the alias is not specified or is empty.

But personally I'd rather avoid any randomness in the code, that can make things worse to reproduce. And it is easier to grep the logs if you know what you are looking for.

As we use only a single DUD repository even for multiple DUDs there is basically no risk for conflicts. (Unless some user really wants to shoot his leg and deliberately uses the same reserved name.)

imobachgs (approved):

PR #2547, opened 2025-07-09
Changes on the storage-selector branch

ancorgs:
Includes:

- Merging master (there were relevant aesthetic changes there)
- Changes in the behavior, labels and texts when defining partitions and space policy
- Removed the now useless `useLongestDiskTitle`
- Redefine the fields displayed on the device selector (hopefully @dgdavid can improve the look & feel).

This is how it looks without mocking in a system where the model of the disks can be read

real

This is mocked with some MD RAIDs

mock_raid

This is mocked with multipath (long names)

mock_multipath


web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -172,10 +170,13 @@ const ChangeDeviceMenuItem = ({
   device,
   ...props
 }: ChangeDeviceMenuItemProps): React.ReactNode => {
+  const onlyOneOption = UseOnlyOneOption(modelDevice);
```
dgdavid, line 173: NP: not part of your changes and probably simply a typing mistake, but we usually use PascalCase for components, not for hooks. I.e., should be `useOnlyOneOption` or not use `use` prefix if it is not a hook.

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -208,6 +209,10 @@ const RemoveEntryOption = ({ device, onClick }: RemoveEntryOptionProps): React.R
     return !onlyToBoot;
   };
 
+  // If the target device cannot be changed, this button will always be disabled and would only
+  // provide redundant information.
+  if (UseOnlyOneOption(device)) return;
```
dgdavid, line 214: Ah! looks like `UseOnlyOneOption` is not a hook. So, we need better name for it **plus* go for camelCase since it is not a component either. Again, I'm completely aware that this is not part of your work here.

dgdavid (commented):
I have done a first review of your commits only. Everything looks good, but I'm postponing the approval for a final review when the PR is actually marked ready for review.

PR #2550, opened 2025-07-10
Fix base product names

lslezak:
## Problem

- The product definitions contain some incorrect product names
  - The openSUSE Leap product has been renamed from `openSUSE` to `Leap` (since Leap 15.3)
  - Similarly for LeapMicro there is `Leap-Micro` product
  - And there actually is `Kalpa` product in the Tumbleweed OSS repository, so let's use it

## Solution

- Fixed the installed base product names for Kalpa, Leap 16.0 and Leap-Micro 6.2 distributions.

### DIFF
--- a/products.d/kalpa.yaml
+++ b/products.d/kalpa.yaml
@@ -37,7 +37,7 @@ software:
     - NetworkManager
     - openSUSE-repos-MicroOS
   optional_packages: null
-  base_product: MicroOS
+  base_product: Kalpa
 
 security:
   lsm: selinux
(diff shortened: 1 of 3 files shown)


lkocman:
@lslezak how are these used and what exactly would you say is the source of truth for these base product names or what dot you validate them against? Shall it be openSUSE Leap or just Leap.

Invididual product names are currently inconsistent some use openSUSE some don't, so let's have these products consistent at least.

lkocman (approved):
Matches with release package. Thank you @lslezak!

Provides:       product() = Leap
Provides:       product(Leap) = 16.0-1

PR #2551, opened 2025-07-10
feat: support for unattended zFCP configuration

jreidinger:
## Problem

There is missing support to activate zFCP devices in unattended agama.

## Solution

Implement it and also together with it fix also wrong dasd export at https://bugzilla.suse.com/show_bug.cgi?id=1246102

## Testing

- *Added a new unit test*
- *Tested manually*

Manually tested with following profile ( first select product and then test it using `agama config load`. We cannot test it with full unatttended profile as our only s390 machine with zfcp available is already installed, so we tested it on installed system )

```json
{
  "zfcp": {
    "devices": [
      {
        "channel": "0.0.fa00",
        "wwpn": "0x500507630b181216",
        "lun": "0x4020404900000000"
      }
    ]
  }
}

```

## Documentation

related PR https://github.com/agama-project/agama-project.github.io/pull/96


rust/agama-lib/share/profile.schema.json
```diff
@@ -119,6 +119,34 @@
         }
       }
     },
+    "zfcp": {
+      "title": "zFCP device activation (s390x only)",
+      "type": "object",
+      "properties": {
+        "devices": {
+          "title": "List of DASD devices",
```
```suggestion
          "title": "List of zFCP devices",
```

rust/agama-lib/src/install_settings.rs
```diff
@@ -96,6 +97,9 @@ pub struct InstallSettings {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub scripts: Option<ScriptsConfig>,
+    #[serde(default)]
```
imobachgs: I am not sure whether the `default` is needed for optional field. I know, we are using it all over the place, but...

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -217,4 +218,23 @@ impl ZFCPClient<'_> {
             Err(ServiceError::UnsuccessfulAction(text))
         }
     }
+
+    pub async fn set_config(&self, config: ZFCPConfig) -> Result<(), ServiceError> {
```
imobachgs: We need to take a decision whether we should use `ZFCPConfig` or `&ZFCPConfig` to be consistent. Checking the function's code it looks like a ref should be enough.

rust/agama-lib/src/storage/http_client/dasd.rs
```diff
@@ -41,10 +41,24 @@ impl DASDHTTPClient {
...
     pub async fn set_config(&self, config: &DASDConfig) -> Result<(), DASDHTTPClientError> {
+        if !self.supported().await? {
+            // TODO: should we add tracing error here?
+            return Ok(());
+        }
         Ok(self.client.put_void("/storage/dasd/config", config).await?)
     }
+
+    pub async fn supported(&self) -> Result<bool, DASDHTTPClientError> {
+        Ok(self.client.get("/storage/dasd/supported").await?)
```
imobachgs, line 62: Perhaps it might be good to have a `/storage/supported` or `/storage/features` that returns which features are actually supported in a single place. But I guess it is out of scope here.

rust/agama-lib/src/storage/http_client/zfcp.rs
```diff
@@ -0,0 +1,64 @@
...
+            Ok(None)
+        } else {
+            Ok(Some(config))
+        }
+    }
+
+    pub async fn set_config(&self, config: &ZFCPConfig) -> Result<(), ZFCPHTTPClientError> {
+        if !self.supported().await? {
+            // TODO: should we add tracing error here?
+            return Ok(());
```
imobachgs, line 56: By now I would most probably add a warning.

jreidinger: question is if we should add `tracing` to agama-lib as currently it is only in agama-server and agama-network

rust/agama-lib/src/storage/http_client/dasd.rs
jreidinger: makes sense for me, but I would do it in separate PR as it affects also web UI

rust/agama-lib/src/install_settings.rs
jreidinger: so lets remove it

PR #2555, opened 2025-07-12
fix(web): improve UX around non-persistent connections

dgdavid:
As per feedback received in bug report https://bugzilla.suse.com/show_bug.cgi?id=1245187 (protected link) , this PR aims to polish the user experience when dealing with non-persistent network connections during installation setup:

* Adjusted alert severity
  The alert about no persistent connections is now rendered using the custom variant instead of warning. This change reflects its non-blocking nature and avoids confusing users by presenting it with the same prominence as critical issues.

* Renders such an alert in the connection details page too
  When toggling a connection to "installation-only" results in no connections being persisted, the alert is now shown directly in the connection details page. This provides immediate feedback about the side effect, helping users avoid unintentionally ending up with no configured network after installation, especially since this information is not displayed elsewhere.

The second change is somewhat controversial, as **somehow it goes in a different direction from the first, alerting users too early about a non-critical issue**. While it does help prevent users from unintentionally misconfiguring their system, ideally this kind of feedback should be shown later in the flow, such as in the issues drawer close to the critical ones or in the confirmation dialog, the only point where Agama can be certain the user has finished setting up the installation. This would help avoid interrupting the setup experience while ensuring users has all needed information before proceed with installation. However, due to current limitations in the confirmation dialog and the absent of feedback for non-critical issues, this change serves as a **reasonable temporary mitigation** to prevent users from overlooking a side effect introduced by a recently added feature.

---

| Before | After |
|-|-|
| localhost_8080_ (10) |  localhost_8080_ (9) |
|  localhost_8080_ (11) | localhost_8080_ (12) |


web/src/components/network/NoPersistentConnectionsAlert.tsx
```diff
@@ -36,7 +37,7 @@ export default function NoPersistentConnectionsAlert() {
   if (persistentConnections !== 0) return;
 
   return (
-    <Alert variant="warning" title={_("Installed system may not have network connections")}>
+    <Alert variant="custom" title={_("Installed system may not have network connections")}>
       {_(
         "All network connections managed through this interface are currently set to be \
```
dgdavid, line 42: @teclator is that sentence right?

While testing changes manually I found a bit confusing the "through this interface" part. What do you think?

shundhammer: NP: "**The** installed system"

dgdavid: Good catch!

This actually reminds me of something I mentioned to @ancorgs a while ago. During my research time, I spent some time looking into the GOV.UK Design System, which I found quite practical and insightful. I think there’s a lot we can potentially apply to Agama from their approach, where it makes sense.

Although it’s not directly related to your NP, I came across a point in their service manual (https://www.gov.uk/service-manual/design/writing-for-user-interfaces) that I found particularly relevant:

>    Don’t follow strict grammar conventions if it makes things clearer not to.

Think that they aims for targeting people from 9 years old (https://www.gov.uk/guidance/content-design/writing-for-gov-uk#reading-skills)

teclator (approved):
LGTM

web/src/components/network/NoPersistentConnectionsAlert.tsx
teclator: The point was, that there are connections not managed by Agama at all and that are probably not permanent, that is why we worded it in that way....

dgdavid: I know the point, but which is "this interface"?

dgdavid: Ah, it means "this user interface". Let's keep it as it is and improve it later along other changes.

PR #2559, opened 2025-07-14
feat(storage): directly format devices

joseivanlopez:
## Goal

This pull request enables users of the UI to indicate they want to directly mount a whole disk or a software-defined MD RAID without a partition table. Possibly formatting it in the process.

It can work with software-defined MD RAIDs that are already present in the system, this does not include the ability to create new MD RAIDs using the web UI.

## Screenshots

This is how the whole process looks, step by step.

uno

dos

tres

cuatro

### DIFF
--- a/web/src/components/storage/ConfigureDeviceMenu.tsx
+++ b/web/src/components/storage/ConfigureDeviceMenu.tsx
@@ -24,7 +24,7 @@ import React, { useState } from "react";
 import { useNavigate } from "react-router-dom";
 import MenuButton, { MenuButtonItem } from "~/components/core/MenuButton";
 import { Divider, MenuItemProps } from "@patternfly/react-core";
-import { useCandidateDevices } from "~/hooks/storage/system";
+import { useAvailableDevices } from "~/hooks/storage/system";
 import { useModel } from "~/hooks/storage/model";
 import { useAddDrive } from "~/hooks/storage/drive";
 import { useAddReusedMdRaid } from "~/hooks/storage/md-raid";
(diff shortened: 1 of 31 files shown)


web/src/components/storage/ConfigureDeviceMenu.tsx
```diff
@@ -87,12 +116,6 @@ const AddDeviceMenuItem = ({
 /**
  * Menu that provides options for users to configure storage drives
  *
- * It uses a drilled-down menu approach for disks, making the available options less
```
dgdavid, line 90: Good catch :)

web/src/components/storage/FilesystemMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { sprintf } from "sprintf-js";
+import { _ } from "~/i18n";
+import { filesystemType, formattedPath } from "~/components/storage/utils";
+
+function deviceDescription(deviceModel: FilesystemMenuProps["deviceModel"]): string {
+  const fs = filesystemType(deviceModel.filesystem);
+  const mountPath = deviceModel.mountPath;
+  const reuse = deviceModel.filesystem.reuse;
+
+  // I don't think this can happen, maybe when loading a configuration not created with the UI
```
dgdavid, line 41: NP:


```suggestion
  // Unlikely to happen, but possible specially when loading a configuration not created with the UI
```

web/src/components/storage/LvmPage.tsx
```diff
@@ -53,24 +52,19 @@ import { _ } from "~/i18n";
 /**
  * Hook that returns the devices that can be selected as target to automatically create LVM PVs.
  *
- * FIXME: temporary and weak implementation that relies on the current model to offer only the
```
dgdavid, line 56: Nice to start paying debt :)

dgdavid (commented):
LGTM, thanks guys 👍

web/src/components/storage/FilesystemMenu.tsx
ancorgs: The point is that I literally think is not possible, even by loading a configuration. So I really wanted to be quite assertive here.

PR #2560, opened 2025-07-14
fix(ruby): LSM pre-selection does not overwrite the list of patterns

imobachgs:
## Problem

We have been reported that using auto-installation Agama does not honor the list of patterns included in the profile. This behavior is related to the recent feature we introduced to disable SELinux (see #2540).

## Solution

Change the approach to select the default LSM only after selecting the product instead of during the "probing" phase.

## Testing

- Added a new unit test
- Tested manually

## Pre-selecting patterns

This PR introduces a mechanism to determine whether a user pattern should be enabled by default:

```yaml
user_patterns:
  - cockpit
  - name: selinux
    selected: true
  - fips
  - documentation
```


teclator (approved):
LGTM

service/lib/agama/software/manager.rb
```diff
@@ -282,14 +287,17 @@ def patterns(filtered)
...
             # the pattern is not from a base repository or is included in the display list
-            !base_repos.include?(p.source) || product.user_patterns.include?(p.name)
+            !base_repos.include?(p.source) || user_patterns_names.include?(p.name)
           end
         end
 
         patterns
       end
+      # rubocop:enable Metrics/CyclomaticComplexity
+      # rubocop:enable Metrics/PerceivedComplexity
```
jreidinger, line 300: well, not so happy about this, but as we are close to release, lets not refactor code.

service/lib/agama/software/product.rb
```diff
@@ -84,12 +87,14 @@ class Product
 
       # Optional patterns.
       #
+      # These patterns are always installed if they are available.
```
jreidinger, line 90: :+1:

PR #2562, opened 2025-07-15
Display only Tumblweed based distros on Tumbleweed

lkocman:
- Similarly display only SLFO based distros on Leap

### DIFF
--- a/products.d/agama-products.spec
+++ b/products.d/agama-products.spec
@@ -41,6 +41,20 @@ env \
   datadir=%{_datadir} \
   %{_builddir}/agama/install.sh
 
+
+# Keep only Leap based distros on Leap
+%if 0%{?is_opensuse} && 0%{?suse_version} == 1600
+rm -f %{buildroot}%{_datadir}/agama/products.d/kalpa.yaml
+rm -f %{buildroot}%{_datadir}/agama/products.d/microos.yaml
+rm -f %{buildroot}%{_datadir}/agama/products.d/tumbleweed.yaml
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

products.d/agama-products.spec
```diff
@@ -52,12 +66,16 @@ Definition of openSUSE products (Tumbleweed, Leap, MicroOS and Slowroll) for the
...
 %dir %{_datadir}/agama/products.d
+%if 0%{?is_opensuse} && 0%{?suse_version} > 1600
 %{_datadir}/agama/products.d/microos.yaml
 %{_datadir}/agama/products.d/tumbleweed.yaml
-%{_datadir}/agama/products.d/leap_160.yaml
-%{_datadir}/agama/products.d/leap_micro_62.yaml
 %{_datadir}/agama/products.d/slowroll.yaml
 %{_datadir}/agama/products.d/kalpa.yaml
+%endif
+%if 0%{?is_opensuse} && 0%{?suse_version} >= 1600
```
jreidinger, line 75: this cause failure at https://build.opensuse.org/package/live_build_log/systemsmanagement:Agama:Devel/agama-products/openSUSE_Tumbleweed/x86_64

basically above it removes leap for > 1600, but here it is included. it should be `%if 0%{?is_opensuse} && 0%{?suse_version} > 1600`

imobachgs: I would say it should be `== 1600`.

jreidinger: ah, right...my fault. I invert condition

PR #2565, opened 2025-07-15
Add reprobe button

jreidinger:
## Problem

There is no way how to reprobe storage devices when new device is attached or activated outside of agama.

On top of that, if you skip the opportunity to provide the password for an encrypted device, then you never have a second change.

trello: https://trello.com/c/W9FyBIHU/5092-1-reprobing-button

## Solution

Add the reprobe button to the "Other options" menu.

## Screenshots

rescan

### DIFF
--- a/web/src/components/core/ProgressReport.tsx
+++ b/web/src/components/core/ProgressReport.tsx
@@ -116,7 +116,6 @@ function findDetail(progresses: ProgressType[]) {
  */
 function ProgressReport({ title, firstStep }: { title: string; firstStep?: React.ReactNode }) {
   useResetProgress();
-  // TODO: add a way to retrieve the right progress for rescan.
   const progress = useProgress("manager", { suspense: true });
   const [steps, setSteps] = useState(progress.steps);
   const softwareProgress = useProgress("software");


web/src/components/storage/ConfigEditorMenu.tsx
```diff
@@ -69,6 +70,13 @@ export default function ConfigEditorMenu() {
         >
           {_("Change boot options")}
         </DropdownItem>
+        <DropdownItem
+          key="reprobe-link"
+          onClick={() => reprobe()}
+          description={_("Reprobe available devices. Useful when activating them outside of agama")}
```
dgdavid: Remember that we’re not mentioning the project name anywhere, so I’d suggest reworking that explanation

jreidinger: Here is what AI suggest to me:

```
Here are some good names for your menu entry, along with their descriptions, keeping your requirements in mind:

Menu Entry Name Options
Option 1: Detect New Devices

Description: Scans the system for newly appeared devices that can be activated or attached.

Option 2: Rescan for Devices

Description: Initiates a scan to detect and make available new devices.

Option 3: Find New Devices

Description: Searches the system for recently connected devices, making them available for use.

I've focused on names that clearly convey the action of finding or rescanning for devices, without using "Agama" or "Reprobe." Let me know if you'd like more options!
```

ancorgs: Just for the record, the YaST Partitioner used "Rescan devices"

jreidinger: Ok, after a bit more conversation with AI, I think this is good proposal that sounds good to me, but I am not storage expert, so feel free to comment it:

```
Menu Entry: Rescan Devices
Description: Searches the system for newly connected or recently activated devices, making them available for use.
```

dgdavid: Cool, but maybe too long description. What about something shorter like

Detects newly connected or activated devices

In any case, @ancorgs has more wording skills than me, so whatever you both choose LGTM.

ancorgs: In general I like latest proposal from @dgdavid. But the other entries uses imperative, so it would be.

Rescan Devices
Detect newly connected or activated devices

jreidinger: thanks, so lets update it

ancorgs: Now I like the wording... but I should also have mentioned that I would place the new option at the very end of the menu - in the second "logical block" of the menu. After all the actions to configure iSCSI, DASD, etc.

I feel the new action is more connected to that part (adding / removing devices) than to the first block of the menu (playing with the configuration).

ancorgs: Just for the record, in the YaST partitioner all those options like "Rescan", Configure iSCSI" or "Configure DASD" are in the dedicated menu called "System". Maybe we could consider a drill-down for all that (or even a separate button, which I consider too much). @dgdavid What do you think?

ancorgs: To clarify my previous comment, I mean:

- Other options
  - Change boot options
  - Reset to defaults
  - System
    - Configure iSCSI
    - [Configure DASD]
    - [Configure zFCP]
    - Rescan Devices

Just a wild idea.

ancorgs:
Another question. If I omitted introducing an encryption password during the original probing. Will this new option ask me the password again?

web/src/components/storage/ConfigEditorMenu.tsx
ancorgs: Or maybe:

- Other options
  - Change boot options
  - Reset to defaults
  - Configure system
    - iSCSI
    - [DASD]
    - [zFCP]
    - Rescan Devices

dgdavid: Actually no strong opinion to that regard yet, but I'd keep as it is by now. It's more straightforward and these options are clearly divided in two logical blocks, as you commented before. Additionally, we have to make possible adding the headers to these blocks, for which PatternFly has support.

dgdavid: I mean, I do not see a clear advantage of hiding them in a submenu. It even defeats a bit the discoverability of these actions for not advantage in exchange. So, since that "Other options" menu is not that long, I feel is more straight forward and even better for users to keep it in just one level.

jreidinger: For me it is already a bit hidden to find it. If we add submenu, it will be even more hidden

dgdavid:
I've sent a commit that changes the progress interface approach by using the plain layout to somehow block user interaction while _reprobing_ is being done. Essentially, the screenshots in the PR description will looks the same but with below layout


localhost_8080_ (15)

web/src/components/storage/ConfigEditorMenu.tsx
```diff
@@ -102,6 +104,13 @@ export default function ConfigEditorMenu() {
             {_("Configure DASD")}
           </DropdownItem>
         )}
+        <DropdownItem
+          key="reprobe-link"
+          onClick={reactivate}
+          description={_("Update the list of available disks and activate crypt devices")}
```
dgdavid: NP: I wonder if something three words shorter would work


```suggestion
          description={_("Update available disks and activate crypt devices")}
```

web/src/components/core/ProgressReport.tsx
```diff
@@ -116,6 +116,7 @@ function findDetail(progresses: ProgressType[]) {
  */
 function ProgressReport({ title, firstStep }: { title: string; firstStep?: React.ReactNode }) {
   useResetProgress();
+  // TODO: add a way to retrieve the right progress for rescan.
```
dgdavid:
```suggestion
```

dgdavid (approved):
Just two minor suggestions, one for deleting a comment added by myself 🙈 

LGTM, thanks!

PR #2567, opened 2025-07-16
(backend): Set connection method4 and method6 as auto in case of absense in the profile (bsc#1245798)

teclator:
## Problem

The default for **IPv4** method and **IPv6** method is to have it disabled. But the defaults for NetworkManager and what most users would expect is to set it to auto.

- https://bugzilla.suse.com/show_bug.cgi?id=1245798

## Solution

Set the default to auto for IPv6 and IPv4 methods.

## Testing

- *Tested manually*

## Example

**profile.jsonnet**
```jsonnet
{
  "network": {
    "connections": [
      {
        "id": "Bridge0",
        "interface": "br0",
        "bridge": {
          "ports": ["enp5s0f4u1u1c2"],
          "stp": false
        }
      }
    ]
  }
}
```

```bash
agama config load < profile.jsonnet
agama config show 
{
  ...
  "network": {
    "connections": [
      {
        "id": "Bridge0",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "interface": "br0",
        "bridge": {
          "stp": false,
          "ports": [
            "enp5s0f4u1u1c2"
          ]
        },
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  }
  ...
}
✓ The profile is valid.

```

### DIFF
--- a/rust/agama-network/src/model.rs
+++ b/rust/agama-network/src/model.rs
@@ -1074,8 +1074,8 @@ pub struct UnknownIpMethod(String);
 #[derive(Debug, Default, Copy, Clone, PartialEq, Deserialize, Serialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub enum Ipv4Method {
-    #[default]
     Disabled = 0,
+    #[default]
     Auto = 1,
     Manual = 2,
     LinkLocal = 3,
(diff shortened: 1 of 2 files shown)


teclator:
> Very nice change! Thank you! I was also wondering why the default was _Disabled_.

Not sure, it was the default long time ago, but probably we were doing an unwrap and not handled correctly if not declared in the profile...

See https://github.com/agama-project/agama/commit/0c438180a99ad934d98d056b9167b78fcef40b0e

So, bringing back the default to auto makes sense.

imobachgs (approved):

PR #2568, opened 2025-07-16
feat(cli): Added -i/--insecure optional argument to the "agama download" command

lslezak:
## Problem

- The `agama download` command fails when downloading from HTTPS with a self-signed certificate
- It should be possible to ignore the problem and download the file anyway (only on explicit user request)
- Useful for testing, might be used by customers in local networks
- https://bugzilla.suse.com/show_bug.cgi?id=1245393

## Solution

- Ignore SSL problems in HTTPS downloads (self-signed certificate, non-matching host, unknown CA...)

## Testing

- Tested manually

### Fails by default 

Downloading from HTTPS server using a self-signed certificate fails:

```console
> ./agama download https://localhost:9000 index.html
Could not retrieve https://localhost:9000/

Caused by:
    [60] SSL peer certificate or SSH remote key was not OK (SSL certificate problem: self-signed certificate)
```

### Using the `--insecure` option

When using the `--insecure` option the problem is ignored and the download succeeds:

```console
> ./agama download --insecure https://localhost:9000 index.html
File saved to index.html
```

### Help

The option is described in the help output:

```console
> ./agama download --help
Download file from a given (AutoYaST) URL

The purpose of this command is to download files using AutoYaST supported schemas (e.g. device://).
It can be used to download additional scripts, configuration files and so on. You can use it for
downloading Agama autoinstallation profiles. If you want to convert an AutoYaST profile, use "agama
config generate".

Usage: agama download [OPTIONS] <URL> <DESTINATION>

Arguments:
  <URL>
          URL reference pointing to file for download. If a relative URL is provided, it will be
          resolved against the current working directory

  <DESTINATION>
          File name

Options:
  -i, --insecure
          Disables SSL verification for HTTPS downloads

  -h, --help
          Print help (see a summary with '-h')
```

### DIFF
--- a/rust/agama-lib/src/utils/transfer/handlers/generic.rs
+++ b/rust/agama-lib/src/utils/transfer/handlers/generic.rs
@@ -32,11 +32,18 @@ use crate::utils::{TransferError, TransferResult};
 pub struct GenericHandler {}
 
 impl GenericHandler {
-    pub fn get(&self, url: Url, out_fd: &mut impl Write) -> TransferResult<()> {
+    pub fn get(&self, url: Url, out_fd: &mut impl Write, insecure: bool) -> TransferResult<()> {
         let mut handle = Easy::new();
         handle.follow_location(true)?;
         handle.fail_on_error(true)?;
         handle.url(url.as_ref())?;
+
+        if insecure {
+            // allow self-signed certificate, ignore verification failures
+            handle.ssl_verify_peer(false)?;
+            // allow not matching hostname
+            handle.ssl_verify_host(false)?;
+        }
 
         let mut transfer = handle.transfer();
         transfer.write_function(|buf| Ok(out_fd.write(buf).unwrap()))?;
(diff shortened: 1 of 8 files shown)


rust/agama-lib/src/utils/transfer/handlers/generic.rs
```diff
@@ -32,12 +32,19 @@ use crate::utils::{TransferError, TransferResult};
...
-    pub fn get(&self, url: Url, out_fd: &mut impl Write) -> TransferResult<()> {
+    pub fn get(&self, url: Url, out_fd: &mut impl Write, insecure: bool) -> TransferResult<()> {
         let mut handle = Easy::new();
         handle.follow_location(true)?;
         handle.fail_on_error(true)?;
         handle.url(url.as_ref())?;
 
+        if insecure {
+            // allow self-signed certificate, ignore verification failures
+            handle.ssl_verify_peer(false)?;
```
imobachgs, line 43: np: you can do `handle.ssl_verify_peer(!insecure)?` instead.

imobachgs (approved):

rust/agama-lib/src/utils/transfer/handlers/generic.rs
lslezak: Yes, I even had this in the initial version, but later changed to this.

- Calling with `true` is pointless, that are the defaults.
- The security related block is more visible in the code.

PR #2570, opened 2025-07-16
initial attempt to ensure that product is properly selected when needed

jreidinger:
## Problem

It is possible to not specify product in agama profile and it can lead to crash in ruby code or unexpected results.

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1246601

## Solution

For parts that needs product to be already selected add check and report error if it is not the case.

## Testing

- *Tested manually*

## Screenshots

output before:

```shell
agama2:~ # agama config load < profile.jsonnet 
✓ The profile is valid.

Error processing storage settings: Backend call failed with status 400 and text '{"error":"Agama service error: D-Bus service error: org.freedesktop.DBus.Error.Failed: undefined method 'reject' for nil; caused by 1 sender=:1.1 -> dest=org.opensuse.Agama.Storage1 serial=219 reply_serial= path=/org/opensuse/Agama/Storage1; interface=org.opensuse.Agama.Storage1; member=Reprobe error_name="}'

Caused by:
    Backend call failed with status 400 and text '{"error":"Agama service error: D-Bus service error: org.freedesktop.DBus.Error.Failed: undefined method 'reject' for nil; caused by 1 sender=:1.1 -> dest=org.opensuse.Agama.Storage1 serial=219 reply_serial= path=/org/opensuse/Agama/Storage1; interface=org.opensuse.Agama.Storage1; member=Reprobe error_name="}'
```

and output with fix:

```shell
agama2:~ # agama config load < profile.jsonnet 
✓ The profile is valid.

Cannot proceed with profile without specified product
```


rust/agama-lib/src/store.rs
```diff
@@ -246,6 +266,20 @@ impl Store {
         Ok(())
     }
 
+    async fn detect_selected_product(&self) -> Result<bool, ProductStoreError> {
+        let product_client = ProductHTTPClient::new(self.http_client.clone());
+        let product = product_client.product().await?;
+        Ok(!product.is_empty())
+    }
+
+    fn check_selected_product(selected: bool) -> Result<(), StoreError> {
```
imobachgs: I would prefer `ensure_selected_product` because it communicates better the intention.

rust/agama-lib/src/store.rs
```diff
@@ -85,6 +85,8 @@ pub enum StoreError {
     ZFCP(#[from] ZFCPStoreError),
     #[error("Could not calculate the context")]
     InvalidStoreContext,
+    #[error("Cannot proceed with profile without specified product")]
+    ProductMissing,
```
imobachgs: np: `MissingProduct` (as `InvalidStoreContext`, `MissingDevice`, etc.).

rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jul 16 19:05:48 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix crash when agama profile contain only zfcp section by
```
```suggestion
- Fix crash when an Agama profile contains only the zfcp section by
```

rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
...
+  providing better error report when product selection have to be
+  done before or in same profile (bsc#1246601)
```
```suggestion
  done before or in the same profile (bsc#1246601)
```

imobachgs (approved):

PR #2571, opened 2025-07-16
fix: better handling of missing license translations

imobachgs:
## Problem

* [bsc#1238364](https://bugzilla.suse.com/show_bug.cgi?id=1238364)

When displaying a product license, Agama allows to select the language to display the license. In general, it offers the same list of languages Agama is translated to. This has a few problems:

* In some cases, the license is no translated to that language. The Agama falls back to English without informing the user.
* It is just another language selector, which is kind of confusing.

## Solution

Now the license is directly shown in the same language that Agama is using. No additional language selector. If the translation is missing:

* It tries to find an alternative language (e.g., "Spanish" for "Catalan".
* If there is no alternative, it falls back to English.

## Alternative language

To search for an alternative language, Agama choses the first language in the same "territory". For instance, if you pick "Catalan" ("ca-ES"), Agama uses the [langtable data](https://pypi.org/project/langtable/) to search for the first language in the same territory, which in this case it is "Spanish" (es).

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>
<summary>The license is translated. No warning and no selector.</summary>

license-found-license
</details>

<details>
<summary>The license is not translated to Catalan, but to Spanish. Warn the user.</summary>

license-missing-translation
</details>

<details>
<summary>Fall back to English. Warn the user.</summary>

license-fallback-english
</details>

## Known problems

* It takes a moment to refresh the page when you change the language. It may confuse QA but it is a known problem, not related to this PR.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -54,6 +54,7 @@ dependencies = [
 name = "agama-lib"
 version = "1.0.0"
 dependencies = [
+ "agama-locale-data",
  "agama-network",
  "agama-utils",
  "anyhow",
--- a/rust/agama-lib/Cargo.toml
+++ b/rust/agama-lib/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 anyhow = "1.0"
 agama-utils = { path = "../agama-utils" }
 agama-network = { path = "../agama-network" }
+agama-locale-data = { path = "../agama-locale-data" }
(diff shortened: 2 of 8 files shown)


dgdavid (changes requested):
Arg! We forgot the tests 🙄

dgdavid (approved):

PR #2575, opened 2025-07-17
fix: improve the logging information

imobachgs:
Agama logs are growing on every release. That's perfectly fine, but sometimes we are:

- Logging too much (locales),
- Logging with the wrong (or not consistent) level.
- Just dumping Rust structs. This is fine, but sometimes it might be better to use a JSON representation (like when dealing with events).


joseivanlopez (approved):
LGTM

Do we need changelog?

joseivanlopez (approved):

PR #2576, opened 2025-07-17
fix(web): encode URL params for safer URL generation

dgdavid:
## Problem

Bug report https://bugzilla.suse.com/show_bug.cgi?id=1246551 (private link) revealed that some URL parameters were not being properly encoded. This led to issues where users could not access certain resources when special characters (e.g., #, $, spaces) were present in the URL path.

## Solution

Introduced a new utility function, `generateEncodedPath`, which wraps React Router's `generatePath` and ensures all path parameters are safely URI-encoded using [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).

Replaced all occurrences of `generatePath` in the codebase with `generateEncodedPath`, adapting parameters where needed to ensure compatibility.

## Testing

* Updated and adapted existing unit tests to validate encoding behavior.
* Performed manual testing to verify that URLs with special characters are now correctly handled and functional.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 17 11:26:17 UTC 2025 - David Diaz <dgonzalez@suse.com>
+
+- Replaced all usage of generatePath with generateEncodedPath to
+  ensure proper URL encoding of route parameters and prevent issues
+  with special characters (bsc#1246551).
+
+-------------------------------------------------------------------
 Wed Jul 16 15:39:14 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Display a message when the license is not available in the
--- a/web/src/api/network.ts
+++ b/web/src/api/network.ts
@@ -37,7 +37,7 @@ const fetchDevices = (): Promise<APIDevice[]> => get("/api/network/devices");
  * Returns data for given connection name
  */
 const fetchConnection = (name: string): Promise<APIConnection> =>
-  get(`/api/network/connections/${name}`);
+  get(`/api/network/connections/${encodeURIComponent(name)}`);
 
 /**
  * Returns the list of known connections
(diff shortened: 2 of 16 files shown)


dgdavid:
> Sadly, problem looks bigger than just an issue with routing. These id's has to be encoded to when doing request to the API

Having a look to the code, only the network part looks problematic at src/api level


```
❯ grep "\${" src/api/**.* 
src/api/issues.ts:const fetchIssues = (scope: IssuesScope): Promise<Issue[]> => get(`/api/${URLS[scope]}`);
src/api/network.ts:  get(`/api/network/connections/${name}`);
src/api/network.ts:  put(`/api/network/connections/${connection.id}`, connection);
src/api/network.ts:const deleteConnection = (name: string) => del(`/api/network/connections/${name}`);
src/api/network.ts:const connect = (name: string) => post(`/api/network/connections/${name}/connect`);
src/api/network.ts:const disconnect = (name: string) => post(`/api/network/connections/${name}/disconnect`);
src/api/progress.ts:  const progress: APIProgress = await get(`/api/${service}/progress`);
src/api/questions.ts:  await put(`/api/questions/${question.id}/answer`, answer);
src/api/software.ts:  get(`/api/software/licenses/${id}?lang=${lang}`);
src/api/storage.ts:  return get(`/api/storage/config_model/solve?model=${serializedModel}`);
src/api/storage.ts:  return get(`/api/storage/product/volume_for?mount_path=${path}`);
```

@teclator, I will encode network calls to avoid below

Screenshot From 2025-07-17 12-44-52



@ancorgs, @joseivanlopez, please confirm me that `mount_path` will not have special characters that needs to be encoded.

joseivanlopez (approved):
LGTM

dgdavid:
Now everything seems work as expected in network pages. Tested manually. @teclator please double check it.

teclator (approved):
LGTM, tested manually

PR #2579, opened 2025-07-17
fix(live): disable storage services

joseivanlopez:
## Problem

Some services interfers with libstorage-ng activation.

https://bugzilla.suse.com/show_bug.cgi?id=1246133

## Solution

~~Disable services.~~ Mask the services. Disabling them was not enough.

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul 17 21:51:19 UTC 2025 - Ancor Gonzalez Sosa <ancor@suse.com>
+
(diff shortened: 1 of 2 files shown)


ancorgs:
Looks like I finally found the option that works: masking (not disabling) `lvm2-monitor.service` and `lvm-activate-system.service`.

I'm pretty sure we don't need the because libstorage-ng will take care of activating LVM. I'm also pretty sure that disabling them is not enough.

So I modified this pull request according to those findings.

ancorgs (approved):
I manually checked this and works as expected. I was able to complete the installation in a multipath system with a previous LVM and everything was correctly detected and later installed.

PR #2580, opened 2025-07-17
Added VLAN CLI support

teclator:
## Problem

The VLAN model has been added to the backend but it is not supported by the CLI or the HTTP API.

- https://trello.com/c/U3TS6jy3/5099-vlan-unattended

## Solution

Added the VlanSettings and extend the schema supporting VLAN configuration.

## Testing

- *Tested manually*


mvidner:
> Tested manually

Not to pick on you, but I like to make this kind of statement more informative, to help reviewers who may not be domain experts:
If the testing needs some infrastructure, what tools did you use to set it up? Pasteable commands are awesome.

rust/agama-lib/share/profile.schema.json
```diff
@@ -673,6 +673,29 @@
...
+                "required": ["id"],
+                "properties": {
+                  "id": {
+                    "title": "VLAN Identifier",
+                    "type": "integer",
+                    "minimum": 0,
+                    "maximum": 4095
+                  },
+                  "parent": {
+                    "title": "VLAN protocol for encapsulation",
```
mvidner: Pasting mistake. BTW I tried to find if there is an "official" term for "parent" but failed to find one.

```suggestion
                    "title": "Parent interface",
```

rust/agama-network/src/model.rs
```diff
@@ -1289,7 +1337,7 @@ impl TryFrom<WirelessSettings> for WirelessConfig {
         let bssid = if let Some(bssid) = &settings.bssid {
             Some(
                 macaddr::MacAddr6::from_str(bssid)
-                    .map_err(|_| NetworkStateError::InvalidBssid(bssid.to_string()))?,
+                    .map_err(|_| NetworkStateError::InvalidVlanProtocol(bssid.to_string()))?,
```
mvidner: What?!

rust/agama-lib/share/profile.schema.json
```diff
@@ -673,6 +673,29 @@
                     "type": "boolean"
                   }
                 }
+              },
+              "vlan": {
+                "type": "object",
+                "title": "VLAN configuration",
+                "additionalProperties": false,
+                "required": ["id"],
```
mvidner:
```suggestion
                "required": ["id", "parent"],
```

also, what does it mean if I omit the `protocol` property? Please consider adding a [`default` annotation](https://json-schema.org/understanding-json-schema/reference/annotations) to it

rust/agama-lib/share/profile.schema.json
teclator: I was also thinking in name it different for the profile attribute, but failed in the same way so kept the NM name.

rust/agama-network/src/model.rs
teclator: uh, that was introduced by error :(, thanks for catching

rust/agama-lib/share/profile.schema.json
mvidner: NM using the term is a good reason

rust/agama-lib/share/profile.schema.json
mvidner: OK, so `protocol` is optional and we have documented the default,
But `parent` is IMHO required.

teclator: At least in this implementation we should require it.. it is not in the backend, but in that case the connection should have the mac-address set... so, by now it is preferable to have it as mandatory

teclator: From https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html:

> parent	string	 	If given, specifies the parent interface name or parent connection UUID from which this VLAN interface should be created. If this property is not specified, the connection must contain an "802-3-ethernet" setting with a "mac-address" property.

PR #2581, opened 2025-07-18
use more generic fix to disable lvm auto assembly

jreidinger:
## Problem

LVM with name different to system is still assembled and blocking  libstorage-ng actions.

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1246133#c38

## Solution

Try to disable udev lvm assembly in `/etc/lvm/lvm.conf`.

## Testing

- *Added a new unit test*
- *Tested manually*

Tested manually and current devel iso contain some issues with products so hard to test, but still I see some EBusy errors in journalctl. But lvm-activate systemd unit is not generated.

### DIFF
--- a/service/lib/agama/software/manager.rb
+++ b/service/lib/agama/software/manager.rb
@@ -148,15 +148,12 @@ module Agama
 
         raise ArgumentError unless new_product
 
-        # order is important here as we want to first set internally new product to ensure that cache read will read it properly
-        @product = new_product
-
-        update_repositories(new_product)
-
-        # here it will reset config cache, so do it last to have all stuff in place
         proposal.set_resolvables(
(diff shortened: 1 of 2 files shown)


service/lib/agama/software/manager.rb
```diff
@@ -148,12 +148,15 @@ def select_product(id)
 
         raise ArgumentError unless new_product
 
+        # order is important here as we want to first set internally new product to ensure that cache read will read it properly
+        @product = new_product
+
+        update_repositories(new_product)
+
+        # here it will reset config cache, so do it last to have all stuff in place
```
jreidinger: I am not sure why, but even with ordering change it still sometimes does not work and looks like it can still create race conditions when it cache empty product.

ancorgs:
Closed in favor of #2584

PR #2582, opened 2025-07-18
nm/client: fix deactivate_connection() race

cfconrad:
## Problem

There was a potential race condition where the ActiveConnection could be removed directly after the call to `nm_proxy.active_connections()`. 

## Solution

This situation should not throw an error but instead return `Ok(None)`, as it's not fatal and the caller can handle it gracefully.

It's seems to be similar to the error handling here https://github.com/agama-project/agama/blob/9d76c6780e8a86d37a326f6c54270ed15154992a/rust/agama-network/src/nm/client.rs#L395 but one step before.

## Testing

- *Tested manually*

- The issue was discovered by one out of 10 runs on migrating an OvsBridge from wicked to Networkmanager https://github.com/openSUSE/wicked2nm/pull/33
  To reproduce, one can run this inside the `wicked2nm` repo:
  ```bash
  cd tests;
  cnt=0; err=0; 
  while true; do 
    ((cnt++)); 
    ./test.sh --nm-cleanup ovs-bridge2.1 || ((err++)); 
    echo "cnt:$cnt err:$err"; 
    sleep 2; 
  done
  ```

### DIFF
--- a/rust/agama-network/src/nm/client.rs
+++ b/rust/agama-network/src/nm/client.rs
@@ -411,6 +411,18 @@ impl<'a> NetworkManagerClient<'a> {
         Ok(proxy)
     }
 
+    /// Retrieves the `org.freedesktop.NetworkManager.ActiveConnection` object
+    /// path corresponding to a given `org.freedesktop.NetworkManager.Settings` object path.
+    ///
+    /// # Arguments
+    ///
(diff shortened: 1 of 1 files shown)


teclator (changes requested):
LGTM, but please, add the changelog entry.

PR #2587, opened 2025-07-21
Fixed crash when registering an extension without registration key

lslezak:
## Problem

- Registering the HA extension from an RMT server crashes the web UI
- https://bugzilla.suse.com/show_bug.cgi?id=1246690

agama-rmt-ha-broken

## Solution

- When registering the HA extension via an RMT server no registration key is required. In that case it set to `null` which was not handled in the code properly.

## Testing

- Tested manually

agama-rmt-ha-fixed

### DIFF
--- a/web/src/components/product/RegistrationExtension.tsx
+++ b/web/src/components/product/RegistrationExtension.tsx
@@ -44,11 +44,16 @@ import RegistrationCodeInput from "./RegistrationCodeInput";
  * @param param0
  * @returns
  */
-const RegisteredExtensionStatus = ({ registrationCode }: { registrationCode: string }) => {
+const RegisteredExtensionStatus = ({ registrationCode }: { registrationCode: string | null }) => {
   const [showCode, setShowCode] = useState(false);
 
   // TRANSLATORS: %s will be replaced by the registration key.
   const [msg1, msg2] = _("The extension has been registered with key %s.").split("%s");
+
+  // free extension or registered via RMT
+  if (registrationCode === null) {
+    return <span>{_("The extension was registered without any registration code.")}</span>;
+  }
 
   return (
     <span>
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul 21 10:49:02 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Fixed crash when registering an extension without registration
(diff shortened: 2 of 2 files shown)


web/src/components/product/RegistrationExtension.tsx
```diff
@@ -44,12 +44,17 @@ import RegistrationCodeInput from "./RegistrationCodeInput";
...
-const RegisteredExtensionStatus = ({ registrationCode }: { registrationCode: string }) => {
+const RegisteredExtensionStatus = ({ registrationCode }: { registrationCode: string | null }) => {
   const [showCode, setShowCode] = useState(false);
 
   // TRANSLATORS: %s will be replaced by the registration key.
   const [msg1, msg2] = _("The extension has been registered with key %s.").split("%s");
 
+  // free extension or registered via RMT
+  if (registrationCode === null) {
+    return <span>{_("The extension was registered without any registration code.")}</span>;
```
joseivanlopez, line 55: Do we need to show this? We omit it in the summary of the product registration.

dgdavid: IMHO, just displaying "Registered" here and "Registered with {code}" in the others would be more than enough for now. I guess that page will be refactored in a future and we can think in other strategies, if really needed.

dgdavid: Good opportunity for extracting some code to an internal component

```tsx
const RegistrationInfo ({ regCode }) => {
  const [showCode, setShowCode] = useState(false);
  
  if (isNull(regCode)  return <span>{_("Registered")}</span>

  return (
    <span>{sprintf(_("Registered with key %s"), showCode ? regCode :  mask(regCode))</span> <Button....

```

dgdavid: Although tricky part is the fact that "null" means 

>   // free extension or registered via RMT

lslezak: The text is displayed instead of the "Register" button, without the label it could be confusing why the button is not there.

dgdavid: I see it different, since the absent of "Register" button makes more clear that no action is needed. Thus, displaying "Registered" text with the appropriated look and feel should be enoug.

That said, it is fine to me going ahead with proposed changes to at least stop crashing. We can invest some time in the future refining the whole approach of that page anyway.

dgdavid (approved):

PR #2589, opened 2025-07-21
feat(rust): add support for the inst.script option

imobachgs:
## Problem

In RC2 we dropped the option to run an arbitrary script to orchestrate the unattended installation. 

## Solution

Bring back this feature but using a separate option. So use `inst.script` instead of `inst.auto` when you need to run a script.

## Testing

- *Added a new unit test*
- *Tested manually*

## Documentation

To be written.

## Pending

* Retrieve the scripts from HTTPS without checking the certificate. This feature is not supported `inst.auto`, so we should address both.
* When using a URL that does not contain a file name (or the same filename), they will conflict. It is something that can be fixed later.


rust/agama-autoinstall/src/scripts.rs
```diff
@@ -0,0 +1,157 @@
...
+    io::Write,
+    os::unix::fs::OpenOptionsExt,
+    path::{Path, PathBuf},
+    process::Output,
+};
+
+use agama_lib::utils::Transfer;
+use url::Url;
+
+/// Downloads and runs user-defined scripts.
```
mvidner:
```suggestion
/// Downloads and runs user-defined scripts for inst.script.
```

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -0,0 +1,157 @@
...
+            insecure,
+        }
+    }
+
+    /// Downloads and runs the script from the given URL.
+    ///
+    /// It downloads the script from the given URL to the runner directory.
+    /// It saves the stdout, stderr and exit code to separate files.
+    ///
+    /// * url: script URL.
```
mvidner:
```suggestion
    /// * url: script URL, supporting agama-specific schemes.
```

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn save_script(&self, url: &str, path: &PathBuf) -> anyhow::Result<()> {
+        let mut file = Self::create_file(&path, 0o700)?;
+        Transfer::get(url, &mut file, self.insecure)?;
+        Ok(())
+    }
+
+    fn save_logs(&self, path: &Path, output: Output) -> anyhow::Result<()> {
+        if !output.stdout.is_empty() {
+            let mut file = Self::create_file(&path.with_extension("stdout"), 0o600)?;
+            file.write_all(&output.stdout).unwrap();
```
mvidner:
```suggestion
            file.write_all(&output.stdout)?;
```

For consistency with the other files.

rust/agama-autoinstall/src/main.rs
```diff
@@ -43,6 +43,16 @@ async fn main() -> anyhow::Result<()> {
     let manager_client = ManagerHTTPClient::new(http.clone());
     let loader = ConfigAutoLoader::new(http.clone())?;
 
+    let scripts = args.get("inst.script");
+    // TODO: add support to disable SSL checks.
+    let mut runner = ScriptsRunner::new("/run/agama/inst-scripts", false);
```
mvidner, line 48: I forgot to consider log collection at first. How does this play along?

imobachgs: Good point, we need to add this path to logs store command.

mvidner (approved):

PR #2594, opened 2025-07-22
Fix Rust changelog sequence

dgdavid:
## Problem

Automated validation tools detected that a changelog entry was not in the correct chronological order, resulting in validation failures.

## Solution

Reordered the changelog entries to ensure they follow the expected sequence, placing the problematic entry in its correct chronological position.


dgdavid:
Closing in favor of #2593, which is not only more complete but also was created before that one.

PR #2605, opened 2025-07-23
Encode connection ID for the URL path

teclator:
## Problem

The CLI needs to encode the connection Id in order to use it in the URL for adding or updating it.

- https://bugzilla.suse.com/show_bug.cgi?id=1246930
- https://trello.com/c/H8mxZJxy/5185-2-agama-loading-a-profile-tries-to-create-an-existing-network-

## Solution

Encode the connection id in order to use it in the URL path for adding or updating a connection

## Testing

- *Tested manually*
**Example profile net.json**
```
suse@vikingo:~$ cat net.json 
{
  "product": {
    "id": "Tumbleweed"
  },
  "network": {
    "connections": [
      {
        "id": "Wired #1",
        "status": "up"
      }
    ]
  }
}

suse@vikingo:~$ agama config load < net.json 
✓ The profile is valid.

suse@vikingo:~$ agama config load < net.json 
✓ The profile is valid.

## REMOVED IRRELEVANT DATA
suse@vikingo:~$ agama config show 
{
  "network": {
    "connections": [
      {
        "id": "Wired #1",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  }
}
```

### DIFF
--- a/rust/agama-lib/src/network/client.rs
+++ b/rust/agama-lib/src/network/client.rs
@@ -20,7 +20,7 @@
 
 use super::{settings::NetworkConnection, types::Device};
 use crate::http::{BaseHTTPClient, BaseHTTPClientError};
-use crate::utils::url::encoded;
+use crate::utils::url::encode;
 
 #[derive(Debug, thiserror::Error)]
 pub enum NetworkClientError {
@@ -57,7 +57,7 @@ impl NetworkClient {
 
     /// Returns an array of network connections
     pub async fn connection(&self, id: &str) -> Result<NetworkConnection, NetworkClientError> {
-        let encoded_id = encoded(id.to_string());
+        let encoded_id = encode(id);
         let json = self
             .client
             .get::<NetworkConnection>(format!("/network/connections/{encoded_id}").as_str())
@@ -72,7 +72,7 @@ impl NetworkClient {
         connection: NetworkConnection,
     ) -> Result<(), NetworkClientError> {
         let id = connection.id.clone();
-        let encoded_id = encoded(id.to_string());
+        let encoded_id = encode(id.as_str());
         let response = self.connection(id.as_str()).await;
 
         if response.is_ok() {
--- a/rust/agama-lib/src/network/client.rs
+++ b/rust/agama-lib/src/network/client.rs
@@ -20,6 +20,7 @@
 
 use super::{settings::NetworkConnection, types::Device};
 use crate::http::{BaseHTTPClient, BaseHTTPClientError};
+use crate::utils::url::encoded;
 
 #[derive(Debug, thiserror::Error)]
 pub enum NetworkClientError {
@@ -56,9 +57,10 @@ impl NetworkClient {
 
     /// Returns an array of network connections
     pub async fn connection(&self, id: &str) -> Result<NetworkConnection, NetworkClientError> {
+        let encoded_id = encoded(id.to_string());
         let json = self
             .client
-            .get::<NetworkConnection>(format!("/network/connections/{id}").as_str())
+            .get::<NetworkConnection>(format!("/network/connections/{encoded_id}").as_str())
             .await?;
 
         Ok(json)
@@ -70,10 +72,11 @@ impl NetworkClient {
         connection: NetworkConnection,
     ) -> Result<(), NetworkClientError> {
         let id = connection.id.clone();
+        let encoded_id = encoded(id.to_string());
         let response = self.connection(id.as_str()).await;
 
         if response.is_ok() {
-            let path = format!("/network/connections/{id}");
+            let path = format!("/network/connections/{encoded_id}");
             self.client.put_void(path.as_str(), &connection).await?
         } else {
             self.client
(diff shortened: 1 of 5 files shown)


rust/agama-lib/src/network/client.rs
```diff
@@ -56,9 +57,11 @@ impl NetworkClient {
 
     /// Returns an array of network connections
     pub async fn connection(&self, id: &str) -> Result<NetworkConnection, NetworkClientError> {
+        let encoded_id: String = byte_serialize(id.as_bytes()).collect();
+        let encoded_id = encoded_id.replace("+", "%20");
```
dgdavid: As far as I know, %20 is for Space, not for `+` Check https://developers.google.com/maps/url-encoding or try encoding yourself a URL with `+` by using the `encodeURIComponent` JavaScript function in the console of your browser developer tools

```js
encodeURIComponent("Wired+1")
/ / Returns "Wired%2B1" 
```

Am I missing something here?

dgdavid (changes requested):

rust/agama-lib/src/network/client.rs
dgdavid: Ah, I see.

`byte_serialize` is encoding spaces as `+` and you converting them to `%20`. A bit weird, honestly. What if instead of using `Wired #1` in your example you use `Wired +1`? Is that `+` properly encoded by the mentioned function?

dgdavid: I've tested by myself in a playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b524dd98aaf9b814123029237b2eb731

```rust
use url::form_urlencoded::byte_serialize;

fn main() {
    let encoded_id: String = byte_serialize("Wired +1".as_bytes()).collect();
    
    println!("{}", encoded_id);
}
```

It indeed returns `Wired+%2B1`

I wonder if there is a way to do the encoding without such a `+` replacement.

dgdavid: JFTR, I found a way

```rust
use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};


fn main() {
  let id: String = utf8_percent_encode("Wired #1+2", NON_ALPHANUMERIC).to_string();

  println!("{}", id);
}
```

```
Standard Error

   Compiling playground v0.0.1 (/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.53s
     Running `target/debug/playground`

Standard Output

Wired%20%231%2B2
```

Maybe for a helper function or method (don't know how that is called in Rust)

To read more, https://docs.rs/percent-encoding/2.3.1/percent_encoding/ part of https://github.com/servo/rust-url

teclator:
> We want this included at RC3. So, as I learned, this should be based on (and targeting to) the corresponding `rc3` branch.

As it was not marked as fastrack I was not sure. I will switch the branch

rust/agama-lib/src/network/client.rs
teclator: Thanks for all the comments and suggestions! 

It should be already fixed by https://github.com/servo/rust-url/pull/1028

dgdavid: As discussed in Slack, looks like byte_serialize is not using the general URI percent-encoding rules.

Read more at https://en.wikipedia.org/wiki/Percent-encoding. In _The application/x-www-form-urlencoded type_ chapter is stated that 

>  The encoding used by default is based on an early version of the general URI percent-encoding rules,[5] with a number of modifications such as newline normalization and replacing spaces with + instead of %20.

And, as you found, there is even a PR open against rust-url to that regard: https://github.com/servo/rust-url/pull/1028

dgdavid: > It should be already fixed by [servo/rust-url#1028](https://github.com/servo/rust-url/pull/1028)

It's a not merged PR :)

rust/agama-lib/src/utils/url.rs
```diff
@@ -0,0 +1,41 @@
...
+// find current contact information at www.suse.com.
+
+use url::form_urlencoded::byte_serialize;
+
+pub fn encoded(value: String) -> String {
+    let serialized_value: String = byte_serialize(value.as_bytes()).collect();
+    // Encode space to '%20' as per url standard
+    // Should be fixed by https://github.com/servo/rust-url/pull/1028
+    serialized_value.replace("+", "%20")
+}
```
dgdavid: Thanks for this! I think it's more useful to have the encoding logic here rather than buried in a network-related file. While it's true that encoding issues have only bitten us in the network layer so far, it's generally safer to always encode URL query strings to prevent future problems.

That said, I have two minor non-blocking comments:

* I found the alternative solution proposed in [this comment](https://github.com/agama-project/agama/pull/2605#discussion_r2227843048) a bit more straightforward and even sensible, as the code is simpler and not tied to forms but to the [percent encoding](https://en.wikipedia.org/wiki/Percent-encoding) instead. That said, I have no experience with Rust, so I might be missing something. 
* I would personally name the function `encode` rather than `encoded`, but again, I’m unfamiliar with Rust naming conventions.

So, overall it LGTM!

dgdavid (approved):

PR #2606, opened 2025-07-23
fix(live): avoid duplicated iSCSI initiator name

joseivanlopez:
### Problem

The live media always includes the same iSCSI initiator name (/etc/iscsi/initiatorname.iscsi) which is copied to the target system. This implies all the installed systems would have the same initiator name, but it must be unique. 

https://bugzilla.suse.com/show_bug.cgi?id=1246280

### Solution

Remove /etc/iscsi/initiatorname.iscsi from the live media, so a new one is always generated when iscsi.service starts on each boot.


teclator (changes requested):
Please add the changelog otherwise, LGTM

PR #2607, opened 2025-07-23
web: Display RAID level at the storage result (bsc#1244697)

ancorgs:
## Problem

The results table of Agama does not show the RAID level for RAID devices.

nolevel

https://bugzilla.suse.com/show_bug.cgi?id=1244697

## Solution

Add the RAID level.

level

The current representation may look a bit redundant in some cases, but it's the best we can do with minimal impact (eg. not modifying yast2-storage-ng). And we want impact to be minimal at the current development stage. 

## Testing

Tested manually as proven by the screenshots.


ancorgs:
Closed in favor of #2608 (right branch RC3)

PR #2609, opened 2025-07-24
fix(storage): format and encrypt devices

joseivanlopez:
### Problem

Agama supports formatting drives and MD RAIDS, but the devices were not encrypted when the encryption is enabled in the UI, see https://bugzilla.suse.com/show_bug.cgi?id=1246939.

Moreover, the generated JSON does not match the schema if there is a formatted device, see https://bugzilla.suse.com/show_bug.cgi?id=1246940.

### Solution

* Adapt conversion from/to model to properly encrypt formatted devices . 
* Fix conversion to JSON.
* Improve the model support checker to properly check the encryption configs.

### DIFF
--- a/service/lib/agama/storage/config.rb
+++ b/service/lib/agama/storage/config.rb
@@ -190,14 +190,37 @@ module Agama
         drives + drives.flat_map(&:partitions)
       end
 
-      # All encryption configs.
+      # Encryption configs, excluding encryptions from skipped devices.
       #
       # @return [Array<Configs::Encryption>]
-      def encryptions
+      def valid_encryptions
+        valid_devices = supporting_encryption.reject { |c| skipped?(c) }
+
         [
-          supporting_encryption.map(&:encryption),
+          valid_devices.map(&:encryption),
           volume_groups.map(&:physical_volumes_encryption)
         ].flatten.compact
+      end
+
+      # Drive configs, excluding skipped ones.
+      #
+      # @return [Array<Configs::Drive>]
+      def valid_drives
+        drives.reject { |d| skipped?(d) }
+      end
+
+      # MD RAID configs, excluding skipped ones.
+      #
+      # @return [Array<Configs::MdRaid>]
+      def valid_md_raids
+        md_raids.reject { |r| skipped?(r) }
+      end
+
+      # Partitions configs, excluding skipped ones.
+      #
+      # @return [Array<Configs::Partition>]
+      def valid_partitions
+        partitions.reject { |p| skipped?(p) }
       end
 
       # Configs directly using a device with the given alias.
@@ -277,6 +300,16 @@ module Agama
         else
           false
         end
+      end
+
+      # Whether the config is skipped.
+      #
+      # @param config
+      # @return [Boolean]
+      def skipped?(config)
+        return false unless config.respond_to?(:skipped?)
+
+        config.skipped?
       end
     end
   end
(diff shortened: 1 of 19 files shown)


service/lib/agama/storage/config.rb
```diff
@@ -190,6 +190,39 @@ def potential_for_md_device
         drives + drives.flat_map(&:partitions)
       end
 
+      # Encryption configs, excluding encryptions from skipped devices.
+      #
+      # @return [Array<Configs::Encryption>]
+      def valid_encryptions
```
ancorgs, line 196: I'm not sure about these `valid_whatever`. I had no problem with the name when they were internal methods of a conversion. But now that they are part of the `Storage::Config` API I would be more picky. And I don't think "valid" represents its meaning.

service/lib/agama/storage/config_conversions/to_model_conversions/with_partitions.rb
```diff
@@ -29,15 +29,11 @@ module ToModelConversions
         module WithPartitions
           # @return [Array<Hash>]
           def convert_partitions
-            valid_partitions
+            config.partitions
+              .reject(&:skipped?)
```
ancorgs, line 33: Shouldn't this use `config.valid_partitions`?

joseivanlopez: In this context, config is `Configs::Drive` or `Configs::MdRaid`. Do you suggest to add `#valid_partitions` to those classes?

service/lib/agama/storage/config.rb
joseivanlopez: I don't like `#valid_whatever` neither, but my brain is not able to find anything better right now :(

ancorgs: Some suggestions from the top of my head: `relevant_`, `non_skipped_`, `final_`, `processable_`.

service/lib/agama/storage/config_conversions/to_model_conversions/with_partitions.rb
ancorgs: Not really. I thought it was a `Config` object. But actually it may make some sense.

ancorgs (approved):
Despite my comments about some method names, this generally looks good to me. So approving.

ancorgs:
BTW, I noticed this is not longer considered to result in a supported model. It was valid at some point of the lifecycle of this PR.

```json
{
  "storage": {
    "drives": [
      {
        "encryption": {
          "luks2": {
            "password": "linuxes"
          }
        }
      }
    ]
  }
}
```

service/lib/agama/storage/config.rb
ancorgs: More suggestions: `honored_`, `_to_be_honored`

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Jul 24 21:09:25 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Make encryption work with whole formatted devices (bsc#1246939).
+- Do not export partitions to json if there is no partition
+  (bsc#12469409).
```
ancorgs: Wrong bug number. The correct is 1246940, without the final 9.

PR #2610, opened 2025-07-24
Added "--insecure" option to "agama config load" and "agama config generate" commands

lslezak:
## Problem

- Agama does not allow to disable SSL certificate check for `inst.auto` and `inst.script` boot options
- https://bugzilla.suse.com/show_bug.cgi?id=1246836

## Solution

- This is the first step, add `--insecure` option to `agama config load` and `agama config generate` commands which are internally used for these boot options
- Use the global `--insecure` option

## Testing

- Tested manually

```console
agama:~ # agama config generate https://192.168.1.8:4433/auto.json
Retrieving data from URL https://192.168.1.8:4433/auto.json

Caused by:
    0: Could not retrieve https://192.168.1.8:4433/auto.json
    1: [60] SSL peer certificate or SSH remote key was not OK (SSL certificate problem: self-signed certificate)
agama:~ # agama --insecure config generate https://192.168.1.8:4433/auto.json
{
  "user": {
    "fullName": "test",
    "password": "test",
    "hashedPassword": false,
    "userName": "test"
  },
  "root": {
    "password": "linux",
    "hashedPassword": false
  },
  "product": {
    "id": "Tumbleweed"
  }
}
✓ The profile is valid.

agama:~ # agama config load https://192.168.1.8:4433/auto.json
Retrieving data from URL https://192.168.1.8:4433/auto.json

Caused by:
    0: Could not retrieve https://192.168.1.8:4433/auto.json
    1: [60] SSL peer certificate or SSH remote key was not OK (SSL certificate problem: self-signed certificate)
agama:~ # agama --insecure config load https://192.168.1.8:4433/auto.json
✓ The profile is valid.

[1/2] Analyze disks
[2/2] Configure software
```

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -52,9 +52,6 @@ pub enum ConfigCommands {
     Load {
         /// JSON file: URL or path or `-` for standard input
         url_or_path: Option<CliInput>,
-        /// Disables SSL verification for HTTPS downloads
-        #[arg(short, long)]
-        insecure: bool,
     },
 
     /// Validate a profile using JSON Schema
@@ -83,9 +80,6 @@ pub enum ConfigCommands {
     Generate {
         /// JSON file: URL or path or `-` for standard input
         url_or_path: Option<CliInput>,
-        /// Disables SSL verification for HTTPS downloads
-        #[arg(short, long)]
-        insecure: bool,
     },
 
     /// Edit and update installation option using an external editor.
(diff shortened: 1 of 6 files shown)


imobachgs (commented):
We already have [a global `--insecure` switch](https://github.com/agama-project/agama/blob/master/rust/agama-cli/src/lib.rs#L73). And I think we should use the same.

Actually, I did not noticed when reviewing #2568.

rust/agama-cli/src/config.rs
```diff
@@ -99,6 +99,7 @@ pub async fn run(
     http_client: BaseHTTPClient,
     monitor: MonitorClient,
     subcommand: ConfigCommands,
+    insecure: bool,
```
imobachgs: Why not pass the GlobalOpts? If tomorrow we add a, let's say, `--verbose` argument, you might need to add another argument.

imobachgs (approved):
LGTM. Just a minor suggestion.

PR #2611, opened 2025-07-24
feat: disable SSL checks for inst.auto and inst.script

imobachgs:
## Problem

[bsc#1246836](http://bugzilla.suse.com/show_bug.cgi?id=1246836)

It is not possible to disable the SSL checks for the `inst.auto` and `inst.script` boot options.

## Solution

Add a pair of options, `inst.auto_insecure` and `inst.script_insecure`, to allow disabling the SSL checks.

It was required to extend the `agama config generate` subcommand to support the `--insecure` flag. See #2610 for further details.

Additionally, report the user when the script cannot be downloaded.

## Testing

- Tested manually.

## Tasks

- [x] Implement `inst.auto_insecure`
- [x] Implement `inst.script_insecure`
- [x] (optional) Inform when the script cannot be fetched.

### DIFF
--- a/rust/agama-autoinstall/src/auto_loader.rs
+++ b/rust/agama-autoinstall/src/auto_loader.rs
@@ -47,6 +47,7 @@ impl ConfigAutoLoader {
     /// Builds a new loader.
     ///
     /// * `http`: base client to connect to Agama.
+    /// * `insecure`: whether to skip SSL cert checks.
     pub fn new(http: BaseHTTPClient, insecure: bool) -> anyhow::Result<Self> {
         Ok(Self {
             questions: UserQuestions::new(http),
--- a/rust/agama-autoinstall/src/scripts.rs
+++ b/rust/agama-autoinstall/src/scripts.rs
@@ -43,8 +43,9 @@ pub struct ScriptsRunner {
 impl ScriptsRunner {
     /// Creates a new scripts runner.
     ///
-    /// * path: working directory for the runner.
-    /// * insecure: whether to check certificates when downloading scripts.
+    /// * `http`: base client to connect to Agama.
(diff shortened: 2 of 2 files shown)


rust/agama-autoinstall/src/auto_loader.rs
```diff
@@ -45,16 +39,19 @@ const PREDEFINED_LOCATIONS: [&str; 6] = [
...
 }
 
 impl ConfigAutoLoader {
     /// Builds a new loader.
     ///
     /// * `http`: base client to connect to Agama.
-    pub fn new(http: BaseHTTPClient) -> anyhow::Result<Self> {
-        let questions = QuestionsHTTPClient::new(http)?;
-        Ok(Self { questions })
+    pub fn new(http: BaseHTTPClient, insecure: bool) -> anyhow::Result<Self> {
```
jreidinger, line 51: new param not documented in method documentation

rust/package/agama.changes
```diff
@@ -24,8 +30,8 @@ Mon Jul 21 15:07:40 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 -------------------------------------------------------------------
 Mon Jul 21 14:35:57 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
-- Add support for an inst.script that allows to run an arbitrary
-  script (bsc#1246702, gh#agama-project/agama#2589).
+- Add support for an inst.script option that allows to run an
```
jreidinger, line 33: well, to be honest usually OBS reviewers are not that happy about changing history.

imobachgs: I do not expect it to be a problem, especially given that the original change was a few days ago.

PR #2612, opened 2025-07-25
feat: embed questions configuration in the profile

imobachgs:
## Problem

* Bug [bsc#1246997](https://bugzilla.suse.com/show_bug.cgi?id=1246997)

To answer questions in advance when using the unattended mode, it is required to use a pre-script. If you just want Agama to use the default values, it is kind of straight-forward:

```jsonnet
  scripts: {
    pre: [
      {
        name: 'disable-questions',
	content: |||
          #!/usr/bin/bash
          agama questions mode non-interactive
	|||
      }
    ]
  }
```

However, if you want to answer an specific question (e.g., providing a password for an existing encrypted file system) it is kind of tricky. You might need to create a file like this:

```json
  {
    "answers": [
      {
        "password": "my-password",
        "class": "storage.luks_activation"
      },
    ]
  }
```

And the load it using a pre-script:

```jsonnet
  scripts: {
    pre: [
      {
        name: 'disable-questions',
	content: |||
          #!/usr/bin/bash
          agama download http://my-server.lan/answers.json answers.json
          agama questions answers answers.json
	|||
      }
    ]
  }
```

## Solution

Allow embedding the answers in the profile.

```json
  "questions": {
    "policy": "auto",
    "answers": [
      { "class": "storage.luks_activation", "password": "nots3cr3t" }
    ]
```

* `policy`: ask the user (`user`) or choose the default option (`auto`).
* `answers`: list of answers to the given question.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -999,6 +999,12 @@
           "title": "Answer for a password-based question",
           "description": "Answer to use for a question that expects an answer",
           "type": "string"
+        },
+        "data": {
+          "title": "Additional data for matching questions",
+          "description": "Additional data for matching questions and answers",
+          "type": "object",
+          "example": [{ "device": "/dev/sda" }]
         }
       }
     }
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -271,7 +271,8 @@
           "title": "Policy to answer questions",
           "description": "Ask questions to the user or select the default value",
           "type": "string",
-          "enum": ["auto", "user"]
+          "enum": ["auto", "user"],
+          "default": "user"
         },
         "answers": {
           "title": "Pre-defined answer for a matching question",
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -267,10 +267,11 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "interactive": {
-          "title": "Interactive mode",
-          "description": "Whether to run Agama in interactive mode. If set to false, Agama will use the default values for each question.",
-          "type": "boolean"
+        "policy": {
+          "title": "Policy to answer questions",
+          "description": "Ask questions to the user or select the default value",
+          "type": "string",
+          "enum": ["auto", "user"]
         },
         "answers": {
           "title": "Pre-defined answer for a matching question",
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -262,6 +262,25 @@
         }
       }
     },
+    "questions": {
+      "title": "How to handle Agama questions",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "interactive": {
+          "title": "Interactive mode",
+          "description": "Whether to run Agama in interactive mode. If set to false, Agama will use the default values for each question.",
+          "type": "boolean"
+        },
+        "answers": {
+          "title": "Pre-defined answer for each question.",
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/answer"
+          }
+        }
+      }
+    },
     "product": {
       "title": "Product to install",
       "type": "object",
@@ -954,6 +973,34 @@
       },
       "required": ["destination"],
       "oneOf": [{ "required": ["url"] }, { "required": ["content"] }]
+    },
+    "answer": {
+      "title": "Automatic answer to questions",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "class": {
+          "title": "Question class",
+          "description": "Each question has a \"class\" which works as an identifier.",
+          "type": "string",
+          "examples": ["storage.activate_multipath"]
+        },
+        "text": {
+          "title": "Question text",
+          "description": "Question full text",
+          "type": "string"
+        },
+        "answer": {
+          "title": "Question answer",
+          "description": "Answer to use for the question.",
+          "type": "string"
+        },
+        "password": {
+          "title": "Answer for a password-based question",
+          "description": "Answer to use for a question that expects an answer",
+          "type": "string"
+        }
+      }
     }
   }
 }
--- a/rust/agama-lib/share/profile.schema.json
+++ b/rust/agama-lib/share/profile.schema.json
@@ -273,7 +273,7 @@
           "type": "boolean"
         },
         "answers": {
-          "title": "Pre-defined answer for each question.",
+          "title": "Pre-defined answer for a matching question",
           "type": "array",
           "items": {
             "$ref": "#/$defs/answer"
@@ -996,8 +996,7 @@
           "type": "string"
         },
         "password": {
-          "title": "Answer for a password-based question",
-          "description": "Answer to use for a question that expects an answer",
+          "title": "Password provided as response to a password-based question",
           "type": "string"
         },
         "data": {
--- a/rust/agama-lib/src/questions/store.rs
+++ b/rust/agama-lib/src/questions/store.rs
@@ -1,0 +1,53 @@
+// Copyright (c) [2025] SUSE LLC
+//
+// All Rights Reserved.
+//
+// This program is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 2 of the License, or (at your option)
+// any later version.
+//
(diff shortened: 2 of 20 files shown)


dgdavid:
> Allow embedding the answers in the profile.
> 
> ```json
>   "questions": {
>     "interactive": false,
>     "answers": [
>       { "class": "storage.luks_activation", "password": "diekn3m." }
>     ]
> ```

What if there are more than one "luks activation" questions?

imobachgs:
> > Allow embedding the answers in the profile.
> > ```json
> >   "questions": {
> >     "interactive": false,
> >     "answers": [
> >       { "class": "storage.luks_activation", "password": "diekn3m." }
> >     ]
> > ```
> 
> What if there are more than one "luks activation" questions?

The first answer wins, always. But you can be more specific:

```json
  "questions": {
    "interactive": false,
    "answers": [
      {
        "class": "storage.luks_activation",
        "password": "nots3cr3t",
        "data": {
          "device": "/dev/vda1", # or label or...
        }
      }
    ]
  }
```

You can find the questions and the additional fields in the [questions internal doc](https://github.com/agama-project/agama/blob/master/doc/questions.md#question-types).

dgdavid:
> 
> The first answer wins, always. But you can be more specific:
> 
> ```json
>   {
>     "questions": {
>       "interactive": false,
>       "answers": [
>         {
>           "class": "storage.luks_activation",
>           "password": "nots3cr3t",
>           "data": {
>             "device": "/dev/vda1"
>           }
>         }
>       ]
>     }
>   }
> ```
> 
> You can find the questions and the additional fields in the [questions internal doc](https://github.com/agama-project/agama/blob/master/doc/questions.md#question-types).

Thanks a lot for the answer. Doubt solved.

rust/agama-lib/share/profile.schema.json
```diff
@@ -262,6 +262,25 @@
...
       }
     },
+    "questions": {
+      "title": "How to handle Agama questions",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "interactive": {
+          "title": "Interactive mode",
+          "description": "Whether to run Agama in interactive mode. If set to false, Agama will use the default values for each question.",
```
jreidinger: I would document that default is `true`

rust/agama-lib/share/profile.schema.json
```diff
@@ -262,6 +262,25 @@
...
+          "type": "boolean"
+        },
+        "answers": {
+          "title": "Pre-defined answer for each question.",
```
jreidinger: I find it a bit misleading  as it can feel like all possible questions needs its answer which is not true. so probably I would write something like `"Pre-defined answers for defined questions."

rust/agama-lib/share/profile.schema.json
```diff
@@ -954,6 +973,40 @@
...
+          "type": "string"
+        },
+        "answer": {
+          "title": "Question answer",
+          "description": "Answer to use for the question.",
+          "type": "string"
+        },
+        "password": {
+          "title": "Answer for a password-based question",
+          "description": "Answer to use for a question that expects an answer",
```
jreidinger: I would write `"Password to use for a question that beside an answer expects also provided password"`

rust/agama-lib/src/questions/store.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl QuestionsStore {
+    pub fn new(client: BaseHTTPClient) -> Self {
+        Self {
+            questions_client: QuestionsHTTPClient::new(client),
+        }
+    }
+
+    pub async fn load(&self) -> QuestionsStoreResult<Option<QuestionsConfig>> {
+        Ok(None)
```
jreidinger, line 47: this is a bit inconsistent with rest. With this the combination of `agama config load` and `agama config show` will omit answers, which I am not sure if it is intention or expected.

rust/agama-lib/share/profile.schema.json
dgdavid: > I would write `"Password to use for a question that beside an answer expects also provided password"`

Too complex for my English parser :stuck_out_tongue_winking_eye: 

My two cents, 

```json
"password": {
  "title": "Password",
  "description": "The password provided in response to a password-based question.",
  "type": "string"
}
```

rust/agama-lib/src/questions/store.rs
imobachgs: It is intentional. I did not find a good way to export the list of questions. They are stored as trait objects if I am not wrong. I decided we can live with that for 16.0.  But I will reevaluate it if needed.

imobachgs: I had a closer look to this issue. The main problem is that the elements in the [answer_strategies](https://github.com/agama-project/agama/blob/b191edc46d26a6446805e25fdd9c0641e4b47303/rust/agama-server/src/questions.rs#L159) should implement `Serialize` too, but this trait is not compatible with `dyn` (it is generic). We could rely on the [erased_serde crate](https://docs.rs/erased-serde/latest/erased_serde/), but I have not tried.

To be honest, given that we plan to do a major refactor of this part (dropping the D-Bus interface), I would consider a different implementation (e.g., using enums instead of trait objects).

jreidinger (approved):

rust/agama-lib/share/profile.schema.json
```diff
@@ -954,6 +975,39 @@
...
+          "title": "Password provided as response to a password-based question",
+          "type": "string"
+        },
+        "data": {
+          "title": "Additional data for matching questions",
+          "description": "Additional data for matching questions and answers",
+          "type": "object",
+          "example": [{ "device": "/dev/sda" }]
+        }
+      }
```
joseivanlopez, line 1010: Sorry, because I did not have a look to this PR before. I think we can improve the current schema for answers. With the current schema I see these problems:

* Some properties (e.g., "password") only make sense for certain answers.
* The "data" property is an open object. You have to check the documentation to know what properties you can use for each class of question.

Because of that, I think the current schema is too open and not very self-descriptive. I would prefer a schema that perfectly define all the possible answers and how to use each one. For example, for LUKS activation we could have:

Skip:

~~~json
{
  "answers": [
    {
      "storage.luks_activation": "skip" 
    }
  ]
}
~~~

Decrypt:

~~~json
{
  "answers": [
    {
      "storage.luks_activation": {
        "decrypt": "12345"
      } 
    }
  ]
}
~~~

Skip with conditions:

~~~json
{
  "answers": [
    {
      "storage.luks_activation": { 
        "skip": {
          "device": "/dev/vda1"
        }
      }
    }
  ]
}
~~~

Decrypt with conditions:

~~~json
{
  "answers": [
    {
      "storage.luks_activation": { 
        "decrypt": {
          "password": "12345",
          "device": "/dev/vda1"
        }
      }
    }
  ]
}
~~~

And the schema would be something like this:

~~~json
{
  "answer": {
    "anyOf": [
      { "$ref": "#/$defs/storageLuksActivation" }
    ]
  },
  "storageLuksActivation": {
    "title": "Automatic answer for LUKS activation",
    "type": "object",
    "properties": {
      "storage.luks_activation": {
        "anyOf": [
          { "$ref": "#/$defs/storageLuksActivationSkip" },
          { "$ref": "#/$defs/storageLuksActivationDecrypt" },
          { "$ref": "#/$defs/storageLuksActivationSkipWithData" },
          { "$ref": "#/$defs/storageLuksActivationDecriptWithData" }
        ]
      }
    },
    "storageLuksActivationSkip": { "const": "skip" },
    "storageLuksActivationDecrypt": {  
      "type": "object",
      "properties": {
        "decrypt": { 
          "title": "password",
          "type": "string"
        } 
      }
    },
    "storageLuksActivationSkipWithData": {  
      "type": "object",
      "properties": {
        "skip": { 
          "type": "object",
          "properties": {
            "device": { "type": "string" }
          }
        } 
      }
    },
    "storageLuksActivationDecryptWithData": {  
      "type": "object",
      "properties": {
        "skip": { 
          "type": "object",
          "properties": {
            "password": { "type": "string" },
            "device": { "type": "string" }
          }
        } 
      }
    }
  }
}
~~~

imobachgs: [..]

> Because of that, I think the current schema is too open and not very self-descriptive. I would prefer a schema that perfectly define all the possible answers and how to use each one. For example, for LUKS activation we could have:

[..]

I agree. However, I went for the simplistic approach just because I did not have enough time. A first step would be to document all the questions (at this point the `questions.md` document is outdated).

PR #2613, opened 2025-07-25
fix(web): do not encode connection name in POST payload

dgdavid:
## Problem

#2576 went too far and encoded not only the connection name in the URL, but also the POST payload for the _persist_ request, which is not correct.

As a result, the _Use for installation only_ switch is actually not working in the web interface. What is worse, it  even appears to work, due to the fatal combination of an optimistic interface update and the backend not complaining but still responding with a `204` status (i.e., no chances for rolling back the optimistic update).

## Solution

Stop encoding the connection name in the POST payload.

## Testing

Tested manually only, as there's currently no support for automating this type of API test.

### Payload before the change

```json
{"only":["Wired%20%231"],"value":false}
```

### Payload after the change

```json
{"only":["Wired #1"],"value":false}
```

### DIFF
--- a/web/src/api/network.ts
+++ b/web/src/api/network.ts
@@ -91,7 +91,7 @@ const disconnect = (name: string) =>
  * Make the connection persistent after the installation
  */
 const persist = (name: string, value: boolean) =>
-  post(`/api/network/connections/persist`, { only: [encodeURIComponent(name)], value });
+  post(`/api/network/connections/persist`, { only: [name], value });
 
 export {
   fetchState,
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jul 25 19:42:18 UTC 2025 - David Diaz <dgonzalez@suse.com>
+
+- Fix an issue preventing connections from being set as
+  "installation-only" in the web interface
+  (gh#agama-project/agama#2613).
+
+-------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


web/src/api/network.ts
```diff
@@ -91,7 +91,7 @@ const disconnect = (name: string) =>
  * Make the connection persistent after the installation
  */
 const persist = (name: string, value: boolean) =>
-  post(`/api/network/connections/persist`, { only: [encodeURIComponent(name)], value });
+  post(`/api/network/connections/persist`, { only: [name], value });
```
dgdavid, line 94: **Honest question**

Now that I've run into this issue, I’m wondering why we don’t just use two separate requests to stay consistent with the rest of the network-related API.

Something like:

`POST /connections/:id/persist`, to set the connection as persistent.
`DELETE /connections/:id/persist` (or maybe `POST /connections/:id/transient`), to set a connection as transient.

BTW, it has been weird to me to receive a `204` for a request to change a "not existent connection" instead of an error or so.

dgdavid:
Already merged in `rc3`. Thanks @imobachgs

imobachgs (commented):
IMHO, this change should be included in `master` in a sync pull request (instead of having two PRs for any feature we fix in `rc3`).

web/src/api/network.ts
teclator: Mainly because it allows to persist a list of connections

dgdavid: I see, although I still having doubts about the right implementation for it. Let's see if I manage to propose some sensible changes to support all use cases in a more consistent way.

Thanks for the reply 🙏

PR #2616, opened 2025-07-27
[master] live: do not fail if there are no rpm scriplets to clean

g7:
Recent libzypp versions would ensure that there are no stray rpm scriptlets left in /var/tmp [0], so ensure we don't fail if there aren't any.

[0] https://github.com/openSUSE/libzypp/commit/760882b533903da9420fd4525938a73c743b5764

## Testing

- *Tested manually*


imobachgs:
This code was merged to master as part of #2621.

PR #2617, opened 2025-07-27
[rc3] live: do not fail if there are no rpm scriplets to clean

g7:
Recent libzypp versions would ensure that there are no stray rpm scriptlets left in /var/tmp [0], so ensure we don't fail if there aren't any.

[0] https://github.com/openSUSE/libzypp/commit/760882b533903da9420fd4525938a73c743b5764

Backport of #2616 to `rc3` branch.


g7:
Forwarded to SLES to https://src.suse.de/products/SLES/pulls/362

imobachgs (approved):

PR #2622, opened 2025-07-28
fix(storage): refresh on changes

joseivanlopez:
The storage UI does not automatically refresh when other clients perform changes (e.g., after loadig a config using the CLI). This happens to both the storage page and the storage section in the overview page.


web/src/components/core/AlertOutOfSync.tsx
```diff
@@ -72,7 +72,7 @@ export default function AlertOutOfSync({ scope, ...alertProps }: AlertOutOfSyncP
     if (missingScope) return;
 
     return client.onEvent((event) => {
-      event.type === `${scope}Changed` && setActive(event.clientId !== client.id);
+      event.type === `${scope}Changed` && setActive(event.client_id !== client.id);
```
dgdavid: I actually expect `clientId` here. I though they were using something like `#[serde(rename_all = "camelCase")]` in Rust part, but most probably I'm missing something.

I even would say that this worked with `clientId` form when I tested it in previous PRs.

joseivanlopez: Actually, the client id was wrongly reported in the `StorageChanged` event.

web/src/components/core/AlertOutOfSync.tsx
```diff
@@ -46,14 +40,12 @@ type AlertOutOfSyncProps = Partial<AlertProps> & {
...
- * caused by unexpected refreshes.
+ * It warns that the interface may be out of sync and forces reloading
+ * before continuing to avoid issues and data loss.
  *
  * It works by listening for "Changed" events on the specified scope:
  *
- * - Displays a toast alert if the event originates from a different client
+ * - Displays a popup if the event originates from a different client
  *   (based on client ID).
  * - Automatically dismisses the alert if a subsequent event originates from
```
dgdavid: NP: Although this still somehow true, it's unlikely to happen since the client displaying the alert is now in a blocking mode. Thus, documentation can be simplified.

web/src/components/core/AlertOutOfSync.tsx
```diff
@@ -21,20 +21,14 @@
...
-} from "@patternfly/react-core";
+import { Content } from "@patternfly/react-core";
 import { useInstallerClient } from "~/context/installer";
 import { isEmpty } from "radashi";
 import { _ } from "~/i18n";
 import { locationReload } from "~/utils";
+import Popup, { PopupProps } from "~/components/core/Popup";
 
-type AlertOutOfSyncProps = Partial<AlertProps> & {
+type AlertOutOfSyncProps = Partial<PopupProps> & {
```
dgdavid: NP: we could also `<Omit>` the _title_ and _isOpen_ props. They are actually overwritten by the component anyway, so better to let consumer know they are not available

web/src/components/core/AlertOutOfSync.tsx
```diff
@@ -84,32 +76,19 @@ export default function AlertOutOfSync({ scope, ...alertProps }: AlertOutOfSyncP
...
-    </AlertGroup>
+    <Popup
+      {...alertProps}
+      title={title}
+      isOpen={active}
+      backdropClassName="agm-backdrop-gray-and-blur"
+    >
+      <Content component="p">{_("The configuration has been updated externally.")}</Content>
+      <Content component="p">
+        {_("The page must be reloaded to get the latest data and avoid issues or data loss.")}
```
dgdavid: It's not the page, but the whole _app_. What about



> 
```suggestion
        {_("Reloading is required to get the latest data and avoid issues or data loss.")}
```

dgdavid (commented):
Thanks @joseivanlopez 

Web part LGTM despite the few minor comments I've left.

web/src/components/core/AlertOutOfSync.tsx
dgdavid: I mean, technically _is the page_, but I would avoid such a term and go for something _more generic_ or _less detailed_ like the above proposal.

ancorgs (approved):
I see there is room to improve the behavior when you are at a form that stops making sense (eg. editing a partition that is not longer there in the new config).

But the current status is already much better than what we had before. So LGTM.

PR #2623, opened 2025-07-29
bsc#1247248: only complain about the missing product in some cases

imobachgs:
## Problem

#2570 introduced a check to make sure that a product is selected in some cases when the user tries to load a configuration that contains some specific sections (basically software and storage-related ones).

However, `agama config load` now complains *always*.

## Solution

The root cause is that, when deserializing the settings, a users section is *always* present (still investigating why). However, the users section does not require the product anymore.

## Testing

- Tested manually


joseivanlopez (approved):
LGTM

PR #2624, opened 2025-07-29
bsc#1247266: make RPMs given via inst.dud installable

imobachgs:
## Problem

The Agama DUD mechanism works in the following way when it comes to RPM handling:

* If you are using an RPM, it just patches the installation media.
* If you are using a DUD (created by mkdud), it sets up a repository containing the RPMs to make them available during installation (if the `--install repo` method is included).
* 
The reporter of the [bsc#1247266](https://bugzilla.suse.com/show_bug.cgi?id=1247266) expected that an RPM given as `inst.dud=` option would be available for installation.

## Solution

Include any RPM package specified as `inst.dud=` in the DUD repository for installation.

## Testing

- *Tested manually*

## Open question

* Is this actually a good idea? We could ask the users to use a DUD if they want this behavior.

### DIFF
--- a/live/live-root/usr/lib/dracut/modules.d/99agama-dud/agama-dud-apply.sh
+++ b/live/live-root/usr/lib/dracut/modules.d/99agama-dud/agama-dud-apply.sh
@@ -67,7 +67,8 @@ apply_updates() {
 
 # Applies an update from an RPM package
 #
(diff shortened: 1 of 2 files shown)


imobachgs:
We decided to not implement this change. Check the bug report for further details.

PR #2626, opened 2025-07-30
Security adapt

jreidinger:
## Problem

For SLES plan is to have selinux policy driven by rpm and agama overwrites it even if selinux policy that uses permissive is installed.

- https://trello.com/c/QdrVfr8F/5198-slesfor-sap-applications-160-p1-1247046-sles160-build1221incorrect-kernel-cmdline-security-generated-when-network-installation-w

## Solution

Do not set selinux policy and keep rpm defaults.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/service/lib/agama/security.rb
+++ b/service/lib/agama/security.rb
@@ -25,6 +25,8 @@ require "yast2/execute"
 require "agama/config"
 require "agama/http"
 
+Yast.import "Bootloader"
+
 # FIXME: monkey patching of security config to not read control.xml and
 # instead use Agama::Config
 # TODO: add ability to set product features in LSM::Base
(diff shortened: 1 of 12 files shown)


service/lib/agama/security.rb
```diff
@@ -73,11 +75,18 @@ def initialize(logger, config)
     end
 
     def write
+      # at first clear previous kernel params
+      selected = lsm_selected
+      selected&.reset_kernel_params
```
teclator, line 80: is necessary at all? I mean, the lsm_selected is already different to nil?

teclator (approved):
Just a minor note about the reset of the kernel parameters otherwise it looks good to me.

service/lib/agama/security.rb
jreidinger: I am not sure, I just want to be on safe side if it gets somehow on CLI like from proposal. Method docu mentions that it should be called when changing lsm.

PR #2627, opened 2025-07-30
feat: add a client ID to be used in signals

imobachgs:
## Problem

This PR adds a client ID that can be used in signals, making it easy to find out who originated a change.

## Solution

Add an UUID as a client ID to the JWT. Any handler can access the client ID using the [Axum's Extension extractor](https://docs.rs/axum/latest/axum/struct.Extension.html). Accessing the client ID is easy (see 018478c).

## Testing

- *Tested manually*


joseivanlopez (approved):
LGTM

PR #2629, opened 2025-07-30
Log error when activation fails but continue

teclator:
## Problem

Currently the add_or_update_connection client method not only add_or_update a connection in NetworkManager but it also tries to activate or deactivate it based on the status property. The problem is that the activation could fail in some situations like:

  - The interface bind to the connection is not present.
  - The interface bind to the connection is already applying other connection.
  - In case of Wireless, wireless could be disabled by hardware.
  
.... and other cases.

When this happen it returns with an error and the write operation does a rollback to the checkpoint created before starting with the write.

## Solution

- When the activation or deactivation of a connection return an error it will be just logged in order to finish with the write of the connections.
- Before activating a connection, if it is bind to an specific interface which is applying a different connection then the device will be disconnected in order to be able to activate the latest one. Example, the default connection created by dracut which could be applied to multiple-connections.

## Testing

- https://trello.com/c/c7HAWvBz/5020-1245548-modifying-a-network-connection-binding-settings-does-not-apply-the-changes-to-the-active-connection

### DIFF
--- a/rust/agama-network/src/nm/client.rs
+++ b/rust/agama-network/src/nm/client.rs
@@ -297,6 +297,8 @@ impl<'a> NetworkManagerClient<'a> {
                 .map_err(NmError::FailedNmVersionParse)?,
         );
 
+        let devices = self.devices().await?;
+
         let path = if let Ok(proxy) = self.get_connection_proxy(conn.uuid).await {
             let original = proxy.get_settings().await?;
             let merged = merge_dbus_connections(&original, &new_conn)?;
@@ -305,6 +307,7 @@ impl<'a> NetworkManagerClient<'a> {
             } else {
                 UpdateFlags::InMemoryOnly
             };
+
             proxy
                 .update2(merged, persist as u32, Default::default())
                 .await?;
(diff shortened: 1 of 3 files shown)


rust/agama-network/src/nm/client.rs
```diff
@@ -328,17 +331,47 @@ impl<'a> NetworkManagerClient<'a> {
         // FIXME: Do not like that an activation/deactivation could not apply the changes because of
         // a roolback when calling this method with an error.
         if conn.is_up() {
+            if let Some(interface) = &conn.interface {
+                for device in devices {
+                    if interface.to_string() == device.name {
```
imobachgs:
```suggestion
                    if interface == &device.name {
```

And if you want to save some nesting, what about this?

```rust
if interface != &device.name {
  continue;
}
```

rust/agama-network/src/nm/client.rs
```diff
@@ -328,17 +331,47 @@ impl<'a> NetworkManagerClient<'a> {
...
+                        if let Some(device_conn) = device.connection {
```
imobachgs: JFYI, in the future it would be possible to combine `if let` and `&&` expressions. See https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains.

rust/agama-network/src/nm/client.rs
```diff
@@ -411,6 +448,37 @@ impl<'a> NetworkManagerClient<'a> {
         Ok(proxy)
     }
 
+    async fn get_device_proxy(&self, name: String) -> Result<DeviceProxy, NmError> {
```
imobachgs, line 456: Please, document the function.

rust/agama-network/src/nm/client.rs
```diff
@@ -411,6 +448,37 @@ impl<'a> NetworkManagerClient<'a> {
...
+            let proxy = DeviceProxy::builder(&self.connection)
+                .path(path.as_str())?
+                .build()
+                .await?;
+
+            if let Ok(device) = DeviceFromProxyBuilder::new(&self.connection, &proxy)
+                .build()
+                .await
+            {
+                if device.name == name {
```
imobachgs, line 468: Can this happen more than one time? Because if that's not the case, you could break the loop here. You found what you were looking for.

rust/agama-network/src/nm/client.rs
```diff
@@ -411,6 +448,37 @@ impl<'a> NetworkManagerClient<'a> {
...
+                    device_path = Some(path.to_owned());
+                }
+            } else {
+                tracing::warn!("Ignoring network device on path {}", &path);
```
imobachgs: Is this really a warning?

imobachgs (changes requested):
In general it looks good. Just some minor things.

rust/agama-network/src/nm/client.rs
teclator: We need rust version 2024 isn't it?

rust/agama-network/src/nm/client.rs
teclator: Is even not needed, I added for debugging mainly

PR #2632, opened 2025-07-31
Boot from the hard disk by default

jreidinger:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1247438
- https://trello.com/c/9yAyyDZ1/1952-sles160-agama-live-iso-reboots-into-the-installer-again

The default option in grub menu is Install XXX. It has consequence that unattended installation where boot entry cannot be modified is still the same in endless loop.

## Solution

Use same behavior as SLES15 and have boot from hard disk first and default option.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

new_boot_order


mvidner (approved):
LGTM, thanks

PR #2636, opened 2025-08-05
Decrease the Live ISO size

lslezak:
## Problem

- After switching to Wayland and Gnome kiosk the Live ISO size increased significantly

## Solution

- Delete some not needed packages or individual big files 
- Unfortunately most of them are pulled in by some dependencies so they need to be deleted by force (ignoring dependencies)
- It turned out that using the `patternType="onlyRequired"` option in the Kiwi file as [suggested here](https://github.com/agama-project/agama/compare/kiwi-only-required?expand=1) does not change anything. (Probably it applies only to patterns, but we install Wayland and Gnome kiosk as individual packages, we do not use a pattern for that.)

## Result

- The image size decreased from 693MB to 662MB, saving about 31MB (-4.5%).
- We can apply only a subset of the proposal if we consider some removal is dangerous.
-  Saved size for each individual part:
    - python: -11MB
    - libyui-ncurses: -1MB
    - iso-codes: -2MB
    - gtk4-tools: -2MB
    - gnome-control-center: -2MB
    - gnome-themes-accessibility: -2MB
    - gweather4-data: -1MB
    - codecs: -6MB
    - gstreamer: -3MB
    - jsonnet-lint: -3MB
    - Firefox crashreporter: -2MB

### Notes
- Due to rounding the sum of the saved parts do not exactly match the total saved space, the numbers are only approximate.
- The numbers are for the x86_64 architecture, on the other architectures the numbers will be very likely different.

## Testing

- Tested manually, the built Live ISO still boots :smiley:

### DIFF
--- a/live/src/config.sh
+++ b/live/src/config.sh
@@ -253,7 +253,9 @@ if [ -n "$python" ]; then
   echo "Python package: $python"
   python_deps=$(rpm -e "$python" 2>&1 || true)
   # avoid removing python accidentally because of some new unknown dependency
-  python_deps=$(echo "$python_deps" | grep -v -e "Failed dependencies" -e "needed by .* libpython" -e "needed by .* bcache-tools" -e "needed by .* xfsprogs" || true)
+  python_deps=$(echo "$python_deps" | grep -v -e "Failed dependencies" -e "needed by .* libpython" \
+    -e "needed by .* bcache-tools" -e "needed by .* xfsprogs" -e "needed by .* hyper-v" \
+    -e "needed by .* malcontent" -e "needed by .* gnome-shell" || true)
 
   if [ -z "$python_deps" ]; then
     echo "Removing Python..."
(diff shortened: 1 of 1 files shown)


DimStar77:
#### The dep chain leading to gnome-control-center is: 
* rpm -qgdm-systemd -> gnome-shell -> gnome-control-center
The dep generally does make sense, as gnome-shell does integrate the control-center directly into the panel (it's absence should not cause a problem; for TW, we might move that dep to the pattern and remove it from gnome-shell)


#### the dep chain leading to gweather4-data is:
* gweather4-data is a dependency (brings glib schema and location database) for libgweather, linked into gnome-settings-daemon; settings-daemon is needed by mutter and gnome-session; 

#### dep chain leading to python313-base:
* /usr/bin/python3 is needed by hyper-v, xfsprogs, and bache-tools

### some other potential candidates:
#### nodejs22
* nodejs22 is 6MB in size and is only pulled in as a dependency to agama-integration-tests; Is that really needed on the actual live image?

#### epiphany vs Firefox?
* It might be worthy to check if Firefox (~270MB) could be replaced with GNOME Web (aka epiphany, ~ 135MB)

PR #2639, opened 2025-08-05
feat(web): improve layout and accuracy of wired connection details

dgdavid:
This PR introduces several enhancements to the wired connection details view, aiming to improve both the accuracy of the shown data and the usability of the layout, especially in cases where multiple devices share a connection.

To achieve this, the following actions were taken:

  
*   **Fixed connection settings display**  
    The details page now shows the actual connection settings, rather than using those from the first connected device. This drop the mismatch between the details view and the corresponding connection form.

*   **Added "Binding Settings" section**  
    To clearly displays whether the connection is bound to a specific device and how (by MAC address or device name), rather than just showing information for the first device using it.
    
*   **Added support for multiple devices**  
    When multiple devices share a connection, users can switch between them using a tabbed section, making it easier to view and compare individual device details.
    
*   **Reorganized layout for clarity**  
    The layout has been adjusted to better group related information and improve overall readability.

<details>
  <summary>Click to show/hide some screenshots</summary>

  ---

_Before changes_

localhost_8080_ (25)

---

**NOTE: below screenshots might be slightly outdated due to suggestions from code review. They will be updated only if the final result changes noticeable**

_After changes, with only one device using the connection_

localhost_8080_ (30)

---

_After changes, with more than one device sharing the connection_

localhost_8080_ (29)

</details>

---

:point_right: **Follow up work:** do the same for `WifiConnectionDetails`, once this is reviewwed, adjusted, approved, and merged.

---

Related to https://trello.com/c/MqZ19UnL (protected link)

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug  6 12:44:13 UTC 2025 - Lubos Kocman <lubos.kocman@suse.com>
+
+- Drop Beta from Leap 16.0.
+  We've agreed to not use RC and simply go with Leap 16.0
+  just like on boot menu. 
+
+-------------------------------------------------------------------
 Thu Jul 31 07:44:04 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
 
 - Remove selinux policy from products definition as it is noop
--- a/products.d/leap_160.yaml
+++ b/products.d/leap_160.yaml
@@ -1,5 +1,5 @@
 id: Leap_16.0
-name: Leap 16.0 Beta
+name: Leap 16.0
 # ------------------------------------------------------------------------------
 # WARNING: When changing the product description delete the translations located
 # at the at translations/description key below to avoid using obsolete
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -88,6 +88,7 @@ dependencies = [
  "tokio-tungstenite 0.26.2",
  "url",
  "utoipa",
+ "uuid",
  "zbus",
 ]
 
@@ -4694,12 +4695,14 @@ dependencies = [
 
 [[package]]
 name = "uuid"
-version = "1.16.0"
+version = "1.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "458f7a779bf54acc9f347480ac654f68407d3aab21269a6e3c9f922acd9e2da9"
+checksum = "3cf4199d1e5d15ddd86a694e4d0dffa9c323ce759fea589f00fef9d81cc1931d"
 dependencies = [
  "getrandom 0.3.2",
+ "js-sys",
  "serde",
+ "wasm-bindgen",
 ]
 
 [[package]]
--- a/rust/agama-lib/Cargo.toml
+++ b/rust/agama-lib/Cargo.toml
@@ -45,6 +45,7 @@ fluent-uri = { version = "0.3.2", features = ["serde"] }
 tokio-tungstenite = { version = "0.26.2", features = ["native-tls"] }
 tokio-native-tls = "0.3.1"
 percent-encoding = "2.3.1"
+uuid = { version = "1.17.0", features = ["serde", "v4"] }
 
 [dev-dependencies]
 httpmock = "0.7.0"
--- a/rust/agama-lib/src/auth.rs
+++ b/rust/agama-lib/src/auth.rs
@@ -45,7 +45,7 @@ const USER_TOKEN_PATH: &str = ".local/agama/token";
 const AGAMA_TOKEN_FILE: &str = "/run/agama/token";
 
 use std::{
-    fmt::Display,
+    fmt,
     fs::{self, File},
     io::{self, BufRead, BufReader, Write},
     os::unix::fs::OpenOptionsExt,
@@ -56,6 +56,7 @@ use chrono::{Duration, Utc};
 use jsonwebtoken::{DecodingKey, EncodingKey, Header, Validation};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use uuid::Uuid;
 
 #[derive(Error, Debug)]
 #[error("Invalid authentication token: {0}")]
@@ -183,8 +184,8 @@ impl AuthToken {
     }
 }
 
-impl Display for AuthToken {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl fmt::Display for AuthToken {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.0)
     }
 }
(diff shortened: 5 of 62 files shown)


web/src/components/network/WiredConnectionDetails.tsx
```diff
@@ -20,153 +20,260 @@
...
-          <DescriptionListDescription>{device.state}</DescriptionListDescription>
-        </DescriptionListGroup>
-        <DescriptionListGroup>
-          <DescriptionListTerm>{_("MAC")}</DescriptionListTerm>
-          <DescriptionListDescription>{device.macAddress}</DescriptionListDescription>
-        </DescriptionListGroup>
-      </DescriptionList>
+      <Content>
+        {bindingMode === "none"
+          ? sprintf(_("Connection is bind %s."), BINDING_MODES[bindingMode])
```
ancorgs: Too much composing of strings -> problems for the translators.

web/src/components/network/WiredConnectionDetails.tsx
```diff
@@ -20,153 +20,260 @@
...
+          : sprintf(_("Connection is bind %s to %s."), BINDING_MODES[bindingMode], bindingInfo)}
```
ancorgs: Translations here will only be possible for languages with a structure very similar to English.

web/src/components/network/WiredConnectionDetails.tsx
```diff
@@ -20,153 +20,260 @@
...
+  const onlyOne = connectedDevices.length === 1;
+  const multiple = connectedDevices.length > 1;
 
   return (
     <Page.Section
-      title={_("IP settings")}
+      title={onlyOne ? _("Connected device") : _("Connected devices")}
+      pfCardProps={{ isPlain: false, isFullHeight: false }}
+    >
+      {none && _("None")}
```
ancorgs: I guess we could use a more explanatory string other than simply "None".

ancorgs (commented):
This generally looks great! I just have some reservations about the definition and usage of some strings.

web/src/components/network/WiredConnectionDetails.tsx
dgdavid: Thanks for the comment. You're right. Indeed, I was actually not happy with this piece of code, but I then forget to refactor it before opening de PR. Let's say my brain was already empty of energy yesterday 🥲

If you or someone else don't mind, feel free to change it. I thin a simply 


```js

...

if (bindingMode === "none") return ("the text")
if (bindingMode === "mac") return ("the text %s", theMac)
if (bindingMode === "iface") return ("the text %s", theIface)
```

Would work. 

I also forgot comments for translators

web/src/components/network/WiredConnectionDetails.tsx
dgdavid: I fully agree, but don't know what would be better there.

Visually I was thinking in using something similar to an empty state, but I prefer to see a meaningful sentence there first. Help is welcome, thanks.

web/src/components/network/WiredConnectionDetails.tsx
dgdavid: Agree. Read https://github.com/agama-project/agama/pull/2639#discussion_r2256388809

web/src/components/network/WiredConnectionDetails.test.tsx
```diff
@@ -52,50 +52,182 @@ const mockDevice: Device = {
...
-    const section = screen.getByRole("region", { name: "IP settings" });
-    const editLink = within(section).getByRole("link", { name: "Edit" });
-    expect(editLink).toHaveAttribute("href", "/network/connections/Network%20%231/edit");
+  describe("Binding settings section", () => {
+    it("renders information aobut the binding mode", () => {
+      const { rerender } = plainRender(
+        <WiredConnectionDetails connection={new Connection("Network #1")} />,
+      );
+      const section = screen.getByRole("region", { name: "Binding" });
+      within(section).getByText("Connection is bind to any interface.");
```
joseivanlopez: "is bound". The same in more strings below.

web/src/components/network/WiredConnectionDetails.tsx
dgdavid: I've done it.

web/src/components/network/WiredConnectionDetails.tsx
ancorgs: I was thinking about a better sentence and I realized we use the word "interface(s)" at the "Binding" box but the word "device(s)" at the "Connected devices" one. I'm not sure what is more correct, but it makes it hard to establish a mental connection.

ancorgs: Some reflections (some of them pretty subjective) about my comment above. ☝️☝️

The connection between the binding settings and the list of connected devices should be relatively obvious. That's especially important when the list of devices is empty. Changing the binding settings may "fix" the empty list or not, it depends on other factors, but at least it should be somehow "suggested" by the UI.

I personally see the connection in a more obvious way if the boxes are sorted like this, although I understand that's very subjective.

network

Regardless of that, I see the word "interface" is used all over the place in the card to display the binding and in the form to configure it.

iface

So I wonder:

1. Whether we should use "Connected interfaces" instead of "Connected devices" (or the other way around, use "device" instead of "interface" for all the binding part).
2. Whether the "None" string should use the word "interface" or even suggest to take a look to the binding. Maybe the second part is too much and it may make no sense (eg. if no explicit binding is configured). I don't know. Maybe replacing the current "None" with something like "There is no interface associated to this connection".

ancorgs: I guess @teclator has a more authorized opinion about terminology.

dgdavid: I almost fully agree with your comments, although I'm a bit lost with these networking concepts. Just trying to shape a reasonable interface for them.


> "There is no interface associated to this connection".

I was actually thinking on "There is no device (interface?) using this connection", since I understood that interfaces are shown there only if they are "connected". But let's wait for @teclator to help with this.

dgdavid: > I personally see the connection in a more obvious way if the boxes are sorted like this, although I understand that's very subjective.
> network
> 

The sections layout it is, indeed, subjective. But I agree with your proposal until we have good reasons to make changes again. I moved it to the _left column_ in order to

* _Keep editable stuff in the same column_, which means to keep them at the top when the screen size only support one column. Your proposal keeps editable stuff at the top always, which also looks good to me but somehow I though it could be better having them at left in wide screen sizes because next point.
* _Retain space for vertical tabs_, which as my first approach for displaying multiple devices. However, I finally used the horizontal tabs because vertical ones does not work as expected.: tabs at left and content at right. Instead, tabs are laid out vertically and content under the tabs, which looks actually like a bug in the component :confused: 

That explained, I have no strong opinion on keeping them as they were proposed in this PR or going for your proposal if you and @teclator feel it's better. But please, do not forget about the order when layout become one column only.

dgdavid: localhost_8080_ (27)

ancorgs (approved):
LGTM

teclator:
Also LGTM, it really fixes some of the issues we had.

PR #2642, opened 2025-08-06
Retry registration when a network error happens during autoinstallation

lslezak:
## Problem

- In most cases the network is already configured when Agama starts.
- But in some rare scenarios the network needs to be configured from the installation profile itself to get it running.
- The problem is that the NetworkManager configures the network asynchronously (in backgroung), at the point when the system is being registered the network might not be configured yet and the registration fails.
- https://bugzilla.suse.com/show_bug.cgi?id=1246990

image

## Solution

- Pass the registration error details (the error code) from the HTTP client
- The CLI tool then checks the error code and if it is a network error or a timeout then it tries the registration again after a small delay (4 attempts with exponential backoff: 2,4,8,16 second delays)
- The Web UI is not changed, it fails immediately without any retry. If a retry is needed user can just press the button again.

image

## Testing

- Tested manually

### DIFF
--- a/rust/agama-lib/src/product/http_client.rs
+++ b/rust/agama-lib/src/product/http_client.rs
@@ -30,7 +30,7 @@ pub enum ProductHTTPClientError {
     #[error(transparent)]
     HTTP(#[from] BaseHTTPClientError),
     #[error("Registration failed: {0}")]
-    FailedRegistration(String),
+    FailedRegistration(String, Option<u32>),
 }
 
 pub struct ProductHTTPClient {
@@ -113,15 +113,18 @@ impl ProductHTTPClient {
             return Ok(());
         };
 
+        let mut id: Option<u32> = None;
+
         let message = match error {
             BaseHTTPClientError::BackendError(_, details) => {
                 let details: RegistrationError = serde_json::from_str(&details).unwrap();
+                id = Some(details.id);
                 format!("{} (error code: {})", details.message, details.id)
             }
             _ => format!("Could not register the product: #{error:?}"),
         };
 
-        Err(ProductHTTPClientError::FailedRegistration(message))
+        Err(ProductHTTPClientError::FailedRegistration(message, id))
     }
 
     /// register addon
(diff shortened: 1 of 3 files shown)


rust/agama-lib/src/product/http_client.rs
```diff
@@ -30,7 +30,7 @@ pub enum ProductHTTPClientError {
     #[error(transparent)]
     HTTP(#[from] BaseHTTPClientError),
     #[error("Registration failed: {0}")]
-    FailedRegistration(String),
+    FailedRegistration(String, Option<u32>),
```
mvidner, line 34: Add a comment before the `#[error]:
// If present, the number is already printed in the String part

rust/agama-lib/src/product/store.rs
```diff
@@ -117,6 +126,52 @@ impl ProductStore {
...
+                    match error {
+                        ProductHTTPClientError::FailedRegistration(_msg, code) => {
+                            match code {
+                                // see service/lib/agama/dbus/software/product.rb
+                                // 4 => network error, 5 => timeout error
+                                Some(4) | Some(5) => {
+                                    attempt += 1;
+                                    if attempt > RETRY_ATTEMPTS {
+                                        // still failing, report the error
+                                        return Ok(result?);
```
mvidner: `result` is an `Err` in all 3 cases, right?
```suggestion
                                        return result;
```

rust/agama-lib/src/product/store.rs
```diff
@@ -24,6 +24,13 @@ use crate::{
     http::BaseHTTPClient,
     manager::http_client::{ManagerHTTPClient, ManagerHTTPClientError},
 };
+use std::time;
+use tokio::time::sleep;
+
+// registration retry attempts
+const RETRY_ATTEMPTS: u32 = 3;
+// delay between attempts
+const RETRY_DELAY: time::Duration = time::Duration::from_secs(10);
```
mvidner: NP: I would prefer a slightly more complex but user friendlier retry scheme: with an exponential backoff:
delay for 1, 2, 4, 8, 16 seconds?

rust/agama-lib/src/product/store.rs
```diff
@@ -117,6 +126,52 @@ impl ProductStore {
...
+                                // 4 => network error, 5 => timeout error
+                                Some(4) | Some(5) => {
+                                    attempt += 1;
+                                    if attempt > RETRY_ATTEMPTS {
+                                        // still failing, report the error
+                                        return Ok(result?);
+                                    }
+
+                                    // wait a bit then retry (run the loop again)
+                                    println!(
```
mvidner: I haven't checked, is stdout the correct way for feedback here, not stderr?

rust/agama-lib/src/product/store.rs
lslezak: OK, makes sense. In the end I used 4 retries with 2, 4, 8, 16 seconds.  I think 5 attempts would be too much.

(Do not forget that the retry happens also in case of a timeout error. If a timeout happens after 30 seconds then the real retry time is 30+backoff, i.e. 31, 32, 34, 38 and 46. That would be too long IMHO...)

mvidner: Oh, the "built-in timeout" undermines my point, but it happens only in some cases and the resulting delay sequence is still acceptable, thanks

PR #2651, opened 2025-08-11
fix(rust): fix set_config_locales test

imobachgs:
## Problem

After introducing the [client ID](https://github.com/agama-project/agama/pull/2627), the `test_set_config_locales` does not work anymore.

## Solution

Add the extension to the request. It should fix [bsc#1247744](https://bugzilla.suse.com/show_bug.cgi?id=1247744).

## Testing

- Added a new unit test
- Tested manually


teclator (approved):
LGTM

PR #2653, opened 2025-08-12
fix(web): move progress details to separate line below message

### LINKED ISSUE #2647: [Suggestion] Place the progress indicator numbers in a different row
Hi,

Just a small improvement/suggestion ticket. Today I used the new Agama installer, it’s really awesome, thanks for your work! :)

When I tried to follow the progress numbers with my eyes, it was a bit difficult because they were zigzagging due to the varying lengths of the package names. This caused the (xxxx/yyyy) progress numbers to quickly zig-zag back and forth horizontally, making them hard to read.

I made a short video to better demonstrate the issue, as well as an image showing where I think the progress numbers should be aligned instead.

Video demonstrating the problem:
https://github.com/user-attachments/assets/67dc6e77-a95b-4ad8-8d73-5b9a7b7b4e4c

Image showing suggested alignment:
Image

dgdavid:
To prevents the numbers from zig-zagging horizontally due to varying message lengths during software installation. It also provides more space for displaying package names.

Fixes  #2647

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug 12 09:45:01 UTC 2025 - David Diaz <dgonzalez@suse.com>
+
+- Split progress details into two lines for better readability
+  (gh#agama-project/agama#2647).
+
+-------------------------------------------------------------------
 Tue Aug  5 23:28:43 UTC 2025 - David Diaz <dgonzalez@suse.com>
 
 - Improve layout and accuracy of wired connection details view


dgdavid:
I've not clear if still using these parenthesis surrounding the numbers or not after this change. 



Screenshot From 2025-08-12 07-27-33
Screenshot From 2025-08-12 07-27-43
Screenshot From 2025-08-12 07-27-20
Screenshot From 2025-08-12 07-27-53

dgdavid:
> I've not clear if still using these parenthesis surrounding the numbers or not after this change.
> 
>

As discussed in a daily call, let's keep format as it was and simply move these numbers below.

imobachgs (approved):

PR #2665, opened 2025-08-18
fix(live): add 'Boot from Hard Disk' to ppc64le

imobachgs:
## Problem

* [bsc#1248161](https://bugzilla.suse.com/show_bug.cgi?id=1248161)

The option "Boot from Hard Disk" is not available for `ppc64le`. Actually, it was not handled when we set that option as the default (see https://github.com/agama-project/agama/pull/2632) because it did not exist.

## Solution

By now, I have converted the "local" option to "Boot from Hard Disk". At this point it is the same option that [we have for aarch64](https://github.com/agama-project/agama/blob/master/live/config-cdroot/fix_bootconfig.aarch64#L79-L81).

However, in ppc64le we have an extra option [Exit to Open Firmware](https://github.com/agama-project/agama/blob/master/live/config-cdroot/fix_bootconfig.ppc64le#L128-L130) which is basically the same (`exit`). For that reason, I decided remove it.

## Testing

- Manually tested (see the video below).

## Screenshots

https://github.com/user-attachments/assets/b64daf70-a5cd-4b33-8698-bf8d3a854afd

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Mon Aug 18 15:48:09 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
-- Add a "Boot from Hard Disk" option for ppc64le (bsc#1248161).
+- Add a "Boot from Hard Disk" option for ppc64le and drop the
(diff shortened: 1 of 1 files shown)


ancorgs (approved):
LGTM

PR #2670, opened 2025-08-20
Implemented --local option to allow profile validation without accessing agama server

mchf:
## Problem

Originally agama-cli performed profile validation by asking agama server through HTTP API. This might complicate migration from older products sometimes.

## Solution

Implemented --local option so you can ask ```agama config --local validate``` for checking the profile and CLI won't send it to agama server.

## Testing

- *Tested manually*


rust/agama-cli/src/config.rs
```diff
@@ -148,6 +157,33 @@ pub async fn run(
...
+        ConfigCommands::Validate { url_or_path } => validate_json(url_or_path, opts.insecure),
+        _ => {
+            eprintln!("This subcommand doesn't support --local option");
+            Ok(())
+        }
+    }
+}
+
+/// Validates a JSON profile with locally available tools only
+fn validate_json(url_or_path: CliInput, insecure: bool) -> anyhow::Result<()> {
```
jreidinger: I do not like the name as there are already validate...maybe I would use something like `validate_locally` ? that emphasize, that it is validation for local only purpose.

jreidinger (approved):

mchf:
Moving --local option into agama config validate was discussed with @imobachgs off the github

PR #2672, opened 2025-08-22
feat(web): add support for sortering to device selector

dgdavid:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`


web/src/utils.ts
```diff
@@ -178,6 +179,32 @@ const generateEncodedPath = (...args: Parameters<typeof generatePath>) => {
...
+ * A lightweight wrapper around `fast-sort`.
+ *
+ * Rather than using `fast-sort`'s method-chaining syntax, this function accepts
+ * the sort direction (`"asc"` or `"desc"`) as a direct argument, resulting in a
+ * cleaner and more declarative API for Agama components, where sorting is often
+ * built dynamically.
+ *
+ * @example
+ * ```ts
+ * sortCollection(devices, 'size', 'asc');
```
ancorgs: I fear these examples have the arguments in the wrong order.  Shouldn't it be direction first and key later?

dgdavid: Right, these examples kept outdated. Initially I had that order (more natural to me) but finally I decided to keep the same order that the library we're using under the hood at this moment. I'll update them, thanks for catching.

web/package/agama-web-ui.changes
```diff
@@ -1,7 +1,13 @@
+-------------------------------------------------------------------
+Tue Sep 2 08:24:38 UTC 2025 - David Diaz <dgonzalez@suse.com>
```
ancorgs, line 2: I see you were cleaning surplus spaces. Great!
But I was actually expecting two spaces between "Sep" and "2" (I think that's the default in this format for days with just one digit). But I don't think that really matters.

ancorgs (approved):
LGTM

web/package/agama-web-ui.changes
dgdavid: > But I was actually expecting two spaces between "Sep" and "2"

Me too.

Originally I "kept" it. In fact, I added it because I solved the conflict manually using the Github web interface... but CI complained about wrong date and the only thing I found the extra space. I can add it back, though.

ancorgs (approved):
Still LGTM after the latest merge. Hurry up and merge before another changelog conflict kicks in. :-)

PR #2685, opened 2025-08-27
feat: display config load error

imobachgs:
## Problem

When Agama cannot load a configuration, it offers not details about the problem.

## Solution

config-load-error

Display the output of the "agama config" commands. We would like to have a better
representation, but it is good enough to do not force the users to check the logs.

### DIFF
--- a/web/src/components/questions/LoadConfigRetryQuestion.test.tsx
+++ b/web/src/components/questions/LoadConfigRetryQuestion.test.tsx
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) [2025] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import { screen } from "@testing-library/react";
+import { plainRender } from "~/test-utils";
+import { Question } from "~/types/questions";
+import LoadConfigRetryQuestion from "~/components/questions/LoadConfigRetryQuestion";
+
+const question: Question = {
+  id: 1,
+  text: "It was not possible to load the configuration from http://wrong.config.file. It was unreachable or invalid. Do you want to try again?",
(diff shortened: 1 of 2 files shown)


dgdavid:
Hi @imobachgs.

I've tried to make some improvements to the structure and formatting of the question, keeping the following in mind:

*   Use a title  
    This improves scannability and gives users immediate context thanks to its prominent size. It also acts as a summary of the issue.
    
*   Use the `isEditorial` prop for the first paragraph
    Slightly increases font size to improve readability of the main message.
    
*   Add paragraph spacing
    Improves readability and visual clarity by separating content into distinct sections.
    
*   Reposition or drop the follow-up question**  
    IMHO, the "Do you want to try again?" question should be closer to the actions. Eventually, it could even be drop in favor of clearer button labels, such as \[Try again\] and \[Continue without configuration\] or \[Skip configuration file\].
    
*   Avoid redundancy between title and first paragraph  
    The goal is to make the first line less repetitive with the title.
    

However, I couldn’t address the last two points because the first paragraph currently comes directly from the raw error message :/

Ideally, the error content would be separated from the user-facing message, allowing us to write clearer, more helpful copy.

What do you think? Would it be possible to change or adapt the error message so it only states the error?


The idea is to have something closer to below examples, 

--

### Configuration unreachable or invalid

It was not possible to load the configuration from http://wrong.config.file

```
Could not generate the configuration: Retrieving data from URL http://wrong.config.file

Caused by:
    0: Could not retrieve http://wrong.config.file/
    1: [6] Could not resolve hostname (Could not resolve host: wrong.config.file)
```

Do you want to try again?

[Yes] [No]

---

or even better 

---

### Configuration unreachable or invalid

It was not possible to load the configuration from http://wrong.config.file

```
Could not generate the configuration: Retrieving data from URL http://wrong.config.file

Caused by:
    0: Could not retrieve http://wrong.config.file/
    1: [6] Could not resolve hostname (Could not resolve host: wrong.config.file)
```

What would you like to do?

[Try again] [Continue without configuration] (or "Skip configuration")

imobachgs:
> Feel free to change or even revert these changes if they don’t meet your expectations.
> 

It looks much better now, thanks!

> On another note, I’ll be happy to help with testing once we agree on whether the other two points are feasible.

No, it is too late to change the texts (translations), sorry. We can work on that for Agama 18.

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -99,7 +99,7 @@ impl ScriptsRunner {
         let mut file = Self::create_file(&path, 0o700)?;
         while let Err(error) = Transfer::get(url, &mut file, self.insecure) {
             eprintln!("Could not load configuration from {url}: {error}");
-            if !self.should_retry(&url).await? {
+            if !self.should_retry(&url, &error.to_string()).await? {
```
jreidinger, line 102: cannot we just pass error?

imobachgs: > cannot we just pass error?

Yes, sure, but we need to pass a trait object (you can get different kind of errors) or use generics. At the end of the day, you are only interested in the message, so I decided to pass only the representation.

jreidinger: so lets keep it as it is.

PR #2686, opened 2025-08-28
fix: update software config cache on SoftwareProposalChanged

imobachgs:
## Problem

- [bsc#1248826](https://bugzilla.suse.com/show_bug.cgi?id=1248826)

Using the following configuration resets the pre-selected patterns (`selinux`):

```jsonnet
{
  software: {
    patterns: {
      add: ["gnome_basic"]
    }
  }
}
```

The problem is that the software configuration cache is not updated when the list of patterns changes.

## Solution

Update the list of patterns when the cache on the `SoftwareProposalChanged` event.

## Testing

- Tested manually

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Aug 28 10:15:23 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Update the software cache when the patterns selection changes (bsc#1248826).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rust/agama-server/src/software/web.rs
```diff
@@ -226,16 +226,41 @@ fn reason_to_selected_by(
...
+            EventPayload::LocaleChanged { locale: _ } => {
+                let mut cached_products = products.write().await;
+                if let Ok(products) = client.products().await {
+                    *cached_products = products;
+                } else {
+                    tracing::error!("Could not update the products cached");
+                }
             }
+
+            EventPayload::SoftwareProposalChanged { patterns } => {
```
jreidinger, line 243: I think in future we will need to react on more events if we keep this cache.

PR #2693, opened 2025-09-01
fix(web): map zh_* to zh_Han*

imobachgs:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1248876

The description of the Chinese languages should not include the territory.

## Solution

Map `zh-CH` to `zh_Hans` and `zh-TW` to `zh-Hant`.

## Testing

- *Tested manually*

Here is the content of the `locales.json` after applying this change:

```json
{
  "ca-ES": "Català",
  "de-DE": "Deutsch",
  "en-US": "English",
  "es-ES": "Español",
  "fr-FR": "Français",
  "id-ID": "Indonesia",
  "it-IT": "Italiano",
  "ja-JP": "日本語",
  "ka-KZ": "ქართული",
  "pt-BR": "Português",
  "ru-RU": "Русский",
  "sv-SE": "Svenska",
  "tr-TR": "Türkçe",
  "uk-UA": "Українська",
  "zh-CN": "简体中文",
  "zh-TW": "繁體中文"
}
```

### DIFF
--- a/web/share/update-languages.py
+++ b/web/share/update-languages.py
@@ -45,9 +45,6 @@ class Locale:
     def code(self):
         return f"{self.file_language}-{self.territory}"
 
-    def normalized_language(self):
-        return Locale.languages_map.get(self.code(), self.language)
-
     def name(self, include_territory: bool = False):
         if include_territory:
             return language_name(languageId=self.language,
--- a/web/share/update-languages.py
+++ b/web/share/update-languages.py
@@ -35,12 +35,18 @@ class Locale:
     language: str
     territory: Optional[str]
 
+    languages_map = { "zh-CN": "zh_Hans", "zh-TW": "zh_Hant" }
+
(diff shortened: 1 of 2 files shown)


web/share/update-languages.py
```diff
@@ -35,12 +35,18 @@ class Locale:
...
-        self.language = language
+        self.file_language = language
         self.territory = territory
+        self.language = Locale.languages_map.get(self.code(), language)
 
     def code(self):
-        return f"{self.language}-{self.territory}"
+        return f"{self.file_language}-{self.territory}"
+
+    def normalized_language(self):
```
jreidinger: why this code is needed? I see almost identical call that sets already self.langauge to "normalized" one.

imobachgs: Sorry, it was just dead code.

jreidinger (approved):

PR #2704, opened 2025-09-05
testing fix POC

jreidinger:
POC with build fix ( POC for pattern names method now )

zypp c counterpart is https://github.com/agama-project/zypp-c-api/pull/12

PR #2711, opened 2025-09-08
Another concurency attempt

jreidinger:
Draft to ensure zypp runs on one specific thread.

Another attemps require `unsafe Impl Send` which is not guaranteed by libzypp guys.

another attempt with unsafe impl https://github.com/agama-project/agama/pull/2704


mvidner:
> Another attemps require unsafe Impl Send which is not guaranteed by libzypp guys.

The way I understand `unsafe impl Send for Foo {}`, when you declare it, you say "Foo may contain parts that cannot be shared (a zypp pointer), but I have added synchronization which means that Foo the wrapper is safe to share"

jreidinger:
> > Another attemps require unsafe Impl Send which is not guaranteed by libzypp guys.
> 
> The way I understand `unsafe impl Send for Foo {}`, when you declare it, you say "Foo may contain parts that cannot be shared (a zypp pointer), but I have added synchronization which means that Foo the wrapper is safe to share"

no, this is wrong...
basically

Send - it is safe to move object between threads. So e.g. it can be dropped from different thread then original one where it is allocated. So code can be executed in different times in different thread.
Sync - it is safe to execute methods with object across thread. So code can be executed in same time in different threads ( here mutex helps as it ensure that it can be done for code that does not have that guarantee ).

mvidner:
@jreidinger On a closer look, the test assertion is simply wrong, matching earlier incomplete product definition. `Some("16.0")` is the expected value now.

rust/agama-server/src/software_ng/backend/server.rs
```diff
@@ -237,28 +255,28 @@ impl SoftwareServiceServer {
...
+        Ok(zypp)
     }
 
-    fn import_gpg_keys(&self) {
+    fn import_gpg_keys(&self, zypp: &zypp_agama::Zypp) {
         for file in glob::glob(GPG_KEYS).unwrap() {
             match file {
                 Ok(file) => {
-                    if let Err(e) = zypp_agama::import_gpg_key(&file.to_string_lossy()) {
+                    if let Err(e) = zypp.import_gpg_key(&file.to_string_lossy()) {
```
mvidner, line 281: Not related to concurrency, but now I realize that `import_gpg_key` should probably want ` mut Zypp`, not just `&Zypp`.
For future PRs.

rust/agama-server/src/software_ng/backend/server.rs
```diff
@@ -83,17 +83,29 @@ impl SoftwareServiceServer {
...
-            }
+        // see https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html#use-inside-tokiospawn for explain how to ensure that zypp
+        // runs locally on single thread
+
+        let rt = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap();
+
+        std::thread::spawn(move || {
```
mvidner, line 96:
```suggestion
        // drop the returned JoinHandle: the thread will be detached
        // but that's OK for it to run until the process dies
        std::thread::spawn(move || {
```

rust/agama-server/src/software_ng/backend/server.rs
jreidinger: why it would want `mut Zypp`? I think nothing should change zypp pointer and it is only attribute of Zypp structure.

rust/agama-server/src/software_ng/backend.rs
```diff
@@ -96,10 +96,10 @@ pub struct SoftwareService {}
 
 impl SoftwareService {
     /// Starts the software service.
-    pub async fn start(
```
mvidner, line 99: (the async is fine, but)
I get a test failure on the doctest above on line 90, which is clearly just a sketch and not intended to compile as is

```rs
let client = SoftwareService::start(products, http, events_tx).await;
```

jreidinger: I can make it build again with below diff, but still it failed with private module.

```
diff --git a/rust/agama-server/src/software_ng/backend.rs b/rust/agama-server/src/software_ng/backend.rs
index 80e3eb0f1..a9b557a03 100644
--- a/rust/agama-server/src/software_ng/backend.rs
+++ b/rust/agama-server/src/software_ng/backend.rs
@@ -83,15 +83,20 @@ pub enum SoftwareServiceError {
 /// ```no_run
 /// # use tokio_test;
 /// use agama_server::{
-///   software::backend::SoftwareService
-/// };
+///     software_ng::backend::SoftwareService,
+///     products::ProductsRegistry
+/// }
 ///
 /// # tokio_test::block_on(async {
-/// let client = SoftwareService::start(products, http, events_tx).await;
+/// let (events_tx, _) = tokio::sync::broadcast::channel(16);
+/// let products = ProductsRegistry::load().expect("Could not load the products registry.");
+/// let products = std::sync::Arc::new(std::sync::Mutex::new(products));
+/// let client = SoftwareService::start(events_tx, products);
 ///
 /// let products = client.get_products().await
 ///   .expect("Failed to get the products");
 /// # });
+/// ```
 pub struct SoftwareService {}
 
 impl SoftwareService {

```

jreidinger: in general I suggest to drop this code as we can see usage of code by simple grepping crate and having it visible outside of crate is useless as it not usable outside.

rust/agama-server/src/software_ng/backend/server.rs
mvidner: I meant to express that `import_gpg_key` changes the state of the Zypp. So this is a case of interior mutability?

PR #2712, opened 2025-09-08
remove PXE images

teclator:
## Problem

It has been agreed to use the Live images for **PXE** installations using only the **squashfs** images. Therefore, the **PXE** images should be removed to avoid confusions about the images to be used in this kind of deployments.

- https://bugzilla.suse.com/show_bug.cgi?id=1238848#c66

## Solution

Remove the PXE images from the kiwi definition


live/live-root-PXE/etc/dracut.conf.d/90-brd.conf
```diff
@@ -1 +0,0 @@
-add_drivers+=" brd "
```
lslezak, line 1: Um, should we then add the driver to the Live ISO? 🤔

lslezak (approved):
LGTM, we can add the `brd` driver later if needed...

PR #2713, opened 2025-09-08
try to fix vendoring of submodules

jreidinger:
By adding directly members of zypp-C-API cargo vendor should see it and parse it.


lslezak (approved):
Looks reasonable, let's see if that helps...

PR #2725, opened 2025-09-11
Update Leap Micro 6.2 product (boo#1250435)

lkocman:
Update patterns and packages Leap Micro 6.2
    
* Drop traditional
* Fixes install issues of Leap Micro 6.2 (boo#1250435)
* Explicitly add opensuse-migration-tool
* Tested also with all optional packages
* Add most packages from agama-full kiwi profile

image


jreidinger:
@lkocman I am not sure if branch is correct. If it is for Leap16 iso, then it probably should go to Leap-16.0 branch? @lslezak can know for sure.

lkocman:
I think we should in fact merge both, otherwise the Leap Micro 6.X yaml is unusable

PR #2726, opened 2025-09-12
Reset to the text console before exiting grub2 (bsc#1248167)

lcp:
The 'Boot from Hard Disk' menu entry exits grub2 and gives back the control to the system firmware, e.g. BIOS or UEFI. Without resetting the terminal output to the text console, SeaBIOS may not clean up the screen and output messages correctly.

Ref: https://bugzilla.suse.com/show_bug.cgi?id=1248167


lcp:
> @ancorgs this code is only for non-efi, so still relevant...just condition is not needed.

Yes, the fix is for the legacy BIOS. I'll remove the condition since "Boot from hard disk" is dropped for EFI.

PR #2729, opened 2025-09-12
First software fixes: can get past product selection

jreidinger:
## Problem

The first iso for revamped even failed to start web server.

## Solution

Get it into shape by:

- [x] remove remaining software dbus service definition to prevent start of it.
- [x]  fix loading products without registration field.
- [x] convert ruby dbus client to http client at https://github.com/agama-project/agama/blob/software-service/service/lib/agama/dbus/clients/software.rb

## Testing

- *Added a new unit test*
- *Tested manually* using https://build.opensuse.org/project/show/systemsmanagement:Agama:branches:first_software_fixes


service/lib/agama/storage/manager.rb
```diff
@@ -135,7 +135,7 @@ def probe(keep_config: false, keep_activation: true)
           _("Calculating the storage proposal")
         )
 
-        product_config.pick_product(software.selected_product)
+        product_config.pick_product(software.config["product"])
```
mvidner, line 138: The Ruby tests fail in places where the software client was switched from D-Bus to HTTP. Please fix that, the test file already contains similar mocks for other HTTP services:

service/test/agama/storage/manager_test.rb

mvidner (approved):

PR #2737, opened 2025-09-15
fix(storage): generate expected JSON when using current size

joseivanlopez:
## Problem

The storage config is not properly exported to JSON when the size is configured to use the current size of the device, see https://github.com/agama-project/agama/blob/master/doc/storage_proposal_from_profile.md#using-current.

As side effect, devices marked to be shrunk in the UI were actually configured to shrink and grow at the same time. The proposal currently fails with such a config.  

The problem only affects to the proposal calculated using the UI and it started to fail since this PR https://github.com/agama-project/agama/pull/2622. Although the real source of the problem (wrong exported JSON) has been always there.

## Solution

Generate the correct JSON when using the current size of a device.

## Testing

* Unit tests adapted
* Tested manually

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Mon Sep 15 13:50:54 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Fix JSON conversion for not configured sizes (bsc#1249591).
+
+-------------------------------------------------------------------
 Fri Sep 12 08:14:28 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
 - Update translations (bsc#1249378)


service/test/agama/storage/config_conversions/to_json_conversions/examples.rb
```diff
@@ -369,6 +369,29 @@
       )
     end
 
+    context "if min size is not configured" do
+      let(:size) do
```
ancorgs, line 373: Is this `let` needed?

joseivanlopez: I think so, otherwise the config would have a default size.

service/test/agama/storage/config_conversions/to_json_conversions/examples.rb
```diff
@@ -386,6 +409,28 @@
...
+    context "if max size is not configured" do
+      let(:size) do
+        {
+          min: "1 GiB"
+        }
+      end
+
+      before do
+        config.size.max = nil
+      end
```
ancorgs, line 421: Why using both the `let` and `before`? Are not them configuring the same `size`?

joseivanlopez: It is because how the tests are designed. I didn't want to use "max": "current" but directly setting it to nil.

ancorgs: Ok. Now I understood the role of `size` and `config.size` in those tests (I had to read `partition_test.rb` for that) and I see the difference. Sorry for the noise.

service/test/agama/storage/config_conversions/to_json_conversions/examples.rb
ancorgs: I though this was nested in a context that already contained basically the same definition. But I may be wrong (and anyways being explicit will not kill us).

PR #2738, opened 2025-09-15
[PoC] [WIP] uses react-i18next

dgdavid:
# Agama: switch to react-i18next

As you already know, Agama translations are a bit complex, mainly because they come from multiple sources: the web interface, backend services, and even external libraries such as **libstorage-ng** (which returns *actions* in the language set on the backend side).

Throughout development, we’ve found several i18n related issues which have been handled by increasing/polishing our custom, Cockpit-based translation code. However, this solution still doesn’t offer an easy way to address an old challenge related to the not yet implemented welcome screen: the ability to load and apply translations for the selected language **without** requiring a full app reload.

To tackle this, I’ve started exploring **react-i18next** as an alternative. It’s a mature, actively maintained library that could help us to reduce time invested on maintaining a custom i18n implementation, streamline translation handling,  and enable dynamic language switching.

The good news is that it almost works out of the box. That said, to fully integrate it with the Agama translation stack, I’ll need help from someone with deeper knowledge of how translations are currently wired throughout the system.

https://gist.github.com/user-attachments/assets/f35a8cab-2f6f-40bc-9d4f-9ceea4b62690

For reference, please see below the relevant notes from my experiment:

*   **react-i18next** is based on **i18next**, which has a broad ecosystem of plugins and utilities available such as backends, language detectors, extractors, processors, and more. See: [https://www.i18next.com/overview/plugins-and-utils](https://www.i18next.com/overview/plugins-and-utils)
    
*   In the experiment, I used the settings you can check in the diff.
    
*   For testing, I generated locale JSON files using [`i18next-gettext-converter`](https://github.com/i18next/i18next-gettext-converter):
    
    `i18next-conv -l es -s es.po -t web/public/po-es.json`
    
*   Initially, I generated an English JSON file from `agama.pot` using [`msginit`](https://www.gnu.org/software/gettext/manual/html_node/msginit-Invocation.html), before realizing that setting `fallbackLng: false` causes i18next to use keys as fallback:
    
    > `msginit --input=agama.pot --locale=en --no-translator --output-file=web/public/po-en.po`
    
*  From what I understand, plural handling in **i18next** differs slightly from (n)gettext. The `t()` function doesn’t require a plural key by default—just a `count` variable. As a result, if a plural form is missing from the translation file, it will silently fall back to the singular form.

   That said, when using the [i18next-parser](https://github.com/i18next/i18next-parser):

   > `i18next 'src/**/*.{ts,tsx}'`

   It correctly extracts and generates plural keys. For example:

   ```ts
   t("{{count}} languages supported", { count: 4 })
   ```

   Produces the following keys in the output JSON:

   ```json
   {
     "{{count}} languages supported_one": "",
     "{{count}} languages supported_other": ""
   }
   ```

*   **The caveat:** developer-introduced texts alone won’t be enough to fully render the app in the default language (English), because plural forms are no longer present in the source code. Instead, they only would exist in the translation files.
    
    Is this a limitation, or could it be seen as a benefit? It might help ensure that even the default language is reviewed and localized properly, rather than assuming source strings are final. Check https://github.com/i18next/i18next/issues/1690 and https://github.com/i18next/i18next/issues/1096
    
    Still, if this behavior becomes problematic, perhaps it could be improved or automated with a custom plugin: https://www.i18next.com/misc/creating-own-plugins)](https://www.i18next.com/misc/creating-own-plugins

* For translations comming from backend, we can implement a good sync mechanism now that API is being rewritten. Something to manage them via query-cache or so, able to react to an _onBackendLanguageChanged_ or event.


web/src/components/product/ProductSelectionPage.tsx
```diff
@@ -136,7 +138,7 @@ function ProductSelectionPage() {
     // TRANSLATORS: Text used for the license acceptance checkbox. %s will be
     // replaced with the product name and the text in the square brackets [] is
     // used for the link to show the license, please keep the brackets.
-    _("I have read and accept the [license] for %s"),
+    t("I have read and accept the [license] for {{product.name}}", { product: nextProduct }),
```
lslezak, line 141: I'd keep the old way and avoid using the i18next interpolations.  There are several advantages for the old approach:

- Gettext adds the `c-format` flag for such messages and then it checks that all translations also contain the `%s` placeholder. And that check is also done by Weblate, the translators cannot even save a translation without the proper number of `%s` placeholders. That avoids bugs in translations from the very beginning.
- The `%s` is much simpler format and avoids typos or copy&paste errors. What happens if a translator by mistake writes `{{product,name}}` or `{{produtc.name}}` in a translation? With simple `%s` you cannot do much mistakes and if you do then gettext will complain.

web/src/i18next.ts
```diff
@@ -0,0 +1,25 @@
+import i18next from "i18next";
+import { initReactI18next } from "react-i18next";
+import LanguageDetector from "i18next-browser-languagedetector";
+import HTTPApi from "i18next-http-backend";
+
+i18next
```
lslezak, line 6: I do not like when some initialization code is executed directly at `import`. You cannot write a test for such functionality.

I'd wrap it into an `init()` function or something like that and call that function explicitly.

dgdavid: I have not a clear idea nor strong opinion here. It was just copy/pasted form the upstream documentation for doing a quick PoC

lslezak: OK, np as it is still a PoC...

dgdavid:
First of all, thanks for the explanation.

> > If I'm not wrong, that is not possible: "t" is a hook. And that is why it works and immediately change the texts as soon a new translation file is loaded.
> 
> Um, that should be possible, [t()](https://github.com/i18next/i18next/blob/master/src/i18next.js#L445) is just a wrapper around the [translate()](https://github.com/i18next/i18next/blob/master/i18next.js#L510) function which should just return a string in the end... (Hopefully I haven't overlooked something.)

Maybe it was me who overlooked something. BtW, I meant `t` comes from a hook.

> 
> > I'm not sure we should stick with xgettext if it goes against a more React-friendly way of handling translations. For better or worse, React is the framework we're using for the web UI, so it makes sense to align with its ecosystem as much as possible. As mentioned in the PR, there are plugins available that can handle translation extraction.
> 
> I'm fine with changing the tooling if it better matches the used framework. 


> I just do not want to loose the features provided by the current translation stack.

Don't get me wrong. Me neither. Just betting for find the best of two world but avoiding to fight against the framework.

>  I found out that Weblate has native [support for the i18next JSON files](https://docs.weblate.org/en/latest/formats/i18next.html) so that should not be a problem.
> 

Cool.

> However, I can still see benefits of using the GNU gettext and the PO files:
> 
> * The checks for the `%s` format tags as mentioned above
> * Comments for translators which can describe the context of the message or the %s placeholder details, that significantly reduces risk of wrong or confusing translations
> * Location in the source code, in Weblate you can just click a link displayed next to the message to see where exactly the text is located in the Agama sources. That allows the translators to see the context and some related texts around, that can help with writing good translations.
> 
> Another reason for gettext and the PO files is that it is unified for all Agama components, it is also used by the Ruby backend. Using the same approach and the same rules everywhere reduces confusion and mistakes on the translators side.

I see. Let's try put all the pieces together and see what we can do. We can talk about it tomorrow or whenever this task can be moved ahead.

Thanks.

dgdavid:
Closing this PR as the situation has significantly improved. Agama can now translate the interface without requiring a full refresh. This has been addressed in the latest updates: https://github.com/agama-project/agama/pull/3142 and https://github.com/agama-project/agama/pull/3197.

PR #2739, opened 2025-09-15
fix: stop logging the events

imobachgs:
- **fix(rust): do not log the events**
- **fix(ruby): fix logs file permissions**

## Problem

- [bsc#1249622](https://bugzilla.suse.com/show_bug.cgi?id=1249622).

## Solution

- Do not log the events to avoid leaking passwords. We might need a better solution for 16.1, but now we are in a hurry.
- Fix `/var/log/agama-installation/logs.tar.gz`.

### DIFF
--- a/service/lib/agama/storage/finisher.rb
+++ b/service/lib/agama/storage/finisher.rb
@@ -273,6 +273,7 @@ module Agama
           Yast::Execute.locally(
             "agama", "logs", "store", "--destination", path
           )
+          FileUtils.chmod(0o640, Dir.glob("#{path}*"))
         end
 
         def logs_dir
--- a/rust/agama-server/src/web.rs
+++ b/rust/agama-server/src/web.rs
@@ -56,7 +56,6 @@ mod ws;
 use agama_lib::{connection, error::ServiceError, http::Event};
 use common::{IssuesService, ProgressService};
 pub use config::ServiceConfig;
-use event::log_event;
 pub use event::{EventsReceiver, EventsSender};
 pub use service::MainServiceBuilder;
 use std::path::Path;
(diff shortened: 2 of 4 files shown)


service/lib/agama/storage/finisher.rb
```diff
@@ -273,6 +273,7 @@ def collect_logs
           Yast::Execute.locally(
             "agama", "logs", "store", "--destination", path
```
jreidinger: I would say that agama logs store should itself ensure proper permissions of created logs.

service/lib/agama/storage/finisher.rb
```diff
@@ -273,6 +273,7 @@ def collect_logs
...
           )
+          FileUtils.chmod(0o640, Dir.glob("#{path}*"))
```
jreidinger: in yast usually private files use umask 077 so here it would be 0o600, so really only root can read it.

rust/agama-lib/src/manager/http_client.rs
```diff
@@ -106,6 +114,8 @@ impl ManagerHTTPClient {
             ))
         })?;
 
+        let permissions = Permissions::from_mode(0o600);
+        fs::set_permissions(&destination, permissions)?;
```
jreidinger: well, good practice to avoid timing attack and be more atomic is to set permissions directly when opening file with https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.mode
another option is to use set permissions on file before it gets content with https://doc.rust-lang.org/std/fs/struct.File.html#method.set_permissions

jreidinger (approved):

PR #2743, opened 2025-09-16
fix(web): avoid missing translations

dgdavid:
# Problem 1

> The top-level constants are evaluated too early, at that time the
> translations are not available yet and the language is not set.

*Solution*: stop using problematic consts.

# Problem 2

Texts containing double quotes are not marked for translations.

\<lslezak\> said:  I guess the problem is the doubleqotes inside which probably confuse xgettext when collecting translatable strings [...] we use the C language parser for parsing Javascript files: https://github.com/agama-project/agama/blob/master/web/build_pot#L14 [...]  the same problem is at 3 more places in the code https://github.com/search?q=repo%3Aagama-project%2Fagama%20_(%27&type=code :-(

*Solution*: pending

### DIFF
--- a/web/src/components/product/ProductRegistrationPage.tsx
+++ b/web/src/components/product/ProductRegistrationPage.tsx
@@ -61,10 +61,6 @@ import { sprintf } from "sprintf-js";
 import { _ } from "~/i18n";
 
 const FORM_ID = "productRegistration";
-const SERVER_LABEL = _("Registration server");
-const EMAIL_LABEL = _("Email");
-const SCC_SERVER_LABEL = _("SUSE Customer Center (SCC)");
-const CUSTOM_SERVER_LABEL = _("Custom");
 
 const RegisteredProductSection = () => {
   const { selectedProduct: product } = useProduct();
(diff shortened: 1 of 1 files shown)


web/src/components/product/ProductRegistrationPage.tsx
```diff
@@ -61,10 +61,6 @@ import { sprintf } from "sprintf-js";
 import { _ } from "~/i18n";
 
 const FORM_ID = "productRegistration";
-const SERVER_LABEL = _("Registration server");
```
lslezak, line 64: I'd prefer this

```diff
- const SERVER_LABEL = _("Registration server");
+ const SERVER_LABEL = N_("Registration server");
```

and later

```diff
- <DescriptionListTerm>{SERVER_LABEL}</DescriptionListTerm>
+ <DescriptionListTerm>{_(SERVER_LABEL)}</DescriptionListTerm>
```

dgdavid: Ok, as you wish. No strong opinion here.

Actually I though about it, but I simply dropped these constant because we were in kind of a hurry (no longer true) and also due to the incoming changes outlined at #2738.

That said, let me revert the change and follow the path you proposed.

dgdavid: Done

dgdavid: But now we have a bunch of directives for disabling an ESlint rule, https://github.com/agama-project/agama/pull/2743/commits/381eb9b086ad4cb9e009fd28529e73eb5d30ac8d ¯\\_(ツ)_/¯. Not a big deal, but as said, most probably the approach followed here is gonna change.

PR #2746, opened 2025-09-18
fix(web): Preserve installer options values after successful submission

dgdavid:
Ensure installer options form controls are reset to their initial values only on cancellation.  Since the form remains mounted, it should preserve the submitted state when closed after a successful submission.

Fixes  https://bugzilla.suse.com/show_bug.cgi?id=1249636 (protected link)


shundhammer (approved):
LGTM

PR #2747, opened 2025-09-19
integrate agama zypp into agama for easier coordination

jreidinger:
## Problem

having  zypp-c-api as submodule is not practical as we are only users and having it in separate git just makes development harder.

## Solution

Integrate source with a bit different structure into agama. git history is lost, but should not be big deal as it just initial state without any important bug fixes in commit messages.

## Testing

- cargo build pass


mvidner (commented):
Thanks!
- `(cd rust; cargo build)` :heavy_check_mark: 
- `(cd rust; cargo test)` :heavy_check_mark: 
- `testing_using_container.sh` needs fixes, I'll commit them
- the old repo had `clang-format` calls to keep the C++ style consistent, they got lost

rust/zypp-agama/zypp-agama-sys/c-layer/Makefile
```diff
@@ -10,9 +10,15 @@ clean:
 	rm -vf *.o *.a
 
 check:
+	git ls-files | grep '\.[ch]' | \
```
mvidner:
```suggestion
	git ls-files | grep -E '\.(c|h|cxx|hxx)$' | \
```

need better matching or it matches unrelated javascript files from the frontend

rust/zypp-agama/zypp-agama-sys/c-layer/Makefile
```diff
@@ -10,9 +10,15 @@ clean:
...
+	  xargs --verbose clang-format --style=llvm --dry-run
 
 libagama-zypp.a: $(OBJ)
 	$(AR) -crs $@ $^
 
 %.o: %.cxx $(DEPS)
 	$(CXX) -c -o $@ $< $(CXXFLAGS)
+
+format:
+	git ls-files | grep '\.[ch]' | \
```
mvidner:
```suggestion
	git ls-files | grep -E '\.(c|h|cxx|hxx)$' | \
```

mvidner:
Thanks! Now, the clippy CI failure is *mostly* unrelated, but it also includes an error message related to this PR:

```console

error: failed to run custom build command for `zypp-agama-sys v0.1.0 (/home/runner/work/agama/agama/rust/zypp-agama/zypp-agama-sys)`
...
  In file included from lib.cxx:3:
  ./internal/callbacks.hxx:7:10: fatal error: zypp-core/ui/progressdata.h: No such file or directory
```

and it's running on Ubuntu so we will have to move Clippy checking to an openSUSE container too

PR #2756, opened 2025-09-23
Software summary screen

jreidinger:
Goal of this pull request is to get successfully in web UI to the initial summary screen. Mocking and fake values are possible as we need to setup foundation where to start with filling implementation.

Status after this pull request:

- it possible to reach summary screen with fake value about size
- it is not possible to manage patterns
- after click on install, it will do storage part, but failed afterwards

### DIFF
--- a/rust/agama-server/src/software_ng/backend/server.rs
+++ b/rust/agama-server/src/software_ng/backend/server.rs
@@ -52,6 +52,12 @@ pub enum SoftwareAction {
         resolvables: Vec<String>,
         optional: bool,
     },
+    GetResolvables {
+        tx: oneshot::Sender<Vec<String>>,
+        id: String,
+        r#type: ResolvableType,
+        optional: bool,
+    },
 }
 
 /// Software service server.
@@ -161,7 +167,19 @@ impl SoftwareServiceServer {
             } => {
                 let resolvables: Vec<_> = resolvables.iter().map(String::as_str).collect();
                 self.software_selection
-                    .add(&id, r#type, optional, &resolvables);
+                    .set(&id, r#type, optional, &resolvables);
+            }
+
+            SoftwareAction::GetResolvables {
+                tx,
+                id,
+                r#type,
+                optional,
+            } => {
+                let result = self.software_selection
+                    .get(&id, r#type, optional).unwrap_or(vec![]);
+                tx.send(result)
+                    .map_err(|_| SoftwareServiceError::ResponseChannelClosed)?;
             }
         }
         Ok(())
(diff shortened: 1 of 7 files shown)


rust/agama-server/src/software_ng/backend/server.rs
```diff
@@ -161,7 +167,21 @@ impl SoftwareServiceServer {
...
+            SoftwareAction::GetResolvables {
+                tx,
+                id,
+                r#type,
+                optional,
+            } => {
+                let result = self
+                    .software_selection
+                    .get(&id, r#type, optional)
+                    .unwrap_or(vec![]);
```
mvidner: The empty vector on error looks like a prototyping shortcut. Either:
- convert to `Error`
- add a FIXME comment
- explain why it is correct -> ah I see now it is an Option, not a Result

rust/agama-server/src/software_ng/web.rs
```diff
@@ -250,3 +266,36 @@ async fn set_resolvables(
...
+)]
+async fn get_resolvables(
+    State(state): State<SoftwareState>,
+    Path(id): Path<String>,
+    Query(query): Query<HashMap<String, String>>,
+) -> Result<Json<Vec<String>>, Error> {
+    let default = "package".to_string();
+    let typ = query.get("type").unwrap_or(&default);
+    let typ = match typ.as_str() {
+        // TODO: support more and move to Resolvable Kind
```
mvidner:
```suggestion
        // TODO: support more and move to ResolvableKind
```

rust/agama-server/src/software_ng/backend/server.rs
mvidner:
```suggestion
                    .unwrap_or(vec![]); // Option::unwrap is OK
```

jreidinger: this is not Error! It is Option and its method to define default if it is `None` https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

jreidinger: basically None means that software selection have not yet defined resolvables for given id, type and optional flag.

jreidinger: would `unwrap_of_default` method help here with understanding? https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default

mvidner: `unwrap_or_default` would mislead me in the same way as I thought this was a Result

PR #2758, opened 2025-09-25
Software solve commit

jreidinger:
Goal of this non trivial PR is to be able to install system in the most optimistic scenario. It can be without progress, error reporting or anything else. But it should be able to install system and reboot into it.

### DIFF
--- a/rust/agama-lib/src/software/model.rs
+++ b/rust/agama-lib/src/software/model.rs
@@ -49,28 +49,40 @@ impl SoftwareSelection {
         Default::default()
     }
 
-    /// Adds a set of resolvables.
-    ///
-    /// * `id` - The id of the set.
-    /// * `r#type` - The type of the resolvables (patterns or packages).
-    /// * `optional` - Whether the selection is optional or not.
-    /// * `resolvables` - The resolvables to add.
-    pub fn add(&mut self, id: &str, r#type: ResolvableType, optional: bool, resolvables: &[&str]) {
-        let list = self.find_or_create_selection(id, r#type, optional);
-        let new_resolvables: Vec<_> = resolvables.iter().map(|r| r.to_string()).collect();
-        list.resolvables.extend(new_resolvables);
-    }
-
     /// Updates a set of resolvables.
     ///
+    /// * `zypp` - pointer to libzypp to do real action
     /// * `id` - The id of the set.
     /// * `r#type` - The type of the resolvables (patterns or packages).
     /// * `optional` - Whether the selection is optional or not.
     /// * `resolvables` - The resolvables included in the set.
-    pub fn set(&mut self, id: &str, r#type: ResolvableType, optional: bool, resolvables: &[&str]) {
+    pub fn set(
+        &mut self,
+        zypp: &zypp_agama::Zypp,
+        id: &str,
+        r#type: ResolvableType,
+        optional: bool,
+        resolvables: &[&str],
+    ) -> Result<(), zypp_agama::ZyppError> {
         let list = self.find_or_create_selection(id, r#type, optional);
+        // FIXME: use reference counting here, if multiple ids require some package, to not unselect it
+        for res in &list.resolvables {
+            zypp.unselect_resolvable(
+                &res,
+                r#type.into(),
+                zypp_agama::ResolvableSelected::Installation,
+            )?;
+        }
         let new_resolvables: Vec<_> = resolvables.iter().map(|r| r.to_string()).collect();
         list.resolvables = new_resolvables;
+        for res in &list.resolvables {
+            zypp.select_resolvable(
+                &res,
+                r#type.into(),
+                zypp_agama::ResolvableSelected::Installation,
+            )?;
+        }
+        Ok(())
     }
 
     /// Returns a set of resolvables.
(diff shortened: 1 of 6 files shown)


service/lib/agama/http/clients/software.rb
```diff
@@ -39,13 +39,15 @@ def probe
...
+          # post("software/propose", nil)
         end
 
         def install
-          # TODO: implement it
-          post("software/install", nil)
+          http = Net::HTTP.new("localhost")
+          # FIXME: we need to improve it as it can e.g. wait for user interaction.
+          http.read_timeout = 3 * 60 * 60 # set timeout to three hours for rpm installation
+          http.post("/api/software/install", "", headers)
```
mvidner: OK, but the base `post` method also has error checking, please put it in here
```rb
return unless response.is_a?(Net::HTTPClientError)
@logger.warn "server returned #{response.code} with body: #{response.body}"
```

rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -95,6 +96,33 @@ static zypp::ZYpp::Ptr zypp_ptr() {
   return NULL;
 }
 
+void switch_target(struct Zypp *zypp, const char *root,
+                   struct Status *status) noexcept {
+  const std::string root_str(root);
+  try {
+    zypp->zypp_pointer->initializeTarget(root_str, false);
```
mvidner:
```suggestion
    zypp->zypp_pointer->initializeTarget(root_str, false /* rebuild rpmdb: no */);
```

explaining the magic constant, like this or with `const bool rebuild_rpmdb = false;` ...

rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -95,6 +96,33 @@ static zypp::ZYpp::Ptr zypp_ptr() {
...
+  } catch (zypp::Exception &excpt) {
+    status->state = status->STATE_FAILED;
+    status->error = strdup(excpt.asUserString().c_str());
```
mvidner: and then we fall through and set `state` to SUCCEED!
`+ return;`

rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -95,6 +96,33 @@ static zypp::ZYpp::Ptr zypp_ptr() {
...
+  try {
+    zypp::ZYppCommitPolicy policy;
+    zypp::ZYppCommitResult result = zypp->zypp_pointer->commit(policy);
+    status->state = status->STATE_SUCCEED;
+    status->error = NULL;
+    return result.noError();
+  } catch (zypp::Exception &excpt) {
+    status->state = status->STATE_FAILED;
+    status->error = strdup(excpt.asUserString().c_str());
+  }
```
mvidner, line 121: A `return false;` is missing here, and the build script needs to be fixed so that the compiler tells you this early enough

rust/agama-server/src/software_ng/web.rs
```diff
@@ -161,6 +162,24 @@ async fn probe(State(state): State<SoftwareState>) -> Result<Json<()>, Error> {
...
+#[utoipa::path(
+    post,
+    path = "/install",
+    context_path = "/api/software",
+    responses(
+        (status = 200, description = "Installation succeed"),
+        (status = 400, description = "The D-Bus service could not perform the action
+")
+    ),
+    operation_id = "software_probe"
```
mvidner:
```suggestion
    operation_id = "software_install"
```

rust/agama-server/src/software_ng/backend/server.rs
```diff
@@ -218,6 +255,45 @@ impl SoftwareServiceServer {
         })
         .map_err(SoftwareServiceError::LoadSourcesFailed)?;
 
+        let installer_id_string = "installer".to_string();
```
mvidner, line 274: NP: put this new code to a separate `fn`

rust/agama-server/src/software_ng/backend/server.rs
```diff
@@ -187,6 +193,37 @@ impl SoftwareServiceServer {
...
 
+    fn set_resolvables(
+        &mut self,
+        zypp: &zypp_agama::Zypp,
+        id: String,
+        r#type: ResolvableType,
+        resolvables: Vec<String>,
+        optional: bool,
+    ) -> Result<(), SoftwareServiceError> {
+        tracing::info!("Set resolvables for {} with {:?}", id, resolvables);
```
mvidner: Readers may find it annoying to not see `type` and `optional` in the log line?

mvidner (changes requested):

service/lib/agama/http/clients/software.rb
jreidinger: agreed, lets have it even I hope that we can remove this workaround soon.

rust/agama-server/src/software_ng/backend/server.rs
jreidinger: well, usually for me knowing that id and list of resolvables is enough to identify which call it is from code. But of course we can add it.

rust/agama-lib/src/software/model.rs
```diff
@@ -121,22 +123,11 @@ impl SoftwareSelection {
     }
 }
 
+/* Fix tests with real mock of libzypp
```
mvidner:
```suggestion
/* TODO: Fix tests with real mock of libzypp
```

PR #2766, opened 2025-09-29
Rearrange UI of the storage section

ancorgs:
## Problem

The current storage page of Agama allows to configure several common scenario with just a few clicks (although it is still only a first step to the definitive UI we want to have in the future). Its general operation is described at https://agama-project.github.io/docs/user/guides/storage

Current page

But there is a fundamental problem. Users who have not read the document mentioned above sometimes struggle to find their way through the Agama storage UI. The are several factors that contribute to it, but especially there are two circumstances that are crucial:

- The current UI does a very poor job on explaining itself.
- Many traditional (open)SUSE users are biased by years of using the YaST Expert Partitioner and as soon as they see a table representing disks and partitions they try to operate the UI like it would be just a new UI for the known Expert Partitioner.

In fact, if the UI would be more self-explanatory, traditional (open)SUSE users would quickly realize the new UI is more a super-powerful version of YaST Guided Setup than a new version of the Partitioner.

## Solution

This pull request modifies the storage page to make it more self-explanatory and to make the options more understandable.

new-general

Absolutely nothing changes under the hood. All the operations work on the same way, all forms are kept, etc. The changes are just "cosmetic":

- Reorganizing how the information is distributed in the screen.
- Adding some texts to explain the goal of each part of the UI and the workflow.
- Relocating some of the existing options (and exceptionally rewording some labels a bit).

According to all our experiments, these changes dramatically raise the opportunities of new users to understand what they need to do and how, despite not really changing the general philosophy, implementation or operation of the page.

## Details

If needed, we can add a lot of "before vs after" comparisons here, to show how the options are now easier to find and understand.

## Manual testing

An installation ISO incorporating this changes can be downloaded from https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/rethinking-encryption-layout/images/iso/

## In progress

For this pull request to have the required quality, there are still some open tasks like updating the unit tests of the different involved components.


ancorgs:
Closed in favor of #2767

PR #2767, opened 2025-09-29
Rearrange UI of the storage section

ancorgs:
## Problem

The current storage page of Agama allows to configure several common scenario with just a few clicks (although it is still only a first step to the definitive UI we want to have in the future). Its general operation is described at https://agama-project.github.io/docs/user/guides/storage

Current page

But there is a fundamental problem. Users who have not read the document mentioned above sometimes struggle to find their way through the Agama storage UI. The are several factors that contribute to it, but especially there are two circumstances that are crucial:

- The current UI does a very poor job on explaining itself.
- Many traditional (open)SUSE users are biased by years of using the YaST Expert Partitioner and as soon as they see a table representing disks and partitions they try to operate the UI like it would be just a new UI for the known Expert Partitioner.

In fact, if the UI would be more self-explanatory, traditional (open)SUSE users would quickly realize the new UI is more a super-powerful version of YaST Guided Setup than a new version of the Partitioner.

## Solution

This pull request modifies the storage page to make it more self-explanatory and to make the options more understandable.

new-general

Absolutely nothing changes under the hood. All the operations work on the same way, all forms are kept, etc. The changes are just "cosmetic":

- Reorganizing how the information is distributed in the screen.
- Adding some texts to explain the goal of each part of the UI and the workflow.
- Relocating some of the existing options (and exceptionally rewording some labels a bit).

According to all our experiments, these changes dramatically raise the opportunities of new users to understand what they need to do and how, despite not really changing the general philosophy, implementation or operation of the page.

## Details

If needed, we can add a lot of "before vs after" comparisons here, to show how the options are now easier to find and understand.

## Manual testing

An installation ISO incorporating this changes can be downloaded from https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/rethinking-encryption-layout/images/iso/

## In progress

For this pull request to have the required quality, there are still some open tasks like updating the unit tests of the different involved components.


oneeyedc4t:
First I want to thank you for your lengthy reply.  And that you care.  I will help with documentation as soon as I can.

I only have one minor request, that the LUKS+LVM be reconsidered.  This is because not all use cases are insecure or dangerous.  Indeed, one can install a Linux distribution with agama in unencrypted mode, which is quite obviously less than secure if this is the concern.  I intend to run OpenSUSE in my private practice, and there are no contraindications for security in that regard.  As well, running LUKS+LVM as a home user is a very ideal use case because anyone who steals my laptop or desktop is only going to wipe over the data and sell it.

So again, I realize the agama team is busy, and we all appreciate your hard work.  I only ask that it be reconsidered.  If I knew how to program, I'd help do this myself if I was allowed.  (Indeed, I wonder if there is a way to offer a donation to the project, like some Linux FOSS projects, to request the feature, and if so, how much it would likely cost in terms of labor.)

ancorgs:
Since this was not accepted for SLES-16 QU0, I just re-targeted the pull request to the master branch. Will verify everything still works when a new testing image is built.

ancorgs:
> Will verify everything still works when a new testing image is built.

Verified that it works. So I just added the corresponding changelog entry and this should be ready to merge.

PR #2775, opened 2025-10-02
Add a new API to handle issues

imobachgs:
Add a new service to handle issues in the new API. The new service allows any other one to register its issues.

## Registering issues

A service that needs to register issues, receives an `agama_utils::actor::Handler<agama_utils::issue::Service>`. It can update the list of issues through the [Update](https://github.com/agama-project/agama/pull/2775/files#diff-473781875264b7e4d4e408a101a973115f312b6305deff38a2d6e135cdc88082R33) action.

## Dealing with D-Bus services

The new service reads the issues from the D-Bus services too. When booting, it reads the existing issues and starts listening for changes on the `org.opensuse.Agama1.Issues` interfaces.

## Localization issues

As part of this PR, the localization service reports issues now in case of a locale, keymap or timezone is unknown.

## Adapted UI

The code to deal with the issues in the web UI has been adapted to the new API.

## Other smaller changes

- Make `*_db` functions in `ModelAdapter` to return references instead of mutable references.
- In a `TimezoneEntry`, rename `code` to `id` for consistency. Additionally, use `TimezoneId` instead of a string.


rust/agama-utils/src/issue/message.rs
```diff
@@ -0,0 +1,56 @@
...
+    type Reply = HashMap<String, Vec<Issue>>;
+}
+
+// FIXME: consider an alternative approach to avoid pub(crate),
+// making it only visible to the service.
+pub struct Update {
+    pub(crate) list: String,
+    pub(crate) issues: Vec<Issue>,
+    pub(crate) notify: bool,
+}
```
joseivanlopez, line 37: Maybe something like a struct enum?

```rust

pub struct Update(UpdateContent);

struct UpdateContent {
    ...
}
```

imobachgs: JFTR, I added a note because the handler implementation needs to access to the fields but we do not want them to be visible from outside the crate. Alternatively, we could put the messages next to the handlers implementation.

rust/agama-server/src/web/common/issues.rs
```diff
@@ -66,6 +70,8 @@ pub enum IssuesServiceError {
     DBusName(#[from] zbus::names::Error),
     #[error("Could not send the event: {0}")]
     SendEvent(#[from] broadcast::error::SendError<Event>),
+    #[error("Issue conversion error")]
```
joseivanlopez: NP: should we use `#[error(transparent)]`? In general, I am doing so when composing errors from other error type. But I am not sure if that is a common/good practice.

rust/agama-l10n/src/service.rs
```diff
@@ -64,19 +68,57 @@ struct State {
...
+
+    pub fn find_issues(&self) -> Vec<Issue> {
+        let config = &self.state.config;
+        let mut issues = vec![];
+        if !self.model.locales_db().exists(&config.locale) {
+            issues.push(Issue {
+                description: format!("Locale '{}' is unknown", &config.locale),
+                details: None,
+                source: issue::IssueSource::Config,
+                severity: issue::IssueSeverity::Warn,
```
joseivanlopez: Warn or Error?

rust/agama-l10n/src/service.rs
```diff
@@ -120,10 +162,16 @@ impl MessageHandler<message::GetConfig> for Service {
...
-            _ = self.events.send(Event::ProposalChanged);
+        if merged == self.state.config {
+            return Ok(());
         }
+
+        self.state.config = merged;
+        let issues = self.find_issues();
+        _ = self
+            .issues
+            .cast(issue::message::Update::new("localization", issues));
```
joseivanlopez, line 203: Should we propagate or log the possible cast error?

rust/agama-l10n/src/service.rs
```diff
@@ -120,10 +162,16 @@ impl MessageHandler<message::GetConfig> for Service {
...
     async fn handle(&mut self, message: message::SetConfig<Config>) -> Result<(), Error> {
         let merged = self.state.config.merge(&message.config)?;
-        if merged != self.state.config {
-            self.state.config = merged;
-            _ = self.events.send(Event::ProposalChanged);
+        if merged == self.state.config {
+            return Ok(());
         }
+
+        self.state.config = merged;
```
joseivanlopez, line 192: Should the config be updated if there are issues?

rust/agama-server/src/server/types.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Product related issues (product selection, registration, etc.).
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub product: Vec<issue::Issue>,
+    /// Storage related issues.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub storage: Vec<issue::Issue>,
+    /// Software management issues.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub software: Vec<issue::Issue>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
joseivanlopez, line 47: NP: missing documentation for user issues.

rust/agama-server/src/supervisor/start.rs
```diff
@@ -22,13 +22,18 @@ use crate::{
...
+    actor::{self, Handler},
+    issue,
+};
 use tokio::sync::mpsc;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error("Could not start the l10n service")]
     L10n(#[from] l10n::start::Error),
+    #[error("Could not start the issues service")]
```
joseivanlopez: `error(transparent)` ?

rust/agama-utils/src/issue/service.rs
```diff
@@ -0,0 +1,77 @@
...
+pub enum Error {
+    #[error(transparent)]
+    Actor(#[from] actor::Error),
+    #[error(transparent)]
+    Model(#[from] super::model::Error),
+}
+
+pub struct Service {
+    issues: HashMap<String, Vec<Issue>>,
+    events: mpsc::UnboundedSender<IssuesChanged>,
```
joseivanlopez: NP: `events: event::Sender`

rust/agama-utils/src/issue/service.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+#[async_trait]
+impl MessageHandler<message::Update> for Service {
+    async fn handle(&mut self, message: message::Update) -> Result<(), Error> {
+        if message.issues.is_empty() {
+            _ = self.issues.remove(&message.list);
+        } else {
+            self.issues.insert(message.list, message.issues);
```
joseivanlopez, line 68: NP: why not directly insert? It would replace the existing value of the key, right?

rust/agama-utils/src/issue/start.rs
```diff
@@ -0,0 +1,102 @@
...
+pub async fn start(
+    events: event::Sender,
+    dbus: Option<zbus::Connection>,
+) -> Result<Handler<Service>, Error> {
+    let service = Service::new(events);
+    let handler = actor::spawn(service);
+
+    if let Some(conn) = dbus {
+        let monitor = Monitor::new(handler.clone(), conn);
+        tokio::spawn(async move {
```
joseivanlopez: NP: in l10n we have a `monitor::spawn` function.

rust/agama-utils/src/issue/start.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let issues_list = issues.call(message::Get).await.unwrap();
+        assert_eq!(issues_list.len(), 1);
+
+        assert!(events_rx.recv().await.is_some());
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_update_wo_event() -> Result<(), Box<dyn std::error::Error>> {
```
joseivanlopez: what wo_event means?

web/src/api/issues.ts
```diff
@@ -21,18 +21,17 @@
...
-  users: "users/issues",
-  storage: "storage/issues",
-};
+import { Issue, IssuesMap, IssuesScope } from "~/types/issues";
 
 /**
  * Return the issues of the given scope.
  */
-const fetchIssues = (scope: IssuesScope): Promise<Issue[]> => get(`/api/${URLS[scope]}`);
+const fetchIssues = async (): Promise<Issue[]> => {
```
joseivanlopez, line 29: The API returns issues grouped by scope, but the client builds a list. Should we use a list in the API too?

rust/agama-utils/src/issue/monitor.rs
```diff
@@ -60,6 +63,7 @@ impl Monitor {
         Self { handler, dbus }
     }
 
+    /// Run the monitor on a separate Tokio task.
```
joseivanlopez, line 66: NP: this method does not spawn the service. I would create a `spawn` function.

rust/agama-utils/src/issue/start.rs
imobachgs: without (by `w/o`). It is a common abbreviation, but let's use the full word.

rust/agama-utils/src/issue/service.rs
imobachgs: Because I do not want empty lists.

rust/agama-l10n/src/service.rs
imobachgs: Yes, definitely. The configuration could be wrong, but it is a "valid" one. Otherwise, where the issues are coming from?

rust/agama-utils/src/issue/service.rs
```diff
@@ -34,11 +34,11 @@ pub enum Error {
 
 pub struct Service {
     issues: HashMap<String, Vec<Issue>>,
-    events: mpsc::UnboundedSender<IssuesChanged>,
+    events: mpsc::UnboundedSender<Event>,
 }
 
 impl Service {
-    pub fn new(events: mpsc::UnboundedSender<IssuesChanged>) -> Self {
+    pub fn new(events: mpsc::UnboundedSender<Event>) -> Self {
```
joseivanlopez: NP: `events: event::Sender`

web/src/api/issues.ts
imobachgs: For consistency, at API level I prefer to have the issues grouped by scope, as we do with the configuration, the system, the proposal, etc. Even in the future we could just emit an event updating the issues for given scope and, in that case, it looks more natural to me to keep them separated.

But I had the same question when implementing it.

rust/agama-l10n/src/service.rs
joseivanlopez: Ok, it makes sense. But I guess we should not have proposal, right?

rust/agama-utils/src/issue/monitor.rs
imobachgs: OK, let's follow the same approach. However, this code looks confusing to me:

```rust
    let monitor = Monitor::new(handler.clone()).await?;
    monitor::spawn(monitor);
```

The module and the variable are both named `monitor`. What about using a more specific name or just passing the output of `Monitor::new` directly to the `monitor::spawn` function?

rust/agama-l10n/src/service.rs
imobachgs: Not here. It is something we can do in the `cast` function.

rust/agama-l10n/src/service.rs
imobachgs: IMHO it is a warning. The system will mostly work, but I am fine with promoting this to Error. However, what would be a warning them? Does it make sense to have a severity?

rust/agama-server/src/web/common/issues.rs
imobachgs: Me neither. Let's use it.

rust/agama-l10n/src/service.rs
imobachgs: But let's do this on a separate PR. Additionally, I think we should adopt the `tracing` crate here.

web/src/api/issues.ts
imobachgs: Having said that, for me it is fine if we want an array. It is just about moving the conversion to the backend.

rust/agama-l10n/src/service.rs
joseivanlopez: Honestly, I think all our issues are stoppers, right? Probably we are not using warnings yet.

In this case, I would not generate a proposal if there is an issue. If we do that, I see it more like an error instead of a warning.

rust/agama-l10n/src/service.rs
joseivanlopez: I mean, if the cast itself fails (the message cannot be sent). If I am not wrong, we are propating them in other services.

rust/agama-utils/src/issue/monitor.rs
joseivanlopez: Ok, fine. I think the constructor is easy enough to direclty pass it to `monitor::spawn`.

rust/agama-l10n/src/service.rs
imobachgs: If you cannot send a message, you are doomed :-) I would just log it.

imobachgs: OK, you are right. Let's propagate the error (the send about sending the message only).

rust/agama-l10n/src/service.rs
joseivanlopez: We have checked that warnings are currently used for problems with probing. That probing problems could be part of the information of the system instead of an issue. We decided to keep it as it is now, and let's evaluate later whether to remove the severity from our issues.

joseivanlopez (approved):

joseivanlopez (approved):

PR #2778, opened 2025-10-03
Refactoring: Create an http client to handle profile actions

mchf:
## Problem

Not a real problem, just something what was overlooked when during implementation.

## Solution

Create an http client to handle profile validation and other actions to use similar approach as in other cases.


rust/agama-cli/src/cli_input.rs
```diff
@@ -63,23 +63,27 @@ impl CliInput {
     /// so that it works (on localhost) even if server's working directory is different.
     /// See also: [`body_for_web`](Self::body_for_web).
     pub fn add_query(&self, base_url: &mut Url) -> io::Result<()> {
```
imobachgs: Checking the code, this function is not used anymore. Is it OK?

rust/agama-lib/src/profile/http_client.rs
```diff
@@ -0,0 +1,104 @@
...
+        // we use plain text .body instead of .json
+        let mut url = self
+            .client
+            .base_url
+            .join("profile/validate")
+            // unwrap OK: joining a parsable constant to a valid Url
+            .unwrap();
+
+        if let Some(query) = query {
+            url.query_pairs_mut().append_pair(&query.0, &query.1);
```
imobachgs: IMHO, the information included in the query string should be part of the request body itself. In that case, you could even use the `post` method of the base client instead of having to do this workaround (building the URL yourself and using the low-level client which, in my option, should be private).

rust/agama-lib/src/profile/http_client.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        let response = self
+            .client
+            .client
```
imobachgs: The fact that we need to do this workaround (building the URL ourselves and relying on the internal client, which should be private) makes me thing that something is not right.

imobachgs (changes requested):
4. Using `--local` does not work for me (e.g., `agama --local config validate tw.json`). It says I am not authenticated.

--local is valid config specific suboption.

rust/agama-cli/src/cli_input.rs
mchf: most probably that's true ... in fact this function was so weird to me that I even did some drawing to get it ;-)

mchf: However, rust compilation did not complain about unused function, so I kept it for now.

rust/agama-lib/src/profile/http_client.rs
mchf: ok, so I'll "design" some structure to pass in the request body ... However it will clearly require some server side modification(s)

imobachgs: Why? You only need to extract what you want on the server side (the content of the file, the URL or the path).

rust/agama-cli/src/cli_input.rs
imobachgs: Well, it does not complain because it is a library (it is defined in `lib.rs`) so the Rust compiler cannot know if some external crate is using it (even from outside of this repository).

I remove it and everthing built just fine. Please, remove it.

mchf:
> Hi!
> 
> 1. The `*_client` functions are still there. If you want to keep them, it is fine. But, please, find a better name.

... this is "in between step" and in the end I plan to remove them. The main focus currently was on removing those queries.

> 2. Not related to this PR, I do not think `CliInput` should be responsible for converting to a query parameter, a body for a JSON call, etc. But I can live with that until the following refactoring :)

... I think that I'll be able to remove most of those "query" functions in the final cleanup ... at least I hope so.

> 3. The POST actions should use JSON in their bodies instead of a `a=b` format.

... in fact it is posted as json (client.post encapsulates the content as a json). I'll polish it a bit.
> 
> I do not want to make this longer, but it would be nice to at least address points 1 and 3. WDYT?

... definitely that's why I asked before stating it as a final version ;-)

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Nov 10 14:35:25 UTC 2025 - Michal Filka <mfilka@suse.com>
+
+- Created ProfileClient based on HTTP API base client for profile
+  validation requests and refactored / cleaned relevant code
```
imobachgs: Please, add the reference (gh#agama-project/agaam#2778).

PR #2787, opened 2025-10-06
feat: progress service

joseivanlopez:
## Problem

- *Bugzilla link*
- *openQA link*

## Solution

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`


rust/agama-server/src/supervisor/service.rs
```diff
@@ -26,41 +26,107 @@ use crate::supervisor::{
...
+    #[error(transparent)]
     L10n(#[from] l10n::service::Error),
-    #[error("Infallible")]
+    #[error(transparent)]
     Infallible(#[from] Infallible),
 }
 
+#[derive(Clone, Serialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum State {
```
imobachgs, line 53: Is this the equivalent of the old installer "phase"? Just mentioning because we use the word "State" for several things (like Axum and services state). I wonder if it would be a better word.

However, it is not a problem, just a little bit of np.

rust/agama-server/src/supervisor/service.rs
```diff
@@ -26,41 +26,107 @@ use crate::supervisor::{
...
+            state: State::Configuring,
             config: InstallSettings::default(),
         }
     }
+
+    async fn start_install(&mut self) -> Result<(), Error> {
+        self.state = State::Installing;
+        // TODO: translate progress steps.
+        self.progress
+            .call(progress::message::StartWithSteps::new(
```
imobachgs, line 83: If you are not interested in the return value, you can use `cast` instead of `call`. Of course, it would make you not to wait until the message is processed (and perhaps *you want* to wait).

rust/agama-server/src/supervisor/service.rs
```diff
@@ -222,7 +288,12 @@ impl MessageHandler<message::RunAction> for Service {
                 let l10n_message = l10n::message::SetSystem::new(config);
                 self.l10n.call(l10n_message).await?;
             }
-            Action::Install => self.l10n.call(l10n::message::Install).await?,
+            Action::Install => {
+                self.start_install().await?;
```
imobachgs: I would most probably enclose all those steps into a single `install` function. Eventually, it will get longer and longer. But it is something that we can do later.

rust/agama-utils/src/progress/model.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// List of steps
+    pub steps: Option<Vec<String>>,
+    /// Current step
+    pub step: Option<String>,
+    /// Index of the current step
+    pub index: Option<usize>,
+}
+
+impl Progress {
+    pub fn new(scope: String, size: usize) -> Self {
```
imobachgs: If you use `&str` instead of `String`, the API would be a little more flexible and you could do `Progress::new("l10n")` or `Progress::new(SCOPE)` (where `SCOPE` is a const &str).

Additionally, you will save an ugly `clone` when calling it.

rust/agama-utils/src/progress/model.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Progress {
+    /// Scope of the progress
+    pub scope: String,
+    /// Max number of steps
+    pub size: usize,
+    /// List of steps
+    pub steps: Option<Vec<String>>,
+    /// Current step
+    pub step: Option<String>,
+    /// Index of the current step
```
imobachgs, line 35: Do we want to handle the case where the progress has not started yet? What about having the `step` and `index` set to the first step? Would it be too limiting?

rust/agama-utils/src/progress/start.rs
```diff
@@ -0,0 +1,228 @@
...
+                vec![first_step.clone(), second_step.clone(), third_step.clone()],
+            ))
+            .await?;
+
+        let progresses = handler.call(message::Get).await?;
+        let progress = progresses.first().unwrap();
+        assert_eq!(progress.scope, scope);
+        assert_eq!(progress.size, 3);
+        assert_eq!(progress.steps.as_ref().unwrap().len(), 3);
+        assert_eq!(*progress.steps.as_ref().unwrap()[0], first_step);
```
imobachgs: You could do something like this:

```rust
let steps = progress.steps.as_ref().unwrap();
assert_eq!(
    steps,
    &vec![first_step.clone(), second_step.clone(), third_step.clone()]
);
```

imobachgs (commented):
In general, it looks good. Just some minor comments.

rust/agama-server/src/supervisor/service.rs
joseivanlopez: Yes, it represents the "installation state". Do you think that adding documentation is enough?

rust/agama-server/src/supervisor/service.rs
joseivanlopez: I prefer to wait for the progress in order to make sure it is emitted early enough.

rust/agama-utils/src/progress/model.rs
joseivanlopez: Adapted, thanks!

rust/agama-utils/src/progress/message.rs
```diff
@@ -0,0 +1,116 @@
...
+    type Reply = ();
+}
+
+pub struct StartWithSteps {
+    pub scope: String,
+    pub steps: Vec<String>,
+}
+
+impl StartWithSteps {
+    pub fn new(scope: &str, steps: Vec<&str>) -> Self {
```
imobachgs: np: I would pass a reference to a slice (`&[&str]`) instead of the vector. Why? Because it is more flexible and avoids an extra allocation (given that you need to create a new vector anyway).

If you decide to do that, I think you need to replace `to_owned` with `to_string` too. Something like this should work:

```rust
pub fn new(scope: &str, steps: &[&str]) -> Self {
    Self {
        scope: scope.to_string(),
        steps: steps.into_iter().map(ToString::to_string).collect(),
    }
}
```

imobachgs (approved):

rust/agama-utils/src/progress/message.rs
joseivanlopez: Nice!

PR #2788, opened 2025-10-06
implement software queries

jreidinger:
## Problem

The calls that query available or selected packages are mocks only in new software stack.

## Solution

Implement it.

### DIFF
--- a/service/lib/agama/http/clients/software.rb
+++ b/service/lib/agama/http/clients/software.rb
@@ -76,6 +76,8 @@ module Agama
           JSON.parse(get("software/resolvables/#{unique_id}?type=#{type}&optional=#{optional}"))
         end
 
+        # (Yes, with a question mark. Bad naming.)
+        # @return [Array<String>] Those names that are selected for installation
         def provisions_selected?(provisions)
           provisions.select do |prov|
             package_installed?(prov)


rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -381,6 +381,49 @@ bool is_local_url(const char *url, struct Status *status) noexcept {
...
+      status->state = status->STATE_FAILED;
+      status->error = strdup("Internal Error: Package tag is empty.");
+      return false;
+    }
+
+    // look for packages
+    zypp::Capability cap(s_tag, zypp::ResKind::package);
+    zypp::sat::WhatProvides possibleProviders(cap);
+
+    for (auto iter = possibleProviders.begin(); iter != possibleProviders.end();
```
mvidner, line 369: This is a loop over *all* possible providers but it is always exited in the first iteration. I would prefer
```c++
auto iter = possibleProviders.begin();
if (iter != possibleProviders.end()) {
...
}
```
to make it clear it is intentional

service/lib/agama/http/clients/software.rb
```diff
@@ -77,19 +76,18 @@ def get_resolvables(unique_id, type, optional)
           JSON.parse(get("software/resolvables/#{unique_id}?type=#{type}&optional=#{optional}"))
         end
 
-        def provisions_selected?(_provisions)
-          # TODO: implement it, not sure how it should look like
-          []
+        def provisions_selected?(provisions)
+          provisions.select do |prov|
```
mvidner, line 82: In the D-Bus client, it says
```rb
        # @return [Array<Boolean>] An array containing whether each tag is selected or not
```
https://github.com/agama-project/agama/blob/v17/service/lib/agama/dbus/clients/software.rb#L107
```suggestion
          provisions.map do |prov|
```

service/lib/agama/http/clients/software.rb
```diff
@@ -77,19 +76,18 @@ def get_resolvables(unique_id, type, optional)
...
+            package_installed?(prov)
```
mvidner, line 83: Also, this makes one HTTP request per one member of _provisions_. It may be fine for our use cases but we may want a reminder to allow for many provisions?

```suggestion
            # TODO: add bulk API
            package_installed?(prov)
```

mvidner (changes requested):

rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
jreidinger: good catch. My brain was probably on vacation when writing it. Of course it should return true only if any of them are selected like original code did at https://github.com/yast/yast-pkg-bindings/blob/853496f527543e6d51730fd7e3126ad94b13c303/src/Package.cc#L504

jreidinger: so for availability it is ok to finish at first as it means there is at least one available package....for selection, it can be any of them depending on solver and I will fix it.

service/lib/agama/http/clients/software.rb
jreidinger: well, we use it only in one case for Autologin and in fact it will work for both cases :)
https://github.com/agama-project/agama/blob/master/service/lib/agama/dbus/y2dir/modules/Autologin.rb#L179
so it just calls `any?` on result. Maybe we can even change ruby code there?

service/lib/agama/http/clients/software.rb
jreidinger: It was in fact my original plan with axum-extra Query which allows stuff like `software_installed?tag=a&tag=b&tag=c` and return e.g. if any of them is installed or available. I just would like to see more usage in code to design better API as maybe any of tag is not good behavior.

rust/zypp-agama/zypp-agama-sys/c-layer/internal/helpers.hxx
```diff
@@ -9,16 +9,33 @@
...
+/// Macro to help report failure with static string
+#define STATUS_ERROR(status, err_str)                                          \
+  ({                                                                           \
+    status->state = status->STATE_FAILED;                                      \
+    status->error = strdup(err_str);                                           \
+  })
+
+/// Macro to help report failure with own allocated string
+/// which will be later free with free_status method
+#define STATUS_ERR_MSG(status, err_str)                                        \
```
mvidner: Thanks for following up on my suggestion!

The name is terrible :-D `STATUS_ERR_MSG` and `STATUS_ERROR` both include a message, so `MSG` is not helping.

How about this instead?
```c
#define STATUS_ERROR_FORMAT(status, c_format, ...) \
({ \
    status->state = status->STATE_FAILED;                                  \
    status->error = format_alloc(c_format, __VA_ARGS__);             \
  })
```

rust/zypp-agama/zypp-agama-sys/c-layer/internal/helpers.hxx
```diff
@@ -9,16 +9,33 @@
...
+  ({                                                                           \
+    status->state = status->STATE_FAILED;                                      \
+    status->error = strdup(err_str);                                           \
```
mvidner: This is what you meant, right?
```suggestion
    status->error = err_str;                                                   \
```

But I'd like to replace `STATUS_ERR_MSG` with `STATUS_ERROR_FORMAT`anyway

service/lib/agama/http/clients/software.rb
mvidner: You have transposed an API from D-Bus to HTTP while changing its semantics and dropping the docs comment.
The fact that it continues to work is an accident.
Please add a docs comment

service/lib/agama/http/clients/software.rb
```diff
@@ -77,19 +76,18 @@ def get_resolvables(unique_id, type, optional)
           JSON.parse(get("software/resolvables/#{unique_id}?type=#{type}&optional=#{optional}"))
         end
 
-        def provisions_selected?(_provisions)
-          # TODO: implement it, not sure how it should look like
-          []
+        def provisions_selected?(provisions)
```
mvidner, line 81:
```suggestion
        # (Yes, with a question mark. Bad naming.)
        # @return [Array<String>] Those names that are selected for installation
        def provisions_selected?(provisions)
```

PR #2792, opened 2025-10-07
agama-network: add ipv4.link-local for ZeroConf/AutoIP

cfconrad:
Add the field IpConfig.link_local4, which represent NetworkManager `ipv4.link-local` field.

This is needed duo the migration of the following ifconfig
```
BOOTPROTO=dhcp+autoip
```
which lead to
```
ipv4.method=auto
ipv4.link-local=fallback (4)
```

## Testing

- *Tested manually*


rust/agama-network/src/model.rs
```diff
@@ -871,6 +871,35 @@ impl From<InvalidMacAddress> for zbus::fdo::Error {
...
+    #[default]
+    Default = 0,
+    Auto = 1,
+    Disabled = 2,
+    Enabled = 3,
+    Fallback = 4,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct InvalidLinkLocalValue(i32);
```
jcronenberg: Shouldn't this be something like this?

```suggestion
#[derive(Debug, Error)]
#[error("Invalid link-local value: {0}")]
pub struct InvalidLinkLocalValue(i32);
```

cfconrad: good hint, I will take it

rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Tue Oct  7 09:55:46 UTC 2025 - Clemens Famulla-Conrad <cfamullaconrad@suse.com>
+
+- Add IpConfig.link_local4 to specify ZeroConf/AutoIP behavior.
```
teclator, line 4: If there is no bug report then at least reference the PR.

rust/agama-network/src/nm/dbus.rs
```diff
@@ -1081,6 +1090,10 @@ fn ip_config_from_dbus(conn: &OwnedNestedHash) -> Result<IpConfig, NmError> {
             ip_config.dns_priority4 = Some(dns_priority4);
         }
 
+        if let Some(link_local4) = get_optional_property::<i32>(ipv4, "link-local")? {
+            ip_config.link_local4 = link_local4.try_into().unwrap_or(LinkLocal::Default);
```
jcronenberg: I think the `unwrap_or` here is not the right approach. First off `unwrap_or_default()` would probably be better, but in general the approach in `nm/dbus.rs` is that when a parsing error like this occurs it should propagate. So my suggestion would be to implement the corresponding `NmError` and then use `?` instead of `unwrap_or...`

cfconrad: I wasn't sure with this neither. But I thought it is better to map to "default" instead of error out, if NetworkManager upstream devs add an other value.

Ok, but  double checking also [mac_address6_from_dbus()](https://github.com/cfconrad/agama//blob/e80c62a96fca40c382e606f3d254ebd16f350b7d/rust/agama-network/src/nm/dbus.rs#L972) or 
[custom_mac_address_from_dbus()](https://github.com/cfconrad/agama//blob/e80c62a96fca40c382e606f3d254ebd16f350b7d/rust/agama-network/src/nm/dbus.rs#L999)
fail, if there is something invalid.

cfconrad: But here is something similar https://github.com/cfconrad/agama//blob/e80c62a96fca40c382e606f3d254ebd16f350b7d/rust/agama-network/src/nm/dbus.rs#L1306, and we are using `unwrap_or_default()`

jcronenberg: Yea actually thinking about it again I think maybe `unwrap_or_default()` is better since this property isn't really relevant for agama and mostly just for `wicked2nm`

cfconrad: changed to `unwrap_or_default()`.

@teclator WDYT?

rust/agama-network/src/nm/dbus.rs
```diff
@@ -434,12 +434,21 @@ fn ip_config_to_ipv4_dbus<'a>(
         .collect::<Vec<_>>()
         .into();
 
+    let link_local = if ip_config.link_local4 == LinkLocal::Fallback
+        && VersionReq::parse("<1.52.0").unwrap().matches(nm_version)
+    {
+        LinkLocal::Auto
```
cfconrad: Just realized, that `link-local=auto` with `ipv4.method=auto` will not use zeroconf in addition to DHCP on the interface. As `link-local=auto` actually mean, use the value from `ipv4.method`. 

Change this to `link-local=enabled`, then we have a AutoIP and the DHCPv4 address. So the difference between the NetworkManager versions is

`<1.52`: The interface will always have a zeroconf IP address, ALSO after DHCP
`>=1.52`: The zeroconf IP-Address will be removed, once the DHCP lease is applied

Sorry for the late noise!

PR #2793, opened 2025-10-08
feat: manager service

joseivanlopez:
## Problem

The *supervisor* type lives in the *agama-server* package which depends on the *agama_lib* package. The http events are provided by *agama_lib*, and it needs to know the event type of each service. This makes impossible to emit events from the supervisor service because of cyclic dependencies.

## Solution

Extract the *supervisor* to its own package, taking the opportunity for renaming it as *agama-manager*. Note that the supervisor concept is too related to the actors terminology.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Oct  9 10:24:03 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Extract supervisor to agama-manager package (gh#agama-project/agama#2793).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Oct  9 10:24:03 UTC 2025 - José Iván López González <jlopez@suse.com>
+
+- Extract supervisor to agama-manager package (gh#agama-project/agama#2793).
```
jreidinger, line 4: btw that is one part where I get the most conflicts as you do not use the latest master  and solving changelog conflicts is not fun :)

PR #2798, opened 2025-10-10
Drop the unused agama_utils::service module

imobachgs:
Drop the `agama_utils::service` module because it is not used anymore.


joseivanlopez (approved):
LGTM

PR #2801, opened 2025-10-10
Improve events organization

imobachgs:
## Problem

While moving some logic to separate packages, we found out some circular dependency problems with some types, especially with `http::Event` struct from *agama-lib* package.

## Solution

* Extract to *agama-utils* the types used for the HTTP API . In the future, the CLI will use these types without the need to depend on types from all Agama packages.
* The events of the new API are separated from the old events.
* The events are shared to all the packages, making possilbe to use a single broadcast channel to emit events.
* Remove the events listener which is not needed anymore.
* The manager emits a `StatusChanged` event.
* The progress adds information of the `ProgressChanged` event.
* The progress emits a `ProgressFinished` event everytime a progress is finished.

### DIFF
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -31,6 +31,7 @@ use agama_server::{
     logs::init_logging,
     web::{self, run_monitor},
 };
+use agama_utils::api::event::Receiver;
 use anyhow::Context;
 use axum::{
     extract::Request as AxumRequest,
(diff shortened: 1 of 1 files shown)


rust/agama-l10n/src/service.rs
```diff
@@ -18,21 +18,19 @@
...
+use crate::config::Config;
+use crate::message;
+use crate::model::ModelAdapter;
 use agama_locale_data::{InvalidKeymapId, InvalidLocaleId, InvalidTimezoneId, KeymapId, LocaleId};
-use agama_utils::{
-    actor::{self, Actor, Handler, MessageHandler},
-    issue::{self, Issue},
-};
+use agama_utils::actor::{self, Actor, Handler, MessageHandler};
+use agama_utils::api;
```
imobachgs, line 26: JFTR: we agreed to merge these "use" statements, but in a latter PR.

imobachgs (commented):
I have been working on top of this branch and I can confirm that it works. I cannot approve the PR however, because I opened it :-)

PR #2806, opened 2025-10-13
Adapt to new HTTP API

jreidinger:
Adapt the software service, which is still under development, to the new `api-v2` changes. The code will be merged as part of the `software-service` branch.

It contains beside newer api-v2 changes also software adaptation including:

- split of software code to agama-software
- defining actor for software
- separate thread for libzypp and channel to communicate togethere
- split of libzypp oriented model and other parts that works without libzypp like loading producs or licenses so it does not block each other

### DIFF
--- a/rust/agama-software/src/config.rs
+++ b/rust/agama-software/src/config.rs
@@ -34,10 +34,10 @@ use serde::{Deserialize, Serialize};
 pub struct Config {
     /// Product related configuration
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub product: Option<ProductSettings>,
+    pub product: Option<ProductConfig>,
     /// Software related configuration
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub software: Option<SoftwareSettings>,
+    pub software: Option<SoftwareConfig>,
 }
 
 /// Addon settings for registration
@@ -57,7 +57,7 @@ pub struct AddonSettings {
 /// Software settings for installation
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
-pub struct ProductSettings {
+pub struct ProductConfig {
     /// ID of the product to install (e.g., "ALP", "Tumbleweed", etc.)
     #[serde(skip_serializing_if = "Option::is_none")]
     pub id: Option<String>,
@@ -71,7 +71,7 @@ pub struct ProductSettings {
     pub addons: Option<Vec<AddonSettings>>,
 }
 
-impl ProductSettings {
+impl ProductConfig {
     pub fn is_empty(&self) -> bool {
         self.id.is_none()
             && self.registration_code.is_none()
@@ -84,10 +84,10 @@ impl ProductSettings {
 /// Software settings for installation
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
-pub struct SoftwareSettings {
+pub struct SoftwareConfig {
     /// List of user selected patterns to install.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub patterns: Option<PatternsSettings>,
+    pub patterns: Option<PatternsConfig>,
     /// List of user selected packages to install.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub packages: Option<Vec<String>>,
@@ -101,14 +101,14 @@ pub struct SoftwareSettings {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, utoipa::ToSchema)]
 #[serde(untagged)]
-pub enum PatternsSettings {
+pub enum PatternsConfig {
     PatternsList(Vec<String>),
     PatternsMap(PatternsMap),
 }
 
-impl Default for PatternsSettings {
+impl Default for PatternsConfig {
     fn default() -> Self {
-        PatternsSettings::PatternsMap(PatternsMap {
+        PatternsConfig::PatternsMap(PatternsMap {
             add: None,
             remove: None,
         })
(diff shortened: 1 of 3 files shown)


rust/agama-software/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+/// User configuration for the localization of the target system.
+///
+/// This configuration is provided by the user, so all the values are optional.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, utoipa::ToSchema)]
+#[schema(as = software::UserConfig)]
+#[serde(rename_all = "camelCase")]
+pub struct Config {
+    /// Product related configuration
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub product: Option<ProductSettings>,
```
imobachgs: It would be good to rename this to `ProductConfig`. It can be done later, but please, add it to your TODO list :wink:

rust/agama-software/src/extended_config.rs
```diff
@@ -0,0 +1,95 @@
...
+            patterns: PatternsSettings::default(),
+            packages: Default::default(),
+            extra_repositories: Default::default(),
+            only_required: false,
+        }
+    }
+}
+
+impl ExtendedConfig {
+    pub fn merge(&mut self, config: &Config) -> &Self {
```
imobachgs, line 84: It is not merging, it is overwriting the whole section. In other parts, we use [merge_struct](https://docs.rs/merge-struct/latest/merge_struct/) for that kind of stuff. I do not like it that much and I would rather prefer to check an alternative like https://crates.io/crates/merge. But we should use the same approach at least.

rust/agama-software/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+// To contact SUSE LLC about this file by physical or electronic mail, you may
+// find current contact information at www.suse.com.
+
+//! This crate implements the support for software handling in Agama.
+//! It takes care of setting the product, registration and software for the
+//!  target system.
+//!
+//! From a technical point of view, it includes:
+//!
+//! * The [UserConfig] struct that defines the settings the user can
```
imobachgs:
```suggestion
//! * The [Config] struct that defines the settings the user can
```

rust/agama-software/src/message.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl<T> SetSystem<T> {
+    pub fn new(config: T) -> Self {
+        Self { config }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, utoipa::ToSchema)]
+pub struct SystemConfig {
+    pub locale: Option<String>,
```
imobachgs: Copy & paste problem? These settings are from the `l10n` package.

rust/agama-software/src/model.rs
```diff
@@ -0,0 +1,335 @@
...
+pub mod packages;
+pub mod pattern;
+pub mod product;
+pub mod products;
+pub mod registration;
+pub mod software_selection;
+
+/// Abstract the software-related configuration from the underlying system.
+///
+/// It offers an API to query and set different software and product elements of a
```
imobachgs, line 47: The trait does not depend on libzypp, right?

rust/agama-software/src/model.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    /// Finalizes system like disabling local repositories
+    fn finish(&self) -> Result<(), service::Error>;
+}
+
+/// [ModelAdapter] implementation for libzypp systems.
+pub struct Model {
+    zypp_sender: mpsc::UnboundedSender<SoftwareAction>,
+    products: ProductsRegistry,
+    // FIXME: what about having a SoftwareServiceState to keep business logic state?
```
imobachgs, line 98: You mean out of this adapter, right? IMHO, it sounds like the right approach. The adapter should be only for interact with the backend (libzypp).

Additionally, the products registry could live outside of the adapter (or does it depend on libzypp?).

rust/agama-software/src/model.rs
```diff
@@ -0,0 +1,335 @@
...
+    async fn patterns(&self) -> Result<Vec<Pattern>, service::Error> {
+        let Some(product) = &self.selected_product else {
+            return Err(service::Error::MissingProduct);
+        };
+
+        let names = product
+            .software
+            .user_patterns
+            .iter()
+            .map(|user_pattern| match user_pattern {
```
imobachgs, line 125: NP: You can call `cloned()` here and drop the individual clones inside of the closure.

rust/agama-software/src/model.rs
```diff
@@ -0,0 +1,335 @@
...
+        let Some(product) = &self.selected_product else {
+            return Err(service::Error::MissingProduct);
+        };
+
+        let (tx, rx) = oneshot::channel();
+        let repositories = product
+            .software
+            .repositories()
+            .into_iter()
+            .map(|r| r.clone())
```
imobachgs:
```suggestion
            .cloned()
```

(edit: there were spurious backticks)

rust/agama-software/src/zypp_server.rs
```diff
@@ -0,0 +1,458 @@
+// Copyright (c) [2024] SUSE LLC
```
imobachgs:
```suggestion
// Copyright (c) [2025] SUSE LLC
```

rust/agama-software/src/config.rs
jreidinger: you mean `ProductSettings` to `ProductConfig`, right?

rust/agama-software/src/message.rs
jreidinger: yeap, some overlook

rust/agama-software/src/model.rs
jreidinger: yeap...but I am not sure if we need trait at all, as model has channel to libzypp related stuff...we can do the same thing for suse connect stuff and in the end we can just setup in tests testing channels instead of real ones.

rust/agama-software/src/model.rs
jreidinger: well, my idea is that model is what integrates multiple backends we have - libzypp, suseconnect, ProductsRegistry ( on system ).

imobachgs (approved):
The code will be merged as part of the software-service branch. By now, we can merge it there to get the latest updates from the `api-v2` branch.

PR #2814, opened 2025-10-17
Conversion of devicegraph to json

ancorgs:
Pull request created to discuss some aspects about the code organization.

1. I created the concept of `Interface` for stuff that used to be D-Bus interfaces (Drive, Block, Filesystem, etc. But I didn't group all the interfaces at a separate namespace because I felt that `Agama::Storage::DevicegraphConversions::ToJSONConversions::Interface::Block` was already too much.
2. For symmetry with the other conversions, I created a `DevicegraphConversions::ToJSON` class and a `DevicegraphConversions::ToJSONConversions::Devicegraph` one. But I find the second to be unnecessary. It's just one extra level but with the same responsibilities and API.
3. I decided to nest the partitions directly on the "partitionable" device instead of doing it inside of `:partition_table` (that is, inside the PartitionTable interface). That's a change compared to the D-BUS interface (they are nested within PartitionTable there). It kind of makes sense from the organization point of view and also from the code point of view (see below), but I'm not 100% sure.
4. I separated the concept of interfaces and nested devices. There is indeed a semantical difference. But honestly I only separated the concept to avoid some kind of "circular dependency" (it is not elegant if the interfaces create devices because then `ToJSONConversions::Device` would require the interfaces and some interface may require `ToJSONConversions::Device`).

Example output:

```ruby
[
  {
    :block => {
      :active => true,
      :encrypted => false,
      :shrinking =>
       {"Unsupported" => ["Resizing is not supported by this device."]},
      :size => 53687091200,
      :start => 0,
      :systems => [],
      :udev_ids => [],
      :udev_paths => []
    },
    :description => "Disk",
    :name => "/dev/vda",
    :partition_table =>
     {:type => "gpt", :unused_slots => [[62920704, 21471673856]]},
    :partitions =>
      [
        {
          :block => {
            :active => true,
            :encrypted => false,
            :shrinking =>
             {"Unsupported" =>
               ["Neither a file system nor a storage system was detected on the device. In case the device does contain a file system or a storage system that is not supported, resizing will most likely cause data loss."]},
            :size => 2097152,
            :start => 2048,
            :systems => [],
            :udev_ids => [],
            :udev_paths => []
          },
          :description => "BIOS Boot Partition",
          :name => "/dev/vda1",
          :sid => 44
        },
        {
          :block => {
            :active => true,
            :encrypted => false,
            :shrinking =>
             {"Unsupported" => ["Shrinking is not supported by this device"]},
            :size => 21474836480,
            :start => 6144,
            :systems => [],
            :udev_ids => [],
            :udev_paths => []
          },
          :description => "Btrfs Partition",
          :name => "/dev/vda2",
          :sid => 45
        },
        {
          :block => {
            :active => true,
            :encrypted => false,
            :shrinking => {"Supported" => 21474836480},
            :size => 10737418240,
            :start => 41949184,
            :systems => [],
            :udev_ids => [],
            :udev_paths => []
          },
          :description => "XFS Partition",
          :name => "/dev/vda3",
          :sid => 48
        }
      ],
    :sid => 42
  },
  {
    :block => {
      :active => true,
      :encrypted => false,
      :shrinking =>
       {"Unsupported" => ["Resizing is not supported by this device."]},
      :size => 53687091200,
      :start => 0,
      :systems => [],
      :udev_ids => [],
      :udev_paths => []
    },
    :description => "Disk",
    :name => "/dev/vdb",
    :sid => 50
  },
  {
    :block => {
      :active => true,
      :encrypted => false,
      :shrinking =>
       {"Unsupported" => ["Resizing is not supported by this device."]},
      :size => 53687091200,
      :start => 0,
      :systems => [],
      :udev_ids => [],
      :udev_paths => []
    },
    :description => "Ext4 Disk",
    :name => "/dev/vdc",
    :sid => 51
  }
]
```

### DIFF
--- a/service/lib/agama/storage/devicegraph_conversions/to_json.rb
+++ b/service/lib/agama/storage/devicegraph_conversions/to_json.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2025] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "agama/storage/devicegraph_conversions/to_json_conversions/devicegraph"
+
+module Agama
+  module Storage
+    module DevicegraphConversions
+      # Devicegraph conversion to JSON array according to schema.
+      class ToJSON
+        # @param config [Y2Storage::Devicegraph]
+        def initialize(devicegraph)
+          @devicegraph = devicegraph
+        end
+
+        # Performs the conversion to array according to the JSON schema.
+        #
+        # @return [Hash]
+        def convert
+          ToJSONConversions::Devicegraph.new(devicegraph).convert
+        end
+
+      private
+
+        # @return [Y2Storage::Devicegraph]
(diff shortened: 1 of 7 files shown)


ancorgs:
Closed in favor of #2815 (I messed up with the branches)

PR #2816, opened 2025-10-17
feat: storage service

joseivanlopez:
Add storage service to support the new HTTP API.

The D-Bus API is adapted to provide the required information, using D-Bus as a transport layer.

### DIFF
--- a/service/lib/agama/manager.rb
+++ b/service/lib/agama/manager.rb
@@ -231,7 +231,7 @@ module Agama
     #
     # @return [Boolean]
     def valid?
-      users.issues.empty? && !software.errors? && !storage.errors?
+      users.issues.empty? && !software.errors?
     end
 
     # Collects the logs and stores them into an archive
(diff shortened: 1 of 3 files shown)


service/lib/agama/storage/manager.rb
```diff
@@ -49,148 +46,87 @@ module Storage
...
-      #
-      # If the deprecated status changes, then callbacks are executed and the issues are
-      # recalculated, see {#on_deprecated_system_change}.
-      #
-      # @param value [Boolean]
-      def deprecated_system=(value)
-        return if deprecated_system? == value
+      def reset_activation
+        Y2Storage::Luks.reset_activation_infos
+        @activated = false
```
ancorgs: I'm not sure whether setting `activated` to false is correct. Resetting the LUKS activation info gives a second chance to introduce LUKS passwords, but it does not invalidate any previous activation.

teclator (approved):
LGTM

PR #2817, opened 2025-10-17
Remove Kalpa from multi-distribution image.

sfalken:
## Problem

Results in broken installations, and isn't intended to be distributed in this method.

- https://bugzilla.opensuse.org/show_bug.cgi?id=1248982

## Solution

Kalpa will provide it's own agama based installer image.


imobachgs (commented):
Hi! I would expect https://github.com/agama-project/agama/blob/master/products.d/kalpa.yaml to be removed too. Any reason to keep it on the repo?

sfalken:
> Hi! I would expect https://github.com/agama-project/agama/blob/master/products.d/kalpa.yaml to be removed too. Any reason to keep it on the repo?

I'm still trying to make an agama based installer for Kalpa, so it would still need to exist in the Repo, would it not?
I just don't want it showing up on the multi-distro images, as it results in unsupportable configurations, if it creates a bootable install at all, at present.

I'm not completely opposed to it existing on the multi-distribution image in the future, if I can figure out a way to make it work, but right now, I have users clicking on Kalpa in the current development images, and ending up with broken systems.

imobachgs (approved):

imobachgs:
> I'm still trying to make an agama based installer for Kalpa, so it would still need to exist in the Repo, would it not? I just don't want it showing up on the multi-distro images, as it results in unsupportable configurations, if it creates a bootable install at all, at present.

OK, it is fine with me. I think it makes things easier. So let's it here.

PR #2819, opened 2025-10-20
Initial self-update implementation

lslezak:
## Problem

- Missing self-update support

## Solution

- Drop the old self-update service
- Reimplement it as a systemd service started in initrd to update the system earlier.
- So we can boot the already updated root image without need for restarting the updated services (which ones?).

## Features

- This is just the initial implementation (rather a proof of concept)
- So far it can only use a custom self-update repository specified with the `inst.self_update=` boot option.
- The self update can be disabled by the `inst.self_update=0` boot option. However, as there is no fallback or default URL then currently the self-update is also skipped when no boot option is set.

## Details

- Both the dracut part and the systemd service using the `live-self-update` syslog identifier so using `journalctl -t live-self-update` command will show the output of both in a single place. For easier use it has been added to the predefined Bash history.
- I'm creating the PR against a separate branch so we can merge the feature when fully implemented. I expect it will take some time so do the reviews in smaller steps.

## Testing

- Tested manually in a locally built openSUSE Tumbleweed Live ISO
- Tested with `inst.self_update=https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/openSUSE_Tumbleweed/` boot option, optionally with `rd.break` to debug the status in the initrd.

## Additional fixes

- When setting the network in initrd then the system uses the `localhost.localdomain` host name, update the `agama-hostname` service so it sets the `agama` host name also in that case. (When a different hostname is set via the `hostname=` boot option then that name it kept.)

## Screenshots

Note: When using a freshly built Live ISO which contains the latest package then obviously the self-update cannot update anything. It would be nice to test it later again when some real packages can be updated.

### Using the Agama devel project for self update

- Stopped with `rd.break` to avoid scrolling of the messages

agama-self-update-step

### Displaying the service status (still in initrd)

agama-self-update-status-initrd

### Displaying the service status (after switching the root)

- Unfortunately the previous status is lost :worried: 

agama-self-update-status-root

### Displaying the log after switching root to the Live root image

agama-self-update-journal

### Manually restarting the service

- To allow manually retry a failed self-update.
- But he already present self-update repository needs to be manually removed and after the finishing the self-update the updated services need to be restarted manually.
- This is mostly for experts, not for regular use.

agama-self-update-restart

### DIFF
--- a/live/live-root/etc/systemd/system/live-self-update.service
+++ b/live/live-root/etc/systemd/system/live-self-update.service
@@ -6,7 +6,7 @@ DefaultDependencies=no
 # for the dependency order of the dracut services
 
 # after network is configured
-After=network.target
+After=network-online.target
 # after /sysroot is mounted
 After=initrd-fs.target
 # before running the pre-pivot dracut hooks (before the driver update)


live/live-root/usr/bin/live-self-update
```diff
@@ -0,0 +1,102 @@
...
+    if [ -n "$update_url" ]; then
+      repo="self-update-$index"
+      run_command zypper addrepo -f "$update_url" "$repo"
+      self_update_repos+=("$repo")
+      ((++index))
+    fi
+  done <"$REPO_FILE"
+
+  echo "<5>Loading repositories..."
+  run_command zypper refresh "${self_update_repos[@]}"
```
lslezak, line 76: We could possibly run the refresh with some timeout, that would fail the whole self-update in case of network issues:

```shell
timeout 5m zypper refresh ...
```

When there is a network problem it most likely happens in the initial repository refresh. If that succeeds then the package download and installation most likely succeeds as well. (That would run without timeout to ensure it always properly finishes.)

This could avoid getting stuck on network problems.

live/live-root/etc/systemd/system/live-self-update.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=Live system self update
+DefaultDependencies=no
+
+# see https://man7.org/linux/man-pages/man7/dracut.bootup.7.html
+# for the dependency order of the dracut services
+
+# after network is configured
+After=network.target
```
teclator: I guess it should be network-online.target in most cases although it could delay a little bit the boot process

live/live-root/usr/lib/dracut/modules.d/99live-self-update/live-self-update-parser.sh
```diff
@@ -0,0 +1,42 @@
...
+mkdir -p $(dirname "$REPO_FILE")
+
+get_url() {
+  # Agama driver updates
+  local url
+  url=$(getargs "inst.self_update=")
+
+  if [ "$url" == "0" ]; then
+    echo "<5>Self update disabled via boot parameter"
+  elif [ -z "$url" ]; then
```
teclator, line 23: so far it **LGTM** as it does not setup network by default an only if **inst.sel_update** is provided. Once we enabled it by default we should consider cases were the network is disabled explicitly (**rd.neednet=0**) as well as the default **NetworkManager** profile generated by **nm-initrd-generator** which is a multi-connect profile.

Last but not least, maybe we should handle also the proxy option passed as a kernel boot option although it can be part of the repository URL but just for avoiding having to declare it twice.

teclator (approved):
LGTM

PR #2822, opened 2025-10-22
Fix answering questions

imobachgs:
- Fix answering questions, which was not working.
- Flatten the body to PATCH questions.
- Add the missing Agama::HTTP::Base#patch method.

### DIFF
--- a/service/lib/agama/http/clients/questions.rb
+++ b/service/lib/agama/http/clients/questions.rb
@@ -27,6 +27,8 @@ module Agama
     module Clients
       # HTTP client to interact with the files API.
       class Questions < Base
+        class CouldNotAddQuestion < StandardError; end
+
         # Adds a question
         #
         # @param question [Agama::Question]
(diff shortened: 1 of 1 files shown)


service/lib/agama/http/clients/questions.rb
```diff
@@ -79,13 +79,13 @@ def wait_answer(id)
         # @return [Agama::Answer] The question answer, or the result of the block in case a block
         #   is given.
         def ask(question)
-          question_id = add(question)
-          answer = wait_answer(question_id)
+          added_question = add(question)
```
joseivanlopez, line 84: `#add` could return nil. Should we control it?

imobachgs: It is unlikely to fail, to be honest (it will be the lesser of your problems then). But it is true that we can log the error and... raise an exception?

joseivanlopez (approved):
LGTM

PR #2825, opened 2025-10-22
Fix api-v2 branch build

imobachgs:
## Problem

The `api-v2` branch does not build in OBS. There are two problems:

1. Agama crashes if the l10n monitor cannot connect to the systemd locale service, which is not available when running the tests in OBS.
1. The `agama-services` is not part of the Rust package anymore.

## Solution

The solution for the 2nd point is trivial. About the 1st one, we have decided that Agama should not crash, just report (in the logs) that it cannot observe system locale changes.

### DIFF
--- a/rust/package/agama.spec
+++ b/rust/package/agama.spec
@@ -213,7 +213,6 @@ echo $PATH
 %doc README.md
 %license LICENSE
 %{_bindir}/agama-web-server
-%{_datadir}/dbus-1/agama-services
 %{_pam_vendordir}/agama
 %{_unitdir}/agama-web-server.service
 


rust/agama-l10n/src/start.rs
```diff
@@ -53,8 +51,17 @@ pub async fn start(
     let model = Model::from_system()?;
     let service = Service::new(model, issues, events);
     let handler = actor::spawn(service);
-    let monitor = Monitor::new(handler.clone()).await?;
```
joseivanlopez, line 56: Maybe we could continue reporting the error and avoid failing in the test if the error is because the monitor. 

Probably we should continue raising the error when running agama.

imobachgs: I would rather prefer to have the same behavior in both scenarios. Not listening to locale changes is something we can afford even when running for real.

joseivanlopez: We agreed to continue with this approach. Note that any call to *localectl* would fail anyway in a real running.

PR #2831, opened 2025-10-27
Connect the new software service

imobachgs:
* Merge api-v2` and connects the new software service with the new API.
* Allow selecting a product and a set of patterns for installation.
* Performing the installation is out of scope.

The current code allows to configure the product repositories and perform a basic software selection.

## Workflow

We have introduced a new concept: `SoftwareState` (we need a better name, but it is similar to `NetworkState`). It represents the desired status (repositories, patterns, packages, etc.).

When the service receives a `SetConfig`, it generates a new `SoftwareState` using the following inputs:

* The product specification (which might include repositories).
* The current system (which might include the offline repository and, in the future, DUDs).
* The configuration itself.

With that information, Agama generates a new `SoftwareState`. The `ModelAdapter` synchronizes the state with the reality.

Captura desde 2025-10-28 14-12-26

Captura desde 2025-10-28 15-34-43

### DIFF
--- a/service/lib/agama/product_reader.rb
+++ b/service/lib/agama/product_reader.rb
@@ -58,7 +58,7 @@ module Agama
   private
 
     def default_path
-      Dir.exist?(GIT_DIR) || File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
+      File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
     end
   end
 end


rust/agama-cli/src/questions.rs
```diff
@@ -62,53 +62,48 @@ pub enum Modes {
     NonInteractive,
 }
 
-async fn set_mode(proxy: QuestionsProxy<'_>, value: Modes) -> anyhow::Result<()> {
-    proxy
-        .set_interactive(value == Modes::Interactive)
-        .await
-        .map_err(|e| e.into())
+async fn set_mode(client: HTTPClient, value: Modes) -> anyhow::Result<()> {
+    let policy = if value == Modes::Interactive {
```
jreidinger: here it would be better to have match to ensure that newly added modes are covered correctly...and even better would be to have enum Modes implement into Policy.

rust/agama-lib/src/http/event.rs
```diff
@@ -34,21 +34,19 @@ use crate::{
...
 use tokio::sync::broadcast;
 
-pub type Sender = broadcast::Sender<Event>;
-pub type Receiver = broadcast::Receiver<Event>;
+pub type OldSender = broadcast::Sender<OldEvent>;
+pub type OldReceiver = broadcast::Receiver<OldEvent>;
 
 /// Agama event.
 ///
 /// It represents an event that occurs in Agama.
```
jreidinger, line 46: I kind of expect here explanation why it is "Old" event..if it is deprecated or why it is called this way.

rust/agama-lib/src/questions/http_client.rs
```diff
@@ -89,30 +89,56 @@ impl HTTPClient {
...
+        let patch = Patch {
+            update: Some(config),
+        };
+        self.client.patch_void("/v2/config", &patch).await?;
+        Ok(())
+    }
+
+    pub async fn delete_question(&self, id: u32) -> Result<(), QuestionsHTTPClientError> {
+        let update = UpdateQuestion::Delete { id };
+        self.client.patch_void("/v2/questions", &update).await?;
```
jreidinger, line 132: I hope that v2 is just temporary as I do not see point of having killed v1 and only having v2.

imobachgs: Yes, it is a temporary thingie.

rust/agama-lib/src/http/event.rs
imobachgs: We are killing it, no need for documentation.

rust/agama-cli/src/questions.rs
imobachgs: I agree, I will adapt the code.

rust/agama-manager/src/service.rs
```diff
@@ -134,14 +133,14 @@ impl MessageHandler<message::GetExtendedConfig> for Service {
...
-            localization: Some(l10n_config),
-            ..Default::default()
+    async fn handle(&mut self, _message: message::GetExtendedConfig) -> Result<Config, Error> {
+        let l10n = self.l10n.call(l10n::message::GetConfig).await?;
+        let software = self.software.call(software::message::GetConfig).await?;
+        let questions = self.questions.call(question::message::GetConfig).await?;
+        Ok(Config {
+            l10n: Some(l10n),
+            questions: Some(questions),
+            software: Some(software),
```
jreidinger, line 143: does it make sense to always return `Some` ? I kind of think that it should return really option based if there is anything in config or not

rust/agama-software/src/model/packages.rs
```diff
@@ -74,46 +75,20 @@ pub struct ResolvableParams {
 #[derive(Clone, Debug, Deserialize, Serialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Repository {
-    /// repository identifier
-    pub id: i32,
-    /// repository alias. Has to be unique
+    /// Repository identifier
+    pub id: Option<i32>,
```
jreidinger: this is something we should drop soon as unique identifier should be alias and this is just DBus relict.

rust/agama-software/src/model/packages.rs
```diff
@@ -74,46 +75,20 @@ pub struct ResolvableParams {
...
-    /// repository name
+    /// Repository name
     pub name: String,
-    /// Repository url (raw format without expanded variables)
+    /// Repository URL (raw format without expanded variables)
     pub url: String,
-    /// product directory (currently not used, valid only for multiproduct DVDs)
-    pub product_dir: String,
+    /// Product directory (currently not used, valid only for multiproduct DVDs)
+    pub product_dir: String, // FIXME: needed?
```
jreidinger: well, it depends how future full medium will look like if they want somehow integrate addons on full medium. Then it can be useful if they do it per directory.

rust/agama-software/src/model/packages.rs
```diff
@@ -74,46 +75,20 @@ pub struct ResolvableParams {
...
-    /// Repository priority, lower number means higher priority, the default priority is 99
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub priority: Option<i32>,
-    /// Whenever repository can be unsigned. Default is false
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub allow_unsigned: Option<bool>,
-    /// List of fingerprints for GPG keys used for repository signing. By default empty
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub gpg_fingerprints: Option<Vec<String>>,
+    pub loaded: bool, // FIXME: needed?
```
jreidinger: probably not needed, another DBus relict.

rust/agama-software/src/model/state.rs
```diff
@@ -0,0 +1,478 @@
...
+            alias: value.alias.clone(),
+            url: value.url.clone(),
+            enabled: value.enabled,
+        }
+    }
+}
+
+/// Defines a resolvable to be selected.
+#[derive(Debug, PartialEq)]
+pub struct Resolvable {
```
jreidinger: why not have a type here? It is kind of strange to have resolvable without type of it. It cannot be used standalone without context.

mvidner: (_type_ aka _kind_: package, pattern, patch...)

imobachgs: Because it does not bring any value. They are included in different lists, which already tells you what to expect.

I wanted to keep it simple by now. I guess that it will change when we implement the software selection API for other services.

rust/agama-software/src/model/packages.rs
imobachgs: I agree. My point is that we can add it later if needed.

rust/agama-software/src/service.rs
```diff
@@ -178,29 +190,33 @@ impl MessageHandler<message::GetConfig> for Service {
...
         let model = self.model.clone();
+        let issues = self.issues.clone();
         tokio::task::spawn(async move {
             let mut my_model = model.lock().await;
-            // FIXME: convert unwraps to sending issues
-            if need_probe {
-                my_model.probe(&new_product_spec.unwrap()).await.unwrap();
+            let found_issues = my_model.write(software).await.unwrap();
+            if !found_issues.is_empty() {
+                _ = issues.cast(issue::message::Update::new(Scope::Software, found_issues));
```
jreidinger, line 218: this code looks quite tricky, what it actually do and why?

rust/agama-utils/src/api/action.rs
```diff
@@ -0,0 +1,30 @@
...
+// To contact SUSE LLC about this file by physical or electronic mail, you may
+// find current contact information at www.suse.com.
+
+use crate::api::l10n;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, utoipa::ToSchema)]
+pub enum Action {
+    #[serde(rename = "configureL10n")]
+    ConfigureL10n(l10n::SystemConfig),
```
jreidinger, line 27: I still thing we will need explicit probe for software here e.g. when repository refresh failed.

rust/agama-manager/src/service.rs
imobachgs: All the modules have a Config, although it is the default one, but we are using the regular `Config` struct here. I think we need to have another conversation about that ExtendedConfig but out of this PR.

service/lib/agama/config_reader.rb
```diff
@@ -137,7 +137,7 @@ def remote_config
     end
 
     def default_path
-      Dir.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
+      Dir.exist?(GIT_DIR) || File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
```
jreidinger: hmm, this change is a bit wrong as File.exist? check for stat, so Dir.exist? is already included. So it should be just
```suggestion
      File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
```

rust/agama-utils/src/api/action.rs
imobachgs: Sure thing, even for probe. But let's add it when we implement it.

rust/agama-software/src/service.rs
imobachgs: It register issues in the issues service.

service/lib/agama/config_reader.rb
jreidinger: try yourself in irb:
```
jreidinger@localhost:~> irb
irb(main):001:0> File.exist?("/dev/null")
=> true
irb(main):002:0> File.exist?("/dev")
=> true

```

service/lib/agama/product_reader.rb
```diff
@@ -58,7 +58,7 @@ def load_products
   private
 
     def default_path
-      Dir.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
+      Dir.exist?(GIT_DIR) || File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
```
jreidinger: same here:
```suggestion
      File.exist?(GIT_DIR) ? GIT_PATH : SYSTEM_PATH
```

rust/agama-software/src/service.rs
jreidinger: well, nicer method name probably helps here like `issues.add` or `issues.concat`. `cast` invokes to me more something like change type of issues.

imobachgs: It is because it is an actor. `cast` is to send a message and do not wait for the result (we do not care in this case).

PR #2833, opened 2025-10-27
Improved the self-update implementation

lslezak:
##  Description

This is a next step in the self-update implementation for the QU0 release:

- Contact the registration server (SCC/RMT) to get the final self-update URL
- Configure fallback URL if SCC/RMT is not reachable or returns empty response
- Do not configure the network if `rd.neednet=0` boot option is used (the self update then most likely fails, but if that's the user decision let's honor that.)
- In openSUSE self-update is skipped as there is no self-update repository configured, but it is still possible to use a custom self-update repository explicitly (might be useful for testing or using a non-official self-update repo)
- Updated documentation (just removed some obsolete parts)
- Verbose logging (with `set -x`), will be useful for debugging. Let's be more verbose in the initial implementation where we might expect more bug reports. Later we can remove that or enable only on explicit request. Since we have recently enabled the plymouth splash screen the verbose details will not be visible to users by default.

## Details

The self-update is configured via several files in the `/etc/live-self-update` directory:

- `default_reg_server_url` - contains the default registration server URL (`https://scc.suse.com`), the file is missing in openSUSE indicating that SCC/RMT should not be used
- `fallback_url` - contains the fallback URL when contacting SCC/RMT fails or empty response is returned. It can contain several placeholder so it works for multiple architectures or multiple release versions.

Maybe we could use a single JSON file for this and read the values using `jq`. On the other hand reading two plain text files is trivial in bash... 

The temporary files are created in `/run/live-self-update` directory. All files are kept, they might be useful for debugging later.

The file locations are shared in the `/usr/lib/live-self-update/conf.sh` "config" file to avoid typos and ensure that only one place needs to updated in case of changes.

## Testing

- Tested manually in both SLE and Tumbleweed builds

## Screenshots

Self-update log (`journalctl -t live-self-update`) in SLES (the milestone messages originally prefixed with `<5>` string are highlighted). The self-update actually updated nothing as everyting was up to date. This is the default behavior no extra boot parameter was used (except `rd.break` to get the log while still running in initramfs).

agama-self-update-log

---

I tested it also in openSUSE Tumbleweed. By default the self-update is skipped, even the network configuration in initramfs is skipped. But I tried using the custom self update repo from "https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Devel/openSUSE_Tumbleweed/" and this time one package was successfully updated. This is `journalctl -o cat -t live-self-update` output:

agama-self-update-tw-update-log

## TODO

What is still missing (I'll create cards for that):

- [ ] Allow ignoring a self-signed certificate for RMT (basically disabling SSL checks). Add `inst.self_update_insecure` option (similar to `inst.dud_insecure` and `inst.dud_insecure=1` we already have) which adds the `--insecure` parameter to the `curl` call and append `?ssl_verify=no` to repository URL when passing to zypper.  (See `man curl` and `man zypper`.)
- [ ] Allow autoimporting any GPG keys when missing. Reuse `inst.self_update_insecure` or have a separate option for this? If set add the `--gpgcheck-allow-unsigned` option to the `zypper addrepo` call and add `--gpg-auto-import-keys` to the `zypper refresh` call.
- [ ] Proxy configuration. Do we need to do something or it would automagically work? Somebody should test this. If it does not work we need to pass the `--proxy` option to the `curl` call and append `?proxy=<proxy>` to repository URL when passing to zypper. (See `man curl` and `man zypper`.) Or maybe just configure the `$PROXY` environment variable like we do for the Agama service ([here](https://github.com/agama-project/agama/blob/895dde794873409db7802507ef35b054ee317451/live/live-root/usr/lib/dracut/modules.d/99agama-cmdline/agama-cmdline-conf.sh#L8) and [here](https://github.com/agama-project/agama/blob/895dde794873409db7802507ef35b054ee317451/rust/share/agama-web-server.service#L9)), I'm not a network expert... :thinking: 
- [ ] Create documentation for the self-update process

## Questions

Should we actually use the `inst.self_update` parameter name? It is not about the installation, but about the live system itself. Should we rather use `live.*` prefix, like we already do  for the `live.password`? It's a bit complicated that for DUD we already use `inst.dud` and it should be unified... :thinking:

### DIFF
--- a/live/live-root/usr/bin/live-self-update
+++ b/live/live-root/usr/bin/live-self-update
@@ -132,9 +132,10 @@ add_update_repos() {
   done < "$1"
 }
 
-# read the configured fallback repository, the URL might contain several placeholders
-# so it works is multiple architectures and does not contain hardcoded version
-# (the fallback URL should work with SLES-16.0, SLES-16.1, SLES-16.2...)
+# Reads the configured fallback repository, the URL might contain several
+# placeholders. This way, it works across multiple architectures and does not
+# contain a hardcoded version (the fallback URL should work with SLES-16.0,
+# SLES-16.1, SLES-16.2...)
 read_fallback_repos() {
   # Substituting $arch pattern with the architecture of the current system.
   # Substituting these variables with the /etc/os-release content:


live/live-root/usr/bin/live-self-update
```diff
@@ -81,14 +190,13 @@ install_updates() {
...
   done
 
-  echo "<5>Updating packages..."
-  # update the system only using the predefined "agama-devel" repository,
-  # in real implementation we would need to add the self-update repository and 
-  # we should use the "--non-interactive" option to run without asking the user
+  milestone "Updating packages..."
+  # update the system only using the predefined self-update repositories,
+  # allow downgrading and changing vendor (can be useful for testing)
   run_command zypper --non-interactive update --allow-downgrade --allow-vendor-change --details --no-recommends "${repo_options[@]}"
```
kobliha, line 197: Maybe also this?
--auto-agree-with-licenses

lslezak: I'm not sure if adding every possible zypper option "just in case" is a good idea. The self-update should only update the installed packages. Adding a new dependency to the installer is possible but I doubt we would ever add a package which requires explicit license confirmation (a non-free package).

If somebody else needs to add a licensed package (but why?) it is possible to do so via a DUD.

live/live-root/usr/bin/live-self-update
```diff
@@ -23,56 +48,140 @@ run_command() {
...
+  if [ -n "$server" ]; then
+    read_os_release
+
+    local query
+    query="$server/connect/repositories/installer?identifier=$NAME&version=$VERSION_ID&arch=$ARCH"
+    info "Running query $query...."
 
-  echo "<5>Adding self update repositories..."
-  # add the self-update repositories  
+    curl --silent --globoff --location --retry 3 --connect-timeout 20 --max-time 10 --fail --show-error --output "$RESPONSE_FILE" -- "$query"
```
teclator, line 110: did you considered use the **fetch_url** function from the **dracut url-lib** module?

https://github.com/dracutdevs/dracut/blob/master/modules.d/45url-lib/url-lib.sh#L59

live/live-root/usr/bin/live-self-update
```diff
@@ -23,56 +48,140 @@ run_command() {
...
+      run_command zypper removerepo "$repo" || true
       run_command zypper addrepo -f "$update_url" "$repo"
       self_update_repos+=("$repo")
       ((++index))
     fi
-  done <"$REPO_FILE"
+  done < "$1"
+}
+
+# read the configured fallback repository, the URL might contain several placeholders
```
teclator: I guess you meant in multiple architectures.. what about:

> Reads the configured fallback repository, the URL might contain several placeholders. This way, it works across multiple architectures and does not contain a hardcoded version

live/live-root/usr/bin/live-self-update
```diff
@@ -23,56 +48,140 @@ run_command() {
...
+}
 
-  index=0
-  self_update_repos=()
+# ask the registration server (SCC or RMT) for the self-update URL
+query_reg_server() {
+  local server
+  # custom server (RMT)
+  if [ -f "$SERVER_FILE" ]; then
+    server=$(cat "$SERVER_FILE")
```
imobachgs, line 97: np: working on the DUD feature I learned that you do not need to rely on `cat`. Something like this would do the trick:

```
server=$(<"$SERVER_FILE")
```
But your current solution is more portable.

live/live-root/usr/lib/dracut/modules.d/99live-self-update/live-self-update-parser.sh
```diff
@@ -7,34 +7,78 @@
...
+# source the shared self-update configuration variables
+. ./lib/live-self-update/conf.sh
 
-mkdir -p $(dirname "$REPO_FILE")
+mkdir -p "$RUN_DIR"
 
+configure_network() {
+  # skip the network configuration if "rd.neednet=0" option was used by user
+  if getargbool 1 rd.neednet; then
+    echo "<5>Enabling network configuration"
```
imobachgs, line 21: I would expect to use the `milestone` function here.

live/live-root/usr/lib/dracut/modules.d/99live-self-update/live-self-update-parser.sh
```diff
@@ -7,34 +7,78 @@
...
+    echo "rd.neednet=1" > "$FILE"
+    # use DHCP if there is no network configuration provided by user
+    if ! getarg "ip="; then
+      echo "Using default DHCP network configuration"
+      echo "ip=dhcp" >> "$FILE"
+    fi
+  else
+    echo "<5>Network configuration disabled by the rd.neednet=0 boot parameter"
```
imobachgs, line 29: Same than above. `milestone`?

live/live-root/usr/bin/live-self-update
lslezak: Yes, I already tried using `dracut-lib.sh`. But the problem is that it works only when running from dracut. It does not work after switching to the root image. And I'd like to make it work also in the root image as that makes testing and debugging easier. And more importantly it allows users manually restart the self-update if it for whatever reason failed in initrd.

live/live-root/usr/bin/live-self-update
lslezak: Ah, nice, I'm not an expert in shell programming. I already found some Bash tricks in https://dracut-ng.github.io/dracut-ng/developer/bash.html documentation.

live/live-root/usr/lib/dracut/modules.d/99live-self-update/live-self-update-parser.sh
lslezak: Yeah, I wanted use it here as well but then we would need to somehow share the functions from another script. Then I decided it's not worth of doing that just because of a single trivial function which in the end only hides the "<5>" prefix.

live/live-root/usr/bin/live-self-update
teclator: I see, so, it also explain recent changes not relying on dracut builtin functions like getarg

PR #2837, opened 2025-10-30
Software service clean-up

imobachgs:
## Code clean-up

Let's start dropping unused code from the previous software layer.

* `agama_lib::software`
* `agama_lib::product`
* `agama_server::software`
* Old API documentation.
* Some code of the new layer which is not used yet and, most probably, would be implemented in a slightly different way.

## Software types to `agama_utils::api`

Just like we are doing in other modules, the types that will be exposed through the HTTP API are moved to `agama_utils::api`. This PR moves the code but when working on the new `system` and `proposal` we might need to clean and document those types.

### DIFF
--- a/rust/agama-software/src/model.rs
+++ b/rust/agama-software/src/model.rs
@@ -63,8 +63,8 @@ pub trait ModelAdapter: Send + Sync + 'static {
         optional: bool,
     ) -> Result<(), service::Error>;
 
-    /// Probes system and updates info about it.
-    async fn probe(&mut self) -> Result<(), service::Error>;
+    /// Refresh repositories information.
+    async fn refresh(&mut self) -> Result<(), service::Error>;
 
     /// install rpms to target system
     async fn install(&self) -> Result<bool, service::Error>;
@@ -136,7 +136,7 @@ impl ModelAdapter for Model {
             .unwrap_or_default()
     }
 
-    async fn probe(&mut self) -> Result<(), service::Error> {
+    async fn refresh(&mut self) -> Result<(), service::Error> {
         unimplemented!()
     }
 
--- a/rust/agama-software/src/service.rs
+++ b/rust/agama-software/src/service.rs
@@ -199,9 +199,9 @@ impl MessageHandler<message::GetProposal> for Service {
 }
 
 #[async_trait]
-impl MessageHandler<message::Probe> for Service {
-    async fn handle(&mut self, _message: message::Probe) -> Result<(), Error> {
-        self.model.lock().await.probe().await?;
+impl MessageHandler<message::Refresh> for Service {
+    async fn handle(&mut self, _message: message::Refresh) -> Result<(), Error> {
+        self.model.lock().await.refresh().await?;
         self.update_system().await?;
         Ok(())
     }
--- a/rust/agama-software/src/message.rs
+++ b/rust/agama-software/src/message.rs
@@ -77,9 +77,9 @@ impl Message for Install {
     type Reply = bool;
 }
 
-pub struct Probe;
+pub struct Refresh;
 
-impl Message for Probe {
+impl Message for Refresh {
     type Reply = ();
 }
 


rust/agama-lib/src/store.rs
```diff
@@ -184,33 +170,18 @@ impl Store {
...
-        let is_product_selected = self.detect_selected_product().await?;
-        if let Some(software) = &settings.software {
-            Store::ensure_selected_product(is_product_selected)?;
-            self.software.store(software).await?;
-        }
         let mut dirty_flag_set = false;
         // iscsi has to be done before storage
         if let Some(iscsi) = &settings.iscsi {
-            Store::ensure_selected_product(is_product_selected)?;
-            dirty_flag_set = true;
```
jreidinger, line 202: this is inconsistent, below you keep setting dirty flag.

rust/agama-lib/src/store.rs
```diff
@@ -184,33 +170,18 @@ impl Store {
...
             self.users.store(user).await?;
         }
-        // order is important here as network can be critical for connection
-        // to registration server and selecting product is important for rest
-        if let Some(product) = &settings.product {
-            self.product.store(product).await?;
-        }
-        // here detect if product is properly selected, so later it can be checked
-        let is_product_selected = self.detect_selected_product().await?;
-        if let Some(software) = &settings.software {
```
jreidinger, line 194: how ordering will be done now? I expect that afterwards order of loading settings will be managed by manager, but how we achieve order now? Or we keep unattended broken meanwhile?

rust/agama-software/src/lib.rs
```diff
@@ -41,18 +41,14 @@ pub use start::start;
 pub mod service;
 pub use service::Service;
 
-mod model;
+pub mod model;
```
jreidinger: why model needs to be accessible outside of crate?

rust/agama-software/src/model.rs
```diff
@@ -86,7 +64,7 @@ pub trait ModelAdapter: Send + Sync + 'static {
     ) -> Result<(), service::Error>;
 
     /// Probes system and updates info about it.
-    async fn probe(&mut self, product: &ProductSpec) -> Result<(), service::Error>;
+    async fn probe(&mut self) -> Result<(), service::Error>;
```
jreidinger: why you remove parameter here? Will model hold the product specification?

rust/agama-software/src/service.rs
```diff
@@ -233,15 +201,7 @@ impl MessageHandler<message::GetProposal> for Service {
...
-        let Some(product_id) = self.state.config.product.clone().and_then(|c| c.id) else {
-            return Err(Error::MissingProduct);
-        };
-
-        let Some(product) = self.products.find(&product_id) else {
-            return Err(Error::WrongProduct(product_id));
-        };
-
-        self.model.lock().await.probe(product).await?;
+        self.model.lock().await.probe().await?;
```
jreidinger: if probe will be slow we will need to add it to own tokio task.

rust/agama-lib/src/store.rs
imobachgs: Because it is used to determine whether to reprobe storage. Anyway, this code will be deleted completely :-)

rust/agama-lib/src/store.rs
imobachgs: Unattended is already broken. We are just removing the code piece by piece. All this logic will be in the new manager, behind the API.

rust/agama-software/src/model.rs
imobachgs: The probe action has not been implemented yet, but the idea is that you the product is set only in `SetConfig`. The probe action is only to refresh the respositories. And if we need the product spec for some reason, then yes, I guess we will need to keep it.

rust/agama-software/src/service.rs
imobachgs: Yes, most probably. The probe action is not used yet, we will address this issue soon.

rust/agama-software/src/lib.rs
imobachgs: You are right, it is not. I will fix it.

rust/agama-software/src/model.rs
jreidinger: well, it will mean that for setting product you need
1. set product
2. add product repos
3. refresh repos
4. add product itself and all its mandatory patterns and packages.

if this workflow is ok for you, then it can be done without product spec....but in that case it will need at least update comment, maybe even better to update method name to something like refresh_repositories

imobachgs: Then let's rename it.

PR #2840, opened 2025-10-31
Adapt network to the new config based HTTP API.

teclator:
## Problem

There is a new config based HTTP API and all the services needs to be adapted. The aim of this PR is to adapt network.

- https://github.com/agama-project/agama/pull/2715

## Solution

- Implement the support for using the new HTTP API using the already implemented Network service instead of the new services.
- Adapted the web to consume the new storage HTTP API.


rust/agama-lib/share/profile.schema.json
```diff
@@ -336,6 +336,29 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "generalState": {
+          "title": "Network general state settings",
```
```suggestion
          "title": "Network general settings",
```

rust/agama-lib/share/profile.schema.json
```diff
@@ -336,6 +336,29 @@
...
+          "type": "object",
+          "properties": {
+            "connectivity": {
+              "title": "Determines whether the user is able to access the Internet",
```
```suggestion
              "title": "Whether the user is able to access the Internet",
```

Just for consistency.

rust/agama-manager/src/start.rs
```diff
@@ -35,6 +35,8 @@ pub enum Error {
     L10n(#[from] l10n::start::Error),
     #[error(transparent)]
     Storage(#[from] storage::start::Error),
+    #[error(transparent)]
+    NetworkSystem(#[from] network::NetworkSystemError),
```
imobachgs:
```suggestion
    Network(#[from] network::NetworkSystemError),
```

rust/agama-manager/src/start.rs
```diff
@@ -51,8 +53,12 @@ pub async fn start(
     let progress = progress::start(events.clone()).await?;
     let l10n = l10n::start(issues.clone(), events.clone()).await?;
     let storage = storage::start(progress.clone(), issues.clone(), events.clone(), dbus).await?;
+    let network_adapter = network::NetworkManagerAdapter::from_system()
+        .await
+        .expect("Could not connect to NetworkManager");
+    let network = network::NetworkSystem::new(network_adapter).start().await?;
```
imobachgs: As you mentioned, perhaps we could have a `NetworkSystem::for_network_manager` so you do no need to worry about the adapter here, because there is quite some network logic here.

rust/agama-network/src/action.rs
```diff
@@ -42,6 +43,15 @@ pub enum Action {
     GetConnection(String, Responder<Option<Connection>>),
     /// Gets a connection by its Uuid
     GetConnectionByUuid(Uuid, Responder<Option<Connection>>),
+    /// Gets the internal state of the network configuration
+    GetConfig(Responder<Config>),
+    /// Gets the internal state of the network configuration proposal
+    GetProposal(Responder<Proposal>),
+    /// Updates th internal state of the network configuration
```
imobachgs:
```suggestion
    /// Updates the internal state of the network configuration
```

rust/agama-network/src/action.rs
```diff
@@ -42,6 +43,15 @@ pub enum Action {
...
+    UpdateConfig(Box<Config>, Responder<Result<(), NetworkStateError>>),
+    /// Gets the current network configuration containing connections, devices, access_points and
+    /// also the general state
+    GetSystemConfig(Responder<SystemInfo>),
```
imobachgs: It is not the config, but the system itself. The name could be misleading.

rust/agama-manager/src/start.rs
imobachgs: Additionally, why are you using `NetworkSystemError`? Just mentioning because you already have a `agama_network::error:Error`.

rust/agama-network/src/model.rs
```diff
@@ -164,6 +158,52 @@ impl NetworkState {
         Ok(())
     }
 
+    pub fn update_state(&mut self, config: Config) -> Result<(), NetworkStateError> {
```
imobachgs, line 169: Please, document.

rust/agama-network/src/model.rs
```diff
@@ -164,6 +158,52 @@ impl NetworkState {
...
+        if let Some(connections) = config.connections {
+            let mut collection: ConnectionCollection = connections.clone().try_into()?;
+            for conn in collection.0.iter_mut() {
```
imobachgs: That `.0` looks like an internal detail. I would the method or implement the [IterMut](https://doc.rust-lang.org/std/slice/struct.IterMut.html), as you prefer.

rust/agama-network/src/model.rs
```diff
@@ -252,6 +292,20 @@ impl NetworkState {
             )),
         }
     }
+
+    pub fn ports_for(&self, uuid: Uuid) -> Vec<String> {
```
imobachgs: Do you need this function to be public?

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+impl ConnectionCollection {
+    pub fn ports_for(&self, uuid: Uuid) -> Vec<String> {
+        self.0
+            .iter()
+            .filter(|c| c.controller == Some(uuid))
+            .map(|c| {
+                if let Some(interface) = c.interface.to_owned() {
+                    interface
+                } else {
+                    c.clone().id
```
imobachgs:
```suggestion
                    c.id.clone()
```

imobachgs: Perhaps this is more readable:

```rust
 match &c.interface {
     Some(interface) => interface.clone(),
     None => c.id.clone(),
}
```

Or you can use this, which is more concise:

```rust
c.interface.as_ref().unwrap_or(&c.id).clone()
```

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+impl TryFrom<ConnectionCollection> for NetworkConnectionsCollection {
+    type Error = NetworkStateError;
+
+    fn try_from(collection: ConnectionCollection) -> Result<Self, Self::Error> {
+        let network_connections = collection
+            .0
+            .iter()
+            .filter(|c| c.controller.is_none())
+            .map(|c| {
+                let mut conn = NetworkConnection::try_from(c.clone()).unwrap();
```
imobachgs, line 1350: Can this call to `unwrap()` fail?

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+                }
+            })
+            .collect()
+    }
+}
+
+impl TryFrom<ConnectionCollection> for NetworkConnectionsCollection {
+    type Error = NetworkStateError;
+
+    fn try_from(collection: ConnectionCollection) -> Result<Self, Self::Error> {
```
imobachgs, line 1345: Can this function really fail?

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+        let mut controller_ports: HashMap<String, Uuid> = HashMap::new();
+
+        for net_conn in &collection.0 {
+            let mut conn = Connection::try_from(net_conn.clone())?;
+            conn.uuid = Uuid::new_v4();
+            let mut ports = vec![];
+            if let Some(bridge) = &net_conn.bridge {
+                ports = bridge.ports.clone();
+            }
+            if let Some(bond) = &net_conn.bond {
```
imobachgs, line 1379: I see logic like this one in several places. I wonder if we could have some higher level API for that).

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+            }
+            for port in &ports {
+                controller_ports.insert(port.to_string(), conn.uuid);
+            }
+
+            conns.push(conn);
+        }
+
+        for (port, uuid) in controller_ports {
+            let default = Connection::new(port.clone(), DeviceType::Ethernet);
```
imobachgs: Do not get me wrong, this is fine, but you are creating a struct that you might not use at the end. What about something like:

```rust
let mut conn = conns
    .iter()
    .find(|c| c.id == port || c.interface.as_ref() == Some(&port))
    .cloned()
    .unwrap_or_else(|| Connection::new(port, DeviceType::Ethernet));
```

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+                .to_owned();
+            conn.controller = Some(uuid);
+            conns.push(conn);
+        }
+
+        Ok(ConnectionCollection(conns))
+    }
+}
+
+impl TryFrom<NetworkState> for NetworkConnectionsCollection {
```
imobachgs: I have the impression that we are abusing `TryFrom`. This or the implementation of `TryFrom<NetworkState> for SystemInfo` are some examples.

rust/agama-network/src/model.rs
```diff
@@ -1742,6 +1325,179 @@ pub struct BondConfig {
...
+    type Error = NetworkStateError;
+
+    fn try_from(state: NetworkState) -> Result<Self, Self::Error> {
```
imobachgs: The same comments I did for `TryFrom<ConnectionCollection> for NetworkConnectionsCollection` still apply.

rust/agama-manager/src/start.rs
teclator: The idea was to move it to **agama_network::error::Error** instead of **NetworkSystemError** or **NetworkAdapterError** but probably defined it but didn't moved it yet.

rust/agama-manager/src/start.rs
teclator: Exactly, would be nice to have just a call initializing the service using a default adapter as the NetworkManager one.

rust/agama-network/src/model.rs
teclator: not really

rust/agama-network/src/model.rs
teclator: Probably not, but then, we probably should implement the **From** trait instead of the **TryFrom** for **NetworkConnection** and the same for the collections ... what do you think?

rust/agama-network/src/model.rs
teclator: The same mentioned above, it should be a From instead of **TryFrom** but inherited from what we had for the **Connection**.

PR #2845, opened 2025-11-03
Implement proxy support in the self-update, fixed handling fallback

lslezak:
## Problem

- Missing HTTP proxy support in self-update
- Fallback for failing SCC/RMT was not handled correctly (I forgot about the global `set -e` option which fails the whole script on any error :worried: )

## Solution

- Get the configured proxy
- Pass it to the `curl` call which contacts the SCC and zypper  which downloads the packages via `http_proxy` and `https_proxy` environment variables.
- Fixed checking the result of the `curl` call
- Fixed keeping the systemd service status (workaround: enable the service also in the root system, but due to `WantedBy=initrd.target` it runs only once in initramfs.

## Testing

- Tested manually with `proxy=` boot option and also when setting the proxy interactively with the dracut menu (invoked with the `rd.cmdline=menu` boot option)
- Tested with my locally running Squid proxy server
- In both cases the proxy was use for contacting both the registration server (SCC) and the self-update repository.
- So far tested with plain proxy without authentication, later I'll test it with authenticated proxy

I checked the Squid access log and it contained these two lines confirming that the proxy was really used for both connections:

```
1762155964.003    101 192.168.1.112 TCP_TUNNEL/200 5313 CONNECT scc.suse.com:443 - HIER_DIRECT/99.83.188.102 -
1762155967.332    277 192.168.1.112 TCP_TUNNEL/200 8724 CONNECT installer-updates.suse.com:443 - HIER_DIRECT/75.2.43.231 -
```

## Screenshots

Screenshot of the log:

agama-self-update-proxy-log

## Fixed systemd status

Self-update service status in initramfs (started from emergency shell invoked with `rd.break` boot option):

self-update-status-initramfs

Self-update service status in the root image after fully booting the system, the exit status is missing although the log is correctly kept:

self-update-status-root

The service status after enabling it also in the root image (after fully booting the system):

self-update-status-root-fixed

## Updates

- Set the proxy rather via environment variables than via the command line options or URL query. That avoids showing or logging the proxy URL with a password.
- Disable logging the proxy URL in the code
- Tested also with authenticated proxy, the Squid proxy then contains these lines:
```
1762184588.402      0 192.168.1.105 TCP_DENIED/407 3892 CONNECT scc.suse.com:443 - HIER_NONE/- text/html
1762184588.763    359 192.168.1.105 TCP_TUNNEL/200 5313 CONNECT scc.suse.com:443 proxyuser HIER_DIRECT/99.83.188.102 -
1762184591.959      0 192.168.1.105 TCP_DENIED/407 3968 CONNECT installer-updates.suse.com:443 - HIER_NONE/- text/html
1762184592.517    555 192.168.1.105 TCP_TUNNEL/200 8724 CONNECT installer-updates.suse.com:443 proxyuser HIER_DIRECT/99.83.188.102 -
```

The "proxyuser" is the configured proxy account user name. The proxy is connected twice, in the first attempt it checks whether a password is required and which form (basic/digest) authentication is expected. In the second request the actual user name and password is sent.

### DIFF
--- a/live/live-root/usr/bin/live-self-update
+++ b/live/live-root/usr/bin/live-self-update
@@ -108,7 +108,6 @@ cleanup_chroot() {
   fi
 
   # log the result to be checked by openQA or Agama itself
-  # unfortunately systemd loses the service exit status after switching root :-/
   echo "$result" > "$RESULT_FILE"
 
   if [ -n "$NEWROOT" ]; then


live/live-root/usr/bin/live-self-update
```diff
@@ -117,17 +129,31 @@ query_reg_server() {
     info "Running query $query...."
 
     local curl_options
+    curl_options=()
```
teclator: np: I would initialize it with the default options you are appending later

`curl options="--silent --globoff --location --retry 3 --retry-connrefused --connect-timeout 20 --max-time 10 --fail --show-error"`

PR #2846, opened 2025-11-03
Fix broken image URL

pieterdd:
This fixes an image URL in README.md that no longer resolved correctly.


imobachgs:
Thanks for the fix!

imobachgs (approved):

PR #2853, opened 2025-11-06
Move product selection to the manager service

imobachgs:
## Problem

The product selection is handled by the software service. However, we decided to move it to the manager service because it is not only related to the software to install, but to many other settings.

## Solution

* The manager service handles the product selection.
* The software service emits progress during SetConfig.
* Software `SetConfig` action includes the `ProductSpec`. Others will follow.
* If there is a single product, the manager asks the software service to apply its default configuration.

> [!NOTE]
> The proposal is not working properly.

## Testing

- Added some unit test.
- Tested manually

### DIFF
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -49,6 +49,8 @@ pub enum ZyppDispatchError {
     ResponseChannelClosed,
     #[error("Target creation failed: {0}")]
     TargetCreationFailed(#[source] std::io::Error),
+    #[error(transparent)]
+    Progress(#[from] progress::service::Error),
 }
 
 #[derive(thiserror::Error, Debug)]


rust/agama-manager/src/service.rs
```diff
@@ -77,12 +91,43 @@ impl Service {
...
+        self.read_registries().await?;
+        if let Some(product) = self.products.default_product() {
+            let product = Arc::new(RwLock::new(product.clone()));
+            _ = self.software.cast(software::message::SetConfig::new(
+                Arc::clone(&product),
+                None,
+            ));
+            self.product = Some(product);
+        } else {
+            self.notify_no_product()
```
joseivanlopez, line 118: Good point!

rust/agama-manager/src/service.rs
```diff
@@ -177,12 +245,16 @@ impl MessageHandler<message::SetConfig> for Service {
                 .await?;
         }
 
-        if let Some(software) = &message.config.software {
+        if let Some(product) = &self.product {
```
joseivanlopez, line 248: Be aware this branch is not in sync with api-v2. This method was changed, so there will be conflicts.

rust/agama-manager/src/service.rs
```diff
@@ -77,12 +91,43 @@ impl Service {
...
+            system: manager::SystemInfo::default(),
+            product: None,
+            events,
         }
     }
 
+    /// Set up the service by reading the registries and determining the default product.
+    ///
+    /// If a default product is set, it asks the other services to initialize their configurations.
+    pub async fn setup(&mut self) -> Result<(), Error> {
```
joseivanlopez, line 108: NP: in storage we also need to do something at the beginning (reading issues from dbus service). The method was called `#start`. We should unify the names. I am fine with using `#setup` in both.

rust/agama-software/src/message.rs
```diff
@@ -51,16 +54,17 @@ impl Message for GetConfig {
...
 }
 
 impl<T: Send + 'static> Message for SetConfig<T> {
     type Reply = ();
 }
 
 impl<T> SetConfig<T> {
-    pub fn new(config: T) -> Self {
-        Self { config }
+    pub fn new(product: Arc<RwLock<ProductSpec>>, config: Option<T>) -> Self {
```
joseivanlopez, line 66: NP: what do you think about having a type alias for `Arc<RwLock<ProductSpec>>`?

rust/agama-software/src/zypp_server.rs
```diff
@@ -330,6 +346,7 @@ impl ZyppServer {
             }
         }
 
+        _ = progress.cast(progress::message::Next::new(Scope::Software));
```
joseivanlopez: Should we report communication errors?


```suggestion
        progress.cast(progress::message::Next::new(Scope::Software))?;
```

rust/agama-manager/src/service.rs
imobachgs: I would prefer to use "setup" because "start" is for something else IMHO (and we use it extensively).

rust/agama-manager/src/service.rs
imobachgs: There will be many things to change when merging to `api-v2`. So just one more :-) Thanks for the hint anyway.

rust/agama-software/src/message.rs
imobachgs: I am not sure about the name :-) In this case, I think it is important to know that you are behind a lock (and a `RwLock`, not a `Mutex`, that has a different API). So I would keep it as it is to do not loose context.

PR #2859, opened 2025-11-09
Add a new API to set resolvables for installation

imobachgs:
## Problem

Some Agama services need to ask for the installation of some resolvables.

## Solution

This PR adds a new *private* API that allows selecting resolvables for installation. The new end-point is `PUT /api/v2/private/resolvables/:id` and receives the list of resolvables:

```json
[{ "name": "btrfsprogs", "type": "package" }]
```

## Testing

- Added some unit tests (but not integration tests).
- Tested manually.

### DIFF
--- a/rust/agama-software/src/model/state.rs
+++ b/rust/agama-software/src/model/state.rs
@@ -50,7 +50,8 @@ pub struct SoftwareState {
 ///
 /// * [Product specification](ProductSpec).
 /// * [Software user configuration](Config).
-/// * [System information](agama_utils::api::software::SystemInfo).
+/// * [System information](SystemInfo).
+/// * [Agama software selection](SoftwareSelection).
 pub struct SoftwareStateBuilder<'a> {
     /// Product specification.
     product: &'a ProductSpec,
@@ -231,7 +232,6 @@ impl<'a> SoftwareStateBuilder<'a> {
 }
 
 impl SoftwareState {
-    // TODO: Add SoftwareSelection as additional argument.
     pub fn build_from(
         product: &ProductSpec,
         config: &Config,


rust/agama-manager/src/service.rs
```diff
@@ -173,15 +174,34 @@ impl Service {
...
-            .cast(issue::message::Set::new(Scope::Manager, vec![issue]));
+    fn set_product_from_config(&mut self, config: &Config) {
+        let product_id = config
+            .software
+            .as_ref()
+            .and_then(|s| s.product.as_ref())
+            .and_then(|p| p.id.as_ref());
+
+        if let Some(id) = product_id {
+            if let Some(product_spec) = self.products.find(&id) {
```
jreidinger, line 185: do we log somewhere if product is specified, but was not find?

rust/agama-manager/src/service.rs
```diff
@@ -390,3 +408,13 @@ impl MessageHandler<message::SetStorageModel> for Service {
             .await?)
     }
 }
+
+// FIXME: write a macro to forward a message.
```
jreidinger, line 414: :+1:

rust/agama-server/src/server/web.rs
```diff
@@ -109,6 +110,7 @@ pub async fn server_service(
             "/private/storage_model",
             get(get_storage_model).put(set_storage_model),
         )
+        .route("/private/resolvables/:id", put(set_resolvables))
```
jreidinger, line 113: what about get? In past we need it.

rust/agama-software/src/model/state.rs
```diff
@@ -180,19 +224,24 @@ impl<'a> SoftwareStateBuilder<'a> {
...
-            patterns,
-            packages: vec![],
+            resolvables,
             options: Default::default(),
         }
     }
 }
 
 impl SoftwareState {
     // TODO: Add SoftwareSelection as additional argument.
```
jreidinger: todo can be removed

rust/agama-software/src/model/state.rs
```diff
@@ -27,26 +27,39 @@ use agama_utils::{
...
+/// The SoftwareState is built by the [SoftwareStateBuilder] using different
+/// sources (the product specification, the user configuration, etc.).
 #[derive(Debug)]
 pub struct SoftwareState {
     pub product: String,
     pub repositories: Vec<Repository>,
-    // TODO: consider implementing a list to make easier working with them.
-    pub patterns: Vec<ResolvableName>,
-    pub packages: Vec<ResolvableName>,
+    pub resolvables: Vec<ResolvableState>,
```
jreidinger, line 43: Sorry, I do not get how this can work with just list of resolvables. There is no ID for them, right? How it can know how to transit between different states? I have to admit it starts to be more and more magic for me with so many different transitions of data to different states.

imobachgs: Actually I would say it is quite the contrary: we add patterns, packages, repositories, etc. At this point, you do not care about the list of resolvables (with the ID). You just need a set of things to get installed, and whether they are optional or not.

So the `SoftwareState` (I would rather prefer a better name, but it is the name we used for `NetworkState`) just tells you what you want to get in the system: repositories, packages, product, etc. And the Adapter, instead of having a lot of individual and small methods, just takes care of making it a reality. In a single place.

The workflow is like this:

1. The [SoftwareStateBuilder](https://github.com/agama-project/agama/blob/software-selection/rust/agama-software/src/model/state.rs#L54) puts together the information from all the sources (current system, configuration, etc.) . I need to include the SoftwareSelection in the docs, but it is mostly documented.
2. The resulting `SoftwareStage` is send to the `ModelAdapter`, which takes care of synchronizing zypp state.

rust/agama-server/src/server/web.rs
imobachgs: Not really. We do not use it and, in the current version of Agama, [is not implemented](https://github.com/agama-project/agama/blob/master/rust/agama-server/src/software/web.rs#L338).

rust/agama-manager/src/service.rs
imobachgs: No, we just report the issue. But I agree, we should log the problem here.

rust/agama-software/src/model/state.rs
jreidinger: so basically idea is that will unselect all resolvables ( which is not done now ) and then reselect everything again from scratch, right?

imobachgs: Yes, I will address that in a separate PR. We might need to extend the wrapper too.

imobachgs: Having said that, you do not need to reset everything from scratch, just to make sure that you sync the changes (adding the missing packages, removing the ones you do not want, etc.).

PR #2862, opened 2025-11-11
Adapt network to the new config based HTTP API (backend)

teclator:
## Problem

There is a new config based HTTP API and all the services needs to be adapted. The aim of this PR is to adapt the network backend service (replace #2840).

- https://github.com/agama-project/agama/pull/2715

## Solution

In the new architecture all the Agama modules will live in its own package implementing the actors/service as described in (https://github.com/agama-project/agama/pull/2715), but as network was already using a actors model using an enums approach it was decided to use the current service in the first iteration of changes.

### Changes done

- Added the actions for getting the system, proposal and current network configuration using the actual actions passing messages system based on enums.
- All the structs or types that are related to the API (needs to be serialized) has been moved to agama-utils, it was decided to not depend on agama-lib anymore. The types will be re-exported to be used by agama-network when needed.
- As the current merge is concatenating the connections array, the patch **(updateConfig)** will replace the connections with the given one when present, but is something we need to revisit and decide how to do it in a better way.
- Removed the network web service.
- As the proposal, system info and config contains similar collections and just change the absence or presence of some of them we have introduced the **NetworkCollection** and **NetworkConnectionCollections** in order to do the conversion between them.

### TODO

- Move to the actors/service model using traits instead of enums and adding the issues and progress monitor to the network service.
- Remove all the actions which are not needed anymore (old network web service).
- Add unit tests covering the new approach.

### DIFF
--- a/rust/agama-network/src/system.rs
+++ b/rust/agama-network/src/system.rs
@@ -170,18 +170,22 @@ impl NetworkSystemClient {
         self.actions.send(Action::GetConnections(tx))?;
         Ok(rx.await?)
     }
+
+    /// Returns the cofiguration from the current network state as a [Config].
     pub async fn get_config(&self) -> Result<Config, NetworkSystemError> {
         let (tx, rx) = oneshot::channel();
         self.actions.send(Action::GetConfig(tx))?;
         Ok(rx.await?)
     }
 
+    /// Returns the cofiguration from the current network state as a [Proposal].
     pub async fn get_proposal(&self) -> Result<Proposal, NetworkSystemError> {
         let (tx, rx) = oneshot::channel();
         self.actions.send(Action::GetProposal(tx))?;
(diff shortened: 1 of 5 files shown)


rust/agama-network/src/system.rs
```diff
@@ -87,6 +85,15 @@ impl<T: Adapter + Send + Sync + 'static> NetworkSystem<T> {
         Self { adapter }
     }
 
+    /// Returns a new instance of the network configuration system using the [NetworkManagerAdapter] for the system.
+    pub async fn for_network_manager() -> NetworkSystem<NetworkManagerAdapter<'static>> {
+        let adapter = NetworkManagerAdapter::from_system()
+            .await
+            .expect("Could not connect to NetworkManager");
```
imobachgs, line 92: I wonder if it would be better to return a Result instead.

PR #2863, opened 2025-11-11
add to C layer callbacks for download part of commit

jreidinger:
Adds to pkg commit action download callbacks. More will follow.

I try to limit scope of this pull request to setup base for some callbacks to start with.
Real progress calls is not implemented here as from API I am not sure how it will work with software scope which is now splitted into Download and Install parts. @imobachgs feel free to add notes, todos or even code of that progress calls.

### DIFF
--- a/rust/zypp-agama/src/callbacks.rs
+++ b/rust/zypp-agama/src/callbacks.rs
@@ -1,4 +1,4 @@
-use std::str::FromStr;
+use std::{fmt, str::FromStr};
 
 pub mod download_progress;
 pub mod pkg_download;
@@ -37,12 +37,13 @@ impl FromStr for ProblemResponse {
     }
 }
 
-impl ToString for ProblemResponse {
-    fn to_string(&self) -> String {
-        match self {
-            ProblemResponse::ABORT => "Abort".to_string(),
-            ProblemResponse::IGNORE => "Ignore".to_string(),
-            ProblemResponse::RETRY => "Retry".to_string(),
-        }
+impl fmt::Display for ProblemResponse {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            ProblemResponse::ABORT => "Abort",
+            ProblemResponse::IGNORE => "Ignore",
+            ProblemResponse::RETRY => "Retry",
+        };
+        write!(f, "{}", s)
     }
 }


rust/zypp-agama/src/callbacks.rs
```diff
@@ -146,7 +139,7 @@ where
         |url: String, error, description: String| callbacks.finish(&url, error, &description);
     let cb_finish = get_download_progress_finish(&finish_call);
 
-    let callbacks = DownloadProgressCallbacks {
+    let callbacks = zypp_agama_sys::DownloadProgressCallbacks {
```
mvidner: How about I add this before the `start_data` line (which is the first line in this file where the double `as` occurs)
```rs
        // No magic: this is a double type cast, done in two steps because
        // you don't want to change types too much accidentally,
        // thought the language designers.
        // 1. Drop the reference's safety guarantees (&mut T --> *mut T)
        //    Where T is written as _, "infer the complex closure type for me"
        // 2. Erase type information (*mut T --> *mut c_void)
```

mvidner: We could also wrap this in a helper, to be a bit more readable:
```diff
+fn as_c_void<T>(thing: &mut T) -> *mut c_void {
+    thing as *mut T as *mut c_void
+}
...
-  start_data: &mut start_call as *mut _ as *mut c_void,
+  start_data: as_c_void(&mut start_call),

```

mvidner: For completeness, we are currently using Rust 1.81 in Agama, but since Rust 1.82 we could use `&raw mut`:
```rs
start_data: &raw mut start_call as *mut c_void,
```
https://rust-for-c-programmers.com/ch25/25_3_raw_pointers_const_t_and_mut_t.html

___
Sources:

I asked both Claude (Sonnet 4.5) and Gemini (2.5 Flash)

> In a Rust codebase I see this line. I understand that `as` is a type cast, but why is it needed to cast twice?
`start_data: &mut start_call as *mut _ as *mut c_void,`

and

> OK, search the web and find the rationale for why a direct cast to `*mut c_void` is not allowed

jreidinger: maybe adding helper helps to have single place...also I would link there official source for unsafe rust..rustonomicon https://doc.rust-lang.org/nomicon/casts.html

rust/agama-software/src/callbacks/commit_download.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    fn problem(
+        &self,
+        name: &str,
+        error: zypp_agama::callbacks::DownloadResolvableError,
+        description: &str,
+    ) -> zypp_agama::callbacks::ProblemResponse {
+        // TODO: make it generic for any problem questions
+        // TODO: localization
```
mvidner: OK, the localization part is being held back by me,
but we could be marking up translatable texts using the crate that we already are using:
```rs
use gettextrs::gettext;
...
        let actions = [("Retry", gettext("Retry")), ("Ignore", gettext("Ignore"))];
```

rust/agama-software/src/callbacks/commit_download.rs
```diff
@@ -0,0 +1,78 @@
...
+        Self {
+            progress,
+            questions,
+        }
+    }
+}
+
+impl Callback for CommitDownload {
+    fn start_preload(&self) {
+        // TODO: report progress that we start preloading packages
```
mvidner, line 26: "field progress is never read" which I think is related to this TODO.
I would expect that to be part of this PR. A single callback, but going all the way across the stack

rust/agama-software/src/callbacks/commit_download.rs
```diff
@@ -0,0 +1,78 @@
...
+        let actions = [
+            ("Retry", labels[0].as_str()),
+            ("Ignore", labels[1].as_str()),
+        ];
+        let error_str = error.to_string();
+        let data = [("package", name), ("error_code", error_str.as_str())];
+        let question = QuestionSpec::new(description, "software.package_error.provide_error")
+            .with_actions(&actions)
+            .with_data(&data);
+        let result = Handle::current().block_on(async move {
```
mvidner, line 47: NP: I feel the bottom half of the fn is boilerplate and we will want to factor it out as more callbacks appear

rust/agama-software/src/callbacks/commit_download.rs
```diff
@@ -0,0 +1,78 @@
...
+            .unwrap_or(zypp_agama::callbacks::ProblemResponse::ABORT)
+    }
+
+    fn gpg_check(
+        &self,
+        _resolvable_name: &str,
+        _repo_url: &str,
+        _check_result: zypp_agama::callbacks::pkg_download::GPGCheckResult,
+    ) -> Option<zypp_agama::callbacks::ProblemResponse> {
+        // TODO: check if repo_url is DUD repo and kernel specify insecure dud option.
```
mvidner: Again, I expect this to be implemented in this PR.
GPG checks are more important than progress reports: people will nag us to fix the reporting, but we might forget a security check here.

mvidner (changes requested):
Reviewing the high level part (agama-software) only for now, I think the TODOs should get implemented now

rust/agama-software/src/callbacks/commit_download.rs
jreidinger: yes, that is something I do not know how to do it as I will have download phase and install phase for software scope, but not sure how to do it. Hoping that @imobachgs will fill that progress call parts

rust/agama-software/src/callbacks/commit_download.rs
jreidinger: well, what it do now in agama is that it will skip security check for DUD repo if dud.insecure is used. For rest it will just forward to provideDone - https://github.com/agama-project/agama/blob/master/service/lib/agama/software/callbacks/pkg_gpg_check.rb#L72
and exactly this check is kind of tricky for me as I need somehow to get into this callback
1. dud repo url
2. kernel parameters

jreidinger: so in the end I kind of hope that dud support in new api will provide us some helpers for such task like Dud struct with needed info that I can just pass here.

rust/agama-software/src/callbacks/commit_download.rs
jreidinger: yes, I am sure that at least more helpers for ProblemResponse handling will be useful. So lets see how it will end.

jreidinger: I plan to revisit it when more callbacks appears

rust/agama-software/src/callbacks/commit_download.rs
jreidinger: that reminds me that whole dud support in software stack should be readded

rust/agama-software/src/callbacks/commit_download.rs
mvidner: Fair enough, so just add a note to the PR description noting this

rust/zypp-agama/src/callbacks/pkg_download.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn problem(&self, _name: &str, _error: DownloadError, _description: &str) -> ProblemResponse {
+        ProblemResponse::ABORT
+    }
+    // callback after gpg check is done
+    fn gpg_check(
+        &self,
+        _resolvable_name: &str,
+        _repo_url: &str,
+        _check_result: GPGCheckResult,
+    ) -> Option<ProblemResponse> {
```
mvidner, line 137: Oh, why is the ProblemResponse wrapped in an Option?

mvidner: I mean, the API at this layer, the outer zypp-agama  module, should be understandable and documented for people who haven't implemented it :)
I can mostly guess the meaning of many things, but the `Option` occurring here is mysterious.

rust/agama-software/src/callbacks/commit_download.rs
mvidner: Sorry, your comments did not help I will ask better :)

Let's look at the callback signature and the related types:

<details>
<summary>gpg_check, ProblemResponse, GPGCheckResult</summary>

```rs
 fn gpg_check(
        &self,
        resolvable_name: &str,
        repo_url: &str,
        check_result: GPGCheckResult,
    ) -> Option<ProblemResponse>

pub enum ProblemResponse {
    RETRY,
    ABORT,
    IGNORE,
}

pub enum GPGCheckResult {
    Ok,         // Signature is OK.
    NotFound,   // Signature is unknown type.
    Fail,       // Signature does not verify.
    NotTrusted, // Signature is OK, but key is not trusted.
    NoKey,      // Public key is unavailable.
    Error,      // File does not exist or can't be opened.
    NoSig,      // File has no gpg signature (only digests).
}
```
</details>

`ProblemResponse` is fine, but I don't understand why it is in an `Option`

Resolvable name and repo URL are obvious, and so are most of `GPGCheckResult`. Except: `Ok` means that `gpg_check` is called also for *successful* GPG checks? Sounds legit but it should be commented at the `fn` declaration for the benefit of people who do not reat the `GPGCheckResult` contents.

So the most important use cases are:

1. A GPG check happens and succeeds. Does this implementation pass this success and lets the installation proceed? I would guess so. but if not, we will promptly know.
2. A GPG check fails for whatever reason. This minimal implementation should ABORT the installation, not silently IGNORE it. I need to see, in this file, that it is so. Either in some code, or in a comment. a `None` return value is not a sufficient explanation.

Note that I do not care about DUD now, even if that is what is written in the code comment and your initial replies.

jreidinger: OK, let me answer all of those:

1. It is Option, as one of decision can be I do not want to decide and let it pass to ProvideDone resulting in error. Basically here we catched gpg check result and react on it, like ignoring it early if it failed, so result is use this provided file.
2. According to our code I think it can also result in Ok state which means that it is test pass.
3. use case 1 I would say yes, at least it is what original ruby code do which OK result and it works previously.
4. use case 2 will postpone decision so DoneFinish will fail and installation fail and is aborted...None is not ignore...it is do not decide ( it is same as if callback is not implemented at all ).

rust/zypp-agama/src/callbacks/pkg_download.rs
```diff
@@ -0,0 +1,249 @@
...
+
+// generic trait to
+pub trait Callback {
+    // callback when start preloading packages during commit phase
+    fn start_preload(&self) {}
+    // callback when problem occurs during download of resolvable
+    fn problem(&self, _name: &str, _error: DownloadError, _description: &str) -> ProblemResponse {
+        ProblemResponse::ABORT
+    }
+    // callback after gpg check is done
```
mvidner:
```suggestion
    // callback after gpg check is done, even for a successful one
```

rust/agama-software/src/callbacks/commit_download.rs
mvidner: You have also pointed to the earlier Ruby implementation of the callback: https://github.com/agama-project/agama/blob/master/service/lib/agama/software/callbacks/pkg_gpg_check.rb#L68
It makes things clearer, but we need to document it at this library level

So:
1. Document (where this trait is declared, pkg_download.rs):
    > `None` means postpone the decision until FooBar,

    where FooBar is a name in agama-software or zypp-agama. Your mentioned ProvideDone or DoneFinish do not exist here, sorry

2. (I've added a Suggestion to make it clear)
3. good
4. related to 1.

mvidner: For the DUD case, let's write the idea of it in pseudocode already
```rs
// TODO: implement the DUD case:
// DUD (Driver Update Disk)
// ignore the error when the package comes from the DUD repository and
// the DUD package GPG checks are disabled via a boot option
//
// if repo_url == Agama::Software::Manager.dud_repository_url && ignore_dud_packages_gpg_errors? {
//   logger.info "Ignoring the GPG check failure for a DUD package"
//   return Ok(ProblemResponse::IGNORE);
// }
```

jreidinger: regarding question if I tested failed GPG case. I do not try it yet as we do not have yet working commit phase.

rust/zypp-agama/src/callbacks/pkg_download.rs
jreidinger: thanks for suggestion, but in the end I improve whole method documentation to be more explanatory

rust/zypp-agama/src/callbacks/download_progress.rs
```diff
@@ -0,0 +1,132 @@
+use std::os::raw::{c_char, c_int, c_void};
+
+use crate::{
+    callbacks::ProblemResponse,
+    helpers::{as_c_void, string_from_ptr},
+};
+
+// generic trait to
```
mvidner: to what?

rust/zypp-agama/src/callbacks/download_progress.rs
```diff
@@ -0,0 +1,132 @@
...
+pub trait Callback {
+    // callback when download start
+    fn start(&self, _url: &str, _localfile: &str) {}
+    // callback when download is in progress
+    fn progress(&self, _value: i32, _url: &str, _bps_avg: f64, _bps_current: f64) -> bool {
+        true
+    }
+    // callback when problem occurs
+    fn problem(&self, _url: &str, _error_id: i32, _description: &str) -> ProblemResponse {
```
mvidner: `error_id` being i32 is OK in the *-sys crate but not here IMHO. At least comment on the meaning, if not use an enum
Same fort `finish`

rust/zypp-agama/src/callbacks/download_progress.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::{
+    callbacks::ProblemResponse,
+    helpers::{as_c_void, string_from_ptr},
+};
+
+// generic trait to
+pub trait Callback {
+    // callback when download start
+    fn start(&self, _url: &str, _localfile: &str) {}
+    // callback when download is in progress
```
mvidner:
```suggestion
    // callback when download is in progress. return true: continue, false:abort
```

rust/zypp-agama/src/callbacks/pkg_download.rs
```diff
@@ -0,0 +1,283 @@
...
+                DownloadError::Invalid
+            }
+        }
+    }
+}
+
+impl ToString for DownloadError {
+    fn to_string(&self) -> String {
+        match self {
+            DownloadError::NoError => "NoError".to_string(),
```
mvidner: NP: this works too
```diff
         match self {
-            DownloadError::NoError => "NoError".to_string(),
-            DownloadError::NotFound => "NotFound".to_string(),
-            DownloadError::IO => "IO".to_string(),
-            DownloadError::Invalid => "Invalid".to_string(),
+            DownloadError::NoError => "NoError",
+            DownloadError::NotFound => "NotFound",
+            DownloadError::IO => "IO",
+            DownloadError::Invalid => "Invalid",
         }
+        .to_string()
```

but also Clippy says that we should implement `Display` instead to get `ToString` for free, https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#to_string_trait_impl

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
```diff
@@ -53,6 +61,67 @@ struct DownloadProgressCallbacks {
...
+
+enum DownloadResolvableFileError {
+  DRFE_NO_ERROR,
+  DRFE_NOT_FOUND,     // the requested Url was not found
+  DRFE_IO,            // IO error
+  DRFE_ACCESS_DENIED, // user authent. failed while accessing restricted file
+  DRFE_ERROR          // other error
+};
+
+// keep in sync with
```
mvidner, line 92: I wonder why this enum has a "keep in sync" note and explicit numeric values but the others above don't have those.
How do I tell which enum needs the stricter treatment?

In any case, would it be helpful to add libzypp source URLs for reference to more places?

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -118,6 +119,163 @@ struct DownloadProgressReceive : public zypp::callback::ReceiveReport<
...
+
+  void set_callbacks(DownloadResolvableCallbacks *callbacks_) {
+    callbacks = callbacks_;
+  }
+
+  virtual Action problem(zypp::Resolvable::constPtr resolvable_ptr, Error error,
+                         const std::string &description) {
+    if (callbacks != NULL && callbacks->problem != NULL) {
+      enum DownloadResolvableError error_enum;
+      switch (error) {
```
mvidner: NP: IMHO we should factor out these boilerplate enum conversions to separate short methods to make the rest more readable

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -118,6 +119,163 @@ struct DownloadProgressReceive : public zypp::callback::ReceiveReport<
...
+      case PROBLEM_IGNORE:
+        return zypp::repo::DownloadResolvableReport::IGNORE;
+      }
+    }
+    // otherwise return the default value from the parent class
+    return zypp::repo::DownloadResolvableReport::problem(resolvable_ptr, error,
+                                                         description);
+  }
+
+  virtual void pkgGpgCheck(const UserData &userData_r = UserData()) {
```
mvidner, line 172: NP: UserData is a zypp-core concept, basically a dict<String, Any>. I would like to see the type (schema) of the particular UserData for each callback

For example this one is documented at
https://github.com/openSUSE/libzypp/blob/3e26209c0fd4e004374fd7aa6aae8d15e6ab40c7/zypp-logic/zypp/ZYppCallbacks.h#L189
but it's hard to link to 😢

mvidner: https://opensuse.github.io/libzypp/ exists (Doxygen rendering) but has no search function :cry:

rust/zypp-agama/src/callbacks/download_progress.rs
jreidinger: yeah, I want to fill it up same as in other callback and need to add it.

rust/zypp-agama/src/callbacks/download_progress.rs
jreidinger: it is original callback attempt from zypp-agama, that I just copy here. But maybe it is oppurtinity to also improve it and also implement it for repository refresh

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
jreidinger: not sure about this...reason why to keep it in sync is that in C enum is just number and we use this feature to have problem and optional problem just problem with one additional entry appended.

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
jreidinger: agreed. Ideally I would like to see schema against we can check during compilation time as some data has mandatory parts and also optional ones.

mvidner (approved):

PR #2869, opened 2025-11-12
silence a warning about _1, and elided lifetimes

mvidner:
## Problem

1.

```
   Compiling zypp-agama-sys v0.1.0 (/checkout/rust/zypp-agama/zypp-agama-sys)
warning: zypp-agama-sys@0.1.0: In file included from /usr/include/boost/bind.hpp:30,
warning: zypp-agama-sys@0.1.0:                  from callbacks.cxx:1:
warning: zypp-agama-sys@0.1.0: /usr/include/boost/bind.hpp:36:1: note: '#pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.'
warning: zypp-agama-sys@0.1.0:    36 | BOOST_PRAGMA_MESSAGE(
warning: zypp-agama-sys@0.1.0:       | ^~~~~~~~~~~~~~~~~~~~
```

2.

```
warning: hiding a lifetime that's elided elsewhere is confusing
   --> agama-lib/src/storage/client/zfcp.rs:133:9
    |
133 |         &self,
    |         ^^^^^ the lifetime is elided here
134 |         controller_id: &str,
135 |     ) -> Result<ControllerProxy, ServiceError> {
    |                 ^^^^^^^^^^^^^^^ the same lifetime is hidden here
    |
    = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
```
## Solution

Follow the hint in the warning message

## Testing

It still compiles

### DIFF
--- a/rust/agama-lib/src/storage/client/zfcp.rs
+++ b/rust/agama-lib/src/storage/client/zfcp.rs
@@ -50,7 +50,7 @@ pub struct ZFCPClient<'a> {
     introspectable_proxy: IntrospectableProxy<'a>,
 }
 
-impl ZFCPClient<'_> {
+impl<'a> ZFCPClient<'a> {
     pub async fn new(connection: Connection) -> Result<Self, ServiceError> {
         let manager_proxy = ManagerProxy::new(&connection).await?;
         let object_manager_proxy = ObjectManagerProxy::builder(&connection)
@@ -130,9 +130,9 @@ impl ZFCPClient<'_> {
     }
 
     async fn get_controller_proxy(
-        &self,
+        &'a self,
         controller_id: &str,
-    ) -> Result<ControllerProxy, ServiceError> {
+    ) -> Result<ControllerProxy<'a>, ServiceError> {
         let dbus = ControllerProxy::builder(&self.connection)
             .path(ZFCP_CONTROLLER_PREFIX.to_string() + "/" + controller_id)?
             .build()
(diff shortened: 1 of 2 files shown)


jreidinger (approved):

rust/agama-lib/src/storage/client/zfcp.rs
```diff
@@ -50,7 +50,7 @@ pub struct ZFCPClient<'a> {
     introspectable_proxy: IntrospectableProxy<'a>,
 }
 
-impl ZFCPClient<'_> {
+impl<'a> ZFCPClient<'a> {
```
jreidinger, line 53: are both lifetime needed? I expect just one of them here

mvidner: Whichever you mean, I disagree:

1.
```rs
impl ZFCPClient<'a>
```
error[E0261]: use of undeclared lifetime name `'a`

2.
```rs
impl<'a> ZFCPClient
```
error[E0726]: implicit elided lifetime not allowed here

3.
```rs
impl<'a> ZFCPClient<'_>
```
This does compile but serves only to obscure the meaning.

jreidinger: in fact the third one is correct for me as it said impl use lifetime a, and let auto generate lifetime of struct itself. But  as it has same lenght, it does not matter much for me.

PR #2871, opened 2025-11-12
Drop unused Ruby services

imobachgs:
This PR drops the Ruby-based services that are not used anymore.

### DIFF
--- a/service/share/dbus.conf
+++ b/service/share/dbus.conf
@@ -36,14 +36,10 @@
     <allow eavesdrop="true"/>
 
     <allow own="org.opensuse.Agama1" />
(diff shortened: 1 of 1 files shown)


service/lib/agama/dbus/y2dir/modules/Package.rb
```diff
@@ -0,0 +1 @@
+Package.rb
```
joseivanlopez: Is this still needed?

imobachgs: Good question. It does not seem to be used by anyone. And what about the InstFunctions?

joseivanlopez (approved):
LGTM

PR #2874, opened 2025-11-12
D-Bus storage manager receives the product configuration

ancorgs:
Before this, the storage service was responsible for parsing the Agama configuration files and, based on an identifier received from D-Bus, loading the configuration for the product.

Now it receives the final configuration of the product via D-Bus.

**Bonus:** This also backports from master a commit used to make the tests more reliable. See https://github.com/agama-project/agama/pull/2860

### DIFF
--- a/service/lib/agama/config.rb
+++ b/service/lib/agama/config.rb
@@ -20,7 +20,6 @@
 # find current contact information at www.suse.com.
 
 require "agama/copyable"
-require "yast2/equatable"
 
 module Agama
   # Class representing the current product configuration.
@@ -31,13 +30,31 @@ module Agama
     # @return [Hash] configuration data
     attr_reader :data
 
-    eql_attr :data
-
     # Constructor
     #
     # @param config_data [Hash] configuration data
     def initialize(config_data = {})
       @data = config_data
+    end
+
+    # Updates the internal values if needed
+    #
+    # This update mechanism exists because the current implementation of Agama relies on the
+    # previous behavior of this class, in which a single shared Config object was constructed only
+    # once (when the configuration files were read) and the values returned by that object were
+    # later adjusted by subsequent calls to #pick_product on that shared object.
+    #
+    # To keep a similar behavior, this method provides a way to update the configuration values
+    # without creating a new Config object.
+    #
+    # @param new_config_data [Hash] new values for the configuration data
+    # @return [boolean] true if the internal values were really modified, false if there was no need
+    #   to do so because the values are the same
+    def update(new_config_data)
+      return false if new_config_data == @data
+
+      @data = new_config_data
+      true
     end
 
     # Default paths to be created for the product.
--- a/service/lib/agama/dbus/storage/manager.rb
+++ b/service/lib/agama/dbus/storage/manager.rb
@@ -184,14 +184,9 @@ module Agama
         # @param serialized_product [String] Serialized product config.
         # @param serialized_config [String] Serialized storage config.
         def configure(serialized_product, serialized_config)
-          system_changed = false
-
-          new_product_config = Config.new(JSON.parse(serialized_product))
-          if new_product_config != backend.product_config
-            backend.product_config = new_product_config
-            # Potential change in system - productMountPoints, encryptionMethods, volumeTemplates
-            system_changed = true
-          end
+          new_product_data = JSON.parse(serialized_product)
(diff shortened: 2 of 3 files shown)


service/lib/agama/dbus/storage/manager.rb
```diff
@@ -201,11 +181,46 @@ def recover_config_model
...
+            backend.probe
+            # Potential change in system - devices, issues, candidateX, availableX
+            system_changed = true
+          end
+
+          emit_system_changed if system_changed
+
+          next_progress_step(CONFIGURING_STEP)
+          config_json =
+            if serialized_config.empty?
```
joseivanlopez: I think the serialized config is `"null"` instead of empty.

joseivanlopez: Probably we can simply parse the serialized config. I would say it returns nil if the json is "null".

service/lib/agama/dbus/storage/manager.rb
```diff
@@ -201,11 +181,46 @@ def recover_config_model
...
+        # @param serialized_product [String] Serialized product config.
         # @param serialized_config [String] Serialized storage config.
-        def configure(serialized_config)
-          start_progress(1, CONFIGURING_STEP)
+        def configure(serialized_product, serialized_config)
+          system_changed = false
+
+          new_product_config = Config.new(JSON.parse(serialized_product))
+          if new_product_config != backend.product_config
+            backend.product_config = new_product_config
```
joseivanlopez: If I am not wrong, the `#product_config` is not updated in the proposal object. Than means the default config is always calculated according to the first configured product.

service/lib/agama/config.rb
```diff
@@ -20,173 +20,32 @@
...
-
-      elements.map do |element|
-        if !element.is_a?(Hash)
-          element
-        elsif arch_match?(element["archs"])
-          property ? element[property.to_s] : element
-        end
-      end.compact
+    # @param path [String|Pathname] File path
+    def self.from_file(path)
```
joseivanlopez: Is this really needed? If we are using it for loading data in tests, then we could define this oneliner directly in the test file.

service/lib/agama/dbus/storage/manager.rb
ancorgs: I kind of remember that in the past we used to pass `""` as a way to omit a string argument in D-Bus (since there are no optional arguments).

service/lib/agama/config.rb
ancorgs: Yes, I considered to create a helper in the tests. I guess it's time for it.

service/lib/agama/dbus/storage/manager.rb
ancorgs: Not sure if I follow.

service/lib/agama/dbus/storage/manager.rb
ancorgs: So the current implementation is more robust. It works with the empty string and also with "null".

service/lib/agama/config.rb
ancorgs: Although you are right. It's a oneliner so maybe not worth a helper. We have more complicated stuff repeated over some tests.

ancorgs: And is even a simpler one-liner. I added a commit simplifying it.

service/lib/agama/dbus/storage/manager.rb
joseivanlopez: Our D-Bus API expects a serialized JSON and an empty string is not a valid json. IMHO, we should complain if we get an empty string or whatever that is not "parseable" to json.

We used to use empty string as no value for string attributes/parameters, but this case is different. I would say bout "" and "what ever no json string" should be wrong.

service/lib/agama/dbus/storage/manager.rb
ancorgs: I see what you mean now. In the current code (before this PR), there is only one `Agama::Config` object that is constructed when the configuration files are read and that is shared by all other objects.

We change the values that object return by calling `#pick_product` on it, but it's the same shared object during the whole execution.

That would not happen now and is a problem.

ancorgs: It should be solved now.

PR #2875, opened 2025-11-12
Fix build of rubygem-agama-yast

imobachgs:
Fix build of the `rubygem-agama-yast` package. The proxy support was temporarily removed in https://github.com/agama-project/agama/pull/2871.


ancorgs (approved):
LGTM

PR #2881, opened 2025-11-13
Add product to storage calls

joseivanlopez:
Adapt storage service to send the config of the selected product through D-Bus (requires https://github.com/agama-project/agama/pull/2874). 

Bonus: makes manager fail when trying to apply a config with missing/unknown product.


rust/agama-manager/src/service.rs
```diff
@@ -28,18 +26,21 @@ use agama_utils::{
         Proposal, Scope, Status, SystemInfo,
     },
     issue,
-    license::{Error as LicenseError, LicensesRegistry},
-    products::{ProductSpec, ProductsRegistry, ProductsRegistryError},
+    licenses_registry::{self, LicensesRegistry},
```
imobachgs: I am not sure I like this `licenses_registry`. I think I would prefer `licenses::Registry`. The same below.

joseivanlopez: I agree. We are repeating the workd licenses (and the same for products) in both the module name and type. I also prefer something like `products::Registry` and `licenses::Registry`. What about the spec? `products::Spec` or  `products::ProductSpec`?

imobachgs: The problem with `products::Spec` is that you might have more *specs (e.g., `SoftwareSpec`) and then it does not look consistent at all.

imobachgs (approved):

PR #2882, opened 2025-11-14
[web]: adapt product selection to API v2

dgdavid:
This PR is not final code, but a step forward in adapting the web UI to the new API. It enables product selection and navigation to the overview page. Some sections of code, particularly those related to product registration, are temporarily disabled, as that functionality is not yet available in the new API.


teclator (approved):
There are still some check-types failing but so far, its LGTM.

web/src/api/config.ts
```diff
@@ -22,10 +22,12 @@
 
 import * as l10n from "~/api/l10n/config";
 import * as storage from "~/api/storage/config";
+import * as product from "~/api/product/config";
```
teclator, line 25: @dgdavid looks like the `~/api/product` folder was not added to git :(

web/src/api/proposal.ts
```diff
@@ -22,10 +22,12 @@
 
 import * as l10n from "~/api/l10n/proposal";
 import * as storage from "~/api/storage/proposal";
+import * as software from "~/api/software/proposal";
```
teclator, line 25: And the same with software.

PR #2884, opened 2025-11-14
Improve issues

joseivanlopez:
Changes in the issues API:

* Report issues as a plain list including the scope for each issue. 
* Remove severity and source.

This PR also removes some dead ruby code and speeds ruby tests up.

### DIFF
--- a/rust/agama-utils/src/api/issue.rs
+++ b/rust/agama-utils/src/api/issue.rs
@@ -39,6 +39,7 @@ pub type IssueMap = HashMap<Scope, Vec<Issue>>;
 pub struct Issue {
     pub class: String,
     pub description: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub details: Option<String>,
 }
 


rust/agama-storage/src/lib.rs
```diff
@@ -47,7 +47,7 @@ mod tests {
         async fn setup() -> Context {
             let (events_tx, _events_rx) = broadcast::channel::<Event>(16);
             let dbus = test::dbus::connection().await.unwrap();
-            let issues = issue::start(events_tx.clone(), dbus.clone()).await.unwrap();
+            let issues = issue::Service::starter(events_tx.clone()).start();
```
imobachgs, line 50: Perhaps we could consider using `Service::start()` and keep the `starter()` function to the case where you want to tweak the service. But, of course, it is out of scope.

service/test/agama/storage/configurator_test.rb
```diff
@@ -60,6 +60,9 @@
 
   before do
     mock_storage(devicegraph: scenario)
+    # To speed-up the tests
+    allow(Y2Storage::BootRequirementsStrategies::Analyzer)
+      .to receive(:bls_bootloader_proposed?).and_return(false)
```
imobachgs, line 65: 👍🏻 ⚡

PR #2892, opened 2025-11-18
[web]: add missing files

dgdavid:
Add some new files that were forgotten at https://github.com/agama-project/agama/pull/2882 and fix some types.

### DIFF
--- a/web/src/components/overview/SoftwareSection.tsx
+++ b/web/src/components/overview/SoftwareSection.tsx
@@ -27,15 +27,16 @@ import { _ } from "~/i18n";
 import { useProposal, useSystem } from "~/hooks/api";
 
 export default function SoftwareSection(): React.ReactNode {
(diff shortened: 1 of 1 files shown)


web/src/api/software/proposal.ts
```diff
@@ -0,0 +1,30 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { PatternsSelection } from "~/types/software";
+
+type Proposal = {
+  size?: string;
```
teclator: Are both optional at all?, I would say that for the proposal them are always present, isn't it?

dgdavid: Fixed in latest commit. Thanks for catching it.

PR #2893, opened 2025-11-18
Fixed generating profile from AY xml

mchf:
Fixed passing arguments when "agama config generate" is called with an AY profile

Cleaned some leftovers
- unhandled unwrap
- squashed some functions regarding validation reporting

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -4,7 +4,7 @@ Tue Nov 18 08:47:11 UTC 2025 - Michal Filka <mfilka@suse.com>
 - Fixed passing arguments when "agama config generate" with an
   AutoYast profile.
 - Removed unhandled unwrap, reorganized validation result reporting 
-- gh#agama-project/agama#2893
+  (gh#agama-project/agama#2893).
 
 -------------------------------------------------------------------
 Wed Nov 12 15:42:27 UTC 2025 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>


rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov 18 08:47:11 UTC 2025 - Michal Filka <mfilka@suse.com>
+
+- Fixed passing arguments when "agama config generate" with an
+  AutoYast profile.
```
imobachgs, line 5: Please, add a reference. Given that we do not have a proper bug report, let's use this PR (`gh#agama-project/agama#2893`).

rust/agama-server/src/profile/web.rs
```diff
@@ -111,7 +111,7 @@ impl ProfileBody {
     /// Expected format is a HashMap<String, String>, expecte keys are
     /// path, url or profile
     fn from_string(string: String) -> Self {
-        let map: HashMap<String, String> = serde_json::from_str(&string).unwrap();
+        let map: HashMap<String, String> = serde_json::from_str(&string).unwrap_or(HashMap::new());
```
imobachgs:
```suggestion
        let map: HashMap<String, String> = serde_json::from_str(&string).unwrap_or_default();
```

rust/agama-cli/src/config.rs
```diff
@@ -193,24 +193,19 @@ fn validate_local(url_or_path: CliInput, insecure: bool) -> anyhow::Result<Valid
     }
 }
 
-async fn validate_silently(
+async fn validate(
     client: &BaseHTTPClient,
     url_or_path: CliInput,
+    silent: bool,
```
imobachgs, line 199: To be honest, I do not like that `validate()` takes care of printing anything. And adding a `silent` argument does not make it any better.

However, let's consider that for a future refactoring during the `api-v2` adoption.

mchf: I hate this function. I was circling around it and tried several approaches and disliked them all :-/

rust/agama-server/src/profile/web.rs
mchf: +1

rust/package/agama.changes
```diff
@@ -1,3 +1,11 @@
...
+- Removed unhandled unwrap, reorganized validation result reporting 
+- gh#agama-project/agama#2893
```
```suggestion
  (gh#agama-project/agama#2893).
```

PR #2894, opened 2025-11-18
Improve testing of the Rust code

imobachgs:
## Problem

Testing our Rust code is kind of tricky. Many times, the code has to interact with the underlying system (systemd, NetworkManager, etc.) which makes hard even to test logic that it is not directly related.

## Solution

* Write a set of *builders* that allows to spawn services replacing some parts with testing versions. They allow to create services replacing the "adapters" (e.g., using a dummy network adapter).
* Add a set of `test_utils` modules with some utilities to use during tests.
* Expand the current tests, going from 21% to 26%.

## Testing

- Expanded the test coverage.

## Follow-up

- Create a test adapter for the software service.
- Consider to do not inject any service by default in the builders. I would allow to replace any piece, but it is not needed by now.

### DIFF
--- a/rust/agama-l10n/src/service.rs
+++ b/rust/agama-l10n/src/service.rs
@@ -77,14 +77,14 @@ pub enum Error {
 /// - A monitor which checks for changes in the underlying system (e.g., changing the keymap)
 ///   and signals the main service accordingly.
 /// - It depends on the issues service to keep the installation issues.
-pub struct Builder {
+pub struct Starter {
     model: Option<Box<dyn ModelAdapter + Send + 'static>>,
     issues: Handler<issue::Service>,
     events: event::Sender,
 }
 
-impl Builder {
-    /// Creates a new builder.
+impl Starter {
+    /// Creates a new starter.
     ///
     /// * `events`: channel to emit the [localization-specific events](crate::Event).
     /// * `issues`: handler to the issues service.
@@ -108,8 +108,11 @@ impl Builder {
         self
     }
 
-    /// Spawns the l10n service and the D-Bus monitor.
-    pub async fn spawn(self) -> Result<Handler<Service>, Error> {
+    /// Starts the service and returns a handler to communicate with it.
+    ///
+    /// The service uses a separate monitor to listen to system configuration
+    /// changes.
+    pub async fn start(self) -> Result<Handler<Service>, Error> {
         let model = match self.model {
             Some(model) => model,
             None => Box::new(Model::from_system()?),
@@ -126,11 +129,11 @@ impl Builder {
             events: self.events,
         };
         let handler = actor::spawn(service);
-        Self::spawn_monitor(handler.clone()).await;
+        Self::start_monitor(handler.clone()).await;
         Ok(handler)
     }
 
-    pub async fn spawn_monitor(handler: Handler<Service>) {
+    pub async fn start_monitor(handler: Handler<Service>) {
         match Monitor::new(handler.clone()).await {
             Ok(monitor) => monitor::spawn(monitor),
             Err(error) => {
@@ -160,8 +163,8 @@ pub struct Service {
 }
 
 impl Service {
-    pub fn builder(events: event::Sender, issues: Handler<issue::Service>) -> Builder {
-        Builder::new(events, issues)
+    pub fn starter(events: event::Sender, issues: Handler<issue::Service>) -> Starter {
+        Starter::new(events, issues)
     }
 
     fn get_proposal(&self) -> Option<Proposal> {
--- a/rust/agama-manager/src/lib.rs
+++ b/rust/agama-manager/src/lib.rs
@@ -88,7 +88,7 @@ mod test {
                 }
             });
 
-            let handler = test_utils::spawn_service(events_tx, dbus).await;
+            let handler = test_utils::start_service(events_tx, dbus).await;
             Context { handler }
         }
     }
--- a/rust/agama-manager/src/service.rs
+++ b/rust/agama-manager/src/service.rs
@@ -72,7 +72,7 @@ pub enum Error {
     NetworkSystem(#[from] network::NetworkSystemError),
 }
 
-pub struct Builder {
+pub struct Starter {
     questions: Handler<question::Service>,
     events: event::Sender,
     dbus: zbus::Connection,
@@ -84,7 +84,7 @@ pub struct Builder {
     progress: Option<Handler<progress::Service>>,
 }
 
-impl Builder {
+impl Starter {
     pub fn new(
         questions: Handler<question::Service>,
         events: event::Sender,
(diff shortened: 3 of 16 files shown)


rust/agama-l10n/Cargo.toml
```diff
@@ -18,6 +18,7 @@ zbus = "5.11.0"
 async-trait = "0.1.89"
 
 [dev-dependencies]
+test-context = "0.4.1"
```
jreidinger, line 21: :+1: this make code feel more like other test frameworks.

rust/agama-l10n/src/service.rs
```diff
@@ -65,6 +66,83 @@ pub enum Error {
     MissingProposal,
 }
 
+/// Builds and spawns the l10n service.
```
jreidinger, line 69: I have to say that I do not like much the change that basically make Builder to call spawn. It does not make much sense for me to say "Builder spawn service". if we go this way I would prefer to follow convention and have usual "new->with_*->build" flow for builder and for spawning having dedicated spawner or have it as method in that service like `ServiceBuilder::new().build()?.spawn()?`
This way it is inconsistent with rest of rust builder and is just confusing.

rust/agama-lib/src/profile.rs
```diff
@@ -116,19 +121,23 @@ pub struct ProfileValidator {
...
 impl ProfileValidator {
     pub fn default_schema() -> Result<Self, ProfileError> {
-        let relative_path = Path::new("agama-lib/share/profile.schema.json");
+        let relative_path = PathBuf::from("agama-lib/share/profile.schema.json");
         let path = if relative_path.exists() {
             relative_path
         } else {
-            Path::new("/usr/share/agama-cli/profile.schema.json")
+            let schema_dir =
+                std::env::var("AGAMA_SCHEMA_DIR").unwrap_or("/usr/share/agama-cli".to_string());
```
jreidinger: I would use constant for default SCHEMA_DIR location to make it easier to find or grep.

rust/agama-manager/src/lib.rs
```diff
@@ -30,3 +27,174 @@ pub use agama_l10n as l10n;
...
+        assert!(l10n_config.keymap.is_some());
+        assert!(l10n_config.timezone.is_some());
+
+        Ok(())
+    }
+
+    #[test_context(Context)]
+    #[tokio::test]
+    async fn test_patch_config_without_product(ctx: &mut Context) -> Result<(), Error> {
+        let input_config = Config {
```
jreidinger, line 180: to be honest in this case for better readability I would probably prefer to have that config in json and simply read json in this test and pass it. This will be easier to see exact data used for testing and it can make our life easier when we need to add regression testing with specific user data.

rust/agama-manager/src/service.rs
```diff
@@ -65,54 +65,175 @@ pub enum Error {
...
+        self
+    }
+
+    pub fn with_progress(mut self, progress: Handler<progress::Service>) -> Self {
+        self.progress = Some(progress);
+        self
+    }
+
+    /// Spawns the manager service.
+    pub async fn spawn(self) -> Result<Handler<Service>, Error> {
```
jreidinger: same note as above. In this case it is even more visible that it do building ( aka setting defaults for not specified stuff ) and then spawning together with service setup. Which I see as different tasks.

rust/agama-l10n/src/service.rs
imobachgs: I see your point. Perhaps we need a better name instead of having a separate concept for spawning the service. For instance, Rust [Command](https://doc.rust-lang.org/std/process/struct.Command.html) follows the builder pattern too, but it has better naming.

rust/agama-manager/src/test_utils.rs
```diff
@@ -0,0 +1,43 @@
...
+//! This module implements a set of utilities for tests.
+
+use agama_l10n::test_utils::spawn_service as spawn_l10n_service;
+use agama_network::test_utils::spawn_service as spawn_network_service;
+use agama_storage::test_utils::spawn_service as spawn_storage_service;
+use agama_utils::{actor::Handler, api::event, issue, progress, question};
+
+use crate::Service;
+
+/// Spawns a testing manager service.
```
jreidinger: is there some magic to avoid compilation of this or should be here `#cfg[test]` ?

rust/agama-manager/src/service.rs
imobachgs: For me it is similar to the standard `Command` struct.

rust/agama-manager/src/lib.rs
imobachgs: I agree. I will change it.

rust/agama-software/src/service.rs
```diff
@@ -63,14 +64,77 @@ pub enum Error {
     ZyppError(#[from] zypp_agama::errors::ZyppError),
 }
 
-/// Localization service.
+/// Budils and spawns the software service.
```
jreidinger: typo budils

rust/agama-storage/src/lib.rs
```diff
@@ -18,12 +18,79 @@
...
-
 pub mod service;
 pub use service::Service;
 
-mod client;
+pub mod client;
 pub mod message;
 mod monitor;
+
+pub mod test_utils;
```
jreidinger, line 28: as mentioned above I kind of expect to have this covered by cfg test to not be available outside of tests or inside test utils.

rust/agama-storage/src/service.rs
```diff
@@ -38,20 +39,82 @@ pub enum Error {
     Client(#[from] client::Error),
     #[error(transparent)]
     Issue(#[from] issue::service::Error),
+    #[error(transparent)]
+    Monitor(#[from] monitor::Error),
+}
+
+/// Builds and spawns the storage service.
+//
+// TODO: Allow replacing the client for testing purposes.
```
jreidinger: I am not sure if I get TODO as builder allow replacing client, not?

rust/agama-manager/src/test_utils.rs
imobachgs: No, it cannot be `#cfg[test]` because it cannot be used by other crates. The solution could be to put it behind a feature flag, but I decided to stop here (I think we have more urgent things to work on at this moment). But it could be a nice task for a Friday afternoon.

rust/agama-l10n/src/service.rs
imobachgs: Actually, you always spawn the service. After all, you communicate with it through messages. So I do not see the point of having separate `build` and `spawn` methods. Perhaps something like `Launcher` and `launch`. I do not like `Spawner` because the word does not exist :-) But I am open to other suggestions (perhaps something related to actors).

rust/agama-storage/src/service.rs
imobachgs: You are right, it was a left-over.

rust/agama-l10n/src/service.rs
jreidinger: Launcher and launch sounds good to me. I just find confusing to have Builder and spawn. Alternative can be also "Starter"/"start" or "Bootstrap"/"start" or "Runner"/"run"

rust/agama-manager/src/service.rs
jreidinger: In that case I would not call it `Builder` but `Task`. as you spawn tasks and not Builder

PR #2897, opened 2025-11-19
Reset the currently selected libzypp resolvabls before new selection

lslezak:
## Problem

- The currently selected resolvables are not reset before doing a new selection

## Solution

- Completely reset the libzypp pool

## Testing

- Tested manually


rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -319,12 +321,17 @@ void resolvable_unselect(struct Zypp *_zypp, const char *name,
   selectable->unset(value);
 }
 
+void resolvable_reset_all(struct Zypp *_zypp) noexcept {
+  MIL << "Resetting status of all resolvables" << std::endl;
+  for (auto &item: zypp::ResPool::instance()) item.statusReinit();
```
jreidinger: why not use statusReset which should be even faster as it does not check lock status?
https://github.com/openSUSE/libzypp/blob/929bff6fba87bc8204fad6483cebcf8923d77a32/zypp-logic/zypp/PoolItem.cc#L66

PR #2898, opened 2025-11-19
Identify whether a pattern was selected by the user

imobachgs:
## Problem

The software proposal does not include information about whether a pattern was selected by the user or by the installer itself.

```json
{
  "software": {
    "size": "3944742 KiB",
    "patterns": {
      "kde": "None",
      "multimedia": "Auto",
      "apparmor": "None",
      "basic_desktop": "None",
      "gnome": "Auto",
      "xfce": "None",
      "selinux": "Auto",
      "office": "Auto",
      "yast2_basis": "None",
      "yast2_desktop": "None",
      "yast2_server": "None"
    }
}
```

Additionally, there are other problems:

- Capitalization looks wrong.
- The "size" field is already formatted, whether it should be up to the client.
- The proposal is not updated when using the set resolvables API.

## Solution

* Distinguish installer (`auto`) from user-selected (`user`) patterns and use *camelCase*.
* Return the size as a number (in KiB).

```json
{
  "size": 6240723,
  "patterns": {
    "basic_desktop": "none",
    "office": "auto",
    "apparmor": "none",
    "kde": "none",
    "yast2_server": "none",
    "yast2_basis": "auto",
    "yast2_desktop": "none",
    "multimedia": "auto",
    "selinux": "none",
    "xfce": "none",
    "gnome": "user"
  }
}
```

## Other changes

The `Model` struct takes care of talking to libzypp. However, it relies on `ZyppServer` for communicating with libzypp, which runs on a dedicated thread. As part of this PR, we as simplified `ZyppServer` API a bit and, instead of implementing smaller actions (like getting the patterns metadata), it now returns `SystemInfo` and `Proposal` structs.

Last but not least, now the software proposal is recalculated when using the new set resolvables API.

### DIFF
--- a/rust/agama-software/src/service.rs
+++ b/rust/agama-software/src/service.rs
@@ -38,6 +38,7 @@ use agama_utils::{
 use async_trait::async_trait;
 use std::{path::PathBuf, process::Command, sync::Arc};
 use tokio::sync::{broadcast, Mutex, MutexGuard, RwLock};
+use url::Url;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
@@ -390,15 +391,25 @@ fn find_mandatory_repositories<P: Into<PathBuf>>(root: P) -> Vec<Repository> {
     repos
 }
 
+/// Returns the repository for the given directory if it exists.
 fn find_repository(dir: &PathBuf, name: &str) -> Option<Repository> {
     if !std::fs::exists(dir).is_ok_and(|e| e) {
         return None;
     }
 
+    let url_string = format!("dir:{}", dir.display().to_string());
+    let Ok(url) = Url::parse(&url_string) else {
+        tracing::warn!(
+            "'{}' is not a valid URL. Ignoring the repository.",
+            url_string
+        );
+        return None;
+    };
+
     Some(Repository {
         alias: name.to_string(),
         name: name.to_string(),
-        url: format!("dir:{}", dir.display().to_string()),
+        url: url.to_string(),
         enabled: true,
         predefined: true,
     })
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -241,6 +241,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tracing",
+ "url",
  "utoipa",
  "zypp-agama",
 ]
@@ -4309,7 +4310,7 @@ dependencies = [
  "getrandom 0.3.2",
  "once_cell",
  "rustix 1.0.5",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/rust/agama-software/Cargo.toml
+++ b/rust/agama-software/Cargo.toml
@@ -18,6 +18,7 @@ thiserror = "2.0.12"
 tokio = { version = "1.40.0", features = ["macros", "rt-multi-thread", "sync"] }
 tokio-stream = "0.1.16"
 tracing = "0.1.41"
+url = "2.5.7"
 utoipa = { version = "5.2.0", features = ["axum_extras", "uuid"] }
 zypp-agama = { path = "../zypp-agama" }
 


rust/agama-software/src/model.rs
```diff
@@ -95,33 +98,14 @@ impl Model {
...
-        let (tx, rx) = oneshot::channel();
-        self.zypp_sender
-            .send(SoftwareAction::GetPatternsMetadata(names, tx))?;
-        Ok(rx.await??)
-    }
 }
 
 #[async_trait]
 impl ModelAdapter for Model {
+    // FIXME: the product should be mandatory.
```
jreidinger: you mean like that model should not be instantiated before user select product? It can make sense as we cannot do much without this info.

rust/agama-software/src/model.rs
```diff
@@ -143,11 +127,23 @@ impl ModelAdapter for Model {
...
+            return Err(service::Error::MissingProduct);
+        };
+
+        let (tx, rx) = oneshot::channel();
+        self.zypp_sender
+            .send(SoftwareAction::GetSystemInfo(product_spec, tx))?;
+        let mut system_info = rx.await??;
+
+        // set "mandatory" field as the info to decide whether the repository is mandatory or not
+        // lives in this struct.
```
jreidinger: I am not sure if this comment is not more confusing. I would write here simply that "set as mandatory local repositories".

rust/agama-software/src/model/state.rs
```diff
@@ -279,26 +298,45 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
 /// Defines a resolvable to be selected.
 #[derive(Debug, PartialEq)]
-pub struct ResolvableState {
+pub struct SelectedResolvable {
     /// Resolvable name.
     pub resolvable: Resolvable,
-    /// Whether this resolvable is optional or not.
-    pub optional: bool,
+    /// The reason to select the resolvable.
+    pub reason: SelectedReason,
```
jreidinger, line 305: btw question is if we need to extend it to know if it is installer or product who selects it or not. So far I do not see need, but maybe in future we will need it. Libzypp supports 4 levels.

rust/agama-software/src/service.rs
```diff
@@ -177,6 +179,92 @@ impl Service {
...
+        tracing::info!("Wanted software state: {new_state:?}");
+        let model = self.model.clone();
+        let progress = self.progress.clone();
+        let issues = self.issues.clone();
+        let state = self.state.clone();
+        let events = self.events.clone();
+
+        tokio::task::spawn(async move {
+            let mut my_model = model.lock().await;
+            my_model.set_product(product);
```
jreidinger, line 212: do we need to always set product? I am not sure how time consuming the operation is.

rust/agama-software/src/service.rs
```diff
@@ -202,57 +290,25 @@ impl MessageHandler<message::GetConfig> for Service {
...
-                scope: Scope::Software,
-            });
-        });
+        self.update_proposal().await?;
 
         Ok(())
     }
 }
 
+impl Service {}
```
jreidinger: what is the point of this line?

rust/agama-software/src/service.rs
```diff
@@ -303,21 +359,52 @@ impl MessageHandler<message::Finish> for Service {
...
+        if self.product.is_some() {
+            self.update_proposal().await?;
+        }
         Ok(())
     }
 }
 
-const LIVE_REPO_DIR: &str = "/run/initramfs/live/install";
+const LIVE_REPO_DIR: &str = "run/initramfs/live/install";
+const DUD_REPO_DIR: &str = "var/lib/agama/dud/repo";
```
jreidinger, line 365: I think we will need in future also the one from kernel cmdline....also is somehow handled that new property @lslezak say about adding it as insecure if additional kernel param is added?

rust/agama-software/src/service.rs
```diff
@@ -303,21 +359,52 @@ impl MessageHandler<message::Finish> for Service {
...
     }
 
-    normalize_repository_url(LIVE_REPO_DIR, "/install").map(|url| Repository {
-        alias: "install".to_string(),
-        name: "install".to_string(),
-        url,
+    Some(Repository {
+        alias: name.to_string(),
+        name: name.to_string(),
+        url: format!("dir:{}", dir.display().to_string()),
```
jreidinger: this can be quite fragile as unix dir is not url encoded. I worry that it will have problems sooner or later.

rust/agama-software/src/zypp_server.rs
```diff
@@ -46,6 +47,8 @@ const GPG_KEYS: &str = "/usr/lib/rpm/gnupg/keys/gpg-*";
 pub enum ZyppDispatchError {
     #[error("Failed to initialize libzypp: {0}")]
     InitError(#[from] ZyppError),
+    #[error("libzypp error: {0}")]
+    ZyppServer(#[from] ZyppServerError),
```
jreidinger, line 51: this looks wrong to me as libzypp errors are only `zypp_agama::ZyppError`. ZyppServerError is much broader. If you need to add all of those, then I would add it transparent.

rust/agama-software/src/zypp_server.rs
```diff
@@ -333,22 +336,38 @@ impl ZyppServer {
         zypp.reset_resolvables();
 
         _ = progress.cast(progress::message::Next::new(Scope::Software));
+        let result = zypp.select_resolvable(
+            &state.product,
+            zypp_agama::ResolvableKind::Product,
+            zypp_agama::ResolvableSelected::Installation,
+        );
+        if let Err(error) = result {
+            let message = format!("Could not select the product '{}'", &state.product);
```
jreidinger, line 325: here it should be at least FIXME with the localization ( I know we do not solve yet how to pass variables into it )

rust/agama-software/src/zypp_server.rs
```diff
@@ -333,22 +336,38 @@ impl ZyppServer {
...
         for resolvable_state in &state.resolvables {
             let resolvable = &resolvable_state.resolvable;
             // FIXME: we need to distinguish who is selecting the pattern.
             // and register an issue if it is not found and it was not optional.
-            let result = zypp.select_resolvable(
-                &resolvable.name,
-                resolvable.r#type.into(),
-                zypp_agama::ResolvableSelected::Installation,
-            );
+            let reason = match resolvable_state.reason {
```
jreidinger: I think we can have it as From trait elsewhere, not?

rust/agama-software/src/zypp_server.rs
```diff
@@ -333,22 +336,38 @@ impl ZyppServer {
...
+
+            let result = zypp.select_resolvable(&resolvable.name, resolvable.r#type.into(), reason);
 
             if let Err(error) = result {
-                let message = format!("Could not select pattern '{}'", &resolvable.name);
-                issues.push(
-                    Issue::new("software.select_pattern", &message)
-                        .with_details(&error.to_string()),
-                );
+                if !resolvable_state.reason.is_optional() {
```
jreidinger: I would at least log failure for optional ones, so we know about it.

rust/agama-software/src/zypp_server.rs
```diff
@@ -430,40 +449,75 @@ impl ZyppServer {
...
+    async fn patterns(
+        &self,
+        product: &ProductSpec,
+        zypp: &zypp_agama::Zypp,
+    ) -> Result<Vec<Pattern>, ZyppServerError> {
+        let pattern_names: Vec<_> = product
+            .software
+            .user_patterns
+            .iter()
+            .map(|p| match p {
```
jreidinger: there is method `name` exactly for this purpose - https://github.com/agama-project/agama/blob/api-v2/rust/agama-utils/src/products.rs#L193

rust/agama-software/src/zypp_server.rs
```diff
@@ -430,40 +449,75 @@ impl ZyppServer {
...
+    ) -> Result<Vec<api::software::Repository>, ZyppServerError> {
+        let result = zypp
+            .list_repositories()?
+            .into_iter()
+            .map(|r| api::software::Repository {
+                alias: r.alias.clone(),
+                name: r.alias,
+                url: r.url,
+                enabled: r.enabled,
+                mandatory: false,
```
jreidinger: I think short comment explaining why setting mandatory  to false can help future readers.

rust/agama-software/src/zypp_server.rs
```diff
@@ -497,44 +551,48 @@ impl ZyppServer {
...
-                .map_err(|_| ZyppDispatchError::ResponseChannelClosed)?;
-            return Ok(());
-        };
-        let size = computed_mount_points.first().unwrap().used_size;
-        // TODO: format size
-        let size_str = format!("{size} KiB");
-        tracing::info!("Software size: {size_str}");
+        let computed_mount_points = zypp.count_disk_usage(mount_points)?;
+        let size = computed_mount_points.first().map(|m| m.used_size);
+        Ok(size.unwrap_or(0))
```
jreidinger: not sure if  it is correct here to mask invalid data from libzypp as zero. I would say that I prefer here error as data from libzypp failed for unknown reason.

web/src/api/software/proposal.ts
```diff
@@ -23,7 +23,9 @@
 import { PatternsSelection } from "~/types/software";
 
 type Proposal = {
-  size: string;
+  /** Used space in human-readable form */
```
jreidinger: I think it is exactly opposite. It is number in KiB, so not so easy human readable.

web/src/components/overview/SoftwareSection.tsx
```diff
@@ -25,18 +25,20 @@ import { Content, List, ListItem } from "@patternfly/react-core";
 import { isEmpty } from "radashi";
 import { _ } from "~/i18n";
 import { useProposal, useSystem } from "~/hooks/api";
+import xbytes from "xbytes";
```
jreidinger, line 28: :+1: no new dependency in nodejs? Even nicer.

rust/agama-software/src/model.rs
imobachgs: I would say that even the term "mandatory" is confusing. It does not refer exactly to "local repositories" but to those repositories that are kind of "fixed" or "automatic" (the off-line repo, the DUD, etc.). They cannot be altered using the configuration. Perhaps we should use `fixed` or `auto`. WDYT?

rust/agama-software/src/model/state.rs
imobachgs: We can do it when needed.

rust/agama-software/src/service.rs
imobachgs: It takes no time. Initially, I thought about removing this function, but one thing at a time :-)

rust/agama-software/src/service.rs
imobachgs: It is a test to check whether the reviewer is actually reading the code 😉. It is just a left-over, I will remove it.

rust/agama-software/src/service.rs
imobachgs: I am not sure. It is a different kind of repository and I think it is not signed.

rust/agama-software/src/service.rs
imobachgs: Can you be more specific, please?

rust/agama-software/src/zypp_server.rs
imobachgs: OK, I have simplified error handling a bit and removed some unused errors.

rust/agama-software/src/zypp_server.rs
imobachgs: Well, we need to check all the issues for the localization anyway.

rust/agama-software/src/zypp_server.rs
imobachgs: Sorry, I got the impression that the original implementation was kind of temporary because the hardcoded "btrfs" string and [the call to `unwrap`](https://github.com/agama-project/agama/blob/api-v2/rust/agama-software/src/zypp_server.rs#L521), etc. So I just wanted to have a safe default.

But you are right, I have introduced a new error variant for this case.

web/src/api/software/proposal.ts
imobachgs: Copy & paste error. IMHO, the API should provide a number and it is up to the client to decide how to display it.

rust/agama-software/src/model.rs
jreidinger: yeap, better name will help for sure...maybe "predefined"? Like that it is not the ones coming from user?

rust/agama-software/src/zypp_server.rs
imobachgs: Yes, you are right.

rust/agama-software/src/service.rs
jreidinger: well, here you construct url by just appending unix path after "dir:"...so any e.g. space like " " will kill it as it needs to be url encoded to "%20". What I mean here, is that we should sanitize or convert dir into url compatible string.

rust/agama-software/src/model.rs
imobachgs: I like predefined. I will adapt the code.

rust/agama-software/src/service.rs
imobachgs: I tend to agree, but in this case I would say the URLs are under our control.

jreidinger: It is, but as said in future if we extend it, it can fail. So at least document it in method documentation as note for anyone who would like to re-use method.

imobachgs: I am still not convinced of introducing a dependency just for this, but let's unblock the situation. I decided to go with [url](https://docs.rs/url/latest/url/) because it is in the dependencies tree anyway (through `zypp-agama`).

jreidinger (approved):

PR #2900, opened 2025-11-20
Adapted network UI to the new config HTTP API

teclator:
## Problem

The network backend has been already adapted to the new HTTP config-based API which causes the UI to be completely broken. Therefore, it needs to be adapted.

## Solution

Adapted the network UI to the new config-based HTTP API.

## Pending

- Try to move the current types to the API ones.
- Handle correctly the backend changes.

## Testing

- *Adapted unit tests*
- *Tested manually*

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -269,6 +269,7 @@ dependencies = [
  "agama-locale-data",
  "async-trait",
  "cidr",
+ "fs-err",
  "gettext-rs",
  "macaddr",
  "regex",
@@ -1670,6 +1671,15 @@ checksum = "0f158e3ff0a1b334408dc9fb811cd99b446986f4d8b741bb08f9df1604085ae7"
 dependencies = [
  "lazy_static",
  "num",
+]
+
+[[package]]
+name = "fs-err"
+version = "3.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "62d91fd049c123429b018c47887d3f75a265540dd3c30ba9cb7bae9197edb03a"
+dependencies = [
+ "autocfg",
 ]
 
 [[package]]
--- a/rust/agama-lib/src/error.rs
+++ b/rust/agama-lib/src/error.rs
@@ -73,7 +73,7 @@ pub enum ProfileError {
     Unreachable(#[from] TransferError),
     #[error("Jsonnet evaluation failed:\n{0}")]
     EvaluationError(String),
-    #[error("I/O error")]
+    #[error("I/O error: {0}")]
     InputOutputError(#[from] io::Error),
     #[error("The profile is not a well-formed JSON file")]
     FormatError(#[from] serde_json::Error),
--- a/rust/agama-utils/Cargo.toml
+++ b/rust/agama-utils/Cargo.toml
@@ -24,6 +24,7 @@ serde_yaml = "0.9.34"
 uuid = { version = "1.10.0", features = ["v4"] }
 cidr = { version = "0.3.1", features = ["serde"] }
 macaddr = { version = "1.0.1", features = ["serde_std"] }
+fs-err = "3.2.0"
 
 [dev-dependencies]
 test-context = "0.4.1"
--- a/rust/agama-utils/src/licenses.rs
+++ b/rust/agama-utils/src/licenses.rs
@@ -22,9 +22,9 @@
 
 use crate::api::manager::{InvalidLanguageCode, LanguageTag, License, LicenseContent};
 use agama_locale_data::get_territories;
+use fs_err::read_dir;
 use std::{
     collections::HashMap,
-    fs::read_dir,
     path::{Path, PathBuf},
 };
 use thiserror::Error;
@@ -33,7 +33,7 @@ use thiserror::Error;
 pub enum Error {
     #[error("Not a valid language code: {0}")]
     InvalidLanguageCode(#[from] InvalidLanguageCode),
-    #[error("I/O error")]
+    #[error("I/O error: {0}")]
     IO(#[from] std::io::Error),
 }
 
@@ -117,7 +117,7 @@ impl Registry {
     ///
     /// * `path`: directory to search translations.
     fn find_translations(path: &PathBuf) -> Result<Vec<LanguageTag>, std::io::Error> {
-        let entries = read_dir(path).unwrap().filter_map(|entry| entry.ok());
+        let entries = read_dir(path)?.filter_map(|entry| entry.ok());
 
         let files = entries
             .filter(|entry| entry.file_type().is_ok_and(|f| f.is_file()))
--- a/rust/install.sh
+++ b/rust/install.sh
@@ -52,6 +52,9 @@ install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-autoinstall.servic
 install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-scripts.service
 install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-web-server.service
 
+# instal licenses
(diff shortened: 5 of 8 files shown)


web/src/api/system.ts
```diff
@@ -23,13 +23,15 @@
 import * as l10n from "~/api/l10n/system";
 import * as storage from "~/api/storage/system";
 import { AddonInfo, Pattern, Product, Repository } from "~/types/software";
+import * as network from "~/api/network/system";
 
 type System = {
   l10n?: l10n.System;
-  storage?: storage.System;
+  network: network.System;
```
dgdavid, line 30: Is "network" the only one that is not optional? Why?

web/src/components/core/ChangeProductOption.test.tsx
```diff
@@ -43,13 +44,26 @@ const microos: Product = {
...
+};
+
 // let registrationInfoMock: RegistrationInfo;
 const mockSystemProducts: jest.Mock<Product[]> = jest.fn();
 
 jest.mock("~/hooks/api", () => ({
   ...jest.requireActual("~/hooks/api"),
   useSystem: (): ReturnType<typeof useSystem> => ({
     products: mockSystemProducts(),
+    network,
```
dgdavid, line 66: Same doubt here, why is network required?

dgdavid: It's, indeed, forcing to make a lot of changes in tests. I simply would like to understand the reason behind this decision.

web/src/components/network/BindingSettingsForm.tsx
```diff
@@ -60,7 +62,7 @@ function DevicesSelect({
   valueKey,
   ...formSelectProps
 }: DevicesSelectProps): React.ReactNode {
-  const devices = useNetworkDevices();
+  const devices = useDevices({ suspense: true });
```
dgdavid: We're trying to get ride of this `{ suspense: true }` option everywhere in favor of always using `useSuspenseQuery` instead. 

Same comment applies to all of similar lines.

web/src/components/network/InstallationOnlySwitch.test.tsx
```diff
@@ -40,10 +40,10 @@ const mockConnection = (options: Partial<ConnectionOptions> = {}) =>
...
   });
 
-jest.mock("~/queries/network", () => ({
-  ...jest.requireActual("~/queries/network"),
-  useConnectionPersistMutation: () => ({
-    mutateAsync: mockPersistMutation,
+jest.mock("~/hooks/network/config", () => ({
+  ...jest.requireActual("~/hooks/network/config"),
+  useConnectionMutation: () => ({
+    mutateAsync: mockConnectionMutation,
```
dgdavid, line 46: I remember we decided to keep mutations by now and change them, when possible, with the current approach: to send an _action_. Am I right?

web/src/hooks/network/config.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+  };
+  return useMutation(query);
+};
+
+const selectConnections = (data: network.Proposal): Connection[] =>
+  data.connections.map((c) => Connection.fromApi(c));
+
+const useConfig = (options?: QueryHookOptions): NetworkConfig => {
+  const func = options?.suspense ? useSuspenseQuery : useQuery;
```
dgdavid: As said, we can stop using `useQuery` in favor of `useSuspenseQuery` form almost all request to the new API. But I fully understand if you wanna keep this adaptation as it is and go for this kind of small refactor in a subsequent PR.

web/src/hooks/api/proposal/network.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  return data;
+};
+
+const useConnections = (options?: QueryHookOptions): Connection[] => {
+  const func = options?.suspense ? useSuspenseQuery : useQuery;
+  const { data } = func({
+    ...proposalQuery(),
+    select: (d: Proposal) => selectConnections(d.network),
```
dgdavid, line 54: Sadly, *I was wrong* the first time we saw the selectors thingy and they are hardly reusable. So, if the `selectConnections` will be not reused in other place I'd move the map directly here.

Also, looks like the usage of `useCallback` is needed here, although I'm not pretty sure.

dgdavid: Actually it's okish keeping it as it is, since you're defining a function that provides an stable reference for the callback. Check https://github.com/agama-project/agama/pull/2902#discussion_r2564905462

web/src/api/system.ts
teclator: because network is always present, it does not need to be probed, not sure why l10n is not.

web/src/components/core/ChangeProductOption.test.tsx
teclator: As commented, not sure why it is not present in the rest, but getting the system configuration for network will always contain the system information, I would expect the same for others.

web/src/components/network/BindingSettingsForm.tsx
teclator: yes, but I already had it adapted according to the existent system query, so, will need to adapt all the queries for removing it once the system query is adapted too.

web/src/components/network/InstallationOnlySwitch.test.tsx
teclator: Somehow yes, it will basically apply the config and the subscription to changes should do the rest, but by now I'm invalidating the queries.

web/src/hooks/network/config.ts
teclator: Yes, planned to do it in subsequent PRs but if it is done already in storage or others too can adapt it.

dgdavid: > in subsequent PRs

Better, IMHO

dgdavid (approved):

web/src/api/config.ts
```diff
@@ -20,17 +20,19 @@
...
-  product?: Product;
   l10n?: l10n.Config;
+  network?: network.Config;
+  product?: Product;
   storage?: storage.Config;
 };
 
 type Product = {
   id?: string;
 };
```
dgdavid, line 36: For some reason (multiple file reorganization) this has ended up here while I expect it at _~/api/config/software.ts_ or so. Well, actually I would expect it at _~/types/api/software_, but this is yet another discussion out of the scope of this PR 😜 

I guess we need a single PR fixing this kind of things once everything is adapted.

web/src/hooks/api/system/network.ts
```diff
@@ -0,0 +1,195 @@
...
+  GeneralState,
+  WifiNetworkStatus,
+  DeviceState,
+  WifiNetwork,
+  ConnectionState,
+} from "~/types/network";
+import { useInstallerClient } from "~/context/installer";
+import React, { useCallback } from "react";
+
+const selectSystem = (data: System | null): NetworkSystem =>
```
dgdavid, line 40: What about


```suggestion
const mapFromApi = (data: System | null): NetworkSystem =>
```

Or something better. The key is that "selectX" suggest you're going to select a subset of data while actually you are here simply transforming the whole set.

Sorry for such a huge NP 😅

dgdavid: Also, would be nice to have these methods/selectors/maps documented.

web/src/hooks/api/system/network.ts
```diff
@@ -0,0 +1,195 @@
...
+      }
+
+      return {
+        ...ap,
+        settings,
+        device,
+        status,
+      };
+    });
+};
```
dgdavid, line 116: I'm looking forward to the day we can kill this 😄 Meanwhile, we could maybe use a selector for such a big transformation. Not sure, but maybe we can. Probably, not using `useSystem()` here but the

```tsx
const { data } = useSuspenseQuery({
    ...systemQuery,
    select: useCallback((data) => {
      // transformation here
    }, []),
  });

  return data;
```

But if we're kill this at some point in the near future, don't waste more time here.

PR #2901, opened 2025-11-20
Security callbacks

jreidinger:
## Problem

There is missing in api-v2 security callbacks for GPG and checksum handling

## Solution

Use new API to implement it.

## Testing

:warning: NONE :warning:
TODO: test this as part of this PBI:
- [ ] https://trello.com/c/HRqBeOi7/5450-support-for-instinstallurl

### DIFF
--- a/rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
+++ b/rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
@@ -137,7 +137,7 @@ struct DownloadResolvableCallbacks {
 
 /**
  * @brief What to do with an unknown GPG key.
- * @see KeyRingReport::KeyTrust in libzypp
+ * @see zypp::KeyRingReport::KeyTrust in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/KeyRing.h
  */
 enum GPGKeyTrust {
   /** Reject the key. */
@@ -158,6 +158,7 @@ enum GPGKeyTrust {
  * an empty string if not available.
  * @param user_data User-defined data.
  * @return A GPGKeyTrust value indicating the action to take.
+ *  @see zypp::KeyRingReport::askUserToAcceptKey in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/KeyRing.h
  */
 typedef enum GPGKeyTrust (*GPGAcceptKeyCallback)(const char *key_id,
                                                  const char *key_name,
@@ -201,6 +202,9 @@ typedef bool (*GPGVerificationFailed)(const char *file, const char *key_id,
                                       const char *key_fingerprint,
                                       const char *repository_alias,
                                       void *user_data);
+/**
+ * @see zypp::DigestReport in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/Digest.h
+ */
 typedef bool (*ChecksumMissing)(const char *file, void *user_data);
 typedef bool (*ChecksumWrong)(const char *file, const char *expected,
                               const char *actual, void *user_data);


rust/agama-utils/src/api/question.rs
```diff
@@ -236,6 +236,29 @@ impl QuestionSpec {
...
+    /// # Example
+    ///
+    /// ```
+    ///   use agama_utils::api::question::QuestionSpec;
+    ///   let q = QuestionSpec::new("Continue?", "q.continue").with_yes_no_actions();
+    ///   assert_eq!(q.default_action.as_deref(), Some("No"));
+    /// ```
+    pub fn with_yes_no_actions(self) -> Self {
+        // we have to always call fresh gettext to ensure it is properly localized after change of locale
+        let labels = [gettextrs::gettext("Yes"), gettextrs::gettext("No")];
```
imobachgs, line 263: Why not import `gettext`, to be consistent with other usages? Additionally, it would be great if you add a small unit test.

rust/agama-utils/src/question.rs
```diff
@@ -52,9 +52,72 @@
...
+/// or a default policy), the function returns the answer and deletes the question
+/// from the service to clean up.
+///
+/// # Arguments
+/// * `handler` - A handler to the `question::Service`.
+/// * `question` - The `QuestionSpec` defining the question to be asked.
+///
+/// # Errors
+/// This function can return the following errors:
+/// * `AskError::QuestionNotFound`: If the question is deleted by another process before it can be answered.
```
imobachgs:
```suggestion
/// * [AskError::QuestionNotFound]: If the question is deleted by another process before it can be answered.
```

If you use this notation, it will link to `AskError::QuestionNotFound`. It applies to all the references on this module.

rust/agama-utils/src/question.rs
```diff
@@ -52,9 +52,72 @@
...
+/// * `AskError::Service`: If there is an error within the `question::Service` actor.
+/// * `AskError::Actor`: If there is a communication error with the actor system (e.g., the actor task has terminated).
+pub async fn ask_question(
```
imobachgs, line 103: I wonder why it is not part of the service's API. Perhaps when we replace the polling mechanism with something better we could move it there.

rust/agama-software/src/zypp_server.rs
```diff
@@ -199,12 +205,13 @@ impl ZyppServer {
     fn install(
         &self,
         zypp: &zypp_agama::Zypp,
-        download_callback: &commit_download::CommitDownload,
+        download_callback: &mut commit_download::CommitDownload,
+        security_callback: &mut security::Security,
```
imobachgs: I would rather prefer something like `callbacks::Security`? `security::Security` is repetitive and to vague. Moreover, in the future, we might have a `security` (actually, we have a `security` section already in the profile).

And the same for `commit_download`... what about `callbacks::CommitDownload`? Keeping the same files organization is fine. It is just a matter of exporting the public parts through the `callbacks` module.

rust/agama-software/src/callbacks/security.rs
```diff
@@ -0,0 +1,206 @@
...
+                "The file {file} from {repository_alias} is digitally signed with the \
+                following GnuPG key, but the integrity check failed: {key_id} ({key_name}). \
+                Use it anyway?"
+            )
+        };
+        let data = [("filename", file.as_str())];
+        let question = QuestionSpec::new(&text, "software.verification_failed")
+            .with_yes_no_actions()
+            .with_data(&data);
+        let result = Handle::current()
```
imobachgs, line 138: np: All functions are pretty much the same from this point. I wonder if we could reduce repetition a bit. But it can wait.

rust/agama-software/src/callbacks/security.rs
```diff
@@ -0,0 +1,206 @@
...
+                "The file {file} from {repository_alias} is not digitally signed. The origin \
+                and integrity of the file cannot be verified. Use it anyway?"
+            )
+        };
+        let data = [("filename", file.as_str())];
+        let question = QuestionSpec::new(&text, "software.unsigned_file")
+            .with_yes_no_actions()
+            .with_data(&data);
+        let result = Handle::current()
+            .block_on(async move { ask_question(&self.questions, question).await });
```
imobachgs, line 40: Please, explain at module's level why you need to this.

rust/agama-software/src/callbacks/security.rs
```diff
@@ -0,0 +1,206 @@
...
+        let labels = [gettext("Trust"), gettext("Skip")];
+        let actions = [("Trust", labels[0].as_str()), ("Skip", labels[1].as_str())];
+        let data = [
+            ("id", key_id.as_str()),
+            ("name", key_name.as_str()),
+            ("fingerprint", key_fingerprint.as_str()),
+        ];
+        let question = QuestionSpec::new(&text, "software.import_gpg")
+            .with_actions(&actions)
+            .with_data(&data)
```
imobachgs: np: Just my opinion, I find writing those vaules in-line more readable:

```rust
        let question = QuestionSpec::new(&text, "software.import_gpg")
            .with_actions(&actions)
            .with_data(&[
                ("id", key_id.as_str()),
                ("name", key_name.as_str()),
                ("fingerprint", key_fingerprint.as_str()),
            ])
            .with_default_action("Skip");
```

If constructing the data is more complex, then it is fine to split the logic.

However, it is fine as it is.

rust/agama-utils/src/api/question.rs
jreidinger: well, there is for the default action part already test in rustdoc..but I can add even for labels

rust/agama-software/src/zypp_server.rs
jreidinger: makes sense for me

rust/agama-software/src/callbacks/security.rs
jreidinger: I already fight with that asking questions. I agree that it is really repetitive.

rust/agama-utils/src/api/question.rs
imobachgs: Well, the test only checks the deafult_action, not that the question has the expected answers. Not to mention that it is the only method here that has a doctest (perhaps we should move the rest to doctests too, just for consistency).

mvidner:
How much testing did this get?

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
```diff
@@ -134,6 +134,120 @@ struct DownloadResolvableCallbacks {
   ZyppDownloadResolvableFileFinishCallback file_finish;
   void *file_finish_data;
 };
+
+/**
+ * @brief What to do with an unknown GPG key.
+ * @see KeyRingReport::KeyTrust in libzypp
```
mvidner:
```suggestion
 * @see zypp::KeyRingReport::KeyTrust in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/KeyRing.h
```

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
```diff
@@ -134,6 +134,120 @@ struct DownloadResolvableCallbacks {
...
+/**
+ * @brief Callback to decide whether to accept an unknown GPG key.
+ * @param key_id The ID of the GPG key.
+ * @param key_name The name of the GPG key.
+ * @param key_fingerprint The fingerprint of the GPG key.
+ * @param repository_alias The alias of the repository providing the key. Can be
+ * an empty string if not available.
+ * @param user_data User-defined data.
+ * @return A GPGKeyTrust value indicating the action to take.
+ */
```
mvidner, line 162:
```suggestion
 *  @see zypp::KeyRingReport::askUserToAcceptKey in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/KeyRing.h
 */
```

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
```diff
@@ -134,6 +134,120 @@ struct DownloadResolvableCallbacks {
...
+ * if not available.
+ * @param user_data User-defined data.
+ * @return true to continue, false to abort.
+ */
+typedef bool (*GPGVerificationFailed)(const char *file, const char *key_id,
+                                      const char *key_name,
+                                      const char *key_fingerprint,
+                                      const char *repository_alias,
+                                      void *user_data);
+typedef bool (*ChecksumMissing)(const char *file, void *user_data);
```
mvidner, line 208:
```suggestion
/**
 * @see zypp::DigestReport in https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/Digest.h
 */
typedef bool (*ChecksumMissing)(const char *file, void *user_data);
```

PR #2902, opened 2025-11-20
Fix storage UI

joseivanlopez:
Fix the storage UI after the migration to the new API.

Note: the failing tests will be adapted in a separate PR.

### DIFF
--- a/web/src/components/storage/BootSelection.tsx
+++ b/web/src/components/storage/BootSelection.tsx
@@ -73,10 +73,12 @@ export default function BootSelection() {
   const navigate = useNavigate();
   const devices = useDevices();
   const model = useModel();
-  const candidateDevices = filteredCandidates(useCandidateDevices(), model);
+  const allCandidateDevices = useCandidateDevices();
   const setBootDevice = useSetBootDevice();
   const setDefaultBootDevice = useSetDefaultBootDevice();
   const disableBootConfig = useDisableBootConfig();
+
+  const candidateDevices = filteredCandidates(allCandidateDevices, model);
 
   useEffect(() => {
     if (state.load || !model) return;


rust/agama-utils/src/api/query.rs
```diff
@@ -18,11 +18,19 @@
...
 use serde_json::Value;
 
 #[derive(Deserialize, utoipa::IntoParams, utoipa::ToSchema)]
 pub struct SolveStorageModel {
-    /// Serialized storage model.
+    #[serde(deserialize_with = "deserialize_model")]
     pub model: Value,
 }
+
+fn deserialize_model<'de, D>(deserializer: D) -> Result<Value, D::Error>
```
jreidinger, line 33: I think it would be good to document **why** it is needed.

service/lib/agama/storage/config_checkers/alias.rb
```diff
@@ -122,7 +122,7 @@ def partitioned_issue
               ),
               config.alias
             ),
-            kind: :partitioned_with_user
+            kind: IssueClasses::Config::OVERUSED
```
jreidinger, line 125: NP: overused sounds a bit strange. I kind of expect name like `ALREADY_USED` or `IN_USE`

service/lib/agama/storage/config_checkers/boot.rb
```diff
@@ -84,7 +84,7 @@ def invalid_alias_issue
           # TRANSLATORS: %s is replaced by a device alias (e.g., "boot").
           error(
             format(_("There is no boot device with alias '%s'"), device_alias),
-            kind: :no_such_alias
+            kind: IssueClasses::Config::ALIAS
```
jreidinger, line 87: name is also a bit inconsistent. I kind of expect something like `NO_ALIAS`

jreidinger: especially if you compare it with above `NO_ROOT`. I think issue should be either identified the part where is issue like `ROOT`, `ALIAS`, `ENCRYPTION` or by problem it describe like missing something, wrong entry, etc.

service/lib/agama/storage/config_checkers/search.rb
```diff
@@ -62,7 +62,7 @@ def search
 
         # @see Base
         def error(message)
-          super(message, kind: :search)
+          super(message, kind: IssueClasses::Config::SEARCH)
```
jreidinger, line 65: to be honest I am not sure on which level the issue  classes acts...as here is generic search issue, but when alias is not found or root device is not found, it has dedicated entry. Overall I found it inconsistent in naming and in usage. For me if I need add new constant there, I will be lost how it should look like.

web/src/api/config.ts
```diff
@@ -20,15 +20,18 @@
...
-import * as storage from "~/api/storage/config";
-import * as product from "~/api/product/config";
+import * as l10n from "~/api/config/l10n";
+import * as storage from "~/api/config/storage";
 
 type Config = {
+  product?: Product;
   l10n?: l10n.Config;
   storage?: storage.Config;
-  product?: product.Config;
```
jreidinger, line 30: this looks strange. I think @imobachgs should probably check it as I kind of expect we should have soon proper type for product and software config here.

web/src/api/proposal/software.ts
```diff
@@ -20,11 +20,20 @@
...
+type PatternsSelection = { [key: string]: SelectedBy };
+
+enum SelectedBy {
+  /** Selected by the user */
+  USER = 0,
+  /** Automatically selected as a dependency of another package */
+  AUTO = 1,
+  /** No selected */
+  NONE = 2,
+}
```
jreidinger, line 39: I smell high probability of collision with @imobachgs PR. So please coordinate ;)

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+  recommended: boolean;
+  description: string;
+  type: string;
+  release: string;
+};
+
+/**
+ * Enum for the reasons to select a pattern
+ */
+enum SelectedBy {
```
jreidinger, line 71: this looks strange. Why it is duplicitely also at https://github.com/agama-project/agama/pull/2902/files#diff-baa3da095b381d915a0bcd93c0c2a72395b98c122a2b8beea3f2bf4436ec2004R30 ?

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+type Pattern = {
+  /** Pattern name (internal ID) */
+  name: string;
+  /** Pattern category */
+  category: string;
+  /** User visible pattern name */
+  summary: string;
+  /** Long description of the pattern */
+  description: string;
+  /** {number} order - Display order (string!) */
```
jreidinger, line 38: I am really confused here what means that `string!` part in comment.

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+  order: number;
+  /** Icon name (not path or file name!) */
```
jreidinger, line 40: I think this comment deserve example how it should look like.

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+type Repository = {
+  repo_id: number;
+  alias: string;
+  name: string;
+  raw_url: string;
+  product_dir: string;
+  enabled: boolean;
+  loaded: boolean;
```
jreidinger, line 53: @imobachgs I kind of expect we will kill that loaded soon, right?

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+  NONE = 2,
+}
+
+type Product = {
+  /** Product ID (e.g., "Leap") */
+  id: string;
+  /** Product name (e.g., "openSUSE Leap 15.4") */
+  name: string;
+  /** Product description */
+  description?: string;
```
jreidinger, line 86: is it really optional? I think we have it defined always, not?

web/src/api/system/software.ts
```diff
@@ -0,0 +1,95 @@
...
+  /** Product icon (e.g., "default.svg") */
```
jreidinger, line 87: so here it is not icon name, but icon filename? ( so different to pattern icon?)

web/src/components/storage/ProposalFailedInfo.tsx
```diff
@@ -86,8 +88,11 @@ const Description = () => {
  *   - The generated proposal contains no errors.
  */
 export default function ProposalFailedInfo() {
-  const issues = useScopeIssues("storage");
-  if (issues.length === 0) return;
+  const configErrors = useConfigIssues();
+  const errors = useIssues("storage");
```
jreidinger: to be honest I do not get rename of issues to errors here. It is a bit confusing. I would probably keep it in original form `configIssues` and `issues`

web/src/components/storage/ProposalPage.tsx
```diff
@@ -300,10 +300,10 @@ function ProposalSections(): React.ReactNode {
  *  and test them individually. The proposal page should simply mount all those components.
  */
 export default function ProposalPage(): React.ReactNode {
-  const model = useConfigModel({ suspense: true });
+  const model = useStorageModel();
   const availableDevices = useAvailableDevices();
   const proposal = useProposal();
-  const issues = useScopeIssues("storage");
+  const configErrors = useConfigIssues();
```
jreidinger: and same here. I would pick one term and stick to it. So if it is issues, I would keep it consistent across code.

web/src/components/storage/ProposalPage.tsx
```diff
@@ -324,11 +324,16 @@ export default function ProposalPage(): React.ReactNode {
...
   }, [resetNeeded, setUiState]);
 
-  const fixable = ["no_root", "required_filesystems", "vg_target_devices", "reused_md_member"];
-  const unfixableErrors = issues.filter((e) => !fixable.includes(e.kind));
+  const fixable = [
+    "configNoRoot",
+    "configRequiredPaths",
+    "configOverusedPvTarget",
+    "configOverusedMdMember",
+  ];
```
jreidinger, line 332: I know that it is same as original code, but I really think that it is something that should come from backend as issue property. Or it is more like "fixable" in UI? and for some other issues, it is fixable, but UI does not support it?

web/src/components/storage/utils.test.ts
```diff
@@ -42,16 +42,15 @@ const volume = (properties: object = {}): Volume => {
   const testVolume: Volume = {
     mountPath: "/test",
     mountOptions: [],
-    target: "default",
-    fsType: "Btrfs",
+    fsType: "btrfs",
```
jreidinger, line 45: so here we really change capitalization of fs? Does it affect also user configuration provided for unattended installation?

web/src/hooks/api/config.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function useExtendedConfig(): Config | null {
+  return useSuspenseQuery(extendedConfigQuery)?.data;
+}
+
+function selectProduct(
+  products: system.Product[],
+  product: config.Product | null,
+): system.Product | null {
+  return products.find((p) => p.id === product?.id) || null;
```
jreidinger: I probably do not get how this method works as it looks to me like if it only search for given product.id. What exactly it did?

web/src/hooks/api/config/storage.ts
```diff
@@ -20,24 +20,23 @@
...
-import { QueryHookOptions } from "~/types/queries";
-import { extendedConfigQuery } from "~/hooks/api";
+import { useSuspenseQuery } from "@tanstack/react-query";
+import { extendedConfigQuery } from "~/hooks/api/config";
 import { putConfig, Response } from "~/api";
 import { Config } from "~/api/config";
 
-const resetConfig = (data: Config | null): Config => (!data ? {} : { ...data, storage: null });
+const removeStorageConfig = (data: Config | null): Config =>
+  !data ? {} : { ...data, storage: null };
```
jreidinger: why so negative. It is usually easier to read for humans in positive form
```suggestion
  data ? { ...data, storage: null } : {};
```

web/src/hooks/api/config/storage.ts
```diff
@@ -20,24 +20,23 @@
...
+    ...extendedConfigQuery,
+    select: removeStorageConfig,
   });
   return () => putConfig(data);
 }
 
-export { useResetConfig };
-export type { ResetConfigFn };
+export { useReset };
+export type { ResetFn as ResetConfigFn };
```
jreidinger: what is reason for this rename? I would not rename types unless there is good reason, as it make grepping source codes for its definition harder.

web/src/hooks/api/proposal.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  React.useEffect(() => {
+    if (!client) return;
+
+    // TODO: replace the scope instead of invalidating the query.
+    return client.onEvent((event) => {
+      if (event.type === "ProposalChanged") {
+        queryClient.invalidateQueries({ queryKey: ["extendedConfig"] });
+        queryClient.invalidateQueries({ queryKey: ["storageModel"] });
+        queryClient.invalidateQueries({ queryKey: ["proposal"] });
```
jreidinger, line 50: Hmm, I am not sure if we can get here into infinite loop when software will implement calculating size based on partitioning, then when software proposal change, it can result in recompute of storage proposal, that again can change software proposal regarding used size.

web/src/hooks/api/status.ts
```diff
@@ -0,0 +1,87 @@
...
+  queryKey: ["status"],
+  queryFn: getStatus,
+};
+
+function useStatus(): Status | null {
+  return useSuspenseQuery(statusQuery)?.data;
+}
+
+// FIXME: Borrowed from radashi 12.7. Simply import it after updating the dependency.
+function isArrayEqual<T>(array1: T[], array2: T[]): boolean {
```
jreidinger, line 40: I think I already see it in @dgdavid PR. Can it be just on single place?

web/src/hooks/api/config.ts
jreidinger: ah, now I get it, it basically filter out from config just product definition. So it combines from system list of products with selected product id to get specification of selected product....well, I think it deserve to be documented in function documentation.

web/src/hooks/storage/issue.ts
```diff
@@ -0,0 +1,39 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { useSuspenseQuery } from "@tanstack/react-query";
+import { issuesQuery } from "~/hooks/api/issue";
+import { Issue } from "~/api/issue";
+
+function selectIssues(issues: Issue[]) {
+  return issues.filter((i: Issue) => i.scope === "storage" && i.class !== "proposal");
```
jreidinger, line 28: I see this quite fragile comparison. When any new non config class will be added, then there will be problem. What about having prefixes in classes like `config.no_root` or `config.label`?

web/src/storage/boot.ts
```diff
@@ -20,9 +20,9 @@
  * find current contact information at www.suse.com.
  */
 
-import { model } from "~/types/storage";
-import { apiModel } from "~/api/storage";
-import { copyApiModel } from "~/helpers/storage/api-model";
+import { model } from "~/storage";
+import { model as apiModel } from "~/api/storage";
+import { copyApiModel } from "~/storage/api-model";
```
jreidinger, line 23: here it is a bit messy. There is model in ~/storage, then there is a model in ~/api/storage which is imported as apiModel and in ~/storage/api-model is copyApiModel. Naming here is really strange.

web/src/storage/device.ts
```diff
@@ -20,9 +20,8 @@
  * find current contact information at www.suse.com.
  */
 
-// FIXME: the backend should provide the type.
-
-import { system, proposal } from "~/api/storage";
+import { storage as system } from "~/api/system";
+import { storage as proposal } from "~/api/proposal";
```
jreidinger: looks like we no longer use types prefix for types. In such case I think it will really help if we start using `import type { .. } from ...;` syntax to emphasize that we are interested only in types, which in the end also helps optimization as it is instruction only for typescript and js do not need to import anything.

jreidinger: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

web/src/storage/model.ts
```diff
@@ -20,11 +20,85 @@
  * find current contact information at www.suse.com.
  */
 
-import { apiModel } from "~/api/storage";
-import { model } from "~/types/storage";
+/**
+ * Model types.
+ *
+ * Types that extend the apiModel by adding calculated properties and methods.
```
jreidinger, line 26: I really dislike this...There is apiModel which is called model and here it defined another model which is extended version of model. This is extra confusing. I think it should really use different names like `extendedModel` or `modelCalculated`. In rust where it is easy to use namespace like `proposal::Storage` and `system::Storage` it is easy to have types which identical names, but in different namespace. But in typescript having types with identical name is quite hard to read the code, especially if they collide ( basically you always have to check imports in given file to see what it talks about ).

web/src/storage/partition.ts
```diff
@@ -42,11 +42,11 @@ function indexByPath(device: Partitionable, path: string): number {
...
  * */
 function addPartition(
-  apiModel: apiModel.Config,
+  apiModel: model.Config,
   list: "drives" | "mdRaids",
   listIndex: number | string,
   data: data.Partition,
-): apiModel.Config {
+): model.Config {
   apiModel = copyApiModel(apiModel);
```
jreidinger: and here is another example of confusing names as type is no longer apiModel, but method to copy it still contain it.

web/src/storage/search.ts
```diff
@@ -20,10 +20,10 @@
  * find current contact information at www.suse.com.
  */
 
-import { apiModel } from "~/api/storage";
-import { model } from "~/types/storage";
-import { copyApiModel, findDevice, findDeviceIndex } from "~/helpers/storage/api-model";
-import { buildModel } from "~/helpers/storage/model";
+import { model as apiModel } from "~/api/storage";
+import { model } from "~/storage";
+import { build as buildModel } from "~/storage/model";
```
jreidinger: to be honest with namespace support as it is in js, I am not sure if it is good idea to have so generic names like build where it is not clear at all what it builds unless you import it under old name. So from my POV I prefer old name, where it was at least clear what it builds ( sadly no longer as there are two storage model with identical name, but it is discussed elsewhere ).

web/src/storage/model.ts
joseivanlopez: The extended version of the model is going to dead. There will be only one model. Note that the changes in this PR are intended to make the code work with the new API, but all the bad designs are kept. Refactoring will come in another iteration.

web/src/api/config.ts
joseivanlopez: Note that this PR does not define the new types related to sofware and products. It only places the existing types in the correct location. All those types will be revisited when adapting the sofware and product code to the new API.

web/src/api/system/software.ts
joseivanlopez: As mentioned, these types will be revisited.

web/src/api/system/software.ts
joseivanlopez: I only moved the existing code. This will be revisited.

web/src/api/system/software.ts
joseivanlopez: Again, software types are not introduced here and they will be revisted. The same applies for the rest of comments about software types.

web/src/components/storage/ProposalPage.tsx
joseivanlopez: Exactly, it is the list of issues that can be flixable by the current UI . This list will change meanwhile the UI evolves.

web/src/components/storage/utils.test.ts
joseivanlopez: No, it does not affect. Note that the profile schema has not changed. The profile already uses types without capitalization. This change here is because the new "system schema", which is more consistent with the rest of schemas than the info provided by the old API.

web/src/hooks/api/config/storage.ts
joseivanlopez: It was a leftover. Good catch!

web/src/hooks/api/proposal.ts
joseivanlopez: I am not sure if I get your point. Why does invalidating a query end up recalculating a proposal?

web/src/hooks/api/status.ts
joseivanlopez: Yes, it is only moved here as part of solving a git conflict. Note that this function is going to be drop once we update the radashi dependency.

web/src/hooks/storage/issue.ts
joseivanlopez: Actually @ancorgs initially proposed something like that. @ancorgs , what do you think?

web/src/storage/boot.ts
joseivanlopez: I totally agree. Using two different models is a mess. The plan is to keep only the model from the API. This PR already names the model from the API as "model", renaming it as "apiModel" in the places in which the code were already using the other (extended) model.

web/src/hooks/storage/issue.ts
ancorgs: Yes. That was my first thought. And I even proposed that to Imo and Iván.

But since using those prefixes is not a general idea used in other scopes and the comparison with "proposal" looked simpler, I decided to keep it simple.

I can bring back the original idea and add a "config" prefix.

web/src/storage/boot.ts
joseivanlopez: If you don't mind, I would not introduce more changes now because this part is going to change.

web/src/storage/partition.ts
joseivanlopez: After refactoring, it will look like this:

~~~ts
function addPartition(
  model: model.Config,
  list: "drives" | "mdRaids",
  listIndex: number | string,
  data: data.Partition,
): model.Config {
  const newModel = copyModel(model);
~~~

joseivanlopez: In fact, in this code we can already avoid `apiModel` because the "extended" model is not use.

web/src/components/core/ChangeProductOption.tsx
```diff
@@ -26,13 +26,13 @@ import { useHref, useLocation } from "react-router";
...
 import { _ } from "~/i18n";
-import { useSystem } from "~/hooks/api";
+import { useSystem } from "~/hooks/api/system";
 
 /**
  * DropdownItem Option for navigating to the selection product.
  */
 export default function ChangeProductOption({ children, ...props }: Omit<DropdownItemProps, "to">) {
-  const { products } = useSystem({ suspense: true });
+  const { products } = useSystem();
```
dgdavid, line 35: Thanks for addressing this. It makes your PR a bit longer, but it pay off.

dgdavid: I mean, addressing the suspense thingy across all the code.

web/src/components/core/InstallerOptions.tsx
```diff
@@ -551,10 +553,10 @@ export default function InstallerOptions({
   const location = useLocation();
   const {
     l10n: { locales },
-  } = useSystem({ suspense: true });
+  } = useSystem();
   const { language, keymap, changeLanguage, changeKeymap } = useInstallerL10n();
-  const { state } = useStatus({ suspense: true });
-  const selectedProduct = useSelectedProduct();
+  const { state } = useStatus();
+  const selectedProduct = useProduct();
```
dgdavid, line 559: 👌

web/src/components/l10n/KeyboardSelection.tsx
```diff
@@ -25,19 +25,16 @@ import { Content, Flex, Form, FormGroup, Radio } from "@patternfly/react-core";
...
 export default function KeyboardSelection() {
   const navigate = useNavigate();
-  const {
-    l10n: { keymaps },
-  } = useSystem({ suspense: true });
-  const {
-    l10n: { keymap: currentKeymap },
-  } = useProposal({ suspense: true });
+  const keymaps = useSystem()?.keymaps;
+  const currentKeymap = useProposal()?.keymap;
```
dgdavid, line 37: I'm all for not opening discussions about enforcing strict conventions, specially considering that we don't have a formal style guide in place yet, However, I personally prefer to continue using destructuring where possible when it comes to extract data from hooks. It looks like the new useSystem hook might be returning_ undefined_, but I’m wondering if there's a (clean) way to avoid it and return an empty collection instead.

dgdavid: Same applies for similar usage in other files in this PR.

web/src/components/storage/BootSelection.tsx
```diff
@@ -67,12 +67,12 @@ type BootSelectionState = {
...
  */
-export default function BootSelectionDialog() {
+export default function BootSelection() {
   const location = useLocation();
   const [state, setState] = useState<BootSelectionState>({ load: false });
   const navigate = useNavigate();
   const devices = useDevices();
-  const model = useModel({ suspense: true });
+  const model = useModel();
   const candidateDevices = filteredCandidates(useCandidateDevices(), model);
```
dgdavid: Although this line isn’t part of the changes in this PR, it caught my attention. While it doesn’t technically break any [React hook rules](https://react.dev/reference/rules/rules-of-hooks), I find this a bit unconventional and potentially risky. The most seen practice is to call hooks at the top level of a function and then work with the data they return. It probably won't cause any issues now, but I’m concerned that using a hook inline like this could lead (at some point in the future) to subtle issues down the line that might be difficult to track and debug. It might just be a matter of personal preference, but it could be worth reconsidering this pattern for future changes

web/src/hooks/api/system/l10n.ts
```diff
@@ -20,11 +20,18 @@
...
-function timezoneOffset(timezone: Timezone) {
-  return tzOffset(timezone.id, new Date());
+const selectSystem = (data: System | null): l10n.System => data?.l10n;
+
+function useSystem(): l10n.System | null {
+  const { data } = useSuspenseQuery({
+    ...systemQuery,
+    select: selectSystem,
+  });
+  return data;
```
dgdavid, line 34: This could be the place where return an empty object instead of undefined.

```tsx
return data || {}
```

web/src/hooks/api/config.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function useProduct(): system.Product | null {
+  const products = useSystem()?.products;
+  const { data } = useSuspenseQuery({
+    ...extendedConfigQuery,
+    select: useCallback(
+      (data: Config | null): system.Product | null => selectProduct(products || [], data?.product),
```
dgdavid: Is it really necessary to extract product selection to an external function? If there is no a strong reason to do so, I would keep the find directly here.

web/src/hooks/api/issue.ts
```diff
@@ -0,0 +1,60 @@
...
+const useIssuesChanges = () => {
+  const queryClient = useQueryClient();
+  const client = useInstallerClient();
+
+  React.useEffect(() => {
+    if (!client) return;
+
+    return client.onEvent((event) => {
+      if (event.type === "IssuesChanged") {
+        queryClient.invalidateQueries({ queryKey: ["issues"] });
```
dgdavid, line 54: Not to be part of this PR, but we need to address the query keys thingy. Ideally, making it query export its id to be then reused. This would help to avoid typos in the future in a quite cheap way.

web/src/hooks/api/proposal.ts
dgdavid, line 50: Not directly related to this PR, but when we tackle the queryKey thingy, we should consider following the recommendation for using query key factories (see https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories, and https://github.com/agama-project/agama/pull/2447). By doing so, we could easily invalidate queries under the _"proposal/"_ namespace without worrying about potential leftovers in the future.

web/src/hooks/api/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+  queryKey: ["storageModel"],
+  queryFn: getStorageModel,
+};
+
+function useStorageModel(): model.Config | null {
+  return useSuspenseQuery(storageModelQuery)?.data;
+}
+
+const solvedStorageModelQuery = (apiModel?: model.Config) => ({
+  queryKey: ["solvedStorageModel", JSON.stringify(apiModel)],
```
dgdavid, line 37: Weird queryKey, not sure if we want to follow that path.

web/src/storage/device.ts
joseivanlopez: Good point!

web/src/hooks/api/system.ts
```diff
@@ -0,0 +1,59 @@
...
+import React from "react";
+import { useSuspenseQuery, useQueryClient } from "@tanstack/react-query";
+import { getSystem } from "~/api";
+import { useInstallerClient } from "~/context/installer";
+import { System } from "~/api/system";
+
+const systemQuery = {
+  queryKey: ["system"],
+  queryFn: getSystem,
+};
```
dgdavid, line 32: For a future PR, 

```suggestion

const KEY = "system";

const systemQuery = {
  queryKey: [KEY],
  queryFn: getSystem,
};
```

and then export KEY.

web/src/hooks/api/system/l10n.ts
joseivanlopez: If we decide to always return a list, then I would move it to the select function.

web/src/components/l10n/KeyboardSelection.tsx
joseivanlopez: I am not sure if at some point we will need to distinguish between no data vs not fetched.

web/src/components/storage/BootSelection.tsx
joseivanlopez: I think that line is totally equivalent to:

~~~ts
 let candidateDevices = useCandidateDevices();
 candidateDevices = filteredCandidates(candidateDevices, model);
~~~

Probably I would replace all this by a new hook with a select over candidate devices.

web/src/hooks/api/config.ts
joseivanlopez: Good point. It can be moved here to the callback.

web/src/hooks/api/proposal.ts
joseivanlopez: Yes, we should use some robust way for defining queries. But I think we will not have nested queries anymore.

web/src/hooks/api/storage.ts
joseivanlopez: This code is not new. Any suggestion?

web/src/hooks/api/proposal.ts
jreidinger: not much related to queries, more about interdependencies of software and storage proposals...but I think if it occurs, we will soon notice it. So lets ignore for now.

dgdavid: IMHO we have to take time to think about it. If these queries conceptually belongs to "proposal" I do not see a reason for not following key factory recommendation and put them under "proposal". It will make things better to work with.

web/src/hooks/api/storage.ts
dgdavid: Not yet, as I do not have the full picture in mind at this time. Let's go ahead and change it when proceed.

web/src/api/system/software.ts
joseivanlopez: No idea. I was only moving existing types.

web/src/components/storage/BootSelection.tsx
dgdavid: > I think that line is totally equivalent to:
> 
> ```ts
>  let candidateDevices = useCandidateDevices();
>  candidateDevices = filteredCandidates(candidateDevices, model);
> ```

Yes, it is.

As said, it is a personal preference and a kind of bet for a code style, but for sure I would go for this approach keeping hooks always written in same way

const whatever = useHook();

> 
> Probably I would replace all this by a new hook with a select over candidate devices.

Cool, but I still think the suggestion applies. I would prefer to keep hook calls as visible and clean as possible. Maybe it's just me being overly cautious with hooks, but it does not hurt to keep things a bit more predictable and clear, IMHO.

web/src/components/l10n/KeyboardSelection.tsx
dgdavid: Don't understand what you mean.

web/src/hooks/api/proposal.ts
joseivanlopez: I don't get your point. Which queries do you think can be under proposal?

dgdavid: Those you're invalidating here. The fact that you're invalidating them under certain event suggests that these queries could be grouped in some way. "Proposal" was just an example, but the key takeaway is that if you're frequently invalidating a set of queries together, they likely share a common parent or key.

web/src/components/l10n/KeyboardSelection.tsx
joseivanlopez: If we always report data (e.g., `{keymaps: []}, locales: []}`), then we are not able to know if we have already fetched the data from the api or the data is not available yet. But as said, I don't know if we will need that.

web/src/hooks/api/config.ts
dgdavid: While doing https://github.com/agama-project/agama/pull/2900 review and after re-reading the Tanstack blog (https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option) I've realized that to achieve stable selector they could be either, 

  * a stable function reference
  * memoized with useCallback
  
 So, having it as an external function as you had make it stable without the need of useCallback usage. Unless I have misunderstood something. And if you ask me, I prefer the first option. The less hooks, the better.

dgdavid: Ok, I have said nothing 😅 🙈 

Here you were using an anonymous function for the selector in order to access to the `products` defined before. 

🤷‍♂️

web/src/api/config.ts
teclator: If it place the existing types in the correct location why product was removed from Config? I would expect some clarification somewhere because it is an storage PR but this is a product change.

web/src/hooks/api/system/software.ts
```diff
@@ -20,8 +20,18 @@
  * find current contact information at www.suse.com.
  */
 
-type Config = {
```
teclator, line 23: So, the file was removed here but not added in `api/config/product.ts`

service/lib/agama/storage/config_checkers/search.rb
ancorgs: You are right. There should be no generic classes. Each class should correspond to a well defined issue. I'm working on a PR to fix that.

PR #2903, opened 2025-11-21
Fix an opaque I/O error by installing eulas

mvidner:
## Problem

In `api-v2`, agama-web-server crashes at startup with

> Error: I/O error

(tested via /testing_using_container.sh)

Cc @ancorgs @jreidinger 

## Solution

- install the files it is looking for (But are they the right files? The dir content looks patchy!)
- Improve the error message
  -  by using [fs-err](https://docs.rs/fs-err/latest/fs_err/), a NEAT library that provides filenames to ALL such errors (once we `use` it)

> Error: I/O error: failed to read directory `/usr/share/agama/eula`: No such file or directory (os error 2)

## Testing

- Tested manually
- There already were unit tests for this, using the files directly from the git tree. Unit but not integration.

## Screenshots

No

## Documentation

Not crashing is expected

### DIFF
--- a/rust/agama-utils/src/licenses.rs
+++ b/rust/agama-utils/src/licenses.rs
@@ -117,7 +117,7 @@ impl Registry {
     ///
     /// * `path`: directory to search translations.
     fn find_translations(path: &PathBuf) -> Result<Vec<LanguageTag>, std::io::Error> {
-        let entries = read_dir(path).unwrap().filter_map(|entry| entry.ok());
+        let entries = read_dir(path)?.filter_map(|entry| entry.ok());
 
         let files = entries
             .filter(|entry| entry.file_type().is_ok_and(|f| f.is_file()))


rust/install.sh
```diff
@@ -52,6 +52,9 @@ install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-autoinstall.servic
 install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-scripts.service
 install6 -D -t "${DESTDIR}${unitdir}" "${SRCDIR}"/share/agama-web-server.service
 
+# instal licenses
+install6 -D -t "${DESTDIR}${datadir}"/agama/eula/license.final "${SRCDIR}"/share/eula/license.final/*
```
mvidner, line 56: This does look a bit suspicious, more like testing data than the real thing, but we can improve upon that while we have a not-crashing agama-web-server.

```console
$ ls /usr/share/agama/eula/license.final/
license.es.txt  license.txt  license.zh_CN.txt
```

jreidinger (approved):

PR #2904, opened 2025-11-21
Fixed missing translations (bsc#1250059)

lslezak:
## Problem

- Some missing translations in the web UI
- https://bugzilla.suse.com/show_bug.cgi?id=1250059#c0

## Details

The code at the top level like

```
const message = _("foo")
```
does not work properly because the code is evaluated too early, even before the component is actually displayed and before the translations are loaded from the server.

## Solution

There are basically two solutions:

1. Use `N_()` function instead of `_()` at the top level and then use `_()` when using the constant later.
2. Move the constant into a function (or function component) which is called (and evaluated/retranslated) in every rendering of the component.

## Testing

- Tested manually.
- Because most of the problems were in special storage technologies like iSCSI or DASD I was able to test only the error message in the software. And even that was a bit tricky, to trigger the error I configured an invalid installation repository with the `inst.instal_url=http://example.com` boot parameter.
- I also verified the generated POT file after this change to ensure no translation was lost. The generated POT is the very same (except the line numbers in comments) as before the change.

## Additional changes

- Enabled eslint check which finds the top level translations to avoid similar problems in the future
- Use NodeJS v22 in the CI build, that's the same version as in SLE-16. Moreover v18 is obsolete and out of support for several months. :open_mouth: 

## Screenshots

Original SLE-16 system without the patch:
agama-missing-translation-software

Fixed system after applying this patch:
agama-missing-translation-software-fixed

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Fri Nov 21 08:17:06 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
-- Fixed missing translations (bsc#1250059)
+- Fixed missing translations (error cases for software repos, iSCSI, DASD; bsc#1250059)
 
 -------------------------------------------------------------------
 Mon Oct 27 15:51:48 UTC 2025 - Ladislav Slezák <lslezak@suse.com>


web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Nov 21 08:17:06 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Fixed missing translations (bsc#1250059)
```
mvidner: want to specify which translations?
```suggestion
- Fixed missing translations (error cases for software repos, iSCSI, DASD; bsc#1250059)
```

web/src/components/storage/dasd/StatusFilter.tsx
```diff
@@ -31,18 +31,18 @@ import {
...
   value: string;
   onChange: SelectProps["onSelect"];
 };
 
 const options = {
-  all: _("all"),
-  active: _("active"),
-  read_only: _("read_only"),
-  offline: _("offline"),
+  all: N_("all"),
```
mvidner, line 42: NP:
this does fix the translation code but the translators may get it wrong because of the missing context,
often the grammatical gender gets mismatched.
And I see these are applied in different contexts... where the subject is always "devices"?

web/src/components/network/WifiConnectionForm.tsx
```diff
@@ -162,7 +162,8 @@ export default function WifiConnectionForm({ network }: { network: WifiNetwork }
                 <FormSelectOption
                   key={security.value}
                   value={security.value}
-                  label={security.label}
+                  /* eslint-disable agama-i18n/string-literals */
+                  label={_(security.label)}
```
mvidner, line 166: This is good: an instance where the `_(expression)` is in the same file as the matching `N_("literal")`

web/src/components/storage/iscsi/LoginForm.jsx
```diff
@@ -56,7 +56,8 @@ export default function LoginForm({ node, onSubmit: onSubmitProp, onCancel }) {
   };
 
   const startupFormOptions = Object.values(NodeStartupOptions).map((option, i) => (
-    <FormSelectOption key={i} value={option.value} label={option.label} />
+    /* eslint-disable agama-i18n/string-literals */
+    <FormSelectOption key={i} value={option.value} label={_(option.label)} />
```
mvidner, line 60: The `_(expression)` applies to `NodeStartupOptions` which are properly declared as `N_(...)` in another file from this PR

web/src/components/software/SoftwarePage.tsx
```diff
@@ -119,7 +119,7 @@ const ReloadSection = ({
       </>
     ) : (
       <>
-        {errorMsg}{" "}
+        {_(errorMsg)}{" "}
```
mvidner, line 122: NP: Related to testing, I would love to have a handy collection of repositories broken in various ways: missing/mismatched signatures, keys... we should ask the zypp people or start it ourselves

web/eslint.config.mjs
```diff
@@ -41,8 +41,7 @@ export default [
   {
     rules: {
       "agama-i18n/string-literals": "error",
-      // TODO: enable this check after fixing all top level translations (bsc#1250059)
-      // "agama-i18n/top-level-translation": "error",
+      "agama-i18n/top-level-translation": "error",
```
imobachgs, line 44: 👏🏻

imobachgs (approved):
LGTM, but @mvidner has a point about missing context (I guess the problem was already there anyway, right?).

PR #2909, opened 2025-11-25
web: add status monitor for background tasks

dgdavid:
This PR introduces a lightweight status monitor for background tasks (aka _progresses_), allowing users to track ongoing activity in the app. Please, note that **the design and placement of this feature are expected to evolve as we gain further insights into optimal UI integration**. Proposed one is just an simple starting point.

localhost_8080_ (9)

Additionally, the PR **refactors the state logic in _App.tsx_** to reduce unnecessary layout re-renders, which were causing unexpected UI issues such as menus and floating elements closing unexpectedly. This change separates concerns by **moving state management and redirect logic to an internal component defined outside the main one**, improving the overall stability of the layout.

To make everything work, a new hook for listening to and update progress changes has been added.

### DIFF
--- a/web/src/components/core/ProgressStatusMonitor.tsx
+++ b/web/src/components/core/ProgressStatusMonitor.tsx
@@ -34,6 +34,7 @@ import { useStatus } from "~/hooks/api";
 import { Progress as StatusProgress } from "~/api/status";
 import { sprintf } from "sprintf-js";
 import { _, N_, n_ } from "~/i18n";
+import displayStyles from "@patternfly/react-styles/css/utilities/Display/display";
 
 type DetailProps = {
   tasks: StatusProgress[];
@@ -55,10 +56,7 @@ const SCOPE_NAMES = {
  */
 const DetailsHeader = ({ tasks }: DetailProps) => {
   const total = tasks.length;
-
-  return total
-    ? sprintf(n_("%s task active", "%s tasks active", total), total)
-    : _("No tasks active");
+  return sprintf(n_("%s task active", "%s tasks active", total), total);
 };
 
 /**
@@ -108,10 +106,9 @@ export default function ProgressStatusMonitor() {
 
   return (
     <Popover
-      showClose={idle}
+      showClose={false}
       minWidth="400px"
       position="bottom-end"
-      hideOnOutsideClick={!idle}
       headerContent={<DetailsHeader tasks={tasks} />}
       bodyContent={
         <Stack hasGutter>
@@ -119,7 +116,12 @@ export default function ProgressStatusMonitor() {
         </Stack>
       }
     >
-      <Button variant="plain" isDisabled={idle} isLoading={!idle} />
+      <Button
+        variant="plain"
+        className={idle ? displayStyles.displayNone : displayStyles.displayBlock}
+        isDisabled={idle}
+        isLoading={!idle}
+      />
     </Popover>
   );
 }


web/src/components/core/ProgressStatusMonitor.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Renders the header text for the popover, indicating how many tasks are
+ * currently active.
+ */
+const DetailsHeader = ({ tasks }: DetailProps) => {
+  const total = tasks.length;
+
+  return total
+    ? sprintf(n_("1 task active", "%s tasks active", total), total)
```
jreidinger: hardcoded 1 is a bad practice as not all languages has singular just for 1. See [`https://alistapart.com/article/pluralization-for-javascript/`](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html)

jreidinger: Ah, I want original gettext doc at https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

web/src/components/core/ProgressStatusMonitor.tsx
```diff
@@ -0,0 +1,125 @@
...
+ *
+ *  - The number of active tasks
+ *  - A list of progress bars, one per task scope
+ *
+ * Intended as a lightweight monitor for background activity.
+ */
+export default function ProgressStatusMonitor() {
+  const { progresses: tasks } = useStatus();
+
+  const iddle = tasks.length === 0;
```
jreidinger: is there reason for this `iddle` typo? I do not think it is keyword in ts.

web/src/components/core/ProgressStatusMonitor.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  return (
+    <Popover
+      showClose={tasks.length === 0}
```
jreidinger: So if I get logic, you get this task popover when there is some task and it do not allow to close it unless all of them are done, right?

web/src/components/core/ProgressStatusMonitor.tsx
```diff
@@ -0,0 +1,125 @@
...
+      minWidth="400px"
+      position="bottom-end"
+      hideOnOutsideClick={!iddle}
```
jreidinger: and here it uses iddle...so why not above?

web/src/hooks/api.ts
```diff
@@ -59,6 +60,54 @@ function useStatus(options?: QueryHookOptions): Status | null {
   return data;
 }
 
+// Borrowed from radashi 12.7. Simply import it after updating the dependency.
```
jreidinger: I would mark this as TODO: or FIXME: for easier grepping

web/src/components/core/ProgressStatusMonitor.tsx
dgdavid: Arg! Sure, a mistake from my side. It should be %s either, but I have sent draft code :man_shrugging:

web/src/components/core/ProgressStatusMonitor.tsx
dgdavid: Obviously, there is no reason for the typo. Will fix it.

web/src/components/core/ProgressStatusMonitor.tsx
dgdavid: A bit tricky, indeed.

Actually, the user can toggle the Popover visibility using the "loading" button in the header. However, that button is _no longer available_ when there are no active tasks. That’s why I thought it made sense to show the close "X" only once all tasks have finished.

Maybe we could simplify things by letting the Popover disappear automatically when there are no remaining tasks (default behavior). I found that behavior a bit unexpected at first, but it might actually be less confusing overall.

Showing the "X" all the time feels excessive, in my opinion.

web/src/components/core/ProgressStatusMonitor.tsx
dgdavid: Related with previous comment https://github.com/agama-project/agama/pull/2909#discussion_r2559682780

This is avoiding the Popover to hide automatically when ongoing tasks finish.

dgdavid: Ah, right. `showClose` can benefit of idle variable too. I didn't realize it was not updated. In fact, the variable was added to avoid using tasks.length multiple times.

web/src/components/core/ProgressStatusMonitor.tsx
jreidinger: To be honest I probably prefer default behavior. I see no point to check empty task popup. Situation can be different if we display some post-mortem like if probing succeed, or some post installation message ( like rpm messages ). But if it will just disappear, I see no point to keep it open.

imobachgs: I agree, I would just close the popover if the progress finishes.

dgdavid: You're right, thanks. 

I was trying to avoid a different problem (a quick progress making the popover disappear almost immediately) but the solution for such a corner case shouldn't be this one. Reverting to default behavior.

jreidinger (approved):

PR #2910, opened 2025-11-25
Install callbacks

jreidinger:
## Problem

The last missing piece for libzypp callbacks are the ones used for package installation.

Previously in this series: #2901 

## Solution

Implement them. Also improve previous ones with `override` keyword in C++ to ensure we use proper method to avoid "overload" instead of "override".

## Testing

No testing as it is postponed to time when we have working install phase with new stack.

TODO: test this as part of this PBI:
- [ ] https://trello.com/c/HRqBeOi7/5450-support-for-instinstallurl

### DIFF
--- a/rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
+++ b/rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
@@ -54,8 +54,9 @@ struct DownloadProgressReceive : public zypp::callback::ReceiveReport<
 
   int last_reported;
   time_t last_reported_time;
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct DownloadProgressCallbacks *callbacks;
 
   DownloadProgressReceive() { callbacks = NULL; }
@@ -154,8 +155,9 @@ struct DownloadResolvableReport : public zypp::callback::ReceiveReport<
                                       zypp::repo::DownloadResolvableReport> {
   using Super = zypp::repo::DownloadResolvableReport;
 
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct DownloadResolvableCallbacks *callbacks;
 
   DownloadResolvableReport() { callbacks = NULL; }
@@ -270,8 +272,9 @@ static DownloadResolvableReport download_resolvable_receive;
 
 struct CommitPreloadReport
     : public zypp::callback::ReceiveReport<zypp::media::CommitPreloadReport> {
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct DownloadResolvableCallbacks *callbacks;
 
   CommitPreloadReport() { callbacks = NULL; }
@@ -326,8 +329,9 @@ static CommitPreloadReport commit_preload_report;
 
 struct KeyRingReport
     : public zypp::callback::ReceiveReport<zypp::KeyRingReport> {
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct SecurityCallbacks *callbacks;
 
   KeyRingReport() { callbacks = NULL; }
@@ -400,8 +404,9 @@ static KeyRingReport key_ring_report;
 
 struct DigestReceive
     : public zypp::callback::ReceiveReport<zypp::DigestReport> {
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct SecurityCallbacks *callbacks;
 
   DigestReceive() { callbacks = NULL; }
@@ -444,8 +449,9 @@ struct PatchScriptReport
     : public zypp::callback::ReceiveReport<zypp::target::PatchScriptReport> {
   using Super = zypp::target::PatchScriptReport;
 
-  // lifetime of pointer is quite short. Only during operation which takes
-  // callbacks as parameter.
+  // Lifetime: this is set and unset
+  // by a set_zypp_*_callbacks + unset_zypp_*_callbacks pair
+  // around the zypp call using it
   struct InstallCallbacks *callbacks;
 
   PatchScriptReport() { callbacks = NULL; }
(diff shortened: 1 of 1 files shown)


rust/agama-software/src/callbacks/install.rs
```diff
@@ -0,0 +1,109 @@
...
+            progress,
+            questions,
+        }
+    }
+}
+
+impl install::Callback for Install {
+    fn package_start(&self, package_name: String) {
+        tracing::info!("Installing package {}", package_name);
+        // TODO: report progress
```
imobachgs, line 33: ?

rust/agama-software/src/callbacks/security.rs
```diff
@@ -36,8 +33,7 @@ impl security::Callback for Security {
         let question = QuestionSpec::new(&text, "software.unsigned_file")
             .with_yes_no_actions()
             .with_data(&[("filename", file.as_str())]);
-        let result = Handle::current()
-            .block_on(async move { ask_question(&self.questions, question).await });
+        let result = ask_software_question(&self.questions, question);
```
imobachgs, line 36: :+1: that's what I meant with [this comment](https://github.com/agama-project/agama/pull/2901/files#r2555314334).

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -390,39 +389,158 @@ struct KeyRingReport
...
+    case zypp::target::rpm::InstallResolvableReport::Error::IO:
+      return ZyppInstallPackageError::PI_IO;
+    case zypp::target::rpm::InstallResolvableReport::Error::INVALID:
+      return ZyppInstallPackageError::PI_INVALID;
+    }
+    return ZyppInstallPackageError::PI_NO_ERROR;
+  }
+
+  inline zypp::target::rpm::InstallResolvableReport::Action
+  convert_action(PROBLEM_RESPONSE response) {
```
mvidner: Naming: I would expect `convert_response` since that is the argument type.

(Actually I would like to see Rust-like function names, like `to_foo`, `as_foo`  for standalone functions or `from_foo`  for methods)

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
```diff
@@ -252,6 +255,71 @@ struct SecurityCallbacks {
...
+ * @param description A description of the problem.
+ * @param user_data User-defined data.
+ * @return A PROBLEM_RESPONSE value indicating how to proceed.
+ */
+typedef enum PROBLEM_RESPONSE (*ZyppInstallPackageProblemCallback)(
+    const char *package_name, enum ZyppInstallPackageError error,
+    const char *description, void *user_data);
+/**
+ * @brief Callback invoked when the installation of a package finishes.
+ * @note We could add a finish message after package install, but YaST does not.
```
mvidner, line 293: You mean, YaST does nothing special on this callback and just continues with the next package? If it works, then fine, we can keep doing nothing in Agama too

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -22,23 +22,23 @@ struct ProgressReceive : zypp::callback::ReceiveReport<zypp::ProgressReport> {
...
   }
 
-  bool progress(const zypp::ProgressData &task) {
+  bool progress(const zypp::ProgressData &task) override {
     if (callback != NULL) {
       ProgressData data = {task.reportValue(), task.name().c_str()};
       return callback(data, user_data);
     } else {
-      return zypp::ProgressReport::progress(task);
+      return true;
```
mvidner: In the new callbacks (`PatchScriptReport`, `InstallResolvableReport`)  we do call the base class method but here we avoid it. Why?

rust/agama-software/src/callbacks/security.rs
jreidinger: I forgot to commit it in previous PR. It was already there :)

rust/zypp-agama/zypp-agama-sys/c-layer/include/callbacks.h
jreidinger: It is more tricky as YaST for backward compatibility reason use identical callback for Problem and Finish, so it do a bit of magic there. But part specific to finish is really quite trivial logging.

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
jreidinger: yeah, using such method names can be really helpful. So I will change it

rust/agama-software/src/callbacks/install.rs
jreidinger: I am still not sure how to properly report progress when there is two callbacks struct where the first one basically mainly download packages and here it install packages. It also need a bit of additional support from libzypp part as yast precompute for progress amount of packages to install, so it can create nicer sub-progress.

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
jreidinger: right, it should also call the original one which will just call that true. Good catch, I overlook this change.

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -390,39 +389,158 @@ struct KeyRingReport
...
+                                     found.c_str(),
+                                     callbacks->checksum_wrong_data);
   }
 };
 
 static DigestReceive digest_receive;
 
+struct PatchScriptReport
+    : public zypp::callback::ReceiveReport<zypp::target::PatchScriptReport> {
+  struct InstallCallbacks *callbacks;
```
mvidner, line 455: Oh, I've missed this: we need at least a lifetime comment, better a smart pointer that provides some safety.

jreidinger: well, smart pointer won't help you when rust deallocate it. Comment would be OK. Basically lifetime is for commit phase where we set and unset it ( it is same for all callbacks. I want to have it living only for limited time needed for operation ).

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
```diff
@@ -209,7 +209,7 @@ struct DownloadResolvableReport : public zypp::callback::ReceiveReport<
   }
 
   inline DownloadResolvableError
-  from_dre_error(zypp::repo::DownloadResolvableReport::Error error) {
+  from_libzypp_error(zypp::repo::DownloadResolvableReport::Error error) {
```
mvidner: I feel like I'm nitpicking, but:

IMHO something named `from_foo` must be a constructor, but here we return `DownloadResolvableError` in a method of `DownloadResolvableReport`.

So I would use `into_error` or `into_dre_error` here. What do you think?

jreidinger: it is fine for me. Probably I will use simple `into_error` because I like how they are consistent across different callbacks

rust/zypp-agama/zypp-agama-sys/c-layer/callbacks.cxx
mvidner: > // lifetime of pointer is quite short. Only during operation which takes
> // callbacks as parameter.

"Lifetime of pointer is quite short" is a nice philosophical take 🤣
but I wanted something more specific:
Why is `PatchScriptReport::callbacks` valid at all times? It is trivial to see that `PatchScriptReport` has a static instance so the danger of the pointer outliving its target is real.

IIUC, the answer is: `callbacks` is only assigned via `set_callbacks`, which is in turn called only from `{set,unset}_zypp_foo_callbacks`, in this fashion which makes sure that foo_callbacks are only borrowed by FooReport for the duration of the method.

```c++
some_zypp_method(foo_callbacks) {
  set_zypp_foo_callbacks(foo_callbacks);
  zypp::some_method_using_the_callbacks();
  unset_zypp_foo_callbacks();
}
```

As you can see I had to do quite some explaining. It would be nice to design the API in such a way where it would be more obvious and ideally checked by the C++ compiler.

PR #2912, opened 2025-11-26
Add support for files/scripts

imobachgs:
This PR adds support for scripts/files to the new HTTP API. The support for files and classes is included in the new `agama-files` package which reuses the old internal logic. For that reason. We can improve the internals later.

Additionally, it introduces other related changes:

- Extract `agama_lib::utils::transfer` to a separate package (`agama-transfer`).
- Drop the old scripts and files HTTP APIs.
- Add a test_utils module for agama-software.

## Follow-up

Fix a reference to `/mnt` in the `ScriptsRunner`. I decided to do this in a separate PR and take the chance to add progress and error reporting.

### DIFF
--- a/rust/agama-files/src/service.rs
+++ b/rust/agama-files/src/service.rs
@@ -169,7 +169,6 @@ impl Actor for Service {
 impl MessageHandler<message::SetConfig> for Service {
     async fn handle(&mut self, message: message::SetConfig) -> Result<(), Error> {
         self.scripts.clear()?;
-        self.files.clear();
 
         let config = message.config.unwrap_or_default();
 


rust/agama-autoinstall/src/scripts.rs
```diff
@@ -60,6 +62,9 @@ impl ScriptsRunner {
     /// It downloads the script from the given URL to the runner directory.
     /// It saves the stdout, stderr and exit code to separate files.
     ///
+    /// If the script cannot be execute due to the error "Text file busy (os error 26)",
+    /// wait for 50 milliseconds and try again (up to 5 times).
```
jreidinger: well, as exact sleep waiting is internal details of run_with_retry and also retry count. I would probably not specify here and just mention that it use run_with_retry where it will be documented.

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -69,9 +74,9 @@ impl ScriptsRunner {
         let path = self.path.join(&file_name);
         self.save_script(url, &path).await?;
 
-        let output = std::process::Command::new(&path).output()?;
+        let command = tokio::process::Command::new(&path);
```
jreidinger, line 76: should we change all std::Commands to tokio::Command due to async nature?

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -96,13 +101,14 @@ impl ScriptsRunner {
     }
 
     async fn save_script(&self, url: &str, path: &PathBuf) -> anyhow::Result<()> {
-        let mut file = Self::create_file(&path, 0o700)?;
+        let mut file = Self::create_file(&path, 0o755)?;
```
jreidinger: why changing permissions? I do not see any note in commit and as script can contain sensitive information like some credentials I see this a bit risky.

rust/agama-autoinstall/src/scripts.rs
```diff
@@ -96,13 +101,14 @@ impl ScriptsRunner {
...
         while let Err(error) = Transfer::get(url, &mut file, self.insecure) {
-            eprintln!("Could not load configuration from {url}: {error}");
+            eprintln!("Could not load the script from {url}: {error}");
             if !self.should_retry(&url, &error.to_string()).await? {
                 return Err(anyhow!(error));
             }
         }
+        file.sync_all()?;
```
jreidinger, line 110: :+1:

rust/agama-files/src/service.rs
```diff
@@ -0,0 +1,202 @@
...
+    type Error = Error;
+}
+
+#[async_trait]
+impl MessageHandler<message::SetConfig> for Service {
+    async fn handle(&mut self, message: message::SetConfig) -> Result<(), Error> {
+        self.scripts.clear()?;
+        self.files.clear();
+
+        let config = message.config.unwrap_or_default();
```
jreidinger, line 173: So if you send None to this service it result in having files set to None and scripts will be identical? This behavior looks confusing

rust/agama-autoinstall/src/scripts.rs
imobachgs: Argh, sorry, it is a left-over while debugging. I will fix it.

rust/agama-autoinstall/src/scripts.rs
imobachgs: Yes, definitely. I will check them in a separate PR.

rust/agama-files/src/service.rs
imobachgs: If you send `None`, it will clear files and scripts. Take into account that the merging is going in the manager service. Or am I missing something?

The only problem is that, due to the behavior of our merge logic, it will duplicate scripts and files (if any is given). But we will fix that in another place.

jreidinger: ah, it calls clear. you are right....btw does it make sense to call clear on files when we assign it later?

jreidinger (approved):

rust/agama-files/src/service.rs
imobachgs: Probably no. I will remove the call.

PR #2913, opened 2025-11-26
Fix buttons to reset storage configuration to defaults

ancorgs:
## Problem

There are several buttons in the web UI to reset the Agama configuration to the default values. But none of them work.

## Solution

Fix the hook to reset the configuration.

Bonus: remove a wrong comment in the backend (not really part of the fix, but was confusing during debugging).


joseivanlopez (approved):
LGTM

PR #2914, opened 2025-11-27
Drop lists from model

joseivanlopez:
Remove list and listIndex from the extended model. The components receives the collection and index by props.

This is the first step in order to get rid of the extended storage model.

Bonus: fix the result (new and resize labels).

### DIFF
--- a/web/src/components/storage/DriveEditor.tsx
+++ b/web/src/components/storage/DriveEditor.tsx
@@ -99,6 +99,13 @@ export type DriveEditorProps = { index: number };
 export default function DriveEditor({ index }: DriveEditorProps) {
   const driveModel = useDrive(index);
   const drive = useDevice(driveModel.name);
+
+  /**
+   * @fixme Make DriveEditor to work when the device is not found (e.g., after disabling
+   * a iSCSI device).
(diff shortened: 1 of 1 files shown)


web/src/components/storage/ConfigEditor.tsx
```diff
@@ -89,22 +72,12 @@ export default function ConfigEditor() {
...
-        {mdRaids.map((raid, i) => {
-          const device = devices.find((d) => d.name === raid.name);
-
-          return <MdRaidEditor key={`md-${i}`} raid={raid} raidDevice={device} />;
-        })}
-        {drives.map((drive, i) => {
-          const device = devices.find((d) => d.name === drive.name);
-
-          /**
-           * @fixme Make DriveEditor to work when the device is not found (e.g., after disabling
```
ancorgs, line 101: I think this problem remains. So the fixme should have been moved to `DriveEditor`. Isn't it?

joseivanlopez: Good point!

ancorgs (approved):
LGTM now

PR #2915, opened 2025-11-27
Registration crate

jreidinger:
## Problem

With the drop of ruby software service we will also loose suse connect ruby bindings.

## Solution

Write crate that acts like suseconnect bindings. We also have to write headers ourself from suseconnect-ng sources. This can be maybe in future send to their git.

Crate implementation is done for now as it export all methods that ruby agama needed, but it can be extended for future if need arise.

## Testing

- *Tested manually*

To ensure correctness I tested using examples. How to use it:

```
cd rust/suseconnect-agama
cargo test
../target/debug/examples/announce <regcode>
```

Feel free to test both correct and invalid SLES16 registration code.

Another example is for listing addons. I hardcoded sles 16 there and it needs to be run on already registered system.

### DIFF
--- a/web/package/agama-web-ui.spec
+++ b/web/package/agama-web-ui.spec
@@ -40,6 +40,9 @@ Agama web UI for the experimental Agama installer.
 %autosetup -p1 -n agama
 rm -f package-lock.json
 local-npm-registry %{_sourcedir} install --with=dev --legacy-peer-deps || ( find ~/.npm/_logs -name '*-debug.log' -print0 | xargs -0 cat; false)
+# temporary remove tests as its types are broken now
+find src -name *.test.tsx -delete
+rm src/mocks/api.ts
 
 %build
 NODE_ENV="production" npm run build
--- a/web/src/api.ts
+++ b/web/src/api.ts
@@ -21,7 +21,7 @@
  */
 
 import { get, patch, post, put } from "~/http";
-import type { ConfigModel } from "~/model/storage";
+import type { ConfigModel } from "~/model/storage/config-model";
 import type { Config } from "~/model/config";
 import type { Issue } from "~/model/issue";
 import type { Proposal } from "~/model/proposal";
--- a/web/src/components/core/InstallerOptions.tsx
+++ b/web/src/components/core/InstallerOptions.tsx
@@ -555,7 +555,7 @@ export default function InstallerOptions({
     l10n: { locales },
   } = useSystem();
   const { language, keymap, changeLanguage, changeKeymap } = useInstallerL10n();
-  const { state } = useStatus();
+  const { stage } = useStatus();
   const selectedProduct = useProduct();
   const initialFormState = {
     language,
@@ -572,7 +572,7 @@ export default function InstallerOptions({
   // Skip rendering if any of the following conditions are met
   const skip =
     (variant === "keyboard" && !localConnection()) ||
-    state === "installing" ||
+    stage === "installing" ||
     // FIXME: below condition could be a problem for a question appearing while
     // product progress
     [ROOT.login, ROOT.installationProgress, ROOT.installationFinished, PRODUCT.progress].includes(


rust/suseconnect-agama/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+rust-version.workspace = true
+edition.workspace = true
+
+[lib]
+name = "suseconnect_agama"
+path = "src/lib.rs"
+
+
+[[example]]
+name = "announce"
```
imobachgs: The examples are auto-discovered, so I do not think it is needed.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
+use std::{
+    ffi::{CString, IntoStringError},
+    fmt::Display,
+};
+
+use serde_json::{json, Value};
+
+// Safety requirements: inherited from https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr
+// expects that rust get control of string pointer
+// note: it is different then libzypp bindings where string is still owned by libzypp
```
imobachgs:
```suggestion
// note: it is different than libzypp bindings where string is still owned by libzypp
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
+use std::{
+    ffi::{CString, IntoStringError},
+    fmt::Display,
+};
+
+use serde_json::{json, Value};
+
+// Safety requirements: inherited from https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr
+// expects that rust get control of string pointer
```
imobachgs:
```suggestion
// expects that rust gets control of string pointer
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+// note: it is different then libzypp bindings where string is still owned by libzypp
+pub(crate) unsafe fn string_from_ptr(c_ptr: *mut i8) -> Result<String, IntoStringError> {
+    let c_str = CString::from_raw(c_ptr);
+    c_str.into_string()
+}
+
+/// parameters for SUSE Connect calls.
+///
+/// Based on https://github.com/SUSE/connect-ng/blob/main/internal/connect/config.go#L45
```
imobachgs: I wonder if we should use a permalink, because this one might get outdated at some point (yes, I know, permalinks are long and ugly). It would be https://github.com/SUSE/connect-ng/blob/5a200487c50d9c955708b34d0d35ebef47dc5a3e/internal/connect/config.go#L45.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    pub arch: String,
+    /// The product identifier (e.g., "SLES").
+    pub identifier: String,
+    /// The product version (e.g., "15.4").
+    pub version: String,
+}
+
+/// Represents product and also extensions info from SCC.
+/// list of attributes is just selection which agama uses.
+///
```
imobachgs: np: I would remove this line.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+}
+
+/// parameters for SUSE Connect calls.
+///
+/// Based on https://github.com/SUSE/connect-ng/blob/main/internal/connect/config.go#L45
+#[derive(serde::Serialize, Default, Debug, Clone)]
+pub struct ConnectParams {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub language: Option<String>,
+    // TODO: maybe use url instead of string?
```
imobachgs: It would be nice to make sure that it follows the right format. However, I am not sure about how common is to use those types in bindings.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("{message}")]
+    ApiError {
+        message: String,
+        // HTTP error code from API
+        code: u64,
+    },
+    #[error("Unknown error:{0}")]
```
imobachgs:
```suggestion
    #[error("Unknown error: {0}")]
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    UknownError(String),
+    #[error("Unexpected response from suse connect: {0}")]
```
imobachgs:
```suggestion
    #[error("Unexpected response from SUSEConnect: {0}")]
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("{message}")]
+    ApiError {
+        message: String,
+        // HTTP error code from API
+        code: u64,
+    },
+    #[error("Unknown error:{0}")]
+    UknownError(String),
```
imobachgs:
```suggestion
    Unknown(String),
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    #[error("Unexpected response from suse connect: {0}")]
+    UnexpectedResponse(String),
+    #[error(transparent)]
+    JsonParseError(#[from] serde_json::Error),
```
imobachgs: A few comment about naming:

* We need to reach an agreement whether to add the `Error` suffix to variants (not only on this PR). We have the [ErrorKind](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html) example, but I guess there are counter examples.
* Below you have a `JSONError`, so I would be consistent here: `JSONParseError`.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    #[error("Malformed SCC credentials file: {0}")]
+    MalformedSccCredentialsFile(String),
+    #[error("Missing SCC credentials file: {0}")]
+    MissingCredentialsFile(String),
+    #[error("JSON error: {0}")]
+    JSONError(String),
+    #[error("Network error: {0}")]
+    NetError(String),
```
imobachgs: I would use `Network` or `NetworkError`. The problem with the suffix is that it looks redundant: `Error::NetworkError`.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    SSLError {
+        message: String,
+        code: SSLErrorCode,
+        current_certificate: String,
+    },
+    #[error(transparent)]
+    UTF8Error(#[from] std::ffi::IntoStringError),
+}
+
+/// checks response from suseconnect for errors
```
imobachgs:
```suggestion
/// checks response from SUSEConnect for errors
```

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+pub struct Credentials {
+    pub login: String,
+    pub password: String,
+}
+
+impl TryFrom<Value> for Credentials {
+    type Error = Error;
+
+    fn try_from(value: Value) -> Result<Self, Self::Error> {
+        let Some(credentials) = value.get("credentials").and_then(|c| c.as_array()) else {
```
imobachgs: I think you can do (just a suggestion):


```suggestion
        let Some(credentials) = value.get("credentials").and_then(Value::as_array) else {
```

You can do the same with others `as_*` (e.g., `as_str`).

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+/// for subsequent [Self::create_credentials_file] call.
+///
+/// # Errors
+///
+/// Returns an `Err` of type [Error] if the announcement fails due to network issues,
+/// API errors, or problems parsing the response.
+pub fn announce_system(params: ConnectParams, target_distro: &str) -> Result<Credentials, Error> {
+    let result_s = unsafe {
+        let param_json = json!(params).to_string();
+        let params_c_ptr = CString::new(param_json).unwrap().into_raw();
```
imobachgs: Should we document why calling `unwrap` is OK here? Just forget about it if it is the common approach when writing bindings.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+    response.try_into()
+}
+
+/// Activates a product with SUSE Customer Center.
+///
+/// # Arguments
+///
+/// * `product` - The [Product] to activate.
+/// * `params` - Parameters [ConnectParams] for the connection.
+/// * `email` - The email address to associate with the activation. Can be empty.
```
imobachgs, line 363: Why is not the e-mail part of the `params` struct?

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+///
+/// # Errors
+/// Returns an `Err` of type `Error` if writing the configuration fails.
+pub fn write_config(params: ConnectParams) -> Result<(), Error> {
+    let result_s = unsafe {
+        let param_json = json!(params).to_string();
+        let params_c_ptr = CString::new(param_json).unwrap().into_raw();
+
+        let result_ptr = suseconnect_agama_sys::write_config(params_c_ptr);
+        let _ = CString::from_raw(params_c_ptr);
```
imobachgs, line 480: If we are ignoring the result, why is this call needed (just asking)?

rust/suseconnect-agama/src/lib.rs
jreidinger: well, I can convert it at the end to string, which is what suseconnect needs

rust/suseconnect-agama/src/lib.rs
jreidinger: well it is tricky question. Error can happen only if string contain '\0' inside it ( which basically kills C string as C uses \0 for end of string for basic char *). Here when we construct it from json I hope it cannot happen. But maybe we can introduce new error type for malformed strings passed to bindings.

rust/suseconnect-agama/src/lib.rs
jreidinger: it is already there...but do not ask me why this call have it as separate param. It is even part of connect config https://github.com/SUSE/connect-ng/blob/main/internal/connect/config.go#L55 .. so my guess is that it is some backward compatibility relict. But checking code it really uses this param and we also use it in ruby part, so for safety I would also use it here.

rust/suseconnect-agama/src/lib.rs
jreidinger: this is done at multiple parts. Basically here the issue is that SUSEConnect headers use "char *" and not "const char*" like we used in libzypp. Result is that we need to transform CString `into_raw` and basically give away memory ownership ( for const char* we do not need to do it )....but SUSEConnect won't clear passed string, so we need to take back ownership after the call to ensure that CString will free the memory when it gets out of scope. And that is exactly what happens there.
If you are interested here it is explained https://doc.rust-lang.org/stable/std/ffi/struct.CString.html#method.from_raw

rust/suseconnect-agama/src/lib.rs
jreidinger: I agree that suffix looks unnecessary. So lets try to unify on not having it. and I will change that JSON uppercase.

rust/suseconnect-agama/src/lib.rs
jreidinger: that is questionable as on ther hand we are mainly interested in the latest code to see if it changes in a way that can affect us.

rust/suseconnect-agama/src/lib.rs
jreidinger:
```suggestion
```

rust/suseconnect-agama/src/lib.rs
jreidinger: It is a bit shorter, so why not.

rust/suseconnect-agama/src/lib.rs
lslezak: It's easy to get the latest version just by clicking in the GitHub page or by manually rewriting the commit hash in the URL by "master".

But the opposite way, getting the old original status at the time when wrote the code, is more difficult. So I'd also prefer using permalinks.

Moreover having a git commit allows to easily see a diff from master and check the changes.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,796 @@
...
+}
+
+/// parameters for SUSE Connect calls.
+///
+/// Based on https://github.com/SUSE/connect-ng/blob/main/internal/connect/config.go#L45
+#[derive(serde::Serialize, Default, Debug, Clone)]
+pub struct ConnectParams {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub language: Option<String>,
+    // TODO: maybe use url instead of string?
```
lslezak: As a I can see it now uses `url:Url` so the comment can be removed.

rust/suseconnect-agama/examples/print_addons.rs
```diff
@@ -0,0 +1,16 @@
+use suseconnect_agama::{show_product, ConnectParams, ProductSpecification};
+
+// Note: the example has to be run on registered SLES 16, otherwise it returns different errors
+pub fn main() {
+    tracing_subscriber::fmt::init();
+
+    let product_spec = ProductSpecification {
+        identifier: "SLES".to_string(),
+        version: "16.0".to_string(),
+        arch: "x86_64".to_string(),
```
lslezak, line 10: I know this is just a testing example, but how difficult would be to detect the real current architecture? We need it in the real code anyway...

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,796 @@
...
+        };
+
+        Ok(Self {
+            url: url.to_string(),
+            name: name.to_string(),
+        })
+    }
+}
+
+/// SSL Error codes returned from SUSEConnect.
```
lslezak, line 153: I guess a link to the SUSEconnect code would be useful here...

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,796 @@
...
+            extensions,
+        })
+    }
+}
+
+/// Represents a service returned from registration to be added to libzypp.
+#[derive(Debug, Clone)]
+pub struct Service {
+    /// The name of the service that can be used in libzypp.
+    pub name: String,
```
lslezak, line 130: Is that really name or the alias?

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,796 @@
...
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("{message}")]
+    SCCApi {
+        message: String,
+        // HTTP error code from API
+        code: u64,
```
lslezak: I think `u32` should be more than enough for HTTP error codes...

rust/suseconnect-agama/src/lib.rs
```diff
@@ -0,0 +1,796 @@
...
+///
+/// * `login` - The username for the credentials file.
+/// * `pwd` - The password for the credentials file.
+/// * `path` - The path where the credentials file will be created.
+pub fn create_credentials_file(login: &str, pwd: &str, path: &str) -> Result<(), Error> {
+    // unwrap should not happen we do not construct invalid strings
+    let login = CString::new(login)?.into_raw();
+    let pwd = CString::new(pwd)?.into_raw();
+    let path = CString::new(path)?.into_raw();
+    let empty = CString::new("")?.into_raw();
```
lslezak, line 421: Maybe rather use `unused_argument` name? As more important is "why" and not "what", as is obviously visible.

rust/suseconnect-agama/examples/print_addons.rs
jreidinger: Well, in original code we use rpm arch...sadly using `zypper system-architecture` returns different string, so it would probably need a bit of processing. I kind of hope that we can get rpm architecture from libzypp. ( which I do not want to use here )

rust/suseconnect-agama/src/lib.rs
jreidinger: probably alias, but they call it in suse connect "name". That is always a bit challenge for me in such bindings if I should change names or keep original API names.

rust/suseconnect-agama/src/lib.rs
jreidinger: well, even `u16` will be enough, but it is more related to json interger sizes. Maybe even more correct would be i64 as suseconnect also use signed integer https://github.com/SUSE/connect-ng/blob/5a200487c50d9c955708b34d0d35ebef47dc5a3e/pkg/connection/api_error.go#L12

rust/suseconnect-agama/src/lib.rs
jreidinger: Well, I was a bit scaried to send there anything else, so I replicate explicit empty string as in ruby bindings in connect - https://github.com/SUSE/connect-ng/blob/main/third_party/yast/lib/suse/connect/yast.rb#L169

rust/suseconnect-agama/src/lib.rs
jreidinger: something like this mapping? https://github.com/SUSE/connect-ng/blob/5a200487c50d9c955708b34d0d35ebef47dc5a3e/third_party/libsuseconnect/libsuseconnect.go#L292

rust/suseconnect-agama/src/lib.rs
lslezak: Ah, I see...

PR #2916, opened 2025-11-28
Report progress and errors when running scripts

imobachgs:
## Problem

The only way a user can check whether a user script failed is by checking the logs.

## Solution

Implement support to report errors (using the questions mechanism) and ask the user whether Agama should retry running the script. The question includes additional information that can be used (in the future) in the UI to present a better question:

```json
[
  {
    "id": 1,
    "text": "Running the script 'test.sh' failed. Do you want to try again?",
    "class": "scripts.retry",
    "field": {
      "type": "none"
    },
    "actions": [
      {
        "id": "Yes",
        "label": "Yes"
      },
      {
        "id": "No",
        "label": "No"
      }
    ],
    "defaultAction": "No",
    "data": {
      "stderr": "some error message\n",
      "name": "test.sh",
      "exit_code": "1"
    }
  }
]
```

Additionally, Agama now reports the progress of running the scripts.

### DIFF
--- a/rust/agama-files/src/runner.rs
+++ b/rust/agama-files/src/runner.rs
@@ -19,7 +19,8 @@
 // find current contact information at www.suse.com.
 
 use std::{
-    io::Write,
+    fs::File,
+    io::{self, BufReader, Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
     process::ExitStatus,
 };
@@ -147,6 +148,8 @@ impl ScriptsRunner {
     /// * `path`: script's path.
     /// * `chroot`: whether to run the script in a chroot.
     async fn run_command<P: AsRef<Path>>(&self, path: P, chroot: bool) -> Result<(), Error> {
+        const STDERR_SIZE: u64 = 512;
+
         let path = path.as_ref();
         let stdout_file = path.with_extension("stdout");
         let stderr_file = path.with_extension("stderr");
@@ -173,7 +176,7 @@ impl ScriptsRunner {
         }
 
         if !output.status.success() {
-            let stderr = std::fs::read_to_string(&path.with_extension("stderr"))?;
+            let stderr = Self::read_n_last_bytes(&stderr_file, STDERR_SIZE)?;
             return Err(Error::Script {
                 status: output.status,
                 stderr,
@@ -192,6 +195,19 @@ impl ScriptsRunner {
         let steps: Vec<_> = messages.iter().map(|s| s.as_ref()).collect();
         let progress_action = progress::message::StartWithSteps::new(Scope::Files, &steps);
         _ = self.progress.cast(progress_action);
+    }
+
+    /// Reads the last n bytes of the file and returns them as a string.
+    fn read_n_last_bytes(path: &Path, n_bytes: u64) -> io::Result<String> {
+        let mut file = File::open(path)?;
+        let file_size = file.metadata()?.len();
+        let offset = file_size.saturating_sub(n_bytes);
+        file.seek(SeekFrom::Start(offset))?;
+        let bytes_to_read = (file_size - offset) as usize;
+        let mut buffer = Vec::with_capacity(bytes_to_read);
+        _ = file.read_to_end(&mut buffer)?;
+        let string = String::from_utf8_lossy(&buffer);
+        Ok(string.into_owned())
     }
 }
 


rust/agama-files/src/runner.rs
```diff
@@ -0,0 +1,365 @@
...
+    progress,
+    question::{self, ask_question},
+};
+use tokio::process;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error("The script failed")]
```
jreidinger, line 41: should not we write at least stderr?

rust/agama-files/src/runner.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    ///  Asks the user whether it should try to run the script again.
+    async fn should_retry(&self, script: &Script, error: Error) -> bool {
+        let text = format!(
+            "Running the script '{}' failed. Do you want to try again?",
+            script.name()
+        );
+        let mut question = QuestionSpec::new(&text, "scripts.retry").with_yes_no_actions();
```
jreidinger, line 128: do we have that type of question already documented? If not, please write it down to some follow up, to update documentation when new api will be merged.

rust/agama-files/src/runner.rs
```diff
@@ -0,0 +1,365 @@
...
+                .unwrap_or("unknown".to_string());
+            question = question.with_data(&[
+                ("name", script.name()),
+                ("stdout", &stdout),
+                ("stderr", &stderr),
+                ("exit_status", &exit_status),
+            ]);
+        }
+
+        let answer = ask_question(&self.questions, question).await.unwrap();
```
jreidinger: unwrap detected. Are you really sure it cannot fail? If so, please at least document it above the method why you think so.

rust/agama-files/src/runner.rs
```diff
@@ -0,0 +1,365 @@
...
+        } else {
+            process::Command::new(path)
+        };
+
+        let output = run_with_retry(command)
+            .await
+            .inspect_err(|e| println!("Error executing the script: {e}"))?;
+
+        fs::write(path.with_extension("log"), output.stdout.clone())?;
+        fs::write(path.with_extension("err"), output.stderr.clone())?;
```
jreidinger: note: this can be problematic if script will print too much in stdout or stderr as it will eat memory.

rust/agama-files/src/runner.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Ancillary function to start the progress.
+    fn start_progress(&self, scripts: &[&Script]) {
+        let messages: Vec<_> = scripts
+            .iter()
+            .map(|s| format!("Running user script '{}'", s.name()))
```
jreidinger, line 193: small note about localization. If you change localization during progress..it won't switch localization when we have it working of course :)

rust/agama-files/src/runner.rs
imobachgs: I do not think it should be part of the message. It is included already in the error and it is up to the code displaying the error to decide.

rust/agama-files/src/runner.rs
imobachgs: Good point. No, it is not documented, but I [opened agama-project.github.io#134](https://github.com/agama-project/agama-project.github.io/pull/134).

Talking about this, I think we should drop the [questions.md](https://github.com/agama-project/agama/blob/master/doc/questions.md) from this repository.

rust/agama-files/src/runner.rs
imobachgs: You are right. Well, if it fails, I guess we are in a pretty bad situation. But we need to handle it better.

rust/agama-files/src/runner.rs
imobachgs: This behaviour comes from AutoYaST.

rust/agama-files/src/runner.rs
imobachgs: Yes, we need to go through the whole application when we have l10n/i18n working :-)

rust/agama-files/src/runner.rs
jreidinger: that maybe open question if we should document that it is only for 16.1 and newer, but it is topic for that documentation PR

rust/agama-files/src/runner.rs
jreidinger: and does autoyast also store whole output in memory or stream it via fs redirections to that file?

imobachgs: Oh, now I know what you meant. No, AutoYaST writes it directly to a file. So I have adapted the code to behave in the same way, but I think we need to do additional changes:

* If we decide to write the `stdout` to a file because it can be a big file, then we should not send its content as part of the question. So I a dropping the `stdout` key.
* I am not sure about `stderr` (I would expect it to be smaller, but there are no guarantees about it). By now I am still ending the whole content.

Perhaps we could send just part of the stdout/stderr content to the question.

rust/agama-utils/src/command.rs
```diff
@@ -35,10 +41,11 @@ const RETRY_INTERVAL: u64 = 50;
 pub async fn run_with_retry(mut command: Command) -> io::Result<Output> {
     let mut attempt = 0;
     loop {
-        match command.output().await {
-            Ok(output) => {
-                return Ok(output);
-            }
+        // Using the output() function in tokio::process::Command unconditionlly
+        // configures stdout/stderr to be pipes. That's why we use spawn() +
+        // wait_with_output().
```
jreidinger, line 46: :+1: for comment

jreidinger (approved):

PR #2918, opened 2025-11-28
Reorganize data model

joseivanlopez:
Reorganize the code to deal with the API and the APP data model:

* *openapi/*: contains auto-generated types. In the future the API (types and http methods) will be provided by a separate package and this directory will disappear.
* *model/*: contains the types and methods to work with the app data model. Most of the types are reexported from openapi.
* *hooks/model/*: contains hooks to deal with cached data of the app data model.

Components use the model methods and hooks to interact with the API.

### DIFF
--- a/web/src/components/storage/FixableConfigInfo.tsx
+++ b/web/src/components/storage/FixableConfigInfo.tsx
@@ -25,9 +25,7 @@ import { Alert, List, ListItem } from "@patternfly/react-core";
 import { n_ } from "~/i18n";
 import type { Issue } from "~/model/issue";
 
-type DescriptionProps = FixableConfigInfoProps;
-
-const Description = ({ issues }: DescriptionProps) => {
+const Description = ({ issues }: { issues: Issue[] }) => {
   return (
     <List isPlain>
       {issues.map((e, i) => (
--- a/web/src/model/system/storage.ts
+++ b/web/src/model/system/storage.ts
@@ -20,8 +20,6 @@
  * find current contact information at www.suse.com.
  */
 
-// import type {} from "~/api/openapi/system/storage";
-
 import { sift } from "radashi";
 import type { System, Device } from "~/openapi/system/storage";
 


web/src/components/core/ChangeProductOption.test.tsx
```diff
@@ -27,7 +27,7 @@ import { useSystem } from "~/hooks/api";
 import { PRODUCT as PATHS } from "~/routes/paths";
 import { Product } from "~/types/software";
 import ChangeProductOption from "./ChangeProductOption";
-import { System } from "~/api/system/network";
+import { System } from "~/model/system/network";
```
jreidinger, line 30: import type can be used here. ( I think it is in general in each `~/model` import can be used import type.

web/src/components/core/InstallationFinished.test.tsx
```diff
@@ -25,7 +25,7 @@ import React from "react";
 import { screen } from "@testing-library/react";
 import { plainRender } from "~/test-utils";
 import InstallationFinished from "./InstallationFinished";
-import { Encryption } from "~/api/config/storage";
+import { Encryption } from "~/model/config/storage";
```
jreidinger, line 28: Encryption is also type

web/src/components/core/InstallButton.test.tsx
```diff
@@ -25,7 +25,7 @@ import { screen, waitFor, within } from "@testing-library/react";
 import { installerRender, mockRoutes } from "~/test-utils";
 import { InstallButton } from "~/components/core";
 import { PRODUCT, ROOT } from "~/routes/paths";
-import { Issue, IssueSeverity, IssueSource } from "~/api/issue";
+import { Issue, IssueSeverity, IssueSource } from "~/model/issue";
```
jreidinger, line 28: all of those are types.

web/src/components/core/InstallerOptions.test.tsx
```diff
@@ -23,16 +23,16 @@
...
-import { useProduct } from "~/hooks/api/config";
+import { useSystem } from "~/hooks/model/system";
+import { useProduct } from "~/hooks/model/config";
 import { Product } from "~/types/software";
-import { Keymap, Locale } from "~/api/system/l10n";
-import { Progress, State } from "~/api/status";
-import { System } from "~/api/system/network";
+import { Keymap, Locale } from "~/model/system/l10n";
+import { Progress, State } from "~/model/status";
+import { System } from "~/model/system/network";
```
jreidinger, line 31: I would say that all of those are types ( not use hooks, but these with uppercase start )

web/src/components/core/InstallerOptions.tsx
```diff
@@ -52,11 +52,11 @@ import { localConnection } from "~/utils";
...
 import { PRODUCT, ROOT, L10N } from "~/routes/paths";
-import { useProduct } from "~/hooks/api/config";
-import { useSystem } from "~/hooks/api/system";
-import { useStatus } from "~/hooks/api/status";
+import { useProduct } from "~/hooks/model/config";
+import { useSystem } from "~/hooks/model/system";
+import { useStatus } from "~/hooks/model/status";
 import { patchConfig } from "~/api";
-import type { Keymap, Locale } from "~/api/system/l10n";
+import type { Keymap, Locale } from "~/model/system/l10n";
```
jreidinger, line 59: types here.

web/src/components/core/IssuesAlert.test.tsx
```diff
@@ -24,7 +24,7 @@ import React from "react";
 import { screen } from "@testing-library/react";
 import { installerRender } from "~/test-utils";
 import { IssuesAlert } from "~/components/core";
-import { Issue, IssueSeverity, IssueSource } from "~/api/issue";
+import { Issue, IssueSeverity, IssueSource } from "~/model/issue";
```
jreidinger, line 27: and here.

web/src/components/core/IssuesDrawer.test.tsx
```diff
@@ -24,7 +24,7 @@ import React from "react";
 import { screen, within } from "@testing-library/react";
 import { installerRender } from "~/test-utils";
 import { InstallationPhase } from "~/types/status";
-import { Issue, IssueSeverity, IssueSource } from "~/api/issue";
+import { Issue, IssueSeverity, IssueSource } from "~/model/issue";
```
jreidinger, line 27: here

web/src/components/l10n/KeyboardSelection.test.tsx
```diff
@@ -25,7 +25,7 @@ import KeyboardSelection from "./KeyboardSelection";
 import userEvent from "@testing-library/user-event";
 import { screen } from "@testing-library/react";
 import { mockNavigateFn, installerRender } from "~/test-utils";
-import { Keymap } from "~/api/system/l10n";
+import { Keymap } from "~/model/system/l10n";
```
jreidinger, line 28: type here

web/src/components/storage/FixableConfigInfo.tsx
```diff
@@ -23,23 +23,29 @@
...
+      {issues.map((e, i) => (
         <ListItem key={i}>{e.description}</ListItem>
       ))}
     </List>
   );
 };
 
+type FixableConfigInfoProps = {
+  issues: Issue[];
+};
```
jreidinger, line 40: hmm, ordering here looks a bit strange. Above is DescriptionProps that is just `FixableConfigInfoProps`. So alias is used before type definition....and to be honest. I am not sure I see reason why having two distinguished types, when it is identical? I even would call it like `IssuesList`

jreidinger: minimal thing I would expect is to have type defined as DescriptionProps and here have that type alias.

web/src/model/system/storage.ts
```diff
@@ -20,8 +20,10 @@
  * find current contact information at www.suse.com.
  */
 
+// import type {} from "~/api/openapi/system/storage";
```
jreidinger: dead code?

joseivanlopez:
The `import type` were already missing before these chages. All of them will be added in a next iteration.

jreidinger (approved):

PR #2923, opened 2025-12-01
feat(live): Added new boot option "live.net_config_tui=1"

lslezak:
## Problem

- Manual network configuration using the dracut boot options is not easy

## Solution

- Added new boot option `live.net_config_tui=1` to interactively configure network settings using the `nmtui` tool.
- Additional features:
  - Allow configuring the HTTP proxy
  - Display the network configuration summary
  - Added network test button which tries connecting to www.suse.com to check that the network setup is correct.

## Testing

- Tested manually

## Notes

- It turned out that configuring the Wifi network is not possible because the drivers are missing in the initrd. In theory after adding the Wifi drivers (and firmware if required) it should be possible to configure also the Wifi cards and use them for downloading the self-updates.

## Screenshots

The initial summary screen:
agama-net-config-summary-devices

After pressing "Edit" button the `nmtui` tool is started
agama-net-config-edit

Then the user is asked to optionally configure the HTTP proxy:
agama-net-config-proxy

Then the summary dialog is displayed again. After pressing the "Test connection" button a request to `https://www.suse.com` is sent and the result is displayed on the screen:
agama-net-config-test

When the "Continue" button is pressed in the summary the boot process continues as usually.

A proof that the configured network proxy is used in the self-update step later:
nmtui7

### DIFF
--- a/live/live-root/usr/bin/initrd-network-setup.sh
+++ b/live/live-root/usr/bin/initrd-network-setup.sh
@@ -31,6 +31,19 @@ while :; do
     fi
   done
 
+  # list all devices
+  summary=$(
+    cat <<EOF
+Network Devices
+---------------
+$(nmcli d)
+
+Details
+-------
+$summary
+EOF
+  )
+
   if [ -n "$http_proxy" ]; then
     # hide the password in the proxy URL
     proxy_label=$(echo "$http_proxy" | sed 's|://.*@|://*******@|')
@@ -82,8 +95,8 @@ while :; do
     nmtui
 
     proxy_address=$(dialog --clear --no-cancel --backtitle "$BACK_TITLE" --title "HTTP proxy address" \
-		  --stdout --inputbox "Enter the HTTP proxy URL (e.g. http://proxy.example.com:3128)\nor leave empty to not use any proxy." \
-			10 70 "$http_proxy")
+      --stdout --inputbox "Enter the HTTP proxy URL (e.g. http://proxy.example.com:3128)\nor leave empty to not use any proxy." \
+      10 70 "$http_proxy")
 
     exit_status=$?
     if [ $exit_status -eq 0 ]; then


lslezak:
> Would it be possible to implement this simlarly so that a user is just prompted for what we want and with pre-defined / prefilled content for the variables?

Um, that would require to implement such tool from scratch, that's not trivial. The point here is that I basically just wrapped the `nmtui` and `nmcli` tools into a systemd service. This change is trivial and already provides quite a lot of needed functionality.

The network settings from the boot command line, like you used above, are still applied. E.g. if you use `live.net_config_tui=1 nameserver=8.8.8.8` options then the nameserver value is used as a pre-configured value and you can still change it in the `nmtui` tool interactively. The only drawback is that the `nmtui` tool allows setting many options so it is more difficult to navigate there. But in general it does what you need.

Example with the `live.net_config_tui=1 nameserver=8.8.8.8` boot options:
agama-net-config-dns

Martin-Weiss:
I guess there are even some further requirements for the installation media - we might need proxy settings and proxy authentication as well as ssl certificate trust before downloading the squashfs image and/or autoinstall control files in case self-signed certificates are used and the repos / internet etc. are only reachable via ssl-terminating proxy with authentication... but that comes after initial interactive network config at least for for IPv4 (ip/mask/gateway/dns-server/dns zone - search..)

One other thing - not sure if this is a bit out of scope, here - ntp-server - we also need time-sync in air-gapped scenarios in cases where we can not control the time in bios / firmware and where we do not have DHCP providing NTP server info.. not sure if that NTP thing is possible in nmtui? (otherwise SSL certificate validation will fail and timestamps in logs will be wrong)

teclator (approved):
LGTM

lslezak:
> not sure if that NTP thing is possible in nmtui?

No, the `nmtui` can only configure network devices/connections. The network services like NTP are not handled there.

PR #2925, opened 2025-12-02
Add integration test for agama_software::zypp_server::ZyppServer

mvidner:
## Problem

- Want to test especially the security zypp callbacks (wrong package checksum, unknown GPG key...)
- There are no existing tests in agama_software to plug into

## Solution = Testing

It's Hack Week! Learn AI assisted coding. https://hackweek.opensuse.org/25/projects/ai-powered-unit-test-automation-for-agama

Added a test triggering an unknown GPG key callback (with a `software.import_gpg` Question)

Apart from the tests, there are two changes in the library part:
- fix: don't let ZyppServer::initialize_target_dir remove its contents
- fix: remove unwanted repos

## Screenshots

not yet

## Documentation

The test runs as part of existing `cargo test` invoked by CI, no new docs added

### DIFF
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -28,10 +28,8 @@ use agama_utils::{
     products::ProductSpec,
     progress, question,
 };
-use std::{
-    collections::HashMap,
-    path::{Path, PathBuf},
-};
+use camino::{Utf8Path, Utf8PathBuf};
+use std::collections::HashMap;
 
 use tokio::sync::{
     mpsc::{self, UnboundedSender},
@@ -104,14 +102,16 @@ pub enum SoftwareAction {
 /// Software service server.
 pub struct ZyppServer {
     receiver: mpsc::UnboundedReceiver<SoftwareAction>,
-    root_dir: PathBuf,
+    root_dir: Utf8PathBuf,
 }
 
 impl ZyppServer {
     /// Starts the software service loop and returns a client.
     ///
     /// The service runs on a separate thread and gets the client requests using a channel.
-    pub fn start<P: AsRef<Path>>(root_dir: P) -> ZyppServerResult<UnboundedSender<SoftwareAction>> {
+    pub fn start<P: AsRef<Utf8Path>>(
+        root_dir: P,
+    ) -> ZyppServerResult<UnboundedSender<SoftwareAction>> {
         let (sender, receiver) = mpsc::unbounded_channel();
 
         let server = Self {
@@ -528,11 +528,9 @@ impl ZyppServer {
         let target_dir = self.root_dir.as_path();
         std::fs::create_dir_all(target_dir).map_err(ZyppDispatchError::TargetCreationFailed)?;
 
-        // FIXME: use camino::Utf8PathBuf or String
-        let zypp =
-            zypp_agama::Zypp::init_target(target_dir.to_str().unwrap(), |text, step, total| {
-                tracing::info!("Initializing target: {} ({}/{})", text, step, total);
-            })?;
+        let zypp = zypp_agama::Zypp::init_target(target_dir.as_str(), |text, step, total| {
+            tracing::info!("Initializing target: {} ({}/{})", text, step, total);
+        })?;
 
         self.import_gpg_keys(&zypp);
         Ok(zypp)
--- a/rust/agama-software/tests/zypp_server.rs
+++ b/rust/agama-software/tests/zypp_server.rs
@@ -26,6 +26,7 @@ use agama_utils::api::question::{Answer, AnswerRule, Config};
 use agama_utils::api::Issue;
 use agama_utils::progress;
 use agama_utils::question;
+use camino::Utf8Path;
 use glob::glob;
 use std::fs;
 use std::path::Path;
@@ -62,8 +63,8 @@ async fn test_start_zypp_server() {
     let _ = tracing_subscriber::fmt::try_init();
 
     let root_dir =
-        Path::new(env!("CARGO_MANIFEST_DIR")).join("../zypp-agama/fixtures/zypp_repos_root");
-    cleanup_past_leftovers(&root_dir);
+        Utf8Path::new(env!("CARGO_MANIFEST_DIR")).join("../zypp-agama/fixtures/zypp_repos_root");
+    cleanup_past_leftovers(&root_dir.as_std_path());
 
     let client = ZyppServer::start(&root_dir).expect("starting zypp server failed");
 
@@ -103,10 +104,7 @@ async fn test_start_zypp_server() {
     let repo_s = StateRepository {
         name: "signed_repo".to_string(),
         alias: "signed_repo".to_string(),
-        url: root_dir
-            .join("usr/share/signed_repo")
-            .to_string_lossy()
-            .to_string(),
+        url: root_dir.join("usr/share/signed_repo").to_string(),
         enabled: true,
     };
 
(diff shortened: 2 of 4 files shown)


rust/agama-software/src/lib.rs
```diff
@@ -43,6 +43,6 @@ pub use model::{Model, ModelAdapter, Resolvable, ResolvableType};
 
 mod callbacks;
 pub mod message;
-mod zypp_server;
+pub mod zypp_server;
```
jreidinger, line 46: hmm, I do not like this change. Can it be at max `pub(crate)`? I do not want to expose it publicly.

rust/agama-software/src/zypp_server.rs
```diff
@@ -506,14 +510,14 @@ impl ZyppServer {
...
-        let target_dir = Path::new(TARGET_DIR);
+        let target_dir = self.root_dir.as_path();
         if target_dir.exists() {
             _ = std::fs::remove_dir_all(target_dir);
         }
 
         std::fs::create_dir_all(target_dir).map_err(ZyppDispatchError::TargetCreationFailed)?;
 
-        let zypp = zypp_agama::Zypp::init_target(TARGET_DIR, |text, step, total| {
+        let zypp = zypp_agama::Zypp::init_target(target_dir.to_str().unwrap(), |text, step, total| {
```
jreidinger: `unwrap` is for sure something I do not like :)

jreidinger: either change it in a way that to_str cannot fail ( preferred ) or use some ZyppError ( less preferred as I think convert of path to string should not fail ).

mvidner: Right, I will have to start with some policy file to feed the bots at the start of each session.
Something like the SOP (Standard Operating Procedure) from this excellent (and skeptic) article: https://checkeagle.com/checklists/njr/a-month-of-chat-oriented-programming/

Thinking of:
> Any use of `unwrap()` must be preceded by a `// unwrap OK: ... ...` comment explaining why the panic cannot happen

rust/agama-software/src/lib.rs
mvidner: `pub(crate)` does not work even though Gemini also suggests it. I may end up inlining the tests in the fashion of unit tests.

rust/agama-software/src/zypp_server.rs
mvidner: For the case of paths, I like the concept of https://docs.rs/camino/latest/camino/
> camino is an extension of the [std::path](https://doc.rust-lang.org/nightly/std/path/index.html) module that adds new [Utf8PathBuf](https://docs.rs/camino/latest/camino/struct.Utf8PathBuf.html) and [Utf8Path](https://docs.rs/camino/latest/camino/struct.Utf8Path.html) types.
> ...
> Instead, camino allows you to check that your paths are UTF-8 once, and then manipulate them as valid UTF-8 from there on, avoiding repeated lossy and confusing conversions.

rust/agama-software/src/lib.rs
jreidinger: another option can be something like `cfg[test]` together with `pub use` so it will be public only for tests

rust/agama-software/src/callbacks.rs
```diff
@@ -9,11 +9,22 @@ mod security;
...
+use tokio::runtime::Builder;
 
 fn ask_software_question(
     handler: &Handler<question::Service>,
     question: QuestionSpec,
 ) -> Result<Answer, AskError> {
-    Handle::current().block_on(async move { ask_question(handler, question).await })
+    let (tx, rx) = mpsc::channel();
+    let question_handler = handler.clone();
+    std::thread::spawn(move || {
```
mvidner, line 21: @jreidinger 
Hey look, now we have something that works. It does block during the test if there is nothing to answer the question.
A new tokio runtime that is free to block on the question.

rust/agama-software/src/callbacks/commit_download.rs
```diff
@@ -11,6 +11,8 @@ use crate::callbacks::ask_software_question;
 
 #[derive(Clone)]
 pub struct CommitDownload {
+    // FIXME: field is never read
```
jreidinger, line 14: yeap, we plan to use it when we decide how exactly we want to report download and install progress.

rust/agama-software/src/callbacks.rs
```diff
@@ -9,11 +9,22 @@ mod security;
...
+        let runtime = Builder::new_current_thread().enable_all().build().unwrap();
+        let result =
+            runtime.block_on(async move { ask_question(&question_handler, question).await });
+        // unwrap OK: the receiver should not be dropped before the thread finishes
+        tx.send(result).unwrap();
+    });
+    // unwrap OK: the sender should not be dropped before sending a value
+    rx.recv().unwrap()
```
jreidinger, line 30: I worry that this won't work as recv is async method https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv
I am curious why rust do not complain that you do not call async.

jreidinger: ah, it is std mpsc and not tokio one

rust/agama-software/src/callbacks.rs
jreidinger: well, spawning a new thread looks like a bit of overkill performance wise. I still kind of hope that there should be better way to simply block zypp thread, as zypp should be blocking till the answer is provided.

rust/agama-software/src/callbacks.rs
jreidinger: but maybe it is right solution in the end to not use tokio at all for zypp thread and have it fully synchronous. Question probably also for @imobachgs as I am not that familiar of all tokio details.
https://www.linkedin.com/pulse/stdmpscchannel-vs-tokiosyncmpscchannel-rust-amit-nadiger-w5frc

jreidinger: I kind of expect we will probably need to adapt this part which started zypp_server to be able to work properly with asking questions https://github.com/agama-project/agama/pull/2925/files#diff-a680e15e57e7855ed9cab1073cfe8b25c605da1854eebef1521a68f090fc6198R138

rust/agama-software/tests/zypp_server.rs
```diff
@@ -0,0 +1,140 @@
...
+    );
+    let issues = result.unwrap();
+    assert_eq!(issues.len(), 1);
+    assert_eq!(issues[0].class, "software.select_product");
+
+    let questions = question_handler
+        .call(question::message::Get)
+        .await
+        .expect("Failed to get questions");
+    assert!(questions.is_empty());
```
jreidinger, line 143: good that we will have test, so we can play a bit more with how we start service...ideally we should also test if we do not kill e.g. questions service ( so that waiting for zypp answer do not block creating another question meanwhile from other service.

rust/agama-software/src/lib.rs
mvidner: I'm not sure what you mean by this

rust/agama-software/src/callbacks.rs
imobachgs: I guess you both know details better than me. But yes, we could run the whole ZyppServer on a separate thread (I thought we were already doing that) and use a channel to communicate with that. However, if we need to use an standard `mpsc`, we might need to use `spawn_blocking` to wait for the answer. Otherwise, we could block the whole runtime.

jreidinger: yeap, that is part which  I am not sure. For now we use spawn_blocking on libzypp thread, but I am not sure if other part of channel do not block whole runtime of actor.

mvidner:
OK, @jreidinger took this PR as a base for #2939 so I close this as Mission Accomplished :smile:

PR #2928, opened 2025-12-03
web: added TranslatedString type

lslezak:
## Problem

- We already check that the React components for untranslated string literals, e.g. `<Button>Close</Button>` reports an error in the eslint check.
-  Unfortunately it cannot check the texts passed as a component attribute, e.g.  `<Button label="Close" />`

## Solution

- Use the [type-fest](https://www.npmjs.com/package/type-fest) library, it provides quite a lot of useful TypeScript types
- Inspired by [this blog post article](https://idmontie.github.io/blog/post/2023-08-19-localization)
- Define a new `TranslatedString` type which holds a translated string (a string which passed through the `_()` or `n_()` translation functions)
- Require this new type in some core components
- Passing a plain string where a TranslatedString is expected results in a type error reported by TypeScript.

## Testing

- Updated unit tests

Tested with this small code snippet:
```tsx
function foo(label: TranslatedString) {
  return <>{label}</>;
}
// this line fails the type check
foo("label");
```

The last line in the example fails with error "Argument of type 'string' is not assignable to parameter of type 'TranslatedString'.
  Type 'string' is not assignable to type 'Tag<"Translated", never>'."

## Notes

- The texts marked for translation in the unit tests are ignored when building the POT file. Those strings are not uploaded to Weblate.

## Performance impact

The input TypeScript source code
```tsx
function foo(label: TranslatedString) {
  return <>{label}</>;
}
function bar() {
  return foo(_("label"));
}
```

Is compiled into this JavaScript:

```js
function foo(label) {
    return (0, jsx_runtime_1.jsx)(jsx_runtime_1.Fragment, { children: label });
}
function bar() {
    return foo((0, i18n_1._)("label"));
}
```

As you can see there is zero overhead for the `TranslatedString` type, the check is completely done at the compile time without any performance effect at runtime.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Dec 16 08:32:09 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
+
+- Added a new TranslatedString type to require a translated string
+  in the code, this should lower the risk of missing translations
+  (gh#agama-project/agama#2928)
+
+-------------------------------------------------------------------
 Fri Nov 21 08:17:06 UTC 2025 - Ladislav Slezák <lslezak@suse.com>
 
 - Fixed missing translations (error cases for software repos, iSCSI, DASD; bsc#1250059)


web/package.json
```diff
@@ -82,6 +83,7 @@
     "ts-jest": "^29.4.1",
     "ts-loader": "^9.5.4",
     "tsconfig-paths-webpack-plugin": "^4.2.0",
+    "type-fest": "^5.3.0",
```
dgdavid, line 86: I've delayed adding this for some time due to ongoing changes in this area, but I fully support including it now. However, it would be helpful if we could encourage the team to check this package before creating new types, to see if any of the existing utilities or guards might be a good fit. This could save both development and maintenance time in the long run.

web/src/components/core/SplitButton.tsx
```diff
@@ -47,7 +47,7 @@ type SplitButtonBaseProps = React.PropsWithChildren<{
   /** Label for the link or button acting as a main action */
   label: React.ReactNode;
   /** Accessible label for the toggle button */
-  toggleAriaLabel?: MenuToggleProps["aria-label"];
+  toggleAriaLabel?: TranslatedString;
```
dgdavid, line 50: I'm wondering if this is compatible with the expected string for MenuToggle#aria-label. I believe it should be, but I haven’t had the chance to test it yet.

lslezak: I tested this code snippet:

```tsx
import { _, TranslatedString } from "./i18n";

const translated: TranslatedString = _("foo");
const notTranslated: string = "foo";

function fn(params: string) {
  console.log(params);
}

function fn2(params: TranslatedString) {
  console.log(params);
}

export function fn3() {
  // this is OK, passing a TranslatedString where a plain string is expected works just fine,
  // the compiler does not report any error
  fn(translated);

  // but this fails, passing a plain string where a TranslatedString is expected raises error
  // "Argument of type 'string' is not assignable to parameter of type 'TranslatedString'."
  fn2(notTranslated);
}
```

So it works as expected, `TranslatedString` can be used everywhere where a plain string is accepted, but the opposite direction is forbidden.

PR #2934, opened 2025-12-09
Drop storage extended model

joseivanlopez:
## Problem

The storage config model provided by the HTTP API is not directly used by the UI components. A new extended model is defined, which implies to have different types to the API types. This adds an extra layer, making the resulting code more complex without any clear advantage.

## Solution

Drop the extended config model. The components directly use the config model provided by the UI. The extra functionality provided by the extended config model is moved to methods of the app data model.

Note: the hooks under *hooks/storage* and the methods in *storage/* have to be moved too, but it will be done in a separate PR.


web/src/components/storage/BootSelection.tsx
```diff
@@ -37,12 +38,16 @@ import {
 import textStyles from "@patternfly/react-styles/css/utilities/Text/text";
 import { sprintf } from "sprintf-js";
 import { _ } from "~/i18n";
+import { configModelMethods } from "~/model/storage";
+import type { configModel } from "~/model/storage";
```
jreidinger: I have to say that I am a  bit lost about naming convention for types. I see here camelCase, but for others like Issues it start with Upper case. So what is convention?

web/src/components/storage/BootSelection.tsx
```diff
@@ -37,12 +38,16 @@ import {
...
 import type { storage } from "~/model/system";
-import type { Model } from "~/storage/model";
 
-const filteredCandidates = (candidates: storage.Device[], model: Model): storage.Device[] => {
+const filteredCandidates = (
+  candidates: storage.Device[],
+  configModel: configModel.Config,
```
jreidinger: this is a bit tricky to parse...we have variable `configModel` that has type `configModel.Config`...it looks a bit strange for me. I would probably just call it `config` instead of `configModel` especialy when the Type is config

jreidinger: and if I see correctly it is only usage of configModel type here...so is possible to just import Config part of configModel?

web/src/components/storage/DriveHeader.tsx
```diff
@@ -22,23 +22,28 @@
...
     return _("Format disk %s");
   }
 
-  const { isBoot, isTargetDevice: hasPv } = drive;
-  const isRoot = !!drive.getPartition("/");
-  const hasFs = !!drive.getMountPaths().length;
+  const isBoot = configModelMethods.isBootDevice(configModel, drive.name);
+  const hasPv = configModelMethods.isTargetDevice(configModel, drive.name);
+  const isRoot = !!partitionableModelMethods.findPartition(drive, "/");
+  const hasFs = !!partitionableModelMethods.usedMountPaths(drive).length;
```
jreidinger: :+1:

web/src/components/storage/PartitionPage.tsx
```diff
@@ -229,8 +230,7 @@ function useDefaultFilesystem(mountPoint: string): string {
 function useInitialPartitionConfig(): configModel.Partition | null {
   const { partitionId: mountPath } = useParams();
   const device = useDeviceModelFromParams();
-
-  return mountPath && device ? device.getPartition(mountPath) : null;
+  return mountPath && device ? partitionableModelMethods.findPartition(device, mountPath) : null;
```
jreidinger: just idea. Does it make sense to return null from Methods when either of it is null? It makes code more robust to move parameter checks to method then forcing each caller ensuring it.

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -196,42 +217,46 @@ const ChangeDeviceMenuItem = ({
...
-  const model = useModel();
+  const configModel = useConfigModel();
 
   /*
    * Pretty artificial logic used to decide whether the UI should display buttons to remove
    * some drives.
    */
-  const hasAdditionalDrives = (model: model.Model): boolean => {
-    const entries = model.drives.concat(model.mdRaids);
+  const hasAdditionalDrives = (configModel: configModel.Config): boolean => {
```
jreidinger: similar here with variable name. I would probably just call it `config`

web/src/components/storage/SearchedDeviceMenu.tsx
```diff
@@ -261,25 +286,28 @@ const RemoveEntryOption = ({ device, onClick }: RemoveEntryOptionProps): React.R
...
   return availableDevices.filter((availableDev) => {
     if (modelDevice.name === availableDev.name) return true;
 
-    const collection = isDrive(availableDev) ? model.drives : model.mdRaids;
+    const collection = isDrive(availableDev) ? configModel.drives : configModel.mdRaids;
     const device = collection.find((d) => d.name === availableDev.name);
     if (!device) return true;
 
-    return modelDevice.filesystem ? !device.isUsed : !device.filesystem;
+    return modelDevice.filesystem
```
jreidinger: well, after change ternary operator splits over multiple lines...in such case I prefer to use simply if condition for better readability.

web/src/model/storage/config-model.ts
```diff
@@ -20,4 +20,64 @@
...
+}
+
+export {
+  usedMountPaths,
+  bootDevice,
+  hasDefaultBoot,
+  isBootDevice,
+  isExplicitBootDevice,
+  isTargetDevice,
+  isUsedDevice,
```
jreidinger, line 80: to be honest this is quite lot of new code and I do not see any tests for it. So I am not that confident that everything works as expected. I would really like to see some at least basic tests of functionality.

web/src/model/storage/partitionable-model.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { sift } from "radashi";
+import { partitionModelMethods, volumeGroupModelMethods } from "~/model/storage";
+import type { configModel } from "~/model/storage";
+
+type Partitionable = configModel.Drive | configModel.MdRaid;
```
jreidinger: to be honest I find it a bit strange that main type used in this methods are only private to this model. I kind of expect that this Partitionable type will be public so someone using this model methods can define that it needs just device of given type.

web/src/model/storage/partitionable-model.ts
```diff
@@ -0,0 +1,77 @@
...
+function selectConfiguredExistingPartitions(device: Partitionable): configModel.Partition[] {
+  if (device.spacePolicy === "custom")
+    return device.partitions.filter(
+      (p) =>
+        !partitionModelMethods.isNew(p) &&
+        (partitionModelMethods.isUsed(p) || partitionModelMethods.isUsedBySpacePolicy(p)),
+    );
+
+  return device.partitions.filter(partitionModelMethods.isReused);
+}
```
jreidinger, line 69: Also I expect some documentation for those new methods as it is not always obvious what they do or about constrains they have.

web/src/model/storage/volume-group-model.ts
```diff
@@ -0,0 +1,48 @@
...
+
+function candidateTargetDevices(
+  configModel: configModel.Config,
+): (configModel.Drive | configModel.MdRaid)[] {
+  const drives = configModel.drives || [];
+  const mdRaids = configModel.mdRaids || [];
+  return [...drives, ...mdRaids];
+}
+
+function selectTargetDevices(
```
jreidinger: I already notice it during review of usage of methods, but I find it inconsistent with JS to call method "select" like in ruby, when it do inside "filter" so I kind of expect `filterTargetDevices`

web/src/components/storage/BootSelection.tsx
joseivanlopez: Good point. I have asked to gemini. Indeed, PascalCase should be used for grouping types, and camelCase for grouping functions. I will rename it as `ConfigModel`.

BTW, what about the functions? Do you think `configModelMethods` is a good name? Now we can use `configModel` if we want.

joseivanlopez: Perhaphs we could simplify the naming as:

~~~ts
const config: ConfigModel.Config = {...}

configModel.hasDefaultBoot(config);
~~~

* `ConfigModel` for grouping types.
* `configModel` for grouping functions
* `config` for instances of `ConfigModel.Config`

According to gemini, grouping functions is like having  an object instance with all those methods, so using `configModel` could make sense.

jreidinger: what about having different approach and add methods directly to instance of interface like written at https://stackoverflow.com/questions/69651011/is-there-a-way-to-add-methods-to-a-type-in-typescript ?
This way it can be 
```
const config = ConfigModel.Config = { .. }
config.hasDefaultBoot();
```

web/src/components/storage/BootSelection.tsx
joseivanlopez: I hope now it is clearer now because the group type is called `ConfigModel` instead of `configModel`.

web/src/components/storage/PartitionPage.tsx
joseivanlopez: I am not sure. `#findPartition` expects a device and a mountPath. I think the logic of having a device and mountPath does not belong to `#findPartition`.

web/src/model/storage/config-model.ts
joseivanlopez: We have to improve a lot the testing. All this was indirectly tested through the components. In a separate PR we plan: a) fix all existing tests, b) add tests for hooks and model.

web/src/model/storage/partitionable-model.ts
joseivanlopez: Right now it is not needed outside. We can publish it once we need the type.

web/src/model/storage/partitionable-model.ts
joseivanlopez: Yes, but this PR is basically about reorganizing code. Those methods were not documented. For sure we will do it, but we have a lot of things to do (like tests), so we need to prioritize. My plan is to polish documentation once we have everything organized and tested.

web/src/model/storage/volume-group-model.ts
joseivanlopez: I have renamed those methods as filter*, becaue find* usualy returns a single element.

web/src/components/storage/BootSelection.tsx
joseivanlopez: After discussing it with @imobachgs and @dgdavid, we think it does not pay off the effor of extending interfaces or creating classes. There is no a real need, so for now we would like to continue with the current approach. After the renamings, the usage is more clear and elegant:

~~~ts
import configModel from "~/model/storage/config-model";
import type { ConfigModel } from "~/model/storage/config-model";

const config = useConfigModel();
configModel.hasDefaultBoot(config);
~~~

jreidinger (approved):

PR #2936, opened 2025-12-10
Add hardware information to the API

imobachgs:
## Problem

The current API does not include any hardware information of the underlying system. It might be useful to present this information to the end user.

## Solution

Include some basic hardware information as part of the `/system` object.

```json
{
  "cpu": "AMD Ryzen 5 PRO 5650U with Radeon Graphics",
  "memory": 4294967296,
  "model": "QEMU pc-q35-9.2"
}
```

Although in the Rust ecosystem there are interesting libraries, like [hardware_query](https://docs.rs/hardware-query/latest/hardware_query/), we decided to leverage `lshw` because it is the same library we use when processing jsonnet profiles.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/rust/agama-utils/src/api/manager/system_info.rs
+++ b/rust/agama-utils/src/api/manager/system_info.rs
@@ -55,7 +55,7 @@ pub struct Product {
 pub struct HardwareInfo {
     /// CPU description.
     pub cpu: Option<String>,
-    /// Memory size.
+    /// Memory size (in bytes).
     pub memory: Option<u64>,
     /// Computer model.
     pub model: Option<String>,


rust/agama-manager/src/hardware.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        for children in &self.children {
+            children.search_by_class(class, results);
+        }
+    }
+}
+
+impl From<&HardwareNode> for HardwareInfo {
+    fn from(value: &HardwareNode) -> Self {
+        // let system_info = value.find_by_class("system");
```
dgdavid: leftover comment?

rust/agama-manager/src/service.rs
```diff
@@ -257,11 +273,11 @@ impl Service {
         Starter::new(questions, events, dbus)
     }
 
-    /// Set up the service by reading the registries and determining the default product.
+    /// Set up the service by reading the registries, the hardware info and determining the default product.
```
dgdavid, line 276: missing oxford comma?

rust/agama-manager/src/hardware.rs
```diff
@@ -0,0 +1,288 @@
...
+use serde::Deserialize;
+use serde_with::{formats::PreferMany, serde_as, OneOrMany};
+use std::{
+    path::{Path, PathBuf},
+    process::ExitStatus,
+};
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("lshw command failed")]
```
jreidinger: I would add there stderr of failure. It will help with debugging when things go wrong.

jreidinger: I mean to the message which will be printed

rust/agama-manager/src/hardware.rs
```diff
@@ -0,0 +1,288 @@
...
+    Command { status: ExitStatus, stderr: String },
+    #[error("Failed to parse lshw output")]
```
jreidinger: again, here I think will be useful to either include its output ( which would be sadly quite big ) or content of serde Error which I hope can bring a bit of light to it

rust/agama-manager/src/hardware.rs
```diff
@@ -0,0 +1,288 @@
...
+            .find_by_class("processor")
+            .first()
+            .and_then(|c| c.product.clone());
+
+        let memory = value.find_by_id("memory").and_then(|m| m.size);
+
+        let model = value.find_by_class("system").first().map(|s| {
+            format!(
+                "{} {}",
+                s.vendor.clone().unwrap_or_default(),
```
jreidinger: I am not sure if it is better to use empty string like here or explicit Unknown key so it is obvious that we do not know verdor or version.

rust/agama-utils/src/api/manager/system_info.rs
```diff
@@ -47,3 +49,14 @@ pub struct Product {
     /// License ID
     pub license: Option<String>,
 }
+
+/// Represents the hardware information of the underlying system.
+#[derive(Clone, Default, Debug, Serialize, utoipa::ToSchema)]
+pub struct HardwareInfo {
+    /// CPU description.
+    pub cpu: Option<String>,
+    /// Memory size.
```
jreidinger: I would write units here

rust/agama-manager/src/service.rs
imobachgs: Not using the Oxford comma at all. I think you and me are the only ones who use it :-)

dgdavid: Good!

Excuse me, I forgot that it was an already discussed topic. I guess old habits just kicked in!

:point_right:   https://documentation.suse.com/style/current/single-html/style-guide-db/index.html#sec-comma

rust/agama-utils/src/api/manager/system_info.rs
dgdavid: Better to transform to the units you prefer at client side. Or maybe you refer to the comment "Memory size, in bytes"

jreidinger: yeap, I am referring to comment

rust/agama-manager/src/hardware.rs
imobachgs: I have adapted the code to use `None` if the full string is empty.

rust/agama-utils/src/api/manager/system_info.rs
jreidinger: in general any usage of ix or ux should mention units unless it is very obvious.

imobachgs: Actually, I see that the output of the `lshw` is like this:

```json
{
  "id" : "memory",
  "class" : "memory",
  "claimed" : true,
  "description" : "Memoria de sistema",
  "physid" : "0",
  "units" : "bytes",
  "size" : 16106127360
},
```

So I guess we might need something slightly different (e.g., `Size(u64, String)`) or using a proper type.

jreidinger: probably yes, as on system with huge memory like few TB it can cause issue to even fit into type.

imobachgs: I have documented it by now, as it only uses bytes according to [lshw sources](https://github.com/lyonel/lshw/blob/209f8306e9500cf5ac32f01abb11ff6dcce76b0a/src/core/hw.cc#L1544).

PR #2938, opened 2025-12-11
Implement the "install" action

imobachgs:
## Problem

The `api-v2` branch does not implement an action to perform the installation yet.

## Solution

* Implement a first and incomplete version of the `Install` action.
* Move tracking of the installation state (now renamed to Stage) to the progress service.

## To do

- [x] Extend the progress service tests to include the stage.
- [x] Rename `Get` and `Set` to `GetProgress` and `SetProgress`.
- [x] Refuse to run some actions when the state is not `Configuring` (e.g., `SetConfig`, `RunAction`, etc.).

## Testing

- Tested manually

### DIFF
--- a/rust/agama-manager/src/service.rs
+++ b/rust/agama-manager/src/service.rs
@@ -33,6 +33,7 @@ use agama_utils::{
     progress, question,
 };
 use async_trait::async_trait;
+use gettextrs::gettext;
 use merge_struct::merge;
 use network::NetworkSystemClient;
 use serde_json::Value;
@@ -726,10 +727,10 @@ impl InstallAction {
         self.progress
             .call(progress::message::StartWithSteps::new(
                 Scope::Manager,
-                &[
-                    "Prepare the system",
-                    "Install software",
-                    "Configure the system",
+                vec![
+                    gettext("Prepare the system"),
+                    gettext("Install software"),
+                    gettext("Configure the system"),
                 ],
             ))
             .await?;
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -149,6 +149,7 @@ dependencies = [
  "agama-storage",
  "agama-utils",
  "async-trait",
+ "gettext-rs",
  "merge-struct",
  "serde",
  "serde_json",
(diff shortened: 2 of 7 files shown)


rust/agama-manager/src/service.rs
```diff
@@ -480,21 +458,26 @@ impl Service {
         }
         Ok(())
     }
+
+    async fn ensure_stage(&self, expected: Stage) -> Result<(), Error> {
```
mchf: NP: ```ensure_stage``` evokes in my mind that the method will switch to given stage. I'd suggest something like ```check_stage```

imobachgs: Yes, you are right. I changed it to `check_stage`.

rust/agama-manager/src/service.rs
```diff
@@ -679,7 +676,88 @@ impl MessageHandler<message::SolveStorageModel> for Service {
...
+            .await?;
+
+        //
+        // Preparation
+        //
+        self.progress
+            .call(progress::message::StartWithSteps::new(
+                Scope::Manager,
+                &[
+                    "Prepare the system",
```
mvidner: Are these user visible?
Translation infrastructure is still missing, but please mark these with gettextrs::gettext and we will make it work later.

rust/agama-manager/src/service.rs
```diff
@@ -679,7 +676,88 @@ impl MessageHandler<message::SolveStorageModel> for Service {
...
+    storage: Handler<storage::Service>,
+    files: Handler<files::Service>,
+    progress: Handler<progress::Service>,
+}
+
+impl InstallAction {
+    /// Runs the installation process on a separate Tokio task.
+    pub fn run(mut self) {
+        tokio::spawn(async move {
+            self.install().await.unwrap();
```
jreidinger: is unwrap good idea? It will basically kill that task. I hope for some better handling like at least logging it, or setting some FailedInstall Stage or anything else to indicate that installation failed ( the problem we have with current agama )

imobachgs: No, it is not a good idea, but I wanted to have this working to unblock other work on the installation process. I will log the error. Perhaps we could even have a new Stage `Error` so the UI knows that something went wrong.

rust/agama-manager/src/service.rs
imobachgs: However, I am not sure about when to do the real translation. Let's think on a theorical case: you are installing the system, or doing the initial probing. You get the messages in, let's say, English. If you change the language, the rest of the process will be emitted in English.

I am tempted of translating the strings later (at API level) but it is true that, at least, we need to mark the strings here. WDYT?

PR #2939, opened 2025-12-12
Software block on fix

jreidinger:
## Problem

Great test from @mvidner reveals that we panic when any question is asked in software callbacks.

## Solution

I add tests from https://github.com/agama-project/agama/pull/2925 but use different approach as using thread spawning every time question should be asked is very inefficient.

So let me here discuss a bit possible approaches:

1. one from @mvidner PR that spawn thread and do blocking job there. It is very inefficient.
2. use different blocking mechanism like from futures crate..so we block on tokio and then block question with futures. We can do it because we know that answer will come from different thread. On other hand, it adds additional dependency and feels a bit hackish as it escape tokio runtime.
3. synchronous approach. Basically have code in zyppServer thread synchronous and only very small parts that needs to be asynchronous call with block_on for minimal time. This way we ensure that there is no chance to get message when waiting for answer and also it will mark properly asynchronous communication with rest of agama.

For choosen approach 3 I tried few ways that does not work:

- LocalTask and local_spawn cause lockup of test where message is send, but receive do not receive it.
- spawn_blocking approach which result in too much glue around

The working one is spawning common tokio task as it is not that important where it runs. Important part is only that zypp thread waits till it gets answer.

## Testing

- test from mvidner is added


rust/agama-software/tests/zypp_server.rs
```diff
@@ -0,0 +1,151 @@
...
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, contact SUSE LLC.
+//
+// To contact SUSE LLC about this file by physical or electronic mail, you may
+// find current contact information at www.suse.com.
+
+use agama_software::model::state::{Repository as StateRepository, SoftwareState};
+use agama_software::zypp_server::{SoftwareAction, ZyppServer, ZyppServerResult};
+use agama_utils::actor;
```
imobachgs: np: I know, we are not consistent at all, but we are moving to `use agama_utils::{actor, ...}` instead of having multiple lines.

rust/agama-software/tests/zypp_server.rs
```diff
@@ -0,0 +1,151 @@
...
+use agama_utils::question;
+use camino::Utf8Path;
+use glob::glob;
+use std::fs;
+use std::path::Path;
+use std::result::Result;
+use tokio::sync::{broadcast, oneshot};
+use tracing_subscriber;
+
+fn cleanup_past_leftovers(root_dir: &Path) {
```
imobachgs, line 40: In this case it is good enough because we have a single test. But if you plan to add more tests, you might take advantage of https://docs.rs/test-context/latest/test_context/. If it is not the case, just ignore this comment :-)

rust/agama-software/tests/zypp_server.rs
```diff
@@ -0,0 +1,151 @@
...
+        fs::remove_dir_all(rpmdb_dir).expect("removing RPM data failed");
+    }
+}
+
+#[tokio::test]
+async fn test_start_zypp_server() {
+    let _ = tracing_subscriber::fmt::try_init();
+
+    let root_dir =
+        Utf8Path::new(env!("CARGO_MANIFEST_DIR")).join("../zypp-agama/fixtures/zypp_repos_root");
```
imobachgs, line 69: I did not know about `camino`. Interesting...

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
     }
 
     /// Runs the server dispatching the actions received through the input channel.
-    async fn run(mut self) -> Result<(), ZyppDispatchError> {
+    fn run(mut self) -> Result<(), ZyppDispatchError> {
         let zypp = self.initialize_target_dir()?;
 
         loop {
-            let action = self.receiver.recv().await;
+            // well, the code here can look a bit tricky, but what happens in reality here is that we need synchronized code
```
imobachgs: I would make the comment more straightforward :-) (removing the "well, the code here can look a bit tricky..." part).

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
+            // and only for small async interface run it in blocking way. Receiver handling is done to explicit passing
+            // of ownership as receiver do not implement Copy.
+            let rt = tokio::runtime::Builder::new_current_thread()
```
imobachgs, line 138: I would most probably clarify why we cannot just use `spawn::task` directly.

rust/agama-software/src/callbacks.rs
```diff
@@ -9,11 +9,16 @@ mod security;
...
 
 fn ask_software_question(
     handler: &Handler<question::Service>,
     question: QuestionSpec,
 ) -> Result<Answer, AskError> {
-    Handle::current().block_on(async move { ask_question(handler, question).await })
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
```
mvidner, line 21: Do you want to copy my "unwrap OK" comment from below?

rust/agama-software/src/callbacks.rs
```diff
@@ -9,11 +9,16 @@ mod security;
 pub use security::Security;
 mod install;
 pub use install::Install;
-use tokio::runtime::Handle;
 
 fn ask_software_question(
     handler: &Handler<question::Service>,
     question: QuestionSpec,
 ) -> Result<Answer, AskError> {
-    Handle::current().block_on(async move { ask_question(handler, question).await })
```
mvidner, line 18: JFYI for reference and comparison, here is the corresponding bit from my attempt:
```rs
    let (tx, rx) = mpsc::channel();
    let question_handler = handler.clone();
    std::thread::spawn(move || {
        // unwrap OK: fails only when OS resource limits exhausted anyway
        let runtime = Builder::new_current_thread().enable_all().build().unwrap();
        let result =
            runtime.block_on(async move { ask_question(&question_handler, question).await });
        // unwrap OK: rx.recv() does happen
        tx.send(result).unwrap();
    });
    // unwrap OK: tx.send() does happen
    rx.recv().unwrap()
```

rust/agama-software/src/callbacks.rs
```diff
@@ -9,11 +9,16 @@ mod security;
...
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+    let handler = handler.clone();
+    let res = rt.spawn(async move { ask_question(&handler, question).await });
+    rt.block_on(res).unwrap()
```
mvidner, line 25: unwrap comment needed! and as the type is not obvious, I'd prefer an intermediate `let` to make it more obvious

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
+                .enable_all()
+                .build()
+                .unwrap();
+            let res = rt.spawn(async move { (self.receiver.recv().await, self.receiver) });
+            let (action, receiver) = rt.block_on(res).unwrap();
+            self.receiver = receiver;
+
             let Some(action) = action else {
-                tracing::error!("Software action channel closed");
+                tracing::warn!("Software action channel closed");
```
mvidner: I must say that downgrading an error to a warning is still a problem to me, because it shows up in a *successful* test run.
Why not apply my `Quit` message?

https://github.com/agama-project/agama/pull/2925/files#diff-f1fcf26112a69086751592544b49d41f8268f54bda943ee33b8cf96566b76883R145

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
+    ) -> ZyppServerResult<UnboundedSender<SoftwareAction>> {
         let (sender, receiver) = mpsc::unbounded_channel();
 
-        let server = Self { receiver };
+        let server = Self {
+            receiver,
+            root_dir: root_dir.as_ref().to_path_buf(),
+        };
 
         // see https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html#use-inside-tokiospawn for explain how to ensure that zypp
```
mvidner: The comment explains the *previous* implementation

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
 
         loop {
-            let action = self.receiver.recv().await;
+            // well, the code here can look a bit tricky, but what happens in reality here is that we need synchronized code
+            // and only for small async interface run it in blocking way. Receiver handling is done to explicit passing
+            // of ownership as receiver do not implement Copy.
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .unwrap();
```
mvidner, line 141: unwrap safety comment, same as in the other instance of these calls I believe

rust/agama-software/src/zypp_server.rs
```diff
@@ -100,51 +100,54 @@ pub enum SoftwareAction {
...
+            let res = rt.spawn(async move { (self.receiver.recv().await, self.receiver) });
+            let (action, receiver) = rt.block_on(res).unwrap();
```
mvidner, line 144: unwrap OK?

rust/agama-software/src/callbacks.rs
jreidinger: yeap and that thread::spawn is expensive call I do not like...and also needs that blocking std::channel

rust/agama-software/src/zypp_server.rs
jreidinger: well, maybe we can downgrade it even  to info! level as it should happen always when agama finishes...I do not like that Quit Message, as it can emphasize that it can be started again which is something I would like to avoid due to possible libzypp artifacts in memory of this thread.

rust/agama-software/tests/zypp_server.rs
jreidinger: here it is quite tricky as libzypp testing should ideally happen in separate process on separate root. So sadly approach from other tests will be harder to apply. Lets see what more tests bring.

rust/agama-software/tests/zypp_server.rs
jreidinger: it is @mvidner reaction to my complain that with removing constant we need new unwrap, which I am not happy.

rust/agama-software/src/zypp_server.rs
jreidinger: there is no runtime in this thread. So we create new runtime for current thread and use it for blocking. `spawn::task` works well when you use tokio::main or have already existing runtime. I write some comment where I hope it is clear.

rust/agama-software/tests/zypp_server.rs
jreidinger: I also have to admit that I mainly take martin test and focus more on how to make rust code working.

rust/agama-software/src/zypp_server.rs
jreidinger: so to summarize it, I do not have there Quit message as libzypp thread is intended to run whole agama lifetime.

PR #2946, opened 2025-12-16
Fix product and patterns selection

imobachgs:
## Problem

* After selecting the product to install, the web user interface does not navigate to the overview page.
* The patterns selection is not working.

## Solution

* Adapt the product selector to navigate once the user picks a product.
* Fix the software overview and its main page to work with the new API.
* Fix the patterns selection page.

## Changes to the backend

The backend did not allow to explicitly remove a package. In the old version of Agama, that information was in libzypp only. We needed to extend the backend to support this case too.

However, two more things are missing:

* Exposing whether a resolvable was explictly removed by the user (we need to extend the bindings for that).
* Check why it is not possible to unselect a recommended dependency.

Both issues can be fixed later.

## Testing

- Added a new unit test
- Tested manually

## Follow-up

- Clean the old software types.

### DIFF
--- a/rust/agama-lib/share/package.json
+++ b/rust/agama-lib/share/package.json
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "validate": "ajv compile --spec=draft2019 --verbose --all-errors -r storage.schema.json -r iscsi.schema.json -s profile.schema.json && ajv compile --spec=draft2019 --verbose --all-errors -s storage.model.schema.json"
+    "validate": "ajv compile --spec=draft2019 --verbose --all-errors -r storage.schema.json -r iscsi.schema.json -r software.schema.json -s profile.schema.json && ajv compile --spec=draft2019 --verbose --all-errors -s storage.model.schema.json"
   },
   "dependencies": {
     "ajv-cli": "^5.0.0"
--- a/rust/agama-lib/share/software.schema.json
+++ b/rust/agama-lib/share/software.schema.json
@@ -72,6 +72,7 @@
     },
     "repository": {
       "description": "Packages repository",
+      "type": "object",
       "additionalProperties": false,
       "properties": {
         "alias": {


rust/agama-software/src/model/state.rs
```diff
@@ -296,54 +310,98 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
+    /// * `name`: resolvable name.
+    /// * `r#type`: resolvable type.
+    /// * `selection`: selection state.
+    pub fn add_or_replace(
+        &mut self,
+        name: &str,
+        r#type: ResolvableType,
+        selection: ResolvableSelection,
+    ) {
+        println!("Adding {}", name);
```
teclator: leftover?

rust/agama-software/src/model/state.rs
```diff
@@ -296,54 +310,98 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
+        if let Some(entry) = self.0.get(&(name.to_string(), r#type)) {
+            println!("entry {:?}", &entry);
```
teclator: idem

rust/agama-software/src/model/packages.rs
```diff
@@ -39,7 +39,16 @@ impl Resolvable {
...
+    Clone,
+    Copy,
+    Debug,
+    Deserialize,
+    Serialize,
+    strum::Display,
+    utoipa::ToSchema,
+    PartialEq,
+    Eq,
+    Hash,
```
jreidinger, line 51: NP: I prefer to have entries alphabetically ordered, as it is easier to read. I would apply it to all parts that do not depend on order.

rust/agama-software/src/model/state.rs
```diff
@@ -296,54 +310,102 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
+    /// * `r#type`: resolvable type.
+    /// * `selection`: selection state.
+    pub fn add_or_replace(
+        &mut self,
+        name: &str,
+        r#type: ResolvableType,
+        selection: ResolvableSelection,
+    ) {
+        if let Some(entry) = self.0.get(&(name.to_string(), r#type)) {
+            if let ResolvableSelection::AutoSelected { optional: false } = entry {
```
jreidinger: I am still not sure if optional is proper here as sadly we use it in two ways for resolvables:
1. optional like auto-selected by something else ( probably not what is here )
2. optional like mandatory if available, otherwise skip ( as defined for products ). In such case I would not remove it here as it should also not be removed.

rust/agama-software/src/model/state.rs
```diff
@@ -296,54 +310,102 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
+    ) {
+        self.add_or_replace(&resolvable.name, resolvable.r#type, selection);
+    }
+
+    /// Reset the list of resolvables.
+    ///
+    /// The mandatory resolvables are preserved.
+    pub fn reset(&mut self) {
+        self.0.retain(|_, selection| match selection {
+            ResolvableSelection::AutoSelected { optional: false } => true,
```
jreidinger, line 363: and similar issue here.

rust/agama-software/src/model/state.rs
```diff
@@ -296,54 +310,102 @@ impl From<&agama_utils::api::software::Repository> for Repository {
...
+            _ => false,
+        });
+    }
+
+    /// Turns the list of resolvables into a vector.
+    ///
+    /// FIXME: return an interator instead.
```
jreidinger, line 370: :+1: as probably hash set is not needed. I kind of expect that common Vector will be enough.

rust/agama-software/src/zypp_server.rs
```diff
@@ -375,6 +373,39 @@ impl ZyppServer {
...
+        name: &str,
+        r#type: ResolvableType,
+        reason: zypp_agama::ResolvableSelected,
+        optional: bool,
+    ) -> Vec<Issue> {
+        let mut issues = vec![];
+        let result = zypp.select_resolvable(name, r#type.into(), reason);
+
+        if let Err(error) = result {
+            if optional {
```
jreidinger, line 388: well, it can cloak real errors. Ideally we should for optional at first ask if resolvable is available ( not available now in agama-zypp ) and skip if not.

so ideally something like
```
if optional && !zypp.resolvable_available(name, r#typ.into()) {
  return issues;
}
```

rust/agama-software/src/model/state.rs
imobachgs: A `HashSet` is used in the tests to make things easier. We could use a `Vec` but you need to ensure the order (using the name and the type, for instance).

web/src/App.tsx
```diff
@@ -61,11 +59,6 @@ const Content = () => {
     return <Navigate to={ROOT.installationFinished} />;
   }
 
-  if (product?.id === undefined && !isBusy && location.pathname !== PRODUCT.root) {
-    console.log("Navigating to the product selection page");
-    return <Navigate to={PRODUCT.root} />;
```
jreidinger, line 66: Why it can be removed? where redirect to product selection happen?

jreidinger: ah, it is below in overview page. Not sure which location I like more.

web/src/components/software/SoftwarePage.tsx
```diff
@@ -132,24 +137,28 @@ const ReloadSection = ({
...
 
   // Selected patterns section should fill the full width in big screen too when
   // there is no information for rendering the Proposal Size section.
-  const selectedPatternsXlSize = proposal.size ? 6 : 12;
+  const selectedPatternsXlSize = usedSpace ? 6 : 12;
 
   const startProbing = () => {
     setLoading(true);
-    probe();
+    // probe();
```
jreidinger: this should probably also be TODO:

web/src/model/config/software.ts
```diff
@@ -0,0 +1,48 @@
...
+export type FlagIfOnlyMinimalHardDependenciesShouldBeUsedInSolver = boolean;
+export type AliasUsedForRepositoryActingAsIdentifier = string;
+export type URLPointingToRepository = string;
+export type RepositoryPriority = number;
+export type UserVisibleNameDefaultsToAlias = string;
+export type ProductDirectoryOnMultiRepoDVDUsuallyNotNeeded = string;
+export type IfRepositoryShouldBeEnabledDefaultsToTrueUsefulWhenAddingAdditionalRepoThatShouldNotBeImmediatelyUse =
+  boolean;
+export type IfUnsignedRepositoriesAreAllowedMainlyUsefulForRepositoriesThatIsHandCraftedWithoutGPGSignature = boolean;
+export type ListOfGPGFingerprintsThatIsAcceptedForThisRepositoryUsefulForOwnRepositoriesWithProperGPGSignature =
```
jreidinger: just WOW...this name is probably longest name type I ever see :) I am not sure if it is good idea to generate type from description of property :)

web/src/model/proposal/software.ts
```diff
@@ -31,11 +31,13 @@ type PatternsSelection = { [key: string]: SelectedBy };
...
+  AUTO = "auto",
   /** No selected */
-  NONE = 2,
+  NONE = "none",
 }
 
-export type { Proposal, PatternsSelection, SelectedBy };
+export type { Proposal, PatternsSelection };
+
+export { SelectedBy };
```
jreidinger, line 43: why now enum is no longer exported as type?

rust/agama-software/src/model/state.rs
jreidinger: well, in general I would try to not affect production code by tests...having ordered Vec is ok for me. If HashSet is needed only in test, then test should construct it from vector.

web/src/model/config/software.ts
imobachgs: Yes, IMHO the problem is that we should not use _List of GPG fingerprints that is accepted for this repository. Useful for own repositories with proper GPG signature_ as [the title](https://github.com/agama-project/agama/blob/master/rust/agama-lib/share/profile.schema.json#L260). That should be the description.

rust/agama-software/src/model/state.rs
imobachgs: Well, I do not have a strong opinion about that, if the API can be improved and used in tests, it is fine. For instance, we do it for instance everytime we implement `PartialEq` or `Eq` for tests.

Anyway, I am using a `Vec` now.

web/src/model/proposal/software.ts
imobachgs: I would say `export type` is just for types. I tried and it works just fine.

web/src/App.tsx
imobachgs: I will discuss with David. Anyway, we want to stop using `App` as a "router". Most probably the redirection should be in the router itself, but I do not think we are at that point.

jreidinger (approved):

PR #2949, opened 2025-12-16
translate strings originating in Rust code

mvidner:
## Problem

We do not have a mechanism to translate strings originating in the Rust code yet

- https://trello.com/c/DSk54XYG/5303-research-localized-strings-in-rust-code

## Solution

- Translation works for strings originating in Rust code
  (tested on a "Pattern summary" string that was added and reverted)
- Translation of format strings works, via
  https://lib.rs/crates/i18n-format
- Added the appropriate CI Weblate jobs for `../agama-weblate`:
  to put the POT file there and get the PO files from there.
- A new Weblate component to be souced from `agama-weblate/rust/agama.pot`,  :warning: TODO: set it up on https://l10n.opensuse.org/
- Introduced `agama_l10n::helpers::gettext_noop`,
  marks a string for translation but does not translate it
- `QuestionSpec` instantiation simplified with a new `with_action_ids` fn

## Testing

Tested manually: with `/testing_using_container.sh` which exposes Agama at `localhost:10443`, on an artificially added string that is prominently visible in the Software page.

## Screenshots

This screenshot shows a successful translation of a Czech string (added in 2ecb163341e46e0d7d6ee8ac944c8d59bc9ba016 and later reverted, because it is actually useless). The useful strings occur mostly in error cases and are hard to trigger.

agama-cs-pattern-summary2u

## Documentation

- added `rust/po/README.md` and an identical copy, `service/po/README.md`
- https://github.com/agama-project/agama-project.github.io/pull/137

### DIFF
--- a/rust/po/cs.po
+++ b/rust/po/cs.po
@@ -71,8 +71,3 @@ msgstr ""
 #: agama-software/src/zypp_server.rs:486
 msgid "Pattern summary"
 msgstr "Shrnutí vzoru"
-
-#: agama-software/src/zypp_server.rs:561
-#, rust-format
-msgid "Pattern summary: {0} {1}"
-msgstr "Shrnutí vzoru {crashme} {0}"
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -18,7 +18,6 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
-use i18n_format::i18n_format;
 use agama_utils::{
     actor::Handler,
     api::{
@@ -558,7 +557,7 @@ impl ZyppServer {
                     category: p.category,
                     description: p.description,
                     icon: p.icon,
-                    summary: i18n_format!("Pattern summary: {0} {1}", p.summary, ":-)"),
+                    summary: p.summary,
                     order: p.order,
                     preselected,
                 }


rust/agama-autoinstall/src/questions.rs
```diff
@@ -37,12 +38,12 @@ impl UserQuestions {
     /// Asks the user whether to retry loading the profile.
     pub async fn should_retry(&self, text: &str, error: &str) -> anyhow::Result<bool> {
         let question = QuestionSpec::new(text, "load.retry")
-            .with_actions(&[("yes", "Yes"), ("no", "No")])
-            .with_default_action("no")
+            .with_action_ids(&[gettext_noop("Yes"), gettext_noop("No")])
+            .with_default_action("No")
```
jreidinger, line 42: just be aware that case sensitivity is part of questions API for automated answering via "answers" section of profile. In this case it looks like correct fix as documented options are Yes/No - https://agama-project.github.io/docs/user/reference/profile/answers#supported-question-classes

rust/agama-software/src/callbacks/security.rs
```diff
@@ -61,15 +62,16 @@ impl security::Callback for Security {
             key_fingerprint,
         );
         // TODO: support for extra_repositories with specified gpg key checksum
-        // TODO: localization with params
-        let text = format!(
-            "The key {key_id} ({key_name}) with fingerprint {key_fingerprint} is unknown. \
-              Do you want to trust this key?"
+        let text = i18n_format!(
+            // TRANSLATORS: substituting: key ID, (key name), fingerprint
+            "The key {} ({}) with fingerprint {} is unknown. \
```
jreidinger: well, in general I would discourage using generic `{}` for any string with than one parameter, as it disallow translators to move values elsewhere. 
So in this case I would use something like `The key {id} ({name}) with fingerprint {fingerprint} is unknown.` so translators can shuffle around.

rust/agama-utils/src/lib.rs
```diff
@@ -34,3 +34,11 @@ pub mod products;
 pub mod progress;
 pub mod question;
 pub mod test;
+
+/// Does nothing at runtime, marking the text for translation.
+///
+/// This is useful when you need both the untranslated id
+/// and its translated label, for example.
+pub fn gettext_noop(text: &str) -> &str {
```
jreidinger, line 42: NP: alternative name can be `gettext_mark` to emphasize that it marks text for translation. Btw I am not sure if it make sense to allow any `&str`. Will it work if I pass variable there? Or does it need only `&'static str` ?

rust/po/cs.po
```diff
@@ -0,0 +1,73 @@
+# SOME DESCRIPTIVE TITLE.
```
jreidinger, line 1: should this be kept?

rust/agama-software/src/callbacks/security.rs
mvidner: Good point. It turns out your suggested solution will not work but numbered arguments should work `"Eat the {1}, the big {0} surely will"`, as noted in **Limitations** of https://crates.io/crates/formatx

I will verify this and improve the docs

rust/agama-autoinstall/src/questions.rs
mvidner: I was aware but I thought that I was breaking it for the sake of consistency... Thanks for the link

rust/po/cs.po
mvidner: All other PO files have it: https://github.com/search?q=repo%3Aagama-project%2Fagama%20SOME%20DESCRIPTIVE%20TITLE&type=code
... and it is [hardcoded by xgettext](https://gitweb.git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=gettext-tools/src/xgettext.c;hb=d5c190b1fdaeab223e01f2a739fde8e11a3435e9#l2248) so I think we just leave it as is.

rust/agama-utils/src/lib.rs
mvidner: About the reference **lifetime**:
```rs
fn gettext_noop(text: &str) -> &str
```
is an elided form of this (which I've tried explicitly in my working copy)
```rs
fn gettext_noop<'a>(text: &'a str) -> &'a str
```
which means that the output lives just as long as the input. The lifetime does not change, and I guess the call will be optimized away. No problem here.

mvidner: About the **name**: `gettext_noop` is suggested by the Gettext manual so I went with that: https://www.gnu.org/software/gettext/manual/html_node/Special-cases.html

jreidinger (approved):

rust/agama-software/src/callbacks/security.rs
lslezak: Great, xgettext also adds the `rust-format` flag to the message in the POT file. :+1:

PR #2950, opened 2025-12-16
Adapt the CLI to the new API

jreidinger:
## Problem

`agama config` subcommands do not work with new API.

## Solution

Adapt all config subcommands to work with it.

## Testing

- *Tested manually* using new iso and --host option.

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -22,21 +22,18 @@ use std::{io::Write, path::PathBuf, process::Command};
 
 use agama_lib::profile::ProfileHTTPClient;
 use agama_lib::{
-    context::InstallationContext, http::BaseHTTPClient, install_settings::InstallSettings,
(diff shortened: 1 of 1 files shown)


rust/agama-cli/src/config.rs
```diff
@@ -108,67 +107,67 @@ pub async fn run(subcommand: ConfigCommands, opts: GlobalOpts) -> anyhow::Result
...
-            Ok(())
+            let result = edit(&http_client, &response, &editor).await?;
+            patch_config(&http_client, &result).await?;
         }
     }
+
+    Ok(())
+}
+
+async fn patch_config(
```
imobachgs, line 161: Alternatively, we could have a single client which implements this logic. It could be a wrapper of the `BaseClient` or the `BaseClient` itself.

jreidinger: yes, we will see what we will need in other places that uses HTTP api and then can refactor it.

PR #2951, opened 2025-12-16
Implement a new configuration merging logic

imobachgs:
## Problem

In the new API, when merging collections (e.g., a `Vec`), both of them are merged together. For instance, if you send a PATCH with a list of patterns for installation, it would add them to the existing list.

The way to reset the list of patterns is to use PUT, but it forces to send the whole configuration.

The reason for this behavior is that we rely on [merge_struct](https://docs.rs/merge-struct/latest/merge_struct/) for merging the configuration. It was kind of an urgent solution, but we might need to search for a better alternative (or implement our own).

## Solution 

Replace the whole list, so we can add, update, and remove elements using a `PATCH` request. It uses the [merge](https://crates.io/crates/merge) crate. When reviewing the code, take into account that when calling the `merge` function, the receiver is the configuration to update, while the argument to this function is the original one.

```rust
config.update(original);
```

## Notes

Most of the unit tests are generated by Gemini, although some minimal changes were required.

## Open question

* How to proceed in the `storage` section?

### DIFF
--- a/rust/agama-lib/src/questions/http_client.rs
+++ b/rust/agama-lib/src/questions/http_client.rs
@@ -112,7 +112,7 @@ impl HTTPClient {
         answers: Vec<AnswerRule>,
     ) -> Result<(), QuestionsHTTPClientError> {
         let questions = QuestionsConfig {
-            answers,
+            answers: Some(answers),
             ..Default::default()
         };
         let config = Config {


rust/agama-utils/src/api/files/config.rs
```diff
@@ -27,16 +28,19 @@ use crate::api::files::{
     FileSourceError, WithFileSource,
 };
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize, utoipa::ToSchema)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Merge, utoipa::ToSchema)]
 pub struct Config {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    #[merge(strategy = merge::vec::overwrite_empty)]
```
jreidinger: so it means, if I intentionally send empty list of files, it won't be overwritten and instead it will keep old list of files? I find this strange.

rust/agama-utils/src/api/files/config.rs
```diff
@@ -88,3 +92,56 @@ impl ScriptsConfig {
...
+            pre: Some(vec![build_pre_script("updated")]),
+            ..Default::default()
+        };
+
+        updated.merge(original);
+        let pre_scripts = updated.pre.unwrap();
+        assert_eq!(pre_scripts.len(), 1);
+        let script = pre_scripts.get(0).unwrap();
+        assert_eq!(&script.base.name, "updated");
+    }
```
jreidinger, line 146: I miss tests for `files` case.

rust/agama-utils/src/api/files/config.rs
imobachgs: Yes, I think it is wrong. But IMHO the real problem is that we should use an `Option` here too. Otherwise, you cannot differentiate when you want to reset the list.

rust/agama-utils/src/api/network/settings.rs
```diff
@@ -23,12 +23,15 @@
...
+use merge::Merge;
 use serde::{Deserialize, Serialize};
 use std::default::Default;
 use std::net::IpAddr;
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize, utoipa::ToSchema)]
-pub struct NetworkConnectionsCollection(pub Vec<NetworkConnection>);
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Merge, utoipa::ToSchema, PartialEq)]
+pub struct NetworkConnectionsCollection(
+    #[merge(strategy = merge::vec::overwrite_empty)] pub Vec<NetworkConnection>,
```
jreidinger: I would avoid having this on one line, is possible to have it on two?

jreidinger: and question is also here how it will behave when I send intentionally empty list of network connections, will it overwrite or not?

imobachgs: No, I left the formatting to `rustfmt` and it does it in this way.

rust/agama-utils/src/api/question.rs
```diff
@@ -466,4 +469,140 @@ mod tests {
...
+            policy: None,
+            answers: vec![AnswerRule {
+                class: Some("foo".to_string()),
+                text: None,
+                data: None,
+                answer: Answer::new("yes"),
+            }],
+        };
+        updated2.merge(original2.clone());
+        assert_eq!(updated2.answers.len(), 1);
```
jreidinger: and this exactly show why I do not think that overwrite_empty is good one for arrays as it means, that you cannot set intentionally empty list.

jreidinger (approved):

PR #2952, opened 2025-12-17
CLI - Adapt `agama install` to new API

jreidinger:
## Problem

Without `agama install` unattended installation does not work.

## Solution

Adapt it to new API and also adapt Monitor client, so it is possible to wait for result. Adapting `agama progress` was left out of this pull request.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -18,15 +18,14 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
-use std::thread::sleep;
-use std::time::Duration;
-use std::{io::Write, path::PathBuf, process::Command};
+use std::{io::Write, path::PathBuf, process::Command, time::Duration};
 
-use agama_lib::monitor::MonitorClient;
-use agama_lib::profile::ProfileHTTPClient;
 use agama_lib::{
-    context::InstallationContext, http::BaseHTTPClient, profile::ProfileValidator,
-    profile::ValidationOutcome, utils::FileFormat,
+    context::InstallationContext,
+    http::BaseHTTPClient,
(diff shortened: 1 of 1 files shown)


rust/agama-cli/src/config.rs
```diff
@@ -18,8 +18,11 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
+use std::thread::sleep;
+use std::time::Duration;
 use std::{io::Write, path::PathBuf, process::Command};
 
+use agama_lib::monitor::MonitorClient;
```
imobachgs: I would gruop all `agama_lib` imports.

rust/agama-cli/src/lib.rs
```diff
@@ -108,40 +110,27 @@ async fn probe(manager: ManagerHTTPClient, monitor: MonitorClient) -> anyhow::Re
...
-    monitor: MonitorClient,
-    max_attempts: usize,
-) -> anyhow::Result<()> {
+async fn install(http_client: BaseHTTPClient, monitor: MonitorClient) -> anyhow::Result<()> {
     wait_until_idle(monitor.clone()).await?;
 
-    let status = manager.status().await?;
-    if !status.can_install {
+    let status = monitor.get_status().await?;
+    // TODO: own client for issues?
```
imobachgs, line 115: Most probably I would have a single client which exposes the whole API.

jreidinger: fine for me :)

rust/agama-cli/src/config.rs
jreidinger: yeap, vscode start doing this after latest update, not sure why or how to disable it.

PR #2954, opened 2025-12-17
Web: Adapt storage unit tests to api-v2

ancorgs:
## Problem

Many things have changed in the api-v2 branch, so Javascript unit tests got broken.

## Solution

Fix the mocks so the tests are green again

Take the opportunity to improve some of the tests (and also to fix some broken types in other tests).


joseivanlopez (approved):
LGTM

PR #2955, opened 2025-12-18
Merge the storage config

imobachgs:
## Problem

The `storage` configuration is not merged, just replaced.

## Solution

Merge the first level of the `storage` section. The `legacy_autoyast_storage` is still replaced as it only contains a `drives` list.

## Testing

- Added a new unit test

### DIFF
--- a/rust/agama-utils/src/api/storage/config.rs
+++ b/rust/agama-utils/src/api/storage/config.rs
@@ -22,19 +22,87 @@ use merge::Merge;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize, Merge, utoipa::ToSchema)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Config {
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[merge(strategy = merge::option::overwrite_none)]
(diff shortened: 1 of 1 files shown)


rust/agama-utils/src/api/storage/config.rs
```diff
@@ -38,3 +36,73 @@ impl Config {
...
+    let Value::Object(left_object) = left else {
+        return;
+    };
+
+    let Value::Object(right_object) = right else {
+        return;
+    };
+
+    for (k, v) in right_object {
+        left_object.entry(k).or_insert(v);
```
joseivanlopez, line 70: I would expect `right_object` takes precedence, for example:

~~~
config.storage.drives = ...
other.storage.drives = ...
config.storage.drives == other.storage.drives #=> false
config.merge(other)
config.storage.drives == other.storage.drives #=> true
~~~

With the current implementation, `#merge` actually means "set to the left object the missing properties provided by the right object". I expected this other meaning: "set to the left object all the properties provided by the right object". What do you think?

imobachgs: I understand your confusion, but in [merge](https://crates.io/crates/merge) the left side is the "new configuration". Think about the "right" object as the default values (that belongs to the old configuration). So I would expect to use the "left" values if available.

imobachgs: Thinking about this, I could have implemented this as a separate strategy, but as I do not expect to reuse it, implementing `Merge` looks fine.

PR #2958, opened 2025-12-18
web: refactor progress backdrop query tracking 

dgdavid:
Initially, internal `ProgressBackdrop` component at `core/Page`, relied on a global event system to detect when proposal-related queries finished refetching in order to consider an ongoing progress in the given scope was finished (check https://github.com/agama-project/agama/pull/2947). While this worked on the storage page, it failed on other pages becuae the _central_ `useProposalChanges` hook is invalidating an storage-related query, which is not refetched until navigating to a storage-related page. As a result, the progress backdrop could remain stuck on unrelated storage pages.

This commit refactors the tracking mechanism to make it more flexible and page-aware:

 * Start exporting query keys as constants (COMMON_PROPOSAL_KEYS, STORAGE_MODEL_KEY, etc.) from their respective modules

 * Add the `additionalProgressKeys` prop to Page, allowing pages to specify extra query keys to track alongside the common proposal-related queries

 * Use `useTrackQueriesRefetch` directly into ProgressBackdrop, dropping the need for a custom global event system

 * Remove the proposal:updated event infrastructure (onProposalUpdated, dispatchProposalUpdated), which is no longer required

With this change, pages can explicitly declare which additional queries must complete refetching before the UI is unblocked. For example, the storage `ProposalPage` now waits for the `storageModel` query in addition to the common proposal queries.

While requiring pages to be explicit about the queries they depend on is not ideal, this preserves common proposal tracking as the default while preventing the UI from remaining blocked due to unrelated query invalidations.

### DIFF
--- a/rust/agama-files/src/runner.rs
+++ b/rust/agama-files/src/runner.rs
@@ -19,8 +19,9 @@
 // find current contact information at www.suse.com.
 
 use std::{
-    fs::File,
+    fs::{self, File},
     io::{self, Read, Seek, SeekFrom, Write},
+    os::unix::fs::symlink,
     path::{Path, PathBuf},
     process::ExitStatus,
 };
@@ -43,6 +44,11 @@ pub enum Error {
     #[error(transparent)]
     Question(#[from] question::AskError),
 }
+
+// Relative path to the resolv.conf file.
+const RESOLV_CONF_PATH: &str = "etc/resolv.conf";
+// Relative path to the NetworkManager resolv.conf file.
+const NM_RESOLV_CONF_PATH: &str = "run/NetworkManager/resolv.conf";
 
 /// Implements the logic to run a script.
 ///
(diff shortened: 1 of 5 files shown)


web/src/components/storage/ProposalPage.tsx
```diff
@@ -328,7 +328,7 @@ export default function ProposalPage(): React.ReactNode {
   if (resetNeeded) return;
 
   return (
-    <Page progressScope="storage">
+    <Page progressScope="storage" additionalProgressKeys={STORAGE_MODEL_KEY}>
```
imobachgs, line 331: np: Is a "Key" or a "Query"? I know it is a "query key", but it refers to a query.

dgdavid: You're right... I tried to keep the prop shorter but I've another idea in mind for next iteration that might help. Stay tuned :wink:

imobachgs (approved):

PR #2961, opened 2025-12-18
Handle WebSockets disconnection

imobachgs:
## Problem

Agama does not handle WebSockets disconnections properly. When a client gets disconnected, Agama insists on sending the events to the client.

## Solution

Properly handle WebSockets disconnection. In the example below, you can read the logs in that case (they are coming from [tungstenite](https://github.com/snapview/tungstenite-rs)):

```
dic 18 22:22:51 agama agama-web-server[6893]: Received close frame: Some(CloseFrame { code: Away
, reason: "" })
dic 18 22:22:51 agama agama-web-server[6893]: Replying to close with Frame { header: FrameHeader
 { is_final: true, rsv1: false, rsv2: false, rsv3: false, opcode: Control(Close), mask: None }, 
payload: [3, 233] }
```

Additionally, Agama does not send the events from the old API anymore.

## Testing

- Tested manually

### DIFF
--- a/rust/agama-server/src/web/ws.rs
+++ b/rust/agama-server/src/web/ws.rs
@@ -21,7 +21,7 @@
 //! Implements the websocket handling.
 
 use super::state::ServiceState;
-use agama_lib::{auth::ClientId, http};
+use agama_lib::auth::ClientId;
 use agama_utils::api::event;
 use axum::{
     extract::{
@@ -31,7 +31,6 @@ use axum::{
     response::IntoResponse,
     Extension,
 };
-use serde::Serialize;
 use std::sync::Arc;
 use tokio::sync::broadcast;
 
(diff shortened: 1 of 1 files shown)


rust/agama-server/src/web/ws.rs
```diff
@@ -50,45 +49,49 @@ pub async fn ws_handler(
...
 
     loop {
         tokio::select! {
-            msg = old_events_rx.recv() => {
-                if let Err(e) = send_msg(&mut socket, msg).await {
-                    eprintln!("Error sending old event: {e:?}");
+            // Handle messages from the client
+            msg = socket.recv() => {
+                match msg {
+                    Some(Ok(Message::Close(_))) | Some(Err(_)) | None => break,
```
jreidinger, line 70: Maybe this deserve a comment as I need to check exact websocket API to understand all cases (explicit close message, None for closed stream and error for generic axum error ). And I am not sure if breaking in case of generic axum error is good or not as sadly reading axum code for method `recv` does not help understand when it can return Err

jreidinger (approved):
much better than before

PR #2963, opened 2025-12-18
web: improve product selection submit feedback

dgdavid:
Set product selection submit button to a loading state on form submission so users receive immediate visual feedback that the action has started.

It also fix mockings in test.


imobachgs (approved):
LGTM

PR #2965, opened 2025-12-19
web: improve ProgressBackdrop tracking API and its look & feel

dgdavid:
## Tracking API

Use a single options object to configure progress backdrop behavior, instead of passing multiple loosely related props through consumers.

```diff
- <Page progressScope="storage" additionalProgressKeys={STORAGE_MODEL_KEY}>
+ <Page progress={{ scope: "storage", ensureRefetched: STORAGE_MODEL_KEY }}>
```

## Look and feel

The progress message is now always visible, even if the user had previously scrolled the content before the progress appeared.

| Before  | After |
|-|-|
| localhost_8080_ (20) | localhost_8080_ (23) |

Legibility has been improved by increasing the text size and using a gray overlay background from the SUSE brand palette. Although we initially kept the default PF Backdrop background to provide immediate visual feedback about the blocked content area, this feedback is already sufficiently conveyed by the blurred content underneath.

Progress actions are expected to be short-lived; if they do take longer, users can still clearly perceive that interaction is blocked, while benefiting from significantly improved text readability (black on gray instead of black on near-black).

| Before | After |
|-|-|
| localhost_8080_ (19) | localhost_8080_ (18)|

### DIFF
--- a/web/src/components/core/ProgressBackdrop.tsx
+++ b/web/src/components/core/ProgressBackdrop.tsx
@@ -30,16 +30,18 @@ import useTrackQueriesRefetch from "~/hooks/use-track-queries-refetch";
 import type { Scope } from "~/model/status";
 import { _ } from "~/i18n";
 import textStyles from "@patternfly/react-styles/css/utilities/Text/text";
+
 /**
  * Props for the ProgressBackdrop component.
  */
 export type ProgressBackdropProps = {
   /**
-   * Optional scope identifier to filter which progresses trigger the backgrop
+   * Scope identifier to filter which progresses trigger the backgrop
    * overlay. If undefined or no matching tasks exist, the backdrop won't be
    * displayed.
    */
-  progressScope?: Scope;
+  scope: Scope;
+
   /**
    * Additional query keys to track during progress operations.
    *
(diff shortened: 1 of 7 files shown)


web/src/components/core/ProgressBackdrop.tsx
```diff
@@ -52,43 +55,43 @@ export type ProgressBackdropProps = {
...
-   * <Page
-   *   progressScope="network"
-   *   additionalProgressKeys={[NETWORK_CONFIG_KEY, CONNECTIONS_KEY]}
+   * <ProgressBackdrop
+   *   track="network"
+   *   ensureRefecthed={[NETWORK_CONFIG_KEY, CONNECTIONS_KEY]}
    * >
    */
-  additionalProgressKeys?: string | string[];
+  ensureRefetched?: string | string[];
```
dgdavid, line 67: The idea here is to use a [Tagged](https://github.com/sindresorhus/type-fest/blob/main/source/tagged.d.ts) type, similar to what was already done for
[`TranslatedString`](https://github.com/agama-project/agama/pull/2928). This would allow queries to export their keys as a `QueryKey` (or similar), enabling the compiler to verify that valid values are provided to the `ensureRefetched` prop.

PR #2978, opened 2025-12-30
Support for registering the base system

imobachgs:
This PR implements initial support for products and extensions registration. It has a few limitations that should be addressed in separate pull requests:

* The web UI is not adapted.
* It does not allow to register extensions (only the base product). However, the logic to activate them is already implemented.
* It does not copy the credentials at the end of the installation.
* It does not handle certificates.
* It does not support proxies or specific URLs (from kernel's command-line).
* It always uses "en-us" as language.
* (optional) Although not needed by the UI, we could expose the list of added services in the system information object.

## System information

The registration information is exposed through the `/api/v2/system` end-point.

```json
{
  "software": {
    "patterns": [],
    "repositories": [],
    "registration": {
      "code": "XXXXXXXXXXXXXXXX",
      "email": null,
      "url": "https://scc.suse.com/",
      "addons": [
        {
          "id": "sle-ha",
          "version": "16.1",
          "label": "SUSE Linux Enterprise High Availability Extension 16.1 x86_64 (ALPHA)",
          "available": true,
          "free": false,
          "recommended": false,
          "description": "SUSE Linux High Availability Extension provides mature, industry-leading open-source high-availability clustering technologies that are easy to set up and use. It can be deployed in physical and/or virtual environments, and can cluster physical servers, virtual servers, or any combination of the two to suit the needs of your business.",
          "type": "extension",
          "release": "alpha"
        },
        {
          "id": "PackageHub",
          "version": "16.1",
          "label": "SUSE Package Hub 16.1 x86_64",
          "available": true,
          "free": true,
          "recommended": false,
          "description": "SUSE Package Hub is a free-of-charge module providing access to community-maintained packages built to run on SUSE Linux Enterprise Server. Built from the same sources used in openSUSE distributions, these quality packages provide additional software to what is found in the SUSE Linux Enterprise Server product. The packages in this module are delivered without L3 support but do not interfere with the supportability of SUSE Linux Enterprise Server.",
          "type": "extension",
          "release": "released"
        }
      ]
    }
  }
}
```

NOTE: at this point the `type` property is always "extension" because it is not exposed by our bindings.

## Issues

When the system cannot be registered, Agama will create an issue indicating the problem:

```json
[
  {
    "scope": "software",
    "class": "software.register_system",
    "description": "Failed to register the system",
    "details": "Unknown Registration Code."
  }
]
```

### DIFF
--- a/rust/agama-software/src/model/registration.rs
+++ b/rust/agama-software/src/model/registration.rs
@@ -24,7 +24,6 @@
 //! the system and its add-ons and with libzypp (through [zypp_agama]) to add the
 //! corresponding services to `libzypp`.
 
-use agama_utils::api::software::{AddonInfo, RegistrationInfo};
 use agama_utils::{
     api::software::{AddonInfo, RegistrationInfo},
     arch::Arch,
@@ -116,26 +115,20 @@ impl Registration {
     /// but the list of extensions.
     pub fn to_registration_info(&self) -> RegistrationInfo {
         let addons: Vec<AddonInfo> = match self.base_product() {
-            Ok(product) => {
-                product
-                    .extensions
-                    .into_iter()
-                    .map(|e| {
-                        AddonInfo {
-                            id: e.identifier,
-                            version: e.version,
-                            label: e.friendly_name,
-                            available: e.available,
-                            free: e.free,
-                            recommended: e.recommended,
-                            description: e.description,
-                            release: e.release_stage,
-                            // FIXME: missing from suseconnect_agama
-                            r#type: "extension".to_string(),
-                        }
-                    })
-                    .collect()
-            }
+            Ok(product) => product
+                .extensions
+                .into_iter()
+                .map(|e| AddonInfo {
+                    id: e.identifier,
+                    version: e.version,
+                    label: e.friendly_name,
+                    available: e.available,
+                    free: e.free,
+                    recommended: e.recommended,
+                    description: e.description,
+                    release: e.release_stage,
+                })
+                .collect(),
             Err(error) => {
                 tracing::error!("Failed to get the product from the registration server: {error}");
                 vec![]
@@ -234,7 +227,7 @@ impl RegistrationBuilder {
     /// It announces the system, gets the credentials and registers the base product.
     ///
     /// * `zypp`: zypp instance.
-    pub fn build(self, zypp: &zypp_agama::Zypp) -> RegistrationResult<Registration> {
+    pub fn register(self, zypp: &zypp_agama::Zypp) -> RegistrationResult<Registration> {
         let params = suseconnect_agama::ConnectParams {
             token: self.code.clone(),
             email: self.email.clone(),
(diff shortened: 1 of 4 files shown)


jreidinger:
@imobachgs I discuss that type with @lslezak and for SLE16 it does not make sense as we do not have other types there. So question is if we need to expose it at all.

rust/agama-software/src/model/registration.rs
```diff
@@ -18,47 +18,251 @@
...
+                        AddonInfo {
+                            id: e.identifier,
+                            version: e.version,
+                            label: e.friendly_name,
+                            available: e.available,
+                            free: e.free,
+                            recommended: e.recommended,
+                            description: e.description,
+                            release: e.release_stage,
+                            // FIXME: missing from suseconnect_agama
```
jreidinger: if needed I can add it, but not sure if we really need it.

rust/agama-software/src/model/registration.rs
```diff
@@ -18,47 +18,251 @@
...
+    }
+
+    fn base_product_specification(&self) -> suseconnect_agama::ProductSpecification {
+        Self::product_specification(&self.product, &self.version)
+    }
+
+    fn product_specification(id: &str, version: &str) -> suseconnect_agama::ProductSpecification {
+        suseconnect_agama::ProductSpecification {
+            identifier: id.to_string(),
+            arch: std::env::consts::ARCH.to_string(),
```
jreidinger: are you sure that it will match all rpm archs that suse connect uses?

rust/agama-software/src/model/registration.rs
```diff
@@ -18,47 +18,251 @@
...
+        self.email = Some(email.to_string());
+        self
+    }
+
+    /// Registers the system and return a [Registration] object.
+    ///
+    /// It announces the system, gets the credentials and registers the base product.
+    ///
+    /// * `zypp`: zypp instance.
+    pub fn build(self, zypp: &zypp_agama::Zypp) -> RegistrationResult<Registration> {
```
jreidinger: I am not sure if it is not confusing to call it Builder::build and in fact it do `registration`...so I kind of expect more method like `register` or `register_base_product`.

rust/agama-software/src/zypp_server.rs
```diff
@@ -656,4 +672,29 @@ impl ZyppServer {
...
+    fn register_system(
+        &mut self,
+        state: &RegistrationState,
+        zypp: &zypp_agama::Zypp,
+        issues: &mut Vec<Issue>,
+    ) {
+        let mut registration =
+            Registration::builder(self.root_dir.clone(), &state.product, &state.version)
+                .with_code(&state.code);
+        registration = registration.with_code(&state.code);
```
jreidinger: this line is not needed as `with_code` is already set above.

rust/agama-software/src/model/registration.rs
imobachgs: Good point. Actually, the `pcc64le` does not match (it uses `powerpc64` instead). As we will need to do more conversion, I added a new `Arch` enum that allows finding the current architecture and brings the possibility of adding conversions to it.

jreidinger (approved):

PR #2982, opened 2025-12-31
Add support for registering add-ons + refresh services

imobachgs:
Follow-up of #2978.

This PR adds support, at HTTP API level, for registering add-ons. Additionally, it exposes such information as part of the `/api/v2/system` resource.

Additionally, this PR makes sure that the services are refreshed after adding them.

## Notes

* The representation of the add-on status might be improved. But we can postpone it a little (while adapting the UI).

### DIFF
--- a/rust/agama-software/src/model/registration.rs
+++ b/rust/agama-software/src/model/registration.rs
@@ -138,9 +138,10 @@ impl Registration {
         // Add the libzypp service
         zypp.add_service(&service.name, &service.url)
             .map_err(|e| RegistrationError::AddService(service.name.clone(), e))?;
+        let name = service.name.clone();
         self.services.push(service);
-        zypp.refresh_service(&service.name)
-            .map_err(|e| RegistrationError::RefreshService(service.name.clone(), e))?;
+        zypp.refresh_service(&name)
+            .map_err(|e| RegistrationError::RefreshService(name, e))?;
         Ok(())
     }
 


rust/agama-software/src/model/registration.rs
```diff
@@ -56,15 +60,44 @@ pub struct Registration {
     connect_params: ConnectParams,
     creds: Credentials,
     services: Vec<suseconnect_agama::Service>,
+    // Holds the addons information because the status cannot be obtained from SCC yet
+    // (e.g., whether and add-on is register or its registration code).
```
jreidinger, line 64: if needed, maybe we can somehow get that status, but it is not straightforward. It looks like there is list of products we use now and there is also list of "activations" where we can search for what is activated https://github.com/SUSE/connect-ng/blob/75cc6a6ef740809df16a840ab2b63141ace724d5/pkg/registration/product.go#L206
and it is somehow exported ( but looks like not used now ) - https://github.com/SUSE/connect-ng/blob/75cc6a6ef740809df16a840ab2b63141ace724d5/third_party/yast/lib/suse/connect/yast.rb#L20

rust/agama-software/src/model/registration.rs
```diff
@@ -56,15 +60,44 @@ pub struct Registration {
...
+    /// * `addon`: add-on to register.
+    pub fn register_addon(
+        &mut self,
+        zypp: &zypp_agama::Zypp,
+        addon: &Addon,
+    ) -> RegistrationResult<()> {
+        // Use the product's version as default.
+        let version = addon.version.clone().unwrap_or(self.version.clone());
+        let code = addon.code.as_ref().map(|c| c.as_str());
+        self.activate_product(&zypp, &addon.id, &version, code)?;
```
jreidinger, line 85: just note that this can produce some inconsistencies that is not recoverable. Like if activation succeed and then adding service failed ( or its refresh which is more common ), then you cannot again activate it. Not sure what is the best way how to recover from it.

rust/agama-software/src/model/registration.rs
```diff
@@ -105,6 +138,8 @@ impl Registration {
         // Add the libzypp service
         zypp.add_service(&service.name, &service.url)
             .map_err(|e| RegistrationError::AddService(service.name.clone(), e))?;
+        zypp.refresh_service(&service.name)
+            .map_err(|e| RegistrationError::RefreshService(service.name.clone(), e))?;
```
jreidinger: yeap, here is the most common issue with activated and then failed registration, which is not recoverable now. Maybe we can remember also service and add it to list of addons and during some retry try to refresh it again....or at least we should check if we show proper question, so user can retry that refresh if e.g. network glitch.

rust/agama-software/src/model/registration.rs
```diff
@@ -105,6 +138,8 @@ impl Registration {
...
         self.services.push(service);
```
jreidinger: not sure, but maybe we should add to services before trying to refresh it?

jreidinger (approved):
looks good, nothing critical

rust/agama-software/src/model/registration.rs
imobachgs: OK, it is not a problem right now because we can keep track of the status in Agama itself. Perhaps in the future.

rust/agama-software/src/model/registration.rs
imobachgs: Yes, you are right.

rust/agama-software/src/model/registration.rs
imobachgs: I would expect some callback if we cannot refresh the service. By now, I followed your suggestion to add the service to the list before trying to refreshing it.

PR #2990, opened 2026-01-07
fix downloading logs

jreidinger:
## Problem

download logs via CLI does not work and as result also copy of logs at the end of installation does not work.

## Solution

Use the same code as in manager web interface, just move it to internal part and adapt client to use it.

## Testing

- tested manually with CLI and it works. ENOTIME for testing at the end of installation.


rust/agama-lib/src/manager/http_client.rs
```diff
@@ -80,7 +80,7 @@ impl ManagerHTTPClient {
     /// Returns path to logs
     pub async fn store(&self, path: &Path) -> Result<PathBuf, ManagerHTTPClientError> {
         // 1) response with logs
-        let response = self.client.get_raw("/manager/logs/store").await?;
+        let response = self.client.get_raw("/v2/private/download_logs").await?;
```
imobachgs, line 83: I am not sure whether this should be part of the private API. However, until we take a decision, it is good to keep it in the private part.

Additionally, I would name it as `/v2/private/logs`.

imobachgs (commented):
In general, it looks good. I guess we still need to fix the UI.

rust/agama-lib/src/manager/http_client.rs
jreidinger: well, CLI has second param logs_list...not sure how much it is needed or useful, but I want to prevent ambitious names.

PR #2994, opened 2026-01-08
Allow changing the registration URL

imobachgs:
## Problem

The new registration stack always uses the SCC to register the system.

## Solution

Add support to specify the registration URL through:

* The `product.registration_url` element in the configuration.
* The `inst.register_url` kernel command-line argument.

## Testing

- Added a new unit test (partial)
- Tested manually

### DIFF
--- a/rust/agama-software/src/model/state.rs
+++ b/rust/agama-software/src/model/state.rs
@@ -28,8 +28,10 @@ use agama_utils::{
     api::software::{
         Config, PatternsConfig, ProductConfig, RepositoryConfig, SoftwareConfig, SystemInfo,
     },
+    kernel_cmdline::KernelCmdline,
     products::{ProductSpec, UserPattern},
 };
+use url::Url;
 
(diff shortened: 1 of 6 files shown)


rust/agama-software/src/model/state.rs
```diff
@@ -177,12 +188,21 @@ impl<'a> SoftwareStateBuilder<'a> {
             vec![]
         };
 
+        let url = match &config.registration_url {
+            Some(registration_url) => Some(registration_url.clone()),
+            None => self
+                .kernel_cmdline
+                .get_last("inst.register_url")
```
joseivanlopez, line 195: Not related to this PR, but it would be nice to use the same keys in both kernel and config options (register_url vs registration_url).

joseivanlopez (approved):
LGTM

rust/agama-software/src/model/state.rs
imobachgs: Yes, I agree. I think that in the past we chose `register_url` because it was shorter or something like that. Perhaps we could reconsider it.

PR #3006, opened 2026-01-13
Network startup install

teclator:
## Problem

Up to Agama 18, part of the network configuration was done using the [Agama::Network Ruby class](https://github.com/agama-project/agama/blob/master/service/lib/agama/network.rb "‌"). The method `startup` was called during Agama start-up and `install`during the installation process.

We need to implement the same logic in the `api-v2`branch using Rust.

Note: We should consider to copy all the connections from `/etc/NetworkManager/system-connections` as we used to do at least until Agama is aware of new connections added by NetworkManager. See[https://bugzilla.suse.com/show_bug.cgi?id=1254492#c25](https://bugzilla.suse.com/show_bug.cgi?id=1254492#c25 "smartCard-inline")

- https://trello.com/c/ixm9Fz0J/5460-add-support-to-initialize-and-finish-the-network-configuration

## Solution

Implemented the network startup and install methods in the backend network service.

## Tests

- Tested manually

### When starting the service
- [x] Propose default configuration when starting the service.

### When calling install
- [X] Copy connections to the target system.
- [X] Enable NetworkManager service
- [ ]  Configure the proxy in the target system. ?

### DIFF
--- a/rust/agama-network/src/model.rs
+++ b/rust/agama-network/src/model.rs
@@ -220,12 +220,14 @@ impl NetworkState {
             let entry = entry.map_err(|e| NetworkStateError::IoError(e.to_string()))?;
             let path = entry.path();
             if path.is_file() {
-                let file_name = path
-                    .file_name()
-                    .ok_or_else(|| NetworkStateError::IoError("Invalid file name".to_string()))?;
-                let dest = to.join(file_name);
-                std::fs::copy(&path, &dest)
-                    .map_err(|e| NetworkStateError::IoError(e.to_string()))?;
+                if let Some(file_name) = path.file_name() {
+                    let dest = to.join(file_name);
+                    if let Err(e) = std::fs::copy(&path, &dest) {
+                        tracing::error!("It was not possible to copy {:?}: {:?}", &file_name, e);
+                    }
+                } else {
+                    tracing::error!("The path {:?} looks wrong, skipping it.", &path);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


rust/agama-network/src/model.rs
```diff
@@ -160,6 +162,80 @@ impl NetworkState {
...
+        let mut command = process::Command::new("chroot");
+        command.args([path, "systemctl", "enable", "NetworkManager.service"]);
+
+        if let Some(output) = command.output().await.ok() {
+            if output.status.success() {
+                return Ok(());
+            }
+        }
+
+        tracing::error!("Error enabling NetworkManager service");
```
imobachgs, line 208: We are missing the error message here. But we can improve this later.

rust/agama-network/src/model.rs
```diff
@@ -160,6 +162,80 @@ impl NetworkState {
...
+
+        for entry in
+            std::fs::read_dir(from).map_err(|e| NetworkStateError::IoError(e.to_string()))?
+        {
+            let entry = entry.map_err(|e| NetworkStateError::IoError(e.to_string()))?;
+            let path = entry.path();
+            if path.is_file() {
+                let file_name = path
+                    .file_name()
+                    .ok_or_else(|| NetworkStateError::IoError("Invalid file name".to_string()))?;
```
imobachgs: np: in these cases, you can just log the error and continue with the rest.

rust/agama-network/src/service.rs
```diff
@@ -263,6 +263,28 @@ impl NetworkSystemClient {
         let (tx, rx) = oneshot::channel();
         self.actions.send(Action::Apply(tx))?;
         let result = rx.await?;
+        //self.events.send(event::Event::ProposalChanged {
```
imobachgs: left-over?

teclator: yep

PR #2937, opened 2025-12-11
Rewrite users service from ruby to rust

mchf:
## Problem

Current stable version of Agama is mixture of several languages and technologies. Users service was originally written in ruby. As we later on decided to prefer rust for various reasons, it makes sense to rewrite users service to rust.

## Solution

This is initial PR for transition of users service to rust. Currently handles GET, PUT, PATCH /config and GET /proposal. However, no install action yet.

## Testing

- *Tested manually*

PR #3007, opened 2026-01-13
Adapt the "authentication" section of the web UI to the new API

ancorgs:
## Problem

#2937 introduced a new API to configure authentication (root and the first non-root user). But the web interface is still not adapted

## Solution

This adapts the web interface to use the new API.

It also adds a new entry about Authentication for the installation summary, making it possible to reach the corresponding section of the UI.

## Hack alert

To make sure the UI is refreshed after discarding the first user, the page displaying that user gets the information from the proposal, instead of getting it from the configuration.

Once config is fixed in the backend to also emit events on every change that could be reverted. It is done in a separate commit so ease the task.

## Bonus

This pull request also include fixes for RegistrationSummary and StorageSumary

## Disclaimer

The `hasIssues` flag of `UsersSummary` is based on the existence of issues in the scope. Currently such issues are only generated when the service is started. If the configuration is then fixed (let's say a password is set for root) and broken again (let's say the root password is removed afterwards), then the issue will not be re-generated and the interface will not display the corresponding warning icon.

That needs to be fixed in the backend.

## Testing

- Unit tests adapted. Unit tests that were commented out were brought back to life.
- Tested manually

## Screenshots

users1b

users2b

users-error

### DIFF
--- a/web/src/components/overview/StorageSummary.tsx
+++ b/web/src/components/overview/StorageSummary.tsx
@@ -22,7 +22,6 @@
 
 import React from "react";
 import { sprintf } from "sprintf-js";
-import { isEmpty } from "radashi";
 import Summary from "~/components/core/Summary";
 import Link from "~/components/core/Link";
 import { useProgressTracking } from "~/hooks/use-progress-tracking";
@@ -156,15 +155,11 @@ const Description = () => {
  */
 export default function StorageSummary() {
   const { loading } = useProgressTracking("storage");
-  // FIXME: Refactor for avoid duplicating these checks about issues and actions
-  // TODO: extend tests for covering the hasIssues status
-  const actions = useActions();
-  const issues = useIssues("storage");
-  const configIssues = issues.filter((i) => i.class !== "proposal");
+  const hasIssues = !!useIssues("storage").length;
 
   return (
     <Summary
-      hasIssues={!isEmpty(configIssues) || isEmpty(actions)}
+      hasIssues={hasIssues}
       icon="hard_drive"
       title={
         <Link to={STORAGE.root} variant="link" isInline>


web/src/model/proposal/users.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) [2025] SUSE LLC
```
teclator:
```suggestion
 * Copyright (c) [2026] SUSE LLC
```

web/src/components/users/FirstUserForm.tsx
```diff
@@ -138,8 +135,7 @@ export default function FirstUserForm() {
       return;
     }
 
-    setFirstUser
-      .mutateAsync({ ...data })
+    patchConfig({ user: data })
```
teclator, line 138: Is the catch errors still valid? does it work correctly?
 
```suggestion
    patchConfig({ user: data })
```

web/src/model/proposal/users.ts
ancorgs: This is part of a commit that should be reverted soon. So I will fix this via `commit --amend` to ease that task.

web/src/components/users/FirstUserForm.tsx
ancorgs: Yes, it works. I will add a screenshot to probe it.

ancorgs: Screenshot added.

web/src/components/users/FirstUser.tsx
```diff
@@ -33,14 +33,22 @@ import {
 } from "@patternfly/react-core";
 import { Link, Page, SplitButton } from "~/components/core";
 import PasswordCheck from "~/components/users/PasswordCheck";
-import { useConfig } from "~/hooks/model/config";
+// This should be based on the config, not on the proposal. As a temporary hack (introduced in a
+// separate commit that should be easy to revert), we are using the proposal because the config
+// does not emit an event on every change.
+import { useProposal } from "~/hooks/model/proposal";
```
dgdavid, line 39: NP: not related specifically to your code here, but it is kind of a mess sometimes having ~/hooks/model/proposal/whatever and sometimes, like here, not. Hard to follow/understand.

web/src/components/overview/UsersSummary.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const { userName, fullName, password } = config.user;
+  return userName !== "" && fullName !== "" && password !== "";
+};
+
+/**
+ * Renders a summary text describing the authentication configuration.
+ */
+const Value = () => {
+  const config = useConfig();
```
dgdavid, line 54: NP: this is fully ok, but I wanted to mention that I tend to get the config in the main component and pass down via prop. I do so just to save some observers at TanstackQuery layer and because in this kind of simple, one level nesting components looks like a good compromise.

web/src/components/overview/RegistrationSummary.tsx
```diff
@@ -40,7 +40,7 @@ import { useIssues } from "~/hooks/model/issue";
 const Content = () => {
   const { registration } = useSystem();
   const issues = useIssues("software");
-  const hasIssues = issues.find((i) => i.class === "software.register_system") !== undefined;
+  const hasIssues = issues.find((i) => i.class === "software.missing_registration") !== undefined;
```
dgdavid, line 43: Thanks! 🙏

PR #3008, opened 2026-01-13
Adapt iSCSI D-Bus layer to API v2

joseivanlopez:
Adapt the D-Bus service to provide a new iSCSI API suitable to serve the new HTTP API.

Note: the new HTTP API for iSCSI will be implemented in a separate PR.

### DIFF
--- a/service/lib/agama/storage/iscsi/adapter.rb
+++ b/service/lib/agama/storage/iscsi/adapter.rb
@@ -34,7 +34,7 @@ module Agama
       # This class is a wrapper over the YaST code dealing with iSCSI.
       class Adapter
         # Performs actions for activating iSCSI.
-        def self.activate
+        def activate
           Yast::IscsiClientLib.getiBFT
           # Check initiator name, creating one if missing.
           return false unless Yast::IscsiClientLib.checkInitiatorName(silent: true)
--- a/service/lib/agama/dbus/storage_service.rb
+++ b/service/lib/agama/dbus/storage_service.rb
@@ -54,7 +54,7 @@ module Agama
         # Inhibits various storage subsystem (udisk, systemd mounts, raid auto-assembly) that
         # interfere with the operation of yast-storage-ng and libstorage-ng.
         Y2Storage::Inhibitors.new.inhibit
-        Agama::Storage::ISCSI::Adapter.activate
+        Agama::Storage::ISCSI::Adapter.new.activate
         check_multipath
         export
       end
--- a/service/test/agama/dbus/storage_service_test.rb
+++ b/service/test/agama/dbus/storage_service_test.rb
@@ -46,7 +46,7 @@ describe Agama::DBus::StorageService do
       .and_return(manager)
     allow(Agama::DBus::Storage::Manager).to receive(:new).with(manager, logger: logger)
       .and_return(manager_obj)
-    allow(Agama::Storage::ISCSI::Adapter).to receive(:activate)
+    allow_any_instance_of(Agama::Storage::ISCSI::Adapter).to receive(:activate)
   end
 
   describe "#start" do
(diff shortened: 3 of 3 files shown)


service/Gemfile
```diff
@@ -27,7 +27,7 @@ gemspec
 group :development do
   # undeclared dependency of rubocop
   gem "base64"
-  gem "byebug"
+  gem "debug"
```
ancorgs, line 30: Wouldn't this deserve a separate pull request with its own explanation?

joseivanlopez: IMHO, having a PR only for that is too much. I could add some comment if needed.

service/lib/agama/storage/iscsi/adapter.rb
```diff
@@ -34,12 +34,16 @@ module ISCSI
       # This class is a wrapper over the YaST code dealing with iSCSI.
       class Adapter
         # Performs actions for activating iSCSI.
-        def activate
+        def self.activate
```
ancorgs: Why is this a class method? I don't see a fundamental difference compared to the other methods.

service/lib/agama/storage/iscsi/manager.rb
```diff
@@ -253,117 +139,147 @@ def probe_initiator
         # @return [Array<ISCSI::Node>]
         def probe_nodes
           @nodes = adapter.read_nodes
+          # Targets are set as locked if they are already connected at the time of probing for first
```
ancorgs, line 142: NP: I find the usage of the terms "node" and "target" to be confusing all along this class.

If there is a difference, please state it clearly. 

If there is no difference, I would choose one and stick to that in the code. Not sure if using synonyms in variable and method names is a good idea.

service/lib/agama/storage/iscsi/adapter.rb
joseivanlopez: I guess the Adapter could be a module instead of a class.

service/lib/agama/storage/iscsi/manager.rb
joseivanlopez: Right now `Target` is used in the config and `Node` is what it is read from the system. Maybe we could call both as `Target`: `Agama::Storage::ISCSI::Target` and `Agama::Storage::ISCSI::Configs::Target`.

ancorgs: Well, I kind of remember we did put quite some thinking into choosing the term "node" for the `Agama::Storage::ISCSI::Node` class. There is even some note about nomenclature in the class description.

service/lib/agama/storage/iscsi/adapter.rb
ancorgs: Maybe, but I would keep it consistent for now anyways.

service/lib/agama/storage/iscsi/manager.rb
joseivanlopez: Nevertheless, the config (and the JSON) has `#targets` property. I am not sure :)

ancorgs: Yes. That was done to mimic AutoYaST tradition.

ancorgs: Since terminology was mess in YaST, the agreement back then was to use the names we considered more correct for the classes and so on, but keeping the term "target" in the profile and the UI for historical reasons.

joseivanlopez: I agree, although `Configs::Target` was not added by this PR. Do you suggest to rename it as `Configs::Node` ? I don't know what exactly to rename.

ancorgs: Never mind. I have reviewed this pull request more carefully and now I understand the rationale behind the naming.

To be honest, I would even say it is quite clear. So it is fully my fault that I messed up on the previous review because I focused too much on a few lines of code.

service/Gemfile
joseivanlopez: We agreed to keep this change as part of this PR.

PR #3012, opened 2026-01-14
Fix the installation progress and congratulations pages

imobachgs:
Adapt the installation progress and the "congratulations" page to the new HTTP API. It includes navigating when the stage changes, displaying the new progress, and implementing the "Reboot" button again.

Additionally, it fixes the download logs button-

## Testing

- Unit tests are adapted.
- Tested manually.

## Screenshots

Captura desde 2026-01-14 07-11-26

Captura desde 2026-01-14 07-22-08

### DIFF
--- a/rust/agama-autoinstall/src/main.rs
+++ b/rust/agama-autoinstall/src/main.rs
@@ -21,12 +21,8 @@
 use std::str::FromStr;
 
 use agama_autoinstall::{ConfigAutoLoader, ScriptsRunner};
-use agama_lib::{
-    auth::AuthToken,
-    http::BaseHTTPClient,
-    manager::{FinishMethod, ManagerHTTPClient},
-};
-use agama_utils::kernel_cmdline::KernelCmdline;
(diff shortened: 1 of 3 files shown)


web/src/model/manager.ts
```diff
@@ -52,7 +52,7 @@ const startInstallation = () => post("/api/v2/action", install_action);
 /**
  * Clean-up when installation is done.
  */
-const finishInstallation = () => post("/api/manager/finish");
+const finishInstallation = () => post("/api/v2/action");
```
jreidinger, line 55: why you do not send what action you perform here?

imobachgs: Sorry, this is old code. It is not used at all (I blindly adapted the URL). This file should disappear at some point.

imobachgs: The real call is in the `api.ts` file.

jreidinger (approved):

PR #3013, opened 2026-01-14
fix typo

bittin:
fix typo

### DIFF
--- a/web/src/components/core/LoginPage.tsx
+++ b/web/src/components/core/LoginPage.tsx
@@ -47,7 +47,7 @@ const getError = (authError) => {
   if (authError === AuthErrors.AUTH) {
     return {
       title: _("Could not log in"),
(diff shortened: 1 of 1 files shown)


web/src/components/core/LoginPage.tsx
```diff
@@ -47,7 +47,7 @@ const getError = (authError) => {
   if (authError === AuthErrors.AUTH) {
     return {
       title: _("Could not log in"),
-      description: _("Make suere that passsword is correct and try again."),
+      description: _("Make sure that passsword is correct and try again."),
```
imobachgs: I would say we are missing a "the" before the "password": "Make sure that the password is correct and try again."

bittin: yep

bittin: @imobachgs fixed

PR #3014, opened 2026-01-14
web: Drop product registration alert

dgdavid:
Agama already prevents users from starting the installation if any requirements are not met and informs them at the appropriate time. Therefore, continuously drawing the user's attention to product registration on every page during setup is overwhelming and could be disruptive. 

This PR removes the alert, as users will be properly notified about registration in the overview page, which holds the actual installation button that will be disabled if product is not registered since under the hood it is a  blocking issue.

Related to [bsc#1253708](https://bugzilla.suse.com/show_bug.cgi?id=1253708)


dgdavid:
> make sense for me. Maybe changelog entry?

As discussed in IRC, there will be a dedicated PR for adding changelog entries.

Let's merge this as it is.

PR #3022, opened 2026-01-14
Write user service config to installed system

mchf:
## Problem

So far setting users configuration was re-implemented to rust (https://github.com/agama-project/agama/pull/2937), but no data were written.

## Solution

Implemented Install message handler in user's service

## Testing

- *Tested manually*

## TODO:
- ~~set user's details (like full name).~~
- ~~store ssh keys for root~~

### DIFF
--- a/web/src/components/overview/InstallationSettings.tsx
+++ b/web/src/components/overview/InstallationSettings.tsx
@@ -29,6 +29,7 @@ import StorageSummary from "~/components/overview/StorageSummary";
 import NetworkSummary from "~/components/overview/NetworkSummary";
 import SoftwareSummary from "~/components/overview/SoftwareSummary";
 import RegistrationSummary from "~/components/overview/RegistrationSummary";
+import UsersSummary from "~/components/overview/UsersSummary";
 import { _ } from "~/i18n";
 
 import a11yStyles from "@patternfly/react-styles/css/utilities/Accessibility/accessibility";
@@ -54,6 +55,7 @@ export default function InstallationSummarySection() {
           <NetworkSummary />
           <SoftwareSummary />
           <StorageSummary />
+          <UsersSummary />
         </div>
       </NestedContent>
     </>
--- a/web/src/components/overview/RegistrationSummary.tsx
+++ b/web/src/components/overview/RegistrationSummary.tsx
@@ -40,7 +40,7 @@ import { useIssues } from "~/hooks/model/issue";
 const Content = () => {
   const { registration } = useSystem();
   const issues = useIssues("software");
-  const hasIssues = issues.find((i) => i.class === "software.register_system") !== undefined;
+  const hasIssues = issues.find((i) => i.class === "software.missing_registration") !== undefined;
 
   // TRANSLATORS: Brief summary about the product registration.
   // %s will be replaced with the last 4 digits of the registration code.
--- a/web/src/components/overview/StorageSummary.tsx
+++ b/web/src/components/overview/StorageSummary.tsx
@@ -22,7 +22,6 @@
 
 import React from "react";
 import { sprintf } from "sprintf-js";
-import { isEmpty } from "radashi";
 import Summary from "~/components/core/Summary";
 import Link from "~/components/core/Link";
 import { useProgressTracking } from "~/hooks/use-progress-tracking";
@@ -156,15 +155,11 @@ const Description = () => {
  */
 export default function StorageSummary() {
   const { loading } = useProgressTracking("storage");
-  // FIXME: Refactor for avoid duplicating these checks about issues and actions
-  // TODO: extend tests for covering the hasIssues status
-  const actions = useActions();
-  const issues = useIssues("storage");
-  const configIssues = issues.filter((i) => i.class !== "proposal");
+  const hasIssues = !!useIssues("storage").length;
 
   return (
     <Summary
-      hasIssues={!isEmpty(configIssues) || isEmpty(actions)}
+      hasIssues={hasIssues}
       icon="hard_drive"
       title={
         <Link to={STORAGE.root} variant="link" isInline>
--- a/web/src/components/overview/UsersSummary.tsx
+++ b/web/src/components/overview/UsersSummary.tsx
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) [2026] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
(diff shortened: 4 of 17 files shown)


rust/agama-users/src/model.rs
```diff
@@ -33,7 +53,81 @@ pub trait ModelAdapter: Send + 'static {
...
+        };
+        let Some(ref user_password) = user.password else {
+            return Err(service::Error::MissingUserData);
+        };
+
+        let useradd = Command::new("/usr/sbin/useradd").arg(user_name).output()?;
+
+        if !useradd.status.success() {
+            tracing::error!("User {} creation failed", user_name);
+            return Err(service::Error::IO(std::io::Error::other(format!(
```
imobachgs: Perhaps we should have an specific error for this.

rust/agama-users/src/model.rs
```diff
@@ -33,7 +53,81 @@ pub trait ModelAdapter: Send + 'static {
...
+        // push user name and password into the pipe
+        if let Some(mut stdin) = passwd_process.stdin.take() {
+            writeln!(stdin, "{}:{}", user_name, user_password.password);
+        }
+
+        // proceed with the result
+        let passwd = passwd_process.wait_with_output()?;
+
+        if !passwd.status.success() {
+            tracing::error!("Setting passeord for user {} creation failed", user_name);
```
imobachgs:
```suggestion
            tracing::error!("Setting password for user {} creation failed", user_name);
```

rust/agama-users/src/service.rs
```diff
@@ -204,3 +210,18 @@ impl MessageHandler<message::GetProposal> for Service {
...
     }
 }
+
+#[async_trait]
+impl MessageHandler<message::Install> for Service {
+    async fn handle(&mut self, _message: message::Install) -> Result<(), Error> {
+        let Some(proposal) = self.get_proposal() else {
+            tracing::error!("Missing user proposal");
+
+            return Err(Error::MissingProposal);
```
imobachgs: This should not happen, right? In would most probably only log the error (instead of halting the whole installation).

imobachgs (commented):
In general it is looking good. Thanks!

rust/agama-users/src/model.rs
imobachgs: But I can live with this, not a problem.

rust/agama-users/src/service.rs
mchf: I originally had only error log here ... returning the Err was copied from agama-l10n service

imobachgs: And I think it is wrong too :-)

rust/agama-users/src/model.rs
mchf: tried to do so ...

rust/agama-users/src/service.rs
mchf: no problem ... deleting it (in users)

rust/agama-users/src/model.rs
```diff
@@ -33,7 +64,135 @@ pub trait ModelAdapter: Send + 'static {
...
+    /// Reads first user's data from given config and updates its setup accordingly
+    fn add_first_user(&self, user: &FirstUserConfig) -> Result<(), service::Error> {
+        let Some(ref user_name) = user.user_name else {
+            return Err(service::Error::MissingUserData);
+        };
+        let Some(ref user_password) = user.password else {
+            return Err(service::Error::MissingUserData);
+        };
+
+        let useradd = Command::new("/usr/sbin/useradd").arg(user_name).output()?;
```
imobachgs, line 87: We need to reach an agreement on whether we use the full path or not. :-/

rust/agama-users/src/model.rs
```diff
@@ -33,7 +64,135 @@ pub trait ModelAdapter: Send + 'static {
...
+        // push user name and password into the pipe
+        if let Some(mut stdin) = passwd_process.stdin.take() {
+            writeln!(stdin, "{}:{}", user_name, user_password.password)?;
+        }
+
+        // proceed with the result
+        let passwd = passwd_process.wait_with_output()?;
+
+        if !passwd.status.success() {
+            tracing::error!("Setting password for user {} creation failed", user_name);
```
imobachgs: The text looks wrong ("Setting password for user blah creation failed").
```suggestion
            tracing::error!("Setting password for user {} failed", user_name);
```

Or perhaps something like: "Failed to set password for user {}".

rust/agama-users/src/model.rs
```diff
@@ -33,7 +64,135 @@ pub trait ModelAdapter: Send + 'static {
...
+                "Cannot set password for user {}: {}",
+                user_name, passwd.status
+            )));
+        }
+
+        Ok(())
+    }
+
+    /// Updates root's authorized_keys file with SSH key
+    fn update_authorized_keys(&self, ssh_key: &String) -> Result<(), service::Error> {
```
imobachgs:
```suggestion
    fn update_authorized_keys(&self, ssh_key: &str) -> Result<(), service::Error> {
```

rust/agama-users/src/model.rs
```diff
@@ -33,7 +64,135 @@ pub trait ModelAdapter: Send + 'static {
...
+
+        Ok(())
+    }
+
+    /// Sets password for given user name
+    ///
+    /// echo "<user_name>:<password>" | chpasswd
+    fn set_user_password(
+        &self,
+        user_name: &String,
```
imobachgs:
```suggestion
        user_name: &str,
```

rust/agama-users/src/service.rs
```diff
@@ -204,3 +210,16 @@ impl MessageHandler<message::GetProposal> for Service {
...
     }
 }
+
+#[async_trait]
+impl MessageHandler<message::Install> for Service {
+    async fn handle(&mut self, _message: message::Install) -> Result<(), Error> {
+        if let Some(proposal) = self.get_proposal() {
+            self.model.install(&proposal)?;
+        } else {
+            tracing::error!("Missing user proposal");
```
imobachgs: What is really missing is the authentication configuration. The "proposal" is going to disappear starting next week :-)

rust/agama-users/src/model.rs
mchf: usual pain ... I put full path here because it was so in last version of yast2-users i have. No other reason for that.

rust/agama-users/src/model.rs
mchf: yeah ... i'm guilty by copy pasting lines of code ;-)

rust/agama-users/src/model.rs
imobachgs: Yes, sure, we need to decide what is the best approach. I find it better to not use the full path so you can fake it during testing. But I am not sure whether that's a good idea at all.

imobachgs (approved):

PR #3024, opened 2026-01-14
web: add summary for registration

dgdavid:
Add a short summary for registration area in the overview page.

The PR includes needed changes for improving how areas with issues are rendered in the overview page. As an example, check screenshot below to see two situation on which storage area is "issues mode"

| Invalid settings | Failed to calculate layout |
|-|- |
| localhost_8080_ (49) | localhost_8080_ (50) |

### DIFF
--- a/web/src/components/overview/StorageSummary.tsx
+++ b/web/src/components/overview/StorageSummary.tsx
@@ -22,7 +22,7 @@
 
 import React from "react";
 import { sprintf } from "sprintf-js";
-import { HelperText, HelperTextItem } from "@patternfly/react-core";
+import { isEmpty } from "radashi";
 import Summary from "~/components/core/Summary";
 import Link from "~/components/core/Link";
 import { useProgressTracking } from "~/hooks/use-progress-tracking";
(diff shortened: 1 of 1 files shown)


teclator (approved):
There is no screenshot, but LGTM, although it is out of date with master branch but I guess there will be no conflicts.

web/src/components/overview/StorageSummary.tsx
```diff
@@ -160,9 +156,15 @@ const Description = () => {
  */
 export default function StorageSummary() {
   const { loading } = useProgressTracking("storage");
+  // FIXME: Refactor for avoid duplicating these checks about issues and actions
+  // TODO: extend tests for covering the hasIssues status
+  const actions = useActions();
+  const issues = useIssues("storage");
+  const configIssues = issues.filter((i) => i.class !== "proposal");
```
ancorgs, line 163: Well, in fact, we should not filter out issues by class here. All issues of the scope "storage" block the installation, so all of them deserve the `hasIssues`.

ancorgs: By doing so, you will save checking for actions.

ancorgs: So this is overcomplicated, but it actually covers both situations. So it can be merged as-is and fixed afterwards.

PR #3026, opened 2026-01-14
Fixes for the users service

imobachgs:
Fixes for the users service

- Improve the detection of missing information for the root and the first user.
- Crate the first user and/or set the root password/public key at the end of the installation.


dgdavid (approved):
LGTM

dgdavid (approved):

PR #3029, opened 2026-01-15
Use the YaST arch to filter the list of repos

imobachgs:
## Problem

The recently introduced [Arch](https://github.com/agama-project/agama/blob/master/rust/agama-utils/src/arch.rs) enum uses the following ids to detect each architecture: "aarch64", "ppc64le", "s390x" and "x86_64". These ids come from the `uname -m` command are used, for instance, in OBS.

However, we use different ids in the products specs: "aarch64", "ppc", "s390" and "x86_64" (coming from the YaST code).

## Solution

Instead of changing the product's definition, because it might affect the storage code, maps the new ids to the YaST ones.

## Testing

- *Added a new unit test*

### DIFF
--- a/rust/agama-utils/src/arch.rs
+++ b/rust/agama-utils/src/arch.rs
@@ -49,6 +49,16 @@ impl Arch {
             .try_into()
             .map_err(|_| Error::Unknown(arch_str))
     }
+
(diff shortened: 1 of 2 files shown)


ancorgs (approved):
LGTM

PR #3030, opened 2026-01-15
Add first user into wheel group

mchf:
## Problem

First user is requested to be the member of the wheel group

## Solution

Users service adds the first user into wheel group on creation

### DIFF
--- a/rust/agama-users/src/model.rs
+++ b/rust/agama-users/src/model.rs
@@ -69,11 +69,7 @@ impl Model {
 
         let useradd = self
             .chroot_command()
-            .args([
-                "useradd",
-                "-G",
-                "wheel",
-                &user_name])
+            .args(["useradd", "-G", "wheel", &user_name])
             .output()?;
 
         if !useradd.status.success() {


rust/agama-users/src/model.rs
```diff
@@ -72,6 +80,15 @@ impl Model {
             install_dir: PathBuf::from(install_dir.as_ref()),
         }
     }
+
+    /// Wrapper for creating Command which works in installation chroot
+    fn chroot_command(&self) -> Command {
+        let mut cmd = Command::new("chroot");
+
```
imobachgs, line 55: why

those

empty 

lines? ;-)

rust/agama-users/src/model.rs
```diff
@@ -109,7 +126,7 @@ impl ModelAdapter for Model {
         }
 
         self.set_user_password(user_name, user_password)?;
-
+        self.update_user_groups(user_name, vec!["wheel"])?;
```
imobachgs: I think you can add the user to a group while creating it by just adding the `--groups` argument. So you do not need to run another command.

rust/agama-users/src/model.rs
```diff
@@ -59,6 +59,14 @@ pub trait ModelAdapter: Send + 'static {
     fn update_user_fullname(&self, _user: &FirstUserConfig) -> Result<(), service::Error> {
         Ok(())
     }
+
+    fn update_user_groups(
```
imobachgs: Now that we have a little bit of time, we can talk about the design of this trait. IMHO, its API contains too many details. I would expect the trait to just define an `install` method that receives the configuration and that's it.

The details about how to apply the configuration to the underlying system should not be exposed through the API.

rust/agama-users/src/model.rs
```diff
@@ -72,6 +80,15 @@ impl Model {
             install_dir: PathBuf::from(install_dir.as_ref()),
         }
     }
+
+    /// Wrapper for creating Command which works in installation chroot
+    fn chroot_command(&self) -> Command {
```
imobachgs, line 53: This is a good idea. Thinking about it, and having into account that there is some repetitive code when running the commands, perhaps we could even think about having our own struct.

Something like:

```rust
struct ChrootedCommand(Command);
```

And you could do:

```rust
ChrootedCommand::on("/mnt")
  .command("adduser")
  .arg("blah")
  .output()?; // we could even have a more convenient way of running it if we want to
```

It is just an idea. WDYT? We could even reuse it in other parts of Agama.

rust/agama-users/src/model.rs
mchf: well just my way of formatting .. kind of separates things into logical blocks

rust/agama-users/src/model.rs
mchf: right ... I don't know why I have schema 
```
useradd
usermod
```
in my head

rust/agama-users/src/model.rs
mchf: ok ... I'll shuffle the code a bit.

PR #3031, opened 2026-01-15
Fixed the copy os the hostname to the target system

teclator:
## Problem

Currently the hostname is not copied correctly to the target system

## Solution

Corrected the join of the hostname path with the target dir copying it correctly.


rust/agama-hostname/src/model.rs
```diff
@@ -107,7 +107,7 @@ impl ModelAdapter for Model {
         const HOSTNAME_PATH: &str = "/etc/hostname";
```
imobachgs, line 107: Why not just use `"etc/hostname"` here? Well, disadvantage is that if someone changes the path he/she could introduce the same problem again.

PR #3034, opened 2026-01-16
Removed old users

mchf:
Removed pieces of old users api

## Refactoring todo as part of the cleanup
- ~~struct for ChrootCommand (kind of transparent wrapper for Command)~~
- https://github.com/agama-project/agama/pull/3057#pullrequestreview-3696673521

### DIFF
--- a/rust/agama-users/src/model.rs
+++ b/rust/agama-users/src/model.rs
@@ -44,10 +44,17 @@ pub trait ModelAdapter: Send + 'static {
 /// in chrooted environment is passed as an argument.
 ///
 /// Example use:
-/// ChrootCommand::new(<chroot directory>)
-///   .cmd(<command to be run in chroot env>)
-///   .args(["the command", "args"]
-struct ChrootCommand {
+/// ```
+/// # use agama_users::ChrootCommand;
+/// # use agama_users::service;
+/// # fn main() -> Result<(), service::Error> {
+/// let cmd = ChrootCommand::new("/tmp".into())?
+///   .cmd("echo")
+///   .args(["Hello world!"]);
+/// # Ok(())
+/// # }
+/// ```
+pub struct ChrootCommand {
     command: Command,
 }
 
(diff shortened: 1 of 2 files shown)


rust/agama-users/src/model.rs
```diff
@@ -37,6 +38,41 @@ pub trait ModelAdapter: Send + 'static {
     }
 }
 
+struct ChrootCommand {
```
imobachgs: Please, add some documentation here. An example about how to use it would be nice.

rust/agama-users/src/model.rs
```diff
@@ -37,6 +38,41 @@ pub trait ModelAdapter: Send + 'static {
...
+    command: Command,
+}
+
+impl ChrootCommand {
+    pub fn new(chroot: PathBuf) -> Self {
+        // TODO: check if the dir exists?
```
imobachgs: Please, do :-)

rust/agama-users/src/model.rs
```diff
@@ -38,18 +38,32 @@ pub trait ModelAdapter: Send + 'static {
     }
 }
 
+/// A wrapper for common std::process::Command for use in chroot
+///
+/// It basically creates Command for chroot and command to be run
+/// in chrooted environment is passed as an argument.
+///
+/// Example use:
+/// ChrootCommand::new(<chroot directory>)
```
imobachgs: Sorry, I meant https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html. You can mark the doctest to do not get executed using "```no_run". But the point of using this kind of examples is that they are compiled so we are aware if they get outdated.

imobachgs (commented):
Please, use a documentation test for the example. We use them in several places.

Otherwise, it LGTM.

rust/agama-users/src/model.rs
mchf: yeah ... definitely something new for me ;-) As usual I learn tests the last ;-)

PR #3037, opened 2026-01-16
Autoinstall fix

jreidinger:
## Problem

inst.auto parameter does not work due to calls to old API.

## Solution

Adapt it and also also proper waiting as HTTP calls do not wait unlike DBuS ones.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 16 15:01:32 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix unattended installation
+- Add first user to wheel group
+- Implement using inst.install_url instead of Product repositories
+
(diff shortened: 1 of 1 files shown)


rust/agama-autoinstall/src/main.rs
```diff
@@ -68,8 +72,27 @@ async fn main() -> anyhow::Result<()> {
...
+    }
+
     manager_client.install().await?;
 
+    // wait till install is done.
+    loop {
+        sleep(Duration::from_secs(1)).await;
+        let status = manager_client.status().await?;
+        // TODO: what should we do in case of failure? try reboot or stuck in inst env?
+        if status.stage == Stage::Finished || status.stage == Stage::Failed {
```
imobachgs: If the installation failed, I would rather prefer to print a message and exit.

PR #3044, opened 2026-01-19
fix: disable query refetch

joseivanlopez:
## Problem

By default, tanstack refetches queries, see https://tanstack.com/query/v5/docs/framework/react/guides/important-defaults. This makes the UI to perform unnecessary calls to the backend.  

## Solution

Globally disable query refetch.

### DIFF
--- a/web/src/hooks/model/issue.ts
+++ b/web/src/hooks/model/issue.ts
@@ -29,6 +29,8 @@ import type { Issue, Scope } from "~/model/issue";
 const issuesQuery = {
   queryKey: ["issues"],
   queryFn: getIssues,
+  /* FIXME: disable refetch. */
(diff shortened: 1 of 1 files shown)


web/src/context/app.tsx
```diff
@@ -53,6 +53,13 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       ...sharedOptions,
+      /**
+       * Disable refetch data.
+       * @see https://tanstack.com/query/v5/docs/framework/react/guides/important-defaults
+       */
+      refetchOnMount: false,
+      refetchOnReconnect: false,
```
imobachgs: I am not sure about this one.

PR #3053, opened 2026-01-20
Copy registration files at the end of the installation

jreidinger:
## Problem

Registration does not have its configuration neither credentials after reboot.

PBI: https://trello.com/c/fNM6lTtk/5496-copy-registration-configuration-and-credentials-to-the-final-system

## Solution

Copy the configuration file and the credentials.

## Testing

- Tested manually.
- IBS repo: https://build.suse.de/project/show/home:jreidinger:registration_finish

Scenarios:

- [x] simple reboot after registration
- [ ] RMT
- [x] register with addons

### DIFF
--- a/rust/agama-software/src/model/registration.rs
+++ b/rust/agama-software/src/model/registration.rs
@@ -201,7 +201,7 @@ impl Registration {
         }
     }
 
-    /// Writes to target system all registration configuration that is needed
+    /// Writes to the target system all the registration configuration that is needed
     ///
     /// Beware that, if a certificate was imported, it is copied by the agama-security service.
     pub fn finish(&mut self, install_dir: &Utf8PathBuf) -> Result<(), RegistrationError> {
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Fri Jan 23 13:47:29 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 
-- Copy SCC credentials an libzypp services (gh#agama-project/agama#3053).
+- Copy the SCC credentials and the libzypp services 
+  (gh#agama-project/agama#3053).
 
 -------------------------------------------------------------------
 Fri Jan 23 09:53:37 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>


rust/agama-software/src/model/registration.rs
```diff
@@ -189,6 +194,27 @@ impl Registration {
...
+        Ok(())
+    }
+
+    fn copy_files(&self) -> Result<(), RegistrationError> {
+        for path in &self.config_files {
+            tracing::info!("Copying credentials file {path:?}");
+            let target = Utf8PathBuf::from("/mnt");
+            let target_path = target.join(path);
+            std::fs::copy(path, target_path)
+                .map_err(|e| RegistrationError::IO(path.to_string(), e));
```
jreidinger: this line is missing `?` at the end, so maybe fs::copy failed in my test run.

rust/agama-software/src/model/registration.rs
```diff
@@ -189,6 +194,27 @@ impl Registration {
...
+            .push(suseconnect_agama::DEFAULT_CONFIG_FILE.into());
+        self.copy_files()?;
+        Ok(())
+    }
+
+    fn copy_files(&self) -> Result<(), RegistrationError> {
+        for path in &self.config_files {
+            tracing::info!("Copying credentials file {path:?}");
+            let target = Utf8PathBuf::from("/mnt");
+            let target_path = target.join(path);
```
jreidinger: here in target path we need to strip `/run/agama/software_ng_zypp` prefix from credentials.

rust/agama-software/src/model/registration.rs
```diff
@@ -189,6 +194,27 @@ impl Registration {
...
+            std::fs::copy(path, target_path)
+                .map_err(|e| RegistrationError::IO(path.to_string(), e));
+        }
+
```
jreidinger, line 237: and here we probably should add that copy of repositories and credentials.

service/lib/agama/storage/bootloader.rb
```diff
@@ -127,6 +127,8 @@ def initialize(logger)
       # It proposes the bootloader configuration based on the current system and storage
       # configuration. It also applies the user configuration and installs the needed packages.
       def configure
+        # TODO: get value from product ( probably for TW and maybe Leap?)
+        ::Bootloader::OsProber.package_available = false
```
jreidinger, line 131: we need to update dependency on yast2-bootloader for  this change - https://github.com/yast/yast-bootloader/pull/730/files#diff-30cde0ee6efd85042b30b6654bc0be7ab111ed3667d79a5240ff58150e730b80R20

rust/agama-software/src/zypp_server.rs
```diff
@@ -499,8 +499,15 @@ impl ZyppServer {
...
-    fn registration_finish(&self) -> ZyppServerResult<()> {
-        // TODO: implement when registration is ready
+    fn registration_finish(&mut self) -> ZyppServerResult<()> {
+        let Some(registration) = &mut self.registration else {
+            return Ok(());
+        };
+
+        if let Err(error) = registration.finish() {
+            // just log error and continue as registration config is recoverable
+            tracing::error!("Failed to finish the registration: {error}");
```
imobachgs, line 541: I would improve the description. What "finish the registration" means?

rust/agama-software/src/model/registration.rs
```diff
@@ -189,6 +194,27 @@ impl Registration {
...
+        self.config_files
+            .push(suseconnect_agama::DEFAULT_CONFIG_FILE.into());
+        self.copy_files()?;
+        Ok(())
+    }
+
+    fn copy_files(&self) -> Result<(), RegistrationError> {
+        for path in &self.config_files {
+            tracing::info!("Copying credentials file {path:?}");
+            let target = Utf8PathBuf::from("/mnt");
```
imobachgs: I do not like having references to "/mnt" here and there. IMHO, the ZyppServer struct could have an `install_dir` field (as we do in other places) and pass its value to the "finish" function.

rust/agama-software/src/model/registration.rs
imobachgs: And I wonder if we should change that path to something like `/run/agama/zypp` (now that we have finished the switch to the new software service).

rust/agama-software/src/zypp_server.rs
```diff
@@ -499,8 +499,15 @@ impl ZyppServer {
...
     }
 
-    fn registration_finish(&self) -> ZyppServerResult<()> {
-        // TODO: implement when registration is ready
+    fn registration_finish(&mut self) -> ZyppServerResult<()> {
+        let Some(registration) = &mut self.registration else {
+            return Ok(());
+        };
+
+        if let Err(error) = registration.finish() {
```
imobachgs: np: The name `finish` makes more sense from the service point of view. I wonder if using something like `save` or something like that. But, to be honest, I am not able to come up with a better name :-)

rust/package/agama.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Jan 23 13:47:29 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Copy SCC credentials an libzypp services (gh#agama-project/agama#3053).
```
```suggestion
- Copy the SCC credentials and the libzypp services 
  (gh#agama-project/agama#3053).
```

rust/agama-software/src/zypp_server.rs
teclator: Still pending?

rust/agama-software/src/model/registration.rs
```diff
@@ -195,6 +201,43 @@ impl Registration {
         }
     }
 
+    /// Writes to target system all registration configuration that is needed
```
teclator:
```suggestion
    /// Writes to the target system all the registration configuration that is needed
```

rust/agama-software/src/zypp_server.rs
jreidinger: It is questionable as we can discover more internals like that some copy of configuration failed and try to have it more highlevel pretending we do not know what exactly happens in finish phase...So maybe something like `Failed to complete registration configuration on target system` ?

PR #3054, opened 2026-01-20
service: Honor read_only flag from the product configuration

ancorgs:
## Problem

The product configuration allows to setup the "/" volume as read-only. That should result in an immutable system.

But that setting is currently ignored by the Agama storage proposal. The code was commented out temporarily in order to decide how exactly we wanted to manage immutable systems... but that conversation didn't happen and the code has remained commented until today.

See https://bugzilla.suse.com/show_bug.cgi?id=1253616

## Solution

Honor the read-only flag.

Clarify the current status of the flag in the code.

## Testing

Tested manually installing openSUSE MicroOS with a patched installer.

Everything works as expected:

```
$ cat /etc/fstab | head -n1
UUID=xxx-yyy-zzz  /  btrfs  ro=vfs  0  0

$ btrfs property get -ts /
ro=true
```

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jan 21 10:24:50 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Honor the read_only flag for volumes coming from the product
+  configuration (bsc#1253616).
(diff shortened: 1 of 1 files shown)


ancorgs:
I double checked the network issue by installing Tumbleweed with the same installer (but without the patch). Same result (no NetworkManager started), so I'm sure it was not caused by this patch.

PR #3055, opened 2026-01-21
Import self-signed certificates during registration

imobachgs:
## Problem

It is not possible to register the system with a registration server that uses a self-signed certificate.

## Solution

Introduce a new `agama-security` service which should take care of handling security-related tasks. At this point it only implements two actions:

* `CheckCertificate`: asks the user whether to trust a certificate. If the user accepts, imports the certificate.
* `Finish`: copies the imported certificates to the installed system.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>
<summary>Asking to import a certificate</summary>

Captura desde 2026-01-21 10-35-00
</details>

## Follow-up

* Fix several problems in the registration UI.
* Use `agama-security` to import GPG keys.

### DIFF
--- a/rust/agama-security/src/service.rs
+++ b/rust/agama-security/src/service.rs
@@ -18,7 +18,10 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
-use std::path::{Path, PathBuf};
+use std::{
+    path::{Path, PathBuf},
+    process,
+};
 
 use agama_utils::{
     actor::{self, Actor, Handler, MessageHandler},
@@ -39,6 +42,10 @@ pub enum Error {
     Actor(#[from] actor::Error),
     #[error("Could not write the certificate: {0}")]
     CertificateIO(#[source] std::io::Error),
+    #[error("Could not update the certificates database: {0}")]
+    CaCertificates(String),
+    #[error(transparent)]
+    IO(#[from] std::io::Error),
 }
 
 pub struct Starter {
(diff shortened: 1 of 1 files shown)


rust/agama-security/src/service.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Beware that it does not remove the already imported certificates.
+    pub fn reset(&mut self) {
+        self.trusted.clear();
+    }
+
+    /// Copy the certificates to the given directory.
+    ///
+    /// * `directory`: directory to copy the certificates.
+    pub fn copy_certificates(&self, directory: &Path) {
+        let workdir = self.workdir.strip_prefix("/").unwrap_or(&self.workdir);
```
jreidinger, line 167: sadly copy is not so straightforward :(
Original code is at https://github.com/agama-project/agama/blob/SLE-16/service/lib/agama/registration.rb#L278
so it copy to specific file name ( not sure if it is mandatory ) https://github.com/SUSE/connect-ng/blob/main/third_party/yast/lib/suse/connect/ssl_certificate.rb#L8
and calls update-ca-certificates in chroot as seen at https://github.com/SUSE/connect-ng/blob/main/third_party/yast/lib/suse/connect/yast.rb#L44 which do whatever

imobachgs: OK, so we are only missing the call to `update-ca-certificates`, as we already do for the inst-sys when importing it.

imobachgs: Actually, there is a `ca-certificates.path` and a `ca-certificates.service` in systemd, but I do not know when it runs.

rust/agama-software/src/zypp_server.rs
```diff
@@ -679,15 +712,37 @@ impl ZyppServer {
...
-            self.register_base_system(state, zypp, issues);
-        }
+        match &self.registration {
+            RegistrationStatus::Failed(error) => {
+                issues.push(
+                    Issue::new(
+                        "software.register_system",
+                        &gettext("Failed to register the system"),
+                    )
+                    .with_details(&error.to_string()),
```
jreidinger, line 738: usability note. Details is usually what contain real reason why it failed and I do not see it in UI now.

imobachgs: No, the UI needs quite some fixes, I am non it (different PR).

PR #3056, opened 2026-01-21
Fix enable NetworkManager service in target system

teclator:
## Problem

In #3006 the startup and install methods were added to network but there was an error in the path to enable the NetworkManager service enabling it in the installation systemd instead of the target system.

## Solution

Fixed the enable of the service passing the `target_dir` to the enable service command.

### DIFF
--- a/rust/agama-network/src/model.rs
+++ b/rust/agama-network/src/model.rs
@@ -202,11 +202,11 @@ impl NetworkState {
         match command.output().await {
             Ok(output) => {
                 if !output.status.success() {
-                    tracing::error!("Failed to enable NetworkManager service: {output:?}")
+                    tracing::error!("Failed to enable the NetworkManager service: {output:?}")
                 }
             }
             Err(error) => {
-                tracing::error!("Failed to run the enable NetworkManager service command: {error}");
+                tracing::error!("Failed to run the command to enable the NetworkManager service command: {error}");
             }
         }
 


rust/agama-network/src/model.rs
```diff
@@ -192,20 +192,24 @@ impl NetworkState {
...
-            if output.status.success() {
-                return Ok(());
+        match command.output().await {
+            Ok(output) => {
+                if !output.status.success() {
+                    tracing::error!("Failed to enable NetworkManager service: {output:?}")
+                }
+            }
+            Err(error) => {
+                tracing::error!("Failed to run the enable NetworkManager service command: {error}");
```
imobachgs:
```suggestion
                tracing::error!("Failed to run the command to enable the NetworkManager service command: {error}");
```

rust/agama-network/src/model.rs
```diff
@@ -192,20 +192,24 @@ impl NetworkState {
...
         let mut command = process::Command::new("chroot");
         command.args([path, "systemctl", "enable", "NetworkManager.service"]);
 
-        if let Some(output) = command.output().await.ok() {
-            if output.status.success() {
-                return Ok(());
+        match command.output().await {
+            Ok(output) => {
+                if !output.status.success() {
+                    tracing::error!("Failed to enable NetworkManager service: {output:?}")
```
imobachgs:
```suggestion
                    tracing::error!("Failed to enable the NetworkManager service: {output:?}")
```

PR #3057, opened 2026-01-21
Returns error status if asked for install action and backend detects issues

mchf:
## Problem

It is possible to start installation via CLI / API even when configuration in the backend contains issues.

## Solution

Send reply with status 422 in such case

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-server/src/server/web.rs
+++ b/rust/agama-server/src/server/web.rs
@@ -68,7 +68,7 @@ impl IntoResponse for Error {
             "error": self.to_string()
         });
         let status = if matches!(self, Error::Manager(InstallationBlocked)) {
-            StatusCode::METHOD_NOT_ALLOWED
+            StatusCode::UNPROCESSABLE_ENTITY
         } else {
             StatusCode::BAD_REQUEST
         };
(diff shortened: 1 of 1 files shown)


mchf:
> ## Problem
> It is possible to start installation via CLI / API even when configuration in the backend contains issues.
> 
> ## Solution
> Send reply with status 422 in such case
> 
> ## Testing
> TODO:
> 
> * _Tested manually_

rust/agama-manager/src/service.rs
```diff
@@ -686,6 +688,13 @@ impl MessageHandler<message::GetLicense> for Service {
 impl MessageHandler<message::RunAction> for Service {
     /// It runs the given action.
     async fn handle(&mut self, message: message::RunAction) -> Result<(), Error> {
+        let issues = self.issues.call(issue::message::Get).await?;
+        let progress = self.progress.call(progress::message::GetProgress).await?;
```
imobachgs, line 692: We recently introduced a new action to check whether the progress is empty or not. See https://github.com/agama-project/agama/blob/dce005304f3916a9a9500b1ed8f4b10b4d2786d7/rust/agama-manager/src/service.rs#L524.

Perhaps you want to give it a try (it is available in `master`).

PR #3062, opened 2026-01-22
service: Catch expected bootloader exception

ancorgs:
## Problem

A change introduced lately causes the bootloader proposal to be re-calculated more often.

It is even attempted in situations in which there is not a valid storage proposal, causing the backend error reported at https://bugzilla.suse.com/show_bug.cgi?id=1257013

## Solution

Ignore the failure to calculate the bootloader proposal if there is no storage proposal.

If there is no storage proposal then the installation is blocked with an issue and the bootloader proposal is irrelevant. So such a bootloader failure is kind of expected not a problem.

## Testing

Tested manually with the currently available alpha images of SLE.

I was no able to reproduce the error with more recent versions of the installer (at least with openSUSE based ones), nor in my development machine.

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jan 22 12:50:32 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Stop reporting a bootloader error when there is no valid storage
+  proposal (bsc#1257013).
+
+-------------------------------------------------------------------
 Wed Jan 21 10:24:50 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
 
 - Honor the read_only flag for volumes coming from the product


service/lib/agama/storage/bootloader.rb
```diff
@@ -139,9 +139,10 @@ def configure
         write_config
         # and set packages needed for given config
         install_packages
-        # TODO: error handling (including catching exceptions like Bootloader::NoRoot)
-        # and filling issues
+        # TODO: error handling (including catching exceptions and filling issues)
```
mvidner, line 142: Devil's advocate: When [`::Bootloader::SomethingElse`](https://github.com/search?q=repo%3Ayast%2Fyast-bootloader%20raise&type=code) is raised, we will get a very similar bug report to this one.
Do we want to 
- `rescue` and convert to an Issue
- improve this comment, listing some Bootloader exceptions and explaining why propagating the crash is acceptable?

mvidner (approved):
The explanation and fix look reasonable.

I just have concerns that very similar bugs may still slip through. The TODO told me :)

service/lib/agama/storage/bootloader.rb
ancorgs: Yes. We need to decide how to handle all other kind of exceptions. That's why the TODO remains.

I just rescued the one that is actually known to not be a real problem.

PR #3064, opened 2026-01-22
Add iSCSI to HTTP API

joseivanlopez:
Integrate iSCSI management in the new HTTP API.

The D-Bus service was improved to only re-configure storage or iSCSI if it is really needed.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,3 +1,9 @@
+
+Wed Jan 28 12:59:50 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Add agama-iscsi service and integrate iSCSI into the new HTTP API
+  (gh#agama-project/agama#3064).
+
 -------------------------------------------------------------------
 Wed Jan 28 12:43:05 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
@@ -40,7 +46,7 @@ Fri Jan 23 14:10:16 UTC 2026 - Martin Vidner <mvidner@suse.com>
 -------------------------------------------------------------------
 Fri Jan 23 13:47:29 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 
-- Copy the SCC credentials and the libzypp services 
+- Copy the SCC credentials and the libzypp services
   (gh#agama-project/agama#3053).
 
 -------------------------------------------------------------------
@@ -63,7 +69,7 @@ Wed Jan 21 11:07:34 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Allow importing SSL certificates from registration servers
   (gh#agama-project/agama#3055).
-  
+
 -------------------------------------------------------------------
 Wed Jan 21 10:59:15 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
 


rust/agama-iscsi/src/client.rs
```diff
@@ -0,0 +1,92 @@
...
+pub enum Error {
+    #[error(transparent)]
+    DBus(#[from] zbus::Error),
+    #[error(transparent)]
+    Json(#[from] serde_json::Error),
+}
+
+#[async_trait]
+pub trait ISCSIClient {
+    async fn discover(&self, config: DiscoverConfig) -> Result<u32, Error>;
```
jreidinger: to be honest I do not like much using plain number...especially when there is no documentation for this method. What that u32 means? I can some enum be used instead?

rust/agama-iscsi/src/message.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl Discover {
+    pub fn new(config: DiscoverConfig) -> Self {
+        Self { config }
+    }
+}
+
+impl Message for Discover {
+    type Reply = u32;
```
jreidinger: and now that plain number leaks even to agent communication. Plain number is hard to read unless it is obvious from method name..but number as result of discover is just guessing. Number of discovered systems? Failure code from such discovery?

rust/agama-manager/src/service.rs
```diff
@@ -463,6 +493,13 @@ impl Service {
         Ok(())
     }
 
+    async fn discover_iscsi(&self, config: api::iscsi::DiscoverConfig) -> Result<(), Error> {
+        self.iscsi
+            .call(iscsi::message::Discover::new(config))
```
jreidinger, line 502: as it is call, it waits for result, but here it ignore u32 it returns, is it intentional?

rust/agama-utils/src/api/action.rs
```diff
@@ -18,11 +18,13 @@
...
 // find current contact information at www.suse.com.
 
-use crate::api::l10n;
+use crate::api::{iscsi, l10n};
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize, utoipa::ToSchema)]
 pub enum Action {
+    #[serde(rename = "discoverISCSI")]
+    DiscoverISCSI(iscsi::DiscoverConfig),
```
jreidinger, line 28: I would like to see some documenation that the action really do...it should end up in utoipa scheme, so any documentation is good.

rust/agama-utils/src/api/iscsi/config.rs
```diff
@@ -0,0 +1,151 @@
...
+        let Value::Object(lhs_object) = lhs_value else {
+            return;
+        };
+
+        let Value::Object(rhs_object) = rhs_value else {
+            return;
+        };
+
+        for (k, v) in rhs_object {
+            lhs_object.entry(k).or_insert(v);
```
jreidinger, line 47: well, what is implemented here is basically "ignore" strategy (in conflict prefer lhs over rhs )...I kind of expect it should be "overwrite" one ( in case of conflict use rhs over lhs ). I think it should be documented...especially "why" we use ignore strategy here.

rust/agama-utils/src/api/iscsi/config.rs
```diff
@@ -0,0 +1,151 @@
...
+                "initiator": "ini.test",
+                "targets": [
+                    { "name": "target2.test" }
+                ]
+            }
+            "#,
+        )
+        .unwrap();
+
+        config.merge(original);
```
jreidinger, line 113: this is probably confusing part. I kind of expect to have `original.merge(config)` aka merge to existing configuration changes and not other way around.

service/lib/agama/dbus/storage/iscsi.rb
```diff
@@ -86,10 +86,17 @@ def recover_config
         # @param serialized_config [String] Serialized iSCSI config according to the JSON schema.
         def configure(serialized_config)
           config_json = JSON.parse(serialized_config, symbolize_names: true)
+          logger.info("Configuring iSCSI: #{config_json.inspect}")
```
jreidinger: can it contain passwords to connect to iscsi?

rust/agama-manager/src/service.rs
joseivanlopez: Yes, it is intentional. The current API for actions is not returning any result. IMHO, it is something to consisder for the next iteration of the HTTP API.

service/lib/agama/dbus/storage/iscsi.rb
joseivanlopez: Good point!

rust/agama-utils/src/api/action.rs
```diff
@@ -23,6 +23,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize, utoipa::ToSchema)]
 pub enum Action {
+    /// Performs an iSCSI dicovery, finding nodes from the given portal.
```
jreidinger: typo dicovery

jreidinger (approved):

PR #3076, opened 2026-01-26
web: Log the data sent to the backend when the request fails

lslezak:
## Problem

- When debugging web frontend problems it might be useful to log the data sent to the backend
- Related to https://bugzilla.suse.com/show_bug.cgi?id=1256069

## Solution

- When the `fetch()` calls fail handle the error and log details
- Because we still do not have our own top-level error handler it will report a crash using the default React page. But at least we should be able to collect more details from the browser console.

## Notes

- Added a new `sanitize()` function which creates a copy of the input object with all sensitive properties like "password" or "registrationCode" replaced with text "[FILTERED]".
- This avoids logging sensitive data

## Testing

- Added unit test
- Tested with manually changed URL to invalid `/api/v2/configs` to trigger a server error, the console then contains this:
  agama-console-log-details
  The text in saved console content contains this:
  ```
  12:57:31.707 installHook.js:1 Request /api/v2/configs failed, sent data:
   {
    "update": {
      "root": {
        "sshPublicKey": "",
        "password": "[FILTERED]",
        "hashedPassword": false
      }
    }
  }
  ```

### DIFF
--- a/web/src/utils.test.ts
+++ b/web/src/utils.test.ts
@@ -517,7 +517,6 @@ describe("mergeSources", () => {
   });
 });
 
-
 describe("maskSecrets", () => {
   it("should filter sensitive keys from an object", () => {
     const obj = { user: "test", password: "123" };


web/src/http.ts
```diff
@@ -21,26 +21,46 @@
...
 
 import axios, { AxiosRequestConfig } from "axios";
+import { sanitize } from "~/utils";
 
 const http = axios.create({
   responseType: "json",
 });
 
+// helper function which logs details of a failed HTTP request
+const log_failure = function (url: string, error: Error, data?: unknown) {
```
dgdavid: NP: in TS we tend to use camelCase. Also, why not simply logError? Or logRequestError?

web/src/http.ts
```diff
@@ -21,26 +21,46 @@
...
+  if (data) {
+    // log the error: remove the sensitive data and serialize to string to avoid folding large
+    // data structures
+    console.error(`Request ${url} failed, sent data:\n`, JSON.stringify(sanitize(data), null, 2));
```
dgdavid: NP: why not having a param in sanitize to allow stringify the output?

`sanitize(data, { stringify: true })`

Apart from adding features to the function through an options argument, this would allow to just write one console.error, no matter if there is data or not.

` console.error(`Request ${url} failed\n`,  sanitize(data))`

Assuming sanitize will return null if no data is given.

web/src/utils.ts
```diff
@@ -160,6 +160,59 @@ const mask = (value: string, visible: number = 4, maskChar: string = "*"): strin
   return maskChar.repeat(maskedLength) + visiblePart;
 };
 
+// list of sensitive object properties replaced by the sanitize() function
+const default_sensitive_keys = [
```
dgdavid: NP, camelCase

web/src/utils.ts
```diff
@@ -160,6 +160,59 @@ const mask = (value: string, visible: number = 4, maskChar: string = "*"): strin
...
+ * const data = { user: "John", password: "123" };
+ * // logs { user: "John", password: "[FILTERED]" }
+ * console.log(sanitize(data));
+ * ```
+ *
+ * @param obj - The object or array to sanitize
+ * @param sensitive_keys - Custom keys to replace
+ * @returns Sanitized copy of the input
+ */
+const sanitize = (obj: unknown, sensitive_keys: string[] = default_sensitive_keys): unknown => {
```
dgdavid: NP, according to other comment I left, sensitive keys could be part of an options argument, to allow functions having more features with optional options. { sanitizeKeys: ..., stringify: true }

dgdavid: NP: BTW, `sanitize` is commonly used for strippping HTML or cleanup markup.
 In fact, there is a native  Sanitizer interface https://developer.mozilla.org/en-US/docs/Web/API/Sanitizer

Maybe we can call this `maskSecrets` or `filterSensitive` or something more tied to what the function does.

web/src/utils.ts
```diff
@@ -160,6 +160,74 @@ const mask = (value: string, visible: number = 4, maskChar: string = "*"): strin
...
+    }
+
+    if (Array.isArray(currentObj)) {
+      return currentObj.map((item) => mask(item));
+    }
+
+    const newObj: { [key: string]: unknown } = {};
+
+    for (const key in currentObj) {
+      if (Object.prototype.hasOwnProperty.call(currentObj, key)) {
```
dgdavid: I really do no understand the need of this.

From MDN, 

> The hasOwnProperty() method of [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object) instances returns a boolean indicating whether this object has the specified property as its own property (as opposed to inheriting it).
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

I guess is to avoid masking an inherited property... but I see no the problem of masking everything matches a sensitiveKey. If a "password" is inherited or not should not be a reason for no masking it value.

Also, MDN points that usage of  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn is recommended instead, and it has very good browser support.

web/src/utils.ts
```diff
@@ -160,6 +160,74 @@ const mask = (value: string, visible: number = 4, maskChar: string = "*"): strin
...
+        if (sensitiveKeys.includes(key)) {
+          newObj[key] = "[FILTERED]";
+        } else {
+          newObj[key] = mask(currentObj[key]);
+        }
+      }
+    }
+    return newObj;
+  };
```
dgdavid, line 207: NP: Relying in the usage of the included `radashi` dependency, you could _shorten_ the code going for something like


```suggestion
  const mask = (currentObj: unknown): unknown => {
    if (isArray(currentObj)) {
      return currentObj.map(mask);
    }

    return isPlainObject(currentObj)
      ? mapEntries(currentObj, (k, v) => [k, sensitiveKeys.includes(k) ? "[FILTERED]" : mask(v)])
      : currentObj;
    };
```

PR #3077, opened 2026-01-26
Change how the UI represents immutable products

ancorgs:
## Problem

When a product with read-only Btrfs root was chosen, an informative message was displayed at the top of the storage section.

master-master

On the one hand, this was considered a bit too intrusive and on the other hand it was only visible in the main storage screen and not in the form for adding or editing file-systems.

master-detail

## Solution

Remove the big message and use better wording when describing the file-system to be created. Both in the main storage screen...

new-master

...and in the forms.

new-detail

## Testing

- Tested manually
- Created new unit tests for the model
- Adapted other unit tests

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jan 30 16:27:35 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Update schemas for storage model and volume templates (related to
+  jsc#PED-14307).
+
+-------------------------------------------------------------------
 Fri Jan 30 11:00:08 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
 
 - Added proxy service allowing to configure the proxy through the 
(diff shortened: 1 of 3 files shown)


rust/agama-lib/share/storage.model.schema.json
```diff
@@ -136,7 +136,8 @@
         "default": { "type": "boolean" },
         "type": { "$ref": "#/$defs/filesystemType" },
         "snapshots": { "type": "boolean" },
-        "label": { "type": "string" }
+        "label": { "type": "string" },
+        "transactional": { "type": "boolean" }
```
joseivanlopez: As we discussed, perhaps we could add a new fs type to the model instead of a readonly *transactional* property.

ancorgs: This is an interesting idea. Right now we are using that trick of defining special filesystem types "btrfs with snapshots" in the UI. Maybe moving the idea one layer down to the model makes sense and would simplify the UI.

I will put this PR to rest (there are more urgent tasks in Agama's development), with the idea of exploring that. We should do the same not only for the filesystems but also for the volumes. So it's a bigger change.

ancorgs: So I finally updated the pull request to follow that approach. I squashed the commits so history looks like if we would have taken that decision from the very beginning (that's more clear than adding a `transactional` flag and then removing it).

PR #3079, opened 2026-01-27
Configure the SSH access (bsc#1257212)

lslezak:
## Problem

- The SSH service is not enabled when the SSH key is configured
- https://bugzilla.suse.com/show_bug.cgi?id=1257212

## Solution

- Enable the SSH service and open the SSH port in firewall when an SSH key for the root user is set.

## Testing

- Tested manually in openSUSE Tumbleweed installation

From the Agama web server log:
```
Jan 27 17:12:23 agama agama-web-server[7059]: The sshd service has been successfully enabled
Jan 27 17:12:23 agama agama-web-server[7059]: The SSH port has been successfully opened in firewall
```

After reboot the SSH server is running:

```
localhost:~ # systemctl status sshd
* sshd.service - OpenSSH Daemon
     Loaded: loaded (/usr/lib/systemd/system/sshd.service; enabled; preset: disabled)
     Active: active (running) since Tue 2026-01-27 17:20:54 CET; 7min ago
 Invocation: d2bd9a0b08724845ac0e963f215bc437
    Process: 833 ExecStartPre=/usr/sbin/sshd-gen-keys-start (code=exited, status=0/SUCCESS)
    Process: 925 ExecStartPre=/usr/sbin/sshd -t $SSHD_OPTS (code=exited, status=0/SUCCESS)
   Main PID: 928 (sshd)
      Tasks: 1
        CPU: 559ms
     CGroup: /system.slice/sshd.service
             `-928 "sshd: /usr/sbin/sshd -D [listener] 0 of 10-100 startups"

Jan 27 17:20:54 localhost.localdomain systemd[1]: Starting OpenSSH Daemon...
Jan 27 17:20:54 localhost.localdomain sshd-gen-keys-start[833]: Checking for missing server keys in /etc/ssh
Jan 27 17:20:54 localhost.localdomain sshd-gen-keys-start[837]: ssh-keygen: generating new host
    keys: RSA ECDSA ED25519
Jan 27 17:20:54 localhost.localdomain sshd[928]: Server listening on 0.0.0.0 port 22.
Jan 27 17:20:54 localhost.localdomain sshd[928]: Server listening on :: port 22.
Jan 27 17:20:54 localhost.localdomain systemd[1]: Started OpenSSH Daemon.
Jan 27 17:21:37 localhost.localdomain sshd-session[1239]: Accepted publickey for root from 192.168.1.113
    port 55518 ssh2: ED25519 SHA256:bUer1LwyLc...
Jan 27 17:21:37 localhost.localdomain sshd-session[1239]: pam_unix(sshd:session): session opened for user
    root(uid=0) by root(uid=0)
```

I could log in using the configured SSH key, as mentioned in the log above.

## Notes

- The firewall port is by default open in Tumbleweed, but even opening an already open port works fine. :smiley:
- Because the firewall might not be always installed it ignores the "not found" error. 
- I updated the code which sets the SSH key file permissions so it directly creates the file with the correct permissions. There was a race condition between creating the file and setting the permissions. Although it should not be a problem in Agama as there should be no other running processes but it still should be done correctly.

### DIFF
--- a/rust/agama-users/src/model.rs
+++ b/rust/agama-users/src/model.rs
@@ -201,7 +201,7 @@ impl Model {
                         "Cannot open SSH port in firewall",
                     )));
                 } else {
-                    tracing::info!("The SSH port has been successfully opened in firewall");
+                    tracing::info!("The SSH port has been successfully opened in the firewall");
                 }
             }
 


rust/agama-users/src/model.rs
```diff
@@ -142,19 +144,87 @@ impl Model {
...
+                        "Opening SSH port in firewall failed: exit: {}, stderr: {}",
+                        output.status,
+                        String::from_utf8_lossy(&output.stderr)
+                    );
+
+                    return Err(service::Error::CommandFailed(String::from(
+                        "Cannot open SSH port in firewall",
+                    )));
+                } else {
+                    tracing::info!("The SSH port has been successfully opened in firewall");
```
imobachgs:
```suggestion
                    tracing::info!("The SSH port has been successfully opened in the firewall");
```

rust/agama-users/src/model.rs
```diff
@@ -142,19 +144,87 @@ impl Model {
...
+                // in that case we do need to open the port
+                if e.kind() == std::io::ErrorKind::NotFound {
+                    tracing::warn!(
+                        "Command firewall-offline-cmd not found, firewall not installed?"
+                    );
+                } else {
+                    tracing::error!("Running firewall-offline-cmd failed: {}", e);
+
+                    return Err(service::Error::CommandFailed(String::from(
+                        "Cannot open SSH port in firewall",
```
imobachgs:
```suggestion
                        "Cannot open SSH port in the firewall",
```

rust/agama-users/src/model.rs
```diff
@@ -142,19 +144,87 @@ impl Model {
...
     }
 
+    /// Enables sshd service in the target system
+    fn enable_sshd_service(&self) -> Result<(), service::Error> {
+        let systemctl = self
+            .chroot_command()
+            .args(["systemctl", "enable", "sshd.service"])
+            .output()?;
+
+        if !systemctl.status.success() {
```
imobachgs, line 171: I already talked about this with @mchf: we might want to abstract the chroot and the handling of the success/failures. But definitely, something for the future please, not now :-)

PR #3086, opened 2026-01-28
Add support for product modes

imobachgs:
[PED-14307](https://jira.suse.com/browse/PED-14307)

This PR introduces the concept of product modes.

## Use case

From now on, if you select SLES 16.1, you can choose between installing the system following the traditional partitioning approach (with a read/write root like 16.0) or using the transactional mode.

## Modes

Modes support is not limited to the described use case. Each mode can redefine the software or the storage sections of the product. In the previous case:

* It selects the `transactional_base` or `enhanced_base` depending on the mode.
* Defines a different partitioning layout.

## Defining modes

The modes definition lives under the `modes` key. It includes:

* Some meta-data (`id`, `name`, and `description`).
* Optionally, a `software` key, which is merged with the base one.
* Optionally, a `storage` key which is merged with the base one.

```yaml
modes:
  - id: standard
    name: Standard
    description: Standard system

    software:
      mandatory_patterns:
        - enhanced_base

  - id: immutable
    name: Immutable
    description: Immutable system with atomic updates

    software:
      mandatory_patterns:
        - transactional_base
```

## Testing

- Added a new unit test
- Tested manually (in progress)

## Screenshots

First product selection

Changing the selected product

### DIFF
--- a/web/src/components/product/ProductSelectionPage.tsx
+++ b/web/src/components/product/ProductSelectionPage.tsx
@@ -121,7 +121,7 @@ const ProductFormProductOption = ({
                             <Text component="small">{_("License acceptance required")}</Text>
                           </Label>
                         )}
-                        {product.modes && (
+                        {!isEmpty(product.modes) && (
                           <Label variant="outline" isCompact>
                             <Text component="small">
                               {sprintf(_("%d modes available"), product.modes.length)}


rust/agama-utils/src/products.rs
```diff
@@ -241,17 +377,19 @@ pub struct LabelSpec {
...
-    #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[merge(strategy = merge::option::overwrite_none)]
     boot_strategy: Option<String>,
-    #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[merge(strategy = merge::option::overwrite_none)]
     space_policy: Option<String>,
     #[serde(default)]
+    #[merge(strategy = merge::vec::append)]
```
joseivanlopez, line 389: I am not sure about this. For now, I would assume the storage section is totally overried by the mode.

web/src/components/product/ProductSelectionPage.tsx
```diff
@@ -111,13 +114,20 @@ const ProductFormProductOption = ({
...
                   <Stack hasGutter id={detailsId}>
-                    {product.license && (
+                    {(product.license || product.modes) && (
                       <Split hasGutter>
                         {product.license && (
                           <Label variant="outline" isCompact>
                             <Text component="small">{_("License acceptance required")}</Text>
                           </Label>
                         )}
+                        {product.modes && (
```
joseivanlopez: I would avoid to show this label if modes length is 0.

rust/agama-utils/src/products.rs
jreidinger: well, similar is for software...but in current usage in products it is more like product nothing and modes everything for both.

web/src/components/product/ProductSelectionPage.tsx
jreidinger: agreed. That is also what hit me when I see screenshots.

rust/agama-utils/src/products.rs
imobachgs: I did in this way just for consistency with software. For instance, you might want to have a common list of user patterns and add only the differences, instead of repeating the whole list.

If you want the whole list to be replaced, well, you just specify an empty list in the base configuration.

web/src/components/product/ProductSelectionPage.tsx
imobachgs: Yes, right, in the past the list of modes could be empty. But now it is always present.

PR #3087, opened 2026-01-28
Gpg repo signature

jreidinger:
## Problem

There is missing support for "gpgFingerprints" and "allowUnsigned" keys in extra repositories section of profile. Also in popup for unknown gpg key fingerprint is printed all together which is hard to read and it is regression to SLE-16 version.

## Solution

Add missing support and also print human friendly fingerprint in popup

## Testing

- *Tested manually*

with following profile:

```json
{
	"product": {
		"id": "Tumbleweed"
	},
	"software": {
		"extraRepositories": [
			{
				"alias": "games",
				"url": "https://download.opensuse.org/repositories/games/openSUSE_Tumbleweed/",
				"gpgFingerprints": ["AEF1 EE83 0A27 90AB ABC2 CDA2 8F91 B0E6 C862 B42C"]

			},
			{
				"alias": "mine",
				"url": "ftp://pong.suse.cz/testrepo/",
				"allowUnsigned": true
			}
		],
		"packages": [
			"AtomicWorm"
		]
	}
}
```

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -389,6 +389,7 @@ dependencies = [
  "agama-locale-data",
  "agama-transfer",
  "async-trait",
+ "camino",
  "cidr",
  "fluent-uri 0.4.1",
  "fs-err",
--- a/rust/agama-software/src/model.rs
+++ b/rust/agama-software/src/model.rs
@@ -40,7 +40,7 @@ pub mod software_selection;
 pub mod state;
 
 pub use packages::{Resolvable, ResolvableType};
-pub use registration::{Registration, RegistrationBuilder};
+pub use registration::Registration;
 
 /// Abstract the software-related configuration from the underlying system.
 ///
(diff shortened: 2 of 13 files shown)


rust/agama-software/src/callbacks/security.rs
```diff
@@ -3,16 +3,30 @@ use agama_utils::{actor::Handler, api::question::QuestionSpec, question};
...
     pub fn new(questions: Handler<question::Service>) -> Self {
-        Self { questions }
+        Self {
+            questions,
+            trusted_gpg_keys: vec![],
+            unsigned_repos: vec![],
+        }
+    }
+
+    pub fn set_trusted_gpg_keys(&mut self, trusted_gpg_keys: Vec<RepoKey>) {
```
joseivanlopez, line 24: NP: why not a public field?

jreidinger: well, I consider it as private property that can be set, but read only internally...if I see need to both get/set, then I would change it, but for now I use this set variant as write only. Originally I even have that "with" variant, but it makes code manipulation harder to read.

joseivanlopez (approved):
LGTM

PR #3088, opened 2026-01-28
Actions and registration fixes

imobachgs:
Fixes a few issues found during testing:

* The check to prevent starting an installation when it is not possible was too aggressive, preventing other actions to be executed (see #3057).
* Add a line to the logs when the installation starts and when it finishes.
* Fix the registration screen, which was loosing the "mode".
* Force to choose a mode (in case it is needed) in the product selection screen.


teclator (approved):
LGTM

teclator (approved):
and last but not least... LGTM ;)

PR #3090, opened 2026-01-28
live: Enable the YAML syntax highlighting in vim

lslezak:
## Problem

- When tweaking the Agama product files in the Live ISO the syntax highlighting for YAML files is not active.

## Solution

- Keep the vim YAML syntax files in the ISO

## Testing

- Tested manually in a locally built ISO

## Screenshots

agama-yaml-syntax

### DIFF
--- a/live/src/config.sh
+++ b/live/src/config.sh
@@ -213,7 +213,7 @@ rm /var/log/zypper.log /var/log/zypp/history
 # reduce the "vim-data" content, this package is huge (37MB unpacked!), keep
 # only support for JSON (for "agama config edit"), YAML (the product definition
 # files) and Ruby (fixing/debugging the Ruby service)
-rpm -ql vim-data | grep -v -e '/ruby.vim$' -e '/json.vim$' -e '/yaml.vim$' -e colors | xargs rm 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


live/src/config.sh
```diff
@@ -210,10 +210,10 @@ fi
...
 rm /var/log/zypper.log /var/log/zypp/history
 
-# reduce the "vim-data" content, this package is huge (37MB unpacked!), keep only
-# support for JSON (for "agama config edit") and Ruby (fixing/debugging the Ruby
-# service)
-rpm -ql vim-data | grep -v -e '/ruby.vim$' -e '/json.vim$' -e colors | xargs rm 2>/dev/null || true
+# reduce the "vim-data" content, this package is huge (37MB unpacked!), keep
+# only support for JSON (for "agama config edit"), YAML (the product definition
+# files) and Ruby (fixing/debugging the Ruby service)
+rpm -ql vim-data | grep -v -e '/ruby.vim$' -e '/json.vim$' -e '/yaml.vim$' -e colors | xargs rm 2>/dev/null || true
```
jreidinger: I would probably keep also `bash.vim` for easier tweaking shell scripts.

lslezak: Ah, that already works:

agama-bash-syntax

jreidinger (approved):

PR #3092, opened 2026-01-28
refactor(web): restore interface for managing iSCSI form web client

joseivanlopez:
Bring back the iSCSI user interface fully adapted to the HTTP v2 API and to the latest web client look&feel (https://trello.com/c/S65oeTKz, private link). It also restores the "Rescan devices" option (related to https://bugzilla.suse.com/show_bug.cgi?id=1258089, private link).

---

In addition to other minor changes not strictly related to the iSCSI user interface, during the resurrection of such an interface this PR introduced few important layout updates to make the Header and Breadcrumb more flexible and extensible. These adjustments were necessary to accommodate the placement of the “More storage options” menu while keeping consistence across other areas of Agama.

For more details, please refer to the relevant commits below.

  * [feat(web): add support for sibling menus on breadcrumb items](https://github.com/agama-project/agama/pull/3092/commits/374697f234bee026f13fb483640641607a4e7d10) 
  * [fix(web): pass down all breadcrumb props from header](https://github.com/agama-project/agama/pull/3092/commits/6615e328b806826ab9b8cb49212fcb0f726edd93)
  * [fix(web): rename InstallerOptions to InstallerL10nOptions](https://github.com/agama-project/agama/pull/3092/commits/05b41bd16b0d53a38709e53f9687f32cedc838b3)
  * [refactor(web): introduce slots for Header](https://github.com/agama-project/agama/pull/3092/commits/f8b601bb6452fa985d5e5a342a249a4eb7d39a89)
  * [fix(web): reduce page title font size](https://github.com/agama-project/agama/pull/3092/commits/033e7bc24ef718ac6248a3536e3aca8c1d52f790)
  * [refactor(web): display Overview options in sidebar](https://github.com/agama-project/agama/pull/3092/commits/500cdb65465ef1e927bc41ff77eb11838b6cd343)

---

Related to 

  * https://github.com/agama-project/agama/pull/3008
  * https://github.com/agama-project/agama/pull/3064

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -6,6 +6,11 @@ Thu Feb 12 11:56:57 UTC 2026 - David Diaz <dgonzalez@suse.com>
 - Restore the "Rescan devices" storage option (bsc#1258089).
 
 -------------------------------------------------------------------
+Wed Feb 11 14:24:03 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Honor the language from the backend (gh#agama-project/agama#3142).
+
+-------------------------------------------------------------------
 Wed Feb 11 13:35:13 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
 
 - Show translated product mode name and description
--- a/web/src/components/product/LicenseDialog.test.tsx
+++ b/web/src/components/product/LicenseDialog.test.tsx
@@ -110,7 +110,7 @@ describe("LicenseDialog", () => {
       await waitFor(() => {
         expect(mockFetchLicense).toHaveBeenCalledWith(sle.license, mockUILanguage);
         screen.getByText("El contenido de la licencia");
-        screen.getByText("This license is not available in Deutsch.");
+        screen.getByText("Diese Lizenz ist in Deutsch nicht verfügbar.");
       });
     });
   });
@@ -119,7 +119,7 @@ describe("LicenseDialog", () => {
     const { user } = installerRender(<LicenseDialog product={product} onClose={onCloseFn} />, {
       withL10n: true,
     });
-    const closeButton = screen.getByRole("button", { name: "Close" });
+    const closeButton = screen.getByRole("button", { name: "Schließen" });
     await user.click(closeButton);
     expect(onCloseFn).toHaveBeenCalled();
   });
--- a/web/src/components/product/ProductSelectionPage.test.tsx
+++ b/web/src/components/product/ProductSelectionPage.test.tsx
@@ -106,7 +106,7 @@ describe("ProductSelectionPage", () => {
 
     // No product selected yet
     mockProduct(undefined);
-    const { rerender } = installerRender(<ProductSelectionPage />);
+    const { rerender } = installerRender(<ProductSelectionPage />, { withL10n: true });
     screen.getByRole("radio", { name: tumbleweed.name });
     screen.getByRole("radio", { name: microOs.name });
     screen.getByRole("radio", { name: productWithModes.name });
@@ -134,7 +134,7 @@ describe("ProductSelectionPage", () => {
   // test ensures the selected option remains checked across re-renders with new
   // object references.
   it("keeps product selection across re-renders", async () => {
-    const { user, rerender } = installerRender(<ProductSelectionPage />);
+    const { user, rerender } = installerRender(<ProductSelectionPage />, { withL10n: true });
     const microOsOption = screen.getByRole("radio", { name: microOs.name });
     expect(microOsOption).not.toBeChecked();
     await user.click(microOsOption);
@@ -153,7 +153,7 @@ describe("ProductSelectionPage", () => {
 
   it("force license acceptance for products with license", async () => {
     mockProduct(undefined);
-    const { user } = installerRender(<ProductSelectionPage />);
+    const { user } = installerRender(<ProductSelectionPage />, { withL10n: true });
     expect(screen.queryByRole("checkbox", { name: /I have read and accept/ })).toBeNull();
     const selectButton = screen.getByRole("button", { name: "Select" });
     const microOsOption = screen.getByRole("radio", { name: microOs.name });
@@ -175,7 +175,7 @@ describe("ProductSelectionPage", () => {
       products: [productWithLicense1, productWithLicense2],
     });
 
-    const { user } = installerRender(<ProductSelectionPage />);
+    const { user } = installerRender(<ProductSelectionPage />, { withL10n: true });
 
     // Select first product and accept license
     const product1Option = screen.getByRole("radio", { name: "Product 1" });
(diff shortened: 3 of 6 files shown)


web/src/openapi/system/iscsi.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+export interface Target {
+  name: string;
+  address: string;
+  port: number;
+  interface: string;
+  /**
+   * Whether the target was initiated by iBFT.
+   */
+  ibtf: boolean;
```
dgdavid: @joseivanlopez @ancorgs 

I know this is an autogenerated file, but looks like somewhere there is a typo that makes the interface to own `ibtf` property instead of the expected `ibft`.

joseivanlopez: Good catch. I usually mix "ft" and "tf".

web/src/model/config/iscsi.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function removeTarget(config: Config, name: string): Config {
+  config = clone(config);
+  const index = config.targets?.findIndex((t) => t.name === name);
+  if (index) config.targets.splice(index, 1);
+  return config;
+}
+
+export default { setInitiator, findTargetIndex, findTarget, addTarget, removeTarget };
+export type * from "~/openapi/config/iscsi";
```
dgdavid, line 45: @joseivanlopez, question: what’s the rationale for re-exporting these types here? I’m finding it a bit hard to follow this indirection of exporting them in one place and then re-exporting again under the ISCSI namespace in model/config. I’m probably missing an important design reason, but some clarification would help.

joseivanlopez: The goal is to only use types from the model. In some cases the types of the model are directly the types from openapi, but not always. It was agreed with Imo to follow this approach. What is exactly the drawback? Of course, we can reconsider it anyway.

joseivanlopez: Another thing to consider is that our model is a layer over openapi types. The idea is to build our model using openapi types and providing methods to work with objects of that types. In the future we could have a separate package with everything provided by openapi, so other clients using our API could benefit of it.

dgdavid: > What exactly is the drawback?

There’s no specific drawback per se. I just find the approach hard to follow, with little benefit in return. I understand the rationale behind having a model, config, or similar layer. That part is fine. What I don’t see the value in is re-exporting everything from openapi.

When I look at the model, I can tell that I’ll have access to everything from openapi, but that doesn’t actually tell me what “everything” is. I still need to inspect the openapi package itself to understand which types are available.

This approach would make sense to me if the model were intentionally restricting what the application can use from openapi, or from any external package. For example, if openapi exposed five types but, for some reason, the Agama model only supported Type1 and Type3, then re-exporting just those would be valuable. In that case, a developer could know the available types simply by looking at the model.

I also understand the intention behind treating the openapi output as a third-party package, but I still don’t see the value in it. I think things would be more predictable if the same approach were followed app-wide, for example having app/types in the same way we have app/model, app/hooks, and so on.

Not sure if I’m explaining my concern clearly, but in any case, if you guys have found advantages to working this way, I can of course live with it, even if it still feels unnecessarily complicated to me.

dgdavid: Thinking about this while driving made me realize another drawback, beyond the indirection mentioned earlier: using import * would defeat tree-shaking. In the theoretical case where openapi is a separate package from which we import types, the web interface would end up importing and bundling everything, rather than only the types that are actually used.

joseivanlopez: IMHO, lacking of tree-shaking with types is not a problem. We will need all the types anyway.

dgdavid: I don't fully agree.

We can't be sure whether the web interface will need all of the exported types in the future, nor should we decide upfront that the lack of tree-shaking isn't a problem :cry: In fact, it isn't inherently a problem bundle everything just because it does not hurt (except for the user download bundle size) but it's a nice feature that exists for a reason.

In any case, as said is just another small drawback I personally see, in addition of the complexity we are adding there with these import and export *. 

Anyway, let's move on since it is just me complaining about it. I can survive :wink:

joseivanlopez: Sure, I understand the benefits of tree-shaking when importing features from a big library. In our case, it is our API, and well will need the types. Just a note: in some cases I see benefits of grouping types (`System.Device`, `Proposal.Device`) over fine grain tree-shaking. It saves us to rename types all the time you need both types in the same scope (`SystemDevice`, `ProposalDevice`). I think everything has its pros and cons. 

Anyway, as commented in other channels, I am not against to change the approach and see how it works.

web/src/components/storage/iscsi/TargetsTable.tsx
```diff
@@ -0,0 +1,604 @@
...
+  disconnected: N_("Disconnected"),
+  missing: N_("Missing"),
+};
+
+/**
+ * A map of statuses human-readable status labels.
+ * TODO: Move to a more appropiated place
+ */
+type StatusKey = keyof typeof STATUS_LABELS;
+
```
dgdavid, line 172: **Note:** Ideally, this status logic should live elsewhere. However, I couldn’t make it fit cleanly into the _frontend model layer_ because there are separate models for targets in config and in the system, and the targets handled here may belong to either or even both.

Aside from relocating it (likely in a future iteration without "merged targets"), this logic could be significantly simplified. For now, it works as intended for the initial iSCSI frontend implementation using the new HTTP API. There is still room for improvement, particularly by rethinking what is displayed in the table, which will be addressed in follow-up work.

dgdavid: Additionally, as @ancorgs told me at IRC #yast channel

>  [...] merging the two sources into a single collection should also be model logic

But still no "right" file model (yet) to do so. Moreover, as commented above, the merged collection most probably will be no longer needed in a short/mid-term future, so maybe it does not payoff moving all of this until having a more clear vision about its future.

web/src/components/core/Page.tsx
```diff
@@ -300,92 +301,14 @@ const Content = ({ children, ...pageSectionProps }: PageSectionProps) => {
...
-      {showQuestions && <Questions />}
-    </PFPage>
-  );
-};
-
 /**
  * Common props shared by all page variants.
  */
-interface BasePageProps {
+interface BasePageProps extends HeaderProps {
```
dgdavid, line 307: I'm not really sure, but maybe would be more convenient, predictable, and ergonomics to have a `headerProps` instead. Although `title` and `breadcrumbs` are so convenient to be related to page as well. In any case, I believe this PR is already rather long and changes in that regard should be sent as part of a new one.

It would allow to pass from 

```jsx
<Page breadcrumbs={[...]} noDefaultStartSlot endSlot={...} />
````

to

```jsx
<Page header={breadcrumbs: [...], noDefaultStartSlot: true, endSlot: ... />
```

It has advantages and disadvantages.  Advantages for defining slots contents, disadvantages passing title, breadcrumbs and "noDefaultX".

PR #3098, opened 2026-01-29
Add Leap 16.1

lkocman:
## Problem

* Add openSUSE Leap 16.1 product configuration. 
* Drop 16.0 so we keep only latest verison in main


imobachgs:
Thanks!

PR #3100, opened 2026-01-29
Requires only flag

jreidinger:
## Problem

There is missing support for requiresOnly flag.

trello: https://trello.com/c/Pi5qPdis/5298-5-add-onlyrequired-support

## Solution

Implement it for solver and also use new zypp.conf.d for persistent settings after reboot.

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -441,7 +441,7 @@ impl ZyppServer {
         }
 
         self.only_required = state.options.only_required;
-        tracing::info!("only_required: {}", self.only_required);
+        tracing::info!("Install only required packages: {}", self.only_required);
         // run the solver to select the dependencies, ignore the errors, the solver runs again later
         let _ = zypp.run_solver(self.only_required);
 
@@ -609,7 +609,7 @@ impl ZyppServer {
             let path = self.install_dir.join("etc/zypp/zypp.conf.d/installer.conf");
             let write_result = std::fs::write(path.as_path(), contents);
             if write_result.is_err() {
-                tracing::error!("Failed to write {path}. {write_result:?}");
+                tracing::error!("Failed to write {path}: {write_result:?}");
             }
         }
     }


rust/agama-software/src/zypp_server.rs
```diff
@@ -597,9 +602,16 @@ impl ZyppServer {
...
     }
 
-    fn modify_zypp_conf(&self) -> ZyppServerResult<()> {
-        // TODO: implement when requireOnly is implemented
-        Ok(())
+    fn modify_zypp_conf(&self) {
+        // write only if different from default
+        if self.only_required {
+            let contents = "# Use only hard dependencies as configured in installer\nsolver.onlyRequires = true\n";
+            let path = self.install_dir.join("etc/zypp/zypp.conf.d/installer.conf");
```
imobachgs, line 609: Just a question: in other `.d` directories, the name of the files is usually prefixed with a number (`50-installer.conf`). Is that expected in this case?

rust/agama-software/src/zypp_server.rs
```diff
@@ -597,9 +602,16 @@ impl ZyppServer {
...
+            let write_result = std::fs::write(path.as_path(), contents);
+            if write_result.is_err() {
+                tracing::error!("Failed to write {path}. {write_result:?}");
```
imobachgs:
```suggestion
                tracing::error!("Failed to write {path}: {write_result:?}");
```

rust/agama-software/src/zypp_server.rs
```diff
@@ -438,8 +440,10 @@ impl ZyppServer {
             };
         }
 
+        self.only_required = state.options.only_required;
+        tracing::info!("only_required: {}", self.only_required);
```
lslezak: I'd make the message more clear:

```suggestion
        tracing::info!("Install only required packages: {}", self.only_required);
```

rust/agama-software/src/zypp_server.rs
jreidinger: to mention it also here. I use basically format from example from libzypp developer who use for micro os `micro.conf` without any number.

PR #3101, opened 2026-01-29
Do not export neither network/state nor network/connections when they are None

imobachgs:
## Problem

[bsc#1257400](https://bugzilla.suse.com/show_bug.cgi?id=1257400).

Similar problem to https://github.com/agama-project/agama/pull/3094 but for `network/state` and `network/connections`.

## Solution

Do not export them when they are `None`.

## Tested

* Manually tested.


ancorgs (approved):
LGTM

PR #3103, opened 2026-01-30
Fix selinux support

jreidinger:
## Problem

when selinux pattern is selected, it is still not activated on kernel cmdline.

## Solution

The change is basically to allow from different services call bootloader with own set of kernel parameters. In software it now checks selinux pattern and if it is selected, it adapts kernel parameters to enable it.
Also during manual testing it was found that it is not possible to unselect selinux in pattern selector and this PR fixing it.

## Testing

- *Tested manually*

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Mon Feb  2 08:13:10 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 
-- Implement modification of kernel parameters when selinux pattern
+- Implement modification of kernel parameters when SELinux pattern
   is selected (gh#agama-project/agama#3103)
 
 -------------------------------------------------------------------


rust/agama-software/src/service.rs
```diff
@@ -190,6 +198,28 @@ impl Service {
         Ok(())
     }
 
+    fn update_selinux(&self, state: &SoftwareState) {
```
imobachgs, line 200: Perhaps it would be cleaner to extend `SoftwareState` with a method to search for a selected pattern (or a selected resolvable).

rust/agama-software/src/service.rs
```diff
@@ -190,6 +198,28 @@ impl Service {
...
+            typ == &ResolvableType::Pattern && name == "selinux" && state.selected()
+        });
+
+        let value = if selinux_selected {
+            "security=selinux"
+        } else {
+            "security="
+        };
+        let arg = KernelArg {
+            scope: "selinux".to_string(),
```
imobachgs: I would not use the term `scope` here. It already has [a meaning](https://github.com/agama-project/agama/blob/master/rust/agama-utils/src/api/scope.rs#L41). I guess you can use `Scope::Software` instead. I do not expect to set more than a parameter from a given scope (and we can put all of them in the same call).

If you want something not tied to the scope, I would go for `id` (like the resolvables list).

rust/agama-software/src/service.rs
```diff
@@ -190,6 +198,28 @@ impl Service {
...
+            value: value.to_string(),
+        };
+        let res = self
+            .bootloader
+            .cast(agama_bootloader::message::SetKernelArg { arg });
```
imobachgs: np: I am not sure about the form of this message. In every place I see the kernel args as a `HashMap<String, String>` (it could be `HashMap<Scope, String>`). I would define the message as:


```rust
struct SetKernelArg {
  scope: Scope,
  args: String,
}
```

rust/agama-bootloader/src/client.rs
```diff
@@ -47,6 +50,18 @@ pub trait BootloaderClient {
     async fn get_config(&self) -> ClientResult<Config>;
     /// Sets the bootloader configuration.
     async fn set_config(&self, config: &Config) -> ClientResult<()>;
+    /// Sets the extra kernel args for given scope.
+    async fn set_kernel_arg(&mut self, arg: KernelArg);
```
imobachgs: Why not `set_kernel_arg(&mut self, scope: Scope, args: &str);`? This `KernelArg` looks like an unneeded wrapper.

rust/agama-software/src/service.rs
jreidinger: well, I usually try to make shared method when there are at least two users of it.

imobachgs: Well, the problem is that the caller needs to know too many details about `SoftwareState`. But it is not super important.

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Feb  2 08:13:10 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Implement modification of kernel parameters when selinux pattern
```
```suggestion
- Implement modification of kernel parameters when SELinux pattern
```

PR #2669, opened 2025-08-19
fix(live): make the resolv.conf accesible from /sysroot

imobachgs:
## Problem

- [bsc#1248276](https://bugzilla.suse.com/show_bug.cgi?id=124827)

When applying a `DUD`, it is not possible to access the network when running `agama download` from the `/sysroot`.

## Solution

* Mount the `/run` file system on `/sysroot/run`.
* Link `/sysroot/etc/resolv.conf` to `/sysroot/run/etc/NetworkManager/resolv.conf` file.
* Remove the `/run/agama/dud` directory to release some memory before pivoting.

## Testing

- Tested manually.

A testing build should be available [home:IGonzalezSosa:branches:fix-dud-apply-network](https://build.suse.de/package/show/home:IGonzalezSosa:branches:fix-dud-apply-network/agama-installer) at some point.

PR #2179, opened 2025-03-20
Symlink the resolv.conf around post scripts run

teclator:
## Problem

When running post-scripts in a chroot there is no DNS resolution because there is no resolv.conf

- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1235617
- Trello card: https://trello.com/c/9jDKP62p

## Solution

Create a symling to /run/NetworkManager/resolv.conf if there is no /mnt/etc/resolv.conf before running the post scripts deleting it once finished.

## Testing

- *Tested manually*

PR #3105, opened 2026-01-30
Fix resolv conf link (bsc#1257468)

teclator:
## Problem

After the installation, when running a post-script there is a leftover symlink linking the /etc/resolv.conf to /mnt/run/NetworkManager/resolv.conf

- *https://bugzilla.suse.com/show_bug.cgi?id=1257468*

## Solution

Link the `/mnt/etc/resolv.conf` to `/run/NetworkManager/resolv.conf` as it was done in https://github.com/agama-project/agama/pull/2179 or https://github.com/agama-project/agama/pull/2669

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/rust/agama-files/src/runner.rs
+++ b/rust/agama-files/src/runner.rs
@@ -255,10 +255,6 @@ impl ScriptsRunner {
     fn resolv_link_path(&self) -> PathBuf {
         self.install_dir.join(RESOLV_CONF_PATH)
     }
-
-    fn root_path(&self) -> PathBuf {
-        self.root_dir.clone()
-    }
 }
 
 #[cfg(test)]


rust/agama-files/src/runner.rs
```diff
@@ -228,7 +232,7 @@ impl ScriptsRunner {
     ///
     /// It returns false if the resolv.conf was already linked and no action was required.
     fn link_resolv(&self) -> Result<bool, std::io::Error> {
-        let original = self.install_dir.join(NM_RESOLV_CONF_PATH);
+        let original = self.root_path().join(NM_RESOLV_CONF_PATH);
```
imobachgs: I checked that the `root_path()` just clones the `root_dir` field. I think you do not need to clone it, so you can get rid of that function.


```suggestion
        let original = self.root_dir.join(NM_RESOLV_CONF_PATH);
```

rust/agama-files/src/runner.rs
```diff
@@ -251,6 +255,10 @@ impl ScriptsRunner {
     fn resolv_link_path(&self) -> PathBuf {
         self.install_dir.join(RESOLV_CONF_PATH)
     }
+
+    fn root_path(&self) -> PathBuf {
```
imobachgs: As commented above, you do not need this function.

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Feb  3 07:23:18 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Fixed linking the /etc/resolv.conf file in the chroot to
```
```suggestion
- Correct the /etc/resolv.conf link in the chroot to
```

imobachgs (approved):
Just a minor comment. Otherwise, it looks good.

PR #3110, opened 2026-01-30
Adapt DASD D-Bus API to API v2

joseivanlopez:
Adapt the D-Bus service to provide a new DASD API suitable to serve the new HTTP API.

Note: the new HTTP API for DASD will be implemented in a separate PR.

### DIFF
--- a/service/lib/agama/storage/dasd/manager.rb
+++ b/service/lib/agama/storage/dasd/manager.rb
@@ -58,8 +58,16 @@ module Agama
         def initialize(logger: nil)
           @logger = logger || ::Logger.new($stdout)
           @devices = Y2S390::DasdsCollection.new([])
+          # Keeps whether a configuration was already applied. In some cases it is necessary to
+          # consider that the config has not being applied yet, see {#probe}.
           @configured = false
+          # Keeps the list of locked devices. A device is considered as locked if it is active at
+          # the time of probing. Those devices should not be deactivated when applying a config that
+          # does not include such devices.
           @locked_devices = []
+          # Keeps the list of formatted devices. A device is added to the list when it was formatted
+          # as effect of applying a config. Those devices should not be formatted anymore when
+          # applying a new config.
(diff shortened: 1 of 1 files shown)


service/lib/agama/dbus/storage/dasd.rb
```diff
@@ -19,151 +19,173 @@
...
-        def type_from(device_type)
-          cu_type, dev_type = device_type.split
-          return ECKD if cu_type.to_s.start_with?("3990", "2105", "2107", "1750", "9343")
-          return FBA if cu_type.to_s.start_with?("6310")
-
-          if cu_type.start_with?("3880")
-            return ECKD if dev_type.start_with?("3390")
-            return FBA if dev_type.start_with?("3370")
-          end
+        # @raise if there is an unfinished configuation.
```
teclator:
```suggestion
        # @raise if there is an unfinished configuration.
```

service/lib/agama/storage/dasd/config.rb
```diff
@@ -0,0 +1,54 @@
...
+        # List of devices.
+        #
+        # @return [Array<Configs::Device>]
+        attr_accessor :devices
+
+        def initialize
+          @devices = []
+        end
+
+        # Whether the config includes a devices with the given channel.
```
teclator:
```suggestion
        # Whether the config includes a device with the given channel.
```

service/lib/agama/storage/dasd/manager.rb
```diff
@@ -59,6 +59,8 @@ def initialize(logger: nil)
           @logger = logger || ::Logger.new($stdout)
           @devices = Y2S390::DasdsCollection.new([])
           @configured = false
+          @locked_devices = []
```
teclator, line 67: NP: maybe it deserves its reader method explaining what a locked device is, it is already described in the assignation but you have to go to the method to know more about.

PR #3111, opened 2026-01-30
agama-web-server: keep working even if IPv6 is not available

mvidner:
## Problem

The Agama installer does not start if IPv6 is disabled on the kernel command line with 'ipv6.disable=1'.

- https://bugzilla.suse.com/show_bug.cgi?id=1257251

## Solution

If `:::$PORT` does not work, try on `0.0.0.0:$PORT`

It works but the code is hacky.

## Testing

Tested manually in a VM booted with `ipv6.disable=1`

## Screenshots

## Documentation

- Run: `git ls-files '*.md'`

### DIFF
--- a/rust/agama-server/src/agama-web-server.rs
+++ b/rust/agama-server/src/agama-web-server.rs
@@ -273,19 +273,29 @@ async fn handle_http_stream(
     }
 }
 
+async fn find_listener(addresses: String) -> Option<tokio::net::TcpListener> {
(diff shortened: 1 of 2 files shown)


imobachgs (commented):
Thanks! I was wondering whether we should keep `--address2` given that you can already specify multiple addresses. Or perhaps we could just allow multiple `--address` arguments.

PR #3122, opened 2026-02-03
Run an storage probing after pre-scripts

imobachgs:
## Problem

After running pre-scripts, the storage devices are not "probed" again.

- [bsc#1257501](https://bugzilla.suse.com/show_bug.cgi?id=1257501)

## Solution

Run pre-scripts on a separate task and notify storage about the change.

## Alternative proposals

We thought about two alternative options:

- Introduce a notification mechanism between services. It is not a simple matter and it could make problems harder to debug.
- Make the files service to directly trigger the probing. In this case, the files service knows too much about the general workflow. What if tomorrow we need to do another additional task?

## Testing

- [x] Tested manually

### DIFF
--- a/rust/agama-manager/src/service.rs
+++ b/rust/agama-manager/src/service.rs
@@ -627,10 +627,17 @@ impl Service {
         let files = self.files.clone();
         let storage = self.storage.clone();
         tokio::spawn(async move {
-            let pre_scripts_ran = files
+            let pre_scripts_ran = match files
                 .call(files::message::RunScripts::new(ScriptsGroup::Pre))
                 .await
-                .unwrap();
+            {
+                Ok(result) => result,
+                Err(error) => {
+                    tracing::error!("Failed to run pre-scripts: {error}");
+                    return;
+                }
+            };
+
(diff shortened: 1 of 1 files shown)


rust/agama-manager/src/service.rs
```diff
@@ -623,6 +621,23 @@ impl Service {
...
+
+    /// It runs pre-scripts and asks storage for probing.
+    fn run_pre_scripts(&self) {
+        let files = self.files.clone();
+        let storage = self.storage.clone();
+        tokio::spawn(async move {
+            let pre_scripts_ran = files
+                .call(files::message::RunScripts::new(ScriptsGroup::Pre))
+                .await
+                .unwrap();
```
lslezak: At first I was a bit confused what happens in case of error, but then I realized it always returns `Ok()` and never fails. Maybe add a short comment?

imobachgs: Well, actually the result is that this tokio task will crash. But I have added code to handle that case, so at least we have info in the logs.

PR #3123, opened 2026-02-04
service: Do not confuse iSCSI targets with the same name

ancorgs:
## Problem

An iSCSI target (or node, if you use the `iscsiadm` nomenclature) is identified by its name and its portal. Since the same underlying system (name) can be exposed through several network endpoints (portal).

But Agama was using only the name (ignoring the portal) to identify which nodes need to be connected or disconnected.

## Solution

Use the whole information to match the nodes.

## Testing

Added new unit tests

### DIFF
--- a/service/lib/agama/storage/iscsi/manager.rb
+++ b/service/lib/agama/storage/iscsi/manager.rb
@@ -62,9 +62,9 @@ module Agama
 
         # Probes iSCSI.
         def probe
-          @probed = true
           probe_initiator
           probe_nodes
+          @probed = true
         end
 
         # Performs an iSCSI discovery.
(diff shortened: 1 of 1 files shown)


service/lib/agama/storage/iscsi/manager.rb
```diff
@@ -294,7 +294,7 @@ def disconnect_node(node)
           logger.info("Disconnecting iSCSI node: #{node.inspect}")
           adapter.logout(node).tap do |success|
             # Unlock the node if it was correctly disconnected.
-            @locked_targets&.delete(node.target) if success
+            @locked_nodes&.delete(node) if success
```
joseivanlopez: I think we cannot use #delete because the node objects change after each reprobe.

ancorgs: What's the problem with that? Do you expect any situation in which an old `node` object (from a previous probing that is now invalidated) is passed to `#disconnect_node`?

ancorgs: Now I understood the problem. Commit added to fix the problem.

PR #3124, opened 2026-02-04
Sort locales, keymaps and timezones

imobachgs:
## Problem

The l10n information is not sorted in any way.

- [bsc#1257497](https://bugzilla.suse.com/show_bug.cgi?id=1257497)

## Solution

Expose the information at API level already sorted. Implemented using AI.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/rust/agama-l10n/src/model/keyboard.rs
+++ b/rust/agama-l10n/src/model/keyboard.rs
@@ -40,15 +40,14 @@ impl KeymapsDatabase {
     }
 
     pub fn with_entries(data: &[Keymap]) -> Self {
-        let mut keymaps = data.to_vec();
-        keymaps.sort();
-        Self { keymaps }
+        let mut database = Self::new();
+        database.set_entries(data.to_vec());
+        database
     }
 
     /// Reads the list of keymaps.
     pub fn read(&mut self) -> anyhow::Result<()> {
-        self.keymaps = get_keymaps()?;
-        self.keymaps.sort();
+        self.set_entries(get_keymaps()?);
         Ok(())
     }
 
@@ -59,6 +58,12 @@ impl KeymapsDatabase {
     /// Returns the list of keymaps.
     pub fn entries(&self) -> &Vec<Keymap> {
         &self.keymaps
+    }
+
+    // Set the locales entries.
+    fn set_entries(&mut self, keymaps: Vec<Keymap>) {
+        self.keymaps = keymaps;
+        self.keymaps.sort();
     }
 }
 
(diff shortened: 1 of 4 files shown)


rust/agama-l10n/src/model/keyboard.rs
```diff
@@ -61,6 +62,27 @@ impl KeymapsDatabase {
...
+        let entries = vec![
+            Keymap::new("es".parse().unwrap(), "Spanish"),
+            Keymap::new("us".parse().unwrap(), "English (US)"),
+            Keymap::new("de".parse().unwrap(), "German"),
+        ];
+
+        let db = KeymapsDatabase::with_entries(&entries);
+        let keymaps = db.entries();
+
+        assert_eq!(keymaps[0].id.to_string(), "us");
```
jreidinger: to be honest, This is confusing as `us` is below `de` and `es`. I would use full name, so it is obvious that it is properly sorted.

rust/agama-l10n/src/model/locale.rs
```diff
@@ -56,6 +58,8 @@ impl LocalesDatabase {
     pub fn read(&mut self, ui_language: &str) -> anyhow::Result<()> {
         self.known_locales = Self::get_locales_list()?;
         self.locales = self.get_locales(ui_language)?;
+        self.locales.sort();
+        self.known_locales = self.locales.iter().map(|l| l.id.clone()).collect();
```
jreidinger: this method now gets more tricky, so I would add comment explaining that get_locales needs recent known locales for reading additional data. So it needs to be set twice in this method ( which also kind of smells for me. I kind of expect it should parameter for get_locales and not struct member that is at first not sorted and then later sorted to proper order.

rust/agama-l10n/src/model/locale.rs
```diff
@@ -198,4 +202,45 @@ mod tests {
...
+
+        let db = LocalesDatabase::with_entries(&entries);
+        let locales = db.entries();
+
+        assert_eq!(locales[0].language, "English");
+        assert_eq!(locales[0].territory, "United Kingdom");
+        assert_eq!(locales[1].language, "English");
+        assert_eq!(locales[1].territory, "United States");
+        assert_eq!(locales[2].language, "German");
+        assert_eq!(locales[3].language, "Spanish");
```
jreidinger, line 246: here it is good, obvious ordering.

rust/agama-l10n/src/model/locale.rs
imobachgs: I was actually wrong (I trusted the IA too much).

PR #3128, opened 2026-02-05
Add DASD to HTTP API

joseivanlopez:
Integrate DASD management in the new HTTP API.

### DIFF
--- a/rust/agama-software/src/service.rs
+++ b/rust/agama-software/src/service.rs
@@ -109,7 +109,7 @@ impl Starter {
         self
     }
 
-    const TARGET_DIR: &str = "/run/agama/software_ng_zypp";
+    const TARGET_DIR: &str = "/run/agama/zypp";
     // FIXME: it should be defined in a single place and injected where needed.
     const INSTALL_DIR: &str = "/mnt";
 
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,8 +1,14 @@
 -------------------------------------------------------------------
-Wed Feb 11 14:45:05 UTC 2026 - José Iván López González <jlopez@suse.com>
+Wed Feb 11 15:49:05 UTC 2026 - José Iván López González <jlopez@suse.com>
 
 - Add agama-s390 service and integrate DASD into the new HTTP API
   (gh#agama-project/agama#3128).
+
+-------------------------------------------------------------------
+Wed Feb 11 15:47:38 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Use "/run/agama/zypp" as directory for libzypp
+  (gh#agama-project/agama#3154).
 
 -------------------------------------------------------------------
 Wed Feb 11 13:48:37 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>


rust/agama-utils/src/api/s390/config.rs
```diff
@@ -0,0 +1,34 @@
...
+use serde_with::skip_serializing_none;
+
+#[skip_serializing_none]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, Merge, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+#[merge(strategy = merge::option::recurse)]
+/// s390 configuration.
+pub struct Config {
+    /// Configuration of the DASD devices.
+    pub dasd: Option<RawConfig>,
```
jreidinger, line 33: Well, I have to say I really dislike this. Config struct is important part of API and here is basically nothing, depending on what dbus service provides. Basically throwing away all rust guarantees and also making it quite hard create such config e.g. even for testing purpose and you always have to use json.

rust/agama-utils/src/api/s390/system_info.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use serde::Serialize;
+use serde_json::Value;
+use serde_with::skip_serializing_none;
+
+#[skip_serializing_none]
+#[derive(Clone, Debug, Serialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct SystemInfo {
+    pub dasd: Option<Value>,
```
jreidinger, line 29: and similar here. Basically no info how it looks like or how to construct it. Again even in test it forces to use json text as it is not possible to use API without json.

rust/agama-utils/src/arch.rs
```diff
@@ -111,4 +120,16 @@ mod tests {
...
         assert_eq!(Arch::current().unwrap(), Arch::X86_64);
     }
+
+    #[cfg(target_arch = "s390x")]
+    #[test]
+    fn test_arch_is_s390() {
+        assert!(Arch::is_s390());
+    }
+
+    #[cfg(target_arch = "x86_64")]
```
jreidinger:
```suggestion
   #[cfg(not(target_arch = "s390x"))]
```

rust/Cargo.toml
```diff
@@ -22,9 +23,8 @@ members = [
     "suseconnect-agama/suseconnect-agama-sys",
     "xtask",
     "zypp-agama",
-    "zypp-agama/zypp-agama-sys", 
     "zypp-agama/zypp-agama-sys"
-]
+    ]
```
jreidinger: this indentation looks strange.

rust/agama-utils/src/api/s390/config.rs
joseivanlopez: This was a decision already taken to directly pass the json from the dbus API. It is done so for storage, iscsi, dasd, etc.

joseivanlopez: The validation is done with the schema. Anyway, I think dasd, zfcp and iscsi could be totally moved to rust in the future. Storage would be more complex.

rust/agama-utils/src/api/s390/system_info.rs
joseivanlopez: Yes, same reason than above. For now we provide schema definitions for that data and we don't want to generate the very same json twice, at dbus level and at http level.

rust/agama-utils/src/api/s390/config.rs
jreidinger: sadly validation is still runtime done by checking schema against config, but not against code during compilation, which is what rust doing. So if someone breaks backward compatibility, the config will pass validation and then ruby code will fail.

joseivanlopez: Yes, we are aware of that. Although the ruby code will fail too if the schema changes.

jreidinger (approved):

rust/agama-s390/src/dasd/monitor.rs
```diff
@@ -139,17 +139,19 @@ impl Monitor {
         Ok(())
     }
 
-    fn handle_progress_changed(&self, signal: ProgressChanged) -> Result<(), Error> {
+    async fn handle_progress_changed(&self, signal: ProgressChanged) -> Result<(), Error> {
         let args = signal.args()?;
         let progress_data = serde_json::from_str::<ProgressData>(args.progress)?;
         self.progress
-            .cast(progress::message::SetProgress::new(progress_data.into()))?;
+            .call(progress::message::SetProgress::new(progress_data.into()))
```
jreidinger, line 146: why it needs to be changed? I do not see reason in commit message. Basically original idea is that for progress report, you do not need to wait to be send and just send it and continue to delay if not needed.

joseivanlopez: I did it to guarantee that the signals arrive in order. Otherwise, ProgressFinished could come before than the last ProgressChanged.

PR #3130, opened 2026-02-06
Run SetConfig on a separate task

imobachgs:
## Problems

- It is not possible to load a profile when it does not contain product definition ([bsc#1257067](https://bugzilla.suse.com/show_bug.cgi?id=1257067), [bsc#1257082](https://bugzilla.suse.com/show_bug.cgi?id=1257082)).
- Bootloader SetConfig should be executed *after* storage SetConfig.
- Using "agama questions" in a pre-script causes Agama to hang.

## Solution

- Allow not setting a product in the profile.
- Move the SetConfig stuff to a separate task.
- Run pre-scripts only once to prevent potential infinite loops. Otherwise, we can get into an infinite loop when using "agama config load" (unexpected) or "agama questions" in a pre-script.

## Testing

- Added a new unit test
- Tested manually

## Follow-up

* Reset the proposal and software configuration when issuing a PUT with no product.
* The changes by `agama questions mode non-interactive` will get applied after loading the configuration.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -2,7 +2,7 @@
 Fri Feb  6 07:38:38 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Allow loading a profile with no product (bsc#1257067 and bsc#1257082).
-- Make the bootloader proposal after the storage one is ready.
+- Make the bootloader proposal after the storage one is ready (bsc#1257530).
 - Run pre-scripts only once to prevent potential infinite loops.
 
 -------------------------------------------------------------------


imobachgs:
> note `Bootloader SetConfig should be executed after software SetConfig.` is not correct as it is _storage_ and not software.

Yes, right, I have updated the description.

rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Feb  6 07:38:38 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Allow loading a profile with no product (bsc#1257067 and bsc#1257082).
+- Make the bootloader proposal after the storage one is ready.
```
jreidinger: please add bsc https://bugzilla.suse.com/show_bug.cgi?id=1257530

PR #3137, opened 2026-02-09
Installer: configure libzypp through drop-in config snippet when supported

DimStar77:
## Problem

Libypp, from version 17.38.0 on, supports UAPI style configuration.

- Build fail against new libzypp version: https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:K/agama-installer:openSUSE/images/x86_64

## Solution

Change the build of the live image to no lonber patch /etc/zypp/zypp.conf,
which does not exist in this case anymore (distro default moved to /usr/etc),
but rather create a drop-in configuration in /etc/zypp/zypp.conf.d.
If the image is built against a libzypp not yet supporting UAPI conf, continue
with patching zypp.conf directly


jreidinger (approved):
looks good

PR #3138, opened 2026-02-09
Addons patterns

jreidinger:
## Problem

In master now after registering extension, no patterns from that addon is shown.

- trello https://trello.com/c/tDaI9tu0/5516-add-patterns-from-extensions

## Solution

Add to libzypp API to list all patterns and then do filtering on rust side. There is room for optimization ( and even benchmark is created for this case ), but looks like so far it is not big issue even with tumbleweed big list of patterns ( 215 when I run benchmark and still listing them around 1.2 ms ).

- [x] remove old patterns_info method

## Testing

- *Tested manually*

- [x] with SCC proxy and HA addon (see screenshot from it)
- [x] with TW image to ensure it is still filtered

## Screenshots

ha_addon

### DIFF
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -652,9 +652,7 @@ impl ZyppServer {
             .iter()
             .map(|p| p.name())
             .collect();
-
         let repositories = zypp.list_repositories()?;
-
         let zypp_patterns = zypp.list_patterns()?;
 
         Ok(zypp_patterns.into_iter().filter(move |p| {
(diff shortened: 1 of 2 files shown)


rust/agama-software/src/zypp_server.rs
```diff
@@ -641,14 +641,54 @@ impl ZyppServer {
...
+
+        let zypp_patterns = zypp.list_patterns()?;
+
+        Ok(zypp_patterns.into_iter().filter(move |p| {
+            // lets explain here logic for user selectable patterns
+            // if pattern is listed in product pattern names then use it
+            // else include only patterns coming from repository that is
+            // NOT predefined as agama-* one neither from repository
+            // added by base product registration
+            if product_pattern_names.contains(&p.name.as_str()) {
```
imobachgs, line 664: The code looks a little bit hard to follow. To start with, you could avoid the "else" by just returning earlier (you are using `return` anyway a few lines below).

rust/zypp-agama/benches/list_patterns.rs
```diff
@@ -0,0 +1,52 @@
+use std::fs;
+
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn criterion_benchmark(c: &mut Criterion) {
+    let _ = fs::remove_dir_all("/tmp/zypp");
```
imobachgs, line 6: For temporary directories we use `tempfile` usually, although in this situation I guess no collision (when running in parallel) is possible.

rust/zypp-agama/examples/list_patterns.rs
```diff
@@ -0,0 +1,48 @@
+use std::fs;
```
imobachgs, line 1: This one is basically the same code that the benchmark, right?

rust/zypp-agama/src/lib.rs
```diff
@@ -60,6 +61,19 @@ pub struct PatternInfo {
     pub selected: ResolvableSelected,
 }
 
+// TODO: should we add also e.g. serd serializers here?
```
imobachgs, line 61: They could be behind a feature flag if we want to.

imobachgs (approved):

rust/zypp-agama/benches/list_patterns.rs
jreidinger: well, runnning benchmark in parallel sounds like really bad idea :)

rust/zypp-agama/examples/list_patterns.rs
jreidinger: yes, because benchmark segfault I need to debug it and it is better to not do it when it tries to run it 100 times :)
but I found the example useful in general for integration testing as it is kind of system independent.

PR #3139, opened 2026-02-09
Make sure SetConfig finishes before installing

imobachgs:
## Problem

After moving the "set config" logic to a separate thread, nothing prevents the "install" to run before "set config".

## Solution

Introduce a `TasksRunner` which makes sure that `SetConfigAction` and `InstallAction` do not try to run at the same time (and in the same order they are requested).

## Testing

- Tests keep working.

### DIFF
--- a/rust/agama-manager/src/tasks/runner.rs
+++ b/rust/agama-manager/src/tasks/runner.rs
@@ -310,9 +310,6 @@ impl SetConfigAction {
         //
         // Set the configuration for each service
         //
-        self.progress
-            .call(progress::message::Next::new(Scope::Manager))
-            .await?;
         self.security
             .call(security::message::SetConfig::new(config.security.clone()))
             .await?;


rust/agama-manager/src/tasks.rs
```diff
@@ -0,0 +1,317 @@
...
+        product: Option<Arc<RwLock<ProductSpec>>>,
+    },
+}
+
+/// Runs long tasks in background.
+///
+/// This struct takes care of running the logic to set the configuration (Task::SetConfig]) and to
+/// install the system (Task::Install) in background. Additionally, it makes sure that only one of
+/// those tasks run at the same time and in the same order they are received by using a mpsc
+/// channel.
```
jreidinger: I am thinking if it make sense to have also "Probe" task...on common install it is usually done in SetConfig as there are more products, but for SLES on s390 there is just single product and it can potentially help to ensure that initial probe is done first and then SetConfig is called.

rust/agama-manager/src/tasks.rs
```diff
@@ -0,0 +1,317 @@
...
+                    storage: self.storage.clone(),
+                    files: self.files.clone(),
+                    progress: self.progress.clone(),
+                    users: self.users.clone(),
+                };
+                action.run().await;
+            }
+
+            Task::SetConfig { product, config } => {
+                let action = SetConfigAction {
```
jreidinger: should we check stage here? Like if someone calls setconfig after install, it would be nice to at least log it as it probably won't work.

rust/agama-manager/src/tasks.rs
```diff
@@ -0,0 +1,317 @@
...
+    Install,
+    SetConfig {
+        config: Config,
+        product: Option<Arc<RwLock<ProductSpec>>>,
+    },
+}
+
+/// Runs long tasks in background.
+///
+/// This struct takes care of running the logic to set the configuration (Task::SetConfig]) and to
```
jreidinger: should this actor implement Progress reporting? It kind of makes sense to me. This way it ensure we are ensure that we are always reporting that something is happening.

rust/agama-manager/src/tasks.rs
imobachgs: Sure, we can move this later. And most probably it applies to all "actions".

rust/agama-manager/src/tasks.rs
imobachgs: I agree, but I wonder if I am reporting too many steps now :-)

PR #3143, opened 2026-02-11
refactor(web): restore interface for managing DASD from web client

joseivanlopez:
Bring back the DASD user interface fully adapted to the HTTP v2 API and to the latest web client look&feel (https://trello.com/c/xdvueNDg/, private link)

> [!WARNING]
> 
> The format progress has not been adapted as part of this PR and will be addressed in a follow up one.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Feb 26 16:44:32 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Add schema for DASD config (gh#agama-project/agama#3143).
+
+-------------------------------------------------------------------
 Wed Feb 25 08:13:55 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 
 - Optimize writing software state to reduce time spend when
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Feb 26 16:47:39 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Allow reactivating a DASD if enabling DIAG failed
+  (gh#agama-project/agama#3143).
+
+-------------------------------------------------------------------
 Wed Feb 18 20:28:02 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 
 - Add support for updateNvram bootloader key (jsc#PED-14224)
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Feb 26 16:43:44 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Restore the interface for managing DASD in the web client
+  (gh#agama-project/agama#3143).
+
+-------------------------------------------------------------------
 Fri Feb 20 15:00:44 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Fix the language change (gh#agama-project/agama#3197).


web/src/openapi/config/dasd.ts
```diff
@@ -0,0 +1,30 @@
...
+ * DASD config.
+ */
+export interface Config {
+  /**
+   * List of DASD devices.
+   */
+  devices?: Device[];
+}
+export interface Device {
+  channel: string;
```
dgdavid, line 17: NP: might we document that this is kind of the _id_ of the device? If not, at least add some documentation would be nice. I can do it, but ideas are welcome.

web/src/model/config/dasd.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function addDevice(config: Config, device: Device): Config {
+  return { devices: [...(config.devices || []), device] };
+}
+
+function removeDevice(config: Config, channel: string): Config {
+  const devices = config.devices.filter((d) => d.channel !== channel);
+  return { ...config, devices };
+}
```
dgdavid: ⚠️  Note that these methods may unintentionally break the returned config object in the future if `Config` gets more properties. They are currently too coupled to the current `Config` type structure.

When working with object copies, always return a new object preserving all (potential) existing properties and the ones you wanna overrides. This makes the code more resilient to future changes in the data model.

```js
return { ...config, devices: add(config.devices, newDevice)}
```

```js
return { ...config, devices: remove(config.devices, noLongerNeededDevice)}
```
The key in the examples above is spreading `...config` first.

web/src/openapi/system/dasd.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface System {
+  /**
+   * DASD devices
+   */
+  devices: Device[];
+}
+export interface Device {
+  channel: string;
+  deviceName: string;
+  type: string;
```
dgdavid, line 19: NP: Wondering why we have moved from "deviceType" to simply "type" but kept "deviceName". I would use simply "name" as well.

web/src/hooks/model/config/dasd.ts
```diff
@@ -0,0 +1,60 @@
...
+import { useSuspenseQuery } from "@tanstack/react-query";
+import { configQuery } from "~/hooks/model/config";
+import { patchConfig, Response } from "~/api";
+import dasd from "~/model/config/dasd";
+import type { Config, DASD } from "~/model/config";
+
+const selectConfig = (data: Config | null): DASD.Config => data?.dasd;
+
+function useConfig(): DASD.Config | null {
+  const { data } = useSuspenseQuery({
```
dgdavid, line 56: Note for myself: this module structure make not possible (or at least so difficult) to create a test for these useAddDevice and useRemoveDevice hooks because they live in the same module than useConfig, which makes not possible to mock it successfully.

web/src/model/config/dasd.ts
dgdavid: Above is an outdated comment. To see it in context, refer to code before commit https://github.com/agama-project/agama/pull/3143/commits/cbb75252762e10b18d63cab19775e9057fac804b

web/src/hooks/model/config/dasd.ts
dgdavid: Commit https://github.com/agama-project/agama/pull/3143/commits/24a522ba0b3da74166c937f14dae2b6423c4a84d added a way to workaround "testability" of these modules by directly mocking tanstack-query. Re-evaluation is needed.

web/src/openapi/system/dasd.ts
joseivanlopez: Because we are using `DeviceName` for the name of the associated linux device.

web/src/openapi/config/dasd.ts
joseivanlopez: If you miss some documentation, then I would suggest to add it as `description:` in the schema, so it is automatically exported as comment here.

web/src/components/storage/dasd/DASDTable.tsx
```diff
@@ -477,7 +492,7 @@ const createColumns = () => [
   {
     // TRANSLATORS: table header for a DASD devices table
     name: _("Status"),
-    value: (d: Device) => d.status,
+    value: (d: Device) => STATUS_OPTIONS[d.status],
```
dgdavid, line 551: Good catch!

web/src/components/storage/ConnectedDevicesMenu.tsx
```diff
@@ -26,12 +26,13 @@ import { activateStorageAction } from "~/api";
...
 import Icon from "~/components/layout/Icon";
 import MenuButton from "~/components/core/MenuButton";
+import { useSystem as useDASDSystem } from "~/hooks/model/system/dasd";
 import { _ } from "~/i18n";
 
 export default function ConnectedDevicesMenu() {
   const navigate = useNavigate();
   const isZFCPSupported = false;
-  const isDASDSupported = false;
+  const dasdSystem = useDASDSystem();
```
dgdavid, line 35: I don't like this because the same reason I stated we should not return undefined but an empty object instead. Solution would be just going for isEmpty(dasdSystem) instead, but let's go ahead as it is by now until we finally change these undefined or null values in all similar hooks.

web/src/openapi/config/dasd.ts
dgdavid: Good to know. Let's see if I find a simple short sentence for this one :)

web/src/components/storage/ConnectedDevicesMenu.tsx
dgdavid, line 35: We have to mock that useSystem in the ConnectedDevicesMenu test component to avoid it failing because the test stuck in a suspense

web/src/components/storage/dasd/DASDTable.tsx
```diff
@@ -169,23 +169,37 @@ const buildActions = ({ devices, addOrUpdateDevices, dispatcher }: DASDActionsPr
...
+        addOrUpdateDevices(
+          devices.map((d): ConfigDevice => ({ channel: d.channel, state: "active", diag: true })),
+        ),
     },
     {
       title: _("Set DIAG off"),
-      onClick: () => addOrUpdateDevices(devices.map((d) => ({ channel: d.channel, diag: false }))),
+      onClick: () =>
+        addOrUpdateDevices(
+          devices.map((d): ConfigDevice => ({ channel: d.channel, state: "active", diag: false })),
```
dgdavid, line 224: Ok, I'm going to fix a test that is failing because of this changes and I need to have fixed for testing other changes I'm working on.

web/src/hooks/model/config/dasd.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Used by useSuspenseQuery's select option to transform the query result.
+ * Returns undefined when data is undefined or when dasd property is not present.
+ *
+ * @see {@link https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#select TanStack Query Select}
+ * @see {@link https://tkdodo.eu/blog/react-query-selectors-supercharged#what-is-select Query Selectors Supercharged}
+ *
+ * FIXME: Read todo note below.
+ * @todo Consider returning an empty object ({}) instead of undefined to simplify
+ * consuming code and eliminate the need for fallback checks throughout the codebase.
+ */
```
dgdavid: We discussed this in a meeting and while we haven't reached a final conclusion but different point of view, the direction we're leaning toward is using `null` to represent "nothing found", aligning with language intent despite its well-known `typeof` quirk (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#description). Alongside this, we'd rely on explicit "defaults" (imported from a shared utility module) to reconstruct the expected shape from scratch when needed.

Not defaulting to an empty or default object does make destructuring slightly more harder to write and follow, but beyond small drawbacks the final solution should focus providing a consistent, predictable pattern as key benefits, as well as avoids mixing `null` and `undefined` return types, which can silently become the root cause of unexpected crashes when accessing properties down the render chain (e.g., `null` passing through a `something !== undefined` check and then causing a property access to blow up).

For now, I'm removing these comments to keep the PR diff clean. We'll revisit and document the agreed pattern once the team lands on it.

web/src/components/storage/ConnectedDevicesMenu.tsx
dgdavid: See https://github.com/agama-project/agama/pull/3143#discussion_r2859905366

PR #3145, opened 2026-02-11
Sort products alphabetically

imobachgs:
- Products are sorted alphabetically, making sure the list is always consistent.

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Feb 11 13:58:01 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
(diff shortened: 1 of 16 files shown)


rust/agama-utils/src/products.rs
```diff
@@ -176,7 +177,10 @@ impl Registry {
                     modes,
                 }
             })
-            .collect()
+            .collect();
+
+        products.sort_by(|a, b| a.name.to_lowercase().cmp(&b.name.to_lowercase()));
```
lslezak, line 182: Does that sort according to the current locale or is it a plain sort (C-like locale)?

imobachgs: Plain unicode. We are just ordering the product names, so I would expect that to be enough.

PR #3146, opened 2026-02-11
Add modes to SLES for SAP

imobachgs:
Add "Standard" and "Immutable" modes to SLES for SAP.

- Standard: uses the original SLES for SAP definition.
- Immutable: changes `enhanced_base` with `immutable_base` and uses the same storage layout than SLES.


angelabriel:
Hi @imobachgs 
regarding bsc#1258042 and the _For 'Immutable' mode share the same content between SLES and SLES4SAP_ (comment 8) we should move the `sles_sap_base_sap_server` pattern to the 'Standard' mode in the same way as done with `enhanced_base` pattern.

PR #3155, opened 2026-02-11
Split storage finish

jreidinger:
## Problem

Storage configuration including bootloader one is run after post install scripts. This can interfere with it and overwrite user changes from script.

bsc: https://bugzilla.suse.com/show_bug.cgi?id=1257791

## Solution

Split storage config write into finish part and then finalizing installation into umount part and call the first one before post install script and later after it.

## Testing

- *Tested manually*

### DIFF
--- a/service/lib/agama/storage/manager.rb
+++ b/service/lib/agama/storage/manager.rb
@@ -27,7 +27,7 @@ require "agama/storage/bootloader"
 require "agama/storage/callbacks"
 require "agama/storage/configurator"
 require "agama/storage/finisher"
-require "agama/storage/finalize"
+require "agama/storage/umounter"
 require "agama/storage/iscsi/manager"
 require "agama/storage/proposal"
 require "agama/with_locale"
@@ -149,7 +149,7 @@ module Agama
 
       # Performs the final umount of system before reboot.
       def umount
-        Finalizer.new(logger).run
+        Umounter.new(logger).run
       end
 
       # Storage proposal manager
--- a/service/lib/agama/storage/umounter.rb
+++ b/service/lib/agama/storage/umounter.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2023-2025] SUSE LLC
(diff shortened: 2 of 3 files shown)


doc/dbus/bus/org.opensuse.Agama.Storage1.bus.xml
```diff
@@ -96,6 +96,8 @@
     </method>
     <method name="Finish">
     </method>
+    <method name="Umount">
```
joseivanlopez, line 99: Maybe we don't need these files anymore because the D-Bus API is internal. Anyway, removing that files is out of scope of this PR.

service/lib/agama/storage/finalize.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2023-2025] SUSE LLC
```
joseivanlopez: NP: [2026]

service/lib/agama/storage/manager.rb
```diff
@@ -27,6 +27,7 @@
 require "agama/storage/callbacks"
 require "agama/storage/configurator"
 require "agama/storage/finisher"
+require "agama/storage/finalize"
```
joseivanlopez: The file name does not match the class name : *finalize* vs `Finalizer`. I would use other name for the class because having `Finisher` and `Finalizer` is confusing. What about `Umounter`?

doc/dbus/bus/org.opensuse.Agama.Storage1.bus.xml
jreidinger: yeap, it still can be used for generating rust code for proxies, but to be honest with trivial methods, I write it manually :)

service/lib/agama/storage/manager.rb
jreidinger: fine for me. I want to emphasize there a bit, that finish is consistent with rest and finalizer should be final action, but umounter is also good.

PR #3158, opened 2026-02-12
Recalculate bootloader

joseivanlopez:
## Problem

The UI uses a private endpoint to calculate the storage proposal, and bootloader is not recalculated.

## Solution

Force a call to bootloader after applying a storage config model.

Note: this is a temporary solution. A better approach will be implemented for the next release.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Feb 12 14:43:48 UTC 2026 - José Iván López González <jlopez@suse.com>
+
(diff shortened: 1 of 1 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Feb 12 13:43:48 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Recalulate bootloader proposal after applying a storage config
+  model (gh#agama-project/agama#3158).
```
jreidinger: please mention bsc that ancor reported in this changelog entry.

PR #3164, opened 2026-02-13
Reset bootloader cache

jreidinger:
## Problem

The bootloader config is cached in bootloader factory as in yast it allows switching between bootloaders without loosing previous config. This is not case for agama, where we always need to apply config from scratch.

## Solution

Clear cache in Agama to ensure it always use proper storage settings.

## Testing

- *Tested manually*


ancorgs (approved):
LGTM

PR #3165, opened 2026-02-13
Resolve relative URLs

imobachgs:
- Add some logging when running scripts.
- Do not crash when user files cannot be transferred, just log the error.
- Resolve relative URLs doing `config generate`.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -3,7 +3,7 @@ Fri Feb 13 08:47:07 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Resolve the relative URLs (bsc#1258118).
 - Do not finish the installation as "failed" if it was not possible
-  to fetch the user files (bsc#1258188).
+  to fetch the user files (bsc#1258118).
 
 -------------------------------------------------------------------
 Thu Feb 12 14:43:48 UTC 2026 - José Iván López González <jlopez@suse.com>


rust/agama-utils/src/api/files/config.rs
```diff
@@ -38,6 +38,18 @@ pub struct Config {
     pub scripts: Option<ScriptsConfig>,
 }
 
+impl Config {
+    pub fn resolve_urls(&mut self, base_uri: &Uri<String>) -> Result<(), FileSourceError> {
+        resolve_urls_for(&mut self.files, base_uri)?;
+
+        if let Some(scripts) = &mut self.scripts {
```
jreidinger, line 45: what about relative urls in files location?

jreidinger: ah, it is files at top..it kind of confuse me.

rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Feb 13 08:47:07 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Resolve the relative URLs (bsc#1258118).
+- Do not finish the installation as "failed" if it was not possible
+  to fetch the user files (bsc#1258188).
```
jreidinger: this for me is questionable change. If user want to have specified file on system and it is not possible to add it there, why it should not be failed state?

imobachgs: I do not agree; the whole installation should not fail because of that. For beta 2, I would most probably display a message (using the questions mechanism). Does it make sense?

jreidinger: question can help, as it means that user need to explicitly use answer to skip it in unattended....another option and maybe better would be new flag in files like "critical" or "mandatory" that decide behavior explicitly.

imobachgs: Sure, but I would try to not over-complicate things. Let's discuss about that for beta 2.

PR #3185, opened 2026-02-18
Bunch of bootloader jira features

jreidinger:
## Problem

We are asked to implement few fine tuning of bootloader either for live system or for target one.

For live system we are asked to

- increase timeout to 60 seconds same as on SLE15 (https://jira.suse.com/browse/PED-14986)
- drop boot from hard disk everywhere except legacy x86_64 (https://jira.suse.com/browse/PED-13766)
- unify kernel location for easier network boot (https://jira.suse.com/browse/PED-14303)

and for target system

- add support for update_nvram option to have a way to avoid either broken firmwares or network setup (https://jira.suse.com/browse/PED-14224)

### DIFF
--- a/rust/agama-manager/src/tasks/runner.rs
+++ b/rust/agama-manager/src/tasks/runner.rs
@@ -421,7 +421,6 @@ impl SetConfigAction {
                 self.progress
                     .call(progress::message::Next::new(Scope::Manager))
                     .await?;
-                tracing::info!("setting bootloader config {:?}", &config.bootloader);
                 self.bootloader
                     .call(bootloader::message::SetConfig::new(
                         config.bootloader.clone(),


live/src/agama-installer.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Feb 18 20:32:37 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- increase medium timeout to 60 seconds (jsc#PED-14986)
```
imobachgs:
```suggestion
- Increase medium timeout to 60 seconds (jsc#PED-14986)
```

live/src/agama-installer.changes
```diff
@@ -1,3 +1,11 @@
...
+- Unify kernel and initramfs location (jsc#PED-14303)
+- drop “Boot from Disk” on all archs expect legacy x86_64
```
imobachgs:
```suggestion
- Drop “Boot from Disk” on all archs except legacy x86_64
```

rust/agama-manager/src/tasks/runner.rs
```diff
@@ -421,6 +421,7 @@ impl SetConfigAction {
                 self.progress
                     .call(progress::message::Next::new(Scope::Manager))
                     .await?;
+                tracing::info!("setting bootloader config {:?}", &config.bootloader);
```
imobachgs: This is the only step that has a `tracing::info!` call. I would like to be consistent. Perhaps we could move the "tracing::info!" call to the progress service, so every step would be logged.

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Feb 18 20:28:02 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Add support for updateNvram bootloader key (jsc#PED-14224)
+- Add support for autoyast key bootloader->update_nvram conversion
```
```suggestion
- Add support for AutoYaST key bootloader->update_nvram conversion
```

imobachgs (commented):
Just minor things. Otherwise, LGTM.

rust/agama-manager/src/tasks/runner.rs
jreidinger: well, it was debugging line. I think I can remove it now.

imobachgs (approved):

PR #3192, opened 2026-02-19
Fixes in the boot menu and image build

lslezak:
## Problem

- The image build fails when calling `image-janitor` tool
- The boot from hard disk in the legacy BIOS mode does not work (black screen)

## Fixes

- Fixed boot from hard disk in the legacy BIOS mode, the `exit` command works only in EFI mode (returns back to the firmware), in BIOS boot the machine halts with black screen
- Added missing configuration options for the `image-janitor` tool

## Additional improvements

- When booting from hard disk in the BIOS mode check if the first disk contains a valid MBR sector. If the disk is completely empty it displays an error.
  agama-not-bootable-bios-disk
- Added "p" keyboard shortcut to interactively configure the root password for the installer live system - for me it is annoying to specify it on the boot command line, let's try something easier. This is experimental feature, let's see how it works for others. If it is not widely accepted we can remove it later.
  agama-grub-set-root-password
- Added support for serial console, when the system has a serial console attached the boot menu appears there as well. Tested with virtual serial port in a VirtualBox VM:
  agama-boot-serial-console
- Disable plymouth graphical boot when the grub boot menu is switched to text mode with the "t" keyboard shortcut
agama-grub-textmode-bios

## Testing

- Tested in a VM, x86_64 only. Hopefully Aarch64 behaves similarly, I did the same changes there (except the PC BIOS part)
- Tested with both EFI and legacy BIOS systems

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -12,6 +12,8 @@ Thu Feb 19 12:55:14 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
 - Fixed failed image build because of missing configuration options
   for the "image-janitor" tool
 
+  (gh#agama-project/agama#3192)
+
 -------------------------------------------------------------------
 Wed Feb 18 20:32:37 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
 


live/config-cdroot/fix_bootconfig.x86_64
```diff
@@ -74,58 +72,104 @@ fi
...
+    # this detects that a serial port is present but that does not mean a serial
+    # console is attached there so we cannot for example disable plymouth
+else
+    terminal_input console
+    terminal_output gfxterm
+fi
+
 if [ "\${grub_platform}" != "efi" ]; then
-  menuentry "Boot from Hard Disk" --class opensuse --class gnu-linux --class gnu --class os {
-    exit
```
jreidinger, line 81: strange that for me it works. Do you have disk as boot device in BIOS? if so, then exit should continue to boot with another device in boot order ( at least it works for me when I am testing it in qemu )

lslezak: Interesting... Yes, the boot order is set to "Optical" then "Harddisk".  Without any medium it boots the system from disk (SLES 16.0). But when booting from DVD and using "Boot from Hard Disk" grub option I get just a black screen. When using `exit` from command line it gets stuck as well.

However, with the chainloader it works fine.

agama-bios-boot-exit

jreidinger (approved):

PR #3194, opened 2026-02-20
Fips pattern

jreidinger:
## Problem

When there is `fips=1` on kernel cmdline, agama do not propose fips pattern like it did on 16.0.

bsc: https://bugzilla.suse.com/show_bug.cgi?id=1258306
trello: https://trello.com/c/OLeZ209V/5536-1-1258306-add-fips-pattern-if-the-option-is-set-in-the-kernels-cmdline

## Solution

Implement the fix. It involves also merging agama and kernel command line params, so we can query both.

## Testing

- *Added a new unit test*
- *Tested manually* ( surprisingly TW kernel supports fips, so it can be tested fully on TW )

### DIFF
--- a/rust/agama-utils/src/kernel_cmdline.rs
+++ b/rust/agama-utils/src/kernel_cmdline.rs
@@ -75,7 +75,7 @@ impl KernelCmdline {
         let mut args = self.0;
         for (key, value) in other.0 {
             args.entry(key)
-                // this modify is just for theoreticall correctness as in reality we are merging kernel and agama
+                // this modify is just for theoretical correctness as in reality we are merging kernel and agama
                 // args and they are exclusive
                 .and_modify(|v| v.extend(value.clone()))
                 .or_insert(value);


rust/agama-utils/src/kernel_cmdline.rs
```diff
@@ -57,23 +62,35 @@ impl KernelCmdline {
...
                 .or_insert(vec![value.to_string()]);
         }
         Self(args)
     }
 
+    pub fn merge(self, other: Self) -> Self {
+        let mut args = self.0;
+        for (key, value) in other.0 {
+            args.entry(key)
+            // this modify is just for theoreticall correctness as in reality we are merging kernel and agama
```
imobachgs:
```suggestion
            // this modify is just for theoretical correctness as in reality we are merging kernel and agama
```

imobachgs (commented):
Please, fix rustfmt complains.

PR #3197, opened 2026-02-20
Fix UI language change

imobachgs:
## Problem

Changing the UI language does not work when the checkbox to change the target language is not checked.

## Solution

- Use the suspense API.
- For reloading the application when the language changes.

## Testing

- Adapted unit tests
- Tested manually

### DIFF
--- a/web/src/context/installerL10n.tsx
+++ b/web/src/context/installerL10n.tsx
@@ -138,7 +138,7 @@ async function loadTranslations(locale: string): Promise<string> {
           return agama.language.replace("-", "_");
         })
         .catch(() => {
-          if (locale !== "en-US" && locale) {
+          if (locale && locale !== "en-US") {
             console.error("Cannot load frontend translations for", locale);
           }
           // reset the current translations (use the original English texts)


web/src/components/core/InstallerL10nOptions.test.tsx
```diff
@@ -93,7 +93,7 @@ jest.mock("~/hooks/model/status", () => ({
 }));
 
 jest.mock("~/context/installerL10n", () => ({
-  ...jest.requireActual("~/context/installerL10n"),
```
dgdavid, line 96: I always have the doubt about when still "requireActual" and when not.

web/src/context/installerL10n.tsx
```diff
@@ -115,27 +116,34 @@ function findSupportedLanguage(languages: Array<string>): string | undefined {
...
         `../po/po.${po}`
       )
-        .then((m) => agama.locale(m.default))
+        .then((m) => {
+          agama.locale(m.default);
+          return agama.language.replace("-", "_");
+        })
         .catch(() => {
-          if (locale !== "en-US") {
+          if (locale !== "en-US" && locale) {
```
dgdavid: NP: I usually write it the other way around `if (locale && locale !== "en-US")`, although I know it does not matter.

dgdavid (approved):

PR #3204, opened 2026-02-23
Query available storage

mchf:
## Problem

When disabled questions, autoinstallation can infinite block.

- [bsc#1258466](https://bugzilla.suse.com/show_bug.cgi?id=1258466)

## Solution

Extended fix https://github.com/agama-project/agama/blob/master/rust/agama-manager/src/service.rs#L635 even for Storage service

### DIFF
--- a/rust/agama-manager/src/service.rs
+++ b/rust/agama-manager/src/service.rs
@@ -637,7 +637,11 @@ impl MessageHandler<message::GetSystem> for Service {
         let network = self.network.get_system().await?;
 
         let s390 = if let Some(s390) = &self.s390 {
-            Some(s390.call(s390::message::GetSystem).await?)
+            if self.is_service_available(Scope::Storage).await? {
+                Some(s390.call(s390::message::GetSystem).await?)
+            } else {
+                None
+            }
         } else {
             None
         };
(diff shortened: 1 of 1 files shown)


rust/agama-manager/src/service.rs
```diff
@@ -628,7 +628,12 @@ impl MessageHandler<message::GetSystem> for Service {
             Default::default()
         };
 
-        let iscsi = self.iscsi.call(iscsi::message::GetSystem).await?;
+        let iscsi = if self.is_service_available(Scope::ISCSI).await? {
```
joseivanlopez: I would check the storage scope instad of iscsi, because only storage is able to raise questions.

rust/agama-manager/src/service.rs
```diff
@@ -628,7 +628,12 @@ impl MessageHandler<message::GetSystem> for Service {
...
+            self.iscsi.call(iscsi::message::GetSystem).await?
+        } else {
+            None
+        };
+
         let network = self.network.get_system().await?;
 
         let s390 = if let Some(s390) = &self.s390 {
```
joseivanlopez, line 639: Please, add the check for s390 too. And again, I would check the storage scope.

PR #3207, opened 2026-02-23
Address clippy errors

imobachgs:
## Problem

Address clippy complains.

## Solution

Use Gemini to fix the given issues so we need to go careful in the review process.

### DIFF
--- a/rust/agama-network/src/nm/client.rs
+++ b/rust/agama-network/src/nm/client.rs
@@ -228,9 +228,7 @@ impl<'a> NetworkManagerClient<'a> {
 
             match connection_from_dbus(settings) {
                 Ok(mut connection) => {
-                    let state = states
-                        .get(&connection.id)
-                        .map(|s| NmConnectionState(*s));
+                    let state = states.get(&connection.id).map(|s| NmConnectionState(*s));
                     if let Some(state) = state {
                         connection.state = state.try_into().unwrap_or_default();
                     }
--- a/rust/agama-cli/src/auth_tokens_file.rs
+++ b/rust/agama-cli/src/auth_tokens_file.rs
@@ -43,9 +43,7 @@ impl AuthTokensFile {
     /// Default path for the tokens file in user's home directory.
     pub fn default_path() -> io::Result<PathBuf> {
         let Some(path) = home::home_dir() else {
-            return Err(io::Error::other(
-                "Cannot find the user's home directory",
-            ));
+            return Err(io::Error::other("Cannot find the user's home directory"));
         };
 
         Ok(path.join(AUTH_TOKENS_FILE))
--- a/rust/agama-files/src/service.rs
+++ b/rust/agama-files/src/service.rs
@@ -174,8 +174,7 @@ impl Service {
             }
             packages.push(Resolvable::new("agama-scripts", ResolvableType::Package));
         }
-        self
-            .software
+        self.software
             .call(agama_software::message::SetResolvables::new(
                 "agama-scripts".to_string(),
                 packages,
--- a/rust/agama-lib/src/auth.rs
+++ b/rust/agama-lib/src/auth.rs
@@ -174,9 +174,7 @@ impl AuthToken {
 
     fn user_token_path() -> io::Result<PathBuf> {
         let Some(path) = home::home_dir() else {
-            return Err(io::Error::other(
-                "Cannot find the user's home directory",
-            ));
+            return Err(io::Error::other("Cannot find the user's home directory"));
         };
 
         Ok(path.join(USER_TOKEN_PATH))
--- a/rust/agama-lib/src/logs.rs
+++ b/rust/agama-lib/src/logs.rs
@@ -256,9 +256,7 @@ fn compress_logs(tmp_dir: &TempDir, result: &String) -> io::Result<()> {
     if res.success() {
         set_archive_permissions(PathBuf::from(result))
     } else {
-        Err(io::Error::other(
-            "Cannot create tar archive",
-        ))
+        Err(io::Error::other("Cannot create tar archive"))
     }
 }
 
--- a/rust/agama-security/src/service.rs
+++ b/rust/agama-security/src/service.rs
@@ -128,9 +128,7 @@ impl State {
     /// * `name`: certificate name (e.g., "registration_server")
     pub fn import(&mut self, certificate: &Certificate, name: &str) -> Result<(), Error> {
         let path = self.workdir.join(format!("{name}.pem"));
-        certificate
-            .import(&path)
-            .map_err(Error::CertificateIO)?;
+        certificate.import(&path).map_err(Error::CertificateIO)?;
         self.imported.push(name.to_string());
         Ok(())
     }
--- a/rust/agama-software/src/model/registration.rs
+++ b/rust/agama-software/src/model/registration.rs
@@ -124,9 +124,7 @@ impl Registration {
         let service = suseconnect_agama::activate_product(
             product,
             params,
-            self.connect_params
-                .email.as_deref()
-                .unwrap_or(""),
+            self.connect_params.email.as_deref().unwrap_or(""),
         )?;
 
         if let Some(file) = Self::credentials_from_url(&service.url) {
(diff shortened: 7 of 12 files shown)


rust/agama-lib/src/http/base_http_client.rs
```diff
@@ -154,7 +154,7 @@ impl BaseHTTPClient {
     pub async fn post<T>(
         &self,
         path: &str,
-        object: &impl Serialize,
+        object: &(impl Serialize + ?Sized),
```
jreidinger, line 157: I am not sure what is reason for this change as it compiles before.

rust/agama-manager/src/service.rs
```diff
@@ -591,7 +603,7 @@ impl Service {
         };
 
         let mut software_config = self.config.software.clone().unwrap_or_default();
-        let product_config = software_config.product.get_or_insert_default();
+        let product_config = software_config.product.get_or_insert_with(Default::default);
```
jreidinger, line 606: not sure why here it will change it. Previous one looks like specialized method.

rust/agama-network/src/nm/client.rs
```diff
@@ -101,14 +100,13 @@ impl<'a> NetworkManagerClient<'a> {
...
-                    .write(true)
-                    .open(not_copy_path)
-                {
-                    tracing::error!("Cannot write {} file {:?}", NOT_COPY_NETWORK_PATH, error);
-                }
+            } else if let Err(error) = OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                .open(not_copy_path)
```
jreidinger, line 107: here I would probably use `inspect_err` instead to make it more readable and easier.

rust/agama-network/src/nm/proxies.rs
```diff
@@ -852,6 +852,11 @@ pub trait IP4Config {
     fn wins_servers(&self) -> zbus::Result<Vec<u32>>;
 }
 
+#[allow(dead_code)]
```
jreidinger, line 855: why it is dead_code when it is used as result?

rust/agama-network/src/nm/proxies.rs
```diff
@@ -852,6 +852,11 @@ pub trait IP4Config {
...
+pub type IP6Address = (Vec<u8>, u32, Vec<u8>);
```
imobachgs, line 856: The proxies are auto-generated, but the change looks good.

rust/agama-server/src/server/web.rs
```diff
@@ -72,8 +78,8 @@ impl IntoResponse for Error {
         let mut status = StatusCode::BAD_REQUEST;
 
         if let Error::Manager(error) = &self {
-            if matches!(error, ManagerError::PendingIssues { issues: _ })
-                || matches!(error, ManagerError::Busy { scopes })
+            if matches!(**error, ManagerError::PendingIssues { issues: _ })
```
jreidinger, line 81: not sure what is reason for `**`. Also instead of `issues: _` I would use ` { .. }`

rust/agama-server/src/web/ws.rs
```diff
@@ -32,17 +32,6 @@ use axum::{
     Extension,
 };
 use std::sync::Arc;
-use tokio::sync::broadcast;
-
-#[derive(Debug, thiserror::Error)]
-pub enum Error {
```
jreidinger, line 38: that looks kind of strange that ws Error is not used.

rust/agama-software/src/model/conflict.rs
```diff
@@ -35,29 +35,3 @@ impl From<ConflictSolve> for (u32, u32) {
         (solve.conflict_id, solve.solution_id)
     }
 }
-
-/// Information about possible solution for conflict
```
jreidinger, line 39: well, this is not implemented feature for new software stack. So I would at least tag it in git, so we can resurrect when the need arise.

rust/agama-users/src/password.rs
```diff
@@ -104,7 +103,7 @@ mod test {
     #[test]
     #[cfg(not(ci))]
     fn test_passwords() {
-        let checker = PasswordChecker::default();
+        let checker = PasswordChecker;
```
jreidinger, line 106: this looks kind of strange.

rust/agama-utils/src/api/progress.rs
```diff
@@ -60,28 +60,28 @@ impl Progress {
...
             size: steps.len(),
             steps: steps.clone(),
-            step: steps.first().map_or(String::new(), |s| s.clone()),
+            step: steps.first().cloned().unwrap_or_default(),
             index: 1,
         }
     }
 
-    pub fn next(&mut self) -> Result<(), Error> {
+    pub fn advance(&mut self) -> Result<(), Error> {
```
jreidinger, line 68: this is API change. When I see it in my test run, I am not sure if is good change.

rust/agama-utils/src/test/dbus.rs
```diff
@@ -85,7 +86,14 @@ impl DBusServer<Stopped> {
             .spawn()
             .expect("to start the testing D-Bus daemon");
 
-        let connection = async_retry(|| connection_to(&self.address)).await?;
+        let connection = match async_retry(|| connection_to(&self.address)).await {
+            Ok(c) => c,
+            Err(e) => {
+                let _ = child.kill();
+                let _ = child.wait();
```
jreidinger, line 93: here it duplicate drop method. I am not sure if it is needed?

rust/agama-utils/src/products.rs
```diff
@@ -136,8 +136,7 @@ impl Registry {
     ///
     /// * `id`: product ID.
     /// * `mode`: product mode. Required only if the product has modes.
-    pub fn find(&self, id: &str, mode: Option<&str>) -> Result<ProductSpec, Error> {
-        let mut mode = mode.clone();
+    pub fn find<'a>(&'a self, id: &str, mut mode: Option<&'a str>) -> Result<ProductSpec, Error> {
```
jreidinger, line 139: not sure why it needs lifetimes here when it compile without them.

rust/suseconnect-agama/src/lib.rs
```diff
@@ -302,7 +302,7 @@ impl TryFrom<Value> for Credentials {
                 "Missing credentials key".to_string(),
             ));
         };
-        let Some(login) = credentials.get(0).and_then(Value::as_str) else {
+        let Some(login) = credentials.first().and_then(Value::as_str) else {
```
jreidinger, line 305: here I also see it in my test run and do not like much this change as later is `get(1)` and this way it looks inconsistent. Previously it is sequence of get by position.

rust/agama-lib/src/http/base_http_client.rs
imobachgs: Explained by Gemini:

```
   1. Clippy's Suggestion: Clippy correctly pointed out that passing &Box<RawValue> was inefficient and suggested using &RawValue instead (clippy::ptr_arg).
   2. The Problem: RawValue is an unsized type (it's essentially a wrapper around a str). While Box<RawValue> is Sized (because the pointer has a known size), the bare RawValue is not.
   3. The Error: When I changed the argument to &RawValue, the compiler complained that RawValue does not implement Sized, and therefore couldn't be passed to the HTTP client methods which were defined as &impl
      Serialize.
   4. The Fix: Adding + ?Sized (the "widening" or "relaxed" bound) tells the compiler: "This type must implement Serialize, but it's okay if its size isn't known at compile time."


  This allowed the BaseHTTPClient to be more flexible, accepting both standard sized types and reference-passed unsized types like RawValue or str.
```

rust/agama-manager/src/service.rs
imobachgs: Introduced in Rust 1.83 (and we are aiming at Rust 1.81). I guess we could change the minimal version according to SLES.

rust/agama-utils/src/api/progress.rs
imobachgs: Well, I guess it is proposing that name because `next` is part of the [iterator's API](https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next). So it looks like `Progress` implements that API but it does not.

PR #3208, opened 2026-02-23
Preload error

jreidinger:
## Problem

When libzypp skips some packages during preload due to various errors, installation still reports success.

## Solution

Fix include two parts:

1. report properly when commit method returns false as failure of packages installation
2. change method on CommitResult from checking for errors to the one that checks also if there are skipped packages.

## Testing

- *Tested manually*

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Feb 23 15:10:58 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
+
+- Do not use verbose mode in the "image-janitor" calls, logging
+  so many details significantly increases the ISO build time
+  (roughly two times) (gh#agama-project/agama#3206)
+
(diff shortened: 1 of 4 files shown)


rust/agama-software/src/service.rs
```diff
@@ -409,8 +411,11 @@ impl MessageHandler<message::Refresh> for Service {
 
 #[async_trait]
 impl MessageHandler<message::Install> for Service {
-    async fn handle(&mut self, _message: message::Install) -> Result<bool, Error> {
-        self.model.lock().await.install().await
+    async fn handle(&mut self, _message: message::Install) -> Result<(), Error> {
+        if !self.model.lock().await.install().await? {
```
imobachgs, line 415: np: too much for a single line :) (getting the lock and calling a method).

jreidinger: yeap, it is growing...I promise that next increase will split it :)

PR #3209, opened 2026-02-24
Address clippy issues (take 2)

imobachgs:
## Problem

There are many complains from Clippy.

## Solution

- Apply automatic suggestions.
- Update minimal Rust version to 1.88 (version of the `rust` package in SLES).
- Ask Gemini to fix the rest:
   * Boxed large enum variants in agama-software and agama-utils to reduce memory footprint and satisfy large-enum-variant and result-large-err lints.
   * Updated function signatures to use &Path instead of &PathBuf and &[T] instead of &Vec<T> to avoid unnecessary allocations (ptr-arg lint).
   * Replaced manual boolean comparisons with idiomatic assert! and matches! macros.
   * Removed redundant imports, unused variables, and dead code across multiple crates.
   * Fixed documentation formatting and improved code readability by collapsing nested if let statements.
   * Simplified pointer casts and updated struct initializations to be more concise.

### DIFF
--- a/rust/agama-server/src/server/web.rs
+++ b/rust/agama-server/src/server/web.rs
@@ -72,8 +72,8 @@ impl IntoResponse for Error {
         let mut status = StatusCode::BAD_REQUEST;
 
         if let Error::Manager(error) = &self {
-            if matches!(error, ManagerError::PendingIssues { issues: _ })
-                || matches!(error, ManagerError::Busy { scopes: _ })
+            if matches!(error, ManagerError::PendingIssues { .. })
+                || matches!(error, ManagerError::Busy { .. })
             {
                 status = StatusCode::UNPROCESSABLE_ENTITY;
             }


rust/agama-lib/src/storage/http_client/iscsi.rs
```diff
@@ -48,6 +48,7 @@ impl ISCSIHTTPClient {
         Ok(None)
     }
 
+    #[allow(clippy::borrowed_box)]
```
jreidinger, line 51: I would probably fix that differently then silencing warning

rust/agama-server/src/server/web.rs
```diff
@@ -73,7 +73,7 @@ impl IntoResponse for Error {
 
         if let Error::Manager(error) = &self {
             if matches!(error, ManagerError::PendingIssues { issues: _ })
-                || matches!(error, ManagerError::Busy { scopes })
+                || matches!(error, ManagerError::Busy { scopes: _ })
```
jreidinger: as said, I would probalbly use simple `                || matches!(error, ManagerError::Busy { .. })`

rust/agama-utils/src/test/dbus.rs
```diff
@@ -85,12 +86,17 @@ impl DBusServer<Stopped> {
...
-            extra: Started { child, connection },
-        })
+        match async_retry(|| connection_to(&self.address)).await {
+            Ok(connection) => Ok(DBusServer {
+                address: self.address,
+                extra: Started { child, connection },
+            }),
+            Err(e) => {
+                let _ = child.kill();
+                let _ = child.wait();
```
jreidinger, line 96: I do not know what is reason here for manual cleaning?

rust/agama-lib/src/storage/http_client/iscsi.rs
imobachgs: I agree but I think it does not matter because this code should be removed (is the old iSCSI HTTP client).

rust/agama-utils/src/test/dbus.rs
imobachgs: From Gemini: *By calling wait(), the parent process acknowledges the child's termination, allowing the operating system to fully remove the process from the table and release its PID. In this specific Drop implementation, it ensures that every D-Bus daemon started for a test is cleanly and completely removed when the test finishes or fails.*

PR #3210, opened 2026-02-24
Allow to reuse LVM volume groups, volumes and thin pools

ancorgs:
## Problem

Agama does not allow to reuse an existing LVM volume group and its logical volumes.

## Solution

This adds the possibility to search and, thus, reuse existing volume groups and its logical volumes.

Includes the possibility to create normal volumes, thin pools and thin volumes.

It also includes the possibility to create new thin volumes in a reused thin pool.

It allows to specify which logical volumes should be deleted and resized in order to make space for the new logical volume, in a way totally analogous to what can already be done with partitions when using a disk.

It also allows to add more physical volumes to an existing volume group. But only using explicit volumes (ie. physical volumes cannot be created on demand using `generate`).

Apart from shrinking pre-existing logical volumes, this also allows to grow them in a similar way that what can be done for partitions.

## See also

This depends on https://github.com/yast/yast-storage-ng/pull/1421

## Testing

- Added unit tests, but they cannot pass until https://github.com/yast/yast-storage-ng/pull/1421 is merged
- Tested manually

## To do (follow-ups)

- Detect possible incorrect settings or setups and detect them in the `ConfigChecker`.
- More criteria for searching logical volumes, like searching by volume name or by volume type (normal, pool, thin, etc.).
- Decide whether the proposal should refuse to resize some logical volumes and partitions to prevent unexpected data loss (that is a general topic not really related to LVM).

### DIFF
--- a/service/package/rubygem-agama-yast.changes
+++ b/service/package/rubygem-agama-yast.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Mar  6 13:03:21 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Allow to reuse LVM volume groups, volumes and thin pools
+  (jsc#PED-15104, bsc#1254718 and gh#agama-project/agama#3171).
+
+-------------------------------------------------------------------
 Tue Mar  3 12:35:55 UTC 2026 - José Iván López González <jlopez@suse.com>
 
 - Provide new D-Bus API for zFCP (gh#agama-project/agama#3184).


service/lib/agama/storage/config_conversions/to_json_conversions/with_delete.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2024] SUSE LLC
```
joseivanlopez:
```suggestion
# Copyright (c) [2026] SUSE LLC
```

service/lib/agama/storage/config_conversions/to_json_conversions/with_delete.rb
```diff
@@ -0,0 +1,63 @@
...
+
+require "agama/storage/config_conversions/to_json_conversions/search"
+
+module Agama
+  module Storage
+    module ConfigConversions
+      module ToJSONConversions
+        # Mixin for delete conversion to JSON.
+        #
+        # The class is also expected to include WithSearch and WithSize.
```
joseivanlopez: NP: maybe we can include these mixins here.

service/lib/agama/storage/configs/logical_volume.rb
```diff
@@ -53,9 +55,19 @@ class LogicalVolume
         # @return [Encryption, nil]
         attr_accessor :encryption
 
+        # @return [Boolean]
+        attr_accessor :delete
```
joseivanlopez: Maybe we can create a `WithDelete` mixin.

PR #3328, opened 2026-03-24
Make our GH Actions source YML files more readable by factoring out strings

mvidner:
## Problem

I read our GitHub Actions that autosubmit Agama to OBS and find them hard to read:

- The expression for the target OBS project is repeated
- Shell commands are squashed to a single line

## Solution

- Declare `env.obs_project`
- Use shell variables to break up the monster lines to tame ones

## Testing

- [x] Tested manually: ran `devel/branch2obs.sh` so that the submission actions run
    - [x] Fixed a quoting error in this PR
    - [x] Fixed a bug in branch2obs when you have multiple remotes

## Screenshots

No, but here is a [Workflow run](https://github.com/agama-project/agama/actions/runs/23494790385/job/68372677885) to demonstrate that the `env` expressions indeed work, even in boolean contexts.

## Documentation

- No new functionality
- Mentioned reference docs links in the commits

### DIFF
--- a/.github/workflows/obs-staging-live.yml
+++ b/.github/workflows/obs-staging-live.yml
@@ -12,10 +12,14 @@ on:
   # allow running manually
   workflow_dispatch:
 
+env:
+  obs_project: ${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }}
+
 jobs:
   update_obs_package:
     # do not run in forks which do not set the OBS_PROJECTS variable,
     # or the mapping for the current branch is missing
+    # (BTW, jobs.FOO.if cannot see env.*)
     if: vars.OBS_PROJECTS && fromJson(vars.OBS_PROJECTS)[github.ref_name] && vars.OBS_USER
 
     runs-on: ubuntu-latest
@@ -44,8 +48,8 @@ jobs:
           OBS_USER:     ${{ vars.OBS_USER }}
           OBS_PASSWORD: ${{ secrets.OBS_PASSWORD }}
 
-      - name: Checkout ${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }} agama-installer
-        run: osc co -o dist ${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }} agama-installer
+      - name: Checkout ${{ env.obs_project }} agama-installer
+        run: osc co -o dist ${{ env.obs_project }} agama-installer
         working-directory: ./live
 
       - name: Build sources
@@ -60,6 +64,6 @@ jobs:
         run: osc diff && osc status
         working-directory: ./live/dist
 
-      - name: Commit agama-installer to ${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }}
+      - name: Commit agama-installer to ${{ env.obs_project }}
         run: osc commit -m "Updated to Agama $GITHUB_SHA"
         working-directory: ./live/dist
--- a/.github/workflows/obs-staging-shared.yml
+++ b/.github/workflows/obs-staging-shared.yml
@@ -79,9 +79,11 @@ jobs:
       - name: Update service revision
         # only when a tag has been pushed, or "release" branch updated
         if: inputs.service_file != ''
-        run: |-
+        run: |
           echo "Updating revision to \"${{ github.ref_name }}\""
-          sed -i -e 's#<param name="revision">.*</param>#<param name="revision">${{ github.ref_name }}</param>#' ${{ inputs.service_file }}
+          BEG='<param name="revision">'
+          END='</param>'
+          sed -i -e "s#${BEG}.*${END}#${BEG}${{ github.ref_name }}${END}#" ${{ inputs.service_file }}
 
       - name: Copy optional service file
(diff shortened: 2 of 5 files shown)


.github/workflows/obs-staging-live.yml
```diff
@@ -12,10 +12,14 @@ on:
...
   workflow_dispatch:
 
+env:
+  obs_project: ${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }}
+
 jobs:
   update_obs_package:
     # do not run in forks which do not set the OBS_PROJECTS variable,
     # or the mapping for the current branch is missing
+    # (BTW, jobs.FOO.if cannot see env.*)
```
jreidinger, line 22: it should be able using context. At least it is what doc says https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#using-the-env-context-to-access-environment-variable-values

.github/workflows/obs-staging-shared.yml
```diff
@@ -66,27 +70,34 @@ jobs:
...
         run: |
-          sed -e 's#<param name="url">.*</param>#<param name="url">https://github.com/${{ github.repository }}.git</param>#' ${{ inputs.service_file }} > ./${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }}/${{ inputs.package_name }}/_service
-          if [ -z "$(git tag -l)" ]; then sed -i -e 's#<param name="versionformat">.*</param>##' ./${{ fromJson(vars.OBS_PROJECTS)[github.ref_name] }}/${{ inputs.package_name }}/_service; fi
+          OUT_SERVICE=./${{ env.obs_project }}/${{ inputs.package_name }}/_service
+          BEG='<param name="url">'
+          END='</param>'
+          SED_EXPR="s#${BEG}.*${END}#${BEG}https://github.com/${{ github.repository }}.git${END}#"
+          sed -e "${SED_EXPR}" ${{ inputs.service_file }} > $OUT_SERVICE
+          BEG='<param name="versionformat">'
+          if [ -z "$(git tag -l)" ]; then sed -i -e "s#${BEG}.*${END}##" $OUT_SERVICE; fi
```
jreidinger, line 100: I am not sure if it is more readable now.

devel/branch2obs.sh
```diff
@@ -236,14 +236,14 @@ workflows=(obs-staging-live.yml obs-staging-products.yml obs-staging-rust.yml ob
 if git ls-remote --exit-code --heads origin "$BRANCH" > /dev/null; then
   for workflow in "${workflows[@]}"; do
     echo "Starting GitHub Action $workflow..."
-    gh workflow run "$workflow" --ref "$BRANCH"
+    gh -R "$repo_slug" workflow run "$workflow" --ref "$BRANCH"
```
jreidinger, line 239: nice..I fix that by using gh repo set-default, but your solution looks better.

.github/workflows/obs-staging-live.yml
mvidner: No. That doc uses `env` for `jobs.<job_id>.steps.if`, but `env` is not available for `jobs.<job_id>.if`. Beautiful, right?

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#context-availability

.github/workflows/obs-staging-shared.yml
mvidner: Yes it is complex and dense, but IMHO better than the previous version

jreidinger: well, it makes reading a bit harder as you do not see what sed is in fact doing. I ask AI how to make it more readable and here is AI proposal which is for me more readable ( but still it includes sed expressions :)

```
OUT_SERVICE="./${{ env.obs_project }}/${{ inputs.package_name }}/_service"
REPO_URL="https://github.com/${{ github.repository }}.git"

# Replace the URL parameter and output to the new file
sed "s#<param name=\"url\">.*</param>#<param name=\"url\">${REPO_URL}</param>#" \
  "${{ inputs.service_file }}" > "$OUT_SERVICE"

# Strip the versionformat parameter if there are no git tags
if [ -z "$(git tag -l)" ]; then 
  sed -i "s#<param name=\"versionformat\">.*</param>##" "$OUT_SERVICE"
fi
```

mvidner: IMHO this is a case where your using AI makes things (slightly) harder for your human counterparty, myself. Now I would have to transplant the result to the YAML file and retest it. If you trust the AI output, please make a patch or a commit for me :)

jreidinger: well, I do not trust AI, but paste it here to demonstrate what I do not like at your patch...that sed regexp like in some variable elsewhere, so you need more processing of it. On other hand AI solution ( even if it can have bugs ) is more easier to read for me as there is sed regexp with some long tricky replacements in own variable which is easier to read ( at least for me ).

lslezak: I find it a bit more readable than the original. 😃

lslezak (approved):

PR #3332, opened 2026-03-25
Make curl dependency for agama-network optional via feature

jcronenberg:
## Problem

Due to a dependency chain agama-network pulls in curl via agama-network -> agama-utils -> agama-transfer -> curl. With this comes a dependency on openssl which unfortunately for wicked2nm is a problem, because it fails to build on older SLE 15 SPs.

## Solution

To circumvent this I added a default feature that excludes the affected parts from agama-network so it doesn't depend on agama-transfer.
This should have no actual impact on agama itself.


teclator:
It looks good to me. @jreidinger what do you think?

jreidinger:
@teclator lets write it also here beside private slack conversation. We do not need that feature flag..so I do not mind to have it when it helps, but also there is high chance we accidentally break it in future unless we have some automatic check for build with curl flag disabled. So LGTM just with risk of accidental future break.

teclator:
Merging it as it is by now

teclator (approved):
LGTM

jcronenberg:
> there is high chance we accidentally break it in future unless we have some automatic check for build with curl flag disabled

Yes I am aware, however I think a check like this is not easily possible. I guess it'll be our responsibility if there are regressions in this regard, since we will notice them. That is fine from my point of view, so no need to really take this into consideration from your side.

PR #3350, opened 2026-03-31
Support to configure the bootloader type to use

ancorgs:
## Problem

Quite some logic has been introduced lately in yast2-bootloader and in yast2-storage-ng to decide which bootloader to install in every case (Grub2, systemd-boot, grub2-bls, etc.).

That logic is duplicated in both YaST components but, unfortunately, it is not identical.

Moreover, that logic is based on the system architecture and in some YaST-specific configurations like `Yast::ProductFeatures.preferred_bootloader`. Although initially a `YAST_NO_BLS_BOOT` environment variable was introduced as a mechanism to avoid the YaST changes to affect Agama, that is not longer working after several changes in that area.

## Solution

This introduces a reasonable mechanism to select the bootloader type to use in Agama.

The setting is handled internally in the Agama configuration but not exposed in JSON yet. For now, products can define the default bootloader for EFI environments.

Both yast2-bootloader and yast2-storage-ng will now honor the setting, which value is now handled with a single centralized mechanism.

## Notes

This is a pull request against a feature branch. It introduces new classes like `Y2Storage::Proposal::BootPlanner` or `Agama::Storage::BootloaderType` that would need to be re-evaluated. Before merging the feature branch into master, it will maybe make sense to consolidate some of those with yast2-storage-ng.

### DIFF
--- a/service/lib/agama/storage/bootloader_type.rb
+++ b/service/lib/agama/storage/bootloader_type.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+# Copyright (c) [2026] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Agama
+  module Storage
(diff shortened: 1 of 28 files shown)


ancorgs:
I just realized this could make things worse when the AutoYaST legacy mode (`legacy_autoyast_storage`) is used. I'm considering how to make things better in that regard (it could be part of a separate pull request).

ancorgs:
> I'm considering how to make things better in that regard (it could be part of a separate pull request).

I added a new commit for that. Still not a final solution since that would depend on what we finally do with `Yast::BootloaderChecker` in the end.

service/lib/agama/storage/bootloader_type.rb
```diff
@@ -0,0 +1,92 @@
...
+# find current contact information at www.suse.com.
+
+module Agama
+  module Storage
+    # Enum-like class to represent the possible bootloader types at BootloaderConfig
+    class BootloaderType
+      # Constructor, to be used internally by the class
+      #
+      # @param value [String] see {#value}
+      def initialize(value)
```
joseivanlopez, line 29: Does it make sense to declare the constructor as private?

joseivanlopez (approved):
LGTM

PR #3371, opened 2026-04-10
Use /var/lib/agama/scripts to save scripts

imobachgs:
## Problem

* The scripts cannot be executed from /var/log/agama-installation on a hardened system.

- [bsc#1261787](https://bugzilla.suse.com/show_bug.cgi?id=1261787).

## Solution

Use `/var/lib/agama/scripts`.

## Testing

- Adapted the unit test.
- Tested manually.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Fri Apr 10 06:21:35 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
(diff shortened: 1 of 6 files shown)


joseivanlopez (approved):
LGTM

mvidner (commented):
@imobachgs Do we now more about what kind of "a hardened system" that is? If it is a common and reasonable method, we could add it to our tests

imobachgs:
No, to be honest I do not have more details (just that the `/var/log` mount point does not have exec permissions, which makes sense).

PR #3373, opened 2026-04-10
Return of ipmi

mchf:
## Problem

There used to be support for Ipmi reporting (https://github.com/agama-project/agama/pull/2490) which was lost during transformation of manager service from ruby to rust.

## Solution

Ipmi code rewritten from ruby to rust and hooked into the manager service

## TODO
- changelog (will be added once we get close to end of the review ;-))

## Testing

- *Tested manually*


rust/agama-manager/src/ipmi.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) [2025] SUSE LLC
```
imobachgs:
```suggestion
// Copyright (c) [2026] SUSE LLC
```

rust/agama-manager/src/ipmi.rs
```diff
@@ -0,0 +1,122 @@
...
+        if !self.is_available() {
+            return Ok(());
+        }
+
+        // Create a temporary file that is automatically deleted
+        let mut file = match NamedTempFile::new() {
+            Ok(f) => f,
+            Err(_) => {
+                return Err(Error::Tool {
+                    desc: "ipmitool failed, cannot create temporary event file".to_string(),
```
imobachgs: It would be nice to log the underlying error.

rust/agama-manager/src/ipmi.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub struct Ipmi {}
+
+impl Ipmi {
+    const IPMI_STARTED: u8 = 0x07;
+    const IPMI_FINISHED: u8 = 0x08;
+    const IPMI_ABORTED: u8 = 0x09;
+    const IPMI_FAILED: u8 = 0x0A;
+
+    pub fn new() -> Self {
```
imobachgs: What about returning an error if the ipmi is not available? In that case, this struct does not have any usage. I do not have a strong opinion about this, it is just a suggestion.

imobachgs (commented):
I have a few comments about the overall approach.

You are using an empty struct which does not hold any kind of state. Actually, the `&self` arguments are not needed (e.g., check `is_available`).

I guess you want to somehow group all those functions together and in the future you might want to keep some state (e.g., caching `is_available` or something like that).

Additionally, you are using some hard-coded values (two paths). What about moving those paths to the struct itself?

```rust
struct Ipmi {
  tool: PathBuf, // feel free to use better names
  dev: PathBuf
}

impl Ipmi {
  fn new(tool: &str, dev: &str) -> Self { // you can use a String if you prefer
    Self {
      tool: tool.to_string(),
      dev: dev.to_string()
    }
  }
}

impl Default for Impi {
  fn default() -> Self {
    Self::new("/usr/bin/ipmitool", "/dev/ipmi0")
  }
}
```

This approach would enable you to write unit tests (that are actually missing).

rust/agama-manager/src/ipmi.rs
mchf: Well ipmi is kind of optional from our POV and in fact we do not care whether it is available or not ... And wrapping return value into a Result would make coding little bit more noisy ;-)

imobachgs: Yes, I understand, that's why I say it is just a suggestion. All those functions are now a no-op. At least I would log in the `send_command` function that `IPMI` is not available. Or if you prefer, you could call `is_available` when the manager starts printing whether it is supported or not.

mchf:
> I have a few comments about the overall approach.
> 
> You are using an empty struct which does not hold any kind of state. Actually, the `&self` arguments are not needed (e.g., check `is_available`).
> 
> I guess you want to somehow group all those functions together and in the future you might want to keep some state (e.g., caching `is_available` or something like that).

yes, that was the idea. However originally I wanted to do a (kind of) 1:1 version of original ruby code.  As it wasn't according to your taste I've added basic update according to your comments

rust/agama-manager/src/service.rs
```diff
@@ -814,7 +816,19 @@ impl MessageHandler<message::RunAction> for Service {
                 self.probe_dasd().await?;
             }
             Action::Install => {
-                self.tasks.cast(tasks::message::Install)?;
+                let ipmi = ipmi::Ipmi::default();
+
+                if let Err(e) = ipmi.started() {
+                    tracing::error!("Ipmi coommand failed: {}", e);
```
imobachgs:
```suggestion
                    tracing::error!("IPMI failed: {}", e);
```

rust/agama-manager/src/service.rs
```diff
@@ -814,7 +816,19 @@ impl MessageHandler<message::RunAction> for Service {
...
+                }
+
+                if let Err(error) = self.tasks.cast(tasks::message::Install) {
+                    if let Err(e) = ipmi.failed() {
+                        tracing::error!("Ipmi command failed: {}", e);
```
imobachgs:
```suggestion
                        tracing::error!("IPMI failed: {}", e);
```

rust/agama-manager/src/actions.rs
```diff
@@ -422,6 +422,11 @@ impl FinishAction {
         }
 
         command.arg("now");
+
+        if let Err(e) = Ipmi::default().finished() {
+            tracing::error!("Ipmi command failed: {}", e);
```
imobachgs:
```suggestion
            tracing::error!("IPMI failed: {}", e);
```

imobachgs (commented):
I am still not convinced by the no-op and, more important, the lack of tests.

But please, do not merge until you can try a full installation.

imobachgs (dismissed):

rust/package/agama.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue Apr 14 10:20:25 UTC 2026 - Michal Filka <mfilka@suse.com>
+
+- jsc#PED-12285
+  - re-introduced initial implementation for installer state status
+    report via IPMI. Original implementation lost due to transition
+    from ruby to rust.
```
```suggestion
    from Ruby to Rust.
```

imobachgs (approved):

PR #2399, opened 2025-05-26
feat(rust): add more info to the monitor

imobachgs:
## Problem

The "agama monitor" is good at displaying the progress line by line, but not that good to display the current status.

## Solution

This PR is an experiment to build a new monitor which offers more (and better) information.

Displaying the issues

Displaying the progress

## Testing

- *Tested manually*

PR #3377, opened 2026-04-13
Monitor improvements

jreidinger:
## Problem

Monitor do not print if it wait for questions or if there are issues. Also there is issue with following all states. Monitor also do not end at the end of installation.

PBI: https://trello.com/c/wGvQcQk8/5655-8-cli-agama-monitor-improvements

## Solution

Implement agreed improvements.

## Testing

- *Tested manually*

## Screenshots

Live demo will be shown on review meeting.


mchf:
1. ```MonitorClient``` could be removed from agama-lib, couldn't it?
2. Missing changelog

jreidinger:
@mchf for me question is whole future of agama-lib...for now it looks like lib for agama-cli...so it makes sense to kind of let it there.
Regarding changelog as said I plan to add it when I get some preapproval to avoid too much conflicts.

rust/agama-cli/src/lib.rs
```diff
@@ -110,26 +110,26 @@ async fn probe(manager: ManagerHTTPClient, monitor: MonitorClient) -> anyhow::Re
...
-    let manager = ManagerHTTPClient::new(http_client);
-    manager.install().await?;
+    manager_client.install().await?;
 
     // wait a bit before start monitoring
     sleep(Duration::from_secs(1)).await;
     let progress = tokio::spawn(async {
-        show_progress(monitor, true).await;
+        // just prematurely finish install command if progress monitor failed
+        let _ = show_progress(http_client, ws, true).await;
```
imobachgs: What about printing/logging the error?

rust/agama-cli/src/lib.rs
```diff
@@ -110,26 +110,26 @@ async fn probe(manager: ManagerHTTPClient, monitor: MonitorClient) -> anyhow::Re
...
         return Err(CliError::Validation)?;
     }
 
-    let manager = ManagerHTTPClient::new(http_client);
-    manager.install().await?;
+    manager_client.install().await?;
 
     // wait a bit before start monitoring
     sleep(Duration::from_secs(1)).await;
     let progress = tokio::spawn(async {
```
imobachgs: do we still to run this with a spawn?

rust/agama-cli/src/lib.rs
```diff
@@ -141,24 +141,32 @@ async fn install(http_client: BaseHTTPClient, monitor: MonitorClient) -> anyhow:
...
     Ok(())
 }
 
-async fn wait_until_idle(monitor: MonitorClient) -> anyhow::Result<()> {
-    // FIXME: implement something like "wait_until_idle" in the monitor?
-    let status = monitor.get_status().await?;
-    if !status.progresses.is_empty() {
-        eprintln!("The Agama service is busy. Waiting for it to be available...");
-        show_progress(monitor.clone(), true).await;
+async fn wait_until_idle(http: &BaseHTTPClient, ws: &mut WebSocketClient) -> anyhow::Result<()> {
```
imobachgs, line 161: This function looks much better now.

rust/agama-cli/src/lib.rs
```diff
@@ -263,22 +271,25 @@ fn find_client_token(api_url: &Url) -> Option<AuthToken> {
...
-    let mut progress = ProgressMonitor::new(monitor).stop_on_idle(stop_on_idle);
-    if let Err(e) = progress.run().await {
-        eprintln!("Could not display the progress: {e:?}");
-    }
+pub async fn show_progress(
+    http: BaseHTTPClient,
+    ws: WebSocketClient,
+    stop_on_idle: bool,
+) -> anyhow::Result<()> {
+    // TODO: should monitor has some non interactive version?
```
imobachgs: What do you mean?

rust/agama-cli/src/lib.rs
```diff
@@ -287,21 +298,19 @@ pub async fn run_command(cli: Cli) -> anyhow::Result<()> {
     match cli.command {
         Commands::Config(subcommand) => run_config_cmd(subcommand, cli.opts).await?,
         Commands::Probe => {
-            let (client, monitor) = build_clients(api_url, cli.opts.insecure).await?;
-            let manager = ManagerHTTPClient::new(client.clone());
-            let _ = wait_until_idle(monitor.clone()).await;
-            probe(manager, monitor).await?
+            let (http, mut ws) = build_clients(api_url, cli.opts.insecure).await?;
+            let _ = wait_until_idle(&http, &mut ws).await;
```
imobachgs: I guess we shoudl not ignore an error when calling `wait_until_idle` because it may reveal a connection problem.

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,246 @@
...
+        let questions = questions
+            .into_iter()
+            .filter(|q| q.answer.is_none())
+            .collect();
+        Ok(questions)
+    }
+
+    fn print_stage(stage: &Stage) {
+        match stage {
+            Stage::Configuring => println!("{}", gettext("Installation is ready for start.")),
```
mchf: I guess it should be: 
"... ready to start ..."

imobachgs (changes requested):
Let's summarize what we discussed through other channels because because I think there is some room for improvement.

My main concern is that the UI is coupled with the logic to retrieve the current status of the server. I would separate them in two parts:

* A mechanism that retrieves the current status. It might contain the current installation phase, the progresses, etc. and listen to events to keep that information up-to-date.
* A separate struct that draws the current status. It might access to that information using a some shared memory (like a Mutex) or event a message (or combining both). Using a message would save you from having to set a frequency to update the UI (it would be updated when the message if received).

I tried something similar in [an experiment I did some time ago](https://github.com/agama-project/agama/pull/2399). I am not saying we should use the very same approach, of course.

rust/agama-cli/src/lib.rs
jreidinger: print to stderr probably makes sense,

rust/agama-cli/src/lib.rs
jreidinger: I think it is no longer needed.

rust/agama-cli/src/lib.rs
jreidinger: I mean that we run it also e.g. during agama install or agama config load...and it include those terminal control chars which is probably not friendly to script usage...so maybe we should detect such usage and show something else.

rust/agama-cli/src/lib.rs
```diff
@@ -110,28 +110,30 @@ async fn probe(manager: ManagerHTTPClient, monitor: MonitorClient) -> anyhow::Re
...
     // wait a bit before start monitoring
     sleep(Duration::from_secs(1)).await;
-    let progress = tokio::spawn(async {
-        show_progress(monitor, true).await;
-    });
-    let _ = progress.await;
+
+    let res = show_progress(http_client, ws, true).await;
+    if let Err(e) = res {
+        eprintln!("failed to show progress: {:?}", e);
```
imobachgs:
```suggestion
        eprintln!("Failed to show progress: {:?}", e);
```

rust/agama-cli/src/lib.rs
```diff
@@ -287,21 +302,25 @@ pub async fn run_command(cli: Cli) -> anyhow::Result<()> {
...
         Commands::Config(subcommand) => run_config_cmd(subcommand, cli.opts).await?,
         Commands::Probe => {
-            let (client, monitor) = build_clients(api_url, cli.opts.insecure).await?;
-            let manager = ManagerHTTPClient::new(client.clone());
-            let _ = wait_until_idle(monitor.clone()).await;
-            probe(manager, monitor).await?
+            let (http, mut ws) = build_clients(api_url, cli.opts.insecure).await?;
+            wait_until_idle(&http, &mut ws)
+                .await
+                .context("failed to check if service is busy")?;
```
imobachgs:
```suggestion
                .context("Failed to check if service is busy")?;
```

From user's perspective: "which service?". What about slightly different wording? "Failed to check if the Agama server is busy." or "is available.".

rust/agama-cli/src/lib.rs
```diff
@@ -287,21 +302,25 @@ pub async fn run_command(cli: Cli) -> anyhow::Result<()> {
...
+            probe(http, ws).await?
         }
         Commands::Install => {
-            let (client, monitor) = build_clients(api_url, cli.opts.insecure).await?;
-            let _ = wait_until_idle(monitor.clone()).await;
-            install(client, monitor).await?
+            let (http, mut ws) = build_clients(api_url, cli.opts.insecure).await?;
+            wait_until_idle(&http, &mut ws)
+                .await
+                .context("failed to check if service is busy")?;
```
imobachgs:
```suggestion
                .context("Failed to check if service is busy")?;
```

rust/agama-cli/src/lib.rs
```diff
@@ -287,21 +302,25 @@ pub async fn run_command(cli: Cli) -> anyhow::Result<()> {
...
         }
         Commands::Finish { method } => {
-            let (client, monitor) = build_clients(api_url, cli.opts.insecure).await?;
-            let manager = ManagerHTTPClient::new(client.clone());
-            let _ = wait_until_idle(monitor.clone()).await;
-            finish(manager, monitor, method).await?;
+            let (http, mut ws) = build_clients(api_url, cli.opts.insecure).await?;
+            wait_until_idle(&http, &mut ws)
+                .await
+                .context("failed to check if service is busy")?;
```
imobachgs:
```suggestion
                .context("Failed to check if service is busy")?;
```

rust/agama-cli/src/lib.rs
imobachgs: I am just mentioning because we have *a lot* of TODO, FIXME, etc. And it would be good create a PBI instead.

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+use gettextrs::gettext;
+use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 
 /// Displays the progress on the terminal.
 #[derive(Debug)]
 pub struct ProgressMonitor {
-    monitor: MonitorClient,
-    running: bool,
+    /// The current installation status.
+    state: InstallationStatus,
```
imobachgs: state vs status... I would try to be consistent here (and use one of them).

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+        if status.stage.is_end() {
+            Self::print_final_status(status);
+            return Ok(false);
+        }
+        let progresses = &status.progresses;
+        // if there are some progress, then it has precedence over issues as it can solve them
+        if !progresses.is_empty() {
+            let multibar = MultiProgress::new();
+            let message = if status.stage == Stage::Configuring {
+                gettext("Calculating proposal:")
```
imobachgs:
```suggestion
                gettext("Calculating the installation proposal:")
```

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+            } else {
+                gettext("Installing target system:")
```
imobachgs:
```suggestion
                gettext("Installing the system:")
```

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+    /// Listens for updates from the monitor and updates the progress display.
+    async fn loop_monitor(&mut self) -> anyhow::Result<()> {
+        let mut receiver = self.monitor.subscribe();
         loop {
-            if let Ok(status) = updates.recv().await {
-                if !self.update(&status).await {
-                    return Ok(());
+            let new_state = receiver.recv().await;
+            let Ok(Ok(new_state)) = new_state else {
+                return Err(anyhow::Error::msg("Communication with backend failed"));
```
imobachgs: Again from user's perspective, what is the "backend"? We should use something like "Agama's server" (or something else).

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+
+    /// Listens for updates from the monitor and updates the progress display.
+    async fn loop_monitor(&mut self) -> anyhow::Result<()> {
+        let mut receiver = self.monitor.subscribe();
         loop {
-            if let Ok(status) = updates.recv().await {
-                if !self.update(&status).await {
-                    return Ok(());
+            let new_state = receiver.recv().await;
+            let Ok(Ok(new_state)) = new_state else {
```
imobachgs: I do not find the point of nesting both results if you are going to handle them like this.

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
-                    let bar = bars.entry(progress.scope).or_insert_with(|| {
-                        let style = ProgressStyle::with_template("{spinner:.green} {msg}").unwrap();
-                        let new_bar = ProgressBar::new(progress.size as u64).with_style(style);
-                        new_bar.enable_steady_tick(Duration::from_millis(120));
-                        multibar.add(new_bar)
-                    });
-                    bar.set_message(progress.step);
-                    bar.set_position(progress.index as u64);
+            // then check update of progresses
+            if self.state.status.progresses != new_state.status.progresses {
```
imobachgs: This piece of code looks a little bit complex. I would expect something simple like:

```
if self.state != new_state {
  self.state = new_state;
  self.render(); // or perhaps self.render(new_state);
}
```

And then, the `render` function would contain the logic to display the current status, rendering the state (`render_state`) and updating the progress bars (if any). Or am I missing something?

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+        self.progresses.clear();
+        self.progress_bar = None;
+
+        // and whole terminal
+        Self::clear_terminal();
+
+        // if there is any unaswered question, it has precedence as it affects everything else
+        let questions = &self.state.questions;
+        if !questions.is_empty() {
+            self.print_questions(questions).await?;
```
imobachgs: In some cases you are using associated methods (with no `self`) and in cases like this you are passing `self`. I do not have a strong opinion about which one to use, but I would try to be consistent.

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+        for q in questions {
+            // Should we also print question class?
+            println!("  - {}", q.spec.text);
+        }
+        Ok(())
+    }
+
+    /// Prints any issues blocking the installation.
+    fn print_issues(issues: &Vec<IssueWithScope>) -> anyhow::Result<()> {
+        println!("{}", gettext("Installer is preparing target configuration. But there are issues blocking installation:"));
```
imobachgs: The message looks confusing because it gives the impression that the installer is doing something (and it might be idle).

rust/agama-cli/src/progress.rs
```diff
@@ -18,98 +18,294 @@
...
+
+        let mut grouped: HashMap<&Scope, Vec<&api::Issue>> = HashMap::new();
```
imobachgs, line 261: I think we should have a mechanism to order the scopes so they are always printed in the same order (e.g., manager first). In my tests, I got them in different orders.

rust/agama-lib/src/monitor.rs
```diff
@@ -18,93 +18,153 @@
...
-    #[error("WebSocket error: {0}")]
-    WebSocket(#[from] WebSocketError),
-    #[error(transparent)]
-    Url(#[from] url::ParseError),
+    /// Error connecting to the Manager HTTP API.
+    #[error("Error connecting to the Manager HTTP API: {0}")]
+    Manager(#[from] ManagerHTTPClientError),
+    /// Error connecting to the Questions HTTP API.
+    #[error("Error connecting to the Questions HTTP API: {0}")]
+    Questions(#[from] QuestionsHTTPClientError),
```
imobachgs, line 73: Out of scope, but I think we should tend to have a single HTTP client now that the API surface is significantly smaller.

rust/agama-lib/src/monitor.rs
```diff
@@ -18,93 +18,153 @@
...
 /// It allows connecting to the Agama monitor to get the status or listen for changes.
 ///
 /// It can be cloned and moved between threads.
 #[derive(Clone, Debug)]
 pub struct MonitorClient {
+    /// Channel to send commands to the monitor task.
     commands: mpsc::Sender<MonitorCommand>,
-    pub updates: broadcast::Sender<api::Status>,
+    /// Channel to subscribe to status updates.
+    updates: broadcast::Sender<Result<InstallationStatus, MonitorBackendError>>,
```
imobachgs: I am not sure about sending the errors here. What can do the client about them? Perhaps printing them (for logging purposes) would be enough.

If it is a critical error, you can stop the monitor and the "client" will know because the channel gets closed.

rust/agama-lib/src/monitor.rs
```diff
@@ -18,93 +18,153 @@
...
+    status: Mutex<Result<InstallationStatus, MonitorBackendError>>,
+    /// Channel to broadcast status updates to subscribers.
+    updates: broadcast::Sender<Result<InstallationStatus, MonitorBackendError>>,
 }
 
+/// Commands that can be sent to the monitor task.
 #[derive(Debug)]
-enum MonitorCommand {
-    GetStatus(tokio::sync::oneshot::Sender<api::Status>),
+pub enum MonitorCommand {
```
imobachgs: I am not against, but perhaps we do not need a separate "command" for this (emitting the status when starting the monitor would be enough).

rust/agama-utils/src/api/status.rs
```diff
@@ -47,3 +47,9 @@ pub enum Stage {
     /// Installation failed
     Failed,
 }
+
+impl Stage {
+    pub fn is_end(&self) -> bool {
```
imobachgs: I would use something like `is_last`. `is_end` sounds weird to me.

imobachgs (commented):
Much better than our previous version. However, I have a few comments:

* I am not sure whether we should use the typical actors model here too.
* Testing the result, I wonder if we really need progress bars (let's review that in the upcoming presentation).
* If I connect from the very beginning, I get two issues: one for the missing product and another one for missing the user authentication. That's fine. But after selecting a product, the issues are not updated.
* We should improve the wording a bit.

rust/agama-lib/src/monitor.rs
jreidinger: yeah, completely agree. I would say old separated clients makes sense for dbus, but not that much for http API with much more limited surface.

rust/agama-cli/src/progress.rs
jreidinger: well, one result is receiving from tokio channel and second is result of monitor itself. I am not sure if we want to distinguish it in future.

rust/agama-cli/src/progress.rs
jreidinger: yes, it can be refactored to own method.

rust/agama-cli/src/progress.rs
jreidinger: so maybe something like "Installer tried to prepare target configuration but failed due to following issues:". What do you think?

rust/agama-cli/src/progress.rs
jreidinger: interesting, I always thought that we start with manager...but it is true that if you connect later, it can be messed up, so I should add such mechanism

rust/agama-utils/src/api/status.rs
jreidinger: well, original I want something is `is_finished` but we already have that state ( for me ideally we should have Succeed + Failed and is_finished, but I do not want to break everything now.....is_last sounds good enough.

rust/agama-cli/src/lib.rs
jreidinger: Ok, I add it to card https://trello.com/c/LHvZYUbF/5696-cli-non-interactive-mode and will remove this line.

rust/agama-cli/src/progress.rs
```diff
@@ -157,26 +156,26 @@ impl ProgressMonitor {
...
             // then check update of progresses
-            if self.state.status.progresses != new_state.status.progresses {
-                self.state = new_state;
+            if self.status.status.progresses != new_status.status.progresses {
+                self.status = new_status;
                 // no progress remaining, so just redraw screen
-                if self.state.status.progresses.is_empty() {
+                if self.status.status.progresses.is_empty() {
                     if self.render_state().await? {
                         continue;
```
imobachgs, line 180: Perhaps [my previous](https://github.com/agama-project/agama/pull/3377#discussion_r3091107077) comment was not clear enough. I did not want to extract this to a separate method, but to simplify the workflow. Compare and assign the state once and then update the UI accordingly.

I can give it a try tomorrow because it might be that I am missing something.

jreidinger: well, I want to avoid flickering...so whole comparison is about detection if we need screen redraw ( like when new issue appear, net question, or issue or questions changes ) or if change is only in progress ( very common change ). If there is just progress change, it will just update progress bars...if it need redraw it needs to clear progress bars, reset screen and then print new stuff. So much more complex stuff. And I need to detect what change, having single if that something change and redraw screen will cause flickering during package installation. 
Basically main reason is this progress update detection: https://github.com/agama-project/agama/pull/3377/changes/143c8fe27785b1c7b70eeac07df118b431e51e20#diff-a2165503d8f8178732d3975d30ed71202123400fb1585ffdf5d30fce9dc0b603R186

imobachgs: It looks like too much logic to me. We have been discussing about this issue and we decided to give Ratatui a try. It implements an immediate rendering approach and it might be more flexible in the long-term.

imobachgs (approved):

PR #3379, opened 2026-04-13
Fix TW MicroOS Software Selection

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1261991

## Problem

Agama complains about "Invalid Software Selection", and the user can't fix it:

> Could not select pattern 'sudo-policy-wheel-auth-self' for installation.

## Cause

It's listed in the product definition as a mandatory _pattern_, but it's a _package_.

## Fix

Moved it from _mandatory patterns_ to _mandatory packages_ in the product definition.

## Other Products?

This is the only product that is affected; in all others, it's correctly listed under _mandatory packages_, not _mandatory patterns_.

## Test

Hope and pray and put our faith in Santiago to test it manually :smiley:

### DIFF
--- a/products.d/agama-products.changes
+++ b/products.d/agama-products.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Apr 13 14:47:11 UTC 2026 - Stefan Hundhammer <shundhammer@suse.com>
+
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM, looks like it was introduced by https://github.com/agama-project/agama/pull/3215

PR #3380, opened 2026-04-13
feat: reuse LVM volume groups

joseivanlopez:
Feature for reusing LVM volume groups:

* https://github.com/agama-project/agama/pull/3303
* https://github.com/agama-project/agama/pull/3333
* https://github.com/agama-project/agama/pull/3330


ancorgs (commented):
Sometimes it is possible to get an UI error around `ConfigEditor` because it looks like `config` can be null there. I cannot reproduce it consistently but I'm pretty sure it can happen and we should defend against it before merging.

ancorgs (approved):
Not that I added a guard class to (hopefully) mitigate that error I was not able to reproduce... this LGTM.

PR #3382, opened 2026-04-14
setup-services.sh - fixed modifying the webserver path

lslezak:
## Problem

- The `setup-services.sh` script was not updated in https://github.com/agama-project/agama/pull/3336

## Solution

- Update the script to match the current configuration


mvidner (approved):
I have tested this with `/testing_using_container.sh`

PR #3383, opened 2026-04-14
Improve OpenAPI specification

imobachgs:
This PR enhances the OpenAPI specification for the Agama HTTP API, making it more complete, professional, and easier to consume.

## Unified Specification
- Merged all API specs (config, misc, profile) into a single `agama.json` file
- Eliminated need to manually merge multiple OpenAPI documents
- Better compatibility with standard tooling (Swagger UI, Postman, code generators)

## Complete Documentation
- Added response body schemas to all GET endpoints (`/api/v2/status`, `/api/v2/config`, `/api/v2/proposal`, etc.)
- Added previously undocumented endpoints:
  - `GET /api/v2/status`
  - `GET /api/v2/issues`
  - `GET /api/software/licenses/:id`
- **NEW**: Added all authentication endpoints:
  - `POST /api/auth` - Login with password
  - `GET /api/auth` - Check session
  - `DELETE /api/auth` - Logout
  - `GET /login` - Login from query params

## Server Configuration
- Added server URLs with variable support:
  - `http://localhost` (local development)
  - `https://{agamaHost}` (configurable, defaults to "agama.local")

## Better Metadata
- Improved API titles ("Agama Configuration API" vs "Config HTTP API")
- Added comprehensive descriptions for each API section
- Version now derived from `CARGO_PKG_VERSION`

## Code Quality
- Simplified OpenAPI generation code
- Better code organization while maintaining single output file


rust/install.sh
```diff
@@ -71,7 +71,7 @@ install6 -D -t "${DESTDIR}${datadir}"/zsh/site-functions "${SRCDIR}"/out/shell/_
 install6 -D -t "${DESTDIR}${datadir}"/fish/vendor_completions.d "${SRCDIR}"/out/shell/"${NAME}".fish
 
 # install OpenAPI specification
-install6 -D -t "${DESTDIR}${datadir}"/agama/openapi "${SRCDIR}"/out/openapi/*
+install6 -D -t "${DESTDIR}${datadir}"/agama/openapi.json "${SRCDIR}"/out/openapi.json
```
jreidinger: this does not look correct. 
```
-t, --target-directory=DIRECTORY
              copy all SOURCE arguments into DIRECTORY
```
So I kind of expect that it will create openapi.json directory and into it create file?

jreidinger: So I think it should be something like 

```suggestion
install6 -D -t "${DESTDIR}${datadir}"/agama "${SRCDIR}"/out/openapi.json
```

but please test it.

rust/agama-server/src/server/web.rs
```diff
@@ -149,7 +149,7 @@ pub fn server_with_state(state: ServerState) -> Result<Router, ServiceError> {
     path = "/status",
     context_path = "/api/v2",
     responses(
-        (status = 200, description = "Status of the installation."),
+        (status = 200, description = "Status of the installation.", body = Status),
```
jreidinger, line 152: also for all others:
Does it make sense to add body also to status 400? So user know how error response will look like and can process errors?

jreidinger (changes requested):
that install issue looks critical for me...rest is something that can be improved later.

rust/install.sh
jreidinger: or you can change `-t` to `-T` to thread dest as regular file and not directory.

imobachgs: Good catch. I will give it a try.

rust/agama-server/src/server/web.rs
imobachgs: Good point, I will investigate a bit about that.

imobachgs: I will fix this problem in another PR. I need to check the error codes because, in some cases, we should return a 500 instead of a 400.

jreidinger (approved):

PR #3384, opened 2026-04-14
Add options for pervasive encryption

joseivanlopez:
Extends the storage JSON configuration to support pervasive encryption options (APQNs and key type). Updates the schema, adds JSON conversions for encryption properties, and adapts the encryption planner to properly assign APQN objects and key type to planned encrypted devices.

Example:

~~~json
{
  "storage": {
    "drives": [
      {
        "partitions": [
          {
            "filesystem": {
              "path": "/home"
            },
            "encryption": {
              "pervasiveLuks2": {
                "password": "notsecret",
                "apqns": ["01.0017", "02.0017"],
                "keyType": "CCA-AESCIPHER"
              }
            }
          }
        ]
      }
    ]
  }
~~~

Tested manually in a s390 installation.

Follow-up:

Checks for the selected APQNs will be addressed later in a separate card. Right now, the installation will fail (with a popup) in the commit phase if there is something wrong with the selected APQNs. The required checks are:
  
* APQN not found.
* Whether the APQN is online.
* Whether all the selected APQNs have the same master key (CCA vs EP11). 
* Whether the selected key type is valid according to the selected APQNs.

As a reference:

* AutoYaST filters out and logs an error for the following APQNs from the profile: not found, offline, missing master key. On the other hand, it does not check that all the APQNs use the same master key or whether the selected key type is compatible with the mode of the APQNs (https://github.com/yast/yast-storage-ng/pull/1402).
* The YasT Expert Partitioner is well prepared to only allow selecting APQNs configured with the same master key (https://github.com/yast/yast-storage-ng/pull/1399).

### DIFF
--- a/service/lib/agama/storage/configs/encryption.rb
+++ b/service/lib/agama/storage/configs/encryption.rb
@@ -69,7 +69,7 @@ module Agama
 
         # Type of the generated secure key (only for pervasive encryption).
         #
-        # @return [String, nil] "EP11-AES" or "CCA-AESCIPHER".
+        # @return [String, nil] Accepted key types: "EP11-AES", "CCA-AESCIPHER", "CCA-AESCIPHER".
         attr_accessor :pervasive_key_type
 
         def initialize
--- a/service/lib/agama/storage/configs/encryption.rb
+++ b/service/lib/agama/storage/configs/encryption.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2024-2025] SUSE LLC
+# Copyright (c) [2024-2026] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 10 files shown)


service/lib/agama/storage/configs/encryption.rb
```diff
@@ -61,6 +62,20 @@ class Encryption
...
         attr_accessor :key_size
 
+        # List of APQNs used for generating secure keys (only for pervasive encryption).
+        #
+        # @return [Array<String>]
+        attr_accessor :apqns
+
+        # Type of the generated secure key (only for pervasive encryption).
+        #
+        # @return [String, nil] Accepted key types: "EP11-AES", "CCA-AESCIPHER", "CCA-AESCIPHER".
```
jreidinger: what means nil here? Auto proposed value on some conditions or one is already predefined?

jreidinger (approved):

PR #3388, opened 2026-04-15
Fix copy network link

teclator:
## Problem

Agama is not copying the systemd network link files for renaming devices after the installation. It was originally done by the ruby services and when the rust part was implemented we only added the copy of the connections and the hostname.

- https://bugzilla.suse.com/show_bug.cgi?id=1261774
- Related to https://github.com/agama-project/agama/pull/2404 and https://github.com/agama-project/agama/pull/3006.

## Solution

Copy also the systemd network link files to the target system when present.

## Testing

- *Added a new unit test*
- *Tested manually with `ifname=test0:dc:e4:cc:27:14:a1 ip=test0:dhcp`*

image

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr 15 07:25:18 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Copy Agama the systemd network link files for renaming devices
+  after the installation (bsc#1261774).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Apr 15 07:25:18 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Copy Agama the systemd network link files for renaming devices
```
joseivanlopez:
```suggestion
- Copy the Agama systemd network link files for renaming devices
```

joseivanlopez (approved):
LGTM. Only a typo in the changelog.

PR #3389, opened 2026-04-15
refactor(web): migrate SoftwarePatternsSelection to TanStack Form

dgdavid:
## Summary
This PR transforms `SoftwarePatternsSelection` from a **page of checkboxes** into a **real, declarative form driven by TanStack Form**, making it consistent with how other areas of the app manage settings. Users can now interact with multiple patterns and submit changes explicitly with **Accept** or discard them with **Cancel**, rather than triggering backend updates on every toggle.

This is the first step in a broader **Software area revamp**, laying the foundation for future improvements in UX consistency and visibility of selected patterns.

Key benefits:
- **Consistent UX**: Behaves like other forms, avoiding surprises and making interaction predictable.
- **Reliable submission**: Only patterns that users actually touched are included in the backend update.
- **Maintainable code**: TanStack Form simplifies state management, validation, and submission logic.

## Why
Before this change, the software patterns selection was a list of checkboxes acting like switches: **every toggle immediately called the backend**, triggering a software progress overlay and **blocking user interaction until the request completed**. This made user interaction slow and frustrating.

With this migration:
- Users can make multiple selections and submit once.
- The form tracks which patterns were touched, so only intentional changes are applied.
- The UI now behaves consistently with other forms in the app, supporting proper validation, submission, and completion logic.

Under the hood, each pattern is a separate form field, enabling precise tracking and ensuring correct behavior on save.

## Under the Hood
This PR also brings improvements for developers and the broader Software area:
- **Reusable `usePristineSafeForm` hook**: Detects untouched forms and skips unnecessary business logic, available for other forms across the app.
- **Extracted utilities (`utils/software.ts`)**: Grouping, sorting, filtering, and selection-check helpers are now reusable and testable.
- **Improved test coverage**: Covers form behavior, touched fields, submission logic, and edge cases.
- **Clearer submission logic**: Add/remove operations now compute against initial form state rather than mixing API state and UI state.

Together, these changes make the form **safer, more predictable, and easier to maintain**, while providing tools that will benefit future work in the Software area.

## UX consideration

Before submission, users don’t immediately see which auto-selected patterns resulted from their selections. However, this is not exactly a regression: in the previous form, users might also not have realized some auto-selected patterns were added, especially if they were out of the visible viewport and the user navigated away to the installation summary instead of going back to the software page and been able to realize there were more stuff than they selected.

With the new form, users will be redirected to the main software page, where auto-selected patterns are expected to be _highlighted_ in a future iteration. This improvement is planned as part of the ongoing Software area revamp. Once implemented, the overall experience will be improved and predictable.

### Before

[screen-recording-2026-04-16T07-59-53.webm](https://github.com/user-attachments/assets/0f1fd774-ed45-4ee8-92d4-4c355b5606c6)

### After

[screen-recording-2026-04-16T08-00-28.webm](https://github.com/user-attachments/assets/b1dca04a-daa1-46df-a012-74912341c489)

### DIFF
--- a/web/src/components/software/SoftwarePatternsSelection.tsx
+++ b/web/src/components/software/SoftwarePatternsSelection.tsx
@@ -50,11 +50,10 @@ import { filterPatterns, groupPatterns, isPatternSelected, sortGroupNames } from
 
 /**
  * Form options for pattern selection.
+ * Each pattern is a boolean field (pattern name -> selected state).
  */
 const softwarePatternsFormOptions = formOptions({
-  defaultValues: {
-    selectedPatterns: [] as string[],
-  },
+  defaultValues: {} as Record<string, boolean>,
 });
 
 const NoMatches = (): React.ReactNode => <b>{_("None of the patterns match the filter.")}</b>;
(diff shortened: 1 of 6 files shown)


web/src/components/software/SoftwarePatternsSelection.tsx
```diff
@@ -32,196 +32,178 @@ import {
...
+    ...softwarePatternsFormOptions,
+    defaultValues: initialValues,
+    onSubmit: async ({ value: formValues, formApi }) => {
+      // Add: selected patterns that user touched OR were already USER-selected
+      const add = patterns
+        .filter((p) => {
+          const isSelected = formValues[p.name];
+          if (!isSelected) return false;
+
+          const isDirty = formApi.getFieldMeta(p.name)?.isDirty;
```
jreidinger, line 90: I would probably use `isTouched` here. As tanstack has is consistent, but other libraries including react hook form do not need it and can confuse reader. On other hand isTouched is unambitious.
https://tanstack.com/form/v1/docs/framework/react/guides/basic-concepts#understanding-isdirty-in-different-libraries

web/src/hooks/form.ts
```diff
@@ -46,6 +46,58 @@ const { useAppForm, withForm } = createFormHook({
...
+ *   validators: {
+ *     onSubmitAsync: async ({ value }) => validatePatterns(value),
+ *   },
+ *   onSubmit: async ({ value }) => {
+ *     await patchConfig(value);
+ *   },
+ *   onSubmitComplete: () => navigate(SOFTWARE.root),
+ * });
+ */
+function usePristineSafeForm(options) {
```
jreidinger: Is types auto infered here?

imobachgs (approved):

web/src/components/software/SoftwarePatternsSelection.tsx
dgdavid: I prefer isDirty. User can touch the form but keep it pristine. Subtle difference, but a difference.

web/src/hooks/form.ts
dgdavid: I think so, but has to recheck it.

web/src/components/software/SoftwarePatternsSelection.tsx
dgdavid: About differences, I think TanStack Form doc explain them very well. Even their reasoning.

jreidinger (approved):
if types is properly interfered, then fine for me.

web/src/hooks/form.ts
dgdavid: I was wrong. It was infering "any". Now fixed in https://github.com/agama-project/agama/pull/3389/commits/fb4e4b3c1f551b98f4428df2a6023d4fe2b6e060

imobachgs (approved):

PR #3390, opened 2026-04-16
Add desktop selection option

joseivanlopez:
Extend the yaml configuration of the product with the `desktop_selection` option. 

Possible values:

* `optional`: Suitable for server-oriented products. No warnings or alerts if a DE is not selected.
* `suggested`: The product suggests a DE. The UI will display an info alert in the summary and a confirmation prompt if none are selected.

This is the first step to implement the desktop selection feature. For more details see https://github.com/agama-project/agama/discussions/3335.

**Note for reviewers**: The target of this PR is a feature branch. Changelog entries will be added later.

### DIFF
--- a/products.d/leap_161.yaml
+++ b/products.d/leap_161.yaml
@@ -53,6 +53,7 @@ translations:
       SUSE Linux Enterprise Server.
     zh_CN: 基于 SUSE Linux Enterprise Server 构建的社区发行版的最新版本。
     zh_TW: 根據最新 SUSE Linux Enterprise Server 打造的最新社群版發行版本。
+desktop_selection: suggested
 software:
   installation_repositories:
     - url: https://download.opensuse.org/distribution/leap/16.1/repo/oss/$basearch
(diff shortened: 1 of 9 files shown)


imobachgs (approved):
I am not sure about having two enums `DesktopSelection` (it might be confusing). I guess they have different purposes, but...

joseivanlopez:
> I am not sure about having two enums `DesktopSelection` (it might be confusing). I guess they have different purposes, but...

Maybe having the very same name is a bit confusing, but they are different things. One of them represents the value in the product template, which is read from the yaml file (snake case) and the other one is the API value we report in the system info (camel case). I was thinking on having a simple string in the system info, but that is not good (an enum is the correct thing). And using the same enum type from product template is strange, because we need to publish a type in the api that does not belong to the API.

imobachgs:
I am still unsure. It looks like a simple type to me that it is now kind of duplicated. Not a big issue, but it does not feel right.

PR #3393, opened 2026-04-17
Upgrade axum to version 0.8.9

imobachgs:
## Problem

We are using a rather old axum version. Version 0.8 [was released](https://tokio.rs/blog/2025-01-01-announcing-axum-0-8-0) more than one year ago, so it is time to move forward.

## Solution

Ask Claude to update to the latest version.

- Updated dependency versions in rust/agama-server/Cargo.toml
- Removed #[async_trait] from FromRequestParts implementation (axum 0.8 uses [RPITIT](https://rustc-dev-guide.rust-lang.org/return-position-impl-trait-in-trait.html))
- Changed path parameter syntax from :id to {id} in route definitions
- Replaced nest_service("/", serve) with fallback_service(serve)
- Fixed WebSocket message to use .into() for Utf8Bytes conversion

## Testing

- Tested manually

## Follow-up

It is possible that we can rid of [async_trait](https://crates.io/crates/async-trait) once we have updated our dependencies.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -316,7 +316,7 @@ dependencies = [
  "openssl",
  "pam",
  "pin-project",
- "rand 0.9.2",
+ "rand",
  "sd-notify",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


mchf (approved):
lgtm

dgdavid (approved):

PR #3394, opened 2026-04-17
POC: Try boot partitions with desired and min sizes

ancorgs:
## Useless proof of concept

The logic to define boot-related partitions in yast2-storage-ng defines three sizes for each partition: min, desired and max.

When using that in Agama, desired is ignored and only min and max are taken into account. That's something we always wanted to re-evaluate (together with several other aspects related to proposing boot partitions).

This shows a possible implementation in which several different sets of boot-related partitions are calculated and tried. In this case, it makes a first attempt using the desired sizes as starting point and, if that fails, a second attempt using the min size as starting point.

Although there is some room for improvement, the implementation is fully functional... but kind of useless.

In the process of testing the result I realized there is no practical difference. The distance between min, desired and max is so small (especially in comparison with the rest of partitions to create) that both attempts always produce the same result.

- If the disk is big enough, the max size is always reached. That's the most common case (I would say more than 99% of the scenarios) because the boot-related partitions are tiny compared to the partitions of the operating system and because, as mentioned, the distance between min, desired and max is really small.
- In super-small disks where the desired size cannot be reached, trying first with desired and then with min makes no difference, you will end up using the min size anyways.

So I'm creating (and directly closing) this pull request only for documentation purposes and to be taken as inspiration/starting point if we decide in the future to try a different approach for booting. The general principle of trying several different boot setups can still be useful.

### DIFF
--- a/service/lib/y2storage/agama_proposal.rb
+++ b/service/lib/y2storage/agama_proposal.rb
@@ -134,11 +134,40 @@ module Y2Storage
 
       configure_ptable_types(devicegraph)
       devicegraph = clean_graph(devicegraph)
-      complete_planned(devicegraph)
       return if issues?
 
-      result = create_devices(devicegraph)
+      result = create_devices_with_boot(devicegraph)
       result.devicegraph
+    end
+
+    # Creates the devices doing several attempts, each one for a different set of boot partitions
+    #
+    # @raise [NoDiskSpaceError] if there is no enough space to perform the installation
+    # @raise [NotBootableError] if adding partitions is not enough to make the system bootable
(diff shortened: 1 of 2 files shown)


ancorgs:
> If the disk is big enough, the max size is always reached. That's the most common case (I would say more than 99% of the scenarios) 

I overlooked an important case here. When the mechanism to make space is resizing the existing partitions  "as needed" (without specifying a concrete target size). Then the min becomes very relevant since it is not true that "the max size is always reached".

So maybe time to re-reconsider this.

ancorgs:
> So maybe time to re-reconsider this.

Done as #3559.

PR #3401, opened 2026-04-20
Adapt gnome kiosk options

mchf:
## Problem

New gnome-kiosk (resp. mutter) intentionally disables virtual terminal switching. To enable that gnome-kiosk has to be started with --enable-vt-switch option

- [*Bugzilla link*](https://bugzilla.suse.com/show_bug.cgi?id=1261932)

## Solution

gnome-kiosk service is started with required switch

## Testing

- *Tested manually*

### DIFF
--- a/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
+++ b/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
@@ -1,6 +1,6 @@
 [Service]
 # This is intentional
 ExecStart=
-# Gnome resp. Mutter version >= blocks switching to virtual console by default
+# gnome-kiosk >= 50 blocks switching to virtual console by default
 # --enable-vt-switch enables that. bsc#1261932
 ExecStart=/usr/bin/gnome-kiosk --enable-vt-switch
--- a/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
+++ b/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
@@ -1,5 +1,7 @@
 [Service]
-# This is intentional
+# We have to explicitly clear ExecStart which (most probably)
+# exists in the original service file. If not systemd will try
+# to start every ExecStart (original even the one added bellow)
 ExecStart=
 # gnome-kiosk >= 50 blocks switching to virtual console by default
 # --enable-vt-switch enables that. bsc#1261932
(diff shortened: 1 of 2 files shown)


live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+# This is intentional
+ExecStart=
+# Gnome resp. Mutter version >= blocks switching to virtual console by default
```
DimStar77: gnome-kiosk >= 50; not mutter itself does not (otherwise this would be blocked in regular gnome-sessions too)

mvidner:
```suggestion
# Gnome (Mutter) blocks switching to virtual console by default since version 50
```

"resp" does not mean what you think, https://en.wiktionary.org/wiki/resp

mchf: fixed

live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+# This is intentional
```
mvidner: This is better than no comment, but only partially helpful.
Try forgetting what you know about this and then asking, "WTF".

Why is this here?

mvidner (changes requested):

live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
mchf: ok thanks ... commet rephrased

DimStar77: https://en.wiktionary.org/wiki/resp. - `resp.`, not `resp` (but in any case, this is no longer part of the comment as it was changed t gnome-kiost)

live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
mchf: modified the comment

mvidner (approved):
Thanks!

PR #3407, opened 2026-04-21
Option to enable systemd-boot preview

joseivanlopez:
Add kernel option `inst.systemd_boot_preview=1` for selecting systemd-boot bootloader. It fallbacks to grub2 bootloader if the system is not BLS compliant.  

Note: The target of this PR is a feature branch. The changelog entries will be added later.

### DIFF
--- a/service/lib/agama/cmdline_args.rb
+++ b/service/lib/agama/cmdline_args.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright (c) [2022] SUSE LLC
+# Copyright (c) [2022-2026] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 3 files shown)


service/lib/agama/cmdline_args.rb
```diff
@@ -53,7 +58,7 @@ def self.read_from(path)
         next unless option.start_with?(CMDLINE_PREFIX)
 
         key, value = option.split("=", 2)
-        key.delete_prefix!(CMDLINE_PREFIX)
+        key.delete_prefix!(CMDLINE_PREFIX).downcase!
```
jreidinger: well, I would do that downcase in `standardize` method if you want...as doing it here means that `inst.SYSTEMD_BOOT_PREVIEW` is valid, but `INST.SYSTEMD_BOOT_PREVIEW` will be skipped.

joseivanlopez: Good point! I didn't remember our agreement about forcing downcase and value equal to 1. Let's adapt it.

jreidinger (approved):

PR #3412, opened 2026-04-22
Adding support for configuring a BOND (UI)

teclator:
## Problem

Currently, in the UI there is no support for configuring some connection types like bond, bridge and vlans. This PR concretely is for adding support to bond configuration.

- https://trello.com/c/5vP1FCmd

## Solution

Added type selector allowing to select between different connection types and allowing to setup specific type configuration like the bond one.

## Testing

- *Added a new unit test*
- *Tested manually*

### New bond connection form
BondingConnectionNew

### List of connections with a bond
BondingConnectionList

### Bond connection details
BondingConnectionDetails


web/src/components/network/ConnectionForm.tsx
```diff
@@ -409,9 +469,33 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
   });
 
+  const typeOptions = () => [
+    {
+      value: ConnectionType.BOND,
+      label: N_("Bond"),
+      description: "Bond",
+    },
+    {
+      value: ConnectionType.WIFI,
```
dgdavid: NP: the PR is about adding Bond support but this line is adding the WiFi too. Out of the scope.

dgdavid (commented):
I haven't reviewed in depth it yet. But having just a first look, I would like to ask you to extract the Bond stuff components in similar way IPSettings were extracted. It should help to not have a ConnectionForm component/file crowded more than it is already.

web/src/components/network/ConnectionForm.tsx
teclator: You are right, in fact I originally added also Bridge and removed it.

teclator: But from other point of view, it should not allow to add a new **WIFI** connection but it should be shown when editing. Isn't it?

dgdavid: No, it shouldn't. It supposed you cannot change the type nor the name of a connection

teclator: you cannot that is why I show it disabled but the type is exposed..

dgdavid: But we're not using disabled status for things you cannot change. We can find a way to show user they are editing a X connection. 

Let's discuss it better in a short pair-programming or presentation session.

web/src/components/network/BondSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+              <form.AppField name="bondOptions">
+                {(field) => (
+                  <field.ArrayField
+                    label={
+                      // TRANSLATORS: label for the bond options field.
+                      _("Bond options")
+                    }
+                    helperText={
+                      // TRANSLATORS: helper text for the bond options field.
+                      _("E.g., primary=eth1")
```
dgdavid: What about adding one more example? Reading the code I see that this example is not valid for the default mode option. Not a big dial, though. But might be useful to add another example.

web/src/components/network/BondSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+              <form.AppField name="bondPorts">
+                {(field) => (
+                  <field.ArrayField
+                    label={
+                      // TRANSLATORS: label for the bond ports field.
+                      _("Bond ports")
+                    }
+                    helperText={
+                      // TRANSLATORS: helper text for the bond ports field. %s is a list of available devices.
+                      sprintf(_("Available devices: %s"), devices.map((d) => d.name).join(", "))
```
dgdavid: We have a [formatList](https://github.com/agama-project/agama/blob/f3c0dff9db0cc6586a0891271d01d78d0e1f30f3/web/src/i18n.ts#L145-L154) wrapper around [Intl.List format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat). I believe it fits better than `join` here.

web/src/components/network/BondSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                    displayValue={(name) => {
+                      const device = devices.find((d) => d.name === name);
+                      return device ? sprintf(_("%s - %s"), device.name, device.macAddress) : name;
```
dgdavid: I'm not sure about this. Why display something different user enter?

web/src/components/network/BondSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    }}
+                    validateOnChange={(v) =>
```
dgdavid: This is not aligned with the rest of forms, nor with the rest of ConnectionForm fields. We're not validating on change, but on submit. https://github.com/agama-project/agama/blob/master/web/src/components/form/conventions.md#validation

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -90,6 +90,54 @@ describe("ConnectionForm", () => {
...
+    await user.click(screen.getByLabelText("Type"));
+    await user.click(screen.getByRole("option", { name: "Bond" }));
+    expect(screen.getByLabelText("Name")).toHaveValue("Bond");
+  });
+
+  it("does not show WIFI in the connection type selector for new connections", async () => {
+    const { user } = installerRender(<ConnectionForm />);
+    await user.click(screen.getByLabelText("Type"));
+    expect(screen.queryByRole("option", { name: "WIFI" })).not.toBeInTheDocument();
+  });
```
dgdavid: I would not add a test for something like this. Most probably is a test based on previous behavior you dropped but never land the `master` branch. Moreover, at some point Wi-Fi might be there.

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -90,6 +90,54 @@ describe("ConnectionForm", () => {
...
+
+  describe("Bond connection", () => {
+    it("shows bond fields when type is Bond", async () => {
+      const { user } = installerRender(<ConnectionForm />);
+
+      await user.click(screen.getByLabelText("Type"));
+      await user.click(screen.getByText("Bond"));
+      await waitFor(() => expect(screen.queryByLabelText("Bond mode")).toBeInTheDocument());
+      screen.getByLabelText("Bond mode");
```
dgdavid: These are repeated.

Also, I tend to avoid using .toBeInTheDocument expectation and rely on getByX or findByX


```suggestion
      await screen.findByLabelText("Bond mode");
```

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -90,6 +90,54 @@ describe("ConnectionForm", () => {
...
+
+      // Switch back to Ethernet
+      await user.click(screen.getByLabelText("Type"));
+      await user.click(screen.getByText("Ethernet"));
+
+      // It should not show the Bond's device name field anymore
+      expect(screen.queryByDisplayValue("bond0")).not.toBeInTheDocument();
+      // Binding mode should be back to "Any" (default)
+      expect(screen.getByLabelText("Device")).toHaveTextContent("Any");
+    });
```
dgdavid: I don't fully understand this. 

"Device name" and "Device" are different things, right? If not, please make them different, to allow preserving user entered value if they regret and switch back to Bond type again. Rationale related to https://github.com/agama-project/agama/blob/master/web/src/components/form/conventions.md#6-revealed-by-a-checkbox

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -319,6 +367,21 @@ describe("ConnectionForm", () => {
       expect(screen.queryByLabelText("Name")).not.toBeInTheDocument();
     });
 
+    it("disables the connection type selector", () => {
+      installerRender(<ConnectionForm />);
+      expect(screen.getByLabelText("Type")).toBeDisabled();
```
dgdavid: Why do you disable a selector? We are not doing this. There are defined more useful patterns https://github.com/agama-project/agama/blob/master/web/src/components/form/conventions.md#core-principle

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -319,6 +367,21 @@ describe("ConnectionForm", () => {
...
+    });
+
+    it("shows WIFI as the selected type when editing a WIFI connection", () => {
+      mockUseSystem.mockReturnValue({
+        connections: [buildConnection("wlan0", { wireless: { ssid: "test-wifi" } })],
+      });
+      mockParams({ id: "wlan0" });
+      installerRender(<ConnectionForm />);
+      expect(screen.getByLabelText("Type")).toHaveTextContent("WIFI");
```
dgdavid: NP: It is Wi-Fi.

I wonder, however, how important is this. Nothing against, no strong opinion, but I believe the breadcrumb already has context.

web/src/components/network/ConnectionForm.test.tsx
```diff
@@ -319,6 +367,21 @@ describe("ConnectionForm", () => {
...
+      expect(screen.getByLabelText("Type")).toBeDisabled();
```
dgdavid: Again, don't use disabled stuff.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -173,6 +180,9 @@ function connectionToFormValues(connection: Connection): Partial<FormValues> {
     dnsSearchList: unique(connection.dnsSearchList),
     customDns: connection.nameservers.length > 0,
     customDnsSearch: connection.dnsSearchList.length > 0,
+    bondMode: connection.bond?.mode ?? BondMode.BALANCE_ROUND_ROBIN,
+    bondOptions: connection.bond?.options ? connection.bond.options.split(" ") : [],
+    bondPorts: connection.bond?.ports ?? [],
```
dgdavid, line 196: As said, most probably we should have a bondDevice field too.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -252,7 +262,7 @@ function validateGateway(
  * values are valid. Validation is intentionally done here rather than in
  * per-field onSubmit validators — see the {@link ConnectionForm} remarks.
  */
-function validateConnectionForm(formValues: FormValues): FormFieldErrors | undefined {
+export function validateConnectionForm(formValues: FormValues): FormFieldErrors | undefined {
```
dgdavid: Why is it exported now? If answer is for testing, I believe this can be tested transitively, although I might be wrong.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -315,6 +325,32 @@ function validateConnectionForm(formValues: FormValues): FormFieldErrors | undef
       // TRANSLATORS: validation error for the DNS search domains field.
       _("Some DNS search domains are invalid"),
     ),
+    bondMode:
+      formValues.type === ConnectionType.BOND && !formValues.bondMode.trim()
+        ? // TRANSLATORS: validation error for the bond mode field.
+          _("Bond mode is required")
+        : undefined,
+    bondPorts:
```
dgdavid: Here is where I expect to see the device/ports validation stuff.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -315,6 +325,32 @@ function validateConnectionForm(formValues: FormValues): FormFieldErrors | undef
...
+      formValues.type === ConnectionType.BOND && formValues.bondPorts.length === 0
+        ? // TRANSLATORS: validation error for the bond ports field.
+          _("At least one bond port is required")
+        : undefined,
+    bondOptions:
+      formValues.type === ConnectionType.BOND &&
+      ![BondMode.ACTIVE_BACKUP, BondMode.BALANCE_TLB, BondMode.BALANCE_ALB].includes(
+        formValues.bondMode,
+      ) &&
+      formValues.bondOptions.some((o) => o.startsWith("primary="))
```
dgdavid: NP: this can be simplified by extracting it to helper method. We do not have a clear infrastructure to import validations yet, I know.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -379,16 +423,43 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
+  const syncName = (formApi) => {
+    const type = formApi.getFieldValue("type");
+
+    if (!formApi.getFieldMeta("name")?.isDirty) {
+      const existingIds = new Set(systemConns.map((c) => c.id));
+      formApi.setFieldValue("name", generateConnectionName(type, existingIds), {
+        dontUpdateMeta: true,
+        dontRunListeners: true,
+      });
+    }
```
dgdavid: Don't understand why changing the guard clause for nesting the logic instead 🤷‍♂️ 

This only needed to get and use the type, right?

web/src/components/network/ConnectionForm.tsx
```diff
@@ -379,16 +423,43 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
+        });
+      }
+      if (formApi.getFieldValue("bindingMode") === "iface") {
+        formApi.setFieldValue("bindingMode", "none", {
+          dontUpdateMeta: true,
+          dontRunListeners: true,
+        });
+      }
+    }
+  };
```
dgdavid: I really belive this nesting can be simplified. Hard to follow.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -409,9 +480,31 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
       },
     },
     onSubmit: () => navigate(-1),
-    listeners: isEditing ? undefined : syncNameListeners,
+    listeners: isEditing ? undefined : { onMount: ({ formApi }) => syncName(formApi) },
```
dgdavid, line 309: Which is the rationally for this change?

web/src/components/network/ConnectionForm.tsx
```diff
@@ -409,9 +480,31 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
+        value: ConnectionType.BOND,
+        label: N_("Bond"),
+        description: "Bond",
+      },
+      {
+        value: ConnectionType.WIFI,
+        label: N_("WIFI"),
+        description: "Wireless",
+      },
+      {
```
dgdavid: Still not understanding why adding Wi-Fi here. Out of the scope.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -409,9 +480,31 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
+        value: ConnectionType.ETHERNET,
+        label: N_("Ethernet"),
+        description: "Ethernet",
+      },
+    ];
+
+    return isEditing ? options : options.filter((o) => o.value !== ConnectionType.WIFI);
```
dgdavid: I expect options only be useful/available when creating a new connection.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -450,30 +543,22 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
...
-                  />
-                )}
-              </>
-            )}
-          </form.Subscribe>
-        </Flex>
+        <form.AppField name="type" listeners={{ onChange: () => syncName(form) }}>
+          {(field) => (
+            <field.DropdownField
+              isDisabled={isEditing}
```
dgdavid: No-go. We're not using disabled inputs.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -488,6 +573,39 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
           </form.AppField>
         )}
 
+        <form.Subscribe selector={(s) => s.values.type}>
+          {(type) =>
+            ![ConnectionType.BOND, ConnectionType.BRIDGE, ConnectionType.VLAN].includes(type) && (
```
dgdavid: Wouldn't be easier to mount this only on Ethernet type?

web/src/components/network/ConnectionForm.tsx
```diff
@@ -613,7 +731,9 @@ function EditConnectionForm() {
 
   if (!connection) return <ConnectionNotFound />;
 
-  return <ConnectionFormContent defaults={connectionToFormValues(connection)} isEditing />;
+  const connectionInstance = new Connection(connection.id, connection);
```
dgdavid: Why `connection` is not enough?

web/src/components/network/ConnectionsTable.tsx
```diff
@@ -116,9 +116,10 @@ const filterConnections = (
     }
 
     if (type && type !== "all") {
-      const isWifi = !!c.wireless;
-      if (type === "wifi" && !isWifi) return false;
-      if (type === "ethernet" && isWifi) return false;
+      const connType = c.type();
+      if (type === "wifi" && connType !== ConnectionType.WIFI) return false;
+      if (type === "ethernet" && connType !== ConnectionType.ETHERNET) return false;
+      if (type === "bond" && connType !== ConnectionType.BOND) return false;
```
dgdavid: Now I see it, wondering if we can go for a 


```suggestion
      return c.type() === type;
```

web/src/components/network/ConnectionsTable.tsx
```diff
@@ -153,8 +154,13 @@ const createColumns = (devices: Device[]) => [
   },
   {
     name: _("Type"),
-    value: (c: Connection) => (c.wireless ? _("Wi-Fi") : _("Ethernet")),
-    sortingKey: (c: Connection) => (c.wireless ? "wifi" : "ethernet"),
+    value: (c: Connection) => {
+      const type = c.type();
+      if (type === ConnectionType.WIFI) return _("Wi-Fi");
+      if (type === ConnectionType.BOND) return _("Bond");
+      return _("Ethernet");
```
dgdavid: We should have a reusable enum for the dropdown and this, in a way we can do `return ConnectionType[type].label` or so.

web/src/components/network/ConnectionsTable.tsx
```diff
@@ -285,6 +291,7 @@ export default function ConnectionsTable() {
                   all: _("All"),
                   wifi: _("Wi-Fi"),
                   ethernet: _("Ethernet"),
+                  bond: _("Bond"),
```
dgdavid: Another place where reusable connection type enum might help (or not)

web/src/components/network/ConnectionsTable.tsx
```diff
@@ -347,7 +355,7 @@ export default function ConnectionsTable() {
               title: _("Edit connection"),
               onClick: () => navigate(generatePath(NETWORK.editConnection, { id: c.id })),
             },
-            !isWifi && {
+            !(isWifi || isBond) && {
               id: "editBinding",
               title: _("Edit binding"),
```
dgdavid: I will create a card to kill the binding specific subform. It no longer makes sense, editing a connection should be enough.

web/src/components/network/WiredConnectionDetails.tsx
```diff
@@ -126,6 +126,23 @@ const NetworkDetails = ({ connection }: { connection: Connection }) => {
   );
 };
 
+const BondDetails = ({ connection }: { connection: Connection }) => {
+  return (
+    <Page.Section title={_("Bond")} pfCardProps={{ isPlain: false, isFullHeight: false }}>
+      <DescriptionList aria-label={_("Bond details")} isHorizontal>
+        <DescriptionListGroup>
+          <DescriptionListTerm>{_("Bond mode")}</DescriptionListTerm>
+          <DescriptionListDescription>{connection.bond?.mode}</DescriptionListDescription>
```
dgdavid: What is display when none?

web/src/components/network/WiredConnectionDetails.tsx
```diff
@@ -126,6 +126,23 @@ const NetworkDetails = ({ connection }: { connection: Connection }) => {
...
+        </DescriptionListGroup>
+        <DescriptionListGroup>
+          <DescriptionListTerm>{_("Bond options")}</DescriptionListTerm>
+          <DescriptionListDescription>{connection.bond?.options}</DescriptionListDescription>
```
dgdavid: Same here, which value for none?

web/src/types/network.ts
```diff
@@ -343,6 +361,17 @@ class Connection {
...
+    const { wireless, bond } = this;
+    if (wireless) {
+      return ConnectionType.WIFI;
+    } else if (bond) {
+      return ConnectionType.BOND;
+    } else {
+      return ConnectionType.ETHERNET;
+    }
+  }
+
```
dgdavid: I really don't like this. Would be nice to find a different way to deal whit this.

dgdavid (changes requested):
Thanks for taking care of this and moving it on 🙏 

We have room for improvements yet before considering it ready for master. Have a look to my comments.

web/src/components/network/BondSettings.tsx
teclator: It makes sense, in fact I modified the screenshot in the same way.

web/src/components/network/BondSettings.tsx
teclator: This is what I wanted to discuss, in order to have a better way to add the entries to the list, having more information like the MAC address and not just the device name, like having a table of devices where you can click to add the port.

web/src/types/network.ts
teclator: lets discuss about it

web/src/components/network/ConnectionsTable.tsx
teclator: Do we need a card?

web/src/components/network/BondSettings.tsx
teclator: In fact, maybe makes sense to not validate the names at all. When them are ethernet devices it makes sense to configure when present but in case of bridge, vlan or team ports it could not be the case.

web/src/components/network/ConnectionForm.test.tsx
teclator: The main issue here is when you set the device name like bond0 and you switch to other connection type, then the device selector kept the bond0 device selected where it is a virtual one which does not exist for an Ethernet device. In both cases it is the interface attribute of the connection but in one case it should match with an existing one and in another is how the virtual device will be named.

web/src/components/network/ConnectionForm.test.tsx
teclator: It was clear I would say ... xD

web/src/components/network/ConnectionForm.tsx
teclator: lets remove the field in case of editing being able to remove WIFI completely.

web/src/components/network/ConnectionForm.tsx
teclator: For me was consistent with the new connection form showing the value when editing as well, but just do not show it in case of editing.

web/src/components/network/ConnectionForm.tsx
teclator: it could be also modified in case of WIFI, isn't it? so, I went for using the virtual devices explicitly but let switch the other way around

web/src/components/network/ConnectionForm.tsx
teclator: I was not enough because of the Connection type, the connection is merged using a userConfig or systemConfig and in case of systemConfig it is an APIConnection which does not have the type function. But I have modified the code to not having a function in the Connection object but declare it in utils initializing the FormValues

web/src/components/network/WiredConnectionDetails.tsx
teclator: Fixed, nothing was displayed, so, unified for displaying `None set`. Although it should be always set ;)

web/src/components/network/ConnectionsTable.tsx
teclator: Adapted the code using the namespace pattern for the ConnectionType enum

teclator:
@dgdavid the changelog will be added once code looks good to you :*

web/src/components/network/BondSettings.tsx
dgdavid: > like having a table of devices where you can click to add the port.

As discussed, this is a not yet implemented field out of the scope of this PR.

web/src/components/network/ConnectionForm.test.tsx
dgdavid: Still, this does not apply the principles we are trying to follow in forms. Use two different form fields to manage this.

web/src/components/network/ConnectionForm.test.tsx
dgdavid: Please, check https://github.com/agama-project/agama/pull/3423

web/src/components/network/ConnectionForm.test.tsx
dgdavid: :sweat_smile:

web/src/components/network/ConnectionForm.tsx
dgdavid: > For me was consistent with the new connection form showing the value when editing as well,

But as commented in meeting, long ago was agreed, because reasons, to not use disable inputs in forms. I know I'm repeating myself here and you understood the missunderstanding alrady, but writing it down just in case someone else come to that comment. Check https://github.com/agama-project/agama/pull/3423

web/src/components/network/ConnectionForm.tsx
dgdavid: I was just asking. Overlooked the WiFi thing.

Maybe a `const BINDABLE_CONNECTION_TYPES` would to the work if there can be more types in the future.

web/src/components/network/ConnectionForm.tsx
dgdavid: Thanks a ton for simplifying it instead! Much appreciated :heart:

web/src/components/network/ConnectionsTable.tsx
dgdavid: > Do we need a card?

Don't know. But at least I'm trying to not mix too much stuff in a PR, to keep them managable.

web/src/components/network/ConnectionForm.test.tsx
dgdavid: Please, check https://github.com/agama-project/agama/pull/3424

dgdavid:
> @dgdavid the changelog will be added once code looks good to you :*

I still reviewing it. As part of the review, and in attempt to be more agile without sending changes that might collide with your local ones (if any), I've open 3 PRs already against this branch. Please, check them

* https://github.com/agama-project/agama/pull/3427
* https://github.com/agama-project/agama/pull/3424
* https://github.com/agama-project/agama/pull/3423
* https://github.com/agama-project/agama/pull/3421

web/src/components/network/ConnectionForm.tsx
dgdavid: I have called it bondIface in https://github.com/agama-project/agama/pull/3424

web/src/components/network/ConnectionForm.tsx
dgdavid: I have checked this by myself and addresed at https://github.com/agama-project/agama/pull/3427

web/src/components/network/ConnectionForm.tsx
```diff
@@ -173,6 +183,10 @@ function connectionToFormValues(connection: Connection): Partial<FormValues> {
     dnsSearchList: unique(connection.dnsSearchList),
     customDns: connection.nameservers.length > 0,
     customDnsSearch: connection.dnsSearchList.length > 0,
+    bondIface: connection.iface,
+    bondMode: connection.bond?.mode ?? BondMode.BALANCE_ROUND_ROBIN,
+    bondOptions: connection.bond?.options ? connection.bond.options.split(" ") : [],
+    bondPorts: connection.bond?.ports ?? [],
```
dgdavid, line 196: Wondering why bondOptions performs an split but bondPorts don't

dgdavid: Ah, I see. One is saved after join and the other just as an array of strings. 

https://github.com/agama-project/agama/pull/3412/changes#diff-a968a4f1125424dfa3c57a491dbba03747af997bf991e9e699fe115b52392174R400

I'd like to double check it.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -315,6 +329,32 @@ function validateConnectionForm(formValues: FormValues): FormFieldErrors | undef
...
+      formValues.type === ConnectionType.BOND &&
+      ![BondMode.ACTIVE_BACKUP, BondMode.BALANCE_TLB, BondMode.BALANCE_ALB].includes(
+        formValues.bondMode,
+      ) &&
+      formValues.bondOptions.some((o) => o.startsWith("primary="))
+        ? // TRANSLATORS: validation error for the bond options field when the 'primary' option is used in an invalid mode.
+          _(
+            "The 'primary' option is only valid for 'active-backup', 'balance-tlb', and 'balance-alb' modes",
+          )
+        : undefined,
```
dgdavid: All these `formValues.type === ConnectionType.BOND &&` made me think on extracting each _specific-connection-type-validations_ to their own helper method and ultimately to move all validations to an specific module for better DX.

Please, check https://github.com/agama-project/agama/pull/3437

It will help when adding more and more connection types.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -409,9 +466,11 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
       },
     },
     onSubmit: () => navigate(-1),
-    listeners: isEditing ? undefined : syncNameListeners,
+    listeners: isEditing ? undefined : { onMount: ({ formApi }) => syncName(formApi) },
```
dgdavid, line 309: Still having to test why this

dgdavid: Ah, sorry. It was onMount before, and for a reason: to make possible fill the auto-generated name fill the field in the first render, since it is not possible to do so through the `defaultValues`. I will proposed a change to add a comment there for the future.

I prefer and would kept the `onMount: syncName` signature, though, but I can live with the current change to that regard (`onMount: ({ formApi }) => syncName(formApi)`) if you find it better.

dgdavid, line 309:
```suggestion
    // On mount, auto-fill the name field (e.g., "Ethernet", "Bond 2") since
    // defaultValues can't access systemConns for duplicates. Changing the type
    // updates the name via the type field's onChange listener.
    listeners: isEditing ? undefined : { onMount: ({ formApi }) => syncName(formApi) },
```

web/src/components/network/ConnectionForm.tsx
dgdavid: Fixed in https://github.com/agama-project/agama/pull/3437

web/src/types/network.ts
```diff
@@ -29,6 +30,40 @@ import {
...
+export const ConnectionType = {
+  ...connectionTypeValues,
+
+  /** Returns the translated label for the connection type */
+  // eslint-disable-next-line agama-i18n/string-literals
+  label: (type: ConnectionType): string => _(connectionTypeLabels[type]),
+
+  /** Returns options for a dropdown/select component */
+  options: (types: ConnectionType[]): { value: ConnectionType; label: string }[] =>
```
dgdavid: Looks a bit weird to me having something like this in a type. This is pure dropdown ui stuff. Let me see if I can propose a different way to achieve same intention but decoupling from type.

dgdavid: Check https://github.com/agama-project/agama/pull/3438

https://github.com/agama-project/agama/pull/3438/changes#diff-a968a4f1125424dfa3c57a491dbba03747af997bf991e9e699fe115b52392174R356-R359

dgdavid (approved):
I think there is only a suggestion pending from my side plus my intention to revisit the bondPors/bondOptions types/differences. But let's move on.

Thanks!

PR #3369, opened 2026-04-09
Move encryption methods to the private API

imobachgs:
* [Trello](https://trello.com/c/OY1DkwSm/5638-2-fix-the-small-mess-with-encryption-method-ids)

The storage configuration now accepts encryption method names in both snake_case and camelCase
formats. This improves consistency with the rest of the JSON API, which uses camelCase for
property names (e.g., `tpmFde`, `pervasiveLuks2`).

## Changes

- **JSON Schema**: Updated `encryptionSwap` enum to accept both formats:
  - snake_case (backward compatible): `random_swap`, `protected_swap`, `secure_swap`
  - camelCase (new): `randomSwap`, `protectedSwap`, `secureSwap`
- **Ruby Service**: Added normalization in the encryption conversion layer to automatically convert camelCase to snake_case before looking up the encryption method
  in Y2Storage
- **TypeScript Types**: Regenerated type definitions to reflect the updated schema
- **Tests**: Added comprehensive test coverage for all six encryption method variants

## Example

Users can now use either format in their configuration:

```json
  {
    "drives": [{
      "partitions": [{
        "filesystem": { "path": "swap" },
        "encryption": "randomSwap"
      }]
    }]
  }
```

or the legacy format:

```
{
  "encryption": "random_swap"
}
```

Both map to the same Y2Storage::EncryptionMethod::RANDOM_SWAP internally.

##  Testing

- All existing tests pass (217 examples)
- Added 6 new test cases covering both naming conventions for all three swap encryption methods

##  Backward Compatibility

✅ Fully backward compatible - existing configurations using snake_case continue to work without changes.

PR #3413, opened 2026-04-22
Add bootloader to system JSON

joseivanlopez:
Extends the bootloader API to provide system-level information about which bootloaders can be used and what features they support.

Note: The target of this PR is a feature branch. A changelog entry will be added later.


joseivanlopez:
The private API introduced by https://github.com/agama-project/agama/pull/3369 is not needed anymore.

rust/share/system.bootloader.schema.json
```diff
@@ -0,0 +1,36 @@
...
+    }
+  },
+  "$defs": {
+    "bootloader": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["name", "encryptionAuth"],
+      "properties": {
+        "name": {
+          "enum": ["grub2", "grub2BLS", "systemdBoot"]
```
ancorgs: I'm not sure about this names. According to the name of the packages and the names used to create files and so on, it would be "grub2", "grub2-bls" and "systemd-boot".

I understand we use camelCase instead of camel_case for the names of the attributes and for ids invented by us (eg. the different values of encryptionMethod or encryptionAuthMethod). But in this case is not one of our ids. We don't need to camelcasify the world out there. :-)

service/lib/agama/storage/bootloader_prober.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        if arch.x86? || Yast::Arch.aarch64
+          tpm = bls_tpm?
+          grub2_bls = Bootloaders::Grub2BLS.new(tpm: tpm)
+          systemd_boot = Bootloaders::SystemdBoot.new(tpm: tpm)
+
+          return [grub2, grub2_bls, systemd_boot]
+        end
+
+        []
```
ancorgs: Does it make sense to return an empty array? I would feel better with something like this:

```ruby
def probe
  grub2 = Bootloaders::Grub2.new(tpm: grub2_tpm?)
  return [grub2] unless arch.x86? || (Yast::Arch.aarch64 && !raspberry_pi?)

  tpm = bls_tpm?
  grub2_bls = Bootloaders::Grub2BLS.new(tpm: tpm)
  systemd_boot = Bootloaders::SystemdBoot.new(tpm: tpm)
  [grub2, grub2_bls, systemd_boot]
end
```

service/lib/agama/storage/bootloader_prober.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+    private
+
+      # Whether TPM is usable for grub2.
+      #
+      # @return [Boolean]
+      def grub2_tpm?
+        arch.efiboot? && Y2Storage::EncryptionProcesses::FdeTools.new.tpm_present?
```
ancorgs: Wouldn't be cleaner to use `Y2Storage::EncryptionMethod::TPM_FDE.possible?`

service/lib/agama/storage/bootloader_prober.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+      # Whether TPM is usable for a BLS bootloader.
+      #
+      # @return [Boolean]
+      def bls_tpm?
+        arch.efiboot? && Yast::Arch.has_tpm2
+      end
+
+      # Whether the system is a Raspberry Pi.
```
ancorgs, line 64: This is the third place (two in `agama` and one in `yast-storage-ng`) in which we have this very same logic. Please add a big FIXME so we do not forget to fix it before merging into master.

service/lib/agama/dbus/storage/manager.rb
```diff
@@ -511,6 +537,52 @@ def system_issues_json
...
+            name:           bootloader_name_json(bootloader),
+            encryptionAuth: bootloader_encryption_auth_json(bootloader)
+          }
+        end
+
+        # Name of the bootloader according to the values of the JSON schema.
+        #
+        # @param bootloader [Agama::Storage::Bootloaders::Base]
+        # @return [String]
+        def bootloader_name_json(bootloader)
```
ancorgs: I understand we don't want to mix the bootloader types with this new objects. But things are starting to look weird.

For example, if `bootloader` would not be an `Agama::Storage::Bootloaders::Base` but an `Agama::Storage::BootloaderType` then this method would be implemented simply as this:

```
bootloader.value
```

service/lib/agama/storage/bootloaders/base.rb
```diff
@@ -0,0 +1,47 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Agama
+  module Storage
+    module Bootloaders
+      # Base class for the bootloaders.
+      class Base
```
ancorgs: Hmmm, we have three classes descending from this class and the implementation of the three is identical. That smells.

I have another suggestion that I will write later.

service/lib/agama/dbus/storage/manager.rb
joseivanlopez: Conceptually they are different things.  I would like to avoid mixing the enum (which represents the config options) with the object representing a bootloader probed from the system.

service/lib/agama/storage/bootloaders/base.rb
joseivanlopez: The code is minimal, but I agree for now we could have a single class with a #name or something similar. The current code is prepared to have different logic for each bootloader once they start growing and they differ. But I have no strong opinion.

ancorgs: Doesn't makes more sense to have a single class without inheritance? Something like this:

```ruby
class Whatever
  attr_reader :type

  def initialize(type, tpm: false)
    @type = type
    @tpm = tpm
  end

   def password_encryption_auth?
     true
   end

  def tpm_encryption_auth?
    @tpm
  end
end
```

joseivanlopez: Ok, let's go with something like this.

ancorgs (approved):
LGTM

PR #3414, opened 2026-04-23
Update the OpenAPI specification

imobachgs:
## Changes

- Migrates OpenAPI spec generation from `utoipa` to `aide` for better axum integration and more flexible schema generation with `schemars`.
- Update the OpenAPI specification:
  - Add summaries, operation IDs, descriptions, etc. 
  - Add missing schemas.
  - Add authentication information.

## Migration to aide

- **Replace routers**: `axum::Router` → `aide::axum::ApiRouter` throughout the server
- **Schema generation**: All structs now derive `schemars::JsonSchema` instead of using `utoipa` macros
- **OpenAPI generation**: Updated `xtask` to use aide's `ApiRouter` approach
- **Documentation**: Migrated all endpoints to aide's inline documentation style
- **Cleanup**: Removed `utoipa` dependencies

## Usage

OpenAPI generation command updated:

```bash
  cargo xtask openapi
```

## Having a look with Swagger UI

Screenshot_2026-04-24_09-34-24

## Follow-up

* [Future] Simplify the start-up of the server.
* Move `/profile` to `/private`.
* Reduce the code by using a function to add the common pieces (like response with code 500).
* Serve the `openapi` from Agama server?


rust/agama-network/src/model.rs
```diff
@@ -687,14 +687,15 @@ pub struct GeneralState {
...
 #[serde_as]
 #[skip_serializing_none]
-#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, utoipa::ToSchema)]
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Connection {
     pub id: String,
     pub uuid: Uuid,
-    #[schema(schema_with = schemas::mac_addr6)]
+    #[schemars(with = "Option<String>")]
```
jreidinger, line 695: does this mean that we loose a bit of strictness for value?

rust/agama-server/src/server/web.rs
```diff
@@ -453,17 +541,28 @@ async fn get_license(
...
-    path = "/action",
-    context_path = "/api/v2",
-    request_body(content = Action, description = "Description of the action to run."),
-    responses(
-        (status = 200, description = "Action successfully ran."),
-        (status = 422, description = "Action blocked by backend state", body = ErrorResponse),
-        (status = 500, description = "Internal server error.", body = ErrorResponse)
-    )
-)]
+#[allow(dead_code)]
```
jreidinger: why it is dead code? if it is intentional, I would at least document why it is needed.

rust/agama-utils/src/test/dbus.rs
```diff
@@ -75,14 +75,43 @@ impl DBusServer<Stopped> {
...
+        };
+
+        for candidate in candidates {
+            if candidate.exists() {
+                return candidate;
+            }
+        }
+
+        // Return the first candidate as fallback (will trigger error later)
+        std::path::PathBuf::from("../share/dbus-test.conf")
```
jreidinger, line 102: in fact this is not first candidate as it can be one from share dir. But probably it does not matter if it does not exist.

rust/agama-network/src/model.rs
imobachgs: Yes, by now we loose this one. I added it to the list of things to fix on upcoming iterations.

rust/agama-server/src/server/web.rs
imobachgs: Good point. It is actually not dead code.

jreidinger (approved):

PR #3415, opened 2026-04-23
Fixed medium check (bsc#1262311)

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1262311
- The medium check does not work when booting from HDD or USB flash
- It does not wait for a key press after printing the result

## Solution

- When when booting from HDD or USB flash the label points to a partition, the `checkmedia` tool expects whole disk as input (`/dev/sda`), not a partition (`/dev/sda2`). Try getting the parent device if there is any.
- Fixed waiting for a key press by specifying tty as input in the systemd service file

## Testing

- Tested manually

## Screenshots

Now it properly tests the `/dev/sda` disk and waits for a key press.

### When the check fails 
agama-checkmedium-failed

### When the check succeeds 
agama-checkmedium-success

### DIFF
--- a/live/live-root/etc/systemd/system/checkmedia.service
+++ b/live/live-root/etc/systemd/system/checkmedia.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Installation medium integrity check
 
-# before X11 because it switches the terminal to VT7
+# before starting the installer to possibly avoid installing from a broken medium
 Before=agama.service
 
 # copied from YaST2-Second-Stage.service
@@ -38,6 +38,7 @@ ExecStartPost=dmesg --console-on
 # enable back the systemd status messages on the console
 ExecStartPost=kill -SIGRTMIN+20 1
 
+StandardInput=tty
 StandardOutput=tty
 RemainAfterExit=true
 TimeoutStartSec=infinity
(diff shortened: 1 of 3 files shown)


mchf:
I don't see ```lsblk``` added to the installation image. Can we be sure that it is present?

lslezak:
I do not think we need to require it explicitly in the .kiwi file. It is required by dracut, KIWI and also by Agama itself so we can be pretty sure it is there :

```console
agama:~ # rpm -qf `which lsblk`
util-linux-systemd-2.41.3-5.1.x86_64
agama:~ # rpm -e util-linux-systemd
error: Failed dependencies:
	util-linux-systemd is needed by (installed) fde-tools-0.7.3-9.2.x86_64
	util-linux-systemd >= 2.36.2 is needed by (installed) dracut-110+suse.23.g5d9502c7-1.2.x86_64
	util-linux-systemd is needed by (installed) dracut-kiwi-live-10.2.38-1.3.x86_64
	util-linux-systemd is needed by (installed) ruby4.0-rubygem-agama-yast-20.devel22.00cff6770-1.1.x86_64
	util-linux-systemd is needed by (installed) agama-20+57.818e28a26-409.1.x86_64
```

mchf (approved):

PR #3416, opened 2026-04-23
Network backend fixes

teclator:
## Problem

Some problems have been detected in the API when working on some UI PBIs. 

1. The current connections list exposes the status of the connection, but the status is a field for defining if the connection should be up, down or to keep the current state( a user setting ). Instead of the connection status the list should expose the connection state.
2. The IPv6 method has more options that are not defined in the schema.
3. The network changes are applied even if the user config does not change.
4. If the user config has change it updates all the connections defined in the profile even when there are no real changes (a merge with the backend connection is the same).
5. When a controller ports are modified, them keep being defined as ports or even are created as duplicate.
6. When the network is written, it returns once the connections are updated (written) but the connections could get time to get activated or deactivated.

Related to https://trello.com/c/HXxZRy6T/5702-5-p2-1260465-network-restarts-during-software-config-update

## Solution

1. Bring back the connection state exposing it in the system configuration.
2. Define the missing methods for IPv6.
3. Store the user config and check if it has change or not.
4. Do a merge of the given user config with the backend to compare them properly.
5. Handle better the ports definition removing the orphaned ones.
6. Do not finish writing the configuration until all the generated active connections are activated or deactivated at least once.

This PR was also handled by https://github.com/agama-project/agama/pull/3340, but in order to get it merge ASAP I have moved some code from there.

## Tests

- Tested manually


rust/agama-network/src/nm/model.rs
```diff
@@ -192,6 +192,37 @@ impl TryFrom<NmConnectionState> for ConnectionState {
...
+
+impl From<NmConnectivityState> for u32 {
+    fn from(value: NmConnectivityState) -> u32 {
+        value.0
+    }
+}
+
+impl fmt::Display for NmConnectivityState {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
```
jreidinger, line 207: to be honest this is quite surprising for me as it prints just plain number...I kind of expect here that it will print some string like "Unknown" or "None" and if it needs serialize that u32, it can directly write it as it is as Display is mainly used for user readable strings.

rust/agama-network/src/nm/model.rs
```diff
@@ -192,6 +192,37 @@ impl TryFrom<NmConnectionState> for ConnectionState {
...
+    }
+}
+
+impl TryFrom<NmConnectivityState> for ConnectivityState {
+    type Error = NmError;
+
+    fn try_from(value: NmConnectivityState) -> Result<Self, Self::Error> {
+        match value {
```
jreidinger, line 215: maybe easier code will be to `match.0` so then there will be `0 => ...`

rust/agama-network/src/model.rs
```diff
@@ -1740,7 +1905,7 @@ impl TryFrom<BridgeSettings> for BridgeConfig {
         let stp = settings.stp;
         let priority = settings.priority;
         let forward_delay = settings.forward_delay;
-        let hello_time = settings.forward_delay;
+        let hello_time = settings.hello_time;
```
jreidinger, line 1908: good catch ;)

rust/agama-network/src/model.rs
```diff
@@ -255,25 +258,71 @@ impl NetworkState {
...
+                    tracing::info!("Adding connection {}", &conn.id);
+                    self.add_connection(conn.clone())?;
                 }
             }
 
+            // Although the current sanitized_connection_collection already handled the definition
+            // of port connections with the pertinent controller, the orphaned ports are problematic and
+            // needs to be handled in a particular way.
             for conn in connections.0 {
                 if conn.bridge.is_some() | conn.bond.is_some() {
```
jreidinger: are you sure that there should e bitwise `|` and not `||` ?

rust/agama-network/src/nm/adapter.rs
```diff
@@ -159,29 +192,79 @@ impl Adapter for NetworkManagerAdapter<'_> {
...
+                    }
+                };
 
-                return Err(NetworkAdapterError::Write(anyhow!(e)));
+                if conn.is_up() {
+                    if let Ok(path) = self.activate_connection(conn, path).await {
+                        tracing::info!("Activating connection {} with path {}", &conn.id, &path);
+                        active_paths.push(path)
+                    }
+                } else if conn.is_down() || conn.is_removed() {
```
jreidinger: If I get code, the `conn.is_removed()` cannot happen as this code is already in `else` branch on line 200 which is else of `is_removed` which is true if `conn.is_removed` is true

rust/agama-network/src/nm/adapter.rs
```diff
@@ -159,29 +192,79 @@ impl Adapter for NetworkManagerAdapter<'_> {
...
-                    .await
-                    .map_err(|e| NetworkAdapterError::Checkpoint(anyhow!(e)))?;
-                tracing::error!("Could not process the connection {}: {}", conn.id, &e);
+                        return Err(NetworkAdapterError::Write(anyhow!(e)));
+                    }
+                };
 
-                return Err(NetworkAdapterError::Write(anyhow!(e)));
+                if conn.is_up() {
+                    if let Ok(path) = self.activate_connection(conn, path).await {
```
jreidinger: Error from `self.activate_connection` is silently ignored here. Is it correct?

rust/agama-network/src/nm/adapter.rs
```diff
@@ -159,29 +192,79 @@ impl Adapter for NetworkManagerAdapter<'_> {
...
+                        tracing::info!("Activating connection {} with path {}", &conn.id, &path);
+                        active_paths.push(path)
+                    }
+                } else if conn.is_down() || conn.is_removed() {
+                    tracing::info!("Deactivating connection {}", &conn.id);
+                    let _ = self.client.deactivate_connection(path).await;
```
jreidinger: and same here. Is it logged at least?

rust/agama-network/src/model.rs
teclator: good catch, there were some changes in the code, probably in some of the rewrites i messed it up

rust/agama-network/src/nm/adapter.rs
teclator: Yep, the condition does not need the is_removed anymore, it was originally at 

https://github.com/agama-project/agama/pull/3416/changes/BASE..f3fd25adf3a9c5184767dd748b20306edf437400#diff-9e877651d54b9039c1ab5b6a6133aae49089f56e52424f0de2ee29ec00ebfad2L355

rust/agama-network/src/nm/adapter.rs
teclator: We could add more logs, but yes we ignore them in purpose. We are only interested in getting the active connction.

rust/agama-network/src/nm/model.rs
teclator: Well it mainly follows the same pattern used for NmConnectionState, see https://github.com/agama-project/agama/commit/e2c90dc3d7a791fe3f16bbc66863c7e199715470#diff-e3aa78cbc8b72175c6be1441fceaba9120ca88ac65c4d94df8dc05449e02a976R160-R193

As well as other types around an u32.

rust/agama-network/src/nm/model.rs
teclator: Same here.

rust/agama-network/src/nm/adapter.rs
teclator: Added logs

rust/agama-network/src/nm/adapter.rs
teclator: Added logs

jreidinger (approved):

PR #3417, opened 2026-04-23
feat(web): polish patterns selection page UX

dgdavid:
## TL;DR

Jump directly to Screenshots and Screencast sections at the end of the description.

## Summary

Polish the patterns selection page focused on two small UX wins:

- **Sticky category headers, inline counters, smarter filter feedback**: categories
  stay visible when scrolling; each one shows "X of Y selected"; filter feedback is contextual ("N of M patterns" / "No patterns match") instead of bare numbers.
- **"Auto selected" label fades on touch**: once the user explicitly toggles an auto-selected pattern, the badge steps aside (the choice is no longer purely
  automatic).

## Motivation

After revamping the software page, a few rough edges showed in the selection patterns form

- With long lists or small enough viewports, you lose track of which category you're in once you scroll past the heading.
- It was hidding categories on filter, making user lost the information about patterns selected on them.
- The search input badge was just a number, which works for sighted users who have the list right there but doesn't anchor the count to anything.
- The "auto selected" badge stayed on a pattern even after the user explicitly unchecked or rechecked it, which was misleading: the system isn't choosing it anymore, the user is.
- The badge color was loud enough to compete visually with the pattern title.

### Sticky headers + inline counters

- Category headings are sticky below the search input as the user scrolls.
- Each category shows "X of Y selected" (live, reacts to checkbox toggles).
- When filtering: appends "N matches the filter" or "No patterns match the filter" inline. The body placeholder is gone.
- The SearchInput results badge now reads "N of M patterns" or "No patterns match" instead of a bare number.
- Adds a screen-reader live region announcing filter result changes.
- Pattern rows refactored into a small `PatternCheckbox` so a non-clickable badge (the "auto selected" label) can sit next to the title without being part of the click target.

### "Auto selected" label hides on touch

- Reads `isDirty` from the field's TanStack Form meta and only renders the badge
  while the pattern is both AUTO in the proposal **and** untouched in the form.

### Subtler label color

- Switches `AutoSelectedLabel` to a different color that helps to read it as informative metadata rather than a call to action.

### Before

Screenshot_2026-04-23_10-24-28

### After

Screenshot_2026-04-23_10-23-27

## Screencast

https://github.com/user-attachments/assets/6e664788-c834-4beb-8b71-2f2afa2303af

## Notes for reviewers

> [!NOTE]  
> The target branch is a feature request. The changelog entry will be added later.

### DIFF
--- a/web/src/assets/styles/index.scss
+++ b/web/src/assets/styles/index.scss
@@ -266,6 +266,15 @@ strong {
   top: 1em;
 }
 
+.agm-sticky-category-header {
+  position: sticky;
+  background: var(--pf-t--global--background--color--primary--default);
+  z-index: 2;
+  padding-block-start: var(--pf-t--global--spacer--sm);
+  padding-block-end: var(--pf-t--global--spacer--sm);
+  border-bottom: var(--pf-t--global--border--width--regular) solid rgb(0 0 0 / 10%);
+}
+
 #productSelectionForm {
   input[type="radio"] {
     align-self: center;
--- a/web/src/components/software/AutoSelectedLabel.tsx
+++ b/web/src/components/software/AutoSelectedLabel.tsx
@@ -29,7 +29,7 @@ import { _ } from "~/i18n";
  */
 export default function AutoSelectedLabel(): React.ReactNode {
   return (
-    <Label color="blue" isCompact>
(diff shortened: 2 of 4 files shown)


lslezak (approved):
Nice improvement, thanks! 👍

PR #3418, opened 2026-04-23
fix issue with undefined element in iscsi config

jreidinger:
## Problem

During debugging iscsi issue Accept button do not react and on browser console was error.

## Solution

Fix issue by ensuring that radashi do not get undefined.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Apr 23 11:53:49 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix "Accept" button when connecting to iscsi target
+  (gh#agama-project/agama#3418)
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dgdavid (changes requested):
Great catch!

Thanks a lot for fixing it. We should review similar code to see if we have similar possibilities to crash. 

There are some linter complains, please fix them.

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Apr 23 11:53:49 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix "Accept" button when connecting to iscsi target
```
dgdavid, line 4: This sounds weird to me, sorry. "Accept" button has not be fixed, it was already working ;)

> Do not crash when connecting to iscsi target

dgdavid (approved):
Although I'm not so happy with the changelog entry, the fix LGTM. Let's move on.

PR #3419, opened 2026-04-23
[SLE-16] Replace symlinks by file copies (bsc#1261340)

lslezak:
## Problem

- Broken symlinks in generated *.gem file when creating the package in Leap 15.6 (openSUSE Tumbleweed and Leap 16.0 are fine)
- https://bugzilla.suse.com/show_bug.cgi?id=1261340

## Solution

- Avoid using symlinks, copy the files instead

## Notes

- It turned out that the broken symlinks are generated by Ruby in Leap 15.6. But we should still remove the symlinks to avoid the same problem accidentally in the future.
- The  `gem build` command in openSUSE Tumbleweed and Leap 16.0 actually also replaces the symlinks with file copies so the generated *.gem file should be the same. No change from the OBS POV, it should receive the same files.
- Added CI checks to ensure no new symlinks are introduced and to check that all copies are the same.
- SLE 16.1 is not affected, there are no symlinks in the Ruby sources, this is SLE 16.0 specific.

## Testing

- Tested manually
- Verified the added CI tests:
  - After modifying one file the [CI fails](https://github.com/agama-project/agama/actions/runs/24835688632/job/72695285073#step:9:9)
  - After adding a new symlink the [CI fails](https://github.com/agama-project/agama/actions/runs/24835912519/job/72696082450#step:10:12)

### DIFF
--- a/.github/workflows/ci-service.yml
+++ b/.github/workflows/ci-service.yml
@@ -109,6 +109,24 @@ jobs:
       run: bundle config set --local with 'development' && bundle install && bundle add ostruct &&
         bundle add cgi && bundle add fiddle
 
+    # "gem build" might create broken symlinks so we replaced them with a copy of the file,
+    # ensure that all copies are the same (see bsc#1261340)
+    - name: Check that the copied files are the same
+      run: |
+        cmp InstFunctions.rb ../../modules/InstFunctions.rb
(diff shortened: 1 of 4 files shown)


imobachgs (approved):
LGTM

PR #3422, opened 2026-04-24
Add TPM authentication to the config model

joseivanlopez:
This PR refactors the encryption model to use a `tpm` boolean field instead of explicit encryption method values (`luks2`, `tpmFde`). The actual encryption method is now determined based on the bootloader type and TPM 
  availability.

Note: The target of this PR is a feature branch. The changelog entries will be added later.


rust/agama-lib/share/storage.model.schema.json
```diff
@@ -26,7 +26,8 @@
       "required": ["configure"],
       "properties": {
         "configure": { "type": "boolean" },
-        "device": { "$ref": "#/$defs/bootDevice" }
+        "device": { "$ref": "#/$defs/bootDevice" },
+        "bootloader": { "$ref": "#/$defs/bootloaderType" }
```
jreidinger, line 30: so it means we allow to specify bootloader in autoinstallation profile?

rust/agama-lib/share/storage.model.schema.json
```diff
@@ -38,22 +39,18 @@
...
     },
+    "bootloaderType": {
+      "enum": ["grub2", "grub2-bls", "systemd-boot"]
+    },
     "encryption": {
       "type": "object",
       "additionalProperties": false,
       "required": ["method"],
       "properties": {
-        "method": { "$ref": "#/$defs/encryptionMethod" },
```
jreidinger, line 46: backward compatibility?

rust/agama-lib/share/storage.schema.json
```diff
@@ -930,7 +930,7 @@
     },
     "encryptionSwap": {
       "description": "Swap encryptions.",
-      "enum": ["protected_swap", "secure_swap", "random_swap"]
+      "enum": ["protected_swap", "secure_swap", "random_swap", "protectedSwap", "secureSwap", "randomSwap"]
```
jreidinger, line 933: I would write note to description that snake case methods are deprecated and can be removed in future?

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
```diff
@@ -47,13 +57,16 @@ def conversions
...
-              "luks1"  => Y2Storage::EncryptionMethod::LUKS1,
-              "luks2"  => Y2Storage::EncryptionMethod::LUKS2,
-              "tpmFde" => Y2Storage::EncryptionMethod::TPM_FDE
-            }
+            tpm = model_json[:tpm] || false
+            bootloader_type = bootloader_config.type
+
+            return Y2Storage::EncryptionMethod::TPM_FDE if tpm && bootloader_type&.is?("grub2")
+
+            if tpm && bootloader_type&.is?("grub2-bls", "systemd-boot")
```
jreidinger, line 65: NP: to be honest this "is?" usage is kind of strange when reading it as it miss info that it is any of it...so maybe I would call it `any?` with current arguments or `match?` with array.

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
```diff
@@ -47,13 +57,16 @@ def conversions
...
+              return Y2Storage::EncryptionMethod::SYSTEMD_FDE
+            end
 
-            method_conversions[model_json[:method]]
+            Y2Storage::EncryptionMethod::LUKS2
```
jreidinger, line 69: so LUKS1 is no longer supported?

service/lib/agama/storage/config_conversions/to_model_conversions/boot.rb
```diff
@@ -50,6 +56,14 @@ def convert_device
 
             ToModelConversions::BootDevice.new(config).convert
           end
+
+          # @return [String, nil]
+          def convert_bootloader
+            bootloader_type = bootloader_config&.type
+            return nil if bootloader_type.nil? || bootloader_type.is?(:none)
```
jreidinger, line 63: hmm, so it means that it is same if model do not specify bootloader at all and if it wants explicitly no bootloader to be installed? this looks kind of strange for me.

service/lib/agama/storage/config_conversions/to_model_conversions/encryption.rb
```diff
@@ -39,20 +38,17 @@ def initialize(config)
...
-              Y2Storage::EncryptionMethod::LUKS2.id   => "luks2",
-              Y2Storage::EncryptionMethod::TPM_FDE.id => "tpmFde"
-            }
+          # @return [Boolean, nil]
+          def convert_tpm
+            method = config.method
+            return unless method
 
-            method_conversions[config.method.id] || "luks2"
+            method.is?(:tpm_fde) || method.is?(:systemd_fde)
```
jreidinger, line 51: as you can see, here `is?` acts differently then in bootloader type case, so it is also inconsistent.

web/src/components/core/InstallationFinished.tsx
```diff
@@ -66,26 +67,28 @@ the machine needs to boot directly to the new boot loader.",
...
+//  * Partitions from MD RAIDs are not considered.
+function usingTpm(config?: Storage.Config): boolean {
+  if (!config) return false;
 
   const { drives = [], volumeGroups = [] } = config;
 
   const devices = [
     ...drives,
-    ...drives.flatMap((d) => d.partitions || []),
+    ...drives.flatMap((d) => ("partitions" in d ? d.partitions : [])),
```
jreidinger, line 81: what is reason for this change? can be d non-object? I found previous version more simple and readable.

web/src/components/core/InstallationFinished.tsx
```diff
@@ -66,26 +67,28 @@ the machine needs to boot directly to the new boot loader.",
...
     ...volumeGroups.flatMap((v) => v.logicalVolumes || []),
   ];
 
-  const root = devices.find((d) => d.filesystem?.path === "/");
+  const root = devices.find((d) => "filesystem" in d && d.filesystem?.path === "/");
```
jreidinger, line 85: same here. What case it tries to solve?

web/src/components/core/InstallationFinished.tsx
```diff
@@ -66,26 +67,28 @@ the machine needs to boot directly to the new boot loader.",
...
+  if (!root) return false;
+
+  const encryption = "encryption" in root ? root.encryption : null;
+  if (!encryption) return false;
 
-  return root?.encryption?.tpmFde !== undefined;
+  return typeof encryption === "object" && "tpmFde" in encryption;
```
jreidinger, line 91: can it happen that encryption is not object? it was previously protected against null or undefined.

web/src/components/core/InstallationFinished.tsx
```diff
@@ -66,26 +67,28 @@ the machine needs to boot directly to the new boot loader.",
...
-    ...drives.flatMap((d) => d.partitions || []),
+    ...drives.flatMap((d) => ("partitions" in d ? d.partitions : [])),
     ...volumeGroups.flatMap((v) => v.logicalVolumes || []),
   ];
 
-  const root = devices.find((d) => d.filesystem?.path === "/");
+  const root = devices.find((d) => "filesystem" in d && d.filesystem?.path === "/");
+  if (!root) return false;
+
+  const encryption = "encryption" in root ? root.encryption : null;
```
jreidinger, line 88: simple `  const encryption = root.encryption;` should be enough as it was checked before and it should be always object. We should not be paranoid about our own types.

rust/agama-lib/share/storage.model.schema.json
joseivanlopez: Not yet. The model used by the UI only informs about the configured bootloader type (i.e., grub2, but systemd-boot will be possible thanks to the kernel preview param).

rust/agama-lib/share/storage.model.schema.json
joseivanlopez: Not needed for the model, which is only used by the UI. The model is part of the private API.

web/src/components/storage/EncryptionSettingsPage.tsx
```diff
@@ -25,45 +25,54 @@ import { useLocation, useNavigate } from "react-router";
...
-  const [method, setMethod] = useState<ConfigModel.EncryptionMethod>("luks2");
 
   const passwordRef = useRef<HTMLInputElement>();
   const formId = "encryptionSettingsForm";
 
   useEffect(() => {
     if (configModel?.encryption) {
       setIsEnabled(true);
-      setMethod(configModel.encryption.method);
+      setTpm(configModel.encryption.tpm || false);
```
jreidinger, line 68: config model type specify that encryption can be also undefined. see https://github.com/joseivanlopez/agama/blob/995ec450eca0611160001961104ac0d628596669/web/src/openapi/storage/config-model.ts#L36

in such case this will crash....we really should use more strict typescript to find such things as it is not easy to do it during review....and next line is affected in the same way.

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
joseivanlopez: The `#is?` pattern is commonly adopted in the storage area. It was introduced long time ago to check the type of devices, for example `device.is?(:disk, :partition)`. And then it has been used for similar things. If you propose to change it, then I would prefer to agree the new name/approach with @ancorgs because he introduced this pattern.

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
joseivanlopez: The model (UI) will always use LUKS2.

web/src/hooks/model/system/bootloader.ts
```diff
@@ -0,0 +1,49 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import { useSuspenseQuery } from "@tanstack/react-query";
+import { systemQuery } from "~/hooks/model/system";
+import type { System, Bootloader } from "~/model/system";
+
+const selectSystem = (system: System | null): Bootloader.System | null =>
+  system?.bootloader ?? null;
```
jreidinger, line 28: to be honest here I am not sure if we should return `null` instead of `undefined` and exactly demonstrate problem with "none" set as bootloader...as for "none" bootloader it makes sense to use `null` in term of explicitly set missing value ( still I think that explicit "none" type is better ). But if bootloader is omitted then `undefined` is better value.

service/lib/agama/storage/config_conversions/to_model_conversions/boot.rb
joseivanlopez: The field `boot.bootloader` of the model only indicates the name of the configured bootloader type, if any. The field to indicate if the bootloader partitions have to be configured is `boot.configure`.

Probably you are mixing the meaning of the model and config. Currently, the config does not allow selecting a bootloader. The only way for agama to use a bootloader different to grub2 is by means of `systemd_boot_preview=1` kernel options (this will come in a separate PR). 

The UI needs to know what bootloader is selected (if any). So, for now, we have added `boot.bootloader` to the model to indicate the selected bootloader. This field will be probably removed once the bootloader type can be indicated in the config file (SLES 16.2).

web/src/openapi/system/bootloader.ts
```diff
@@ -0,0 +1,25 @@
...
+ * API description of the bootloader system
+ */
+export interface System {
+  /**
+   * Possible bootloaders to select in the current system.
+   */
+  availableBootloaders: Bootloader[];
+}
+export interface Bootloader {
+  type?: BootloaderType;
```
jreidinger: and here is what I do not like. Another potential null that has to be handled. I would use type in schema which is mandatory and have explicit "none" type as we will need it for support nfs based root.

service/lib/agama/storage/config_conversions/to_model_conversions/encryption.rb
joseivanlopez: Well, in this case I would allow passing more than one param to  `#is?`, because it is the common pattern everywhere in storage.

web/src/components/storage/EncryptionSettingsPage.tsx
joseivanlopez: But here we are safe because the `if(confgiModel?.encryption)` check, righ? But I agree, we need the strict mode as soon as possible.

web/src/components/core/InstallationFinished.tsx
joseivanlopez: The previous version lacks of type checking. Note that this code directly uses a `Config` instead of a config model. The `Config` type is quite complex, and that was one the reasons to introduce the model concept. In this code we need those `is` checks. For example, a drive can be `oneOf` several schemas, and some schemas do not have `#partitions` attribute.

Anyway, this code is going to be revisited in my next PBI. It is quite a transitory code.

web/src/components/core/InstallationFinished.tsx
joseivanlopez: See above.

web/src/components/core/InstallationFinished.tsx
joseivanlopez: Again, one of the schemas for encryption is an enum. In fact, it was very difficult to detect this case for me. Claude helped me with that.

web/src/components/core/InstallationFinished.tsx
joseivanlopez: It is the same. Some of the schemas have no `#encryption` property.

web/src/hooks/model/system/bootloader.ts
joseivanlopez: As said, this value from the model is only used to inform about the type of the configured bootloader, if any. And propably it will be dropped once the config allows configuring the bootloader.

web/src/openapi/system/bootloader.ts
joseivanlopez: This is a bug in the schema. I renamed the field `name` as `type`, but I did not change the `required` list, see https://github.com/joseivanlopez/agama/blob/model-bootloader-tpm/rust/share/system.bootloader.schema.json#L20

service/lib/agama/storage/config_conversions/to_model_conversions/encryption.rb
joseivanlopez: BTW, in this case the method comes from yast-storage-ng, https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/base.rb#L70

joseivanlopez:
We have raised 2 important topics as part of the review:

* The need of activating the TS strict mode to avoid unexpected errors when accessing to values that can be null.
* Avoid having too many optional attributes in the storage config model type, see https://github.com/agama-project/agama/blob/master/web/src/openapi/storage/config-model.ts#L34.

Note that the storage model comming from the backend usually has all the info, but the model you send to the backend can be very minimal. Right now, the very same type is used both cases, and for that reasons it contains a lot of optional attributes. Probably we will need 2 different types. One with mandatory attributes and other with optional ones.

web/src/components/storage/EncryptionSettingsPage.tsx
jreidinger: yeah, you are right

web/src/components/core/InstallationFinished.tsx
jreidinger: so encryption can be either object or enum?

joseivanlopez: In the Config yes, but note the config is never used by the UI. This is the only exception and probably we will change it.

joseivanlopez: The special encryptions for swap haven't properties:

~~~json
{ "encryption": "protectedSwap" }
~~~

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
jreidinger: and what happens if user specify LUKS1 e.g. in autoyast profile? What will happen in UI?

service/lib/agama/storage/config_conversions/to_model_conversions/boot.rb
jreidinger: what I mean here is that "none" is valid bootloader option which means "I want to not install bootloader at all" so it is kind of strange to act same for explicit user selection of none bootloader and missing value.

web/src/components/core/InstallationFinished.tsx
jreidinger: Ok, I have to ask AI why the version is better and its over advantage over `??` is that it allows ts to narrow type scope...so as you said if there is drive type without partitions, it will be skipped. in plain js both works, but ts check if attribute is always defined.

jreidinger (approved):

service/lib/agama/storage/config_conversions/from_model_conversions/encryption.rb
joseivanlopez: In that case, if you use the UI,  then the encryption will be automatically changed to LUKS2. It is something I already discussed with @ancorgs. We consider this switch is quite acceptable, but we will address it with a bug if somebody complains.

service/lib/agama/storage/config_conversions/to_model_conversions/boot.rb
joseivanlopez: Yes, I know we have to distinguish between "not configured" and "not wanted", but this will be probably done with the config instead of the model.

PR #3423, opened 2026-04-24
feat(web): add ReadOnlyField component for forms

dgdavid:
While extending the `Network/ConnectionForm` to add bond support, the use of disabled inputs was proposed but rejected during review, in line with previously established decisions. Instead, forms should handle these values by directly rendering them in the DOM. To maintain a consistent look and feel with the rest of the form, this PR introduces a new `Field` component. It allows read-only information to be displayed alongside editable fields without relying on disabled inputs, which have known accessibility limitations.

Looking ahead, a richer UI element positioned on the right as a sidebar with useful information about the item being edited could be introduced. This would align with the existing widget displayed in both the change product form and the overview page. However, implementing this should be done in a dedicated PR, taking into account all available forms to ensure consistency and usefulness.

localhost_8080_ (37)

### DIFF
--- a/web/src/components/form/ReadOnlyField.test.tsx
+++ b/web/src/components/form/ReadOnlyField.test.tsx
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) [2026] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


teclator (approved):
LGTM

PR #3424, opened 2026-04-24
refactor(web): split bond interface name from binding mode field

dgdavid:
Bond connections require a device name (e.g., `bond0`) that is different from the interface binding mode used by other connection types. The binding mode determines how a connection matches a device (by interface name or MAC address), while a bond's interface name is the name of the virtual bond device itself.

This change introduces a dedicated `bondIface` field to store the bond device name, preventing confusion with the `iface` field used for binding mode selection. It also simplifies the type listener logic by removing bond-specific field synchronization that was overwriting user input when switching connection types.

This aligns with the conventions of preserving user input, allowing users to switch between connection types without losing their configuration.

### DIFF
--- a/web/src/components/network/BondSettings.tsx
+++ b/web/src/components/network/BondSettings.tsx
@@ -70,7 +70,7 @@ const BondSettings = withForm({
           </>
         )}
         {!isEditing && (
-          <form.AppField name="iface">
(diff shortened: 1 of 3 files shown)


teclator (approved):
LGTM

PR #3427, opened 2026-04-24
fix(web): move bond validation tests to ConnectionForm

dgdavid:
`BondSettings` validation tests were importing and testing `ConnectionForm`’s `validateConnectionForm` function. This creates an unnecessary coupling between `BondSettings` tests and the specific validation logic of `ConnectionForm`.

The validation rules such as required bond ports and primary option restrictions, actually belong to `ConnectionForm`, not `BondSettings`. If `BondSettings` were reused in a different form with different rules, these tests would become misleading.

It also moves the existing validation tests to the added `describe("validation")` block. While doing so, it was noticed that some other validations are missing. These, however, will be added in a separate PR that is not related to the implementation of bond connections.

### DIFF
--- a/web/src/components/network/BondSettings.test.tsx
+++ b/web/src/components/network/BondSettings.test.tsx
@@ -21,11 +21,11 @@
  */
 
 import React from "react";
-import { screen, waitFor } from "@testing-library/react";
+import { screen } from "@testing-library/react";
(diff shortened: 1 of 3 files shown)


teclator (approved):
LGTM

PR #3435, opened 2026-04-27
Added support for configuring a network bridge (UI)

teclator:
## Problem

AS already described in #3412, the UI does not support configuring some connection types like bridge and vlans. This PR concretely is for adding support to the UI for configuring a network bridge.

- https://trello.com/c/5vP1FCmd

## Solution

Added the bridge connection type, allowing to select it and configuring the different options based on the STP enablement.

## Testing

- Added a new unit test
- Tested manually

### New Bridge
BridgeSettings

### Bridge STP settings
Bridge STP selector
Bridge STP settings


web/src/components/network/connectionFormValidation.ts
```diff
@@ -235,6 +235,54 @@ function validateBondFields(formValues: FormValues): Partial<FormFieldErrors> {
...
+    }
+    if (bridgeHelloTime < 1 || bridgeHelloTime > 10) {
+      // TRANSLATORS: validation error for the bridge hello time field.
+      errors.bridgeHelloTime = _("Hello time must be between 1 and 10 seconds");
+    }
+    if (bridgeMaxMessageAge < 6 || bridgeMaxMessageAge > 40) {
+      // TRANSLATORS: validation error for the bridge max message age field.
+      errors.bridgeMaxMessageAge = _("Max message age must be between 6 and 40 seconds");
+    }
+  }
```
dgdavid: Why not keep using

```jsx
{
   bridgeForwardDelay: bridgeForwardDelay < 4 || bridgeFormardDelay > 30 ? _("The error") : undefined,
   bridgeHelloTime: ...
}
```

I believe is easier to read and mantain.

teclator: Although not sure what is more readable or easier to maintain... taking into account line lengths.. but fine with me.

dgdavid: Also, for these _in range_ validations you could use Radashi _inRange_ method https://radashi.js.org/reference/number/inRange/ **Bearing in mind that end is exclusive** (which it's surprising to me)

```jsx
{
   bridgeForwardDelay: inRange(bridgeForwardDelay,  4, 31) ? _("The error") : undefined,
   bridgeHelloTime: ...
}
```

or create our own _inRange_ helper if you find the end inclusive hard to deal with. 


---

Spoiler: I have been thinking (a lot) on using an Standard Schema validation library like Zod (https://zod.dev/api, https://tanstack.com/form/latest/docs/framework/react/guides/validation#standard-schema-libraries) but prefer to keep our validations as they are right now until we have more I+D time.

dgdavid: > Although not sure what is more readable or easier to maintain... taking into account line lengths.. but fine with me.

I think so. Or at least, it would be aligned with the same patterns than others instead of mixing nested if in one validator and _plain_ ternaries in others.

the `shake` function in the main/central `validateConnectionForm` is in charge of clean up the undefined values later.

teclator: radashi to the rescue ;)

teclator: > Also, for these _in range_ validations you could use Radashi _inRange_ method https://radashi.js.org/reference/number/inRange/ **Bearing in mind that end is exclusive** (which it's surprising to me)
> 
> ```js
> {
>    bridgeForwardDelay: inRange(bridgeForwardDelay,  4, 31) ? _("The error") : undefined,
>    bridgeHelloTime: ...
> }

yes, that is why I wrote radashi to the rescue ;)

> ```
> 
> or create our own _inRange_ helper if you find the end inclusive hard to deal with.
> 
> Spoiler: I have been thinking (a lot) on using an Standard Schema validation library like Zod (https://zod.dev/api, https://tanstack.com/form/latest/docs/framework/react/guides/validation#standard-schema-libraries) but prefer to keep our validations as they are right now until we have more I+D time.

web/src/components/network/BridgeSettings.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    expect(screen.queryByLabelText("Device name")).not.toBeInTheDocument();
+  });
+
+  it("shows STP options when STP is enabled", async () => {
+    installerRender(<TestForm />);
+
+    expect(await screen.findByLabelText("Priority")).toBeInTheDocument();
+    expect(await screen.findByLabelText("Forward delay")).toBeInTheDocument();
+    expect(await screen.findByLabelText("Hello time")).toBeInTheDocument();
+    expect(await screen.findByLabelText("Max message age")).toBeInTheDocument();
```
dgdavid: NP:

Is STP checked by default? Should it be?

Add an assertion to pair the behavior. I mean, assert first that it is checked and then that these inputs are in the document. Then assert is not checked, and check that these inputs are no longer in the document. Which make me think both tests can be join in just one.

BTW, when checking that something is in the document we tend to simply rely in screen.getByLabelText  or whatever getByX or findByX. These methods will crash if the thing you're looking for is not present. Not additional expect().toBeInTheDocument required. Only when you are checking it is not in the document, i.e queryByX not.toBeInTheDocument.

web/src/components/network/connectionFormValidation.ts
```diff
@@ -235,6 +235,57 @@ function validateBondFields(formValues: FormValues): Partial<FormFieldErrors> {
...
 
+/**
+ * Validates bridge STP specific fields
+ */
+function validateBridgeStp(formValues: FormValues): Partial<FormFieldErrors> {
+  if (!formValues.bridgeStp) return {};
+
+  return {
+    // For radashi inRange the start is inclusive but the end of the range is exclusive,
+    // therefore, take care of adding 1 to the end range.
```
dgdavid:
```suggestion
    // therefore, take care of adding 1 more to the end range.
```

web/src/model/network/types.ts
```diff
@@ -1,506 +0,0 @@
...
-  ConnectionOptions,
-  APIDevice,
-  IPAddress,
-  APINetworkProposal,
-  APINetworkSystem,
-  NetworkGeneralState,
-  Route,
-  APIRoute,
-  WifiNetwork,
-};
```
dgdavid, line 506: NP: I see you have decided to drop this as part of this bridge PR. Fine with me. I would add a note in the PR descriptin, though.

dgdavid (commented):
Thanks @teclator!

In general looks good. I haven't had time to review the helper texts yet. Please, be patient. But I have added a concern I have regarding the default value of STP, which I find wrong and miss-aligned with the way we're delivering forms.

web/src/components/network/BridgeSettings.test.tsx
teclator: STP is enabled by default, it is true that we allow and in case of checked, all the fields are also using their default values. In the schema, only the ports are required but in case of the UI I decided to define all the fields when enabled.

dgdavid: The convention is clear: show users minimal form and let them decide if they are gonna use that thing.

dgdavid: I meant, use STP disabled as default. User can enable it and then face these extra fields with their default values and change them if they want. But avoid revealing them from the very first time without user request.

dgdavid (commented):
@teclator 

I finally managed to find some time to manually check the UI added here (not the behavior). This led me to create three small PRs against this feature branch:

* https://github.com/agama-project/agama/pull/3458
* https://github.com/agama-project/agama/pull/3459
* https://github.com/agama-project/agama/pull/3460

The first two might seem unrelated at first, but this PR gave me the opportunity to work on them. They are small yet useful UI improvements. Please review and merge them if you don’t find any blockers.

The last one could be a bit more controversial. Besides adding the missing nesting, I also simplified the helper texts for STP-related fields. Please check them carefully and let me know if anything important is missing, but the goal is to keep the texts short and easy to digest, while still providing useful information to the users.

Thanks a ton!

dgdavid (commented):
@teclator 

I also faced below error while trying to edit an existing connection.

Image

I think it is related to changes introduces here, at commit https://github.com/agama-project/agama/pull/3435/commits/f28295453fd5b6d901d401ccb546c2eab53e8340

web/src/components/network/BridgeSettings.tsx
```diff
@@ -152,7 +152,9 @@ const BridgeSettings = withForm({
                             // field. Text inside square brackets [] is
                             // semantically and visually emphasized. Keep the
                             // brackets.
-                            _("Root bridge selection (0-61440). [Lower is higher priority].")
+                            _(
+                              "Root bridge selection (0-61440). [Lower is higher priority]. Default is 32768.",
```
dgdavid: I'm not sure to put the default there, since they are already filled the input in first render (https://github.com/agama-project/agama/pull/3460) nor how further we would like to go with helper texts. Something we can re-evaluate as project evolve, though. 

If you're going to keep them in the helper texts, please avoid redundancy

_Root bridge selection (0-61440). [Lower is higher priority]. Defaults to 32768._

More evident in the others

_Listening and learning time (4-30 seconds). Default is 15 seconds._

_Listening and learning time (4-30 seconds). Defaults to 15._



But in my opinion, these defaults value should not be there, specially now that you have switched to STP selector. Custom is custom, no matter defaults.

dgdavid: > But in my opinion, these defaults value should not be there, specially now that you have switched to STP selector. Custom is custom, no matter defaults.

For completeness, I mean that now users has the "Default" option with will use these defaults. When they choose the now called "Enabled" option, I expect hey now what they are doing and the values they want use.

dgdavid: > > But in my opinion, these defaults value should not be there, specially now that you have switched to STP selector. Custom is custom, no matter defaults.
> 
> For completeness, I mean that now users has the "Default" option with will use these defaults. When they choose the now called "Enabled" option, I expect hey now what they are doing and the values they want use.

More over, putting these defaults in the helper texts might make users think it is the value the system will choose if left it empty (especially with my suggestion above), which is not the case (they still unset). That reinforce my suggestion to drop them from the helper texts. Users might use Default option for using defaults or have knowledge about these fields if goes for Enabled option (which I would call Custom)

ancorgs: I agree with @dgdavid view here.

web/src/components/network/ConnectionForm.tsx
```diff
@@ -308,7 +308,7 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
 
     if (formApi.getFieldMeta("name")?.isDirty) return;
 
-    const existingIds = new Set(systemConns.map((c) => c.id));
+    const existingIds = new Set((systemConns || []).map((c) => c.id));
```
dgdavid: NP: I would say just using "?" was enough, 

```ts
new Set(systemConns?.map(....))
```

web/src/components/network/ConnectionForm.tsx
```diff
@@ -563,9 +563,9 @@ function EditConnectionForm() {
   // We have to be careful with the merge as there could be values from a
   // removed connection which is not valid anymore.
   let { connections: configConns } = useConfig();
-  configConns = configConns.filter((c) => c.status !== "removed");
+  configConns = (configConns || []).filter((c) => c.status !== "removed");
```
dgdavid: NP: not sure if it would be more readable to initialize configConns at the very beginning

const { connections: configConns = [] } = useConfig();

web/src/components/network/ConnectionForm.tsx
```diff
@@ -522,8 +605,14 @@ function ConnectionNotFound() {
...
-  const { connections: configConns } = useConfig();
-  const { connections: systemConns } = useSystem();
+
+  // We have to be careful with the merge as there could be values from a
+  // removed connection which is not valid anymore.
+  let { connections: configConns } = useConfig();
+  configConns = (configConns || []).filter((c) => c.status !== "removed");
+  let { connections: systemConns } = useSystem();
+  systemConns = (systemConns || []).filter((c) => c.status !== "removed");
+
```
dgdavid, line 615:
```suggestion

  let { connections: configConns = [] } = useConfig();
  let { connections: systemConns = [] } = useSystem();
  // Filter out removed connections before merging to avoid carrying over
  // stale entries that may still exist in persisted config or system state.
  configConns = configConns.filter((c) => c.status !== "removed";
  systemConns = systemConns.filter((c) => c.status !== "removed";

```

dgdavid (approved):
Once again, thanks a ton for taking care of this set of changes @teclator 

LGTM 🎉

dgdavid:
@teclator, post-approval comment

:point_down: :point_down: :point_down: 

> ### Bridge STP settings
> Bridge STP selector 

---

As said in IRC channel,

> \<dgdavid\> teclator: [....] I still thinking "Custom" is better than "Enabled"
> \<dgdavid\> teclator: "enable" status of such an option is already mentioned in its description. So, for me make sense having "Default", "Custom", and "Disable"

What is more, the description for "Disabled" option is kind of _forced_. Most probably it does not need description, although I understand you added one for _visual pairing_ all options.

dgdavid:
> 
> What is more, the description for "Disabled" option is kind of _forced_. Most probably it does not need description, although I understand you added one for _visual pairing_ all options.

Idea, 

| Option | Label | Description|
|-|-|-|
| Default | Default | Use the system default STP settings|
| Enabled | Custom | Configure STP settings manually|
| Disabled | Disabled | STP will not be used by this bridge|

</body></html>

dgdavid:
> 
> Idea,
> 
> Option	Label	Description
> Default	Default	Use the system default STP settings
> Enabled	Custom	Configure STP settings manually
> Disabled	Disabled	STP will not be used by this bridge

Or

| Option | Label | Description |
|-|-|-|
| Default | Default | Enabled with the system default settings|
| Enabled | Custom | Enabled with manual settings |
| Disabled | Disabled | Not used by this bridge |

dgdavid (approved):

PR #3420, opened 2026-04-23
feature(web): hint users about desktop selection

dgdavid:
## Summary

Desktop environment selection is now a first-class part of Agama web UI. This branch bundles together the changes from 8 previous PRs to prevent users ending up with a command-line-only openSUSE because not visiting the software page and selecting a DE. The installation summary reflects now the software selection in a more complete and informative way, the confirmation screen alerts when no desktop is selected on products that expect one, and the software page itself got a full UX revamp.

- Separate, clearly labeled sections for desktops and additional patterns
- Patterns page migrated to TanStack Form: users can confirm or cancel  the whole selection at once, with no backend calls triggered on pattern checkbox toggle.
- Product-aware behavior via YAML flags (`desktop_selection: suggested | optional`), no per-product code branching
- Accessibility improvements that also benefit sighted users

localhost_8080_ (27)

---

<details>
<summary>Click to show/hide more screenshots</summary>

---

localhost_8080_ (30)

---

localhost_8080_ (29)

---

localhost_8080_ (28)

---

localhost_8080_ (31)

---

localhost_8080_ (32)

</details>

## Motivation

Before this work:

- **Tumbleweed and Leap ship with no default desktop. Not explicitly selecting one from the software page produced a console-only install.** See https://github.com/agama-project/agama/discussions/3335
- The software page mixed desktop environments and patterns in a single checkbox list, with no visible distinction.
- Every checkbox toggle hit the backend instantly, blocking the UI behind a progress overlay more times than needed.
- The installation summary did not reflect the selected desktop.
- Auto-selected patterns (pulled in by dependencies) were indistinguishable from the user's own picks in the software page summary.

### 1. Product-level configuration (PRs #3390, #3403, #3406)

Products now declare their desktop-selection policy:

- `suggested` (Tumbleweed, Leap): the UI hints when no desktop is picked.
- `optional` (SLES and server products): absence is a valid choice, no hint.

Patterns can be flagged as "desktop" so the UI groups them correctly.

### 2. Software page revamp (PRs #3389, #3396)

- Desktops and additional patterns live in separate, labeled sections.
- The patterns page is now a proper form: pick freely, then Accept or Cancel.
- Auto-selected patterns carry a visible label so users see what the system pulled in.
- Long descriptions are truncated in the software summary for better experience.

### 3. Installation summary reflects real choices (PR #3409)

The software line on the overview now shows:

- `GNOME` / `KDE` / `Whatever` when a desktop is picked.
[...]

PR #3436, opened 2026-04-27
fix(web): avoid misleading UI when no desktops are available

dgdavid:
While presenting the work done at https://github.com/agama-project/agama/pull/3420, @joseivanlopez  identified a corner case not handled by it: when products do not offer any desktop environments.

The previous implementation could show misleading messages such as “No desktop selected,” offer links to select a desktop that didn't exist, or display an empty desktop selection page. While this didn't make Agama crash, it created a confusing and unexpected experience.

This PR ensures the UI adapts gracefully with a few behavioral changes:

* **Avoid misleading messages** by only showing “No desktop selected” when desktops are actually available, **regardless of whether the product is configured to suggest a desktop selection**. 
* **Prevent empty views and maintain UI consistency** by redirecting users away from the desktop selection page when no desktops exist and displaying a clear informational message in the software page section. The desktop section is intentionally **not hidden** in this case to preserve layout consistency.

## Before

localhost_8080_ (41)

## After

localhost_8080_ (39)

---

> [!NOTE]
>
> Although no screenshots are included, these changes also suppress the desktop selection alert in the confirmation dialog when no desktops are available.

### DIFF
--- a/web/src/components/overview/SoftwareSummary.tsx
+++ b/web/src/components/overview/SoftwareSummary.tsx
@@ -30,7 +30,11 @@ import Link from "~/components/core/Link";
 import Text from "~/components/core/Text";
 import { useProposal } from "~/hooks/model/proposal/software";
 import { useProgressTracking } from "~/hooks/use-progress-tracking";
-import { useIsDesktopMissing, useSelectedPatterns } from "~/hooks/model/system/software";
+import {
+  useAvailablePatterns,
+  useIsDesktopMissing,
+  useSelectedPatterns,
+} from "~/hooks/model/system/software";
 import { useIssues } from "~/hooks/model/issue";
 import { SOFTWARE } from "~/routes/paths";
 import { _, n_ } from "~/i18n";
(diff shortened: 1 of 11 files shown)


web/src/components/overview/SoftwareSummary.tsx
```diff
@@ -64,9 +68,10 @@ const patternsCount = (qty: number): string =>
 const Value = () => {
   const patterns = useSelectedPatterns();
   const isDesktopMissing = useIsDesktopMissing();
+  const { desktops: availableDesktops } = useAvailablePatterns();
   const desktops = patterns.filter((p) => p.desktop);
 
-  if (isDesktopMissing) {
+  if (isDesktopMissing && availableDesktops.length > 0) {
```
jreidinger: is desktop missing already do check for empty desktops. So this additional check is unnecessary:
https://github.com/agama-project/agama/pull/3436/changes#diff-2b4d237c9a362c4d731a8f94db4eaa5790d4b8fb6b8ec54c53da4b6687c30adeR92

dgdavid: Great catch! I messed up this with the NoDestkopAlert where I wanted to be extra-defensive just in case it is mounted without the usage of isDesktopMissing. But here we are using both, so I'll drop the availableDesktops extra check.

Thanks!

jreidinger (approved):

PR #3437, opened 2026-04-27
refactor(web): extract connection form validation to separate module

dgdavid:
Moves validation logic from `ConnectionForm.tsx` into a dedicated `connectionFormValidation.ts` module to prepare for adding Bridge, VLAN, and other connection types.

The previous approach had inline type checks (formValues.type === ConnectionType.BOND) scattered throughout validateConnectionForm, which would not scale well. The new structure uses type-specific validator functions (validateBondFields, validateCommonFields) that are composed together, making it straightforward to add new types.

Also makes use of  formatList for proper i18n of the mode names list for a primary option bond validation.


teclator (approved):
LGTM

PR #3442, opened 2026-04-29
Add a CLI "agama status" command

jreidinger:
## Problem

There is currently no command to quickly get info about agama status like if something is blocking installation or if it already finishes.

PBI: https://trello.com/c/SjdURfzJ/5656-3-cli-agama-status-command

## Solution

implement such command

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

text format
```
jreidinger@localhost:~/prace/agama/rust> target/debug/agama --host 10.100.51.247 --insecure status
The installer failed to calculate the installation proposal. There are issues blocking the installation.

Details:
Blocking issues:
  - No product has been selected.
  - Defining a user, setting the root password or a SSH public key is required
```

json format

```json
jreidinger@localhost:~/prace/agama/rust> target/debug/agama --host 10.100.51.247 --insecure status --format json
{
  "installation": "issues",
  "status": {
    "stage": "configuring",
    "progresses": []
  },
  "issues": [
    {
      "scope": "manager",
      "class": "no_product",
      "description": "No product has been selected."
    },
    {
      "scope": "users",
      "class": "users.no_auth",
      "description": "Defining a user, setting the root password or a SSH public key is required"
    }
  ],
  "questions": []
}
```

## Documentation

TODO: after approval to avoid CLI reference word changes

### DIFF
--- a/rust/agama-cli/src/status.rs
+++ b/rust/agama-cli/src/status.rs
@@ -62,9 +62,9 @@ impl InstallationEnum {
             return Self::Ready;
         }
         if status.status.stage == Stage::Configuring {
-            return Self::Proposing;
+            Self::Proposing
         } else {
-            return Self::Installing;
+            Self::Installing
         }
     }
 }
--- a/rust/agama-cli/src/status.rs
+++ b/rust/agama-cli/src/status.rs
@@ -22,6 +22,7 @@ use agama_lib::monitor::InstallationStatus;
 use agama_utils::api::status::Stage;
 use gettextrs::gettext;
 use serde::Serialize;
+use std::fmt;
 
 /// A single enum representing current state of the installation process.
 #[derive(Debug, Serialize, PartialEq)]
@@ -38,7 +39,7 @@ pub enum InstallationEnum {
     /// The installation is in progress.
     Installing,
     /// The installation finished successfully.
-    Succeed,
+    Succeeded,
     /// The installation failed.
     Failed,
 }
@@ -46,7 +47,7 @@ pub enum InstallationEnum {
 impl InstallationEnum {
     pub fn from_status(status: &InstallationStatus) -> Self {
         if status.status.stage == Stage::Finished {
-            return Self::Succeed;
+            return Self::Succeeded;
         }
         if status.status.stage == Stage::Failed {
             return Self::Failed;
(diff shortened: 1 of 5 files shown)


rust/agama-cli/src/status.rs
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) [2024] SUSE LLC
```
imobachgs:
```suggestion
// Copyright (c) [2026] SUSE LLC
```

rust/agama-cli/src/status.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// The installation is waiting for a question to be answered.
+    Question,
+    /// The installation has issues blocking it from starting.
+    Issues,
+    /// The installation is ready to start.
+    Ready,
+    /// The installation is in progress.
+    Installing,
+    /// The installation finished successfully.
+    Succeed,
```
imobachgs: np: if we use "failed", I think we should use "Succeeded". Or "fail" and "success".

rust/agama-cli/src/status.rs
```diff
@@ -0,0 +1,238 @@
...
+            return Self::Ready;
+        }
+        if status.status.stage == Stage::Configuring {
+            return Self::Proposing;
+        } else {
+            return Self::Installing;
+        }
+    }
+
+    pub fn print_human_readable(&self) -> String {
```
imobachgs: Why not implement `Display` (or `ToString`)?

rust/agama-cli/src/status.rs
```diff
@@ -0,0 +1,238 @@
...
+
+use agama_lib::monitor::InstallationStatus;
+use agama_utils::api::status::Stage;
+use gettextrs::gettext;
+use serde::Serialize;
+
+/// A single enum representing current state of the installation process.
+#[derive(Debug, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub enum InstallationEnum {
```
imobachgs, line 30: That `Enum` suffix is kind of weird. But if we do not find a better name, it is fine.

imobachgs (commented):
In general, it looks good. I am not a big fan of the `installation` key, but I cannot think of a better name now.

rust/agama-cli/src/status.rs
jreidinger: I am not sure if it is good match when it is localized. let me search a bit.

rust/agama-cli/src/status.rs
jreidinger: Yeah, I would be happy to use better name, but sadly status is already occupied. I will try to ask AI for better name suggestion that is not ambitious as I failed to find better one.

rust/agama-cli/src/status.rs
jreidinger: search found nothing against, so I will change it.

rust/agama-cli/src/status.rs
jreidinger: well, AI suggested to use `InstallationState` and `state` key in json. I am not sure if `state` and `status` in one json output won't be confusing.

jreidinger: I ask for more suggestions and answers are:

- summary
- condition
- overview

imobachgs: I will copy just what I said on IRC:

> while working on the monitor, I was thinking that perhaps we should flatten the "InstallationStatus" struct a bit; I would move the "status" fields to the top-level, so we could have { "status": "issues", "stage": "configuring", "progresses": []", ... }. It looks more consistent (we have "issues", "questions", etc. at top level.

imobachgs (approved):

PR #3444, opened 2026-04-30
Install packages matching the target locale (bsc#1260423)

lslezak:
## Problem

- The language related packages are not installed.
- https://bugzilla.suse.com/show_bug.cgi?id=1260423

## Solution

- When selecting the packages to install query the target system locale and select that locale to install in libzypp.

## Testing

- Tested manually.
- Works both for locales set using the web UI and using CLI.
- The English locale is automatically selected by libzypp, this works the same way as in YaST in SLE15.

After selecting the Czech language the solver run reports these selected locales:

```
PoolImpl.cc(logSat):134: [libsolv] job: install providing namespace:language(cs_CZ)
PoolImpl.cc(logSat):134: [libsolv]   - job Rule #20819:
PoolImpl.cc(logSat):134: [libsolv]     system:system.noarch [1] (w1)
PoolImpl.cc(logSat):134: [libsolv] job: install providing namespace:language(en)
PoolImpl.cc(logSat):134: [libsolv]   - job Rule #20820:
PoolImpl.cc(logSat):134: [libsolv]     system:system.noarch [1] (w1)
PoolImpl.cc(logSat):134: [libsolv] job: install providing namespace:language(cs)
PoolImpl.cc(logSat):134: [libsolv]   - job Rule #20821:
PoolImpl.cc(logSat):134: [libsolv]     system:system.noarch [1] (w1)
```

<details><summary>In the target system the Czech language related packages are installed.</summary>

```console
agama:/ # rpm -qa --provides | grep "locale\(.*:cs\)" | xargs -n1 rpm -q --whatprovides | sort -u
bash-lang-5.3.9-6.3.noarch
cnf-locale-0.9.0~0-1.3.noarch
coreutils-lang-9.11-1.1.noarch
cpupower-lang-6.19.12-13.37.noarch
cryptsetup-lang-2.8.6-1.1.noarch
diffutils-lang-3.12-1.2.noarch
findutils-lang-4.10.0-2.4.noarch
firewalld-lang-2.1.2-9.3.noarch
fwupd-lang-2.0.20-1.2.noarch
gcr-lang-4.4.0.1-3.1.noarch
gdk-pixbuf-lang-2.44.6-1.1.noarch
glib2-lang-2.88.0-1.1.noarch
glibc-lang-2.42-3.3.noarch
glibc-locale-2.42-3.3.x86_64
glib-networking-lang-2.80.1-2.1.noarch
gpg2-lang-2.5.18-1.1.noarch
grep-lang-3.12-1.2.noarch
gsettings-desktop-schemas-lang-50.1-1.1.noarch
gstreamer-lang-1.28.2-1.1.noarch
gstreamer-plugins-base-lang-1.28.2-1.1.noarch
gtk3-lang-3.24.52-1.2.noarch
gtk4-lang-4.22.3-1.1.noarch
gvfs-lang-1.60.0-2.1.noarch
info-lang-7.3-1.1.noarch
json-glib-lang-1.10.8-1.3.noarch
libbytesize-lang-2.12-1.3.noarch
libidn2-lang-2.3.8-1.2.noarch
libpwquality-lang-1.4.5-5.4.noarch
libsecret-lang-0.21.7-2.1.noarch
libsoup-lang-3.6.6-4.1.noarch
libstorage-ng-lang-4.5.314-1.1.noarch
man-pages-cs-4.28.0-1.2.noarch
nano-lang-9.0-1.1.noarch
net-tools-lang-2.10+1-4.2.noarch
NetworkManager-lang-1.54.3-3.3.noarch
notification-daemon-lang-3.20.0-8.3.noarch
parted-lang-3.7-1.1.noarch
plymouth-lang-22.02.122+94.4bd41a3-20.2.noarch
popt-lang-1.19-2.1.noarch
psmisc-lang-23.7-5.2.noarch
sed-lang-4.9-2.9.noarch
shared-mime-info-lang-2.4-3.7.noarch
sharutils-lang-4.15.2-10.2.noarch
systemd-lang-260.1-1.2.noarch
tar-lang-1.35-6.1.noarch
udisks2-lang-2.11.0-2.2.noarch
util-linux-lang-2.41.3-5.2.noarch
wget-lang-1.25.0-2.3.noarch
xkeyboard-config-lang-2.47-2.1.noarch
xz-lang-5.8.3-1.1.noarch
yast2-trans-cs-84.87.20260414.0f82ab3540-1.1.noarch
```

The `*-lang` packages match lots of different languages but the `man-pages-cs` and `yast2-trans-cs` are Czech specific.
</details>

### DIFF
--- a/rust/agama-files/Cargo.toml
+++ b/rust/agama-files/Cargo.toml
@@ -5,7 +5,6 @@ rust-version.workspace = true
 edition.workspace = true
 
 [dependencies]
-agama-l10n = { path = "../agama-l10n" }
 agama-software = { version = "0.1.0", path = "../agama-software" }
 agama-utils = { path = "../agama-utils" }
 async-trait = "0.1.89"
@@ -17,6 +16,7 @@ tokio = { version = "1.48.0", features = ["sync"] }
 tracing = "0.1.41"
 
 [dev-dependencies]
+agama-l10n = { path = "../agama-l10n" }
 tokio-test = "0.4.4"
 test-context = "0.4.1"
 serde_json = "1.0.145"
--- a/rust/agama-files/Cargo.toml
+++ b/rust/agama-files/Cargo.toml
@@ -5,6 +5,7 @@ rust-version.workspace = true
 edition.workspace = true
 
 [dependencies]
+agama-l10n = { path = "../agama-l10n" }
 agama-software = { version = "0.1.0", path = "../agama-software" }
 agama-utils = { path = "../agama-utils" }
 async-trait = "0.1.89"
--- a/rust/agama-software/src/model.rs
+++ b/rust/agama-software/src/model.rs
@@ -85,6 +85,7 @@ pub trait ModelAdapter: Send + Sync + 'static {
     async fn write(
         &mut self,
         software: SoftwareState,
+        l10n: Handler<agama_l10n::Service>,
         progress: Handler<progress::Service>,
     ) -> Result<WriteIssues, service::Error>;
 }
@@ -131,12 +132,19 @@ impl ModelAdapter for Model {
     async fn write(
         &mut self,
         software: SoftwareState,
+        l10n: Handler<agama_l10n::Service>,
         progress: Handler<progress::Service>,
     ) -> Result<WriteIssues, service::Error> {
         let (tx, rx) = oneshot::channel();
+        let l10n_proposal = l10n
+            .call(agama_l10n::message::GetProposal)
+            .await
+            .unwrap_or_default();
+
         self.zypp_sender.send(SoftwareAction::Write {
             state: software,
             progress,
+            l10n: l10n_proposal,
             security: self.security.clone(),
             question: self.question.clone(),
             tx,
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -22,7 +22,7 @@ use agama_security as security;
 use agama_utils::{
     actor::Handler,
     api::{
-        self,
+        self, l10n,
         software::{Pattern, SelectedBy, SoftwareProposal, SystemInfo},
         Issue, Scope,
     },
@@ -120,6 +120,7 @@ pub enum SoftwareAction {
     Write {
         state: SoftwareState,
         progress: Handler<progress::Service>,
+        l10n: Option<l10n::Proposal>,
         question: Handler<question::Service>,
         security: Handler<security::Service>,
         tx: oneshot::Sender<ZyppServerResult<WriteIssues>>,
@@ -226,6 +227,7 @@ impl ZyppServer {
             SoftwareAction::Write {
                 state,
                 progress,
+                l10n,
                 question,
                 security: security_srv,
                 tx,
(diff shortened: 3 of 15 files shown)


rust/agama-files/Cargo.toml
```diff
@@ -5,6 +5,7 @@ rust-version.workspace = true
 edition.workspace = true
 
 [dependencies]
+agama-l10n = { path = "../agama-l10n" }
```
jreidinger: As it is needed only for tests I would add it only as dev depency in `[dev-dependencies]` section. This would speed a bit build of files itself.

rust/agama-software/src/model.rs
```diff
@@ -131,12 +132,19 @@ impl ModelAdapter for Model {
...
         &mut self,
         software: SoftwareState,
+        l10n: Handler<agama_l10n::Service>,
         progress: Handler<progress::Service>,
     ) -> Result<WriteIssues, service::Error> {
         let (tx, rx) = oneshot::channel();
+        let l10n_proposal = l10n
+            .call(agama_l10n::message::GetProposal)
+            .await
+            .unwrap_or_default();
```
jreidinger, line 142: I am not sure if we should mask failure of getting proposal. On other hand if it failed, we will need to create new issue for it, which can be a bit tricky as I am not sure what we should suggest to user......but on other hand, masking that we failed to know user language and silently not install them is also not nice.

rust/agama-software/src/zypp_server.rs
```diff
@@ -497,6 +501,11 @@ impl ZyppServer {
             };
         }
 
+        if let Some(proposal) = l10n {
+            let locale = proposal.locale;
+            zypp.select_locale(locale.language, locale.territory);
```
jreidinger, line 506: are we sure that our locale IDs always match zypp ones? how it will behave if it does not match?

rust/zypp-agama/zypp-agama-sys/c-layer/lib.cxx
```diff
@@ -520,6 +520,33 @@ void free_products(const struct Products *products) noexcept {
   free((void *)products->list);
 }
 
+void select_locale(struct Zypp *_zypp, const char *language,
+                   const char *country) noexcept {
+  LOG_LOCATION("Selecting locale packages");
+
+  if (strlen(country) == 0) {
```
jreidinger: this looks wrong. I think it should check language and not locale as below we handle language without country properly.

lslezak: Ah, forgot to update after copy&paste...

rust/agama-software/src/zypp_server.rs
lslezak: Libzypp [does not care about the actual codes](https://github.com/openSUSE/libzypp/blob/master/zypp-logic/zypp/Locale.h#L35-L37), it is possible to use even an invalid ISO code. In that case it would just not select anything to install. So this should be safe.

rust/agama-software/src/model.rs
lslezak: I cannot image what could happen that we would not get the proposal. If we cannot get the proposal then something very bad happens and missing installed packages would be probably just a minor issue in that case...

jreidinger (approved):

PR #3445, opened 2026-04-30
refactor(web): improve UX of storage mount point inputs

dgdavid:
## Problem

The usage of the internally called `SelectTypeaheadCreatable` component (a wrapper around [PatternFly’s typeahead Select with support for creating new options](https://www.patternfly.org/components/menus/select/#typeahead-with-create-option)) has proven frustrating for users. When a user types a value and presses `Tab`, the input is discarded unless they explicitly select the “Use X” option, which is not intuitive. Also, its dropdown appearance had mislead some users into thinking it only supports predefined selections rather than free text input. [~~This is a deliberate design decision in PatternFly, not a bug,~~](https://github.com/agama-project/agama/pull/3445#issuecomment-4352350413) This makes the component fundamentally unsuitable for a field where free text must always be preserved.

## Solution

To address this usability problem, a lightweight [wrapper around a plain `<TextInput>` and a native HTML `<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist) has been added as a new `SuggestionsTextField` component. The browser handles suggestion presentation natively and does not require custom keyboard navigation, ARIA wiring, or focus management. The input behaves exactly like a normal text field, and suggestions reflect their purely optional, assistive nature instead of behaving like options users must choose from.

## Before vs after

| | Before (`SelectTypeaheadCreatable`) | After (`SuggestionsTextField`) |
|---|---|---|
| **Props per call site** | 5 required (toggleName, listName, inputName, clearButtonName, createText) | 2 required (value, suggestions) |
| **Tab behaviour** | Discards typed value unless "Use X" is explicitly clicked | Always preserves typed value |
| **Free text** | Requires explicit "Use /home" action to commit | Native, no forced selection |
| **Helper text** | "Select or enter a mount point" | "E.g., /home, /var, or swap" |

### Screenshots

On Firefox, with field focused and after a keystroke (not visible in the screenshot)

Screenshot_2026-04-30_13-30-28

---

<details>
<summary>Click to show/hide other screenshots in different browsers</summary>

---

In Chrome. Note the behavior cannot be appreciated in the screenshots, only here for illustrating the look and feel of the input when both, not focused and focused. Suggestions relies in native browser presentation, the same used for autocomple with previously entered values by users. I.e., a familiar auto-complete widget for users.

| Before | After |
|-|-|
| Screenshot_2026-04-30_13-19-29 |Screenshot_2026-04-30_13-19-50 |
| Screenshot_2026-04-30_13-19-23 | Screenshot_2026-04-30_13-18-44 |

---

In Android Chrome browser

| Without focus | With focus |
|-|-|
| photo_6023631039701388571_y | photo_6023631039701388572_y |
</details>

### Approaches evaluated

Several patterns were considered before landing on `datalist`:

The existing component (`SelectTypeaheadCreatable`) had two variants, both unsuitable for this use case:

| Variant | Problem |
|---|---|
| PatternFly `Select` with `variant="typeahead"` | Discards typed value on `Tab`. Deliberate design, not fixable |
| Typeahead with "Use X" create option | Still requires an explicit click/Enter; tabbing away still discards the value |

The following alternatives were evaluated as replacements:

| Approach | Verdict | Reason |
|---|---|---|
| WAI-ARIA combobox (built from `TextInputGroup` + `Menu` + `Popper`) | Deferred | Correct pattern, but all ARIA wiring is fully manual, VoiceOver `aria-selected` dual-state must be managed explicitly, and implementation cost is high. Worth revisiting when rich suggestions (icons, descriptions) are needed |
| Text input + separate "Browse suggestions" button | Rejected | Accessible and well-specified, but adds an extra `Tab` stop for every user. Genuine friction on a field used repeatedly |
| `Select` with "Other (specify…)" option | Rejected | Clean AT support, but penalizes the majority; free text is the norm here, not the exception |
| Native `<datalist>` | **Chosen** | Free text is the default, `Tab` always preserves the value, no ARIA needed, dynamic suggestions work. Ruled out for the full combobox vision (no rich markup in the dropdown), but the right fit for this iteration |

### Temporary workarounds (until TanStack Form migration)

Current storage forms call multiple `setState` calls on every change, causing excessive re-renders, validation on incomplete input, and occasional focus loss mid-typing. The new component uses two short-term mechanisms to handle this:

**400ms debounced `onChange`**  
Reduces parent re-renders while the user is typing. `onChange` is not called on every keystroke, only after the user pauses.

**Focus tracking**  
External value updates are ignored while the field is focused, preventing a race condition where a debounced callback could overwrite what the user just typed:

```
1. User types "/h"     internalValue = "/h" (onChange debounced)
2. User types "ome"    internalValue = "/home"
3. 400ms later         onChange("/h") fires → parent sets value = "/h"
4. Effect runs         without focus tracking, input resets to "/h"  ← bug
                       with focus tracking, external update ignored  ← fixed
```

Both workarounds will be removed after migrating to TanStack Form (validation on submit, no parent re-renders). The component is expected to be simplified at that point, check diff below.

<details>
<summary>Click to show/hide rough diff</summary>
---

```diff
diff --git a/web/src/components/form/SuggestionsTextField.tsx b/web/src/components/form/SuggestionsTextField.tsx
index da4695ead..4b55534ba 100644
--- a/web/src/components/form/SuggestionsTextField.tsx
+++ b/web/src/components/form/SuggestionsTextField.tsx
@@ -140,38 +140,12 @@ export type SuggestionsTextFieldProps = Omit<TextInputProps, "list"> & {
 function SuggestionsTextField({
   id,
   suggestions = [],
-  value: externalValue,
+  value,
   onChange,
-  onBlur,
   ...textInputProps
 }: SuggestionsTextFieldProps): React.ReactElement {
   const datalistId = `${id}-datalist`;
 
-  // Track if input is focused to prevent external updates while typing
-  const isFocused = useRef<boolean>(false);
-
-  // Ref to store the debounce timeout
-  const debounceTimeout = useRef<number | null>(null);
-
-  // Maintain internal state for the input
-  const [internalValue, setInternalValue] = React.useState<string>(String(externalValue || ""));
-
-  // Sync from external value only when not focused
-  useEffect(() => {
-    if (!isFocused.current) {
-      setInternalValue(String(externalValue || ""));
-    }
-  }, [externalValue]);
-
-  // Clean up debounce timeout on unmount
-  useEffect(() => {
-    return () => {
-      if (debounceTimeout.current !== null) {
-        window.clearTimeout(debounceTimeout.current);
-      }
-    };
-  }, []);
-
   // Memoize options to avoid recreating them on every render
   const options = React.useMemo(
     () =>
@@ -181,61 +155,14 @@ function SuggestionsTextField({
     [suggestions],
   );
 
-  const handleFocus = (event: React.FocusEvent<HTMLInputElement>) => {
-    isFocused.current = true;
-    if (textInputProps.onFocus) {
-      textInputProps.onFocus(event);
-    }
-  };
-
-  const handleChange = (event: React.FormEvent<HTMLInputElement>, newValue: string) => {
-    // Update internal state immediately for responsive typing
-    setInternalValue(newValue);
-
-    // Clear any pending debounced onChange
-    if (debounceTimeout.current !== null) {
-      window.clearTimeout(debounceTimeout.current);
-    }
-
-    // Debounce onChange to prevent excessive parent re-renders
-    if (onChange) {
-      debounceTimeout.current = window.setTimeout(() => {
-        onChange(event, newValue);
-      }, 400);
-    }
-  };
-
-  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
-    isFocused.current = false;
-
-    // Cancel debounced onChange and sync immediately on blur
-    if (debounceTimeout.current !== null) {
-      window.clearTimeout(debounceTimeout.current);
-      debounceTimeout.current = null;
-    }
-
-    if (onChange) {
-      onChange(event, internalValue);
-    }
-
-    // Sync with external value on blur in case parent transformed it
-    setInternalValue(String(externalValue || ""));
-
-    if (onBlur) {
-      onBlur(event);
-    }
-  };
-
   return (
     <>
       <TextInput
         id={id}
         type="text"
         list={datalistId}
-        value={internalValue}
-        onChange={handleChange}
-        onFocus={handleFocus}
-        onBlur={handleBlur}
+        value={value}
+        onChange={onChange}
         {...textInputProps}
       />
       <datalist id={datalistId}>{options}</datalist>
```
</details>

### Chrome/Edge dropdown arrow

A global CSS rule hides the native picker indicator (`::-webkit-calendar-picker-indicator`) to match Firefox's behavior and reinforce that this is a free-text field, not a forced-selection one.

## Note for reviewers

The form behavior is unchanged: once a mount point is selected, the rest of the form adapts if it matches one of the product’s known mount points. To avoid breaking the original behavior for now, the field triggers its callback either on onBlur or after a short typing timeout. This timeout, however, makes the _form reset_ more noticeable: a symptom of the form reacting automatically rather than being fully user-driven. Eliminating this automatic adaptation is a goal for future iterations, but it is outside the scope of this PR, which focuses on removing a confusing component.

If necessary, we could increase the timeout to make this behavior less _aggressive_, provided it does not interfere with selecting a suggestion from the datalist.


dgdavid:
> This is a deliberate design decision in PatternFly, not a bug,

I might be mistaken in the statement above, probably due to mixing findings from previous research. 

While self-reviewing the PR description, I revisited the related issues I had found and analyzed them more carefully. It seems this was actually the default behavior in PatternFly v4 back in 2021 (https://github.com/patternfly/patternfly-react/issues/5502). Later in the same year, it was addressed with the introduction of the `isInputValuePersisted` flag (https://github.com/patternfly/patternfly-react/pull/5751). However, the behavior appears to have been lost again in 2023 with the introduction of the new selector implementation in PatternFly v5 (https://github.com/patternfly/patternfly-react/pull/8825/changes#diff-236181a06caba4527b287d04560f995c5c16e0b5d92e21aded03683d966213c1). Or maybe it can be achieved in a different way now, didn't continue investigating.

In any case, the approach proposed in this PR is most likely the better choice for storage forms.

joseivanlopez (approved):
LGTM

dgdavid:
@jknphy, friendly ping.

Not sure if this could hit you. I'll try to check it tomorrow and reach you at Slack to see if you needs to clarify something for speedup the tests adaptation, if needed.

dgdavid:
> @jknphy, friendly ping.
> 
> Not sure if this could hit you. I'll try to check it tomorrow and reach you at Slack to see if you needs to clarify something for speedup the tests adaptation, if needed.

We have reverted this change in https://github.com/agama-project/agama/pull/3490, so it will not hit you by now :)

PR #3446, opened 2026-04-30
Add support for handling NTP configuration

imobachgs:
Implements NTP configuration management for both the running and target systems using chronyd.                                                                                                                                                                                                                                                                                                                                    

- New `agama-ntp` service following the proxy service pattern
  - Supports pool, server, and peer NTP sources with iburst/offline options
  - Writes configuration to `/etc/chrony.d/99-agama.conf` on live system
- Copies configuration to target system during installation
- Reloads chronyd automatically after configuration changes                                                                                                                                                                                                                                                                                                                                                                       

## Changes

- Created `agama-ntp` crate with model, service, and message layers
- Added NTP configuration API types to `agama-utils`
- Integrated NTP service into manager lifecycle (SetConfig/GetConfig/Finish)
- Updated JSON schema with NTP configuration support
- Uses `Vec<Source>` with custom merge strategy (overwrite non-empty only)

## Configuration example

```json
{
  "ntp": {
    "sources": [
      { "type": "pool", "address": "0.opensuse.pool.ntp.org", "iburst": true }
    ]
  }
}
```


rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Failed to write configuration: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("Failed to reload chronyd")]
+    ReloadFailed,
+}
+
+const CHRONY_CONFIG_DIR: &str = "etc/chrony.d";
+const CHRONY_CONFIG_FILE: &str = "99-agama.conf";
```
jreidinger: question is if we want to expose agama or use `99-installer.conf`

rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    lines.push(String::new());
+    lines.join("\n")
+}
+
+fn source_type_to_string(source_type: &SourceType) -> String {
+    match source_type {
+        SourceType::Pool => "pool".to_string(),
+        SourceType::Server => "server".to_string(),
+        SourceType::Peer => "peer".to_string(),
```
jreidinger: this looks kind of strange for me to have it here. What about having method in SourceType .e.g via Display trait?

rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+use agama_utils::api::ntp::{Config, Source, SourceType};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Failed to write configuration: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("Failed to reload chronyd")]
+    ReloadFailed,
```
jreidinger: I failed to find where this error is used.

rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+            .join(CHRONY_CONFIG_FILE)
+    }
+
+    fn install_config_path(&self) -> PathBuf {
+        self.install_dir
+            .join(CHRONY_CONFIG_DIR)
+            .join(CHRONY_CONFIG_FILE)
+    }
+
+    fn reload_chrony(&self) -> Result<(), Error> {
```
jreidinger: I do not see any `?` neither `Err` so I have impression that it cannot return Err Result.

rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+fn generate_chrony_config(sources: &[Source]) -> String {
+    let mut lines = vec!["# Generated by Agama".to_string()];
+
+    for source in sources {
+        let mut parts = vec![
+            source_type_to_string(&source.source_type),
+            source.address.clone(),
```
jreidinger: do we want some sanitization of address here? At least I would log it if user mess a bit string in config.

rust/agama-ntp/src/model.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    fn install(&self, config: &Config) -> Result<(), Error> {
+        let path = self.install_config_path();
+
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent)?;
+        }
+
+        let content = generate_chrony_config(&config.sources);
+        std::fs::write(path, content)?;
```
jreidinger: Is writting config enough or should we ensure that chrony systemd service is enabled?

imobachgs: I think it makes sense to enable the service, yes.

rust/agama-ntp/src/model.rs
imobachgs: We can check whether it is a valid address/name I guess.

rust/agama-ntp/src/model.rs
imobachgs: Good point.

rust/agama-ntp/src/model.rs
imobachgs: :+1:

rust/agama-ntp/src/model.rs
imobachgs: I need to check the code myself.

rust/agama-ntp/src/model.rs
imobachgs: Thinking about it, the point is that this conversion is specific to chrony. If tomorrow we support another daemon, the translation might be not even needed. However, I will do what you suggested. We can even use `strum` as we do in other places.

rust/agama-ntp/src/model.rs
imobachgs: Yes, you are right, it is related with the `Reload` error. I have reorganized the errors.

jreidinger (approved):

jreidinger (approved):

PR #3448, opened 2026-04-30
Add TPM BLS encryption method support

joseivanlopez:
Implements TPM-based encryption for BLS bootloaders (grub2-bls, systemd-boot) alongside the existing TPM FDE for grub2.

In the config file, the TPM unlocking is configured as follows:

~~~json
{
  "encryption": {
    "luks2": {
      "password": "notsecret",
      "tpm": true
    }
  }
}
~~~

Notes for reviewers:

* Tests are failing because they require changes in https://github.com/yast/yast-storage-ng/pull/1426.
* The target is a feature branch and the changelog entry will be added later.


service/lib/agama/storage/config_conversions/from_json_conversions/encryption.rb
```diff
@@ -60,23 +71,26 @@ def luks1?
...
           end
 
           # @return [Boolean]
-          def pervasive_luks2?
-            return false unless encryption_json.is_a?(Hash)
+          def tpm_fde?
+            return true if tpm_fde_schema?
 
-            !encryption_json[:pervasiveLuks2].nil?
+            luks2_schema? && encryption_json.dig(:luks2, :tpm) && !bootloader_config.type&.bls?
```
ancorgs: This would return true if the bootloader is NONE or any other possible value introduced in the future. But TpmFde only works with Grub2 -with the (open)SUSE patched version of Grub2 to be precise-. So I would explicitly check for Grub2.

service/lib/agama/storage/config_checkers/encryption.rb
```diff
@@ -84,41 +90,77 @@ def unavailable_method_issue
...
-        # @return [Issue, nil]
-        def wrong_method_issue
-          method = encryption&.method
-          return unless method&.only_for_swap?
-          return if config.filesystem&.path == Y2Storage::MountPoint::SWAP_PATH.to_s
+        # Whether the bootloader is grub2 and the encryption method cannot be used.
+        #
+        # @return [Boolean]
+        def grub2_error?
+          bootloader_config.type&.is?(:grub2) && encryption&.method&.is?(:tpm_bls)
```
ancorgs: What happens if bootloader is NONE and the method is tpm_bls? I think it should also fail.

service/lib/agama/storage/config_checkers/encryption.rb
```diff
@@ -84,41 +90,77 @@ def unavailable_method_issue
...
-              _("'%{crypt_method}' is not a suitable method to encrypt the device."),
-              crypt_method: method.to_human_string
-            ),
-            kind: IssueClasses::Config::WRONG_ENCRYPTION_METHOD
-          )
+        # Whether the bootloader is BLS-compliant and the encryption method cannot be used.
+        #
+        # @return [Boolean]
+        def bls_error?
+          bootloader_config.type&.bls? && encryption&.method&.is?(:tpm_fde)
```
ancorgs: What happens if bootloader is NONE and the method is tpm_fde? I think it should also fail.

service/lib/agama/storage/config_checkers/encryption.rb
```diff
@@ -84,41 +90,77 @@ def unavailable_method_issue
...
+          method = encryption&.method
+          return unless method&.only_for_swap?
+          return if config.filesystem&.path == Y2Storage::MountPoint::SWAP_PATH.to_s
+
+          error(
+            format(
+              # TRANSLATORS: %{crypt_method} is the identifier of the method to encrypt the device
+              #   (e.g., "luks1", "random swap").
+              _("%{crypt_method} is not suitable for a device different to swap."),
+              crypt_method: method.to_human_string
```
ancorgs, line 113: I fear this errors are going to be confusing. Because we are not exposing methods like TPM_FDE or TPM_BLS in the config (they are activated using a boolean called 'tpm') but we report them as part of the error message.

Maybe not something to fix now, but for sure to consider in the future.

joseivanlopez:
### TPM-Based Encryption Methods

| Encryption Method | GRUB2 | GRUB2_BLS | SYSTEMD_BOOT | BLS_LEGACY | NONE | nil |
|-------------------|-------|-----------|--------------|------------|------|-----|
| **TPM FDE**       | ✓ Pass | ✗ Error | ✗ Error | ✗ Error | ✗ Error | ✓ Pass |
| **TPM BLS**       | ✗ Error | ✓ Pass | ✓ Pass | ✓ Pass | ✗ Error | ✓ Pass |

### Other Encryption Methods

| Encryption Method | Notes |
|-------------------|-------|
| LUKS1 | Compatible with all bootloaders |
| LUKS2  | Compatible with all bootloaders |
| Pervasive LUKS2 | Availability check only |
| Protected Swap | Only valid for swap devices |

ancorgs (approved):
LGTM

PR #3458, opened 2026-05-04
feat(web): reduce visual noise of input instructions

dgdavid:
`ArrayField`'s Instructions now only appear when field is dirty (has content), making empty fields cleaner and less overwhelming. Also repositioned before helper text for better scanning, and updated keyboard hints to use consistent punctuation and more logical ordering.

### Before

https://github.com/user-attachments/assets/7793c071-035f-467e-9867-4c1bf801b49a

### After

https://github.com/user-attachments/assets/cb2791f8-1e1b-4551-89f1-d46ab0e5d71a

### Notes

### DIFF
--- a/web/src/components/form/ArrayField.test.tsx
+++ b/web/src/components/form/ArrayField.test.tsx
@@ -77,10 +77,36 @@ function TestForm({
 }
 
 describe("ArrayField", () => {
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM

PR #3459, opened 2026-05-04
fix(web): decouple field width from form-group wrapper

dgdavid:
Adjust CSS to override default PatternFly behavior so that input/field width is independent of the form-group container, preventing fields from growing to match the width of their parent because of the width of any of its siblings.

### Before

Screenshot_2026-05-04_22-14-03

### After

Screenshot_2026-05-04_22-13-22

### Notes


teclator (approved):
LGTM

PR #3460, opened 2026-05-04
refactor(web): simplify Bridge/STP helper texts for clarity

dgdavid:
Shortens helper texts for bridge STP fields to make them more user-friendly and easier to scan. Removes technical jargon and verbose explanations in favor of concise descriptions.

Priority field now emphasizes the counter-intuitive "lower is higher" concept within a single string.

> [!WARNING]
> Examples were removed from these helper texts because the fields already use them as default input values. In other words, live examples are present, so including them in the helper text would be redundant.

### Comparison table

| Field | Before | After |
|-------|--------|-------|
| **Priority** | Determines the root bridge. Lower values have higher priority. Range: 0 - 61440. E.g., 32768. | Root bridge selection (0-61440). **Lower is higher priority.** |
| **Forward delay** | Time spent in listening and learning states. Range: 4 - 30 seconds. E.g., 15. | Listening and learning time (4-30 seconds). |
| **Hello time** | Interval between sending BPDU packets. Range: 1 - 10 seconds. E.g., 2. | Protocol message interval (1-10 seconds). |
| **Max message age** | Time to store BPDU info before discarding it. Range: 6 - 40 seconds. E.g., 20. | Protocol message retention time (6-40 seconds). |

#### Before

Screenshot_2026-05-04_22-44-33

#### After

Screenshot_2026-05-04_22-44-12

### Notes

### DIFF
--- a/web/src/components/network/BridgeSettings.tsx
+++ b/web/src/components/network/BridgeSettings.tsx
@@ -23,6 +23,7 @@
 import React from "react";
 import { sprintf } from "sprintf-js";
 import Interpolate from "~/components/core/Interpolate";
+import NestedContent from "~/components/core/NestedContent";
 import { connectionFormOptions } from "~/components/network/ConnectionForm";
 import { withForm } from "~/hooks/form";
 import { useDevices } from "~/hooks/model/system/network";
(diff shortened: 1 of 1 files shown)


teclator (approved):
LGTM

PR #3462, opened 2026-05-05
Don't run obs-staging-debug in fork

jcronenberg:
## Problem

This test always fails in my fork.

## Solution

Disable the test in forks.

## Testing

https://github.com/jcronenberg/agama/actions/runs/25382893364/job/74435802793


lslezak:
I have deleted the file completely. The file was created to debug some GitHub Action issues in the past and it is not needed anymore.

Thank you for your contribution anyway!

PR #3468, opened 2026-05-06
feat(web): enable arrow-key opening of closed dropdowns

dgdavid:
## Summary

Implements the [W3C Select-Only Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) keyboard pattern across all PatternFly dropdown components via a new shared `useComboboxKeyboard` hook. Pressing <kbd>ArrowDown</kbd> / <kbd>ArrowUp</kbd> on a closed toggle now opens the menu and focuses the first or last item without immediately committing a value.

## Motivation

Agama makes extensive use of PatternFly’s `Select` component instead of a native `<select>` to display option descriptions alongside labels, something native selects do not support (until broader support for [`appearance: base-select`](https://developer.chrome.com/blog/a-customizable-select) becomes available and accessible).

However, PatternFly `Select` components require users to press <kbd>Enter</kbd> or <kbd>Space</kbd> before arrow-key navigation becomes available, which differs from native `<select>` behavior. Fully matching native behavior of arrow keys immediately commiting values would introduce accessibility and usability issues in the Agama web interface when selections trigger immediate UI updates.

For example, in the IPv4/IPv6 Settings dropdown within the network connection form, pressing <kbd>ArrowDown</kbd> to explore options would instantly trigger visibility changes in related fields. In other contexts, a selection may immediately apply filtering or refresh visible content. In both cases, the UI can change on every keypress, making the experience disorienting for sighted users and even more challenging for screen reader users, whose announcement of an option may be cut off mid-speech by the resulting UI change.

The W3C explore-then-commit pattern solves both issues: arrow keys open the menu and allow free navigation, while the value only changes on <kbd>Enter</kbd>, <kbd>Space</kbd>, or click. This allows all users to understand their options before committing.

### Before

https://github.com/user-attachments/assets/63c11fa7-9e7e-43e3-acb3-b5cafa09f2dc

### After

https://github.com/user-attachments/assets/4f7ad0d5-1c85-49e5-b3dd-d670400c8f47

## Changes

**New hook: `src/hooks/use-combobox-keyboard.ts`**

Provides W3C-aligned keyboard navigation for PatternFly Select and Menu-based components:

- <kbd>Arrow Down</kbd> on a closed toggle, opens menu, focuses first item
- <kbd>ArrowUp</kbd> on a closed toggle, opens menu, focuses last item
- When menu opens, focus is moved in a `setTimeout(fn, 0)` to let Popper finish positioning first (mirrors PatternFly's own `shouldFocusFirstItemOnOpen` pattern)
- When menu closes, focus is restored to the toggle button
- Supports both internal state management and external state via `isOpen`/`setIsOpen` options
- Supports both `"select"` and `"menu"` component types (different item selectors)
- `getToggleRef` merges PatternFly's toggle ref with the hook's internal focus-restoration ref

**Components updated:**

- `DropdownField`
- `SelectWrapper`
  :warning: This component will be deprecated/drop once all forms migrate to TanStack Form
- `SimpleSelector`
- `MenuButton`
  :memo:  Adds `disableArrowKeyOpen` prop for cases where this behavior should be suppressed

## Keyboard behavior after this change

| Action | Before | After |
|---|---|---|
| Tab to toggle, press <kbd>Arrow Down</kbd> | Nothing (menu stays closed) | Menu opens, first item focused |
| Tab to toggle, press <kbd>Arrow Up</kbd> | Nothing (menu stays closed) | Menu opens, last item focused |
| Arrow keys within open menu | Works (PF default) | Works (unchanged) |
| Navigate options without selecting | Not possible without committing | Supported. no value change until <kbd>Enter</kbd> / <kbd>Space</kbd> |
| Escape from open menu | Works, original value preserved  | Works, original value preserved |


joseivanlopez (approved):
LGTM

PR #3469, opened 2026-05-07
Gnome kiosk wrapper

mchf:
## Problem

--enable-vt-switch is mandatory for newer gnome kiosk instances (>= 50). However older versions crashes when started with this option (because of unknown option)

- [*Bugzilla link*](https://bugzilla.suse.com/show_bug.cgi?id=1264099)

## Solution

Start gnome kiosk with --enable-vt-switch option only if current kiosk instance supports it

## Testing

- *Tested manually*

### DIFF
--- a/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
+++ b/live/live-root/root/.config/systemd/user/org.gnome.Kiosk@wayland.service.d/override.conf
@@ -3,6 +3,4 @@
 # exists in the original service file. If not systemd will try
 # to start every ExecStart (original even the one added bellow)
 ExecStart=
-# gnome-kiosk >= 50 blocks switching to virtual console by default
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

lslezak:
Shellcheck complains:

```
In live/live-root/root/.local/bin/gnome-kiosk-wrapper line 12:
if [[ $(gnome-kiosk --help 2> /dev/null | grep 'enable-vt-switch') ]]; then
   ^-- SC3010 (warning): In POSIX sh, [[ ]] is undefined.
      ^-- SC2143 (style): Use grep -q instead of comparing output with [ -n .. ].

For more information:
  https://www.shellcheck.net/wiki/SC3010 -- In POSIX sh, [[ ]] is undefined.
  https://www.shellcheck.net/wiki/SC2143 -- Use grep -q instead of comparing ...
```

lslezak (approved):

PR #3473, opened 2026-05-07
Import the ntp-client section from AutoYaST profiles

imobachgs:
Follow-up of #3446.

Read the `ntp-client` section from AutoYaST profiles.

## Example

``` xml
  <ntp-client>
    <ntp_policy>auto</ntp_policy>
    <ntp_servers config:type="list">
      <ntp_server>
        <address>1.opensuse.pool.ntp.org</address>
        <iburst config:type="boolean">true</iburst>
        <offline config:type="boolean">false</offline>
      </ntp_server>
    </ntp_servers>
  </ntp-client>
</profile>
```

The excerpt above is converted to:

``` json
{
  "ntp": {
    "sources": [
      { "type": "pool", "address": "1.opensuse.pool.ntp.org", "iburst": true, "offline": false }
    ]
  }
}
```

## Testing

- Added a new unit test
- Tested manually

## Follow-up

- Update the documentation about AutoYaST compatibility.

### DIFF
--- a/service/lib/agama/autoyast/converter.rb
+++ b/service/lib/agama/autoyast/converter.rb
@@ -26,6 +26,7 @@ require "agama/autoyast/hostname_reader"
 require "agama/autoyast/iscsi_reader"
 require "agama/autoyast/localization_reader"
 require "agama/autoyast/network_reader"
+require "agama/autoyast/ntp_client_reader"
 require "agama/autoyast/product_reader"
 require "agama/autoyast/root_reader"
 require "agama/autoyast/scripts_reader"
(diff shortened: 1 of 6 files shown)


service/lib/agama/autoyast/converter.rb
```diff
@@ -64,7 +65,8 @@ def to_agama(profile)
           SoftwareReader.new(profile).read,
           StorageReader.new(profile).read,
           UserReader.new(profile).read,
-          ZFCPReader.new(profile).read
+          ZFCPReader.new(profile).read,
+          NtpClientReader.new(profile).read
```
jreidinger: previously it looks like alphabetically sorted and this new entry breaks pattern

jreidinger (approved):

PR #3475, opened 2026-05-07
refactor(web): simplify UX of Hostname Page

dgdavid:
## Problem

The previous implementation used a "Use static hostname" checkbox combined with an alert to convey the current hostname state. This was indirect and visually noisy. Users had to infer the concept of "transient" vs "static" from a checkbox label and a separate alert.

## Solution

Simplifies the hostname configuration UX by replacing the Alert + checkbox driven approach with a cleaner mode selector backed by TanStack Form.

## Key Changes

- Replaces the checkbox with a **Mode** dropdown offering **Transient**
  (provided by the network) and **Static** (set manually) options
- Each mode now has dedicated contextual information shown inline:
  - Transient: displays the current hostname and explains it may change
  - Static: shows a text input with a helper explaining it persists across reboots
- Improves alert copy for clarity (registration warning, error messages)
- Upgrades request error alert from `warning` to `danger` variant

### Options closed

localhost_8080_ (68)

### Options open

Screenshot_2026-05-07_19-12-38

### Before vs After

|Before | After |
|-|-|
|localhost_8080_ (63) | localhost_8080_ (57) | 
| localhost_8080_ (61) |localhost_8080_ (59) | 
| localhost_8080_ (62) | localhost_8080_ (60) |
| localhost_8080_ (64) |  localhost_8080_ (66) |


web/src/components/system/HostnamePage.tsx
```diff
@@ -20,112 +20,241 @@
...
+ * Displays helper text explaining that a static hostname will persist across
+ * reboots and network changes.
+ */
+function StaticModeInfo() {
+  return (
+    <Interpolate
+      sentence={
+        // TRANSLATORS: helper text for static hostname input.
+        // Text in square brackets will be displayed in bold.
+        _("Once set, will [persist across reboots and network changes].")
```
dgdavid, line 89: Need better text here. "Once set," feels weird when it is already using static value. We could show "Once set" when current is transient and omit it if using static. That's perfectly possible, and contextual. For the latest, maybe a "Persisted across reboots..." might work.

WDYT?

web/package/agama-web-ui.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri May  8 11:01:22 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Simplify hostname configuration UX by replacing the checkbox-driven
+  approach with a mode selector (gh#agama-project/agama#3475).
+-------------------------------------------------------------------
+
```
dgdavid:
```suggestion

-------------------------------------------------------------------
```

web/src/components/system/HostnamePage.tsx
```diff
@@ -20,112 +20,241 @@
...
-import { patchConfig } from "~/api";
+import { useAppForm } from "~/hooks/form";
+import { _ } from "~/i18n";
+
+import spacingStyles from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+
+const HOSTNAME_MODE = {
+  TRANSIENT: "transient",
+  STATIC: "static",
+} as const;
```
joseivanlopez, line 41: NP: why `as const`?

web/src/components/system/HostnamePage.tsx
```diff
@@ -20,112 +20,241 @@
...
+          {() => <Text isBold>{transientHostname}</Text>}
+        </Interpolate>
+      </Content>
+      <Content component="small">
+        <Interpolate
+          sentence={
+            // TRANSLATORS: explanation of transient hostname behavior.
+            // Text in square brackets will be displayed in bold.
+            _(
+              "This hostname is dynamic and [may change after a reboot or network update], as configured by the local network administrator.",
```
joseivanlopez, line 69: Maybe we can include the current name in this explanation: 

"Hostname jungle is currently assigned, but it may change after a reboot or network update."

Even the last part ("as configured by the local network administrator") sounds too technical to me.

dgdavid: If'I remember correctly, that last part was something requested by the team last time this page was reworked and reviewed.

It also sounds too technical to me

dgdavid: About including the current value in the explanation, some things:

* I still seeing no need to repeat "Hostname"
* It make interface lose the parity between two modes
* It breaks something was already implemented in network forms: how not editable values are shown

I would go for changing the label and making it appear as the static version

```
Mode
  [ Transient v]
  
Value
  jungle

  Explanation
```

There is a "ReadonlyField" component to do that. The explanation should be the "aria-described-by" part of the field.

dgdavid: If you don't like "Value" I'm open to other labels, but let's keep the presentation consistent.

"Name" or even "Hostname" if there is too much against to "Value", which is the correct label to me.

web/src/components/system/HostnamePage.tsx
```diff
@@ -20,112 +20,241 @@
...
+                <field.DropdownField
+                  // TRANSLATORS: label for hostname mode selector
+                  label={_("Mode")}
+                  options={[
+                    {
+                      value: HOSTNAME_MODE.TRANSIENT,
+                      // TRANSLATORS: hostname mode option
+                      label: _("Transient"),
+                      // TRANSLATORS: description for transient hostname mode
+                      description: _("Provided by the network"),
```
joseivanlopez, line 217: What about something like "Dynamic hostname". The explation once the option is selected already metions the technical detail about when it changes.

web/src/components/system/HostnamePage.tsx
```diff
@@ -20,112 +20,241 @@
...
+                    },
+                    {
+                      value: HOSTNAME_MODE.STATIC,
+                      // TRANSLATORS: hostname mode option
+                      label: _("Static"),
+                      // TRANSLATORS: description for static hostname mode
+                      description: _("Set manually"),
```
joseivanlopez, line 224: And here maybe "Fixed hostname".

web/src/components/system/HostnamePage.tsx
joseivanlopez: Yes, I would prefer something like "Hostname persisted across reboots."

web/src/components/system/HostnamePage.tsx
joseivanlopez: > If'I remember correctly, that last part was something requested by the team last time this page was reworked and reviewed.
> 
> It also sounds too technical to me

I am not an expert in the area, and as so, it gives no value to me. But if it was agreed to include it ... I can live with that.

joseivanlopez: > If you don't like "Value" I'm open to other labels, but let's keep the presentation consistent.
> 
> "Name" or even "Hostname" if there is too much against to "Value", which is the correct label to me.

Ok, I did not realize it was a common pattern adopted in the UI. Then, I would change the label. Some options that sound good to me: "Assigned hostname", "Hostname".

joseivanlopez: If you want to keep the parity, then I would use "Hostname" for both.

web/src/components/system/HostnamePage.tsx
dgdavid: Without repeating "Hostname". I'll do something, most probably dynamically as mentioned

web/src/components/system/HostnamePage.tsx
dgdavid: I'm gonna to insist last time:

Hostname > Mode > Hostname 


Looks too redundant IMHO.

But if there is no better option, I can live with it, without understanding what is wrong with "Value"

web/src/components/system/HostnamePage.tsx
dgdavid: I think It's important to know where the value comes before choosing it. But it is also important to telmuser it will be have a dinnamyc behaviour.

No strong opinion here.

web/src/components/system/HostnamePage.tsx
dgdavid: Fixed sound too technical, doesn't it? 

Set manually sounds more natural

joseivanlopez: I would use fixed if we decide to use dynamic for transient.

web/src/components/system/HostnamePage.tsx
joseivanlopez: I think Value is not a good name because I have selected a Mode and then I am requested for a value. Value of what? I think it is more understandable to use the label of what the field is requesting. But like everything in the UI, I am not an expert and most likely I am totally wrong. I only can say what I feel like a user. I haven't any other argument. Feel free to keep Value if you think it is better. I trust your criteria more ;)

web/src/components/system/HostnamePage.tsx
dgdavid: Then, I'd keep these descriptions as they are. Dynamic/sñfixed behaviour are properly explained in the helper texts

web/src/components/system/HostnamePage.tsx
dgdavid: > Value of what?

Of hostname.

You didn't asked "Mode of what?".

This kind of questions really surprise me

dgdavid: > > Value of what?
> 
> Of hostname.
> 
> You didn't asked "Mode of what?".
> 
> This kind of questions really surprise me

I mean.

We tend to overthink users are completely out of context, which is not the case here.

dgdavid: I mean (again)

It is no's a field labelled "Value" in  a "System Form" with plenty of mixed things. It is a "Value" field in a dedicated "Hostname" form. It makes perfect sense.

All that explained, I'm not against changing it, but don't understand these arguments against such a label because it's perfectly contextualized.

web/src/components/system/HostnamePage.tsx
dgdavid: Sorry, I overlooked this before.

It is a TypeScript assertion that makes all properties `readonly`, turns the object into a kind of strongly-typed enum, infers the exact literal values (`"transient"`, `"static"`) instead of broadening them to `string`, and catches any accidental mutation at compile time.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

web/src/components/system/HostnamePage.tsx
joseivanlopez: I understand we are in the context of configuring hostname, but after selecting a mode (which we can argue if it is also surprising) I have to enter a value. I am simply not sure what value means. Something related to the mode?. Of course, this case is simple and probably we can conclude we have to enter the hostname, but I don't understand why isn't better to make it quite clear.

I think we have already invested quite some time on discussing this little detail. As said before, feel free to keep Value.

joseivanlopez (approved):

joseivanlopez (changes requested):
Sorry, I approved this PR because you have already replied you prefer the current texts. But I forgot you are going to improve the text for static case.

web/src/components/system/HostnamePage.tsx
dgdavid: > (which we can argue if it is also surprising)

We do not have the same view. 

It is surprising when making a comparison with former way. Maybe 1 second of surprise, until users start interacting with the UI (which has exactly one or two control, low room for getting lost). Once users interact with the UI, they are provided with all the information they need to understand what the feature is about. 

And, from my point of view, it is way more understandable than the current version.

>  I have to enter a value. 

Which is exactly what "Set manually" description of the selected option already hint users. I sill not seeing the surprise, sorry.

> Of course, this case is simple and probably we can conclude we have to enter the hostname, but I don't understand why isn't better to make it quite clear.

We have exactly the very same question here. Because I do not understand what is the problem of asking for "Value for the Hostname Static mode". Of course, the "Hostname for the Static Hostname mode"  is not strictly wrong, what I was saying is that I do not see the need of being redundant and losing the connection between _value_ and _mode_, since we bet for adding these stuff at UI level somehow back in the time when we decided to expose the transient thing here. Moreover, I was exactly saying that I do not buy the argument of asking your self "Value for what" without previously asking yourself, in the role of same user, "Mode for what". To me, that reveals the natural and expected resistense to a change.

> I think we have already invested quite some time on discussing this little detail. As said before, feel free to keep Value.

I agree, we have invested much more time that the time invested in the PR itself. But probably for shaping something better.

We can put Hostname as well as Value. To make it defeats a bit the _natural reading_ of the UI, and incur in not needed redundnacey. We can also use "Hostname mode" for the mode.

```
Hostname

Mode
[ Transient   v]

Hostname: 
localhost
Lorem ipsum dolor
```

and 

```
Hostname

Mode
[ Static   v]

Hostname: 
[ localhost ]
Lorem ipsum dolor
```


vs

```
Hostname

Mode
[ Transient   v]

Value: 
localhost
Lorem ipsum dolor
```

and 

```
Hostname

Mode
[ Static   v]

Value: 
[ localhost ]
Lorem ipsum dolor
```

dgdavid: There is also a plan to sketch a SystemPage i nwhich Hostname would be a single fieldset

```
- Hostname -

  Mode              Value
  [ Static  v]      [ localhost ]

  Lorem ipsum here
```

vs

```
- Hostname -

  Mode              Hostname
  [ Static  v]      [ localhost ]

  Lorem ipsum here
```

(no inputs without label, that's why it cant be just a single _Hostname_ label and one or two fields, depending on the mode.

dgdavid: > There is also a plan to sketch a SystemPage i nwhich Hostname would be a single fieldset 

If this plan does not succeed, I will put "Hostname" as label in the current proposal.

```
Hostname

Mode
[ Static   v]

Hostname: 
[ localhost ]
Lorem ipsum dolor
```

If it succeed, I will keep "Value"
```
- Hostname -

  Mode              Value
  [ Static  v]      [ localhost ]

  Lorem ipsum here
```

imobachgs:
> Sorry, I approved this PR because you have already replied you prefer the current texts. But I forgot you are going to improve the text for static case.

I have the impression that we are discussing this wording to the extreme. Please, let's try to move forward ;-)

joseivanlopez:
> I have the impression that we are discussing this wording to the extreme. Please, let's try to move forward ;-)

I already approved it and I have nothing else to add. But @dgdavid commented he is going to propose a new sentence for the static explanation, right? If not, I can approve again.

dgdavid:
> > I have the impression that we are discussing this wording to the extreme. Please, let's try to move forward ;-)
> 
> I already approved it and I have nothing else to add. But @dgdavid commented he is going to propose a new sentence for the static explanation, right? If not, I can approve again.

That is right. But as commented in the daily team call, yesterday we were working in a new PR that will obsolete this (https://github.com/agama-project/agama/pull/3482).  In any case, all feedback provided here will be used in the new one, thanks a lot.

dgdavid:
Superseded by #3482

PR #3482, opened 2026-05-11
feat(web):  adds System section and page

dgdavid:
Built on top of #3475, this set of changes transforms the current `Hostname` section/page into a more generic `System` one. The goal is to allow users to configure not only the hostname, but also other small system-related settings without promoting them to top-level areas, pages, or sections.

We explored a similar idea in the past, but the previous UI implementation did not make it practical. The concept has now resurfaced in the context of adding NTP support to the UI (https://trello.com/c/H5oDLoqj/ protected link), and after several discussions around an initial draft, it was decided to move forward with it.

> [!IMPORTANT]
>
> **`SystemPage` should only contain settings that affect both the currently running installation system and the final installed system**. At the moment, this applies to hostname and NTP configuration.

## Screenshots

> [!NOTE]
> These are just a few screenshot, there are more variants of these pages. Jump into the branch and play with it to see all of them.

### Summary

system_page_summary

system_page_summary_2

### Page/Form

system_page_custom_ntp

system_page_defaut_ntp

---

**Note**: this superseeds https://github.com/agama-project/agama/pull/3475


dgdavid:
> ## Second commit ([7360239](https://github.com/agama-project/agama/commit/736023983ada8894bcc74539924c6583a759a63d)) Hostname + Language and region + NTP
> localhost_8080_ (69)


As per conversation witn @imobachgs, _Language and Region_ does not belong to that new System Page since these are settings only for the installed system, contrary to Hostname and NTP that applies both: the system running the installation and the installed system.

@ancorgs also realized about chances of transient hostname being empty, which looks like not possible but I haven't had time to investigate if it is about something wrong in the web user interface codebase or something that comes empty from backend, nor why.

dgdavid:
> @ancorgs also realized about chances of transient hostname being empty, which looks like not possible but I haven't had time to investigate if it is about something wrong in the web user interface codebase or something that comes empty from backend, nor why.

After trying to have better understanding of hotname API (check https://github.com/agama-project/agama/pull/3482/commits/16e5a6b3d825009c10e94fb357f112eeb5a64d26) I have tried to minimize this problem by using stricter checking of static is not set to show the transient instead (https://github.com/agama-project/agama/pull/3482/commits/f79d43ec15611ed946d2c10bd8fb37f366a8a2e5)

But still not knowing if, as written in the documentation, `hostname` (aka transient)  will always return something,  as expected.

@teclator, friendly ping :point_up: :point_up: :point_up: :point_up:

teclator:
@dgdavid If not specific transient or static hostname is set, the transient hostname should be localhost IIRC...

imobachgs (approved):

web/src/components/overview/SystemSummary.test.tsx
```diff
@@ -0,0 +1,282 @@
+/*
+ * Copyright (c) [2025-2026] SUSE LLC
```
imobachgs:
```suggestion
 * Copyright (c) [2026] SUSE LLC
```

❓

web/src/components/overview/SystemSummary.tsx
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (c) [2025-2026] SUSE LLC
```
imobachgs: ❓ 
```suggestion
 * Copyright (c) [2026] SUSE LLC
```

PR #3484, opened 2026-05-12
Retry pkg install

jreidinger:
## Problem

When installation failed to download some packages e.g. due to unstable network, there is no option to retry it.

## Solution

Add popup with default to not retry. It keeps previously downloaded packages and just try to resume with the rest.

## Testing

- *Tested manually*


rust/agama-software/src/zypp_server.rs
```diff
@@ -273,26 +274,48 @@ impl ZyppServer {
...
+            )));
+            result = zypp.commit(
+                &mut download_callback,
+                &mut install_callback,
+                &mut security_callback,
+            )?;
+            tracing::info!("libzypp commit ends with {}", result);
+            if result {
+                break;
+            } else {
```
imobachgs: If you are "breaking" in the previous line, I guess you do not need to use an `else`.

```rust
if result {
  break;
}

// the rest goes here.
```

imobachgs (approved):
Just a minor comment. Otherwise, it looks good.

imobachgs (approved):

PR #3485, opened 2026-05-12
Use the right strategy and subvolumes in the storage proposal

ancorgs:
## Problem

Agama does not offer good support for installing distributions using a BLS-compliant bootloader. Some of the changes introduced during the last year in YaST made it very difficult to do so in a sane way.

Fortunately, https://github.com/yast/yast-storage-ng/pull/1425 fixes the mess in the yast-storage-ng side, opening the door for a proper implementation in Agama.

## Solution

Rely on the new `Y2Storage` classes and mechanisms to allow a configurable bootloader in Agama, resulting in a reasonable proposal of partitions and subvolumes.

## Notes

This PR does not include a changelog entry because it is open against a feature branch.


service/lib/agama/storage/config_solvers/filesystem.rb
```diff
@@ -79,7 +84,11 @@ def default_filesystem(path = nil)
         # @param path [String, nil]
         # @return [Configs::Btrfs]
         def default_btrfs(path = nil)
-          default_filesystem(path).type.btrfs
+          btrfs = default_filesystem(path).type.btrfs
+          if path == "/" && btrfs.subvolumes
```
joseivanlopez, line 88: Are subvols added only if the default config also define subvols?

ancorgs: Good point. Currently the default config of all products always adds subvols, but we should certainly cover the empty case as well.

ancorgs: I added an extra unit test to cover all cases. Turns out everything worked so the code was correct. But now we are sure. :-)

joseivanlopez (approved):
LGTM

PR #3486, opened 2026-05-12
reset config when switching product (bsc#1264438)

jreidinger:
## Problem

When changing product old partitioning is kept even when it does not make sense with new product.

- bsc https://bugzilla.suse.com/show_bug.cgi?id=1264438

## Solution

Reset whole user config and just set product.

## Testing

- *Tested manually* - https://build.opensuse.org/project/show/systemsmanagement:Agama:branches:reset_storage


dgdavid (commented):
LGTM having in mind that restarting the configuration should be not something unexpected (changing the product selection is almos like starting from the beginning) and therefore no confirmation is required by now for the shake of fluent workflow.

Thanks

web/src/components/product/ProductSelectionPage.tsx
```diff
@@ -704,7 +702,8 @@ const ProductSelectionContent = () => {
   const onSubmit = async (selectedProduct: Product, selectedMode: string) => {
     setIsSubmmited(true);
     setSubmmitedSelection(selectedProduct);
-    putConfig({ ...config, product: { id: selectedProduct.id, mode: selectedMode } });
+    // Put only config there to reset reset of config as it can be affected by product (bsc#1264438)
```
joseivanlopez:
```suggestion
    // Put product and mode only in order to reset the rest of the config, which can depend on the selected product and mode (bsc#1264438)
```

web/src/components/product/ProductSelectionPage.tsx
```diff
@@ -704,7 +702,8 @@ const ProductSelectionContent = () => {
   const onSubmit = async (selectedProduct: Product, selectedMode: string) => {
     setIsSubmmited(true);
     setSubmmitedSelection(selectedProduct);
-    putConfig({ ...config, product: { id: selectedProduct.id, mode: selectedMode } });
+    // Put product and mode only in order to reset the rest of the config, which can depend on the selected product and mode (bsc#1264438)
```
joseivanlopez, line 705: This line is too long, but the linter is not complaining, so I am going to approve.

joseivanlopez (approved):
LGTM

PR #3487, opened 2026-05-12
fix(web): refetch storage model data

joseivanlopez:
## Problem

* When navigating from `ZFCPPage` to `ProposalPage`, the `STORAGE_MODEL_KEY` query was being invalidated but not refetched immediately (since `ZFCPPage` didn't use `useConfigModel()`)
* When arriving at `ProposalPage`, the query would refetch quickly
* But then `useProgressTracking()` would wait for queries to refetch AFTER the progress finished, missing the refetch that already happened.
* https://bugzilla.suse.com/show_bug.cgi?id=1263808

## Solution:

* Added `useConfigModel()` to *App.tsx* `Content` component
* This ensures the query is always actively observed and refetches immediately when invalidated, matching the existing pattern for `useProposal()` and `useIssues()`

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Tue May 12 13:35:54 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Fix query refetch when activating zFCP devices (bsc#1263808).
+
+-------------------------------------------------------------------
 Tue May 12 10:29:09 UTC 2026 - David Díaz <dgonzalez@suse.com>
 
 - Improve keyboard accessibility for dropdown components by enabling
@@ -23,7 +28,7 @@ Thu Apr 30 13:01:05 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
 
 - Added support for configuring a network bridge using the UI
   (gh#agama-project/agama#3435).
-  
+
 -------------------------------------------------------------------
 Tue Apr 28 15:43:35 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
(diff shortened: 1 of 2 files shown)


joseivanlopez:
Claude conversation:

> I have a problem with the tanstak queries in the ZFCPPage component defined at src/components/storage/zfcp/ZFCPPage.tsx. The query with key STORAGE_MODEL_KEY is correctly invalidated thanks to the call to useProposalChanges() in App component, see web/src/App.tsx. But the data of the query is not refetched once I go to proposal page, see web/src/components/storage/ProposalPage.tsx. Could you find the problem?

I found the issue! It's a race condition between query refetching and progress tracking. Here's what's happening:

The Problem

1. When on ZFCPPage, a ProposalChanged event invalidates STORAGE_MODEL_KEY
2. Since ZFCPPage doesn't use useConfigModel(), the query is marked as stale but not refetched yet
3. When you navigate to ProposalPage:
  - useConfigModel() mounts and immediately refetches the stale query
  - The refetch completes quickly
4. Later, when backend progress finishes, useProgressTracking() calls startTracking()
5. startTracking() waits for queries to be refetched AFTER it starts monitoring
6. But STORAGE_MODEL_KEY was already refetched in step 3, so it waits forever

This is exactly the issue mentioned in the FIXME comment at App.tsx:46-47!


The Solution

Follow the same pattern used for useProposal() and useIssues() - call useConfigModel() in the Content component so it's always mounted and refetches immediately when invalidated:

Why This Fixes the Issue

By calling useConfigModel() in a component that's always mounted:

1. Immediate refetch on invalidation: When useProposalChanges() invalidates STORAGE_MODEL_KEY, the query will be refetched immediately (since there's an active observer in the Content
component)
2. Correct timing for progress tracking: When you navigate to ProposalPage and the progress finishes, useProgressTracking() will start monitoring for refetches that happen AFTER that
point. Since the query was already refetched when invalidated (step 1), if there's a subsequent invalidation, it will properly wait for the new refetch.
3. Consistent with existing pattern: This matches the FIXME comment's approach for useProposal() and useIssues()

Now the query will be properly refetched both when you're on ZFCPPage (immediately when invalidated) and when you navigate to ProposalPage (the already-fresh data will be available, or it will properly wait for any ongoing refetch).

dgdavid (approved):
LGTM

It makes sense to follow the patter we already started with the proposal. Although would be nice to find some improvements in the future for these estufr when time permits.

Thanks

PR #3488, opened 2026-05-12
Read NTP configuration from kernel's command-line

imobachgs:
This PR introduces support for reading setting the NTP configuration through kernel command-line option.

1. Adds the new [chrony dracut module](https://github.com/dracut-ng/dracut-ng/pull/2397).
2. Reads the time sources from dracut (located in `/run/chrony/dracut.sources.d/dracut.sources`). They will be included as part of the "extended configuration" (just like we do with the `inst.register_url` argument).

## Testing

- [x] Added a new unit test
- [x] Tested manually


rust/agama-ntp/src/model/chrony.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    async fn install(&self, config: &Config) -> Result<(), Error> {
+        let path = self.install_config_path();
+
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent).map_err(Error::WriteConfig)?;
+        }
+
+        // FIXME: copying the configuration would be enough.
+        let content = generate_chrony_config(config.sources.as_deref().unwrap_or(&[]));
```
jreidinger: NP: default value for None sources is specified multiple times, so it will make change in future harder.

rust/agama-ntp/src/model/chrony.rs
```diff
@@ -0,0 +1,568 @@
...
+        if fs::exists(&path).map_err(Error::RemoveConfig)? {
+            fs::remove_file(&path).map_err(Error::RemoveConfig)?;
+        }
+
+        Ok(())
+    }
+}
+
+fn generate_chrony_config(sources: &[Source]) -> String {
+    let mut lines = vec!["# Generated by Agama".to_string()];
```
jreidinger, line 175: Do we want to specify agama here or just generic `Installer`?

rust/agama-ntp/src/service.rs
```diff
@@ -112,41 +116,57 @@ impl MessageHandler<message::GetConfig> for Service {
         &mut self,
         _message: message::GetConfig,
     ) -> Result<Option<api::ntp::Config>, Error> {
-        Ok(self.config.clone())
+        Ok(Some(self.config.clone()))
```
jreidinger, line 119: so it never return None? So why it is option?

rust/agama-ntp/src/model/chrony.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        sources.push(Source {
+            source_type,
+            address,
+            iburst,
+            offline,
+        });
+    }
+
+    let sources = if sources.is_empty() {
```
jreidinger, line 260: this is quite counter intuitive...if there is config and it has not sources, why returning None instead of empty list?

rust/agama-ntp/src/model/chrony.rs
imobachgs: You are right. Actually, the generate_chrony_config should receive the whole `Config`.

rust/agama-ntp/src/model/chrony.rs
imobachgs: That's a good question. I am not sure.

rust/agama-ntp/src/service.rs
imobachgs: Because it is the standard for all the services. The `Config::ntp` field is an Option.

rust/agama-ntp/src/model/chrony.rs
imobachgs: The Config::sources is an `Option` because we are following the same approach than `packages`, `extraRepositories`, etc. This function basically list the configuration from chrony, so if it does not find any sources, it just do not return the list of sources.

Perhaps we might consider having a different struct.

jreidinger (approved):

PR #3489, opened 2026-05-12
fix(web): prevent focus loss when tabbing out of SuggestionsTextField

dgdavid:
During post-merge manual testing of #3445, @ancorgs noticed a focus-loss issue in storage forms. When users type quickly and press TAB to move between fields, focus could be lost, disrupting the expected workflow and undermining the user experience.

From the main commit of this PR

> When typing quickly in the mount point field and pressing TAB to move to the next field, focus could be lost instead of advancing correctly.
> 
> This happens in certain storage forms that perform intensive callbacks and validations. The blur handler triggered an immediate parent update, causing a React re-render while the browser was still processing the focus transition.
> 
> Defer the blur-time onChange until after the blur event completes using setTimeout(..., 0), so the browser can finish moving focus before re-rendering occurs.
> 
> This issue is expected to stop once these forms are migrated to TanStackForm.

---

 **Note on testing**: This fix addresses a timing-dependent race condition that occurs with fast typing and immediate TAB press. It's difficult to reproduce reliably in unit tests since  RTL `user.type()` deliberately adds realistic delays between keystrokes (https://testing-library.com/docs/user-event/options/#advancetimers) that prevent reproducing the race condition. The fix has been manually verified to work.

---
https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue May 12 16:26:43 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Prevent focus loss and reduce flicker in storage forms using
+  SuggestionsTextField by deferring blur updates and increasing
+  debounce to 600ms (gh#agama-project/agama#3489).
+
+-------------------------------------------------------------------
 Tue May 12 10:29:09 UTC 2026 - David Díaz <dgonzalez@suse.com>
 
 - Improve keyboard accessibility for dropdown components by enabling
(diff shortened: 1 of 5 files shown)


dgdavid:
> This issue is expected to stop once these forms are migrated to TanStackForm.

And aligning them with the rest of forms behavior, not triggering fields changes without user intervention (e.g., to apply the defaults proposed by product) and deferring validations until first submission. These make the forms not only fully user-driven, but wide more performant.

dgdavid:
Closed because of https://github.com/agama-project/agama/pull/3490

PR #3491, opened 2026-05-13
[SLE-16.0] Restore the checkmedia option back

lslezak:
Backporting #3474 to SLE 16.0

### DIFF
--- a/live/src/agama-installer.kiwi
+++ b/live/src/agama-installer.kiwi
@@ -37,17 +37,17 @@
     </preferences>
     <!-- the ISO Volume ID is set by the fix_bootconfig script -->
     <preferences arch="ppc64le" profiles="openSUSE,SUSE_SLE_16,Leap_16.0">
-        <type image="iso" flags="dmsquash"  firmware="ofw" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
+        <type image="iso" flags="dmsquash"  firmware="ofw" mediacheck="true" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
             <bootloader name="grub2" console="serial" timeout="10"/>
         </type>
     </preferences>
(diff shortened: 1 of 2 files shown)


live/src/agama-installer.kiwi
```diff
@@ -37,17 +37,17 @@
...
     </preferences>
     <preferences arch="aarch64,x86_64" profiles="openSUSE,SUSE_SLE_16,Leap_16.0">
-        <type image="iso" flags="dmsquash" firmware="uefi" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
+        <type image="iso" flags="dmsquash" firmware="uefi" mediacheck="true" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
             <bootloader name="grub2" timeout="10"/>
         </type>
     </preferences>
     <preferences arch="s390x" profiles="openSUSE,SUSE_SLE_16,Leap_16.0">
-        <type image="iso" flags="dmsquash" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
+        <type image="iso" flags="dmsquash" mediacheck="true" publisher="SUSE LLC" editbootconfig="fix_bootconfig">
```
mchf, line 50: in original discussion regarding bsc#1262229 was discussed that mediacheck is not supported on other architectures than x86_64. So does checkmedia work on s390, ppc64 and other weird creatures now?

lslezak: The SLE-15 ISO images contain checksums also on aarch64, ppc64le and s390x architectures so I guess it should work there also in SLE-16.

And at least I'm able to verify an s390x medium on an x86_64 machine.

mchf: ok i didn't try myself. Only recalled what kiwi maintainer told me when we're discussing the original fix. So I guess that if it works for you currently, than it is good enough. Thanks

mchf (approved):

PR #3492, opened 2026-05-14
Fix dracut systemd unit installation paths

### LINKED ISSUE #3264: dracut: only use variables expanded to /etc in hostonly mode
dracut-109 (currently in Tumbleweed) brings a compat breaking change that went unnoticed because it was not even mentioned in the NEWS, which is causing issues with external dracut modules: https://github.com/dracut-ng/dracut-ng/commit/2991f74ab46389947ce8f96b87a7abdc1862cbb6

See, e.g., why rpm-ostree systems failed to boot in Fedora: https://github.com/ostreedev/ostree/pull/3572

After checking your codebase, I found two modules with the same problem that need to be fixed:

https://github.com/agama-project/agama/blob/0777bee8746bca161152972f61949c53a8da94a3/live/live-root/usr/lib/dracut/modules.d/99initrd-nmtui/module-setup.sh#L27

https://github.com/agama-project/agama/blob/0777bee8746bca161152972f61949c53a8da94a3/live/live-root/usr/lib/dracut/modules.d/99live-self-update/module-setup.sh#L30

I don't know the severity or where these modules are used, but in non-hostonly initrds (usually what installers create) `$systemdsystemconfdir` is not expanded. So please, replace `$systemdsystemconfdir` with `$systemdsystemunitdir` there to avoid problems.

officialasishkumar:
## Problem

dracut 109 only expands `$systemdsystemconfdir` to `/etc/systemd/system` in hostonly initrds. The Agama live image uses non-hostonly initrds, so the `initrd-nmtui` and `live-self-update` modules can miss their service units when they install them through `$systemdsystemconfdir`.

Fixes #3264.

## Solution

Install both service units through `$systemdsystemunitdir`, matching the existing `98dracut-menu` module and dracut's expanded unit-directory variable.

## Testing

- `bash -n live/live-root/usr/lib/dracut/modules.d/99initrd-nmtui/module-setup.sh`
- `bash -n live/live-root/usr/lib/dracut/modules.d/99live-self-update/module-setup.sh`
- `git diff --check`

`make -C live check` could not run in this environment because `/usr/bin/rspec` is not installed for the live test scripts.


jreidinger (approved):
thanks for contribution

PR #3493, opened 2026-05-14
Filter installer-only boot parameters from target cmdline

### LINKED ISSUE #3172: options meant for the installer are used on the installed system
In some occasion I need to perform a remote installation. I reboot the machine and make it start the agama installer automatically, I point my browser to the IP of the machine and then I can proceed with the installation. However, if somebody is in front of the remote machine, he/she can see what I'm doing and possibly interfere. In order to avoid that, I can pass to agama the option: systemd.unit=multi-user.target as described here: https://agama-project.github.io/docs/user/reference/boot_options.
That does the trick, however there is a drawback. The option "systemd.unit=multi-user.target" meant for the installation process is also present in the installed system in the variable GRUB_CMDLINE_LINUX_DEFAULT in the file /etc/default/grub.
The installed system does not reach the graphic target unless I manually remove such option.

This is not the only option that, while it is meant for the installer, it is also present in the installed system.
Others I have found so fare are:
iso-scan/filename=\<file>
brd.rd_nr=1

And talking about multi-user.target/graphic.target agama should offer in the GUI the option to select the final target on the installed system as yast did.

officialasishkumar:
## Problem

`kernel-cmdline-conf.sh` preserves some parameters that are only meaningful while booting the installer. As reported in #3172, options such as `systemd.unit=multi-user.target`, `iso-scan/filename=...`, and `brd.rd_nr=1` can then be copied into the installed system's bootloader configuration.

Fixes #3172.

## Solution

Filter those installer/live parameters when generating `/run/agama/cmdline.d/kernel.conf`, alongside the existing Kiwi, live, Agama, and network boot-argument filters.

## Testing

- `live/live-root/usr/bin/kernel-cmdline-conf.sh live/test/fixtures/source/cmdline /tmp/agama-3172-cmdline-result`
- `cmp -s live/test/fixtures/expected/cmdline /tmp/agama-3172-cmdline-result`
- `bash -n live/live-root/usr/bin/kernel-cmdline-conf.sh`
- `git diff --check`


live/live-root/usr/bin/kernel-cmdline-conf.sh
```diff
@@ -22,7 +22,8 @@ write_kernel_args() {
       _found=1
       ;;
     # remove the Kiwi PXE boot options or Live options
-    rd.kiwi.* | rd.live.* | ramdisk_size=* | initrd=* | BOOT_IMAGE=*)
+    rd.kiwi.* | rd.live.* | ramdisk_size=* | initrd=* | BOOT_IMAGE=* | \
+    iso-scan/filename=* | brd.rd_nr=* | systemd.unit=*)
```
jreidinger, line 26: yeah, that systemd.unit is source of a bit discussion if we should propose it if user wants terminal only installation or if we should try to let it on installation. I think that in the end, filtering it out and keeping it for rest of config to decide is good enough.

jreidinger (approved):
thanks for contribution

Issue #3483, opened 2026-05-12
configuration option extraRepositories not working anymore


gcomes:
Adding repositories using extraRepositories in the configuration file is not working anymore (it was working just one month ago, see issue 3370). Agama seems to read the content of the repositories specified in extraRepositories, but it fails to add such repositories in zypp.
The command 'zypper lr -u' only shows the builtin repositories.
See the attached log.

[agama-logs.tar.gz](https://github.com/user-attachments/files/27620291/agama-logs.tar.gz)

imobachgs:
According to the logs, I can see that the repositories are added and copied to the installed system:

```
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/fixrpm-16.9.repo to /mnt/etc/zypp/repos.d/fixrpm-16.9.repo
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/non-oss.repo to /mnt/etc/zypp/repos.d/non-oss.repo
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/oss.repo to /mnt/etc/zypp/repos.d/oss.repo
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/agama-2.repo to /mnt/etc/zypp/repos.d/agama-2.repo
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/agama-1.repo to /mnt/etc/zypp/repos.d/agama-1.repo
May 12 00:36:19 <5> agama-web-server[2998]: agama-utils/src/lib.rs:84: Copying from /run/agama/zypp/etc/zypp/repos.d/agama-0.repo to /mnt/etc/zypp/repos.d/agama-0.repo
```

Why do you think it failed to add those repositories? Which are the symptoms? Bear in mind that local repositories are disabled at the end of the installation.

gcomes:
The symptoms I see are that the packages are downloaded from the opensuse servers and not from my repositories, which makes the installation process much slower.
And I figured out what is going on.
The repo files are named:
agama-0.repo
[...]

PR #3494, opened 2026-05-14
Honor extra repository priorities

officialasishkumar:
## Problem

The software profile schema exposes `extraRepositories[].priority`, but Agama currently drops that value before adding repositories to libzypp. That makes configured repository priority ineffective, which matters for setups that need an extra repository, such as a local mirror, to be preferred over default product repositories.

Related to #3483.

## Solution

Carry the configured priority through the software state and zypp wrapper, and set it on `zypp::RepoInfo` before adding the repository. Missing priorities are passed as `0`, which keeps libzypp's default priority.

## Testing

- `cargo fmt --all -- --check`
- `git diff --check`

I also attempted `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo test -p agama-software model::state::tests::test_add_user_repositories -- --nocapture`, but the local host does not have the libzypp development headers required to build `zypp-agama-sys` (`zypp-core/ui/progressdata.h`).


jreidinger (approved):
thanks for your contribution

jreidinger:
@officialasishkumar Ah, CI reveals, that integration test of libzypp bindings at https://github.com/officialasishkumar/agama/blob/623e67720647fb06ba1f504cf07a128cacce5d64/rust/agama-software/tests/zypp_server.rs#L116 is not updated with priority. Otherwise it looks good to me, so if you can also adapt it. Thanks

jreidinger (changes requested):
please adapt test to not fail.

officialasishkumar:
Adapted `rust/agama-software/tests/zypp_server.rs` to include `priority: None` in the `StateRepository` test fixture.

jreidinger (changes requested):
it still does not pass tests due to compilation error. I recommend to test using agama container - registry.opensuse.org/opensuse/tumbleweed:latest see how test runs using it in ci-rust.yml in github actions directory.

rust/package/agama.changes
```diff
@@ -23,6 +23,12 @@ Thu May 14 09:02:02 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 - Replace the user interface of the "agama monitor" command with
   a new version (gh#agama-project/agama#3443).
 
+-------------------------------------------------------------------
+Thu May 14 03:45:49 UTC 2026 - Asish Kumar <officialasishkumar@gmail.com>
```
jreidinger: it is not possible to have retroactively added entries, please use recent date and place it at the top of file.

officialasishkumar: Moved the changelog entry to the top with a current date in e4e829a.

jreidinger (approved):

jreidinger:
looks good now, lets just run CI on it

PR #3495, opened 2026-05-14
Feature bls sync1

ancorgs:
## Problem

The `feature-bls` feature branch is out of sync with `master`.  The `master` branch contains an important fix, without that we cannot build an image for `feature-bls` in OBS.

## Solution

Sync with master to get the latest fixes.


joseivanlopez (approved):
LGTM

PR #3499, opened 2026-05-14
Adapt network UI

teclator:
In https://github.com/agama-project/agama/pull/3416 we did some fixes in the backend, and this PR is for adapting the UI in order to use the correct connectivity enum values as well as use the connections state instead of the status.

The current status is not correct enough, as when there are state changes the current status is not updated accordingly and therefore, it should false values. Therefore, this PR brings back the state value which was used in the past but was removed at some point when working only with the extended config or with the config which does not expose the value as it is not a configuration at all.

## Screenshot

image

### DIFF
--- a/web/src/utils/network.ts
+++ b/web/src/utils/network.ts
@@ -66,7 +66,8 @@ const CONNECTION_STATE_LABELS: Record<ConnectionState, string> = {
 /**
  * Returns the translated label for a connection state.
  */
-const connectionStateLabel = (state: ConnectionState): string => CONNECTION_STATE_LABELS[state];
+// eslint-disable-next-line agama-i18n/string-literals
+const connectionStateLabel = (state: ConnectionState): string => _(CONNECTION_STATE_LABELS[state]);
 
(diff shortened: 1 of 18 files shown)


web/src/utils/network.ts
```diff
@@ -51,6 +52,22 @@ export const CONNECTION_TYPE = {
...
+  activating: N_("Activating"),
+  activated: N_("Activated"),
+  deactivating: N_("Deactivating"),
+  deactivated: N_("Deactivated"),
+};
+
+/**
+ * Returns the translated label for a connection state.
+ */
+const connectionStateLabel = (state: ConnectionState): string => CONNECTION_STATE_LABELS[state];
```
joseivanlopez: Is this actually translated? Or do you need `_(CONNECTION_STATE_LABELS[state])` ?

teclator: It is needed, good catch ;)

joseivanlopez (approved):
LGTM

PR #3502, opened 2026-05-15
Fix handling of AutoYaST rules

imobachgs:
## Problem

Agama is not able to import AutoYaST profiles using rules. Some limitations are expected (e.g., when working with dialogs), but in this case the `agama-autoyast` script just crashes without the user noticing.

The problem is that the Agama D-Bus service holds the libstorage-ng lock and, even if it was not the case, the storage probing is not done.

## Solution

Fortunately, the `AutoInstallRules` module needs to know just a little bit of information about storage. So this solution implements a minimal `StorageManager` which should suffice for most common use cases.

Additionally, this PR fixes the communication using the patched `Yast2::Popup.show` method and the `UI::Password` dialog.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/service/lib/agama/autoyast/profile_fetcher.rb
+++ b/service/lib/agama/autoyast/profile_fetcher.rb
@@ -55,6 +55,7 @@ module Agama
       # @return [ProfileHash,nil] an evaluated AutoYaST profile
       def fetch
         import_yast
+        replace_storage_manager
         read_profile
       end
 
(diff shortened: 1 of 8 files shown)


joseivanlopez (commented):

imobachgs:
The entry is there :-)

joseivanlopez:
Wow, if it works, then LGTM as temporary and quick solution. But I think we need to modify the autoyast code to avoid using Y2Storage at all.

joseivanlopez (approved):
LGTM

imobachgs:
Yes, I completely agree. The AutoYaST part is in need of a deep refactoring.

PR #3503, opened 2026-05-15
Make sure bootloader is installed before final storage steps

ancorgs:
## Problem

The final steps of some encryption processes need to be performed once the bootloader is already configured.

## Solution

Reorder the steps in the storage finish process.

Bonus: update dependency on yast2-storage-ng since we also need some corrections there.

## Testing

Tested manually


joseivanlopez (approved):
LGTM

PR #3504, opened 2026-05-15
Fixed starting the nmtui and self-update services

lslezak:
## Problem

- The self-update service is not started at boot

## Details

- The service files have been moved from `/usr/lib/systemd/system` to `/etc/systemd/system`
- The dracut complains:
```
dracut[I]: *** Including module: live-self-update ***
dracut-install: ERROR: installing '/usr/lib/systemd/system/live-self-update.service'
dracut[E]: FAILED: /usr/lib/dracut/dracut-install -D /var/tmp/dracut.AltRpF/initramfs -a /usr/lib/systemd/system/live-self-update.service live-self-update
Failed to enable unit: Unit live-self-update.service does not exist
```
- Unfortunately dracut continues and ignores the error, that's the reason nobody noticed the problem

## Solution

- Fix the service file location
- The same problem was found in the `initrd-nmtui.service` file
- Use `set -e` to avoid similar problems in the future
-  Return the original exit value so the `systemctl status live-self-update.service` does not report a failure

## Testing

- Tested with a locally built ISO, the self-update service starts properly

### DIFF
--- a/live/live-root/usr/bin/live-self-update
+++ b/live/live-root/usr/bin/live-self-update
@@ -118,6 +118,9 @@ cleanup_chroot() {
     findmnt "$NEWROOT"/proc > /dev/null && umount "$NEWROOT"/proc
     findmnt "$NEWROOT"/dev > /dev/null && umount "$NEWROOT"/dev
   fi
+
+  # return the original exit value
+  exit "$result"
 }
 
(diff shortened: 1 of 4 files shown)


teclator (approved):
LGTM

PR #3510, opened 2026-05-16
Support agama monitor without a terminal

imobachgs:
## Summary
Enables `agama monitor` to run in environments without a TTY (systemd services, automation, background processes). It solves a problem with prevents the autoinstallation, given that the terminal is not available when running as a systemd service (`agama-autoinstall` is a systemd service).

We should consider this one a critical error to besolved before release version 21.

<details>
<summary>Error during auto-installation</summary>

autoinstall-error
</details>

## Changes
- Auto-detect terminal availability using `std::io::IsTerminal`
- Add headless mode with plain text output
- Move monitoring logic from agama-cli to agama-lib (better architecture)
- Monitor owns stop_on_idle logic and emits status updates
- Return Receiver directly (removed MonitorClient wrapper)
- Channel closure properly signals completion (idiomatic Rust)

## Result
- Works in both TUI and headless modes automatically
- -141 lines of code
- Cleaner API and architecture
- Channel closes correctly when monitoring stops


rust/agama-cli/src/monitor.rs
```diff
@@ -62,19 +66,61 @@ fn restore_terminal(terminal: &mut Terminal<CrosstermBackend<io::Stdout>>) -> Re
...
+    stop_on_idle: bool,
+) -> Result<()> {
+    let (mut updates, status) = Monitor::connect(websocket, &http_client, stop_on_idle).await?;
+
+    eprintln!("Agama monitor started (headless mode)");
+    eprintln!("Initial stage: {:?}", status.status.stage);
+
+    // Listen to updates until channel closes
+    while let Ok(status) = updates.recv().await {
+        eprintln!(
```
jreidinger: why having everything on error output? I kind of expect it should use common output, not?

jreidinger (approved):

PR #3518, opened 2026-05-18
Use the "monitor" icon for the "System" group in the web UI

lslezak:
## Problem

- Weird icon for the "System" group

agama-system-icon-old

## Solution

- Use the "monitor" icon instead of "fingerprint"

## Testing

- Tested manually

## Screenshots

agama-system-icon-new

### DIFF
--- a/web/src/components/layout/Icon.tsx
+++ b/web/src/components/layout/Icon.tsx
@@ -35,6 +35,7 @@ import Backspace from "@icons/backspace.svg?component";
 import CheckCircle from "@icons/check_circle.svg?component";
 import ChevronLeft from "@icons/chevron_left.svg?component";
 import ChevronRight from "@icons/chevron_right.svg?component";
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
Since it is so difficult to choose one of available icons for this and, honestly, I have no strong opinion between these two, fine with me to go ahead with the change.

It hurts nobody, and we can change it whenever a fresh proposal suits better. 

Thanks!

lslezak:
The monitor icon is not perfect, but it is still better than the fingerprint. If somebody has a better idea we can always improve it later.

dgdavid:
> The monitor icon is not perfect, but it is still better than the fingerprint.

Icon interpretation is really subjective, so it's hard to say one is better than the other, though I can see why you would  prefer the monitor over the fingerprint. In any case, this is a minor detail, but your contribution definitely helps keep things clear.

> If somebody has a better idea we can always improve it later.

Indeed.

PR #3527, opened 2026-05-20
[SLE-16.0] Use the openSUSE Leap 16.0 system in the CI container

lslezak:
## Problem

- Using the openSUSE Tumbleweed system in CI for the SLE-16.0 builds makes troubles, it might contain newer incompatible packages.

## Solution

- Use the openSUSE Leap 16.0 system in the CI container, that contains the same packages as SLE-16.0.

## Testing

- See the CI test runs.
- The storage unit test was failing even before this change, it is not a regression.

### DIFF
--- a/.github/workflows/ci-live.yml
+++ b/.github/workflows/ci-live.yml
@@ -11,7 +11,7 @@ on:
 
       # this file as well
       - .github/workflows/ci-live.yml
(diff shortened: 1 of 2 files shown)


ancorgs (approved):
LGTM

PR #3532, opened 2026-05-21
Disconnect not bound connection devices

teclator:
## Problem

When a connection bind is modified it does not disconnect current devices using it.

### Example Scenario:

1. A default DHCP profile is created without any device bindings, allowing it to apply to any ethernet interface (e.g., `eth0` and `eth1`).
2. Both `eth0` and `eth1` connect successfully using this profile.
3. The profile is then modified to explicitly bind only to `eth0`.
4. **Actual Behavior:** `eth1` remains connected using the modified profile.
5. **Expected Behavior:** `eth1` should be automatically disconnected since it is no longer compatible with the profile's constraints.

- https://bugzilla.suse.com/show_bug.cgi?id=1265446
- https://trello.com/c/wv1eirIM/5752-2-1265446-sles-161-devices-remain-connected-after-a-connection-bind-setting-is-modified

## Solution

When a connection bind settings are modified, then all the devices that do not match the bind setting anymore will be disconnected. It also ensure a device is disconnected if it is the connection device bind and it is using other connection (for example a device is using the default profile and we create a new connection bind to it).

## Testing

- *Tested manually*

## Screenshots

Screenshots showing how a default connection with `multi-connect=3` and without a bind now disconnect the devices which are not explicitly bound to the connection when bind it to an specific one.

Screenshot From 2026-05-22 07-08-29
Screenshot From 2026-05-22 07-08-44
Screenshot From 2026-05-22 07-09-00
Screenshot From 2026-05-22 07-09-11


rust/agama-network/src/nm/adapter.rs
```diff
@@ -54,29 +54,42 @@ impl<'a> NetworkManagerAdapter<'a> {
...
         tracing::info!("Activating connection {}", &conn.id);
         self.client.activate_connection(path).await
     }
+
+    /// Ensures that devices are in a consistent state before activating a connection.
+    ///
+    /// It disconnects any device that is currently using the connection but is no longer
+    /// compatible with it. Also, it disconnects any device that should be used by the
+    /// connection (is a target) but is currently using a different one.
+    async fn cleanup_devices(&self, conn: &Connection) -> Result<(), NmError> {
```
teclator, line 67: I was not sure about the method name, could rename it to **disconnect_conflicting_devices**, but the description should be enough.

rust/agama-network/src/nm/adapter.rs
```diff
@@ -212,6 +225,10 @@ impl Adapter for NetworkManagerAdapter<'_> {
                     }
                 };
 
+                if let Err(e) = self.cleanup_devices(conn).await {
```
jreidinger, line 227: I am not sure if I get logic here...it is called here and then also below when activate_connection. Why it needs to be run twice? if there is good reason, I would like to see it in comment in code.

rust/agama-network/src/model.rs
```diff
@@ -1068,6 +1068,47 @@ impl Connection {
...
+                    return false;
+                }
+            } else {
+                return false;
+            }
+        }
+        true
+    }
+
+    /// Determines whether the connection matches the given device name or MAC address.
```
jreidinger, line 1096: can it happen that connection specify interface AND mac_address? if so, then this or looks a bit strange as I expect condition is and.

teclator: In general you could, not in the UI but nothing prevents you to doing so in the CLI, it could be even a conflicting one which is wrong. In the UI we allow to select only betweem name or MAC and you have a selector so it is not prone to errors. There is even a matching settings which you could be used but that is out of the scope of this PBI as it is quite more complex to determine which devices match the condition and which not.

rust/agama-network/src/nm/adapter.rs
teclator: Good question, I need to check, I had the code stashed during a while and would need to check if I added for a good reason at all, or just checking some corner case if a connection is updated without activation. Probably evaluating about moving the cleanup out of the activation. For example, a connection which is set to down and the device bind is modified, it only disconnect one device when disconnected. A little bit strange.. but in general, our main goal is to disconnect devices which activating as it is problematic if the bind device is already applying other profile or maintain connected devices which should not anymore.

teclator: If we wanted to disconnect other devices when any connection is updated independently of the status then we should remove the cleanup from the activation.

jreidinger: so ideally please write it down to code as we for sure forget it in few days :)

jreidinger (approved):

PR #3534, opened 2026-05-22
New progress tracking in the CLI

imobachgs:
## Problem

Agama implements an `agama monitor` command to track the progress from the CLI. Additionally, when loading a profile (`agama config load`) it uses the same UI.

The current implementation uses the the full screen through the terminal "alternate screen". This approach has some problems:

- Going to full screen when loading the profile is kind of unexpected.
- In some terminals, the "alternate screen" is not available.
- It does clean-up the screen properly (see [bsc#1265936](https://bugzilla.suse.com/show_bug.cgi?id=1265936)).

## Solution

- This PR introduces a different way to display the information. Now the monitor is shown inline and it properly cleans up the used area.
- It allows stopping the monitor pressing CTRL+c.

image

## Testing

- Tested manually

## Follow-up

These are some ideas for the short-term:

- Improve the raw monitor, which is used when a terminal is not availalble (e.g., when redirecting to a file or running as part of the `agama-autoinstall` service).
- When using CTRL-c in the `agama install` command, it should tell the user that the installation continues although the monitoring stopped.
- Display a summary after the command has finished (`config load`, `install`).
- Remove the `Profile is valid` message.


rust/agama-cli/src/monitor.rs
```diff
@@ -20,49 +20,46 @@
...
 /// Restores the terminal to normal mode
 fn restore_terminal(terminal: &mut Terminal<CrosstermBackend<io::Stdout>>) -> Result<()> {
-    disable_raw_mode()?;
-    execute!(
-        terminal.backend_mut(),
-        LeaveAlternateScreen,
-        DisableMouseCapture
-    )?;
     terminal.show_cursor()?;
+    terminal.clear()?;
```
jreidinger, line 61: is terminal.clear intended? What happens if it is not there? I think it should keep the final state to user

imobachgs: It is a good question. Yes, it is what I want (to clear the work area). However, we plan to display a summary at the end (see the "follow-up" section).

jreidinger (approved):

PR #3537, opened 2026-05-22
live: Added `agama-screenshot` helper script

lslezak:
## Problem

- Taking screenshots of the Agama installer running locally is difficult.

## Solution

- Added `agama-screenshot` helper script which saves the whole current screen.

## Notes

- Enabling the unsafe mode in Gnome Kiosk should be fine in our case, there is nothing else running in the machine besides the Agama installer. And you need root access to communicate with the DBus session bus.
- Ideally we should set some keyboard shortcut for that but unfortunately Gnome Kiosk does not support that. :worried: 

## Testing

- Tested manually

```console
agama:~ # agama-screenshot
Screenshot saved to agama-2026-05-22-Pa5h.png
agama:~ # ll agama-2026-05-22-Pa5h.png
-rw-r--r-- 1 root root 112769 May 22 16:53 agama-2026-05-22-Pa5h.png
agama:~ # agama-screenshot agama.png
Screenshot saved to agama.png
agama:~ # ll agama.png 
-rw-r--r-- 1 root root 236302 May 22 16:53 agama.png

```

## Screenshots

Here are the screenshots taken by the commands above.

The default Agama screen after boot:
agama-2026-05-22-Pa5h

It captures the whole currently visible screen, if you open the Firefox developer tools then the window is captured as well:
agama


live/live-root/usr/bin/agama-screenshot
```diff
@@ -0,0 +1,46 @@
...
+	exit 0
+fi
+
+# Gnome kiosk fails if the target file already exists,
+# create the screenshot in a temporary directory and then move it
+SCREENSHOT_DIR=$(mktemp -d -p /tmp "screenshot-XXXXXXX")
+FILE="$SCREENSHOT_DIR/screenshot.png"
+
+# see the DBus API description at
+# https://gitlab.gnome.org/GNOME/gnome-kiosk/-/blob/main/dbus-interfaces/org.gnome.Shell.Screenshot.xml?ref_type=heads
```
mvidner, line 29:
```suggestion
# https://gitlab.gnome.org/GNOME/gnome-kiosk/-/blob/main/dbus-interfaces/org.gnome.Shell.Screenshot.xml?ref_type=heads
# include cursor: false
# flash screen: false
```

imobachgs:
I wonder if we should integrate this as an agama command instead of a helper script. There is no need to include it in the Rust code. Perhaps we might think about having the `agama-screenshot` executable in `/usr/libexec` and get them execute when calling `agama screenshot...`. Just an idea.

lslezak:
Well, I was thinking about something like this but actually the script is more bound to the Live ISO environment and the Gnome Kiosk running there than the Agama itself.  If you run Agama in some development/testing environment using a full desktop then the script won't work, this is really something only for the Live ISO.

imobachgs (approved):

PR #3529, opened 2026-05-21
Add ignition and combustion to live image

Volcano28:
## Problem

The live image currently does not include the packages needed for the new firstboot mechanism.

## Solution

Add `ignition` and `combustion >= 1.2` to the Agama live image package list.

## Testing

- Tested manually by checking the generated diff.
- No unit tests added; this is a package list change.

## Screenshots

Not applicable.

## Documentation

Not applicable.

PR #3544, opened 2026-05-25
Fixed Kiwi file

lslezak:
## Problem

- Kiwi does not build the Live image and prints error " could not parse build description (kiwi)  element 'packages' closes, but I expected 'package'"

## Solution

- The `>` character in an attribute string must be escaped.
- Replace `combustion >= 1.2` with `combustion &gt;= 1.2`
- Related to #3529

## Testing

- Tested manually, the image can be now built locally

## Note

Later it turned out that both ignition and combustion do not start properly in the Live ISO environment.
See https://bugzilla.suse.com/show_bug.cgi?id=1266257.

As a temporary workaround to fix the booting problem both services have been disabled.

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon May 25 12:06:55 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
+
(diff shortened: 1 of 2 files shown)


kobliha (approved):
Thank you!

imobachgs (approved):

PR #3547, opened 2026-05-25
Ignition + combustion workaround, always force the first boot mode

lslezak:
## Problem

- Ignition and combustion do not start properly in the Live ISO
- https://bugzilla.suse.com/show_bug.cgi?id=1266257

## Solution

- It turned out that the `firstboot-detect.service` which should detect whether the current boot is the first boot or not does not work properly in the Live ISO environment.
- The workaround is to disable the first boot detection and always behave like in the first boot. That's actually true for the installer, we always start with the same read-only root filesystem image from scratch.

## Testing

- Tested manually
- The ISO builds and boots fine
- Tested with a simple combustion script
  - Remote script: boot with the `rd.neednet=1 ip=dhcp combustion.url=http://.../script rd.break` boot options
  - Local script: create disk partition with label `COMBUSTION`, place the script into `/combustion/script` file (do not forget to make it executable)
  - In both cases the script was correctly executed by Combustion .

### DIFF
--- a/live/live-root/usr/lib/dracut/modules.d/95initrd-firstboot/module-setup.sh
+++ b/live/live-root/usr/lib/dracut/modules.d/95initrd-firstboot/module-setup.sh
@@ -18,7 +18,7 @@ install() {
   #
   # The problem is that the first boot detection fails because of the
   # complicated Live ISO setup which uses the device mapper for creating
-  # writable overlay over the squashfs read-only image.)
+  # writable overlay over the squashfs read-only image.
 	$SYSTEMCTL -q --root "$initdir" disable firstboot-detect.service
 	$SYSTEMCTL -q --root "$initdir" enable firstboot.target
 }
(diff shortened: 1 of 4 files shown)


live/live-root/usr/lib/dracut/modules.d/95initrd-firstboot/module-setup.sh
```diff
@@ -0,0 +1,24 @@
...
+}
+
+install() {
+  # Do not try detecting whether this is the first boot or not, consider every
+  # boot as the first boot. That's true for the installer, we always start with
+  # the same read-only root filesystem image from scratch.
+  #
+  # The problem is that the first boot detection fails because of the
+  # complicated Live ISO setup which uses the device mapper for creating
+  # writable overlay over the squashfs read-only image.)
```
jreidinger: ending `)` do not match any opening one...at first I thought it should be smiley :)

jreidinger (approved):

imobachgs (approved):

PR #3548, opened 2026-05-25
feat(web): Add VLAN support to the UI

teclator:
## Problem

The UI does not allow to define a VLAN connection although it is already supported by the backend.
It is expected to integrate it in the connection form, for which it will offer
  - All common connection settings
  - VLAN specific settings
    - Id (numeric between 0 and 4094)
    - Parent Interface
    - Protocol For Encapsulation (selector, default 802.1Q)

-  https://trello.com/c/uAUbfne6/5758-3-network-set-up-a-vlan-connection-through-the-web-ui

## Solution

Implemented the VLAN support in the UI allowing to select it when defining a new connection or modifying an existing one.

## Testing

  - Added a new unit test
  - Tested manually

## Screenshots 

image


web/src/components/form/NumberField.tsx
```diff
@@ -53,7 +53,7 @@ export default function NumberField({ label, helperText, min, max }: NumberField
       <TextInput
         id={field.name}
         name={field.name}
-        value={field.state.value}
+        value={field.state.value ?? ""}
```
dgdavid: Why? I expect the field be initialize in the form initialization, don't you?

web/src/components/network/connection-form/fields.ts
```diff
@@ -220,6 +231,12 @@ const defaultValues = {
   bridgeHelloTime: undefined,
   bridgeMaxAge: undefined,
   bridgePorts: [] as string[],
+
+  // VLAN fields
+  vlanIface: "",
+  vlanId: undefined as number | undefined,
```
dgdavid: We need to check this.

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -0,0 +1,149 @@
...
+        <form.AppField name="vlanId">
+          {(field) => (
+            <field.NumberField
+              label={
+                // TRANSLATORS: label for the VLAN ID field.
+                _("VLAN ID")
+              }
+              helperText={
+                // TRANSLATORS: helper text for the VLAN ID field.
+                _("Numeric between 0 and 4094")
```
dgdavid: What about following same pattern than in Bridge fields?

Image


```suggestion
                _("Numeric identifier (0–4094)")
```

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -0,0 +1,149 @@
...
+        <form.AppField name="vlanParent">
+          {(field) => (
+            <field.TextField
+              label={
+                // TRANSLATORS: label for the VLAN parent device field.
+                _("Parent device")
+              }
+              helperText={
+                // TRANSLATORS: helper text for the VLAN parent device field. %s is a list of available devices.
+                sprintf(_("Available devices: %s"), formatList(devices.map((d) => d.name)))
```
dgdavid: I understand we are not offering a device selector here because users can enter whatever they want there, even though is not an available device at the time of creating the connection, right?

If so, I believe helper text should be improved, since given that one looks like these are the only options I can enter there, which is surprising why not a selector instead.

I know that it is also present for other connections types fields, but I didn't catch them before, sorry.

What about something like, 


```suggestion
                sprintf(_("Physical or virtual interface (e.g. %s)"), devices.map((d) => d.name).join(", "))
```

Also, wondering why only add a list of physical devices if they can use only previously created bridges or bond (sorry if I'm wrong on that statement). Maybe, getting ride of these list is better and keep the helper text to explain what is expected there, no values.

dgdavid (commented):
Thanks @teclator 

It really looks good to me, but I have two wording doubts/proposals that would like to discuss before merging.

web/src/components/network/connection-form/VlanFields.tsx
teclator: In our expectation, it should be one of the existent devices and that is what we list there. Although the true is that it could be event a connection UUID but we are not exposing them in the HTTP API, and there is also the option to be absent requiring the 802-3-ethernet MAC address to be specified.. but that is out of the scope.

If you want to create a VLAN with a bond parent it should exist already otherwise the connection will not be activated. But of course, nothing prevents you to try to define first he VLAN without an existing device. I like the suggestion, but it looks like just an example and not show which are the available ones.. so.. maybe a mix of both options could help...

web/src/components/network/connection-form/fields.ts
teclator: Already fixed

web/src/components/network/connection-form/VlanFields.tsx
teclator: image

dgdavid: My point is that if these are a fixed list of known devices... why not offer then in a selector?

teclator: If we can offer it plus an input for not existing ones then great

dgdavid: As discussed in a meeting, let's keep the helper text as a helper text (what is expected) and use the https://github.com/agama-project/agama/blob/master/web/src/components/form/SuggestionsTextField.tsx component for this input, feeding it with existing devices. I think it will provide what you are looking for. If you need to see it in action, check the new authentication form while defining a user.

dgdavid (approved):
LGTM, thanks for addressing these latest concerns.

PR #3549, opened 2026-05-26
feat(storage): LVM policy for generating physical volumes

joseivanlopez:
New LVM volume groups use the needed space to allocate its logical volumes. For example, if two disks are selected as target devices for creating physical volumes and everything fits into the first disk, then the second disk is not used at all.

This behaviour could be unexpected, so now Agama allows configuring the policy for creating new physical volumes:

## `useNeeded` policy

This is the policy for the current behaviour. Only needed space is used. It is the default value to keep backward compatibility with old profiles.

~~~json
{
  "physicalVolumes": [
    { 
      "generate": {
        "targetDevices": ["disk1", "disk2"],
        "spacePolicy": "useNeeded"
      }
    }
  ]
}
~~~

If a config is loaded using the CLI and `useNeeded` is used, then the UI will inform to the user, but it does not allow changing the policy. In this case, the volume group has to be removed and created again in order to use `useAvailable` policy. 

localhost_8080__

## `useAvailable` policy

Creates the volume group using all the space available in the target devices. Volume groups created through the UI will always use this policy. 

~~~json
{
  "physicalVolumes": [
    { 
      "generate": {
        "targetDevices": ["disk1", "disk2"],
        "spacePolicy": "useAvailable"
      }
    }
  ]
}
~~~

### DIFF
--- a/rust/agama-lib/share/storage.model.schema.json
+++ b/rust/agama-lib/share/storage.model.schema.json
@@ -98,6 +98,9 @@
           "type": "array",
           "items": { "type": "string" }
         },
+        "targetDevicesPolicy": {
+          "enum": ["useNeeded", "useAvailable"]
+        },
         "spacePolicy": { "$ref": "#/$defs/spacePolicy" },
         "logicalVolumes": {
           "type": "array",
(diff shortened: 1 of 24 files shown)


mvidner:
What about `*.changes`, and other Documentation bits of the PR template?

joseivanlopez:
> What about `*.changes`, and other Documentation bits of the PR template?

The documentation is added here https://github.com/agama-project/agama-project.github.io/pull/170

ancorgs (approved):
LGTM

PR #3550, opened 2026-05-26
refactor(web): simplify IssuesAlert component content

dgdavid:
The _Before starting the installation, you need to address the following problems:_ alert title appeared in each section of the Agama web UI. 

This made sense in the former interface, where users could land directly on a section without prior context. It no longer does: users now see the overview first, where installation is clearly blocked with _You need to fix any invalid settings before proceeding with the installation._ and each invalid section is already flagged, inviting users to visit it for details.

Repeating the installation preamble inside each section is therefore redundant and noisy. This PR simplifies `IssuesAlert` to:

- Show the issue description directly as the alert title when there is only one issue.
- Show a concise list of issues under a shorter title when there are more than one.

> [!NOTE]
> Besides reducing noise, this makes the actual problem more immediately visible to the user and saves vertical space by avoiding the extra bullet list when only one issue is present.

The `solver` issue class (now known as  `software_conflict`) and its associated _Review and fix_ link to the conflict resolution page have also been dropped. Conflict resolution is not yet available in the UI following the API changes,
and is expected to be handled differently going forward.

| Before | After |
|-|-|
| localhost_8080_ (96) | localhost_8080_ (95) |

### DIFF
--- a/web/src/components/core/IssuesAlert.tsx
+++ b/web/src/components/core/IssuesAlert.tsx
@@ -21,34 +21,25 @@
  */
 
 import React from "react";
-import { Alert, List, ListItem } from "@patternfly/react-core";
+import { isEmpty } from "radashi";
+import { Alert, AlertProps, List, ListItem } from "@patternfly/react-core";
 import { _ } from "~/i18n";
-import Link from "./Link";
-import { PATHS } from "~/routes/software";
+
 import type { Issue } from "~/model/issue";
 
-export default function IssuesAlert({ issues }) {
-  if (issues === undefined || issues.length === 0) return;
+export default function IssuesAlert({ issues }: { issues: Issue[] }) {
+  if (isEmpty(issues)) return;
+  const props: Partial<AlertProps> = { isInline: true, variant: "warning" };
+
+  if (issues.length === 1) {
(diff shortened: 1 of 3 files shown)


web/src/components/core/IssuesAlert.tsx
```diff
@@ -21,34 +21,25 @@
...
+    return <Alert {...props} title={issues[0].description} />;
+  }
 
   return (
-    <Alert
-      isInline
-      variant="warning"
-      title={_("Before starting the installation, you need to address the following problems:")}
-    >
+    <Alert {...props} title={_("The following issues need to be addressed")}>
```
imobachgs: I would avoid the passive voice, but it is just a personal preference (*You need to address the following issues:*).

imobachgs (approved):

web/src/components/core/IssuesAlert.tsx
dgdavid: As discussed in a meeting, I have chosen "You must fix these issues" https://github.com/agama-project/agama/pull/3550/commits/435865b67fae98f35e2d6a7dfb3d0fc01ddc3e9f

imobachgs (approved):

PR #3553, opened 2026-05-27
No sysinfo in the output of agama status command

mchf:
## Problem

Output of ```agama status``` command displays even system info which has nothing to do with the installer status

## Solution

This happened when an internal structure was extended to contain more data for another reason. This structure was originally directly reused. This fix explicitly picks data to be displayed in the output of ```agama status```

## Testing

- *Tested manually*

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -2,7 +2,7 @@
 Wed May 27 07:40:40 UTC 2026 - Michal Filka <mfilka@suse.com>
 
 - Do not display system information in output of agama status
-  command 
+  command (gh#agama-project/agama#3553).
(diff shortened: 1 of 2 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed May 27 07:40:40 UTC 2026 - Michal Filka <mfilka@suse.com>
+
+- Do not display system information in output of agama status
+  command
```
jreidinger: please add this PR as link to changes

jreidinger (approved):

PR #3554, opened 2026-05-27
bsc#1266276: wait for the software service when setting the configuration

imobachgs:
## Problem

[bsc#1266276](https://bugzilla.suse.com/show_bug.cgi?id=1266276).

Even if the `inst.auto` parameter is set, there is a chance that Agama does not start the installation automatically. It is kind of unfortunate, but [bsc#1266276](https://bugzilla.suse.com/show_bug.cgi?id=1266276) probes that it is possible.

The problem is that `agama-autoinstall` asks the server to load a configuration and it waits until there are no pending progresses, hence there are no running tasks. However, there is a chance that, once Agama loads the configuration (1), the software service does an additional update of its proposal (2). So `agama-autoinstall` can detect (1), ask for starting the installation and then conflict with (2).

## Solution

* The manager's `SetConfig` action waits until the software service finishes.
* The manager's `SetConfig` action is now responsible for setting the packages/patterns (resolvables) that are required by other services. Those services won't talk to the software service anymore, so the manager controls exactly when the whole process of loading the configuration has finished.

## Testing

- Added a new unit test
- Tested manually

## TODO

- [x] Wait for the software service to finish.
- [x] Change the manager to set the resolvables.
- [x] Adapt agama-ntp.
- [x] Adapt agama-files
- [x] Adapt storage-related services (bootloader and storage).


imobachgs:
Replaced with #3577.

PR #3557, opened 2026-05-28
Reset settings when changing the product through the UI

imobachgs:
## Problem

* [bsc#1258032](https://bugzilla.suse.com/show_bug.cgi?id=1258032).

The UI is resetting the full configuration when a product is selected. So if you load a configuration (e.g., adding some pre-configured stuff), it will be cleared when selecting a product through the user interface.

However, in some scenarios, you might want the configuration to get reset. For instance, consider that you choose some software. If you change the product and the software you selected is not available in the new product, you will get an issue (that you cannot fix using the UI).

## Solution

* If there is no product selected yet, then keep the settings.
* If there is a product (product change), then reset all the settings.

## Testing

- Added a new unit test
- Tested manually


mchf (approved):
logic seems to be according to the description and contains quite many descriptive tests -> LGTM

PR #3558, opened 2026-05-28
Remote access

jreidinger:
## Problem

There is no easy way to enable remote access after installation. For ssh ssh keys has to be used and for cockpit there is no option at all.

- https://trello.com/c/k4wwzAJA/5756-8-agm-151-add-an-option-to-enable-remote-access-to-the-installed-system

## Solution

Implement it as option for agama profile. Also use it for agama-users when ssh is requested.

## Testing

- *Added a new unit test* TODO
- *Tested manually* TODO

## Documentation

- Run: `git ls-files '*.md'`


rust/agama-files/src/service.rs
```diff
@@ -302,7 +302,10 @@ impl MessageHandler<message::Finish> for Service {
 
         let scripts = self.scripts.lock().await;
         if !scripts.by_group(ScriptsGroup::Init).is_empty() {
-            enable_service(&self.install_dir, "agama-scripts");
+            let res = enable_service(&self.install_dir, "agama-scripts").await;
```
imobachgs: I wonder if we should have an infallible version, because all of the calls to this function will do the same (logging the error).

We could still have a `try_enable_service` which returns a `Result`. So the caller decides what to do.

rust/agama-manager/src/actions.rs
```diff
@@ -121,6 +122,9 @@ impl InstallAction {
         self.hostname.call(hostname::message::Install).await?;
         self.users.call(users::message::Install).await?;
         self.storage.call(storage::message::Finish).await?;
+        self.remote_access
+            .call(agama_remote::message::Finish)
+            .await?;
 
         // call files before storage finish as it unmounts /mnt/run which is important for chrooted scripts
```
imobachgs: I think this comment does not apply anymore. I know, not related to this PR, but if you can remove it... :-)

rust/agama-manager/src/actions.rs
```diff
@@ -227,6 +232,12 @@ impl SetConfigAction {
         self.security
             .call(security::message::SetConfig::new(config.security.clone()))
             .await?;
+        // as remote access is fast call, lets just use it together with security
```
imobachgs: Many calls are fast enough. Perhaps it would be better to add a step.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    fn is_cockpit_enabled(&self) -> bool {
+        self.user_config.cockpit == Some(api::remote_access::AccessEnum::Enabled)
+            || self
+                .agama_config
+                .values()
+                .any(|config| config.cockpit == Some(api::remote_access::AccessEnum::Enabled))
+    }
+
+    pub fn final_config(&self) -> Config {
```
imobachgs: What "final" means? We have the concept of Config and ExtendedConfig.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        Ok(())
+    }
+
+    pub fn set_user_config(&mut self, config: Config) {
+        self.user_config = config;
+        self.update_resolvables();
+        // ignoring error here is ok as it means just dismissed event
+        let _ = self.events.send(Event::ProposalChanged {
+            scope: Scope::RemoteAccess,
```
imobachgs, line 209: Is it possible that we emit this signal twice?

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,270 @@
...
+                r#type: agama_software::ResolvableType::Pattern,
+            });
+        }
+        if self.is_ssh_enabled() {
+            resolvables.push(Resolvable {
+                name: "openssh-server".to_string(),
+                r#type: agama_software::ResolvableType::Package,
+            });
+        }
+        let res = self.software.cast(agama_software::message::SetResolvables {
```
imobachgs: #3554 changes how resolvables are reported. You should not call `SetResolvables` anymore but just add a `GetResolvables` function that returns what you need. The manager service will take care of syncrhonizing with the software service.

imobachgs (commented):
The approach looks good. Just some minor comments and a warning about how to handle resolvables.

rust/agama-files/src/service.rs
jreidinger: that makes perfect sense for me....but still I hope it will be better if we have installation issues and we can then register issue when error happen like "agama-scripts error, it is possible that init scripts won't run. Error: {}" as logging is weak way how to report such issues to user.

rust/agama-remote/src/service.rs
jreidinger: in users I see full_config......so here it is probably extended_config...I think Config and ExtendedConfig does not make sense here as I do not use Option and instead rely on default which is Default.

rust/agama-remote/src/service.rs
jreidinger: will adapt when you merge it as I need now something that I can use for testing.

rust/agama-remote/src/service.rs
jreidinger: why do you think? It can happen if you load config with Users and RemoteAccess keys (as it can twice affect proposal), but not sure how to avoid it.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,270 @@
...
+        Self {
+            user_config: Config::default(),
+            agama_config: HashMap::new(),
+            software,
+            events,
+        }
+    }
+
+    fn is_ssh_enabled(&self) -> bool {
+        self.user_config.ssh == Some(api::remote_access::AccessEnum::Enabled)
```
jreidinger: here is wrong logic. We need to recognize explicit user Default and if it is specified by user, it should have precedence over agama_config.

rust/agama-files/src/service.rs
imobachgs: Yes, sure, but that's a different topic. We have a PBI for that.

rust/agama-remote/src/service.rs
imobachgs: `Config` is our convention for the user's configuration. ExtendedConfig is the full one, including default values and so on.

rust/agama-remote/src/service.rs
imobachgs: Bear in mind that the services do not need to distinguish between full/user configuration. They always work with the full configuration (the user configuration is kept at manager's level).

imobachgs (commented):
I had a different idea in mind, but it was maybe too simplistic. I was thinking that `agama-remote` could receive the configuration and run after the `agama-user` module. So:

* if SSH is set to `default`, it does `systemctl preset ssh` (so it keeps the distribution default).
* if SSH is set to enable/disable, it just do the right thing.
* if SSH is not configured, it does nothing. So if the user module already enabled the service, then it remains untouched.

Having said that, the API will only return the user settings (if any). No proposal at all. The downside is that, if you add a public key to your `root` configuration, there will be no indication at API level that the service will be open. The benefit is simplicity: 1) `agama-user` module does not need to talk to `agama-remote` and 2) you do not need two configurations ("user" and "module").

Of course, your approach is more extensible, because it enables other modules to open/enable services. But it is something that we are avoid in Agama if possible.

Having said that, I am fine with any approach we follow, I do not have a strong opinion.

rust/agama-manager/src/actions.rs
jreidinger: well, in the end I check more why we split it and improve comment to clarify it and add also reference to bugzilla report.

rust/agama-lib/share/profile.schema.json
```diff
@@ -868,6 +868,23 @@
...
     "iscsi": {
       "$ref": "iscsi.schema.json"
+    },
+    "remoteAccess": {
+      "title": "Remote access settings",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "ssh": {
+          "title": "SSH remote access",
```
imobachgs, line 878: np: I think you should use "description" here. Some of the code generators we have tried will use the "title" as identifier (so you will have `SSHRemoteAccess` as the field name). If you use "description", it will default to "ssh" (the key).

Additionally, I think a description of what each value means would help.

rust/agama-manager/src/service.rs
```diff
@@ -781,6 +790,10 @@ impl MessageHandler<message::GetProposal> for Service {
         let storage = self.storage.call(storage::message::GetProposal).await?;
         let network = self.network.get_proposal().await?;
         let users = self.users.call(users::message::GetProposal).await?;
+        let remote_access = self
+            .remote_access
+            .call(agama_remote::message::GetProposal)
```
imobachgs, line 795: Please, rename `agama_remote` to `remote` to be consistent.

rust/agama-remote/src/test_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+use crate::{Service, Starter};
+
+/// Starts a testing security service.
+pub async fn start_service(
+    software: Handler<agama_software::Service>,
+    events: Sender<Event>,
+) -> Handler<Service> {
+    Starter::new(software, events)
+        .start()
+        .expect("Could not spawn a testing security service")
```
imobachgs:
```suggestion
        .expect("Could not spawn a testing remote service")
```

rust/agama-remote/src/test_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+//
+// To contact SUSE LLC about this file by physical or electronic mail, you may
+// find current contact information at www.suse.com.
+
+use agama_utils::{actor::Handler, api::Event};
+use tokio::sync::broadcast::Sender;
+
+use crate::{Service, Starter};
+
+/// Starts a testing security service.
```
imobachgs:
```suggestion
/// Starts a testing remote service.
```

rust/agama-users/src/service.rs
```diff
@@ -57,24 +60,30 @@ pub enum Error {
...
     events: event::Sender,
+    remote_access: Handler<agama_remote::Service>,
 }
 
 impl Starter {
-    pub fn new(events: event::Sender, issues: Handler<issue::Service>) -> Self {
+    pub fn new(
+        events: event::Sender,
+        issues: Handler<issue::Service>,
+        remote_access: Handler<agama_remote::Service>,
```
imobachgs, line 73: please, use `remote::` to be consistent.

rust/agama-users/src/service.rs
```diff
@@ -126,6 +141,29 @@ impl Service {
         None
     }
 
+    /// checks if users module needs to enable remote ssh access
```
imobachgs:
```suggestion
    /// checks if users service needs to enable remote ssh access
```

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub fn new(software: Handler<agama_software::Service>, events: Sender<Event>) -> Self {
+        Self {
+            user_config: Config::default(),
+            agama_config: HashMap::new(),
+            software: Some(software),
+            events,
+        }
+    }
+
+    /// Checks if SSH access is enabled either by user configuration or module requests.
```
imobachgs:
```suggestion
    /// Checks if SSH access is enabled either by user configuration or service requests.
```

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+    fn is_ssh_enabled(&self) -> bool {
```
imobachgs, line 123: np: I usually prefer to put private functions at the end, so when reading the code I get to the public API first.

rust/agama-utils/src/api/remote_access.rs
```diff
@@ -0,0 +1,61 @@
...
+//! Implements a data model for Bootloader configuration.
+
+use merge::Merge;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Allows to specify explicit enablement of remote access for supported services
+#[derive(Default, Clone, Copy, Debug, Serialize, Deserialize, PartialEq, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum AccessEnum {
```
imobachgs, line 29: Using the `Enum` suffix looks like of redundant to me.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+        if let Some(config) = &self.user_config.ssh {
```
imobachgs, line 124: I would say you do not need the `&` neither in this line nor the following one.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        // no user explicit selection, so compute it from agama requirements
+        self.agama_config
+            .values()
+            .any(|config| config.ssh == Some(api::remote_access::AccessEnum::Enabled))
+    }
+
+    /// Checks if Cockpit access is enabled either by user configuration or module requests.
+    fn is_cockpit_enabled(&self) -> bool {
+        if let Some(config) = &self.user_config.cockpit {
```
imobachgs, line 136: Not needed &.

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// Channel to emit events.
+    events: Sender<Event>,
+}
+
+impl State {
+    /// Creates a new state instance.
+    pub fn new(software: Handler<agama_software::Service>, events: Sender<Event>) -> Self {
+        Self {
+            user_config: Config::default(),
+            agama_config: HashMap::new(),
```
imobachgs, line 116: np: What about something like `services_config`?

rust/agama-remote/src/service.rs
```diff
@@ -0,0 +1,416 @@
...
+        self.user_config = config;
+        self.update_resolvables();
+        // ignoring error here is ok as it means just dismissed event
+        let _ = self.events.send(Event::ProposalChanged {
+            scope: Scope::RemoteAccess,
+        });
+    }
+
+    /// Sets the configuration requested by a specific module and updates software resolvables.
+    pub fn set_module_config(&mut self, id: String, config: Config) {
```
imobachgs, line 214: Please, use `service` instead of `module`. Just for consistency.

rust/agama-remote/src/service.rs
jreidinger: you are right..original code looks differently and I get move of value...but no longer valid.

rust/agama-manager/src/service.rs
jreidinger: I use `remote_access` to be more consistent.

rust/agama-users/src/service.rs
jreidinger: same here, also `remote_access`

rust/agama-manager/src/service.rs
imobachgs: Well, not really, because the package is called `agama-remote` not `agama-remote-access`. But it is fine.

imobachgs (commented):
LGTM

imobachgs (approved):

PR #3559, opened 2026-05-28
Try boot partitions with desired and min sizes

ancorgs:
## Problem

The logic to calculate boot-related partitions in yast2-storage-ng defines three sizes for each partition: min, desired and max.

When using that in Agama, desired is ignored and only min and max are taken into account. That's something we always wanted to re-evaluate (together with several other aspects related to proposing boot partitions).

When installing with enough available space (let's say in an empty disk or allowing the installer to delete partitions) then there is often no practical difference. The distance between min, desired and max is so small in comparison with the rest of partitions to create that any attempt (starting with either min or desired) always produces the same result. See #3394 for the whole rationale.

But the result is non-optimal when resizing existing partitions just as much as needed to make space for the new installation. In those case Agama is creating partitions with min sizes that where intended for extreme cases, not really as the best practice.

## Solution

This introduces a new approach in which several different sets of boot-related partitions are calculated and tried. In this case, it makes a first attempt using the desired sizes as starting point and, if that fails, a second attempt using the min size as starting point.

This results in more reasonable proposal in most cases, without introducing the risk of failing in edge cases that used to work when using the min sizes.

## Testing

- Added a new unit tests

## Note

This is a "resurrection" of #3394, which was archived for the wrong reasons.

### DIFF
--- a/service/lib/y2storage/agama_proposal.rb
+++ b/service/lib/y2storage/agama_proposal.rb
@@ -134,11 +134,40 @@ module Y2Storage
 
       configure_ptable_types(devicegraph)
       devicegraph = clean_graph(devicegraph)
-      complete_planned(devicegraph)
       return if issues?
 
-      result = create_devices(devicegraph)
+      result = create_devices_with_boot(devicegraph)
       result.devicegraph
+    end
+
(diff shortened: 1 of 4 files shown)


joseivanlopez (approved):
LGTM

PR #3561, opened 2026-05-28
Align suseconnect-agama rmt with recent activation updates

rtamalin:
## Problem

The rmt example was incorrectly overwriting the /etc/SUSEConnect config file with credentials file content.

## Solution

Apply the same changes recently made to the activation example to ensure that correct credentials file path is specified.


jreidinger (approved):
thanks for patch

rtamalin:
Thanks!

PR #3565, opened 2026-06-01
Import statically defined schema into the OpenAPI one

mchf:
## Problem

Some schemas (especially storage related) are lying aside from main openapi schema file.

https://trello.com/c/qy1SB7lM/5700-2-1-add-d-bus-schemas-to-the-openapi-specification

## Solution

Import them on-the-fly into openapi schema

Known issues and notes:
- ~~the code needs some polishing (refactoring some pieces to separate method, deal with unwraps)~~

Currently processed schemas:
- dasd.schema.json -> dasd.Config
- iscsi.schema.json -> iscsi.Config
- storage.model.schema.json -> storage.model.Config
- storage.schema.json -> storage.Config
- zfcp.schema.json -> zfcp.Config

- proposal.storage.schema.json -> storage.Proposal

- system.bootloader.schema.json -> bootloader.SystemInfo
- system.dasd.schema.json -> dasd.SystemInfo
- system.iscsi.schema.json -> iscsi.SystemInfo
- system.storage.schema.json -> storage.SystemInfo
- system.zfcp.schema.json -> zfcp.SystemInfo

- device.storage.schema.json -> storage.Device

## Testing

- *Tested manually*

## Screenshots

currently generated openapi schema is here (validated using https://editor.swagger.io): 
- https://paste.opensuse.org/pastes/4d5ce1e71ad8
- obsolete ~~https://paste.opensuse.org/pastes/52b584338cc0~~
- obsolete ~~https://paste.opensuse.org/pastes/a5a2cd2e0655~~
- obsolete ~~https://paste.opensuse.org/pastes/3d8f9916d86a~~
- obsolete ~~https://paste.opensuse.org/pastes/e448153cde2d~~


mchf:
openapi test fails because I already moved the storage.schema.json into rust/share but did not submit the move into the PR - I need to investigate all places influenced by that change.

joseivanlopez:
Would it be possible to use the same name pattern for all schemas? For example `storage.AdvancedLogicalVolumesGenerator` instead of `storage_advancedLogicalVolumesGenerator`.

rust/agama-lib/share/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "validate": "ajv compile --spec=draft2019 --verbose --all-errors -r storage.schema.json -r iscsi.schema.json -r dasd.schema.json -r zfcp.schema.json -r software.schema.json -s profile.schema.json && ajv compile --spec=draft2019 --verbose --all-errors -s storage.model.schema.json"
+    "validate": "ajv compile --spec=draft2019 --verbose --all-errors -r ../../share/storage.schema.json -r ../../share/iscsi.schema.json -r ../../share/dasd.schema.json -r ../../share/zfcp.schema.json -r software.schema.json -s profile.schema.json && ajv compile --spec=draft2019 --verbose --all-errors -s storage.model.schema.json"
```
joseivanlopez: Maybe we can take the opportunity for moving all the schemas (profile.schema.json, etc). And I guess we can move package.json too.

rust/agama-lib/share/profile.schema.json
```diff
@@ -98,9 +98,9 @@
         }
       ]
     },
-    "dasd": { "$ref": "dasd.schema.json" },
+    "dasd": { "$ref": "../../share/dasd.schema.json" },
```
joseivanlopez: As said, I think it would be better to move everything to the same dir.

rust/agama-lib/share/package.json
mchf: moved

rust/agama-lib/share/profile.schema.json
mchf: moved

rust/agama-server/src/web/docs.rs
```diff
@@ -118,6 +119,45 @@ pub async fn build() -> OpenApi {
...
+    // Example:
+    // storage.schema.json will be stored under /components/schema/storage.Config
+    let schemas_to_import = vec![
+        // (file path, schema name, schema parent)
+        // Config schemas
+        ("share/dasd.schema.json", "dasd", "Config"),
+        ("share/iscsi.schema.json", "iscsi", "Config"),
+        ("share/storage.model.schema.json", "storage.model", "Config"),
+        ("share/storage.schema.json", "storage", "Config"),
+        ("share/software.schema.json", "software", "Config"),
```
joseivanlopez: I guess we can generate the software config schema directly from the rust struct, right? For now, we maintain share/software.schema.json for validation (to use it in profile.schema.json), but I think we don't need to generate openAPI result from it.

joseivanlopez:
Some details I see:

* `legacyAutoyastStorage` should be a list of objects, see https://github.com/agama-project/agama/blob/master/rust/agama-lib/share/profile.schema.json#L861
* We don't follow an uniform criteria for the name of the schemas: `ProductConfig`, `l10n.Config`, `iscsi.target`. I would vote for using cameCase if possible: `productConfig`, `l10Config`, `iscsiTarget`, `l10SystemInfo`, etc. Sorry, in another comment we talked about using `x.Y`, but thinking twice I prefer camelCase if possible.
* Where does `SystemInfo2` come from? I think it should be `softwareSystemInfo`.

joseivanlopez:
The `null` variant is not needed. For example, this

~~~
          "bootloader": {
            "anyOf": [
              {
                "$ref": "#/components/schemas/bootloader.Config"
              },
              {
                "type": "null"
              }
            ]
          }
~~~

can be

~~~
"bootloader": { "$ref": "#/components/schemas/bootloader.Config" }
~~~

A property can be `null` if it is not added as required, see https://json-schema.org/understanding-json-schema/reference/object#required.

mchf:
> * We don't follow an uniform criteria for the name of the schemas: `ProductConfig`, `l10n.Config`, `iscsi.target`. I would vote for using cameCase if possible: `productConfig`, `l10Config`, `iscsiTarget`, `l10SystemInfo`, etc. Sorry, in another comment we talked about using `x.Y`, but thinking twice I prefer camelCase if possible.

ok. Changed, so the prefix doesn't contain '.' anymore (generages storageConfig and so on)

> * Where does `SystemInfo2` come from? I think it should be `softwareSystemInfo`.

it doesn't come from imported schemas. It is present in openapi.json even on the master branch (at least when it was in particular state at the time I generated the example openapi.json). I did some quick search and it seems that it comes from ```agama-utils/src/api/software/system_info.rs```

mchf:
> The `null` variant is not needed. For example, this
> 
> ```
>           "bootloader": {
>             "anyOf": [
>               {
>                 "$ref": "#/components/schemas/bootloader.Config"
>               },
>               {
>                 "type": "null"
>               }
>             ]
>           }
> ```
> 
> can be
> 
> ```
> "bootloader": { "$ref": "#/components/schemas/bootloader.Config" }
> ```
> 
> A property can be `null` if it is not added as required, see https://json-schema.org/understanding-json-schema/reference/object#required.

I copied what we had in openapi.json for some other automatically generated cases. E.g. users or l10n. It is not problem to restrict it - just matter of editing a template.

rust/agama-server/src/web/docs.rs
mchf: ok, i'll remove it

joseivanlopez (commented):
Now it looks much better. I only see these 2 details:

* Some names are not really using camelCase:  e.g., "storageissue" instead of "storageIssue".
* Schemas from rust structs start with capital: e.g., "AccessPoint" instead of "accessPoint".

Is it possible to use the same naming style everywhere? If would prefer "accessPoint", "storageIssue", etc. If the names from rust structs cannot be forced to use camelCase, then I would always use PascalCase: "AccessPoint", "StorageIssue".

mchf:
> Now it looks much better. I only see these 2 details:
> 
> * Some names are not really using camelCase:  e.g., "storageissue" instead of "storageIssue".
> * Schemas from rust structs start with capital: e.g., "AccessPoint" instead of "accessPoint".
> 
> Is it possible to use the same naming style everywhere? If would prefer "accessPoint", "storageIssue", etc. If the names from rust structs cannot be forced to use camelCase, then I would always use PascalCase: "AccessPoint", "StorageIssue".

I'd say that modifying setup which generates the json from the code is out of scope. There is a way how to rename structs using serde (e.g. #serde[rename = "accessPoint"]) but it requires more or less manual intervention in a lot of code. So, it seems to be for another openapi update to me.

I'll try to do something about the second case - make the first letter of word joined to a prefix capitalized (e.g. storageIssues) and may be I'll turn it into StorageIssues as it seems as common pattern used in the code anyway (struct names seems to be in PascalCase - e.g. AccessPoint, but internal fields seems to be in camelCase - e.g. AccessPoint::wpaFlags - because of #serde[rename_all = "camelCase"]

joseivanlopez:
> I'll try to do something about the second case - make the first letter of word joined to a prefix capitalized (e.g. storageIssues) and may be I'll turn it into StorageIssues as it seems as common pattern used in the code anyway (struct names seems to be in PascalCase - e.g. AccessPoint, but internal fields seems to be in camelCase - e.g. AccessPoint::wpaFlags - because of #serde[rename_all = "camelCase"]

Perfect! For me it's fine meanwhile we use the same pattern for everything (rust structs and imported schemas). In the future we can decide if we want/need to change to other pattern.

mchf:
> Perfect! For me it's fine meanwhile we use the same pattern for everything (rust structs and imported schemas). In the future we can decide if we want/need to change to other pattern.

So formatting of names modified into final shape ... we have ```StorageIssue``` and so on

joseivanlopez (approved):
LGTM

joseivanlopez (changes requested):
Oh, I approved but changelog entry is missing.

joseivanlopez (approved):

PR #3571, opened 2026-06-01
fix(web): Enhanced VLAN connection form

teclator:
## Problem

This PR aim to apply some suggestions done during the review of the PBI implemented by #3548.

## Solution

- The VLAN device parent is now a device selector.
- The VLAN device name is shown after the parent device and VLAN ID and suggested based on them.
- The connection types are not translated.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots
Screenshot From 2026-06-03 09-53-48
Screenshot From 2026-06-03 09-54-01
Screenshot From 2026-06-03 09-54-06
Screenshot From 2026-06-03 09-54-14


web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -72,8 +73,75 @@ const VlanFields = withForm({
   render: function Render({ form, isEditing }) {
     const devices = useDevices();
 
+    // Suggests a device name based on the parent device and VLAN ID, as long as
+    // the user has not manually edited it.
+    const suggestIface = (parent?: string, id?: number) => {
+      if (isEditing || form.getFieldMeta("vlanIface")?.isDirty) return;
```
dgdavid: Just to double check it: is that correct? Cannot be _vlanIface_ changed when editing? If so, are you displaying it with ReadOnlyField option instead?

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -72,8 +73,75 @@ const VlanFields = withForm({
...
+      }
+    };
+
+    const parentOptions = devices
+      .filter((d) => d.name !== "lo" && d.name !== form.getFieldValue("vlanIface"))
+      .map((d) => ({
+        value: d.name,
+        label: d.name,
+        description: <Text textStyle={["textColorSubtle", "fontSizeXs"]}>{d.macAddress}</Text>,
+      }));
```
dgdavid: I wonder if it would be possible to update DeviceSelectorField to receive a list of excluding and use such a component instead of recreating almost the same selector here.

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -72,8 +73,75 @@ const VlanFields = withForm({
   render: function Render({ form, isEditing }) {
     const devices = useDevices();
 
+    // Suggests a device name based on the parent device and VLAN ID, as long as
+    // the user has not manually edited it.
+    const suggestIface = (parent?: string, id?: number) => {
+      if (isEditing || form.getFieldMeta("vlanIface")?.isDirty) return;
+
+      const vlanParent = parent ?? form.getFieldValue("vlanParent");
+      const vlanId = id ?? form.getFieldValue("vlanId");
```
dgdavid: We have access to these values here, thanks to form.getFieldValue. So, why passing them as arguments?

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -72,8 +73,75 @@ const VlanFields = withForm({
...
+
+      if (vlanParent && vlanId !== undefined) {
```
dgdavid: I would use !isEmpty(vlanParent) && !isEmpty(vlanId)

web/src/utils/network.ts
```diff
@@ -78,23 +78,22 @@ const isVirtual = (type: ConnectionType): boolean =>
   ).includes(type);
 
 /**
- * Translatable labels for connection types.
+ * Labels for connection types.
```
dgdavid, line 81: Would be nice to clarify why these are not marked for translations.

web/src/components/network/connection-form/VlanFields.tsx
teclator: For bridge a bonding we decided it should be readonly in case of Editing and VLAN followed the same approach although could be questionable ...

teclator: The only problem could be if already set a custom name not using the **parent_device.vlan_id** pattern. Which in case that only the parent device is modified then the device name will be modified too.

web/src/components/network/connection-form/VlanFields.tsx
teclator: I started modifying the DeviceSelector but then moved to specific option for VLAN until generalized, so, lets try to use the DeviceSelector

web/src/components/network/connection-form/DeviceSelector.tsx
```diff
@@ -35,6 +35,31 @@ type SyncConfig = {
...
+  /** Whether to show device names or MAC addresses as the primary value. */
+  by: "iface" | "mac";
+  /** Form field name to bind to. Defaults to "iface" or "ifaceMac" based on `by`. */
+  name?: string;
+  /** Label for the dropdown. Defaults to "Device name" or "MAC address" based on `by`. */
+  label?: React.ReactNode;
+  /** Sync configuration to update another field when a device is selected. */
+  sync?: SyncConfig;
+  /** Device names to exclude from the options list, or an object with devices and types to exclude. */
+  exclude?: string[] | ExcludeFilter;
```
dgdavid: I would go for a single type here. If consumer only want to filter by device name, then provide the { devidces: [] }.  If only interested in type, then { types: [] }. If need both, then uses both.  But let's not offer two differents API for the same thing

dgdavid: Less code to maintain while offering exactly the same features :pray:

dgdavid: BTW, if you feel is more discoverable to go for { names: [], types:[] } instead, go for it. I have no naming objection, as long as they are reasonable.

web/src/components/network/connection-form/Form.tsx
```diff
@@ -432,7 +432,7 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
                         <DeviceSelector
                           form={form}
                           by="mac"
-                          label={_("MAC")}
+                          label={_("MAC address")}
```
dgdavid: Don't we expect the "Device" prefix here as well?

teclator: I was not sure about, I mean, if we add it, it will follow the pattern to have the Device prefix and that makes sense. So, yes, thinking twice about it lets add

web/src/components/network/connection-form/BondFields.tsx
```diff
@@ -111,7 +112,7 @@ const BondFields = withForm({
               }
               helperText={
                 // TRANSLATORS: helper text for the bond ports field. %s is a list of available devices.
-                sprintf(_("Available devices: %s"), formatList(devices.map((d) => d.name)))
+                sprintf(_("Available devices: %s"), formatList(availableDevices.map((d) => d.name)))
```
dgdavid, line 115: This make me think again if users can enter there only available devices. If so, we could cook an improvement for these cases: make the DeviceSelector accept multiple selection. But this, of course, has to wait until other things are more up-to-date or implemented.

teclator: exactly

web/src/components/network/connection-form/DeviceSelector.test.tsx
```diff
@@ -139,4 +142,43 @@ describe("DeviceSelector", () => {
...
+      expect(screen.queryByRole("option", { name: /^enp2s0/ })).not.toBeInTheDocument();
+    });
+  });
+
+  describe("when exclude is provided with both devices and types", () => {
+    beforeEach(() => {
+      exclude = { devices: ["enp1s0"], types: [CONNECTION_TYPE.WIFI] };
+    });
+
+    it("does not show devices matching either filter", async () => {
```
dgdavid, line 184: As far as I can see, test is not really testing wifi type devices are not there.

web/src/components/network/connection-form/BondFields.tsx
teclator: Just didn't wanted to expose that device there, but probably we would need to exclude more devices depending on the types. In this case it does not use the DevicesSelector, them are just part of the description.

web/src/components/network/connection-form/DeviceSelector.tsx
```diff
@@ -48,17 +73,41 @@ type SyncConfig = {
...
+      return true;
+    });
+
     const valueKey = by === "iface" ? "name" : "macAddress";
 
-    const name = by === "iface" ? "iface" : "ifaceMac";
+    // When a custom name is provided, it might not match the form schema exactly
+    // (e.g., "vlanParent" instead of "iface"), so we cast to satisfy TypeScript.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const name = (nameProp ?? (by === "iface" ? "iface" : "ifaceMac")) as any;
```
dgdavid: It is not allowed to use "any", we should find a different solution

dgdavid: BTW, I expect vlanParent being part of the form-schema

web/src/components/network/connection-form/DeviceSelector.tsx
```diff
@@ -48,17 +73,41 @@ type SyncConfig = {
...
     // TRANSLATORS: accessible label for the device selector. MAC address refers
     // to the hardware identifier of the network interface.
-    const label = by === "iface" ? _("Device name") : _("MAC address");
+    const defaultLabel = by === "iface" ? _("Device name") : _("MAC address");
```
dgdavid: As said, I though we were put the Device prefix for Mac address too, like it is the case in all IPvX fields. Maybe "Device MAC" is enough.

web/src/components/network/connection-form/Form.tsx
```diff
@@ -423,14 +423,18 @@ function ConnectionFormContent({ defaults, isEditing = false }: ConnectionFormCo
                         <DeviceSelector
                           form={form}
                           by="iface"
+                          label={_("Device name")}
                           sync={{ field: "ifaceMac", with: (d) => d.macAddress }}
+                          exclude={{ devices: ["lo"] }}
```
dgdavid, line 428: I've seen "exclude lo" so many times. Maybe it should be excluded by default with an "includeLoopback" prop for the few cases you wanna to have it there. Out of the scope of this PR, though.

dgdavid (commented):
Still looking good, but I've added some minor comments

web/src/components/network/connection-form/BondFields.tsx
dgdavid: I know these are part of the helper text, but still thinking is user can select only the available ones, we should provide another kind of field. But, as say, this would be an improvement when times permits. Unless, it is a free text because they can put there whatever, that is what I initially understood.

web/src/components/network/connection-form/DeviceSelector.tsx
teclator: we decided it to be **"Device MAC address"**

dgdavid (approved):
LGTM

PR #3572, opened 2026-06-02
Rename remote

jreidinger:
## Problem

On review it was agreed that remoteAccess name and cockpit is not perfect fit.

## Solution

Rename it to simple "access" and "webConsole" keys.

During rename also changelog was changed retroactive. It would be problem in YaST times, but for agama as it is not yet released I think it is not problem.

## Documentation

https://github.com/agama-project/agama-project.github.io/pull/169


ancorgs (approved):
LGTM. Although right now there are some timeouts in the Rust tests.

PR #3573, opened 2026-06-02
Display the checkmedia result and wait on a key press on all consoles

lslezak:
## Problem

- OpenQA has some troubles with pressing a key to continue when using the media check boot option with both normal console and serial console (when using the `console=tty console=ttyS0` boot options).
- https://bugzilla.suse.com/show_bug.cgi?id=1267263

## Solution

- Display the check media result on all configured consoles
- Wait for pressing a key on any console

## Testing

Tested manually, works fine in all these cases:

- System with a standard system console (no extra boot options provided)
- System with a standard console and a serial console
  - The standard console is the main console (booted with `console=ttyS0 console=tty`)
  - The serial console is the main console (booted with `console=tty console=ttyS0`)

## Screenshots

Tested with VirtualBox with a virtual serial port. The big window is the main VBox VM window, the small window is a virtual serial console attached to a virtual serial port of that VM. Booted with the `console=tty console=ttyS0` options. The checkmedia output is displayed only on the main console (ttyS0) but the result and the prompt is displayed on all consoles. Pressing a key on *any console* continues booting the system.

image

### DIFF
--- a/live/live-root/usr/bin/checkmedia-service
+++ b/live/live-root/usr/bin/checkmedia-service
@@ -1,9 +1,79 @@
-#! /bin/sh
+#! /bin/bash
 
-# Note: the LIVE_MEDIUM_LABEL placeholder is replaced by the real label in config.sh script
+# get list of all system consoles from /proc/consoles
+get_consoles() {
+  if [ -f /proc/consoles ]; then
+    while read -r console _rest; do
+      if [ -c "/dev/$console" ]; then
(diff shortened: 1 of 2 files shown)


lslezak:
@jknphy , any comments?

lslezak:
Discussed with Joaquín on Slack, he is fine with the change.

imobachgs (approved):

PR #3575, opened 2026-06-02
Editorconfig fixes, fixed problems reported by the `shfmt` tool

lslezak:
## Problem

- The shell formatting options are not used in the `web/` and `service/` subdirectories.

## Solution

- Remove `root = true` option from the `.editorconfig` files in subdirectories, that prevents from using the shared rules from the top-level `.editorconfig` file, like the `*.sh` definition.
- Added a special `[[shell]]` section for formatting the shell files using the `shfmt` tool, it can autodetect shell files even without the `*.sh` extension (it checks the hashbang on the first line). See the [shfmt documentation](https://manpages.debian.org/testing/shfmt/shfmt.1.en.html).

## Notes

- Additionally I fixed the most severe issues found by the `shfmt` tool.
- It can be used to reformat the sources into an unified style but to make only the minimal changes before RC I'm postponing that fixes after the release. Mostly there is just inconsistent indentation, extra `;` separators , trailing spaces or similar minor issues. The potentially serious issues are already fixed in this pull request.

## Testing

- Tested manually, the shell formatting options are correctly applied also in the `web/` and `service/` subdirectories.

## TODO

- Later I'd like to run the `shfmt` check in the CI, but again, after the release...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,19 @@ root = true
 [*.sh]
 indent_style = space
 indent_size = 2
+trim_trailing_whitespace = true
+
+# A special non-standard configuration for the `shfmt` tool which matches any
+# file with a shell hashbang on the first line, it does not need the *.sh
+# extension in the file name (needs shfmt >= 3.8.0).
+#
+# To reformat all files run `shfmt -w -l .` at the top-level directory.
+[[shell]]
+indent_style = space
+indent_size = 2
+switch_case_indent = true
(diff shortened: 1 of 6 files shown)


mvidner (approved):
Oh nice, thanks! I want to learn more about `shfmt`. Standard formatting is good.

LGTM

Maybe make a draft PR for the shfmt CI now, so that we don't forget?

PR #3576, opened 2026-06-02
Translatable CLI help

mvidner:
## Problem

The help for the `agama` command-line interface (collected in the [CLI reference](https://agama-project.github.io/docs/user/reference/cli)) is not translated.

The solution is not as simple as wrapping the strings in `gettext` calls because the strings are written as _comments_ in the Rust source code (and collected with the _derive_ feature of the `clap` package)

- https://trello.com/c/8qXxSYVq/5790-3-cli-make-help-translatable

## Solution

**POT extraction**: We already have _markdown_ and _manpages_ xtasks (`cargo xtask`) that extract the CLI structure and present it in their ways. So add a **cli-strings** xtask that makes a dummy Rust file which is then seen  by the existing `rust/build_pot` script.
- Added a build.rs script to automatically call that xtask when a relevant source file changes. The resulting `cli_strings.rs` is committed to git so that the weblate-update-pot.yml CI task does not need a Rust toolchain.

**Runtime translation**: Replace the `Cli::parse()`  shorthand with a more explicit routine that includes `translate_command`, which iterates over the CLI data structures and calls `gettext` on them.
- Add logic to reduce translator work: For example, for `agama probe` the short help is _"Analyze the system"_ and the long help is _"Analyze the system.\n\nIn agama's jargon..."_. We compare the short and long helps and call `gettext` on them separately so that translators do not have to repeat themselves.

**`agama-web-server serve`** also has some short CLI help but I consider it an implementation detail, not worth translating.

The **clap built-in** messages like "Usage:" are not translated :-/

NOTE: Clap help strings are of type StyledStr which means that they can use terminal beautifiers (colors, bold...) and by translating them we throw the **styling** away. I did not investigate whether it is possible to bring the styling back.

## Testing

- Added a new unit test (see bottom of [rust/agama-cli/src/translate.rs](https://github.com/agama-project/agama/pull/3576/changes#diff-ae8987f5ddb1fe0064af080f246ca71a19f88a8c95967929a85d267b44dbf404R147))
- Tested manually that it works at runtime (see the description of 50204d610f7db65a346ccd6d0910a908404f1fa0) See "screenshot" below.
- In my agama-weblate fork, the strings are added, yay! https://github.com/mvidner/agama-weblate/commit/e58deabccdb5b8ad5ae36a22b1e82f373fa66c46#diff-2f02cf319a0c4f687830d1de63a36b5a5218f3a4719909f3c58a0eade5713c4c

## Screenshots

```console
$ LANG=cs_CZ ./target/debug/agama |& head -n3
Agama rozhraní příkazové řádky

Usage: agama [OPTIONS] <COMMAND>
$ LANG=cs_CZ ./target/debug/agama --help |& head -n8
Agama rozhraní příkazové řádky

Tento program umožňuje prohlížet nebo měnit nastavení programu Agama, spravovat instalační profily,
spouštět instalaci, sledovat průběh instalace atd.

Další informace získáte pomocí příkazu "help".

Usage: agama [OPTIONS] <COMMAND>
```

## Documentation

- [x] rust/package/agama.changes
- schema not affected
- funnily, the CLI docs need not be changed
- no new documentation needed IMHO, it all ties into existing automation

### DIFF
--- a/.github/workflows/weblate-update-pot.yml
+++ b/.github/workflows/weblate-update-pot.yml
@@ -10,7 +10,7 @@ on:
       # run on every change in master, the action actually checks if there is
       # any change in the translations, if there is no change it does not upload
       # the files to Weblate so it should be OK to run it quite often
-      - master
+      - translatable-cli-help
 
     paths:
       # run only if any file which can contain translatable strings is modified,
@@ -71,7 +71,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           path: agama-weblate
-          repository: ${{ github.repository_owner }}/agama-weblate
+          repository: mvidner/agama-weblate
           token:  ${{ secrets.GH_TOKEN }}
 
       - name: Configure Git user and email
--- a/.github/workflows/weblate-update-pot.yml
+++ b/.github/workflows/weblate-update-pot.yml
@@ -10,7 +10,7 @@ on:
       # run on every change in master, the action actually checks if there is
       # any change in the translations, if there is no change it does not upload
       # the files to Weblate so it should be OK to run it quite often
-      - translatable-cli-help
+      - master
 
     paths:
       # run only if any file which can contain translatable strings is modified,
@@ -71,7 +71,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           path: agama-weblate
-          repository: mvidner/agama-weblate
+          repository: ${{ github.repository_owner }}/agama-weblate
           token:  ${{ secrets.GH_TOKEN }}
 
       - name: Configure Git user and email
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -61,6 +61,7 @@ dependencies = [
 name = "agama-cli"
 version = "1.0.0"
 dependencies = [
+ "agama-l10n",
  "agama-lib",
  "agama-transfer",
  "agama-utils",
@@ -73,6 +74,7 @@ dependencies = [
  "futures-util",
  "gettext-rs",
  "home",
+ "i18n-format",
  "inquire",
  "ratatui",
  "regex",
--- a/rust/agama-cli/Cargo.toml
+++ b/rust/agama-cli/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 clap = { version = "4.5.19", features = ["derive", "wrap_help"] }
+agama-l10n = { path = "../agama-l10n" }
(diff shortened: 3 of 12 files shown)


imobachgs:
Thanks for working on this. It is perhaps a little bit late for this comment but, has you considered migrating our clap related code to a [builder-based approach](https://docs.rs/clap/latest/clap/_tutorial/index.html#tutorial-for-the-builder-api)? AFAIK, you can use `gettext` as usual in that case, so you do not need to extract the strings to a file. We should be able to use some AI for the conversion.

Just my cents.

mvidner:
> have you considered migrating our clap related code to a [builder-based approach](https://docs.rs/clap/latest/clap/_tutorial/index.html#tutorial-for-the-builder-api)?
> AFAIK, you can use `gettext` as usual in that case, so you do not need to extract the strings to a file.

Yes, but I considered that the resulting Builder code would be more verbose and harder to read. So I was happy that Gemini suggested an approach that allowed to keep the Derive approach. Yes it needs some extra code (= all the code in this PR) but I think that this cost is overcome by the benefit of keeping the CLI declaration concise and readable.

> We should be able to use some AI for the conversion.

But I will ask AI for the Builder approach and compare the results.

imobachgs:
> Yes, but I considered that the resulting Builder code would be more verbose and harder to read. So I was happy that Gemini suggested an approach that allowed to keep the Derive approach. Yes it needs some extra code (= all the code in this PR) but I think that this cost is overcome by the benefit of keeping the CLI declaration concise and readable.

Well, the builder code looks quite readable to me, but of course is not straightforward like the derived one. And it saves us to have to maintain another piece of code to extract the strings.

> > We should be able to use some AI for the conversion.
> 
> But I will ask AI for the Builder approach and compare the results.

Thank you! Let's see how it looks like.

mvidner:
@imobachgs JFYI when I explicitly asked Gemini to propose a Builder approach, it said "OK, it would go like this and this... and BTW the runtime usage ends up more verbose and less type-safe so the community tends to go with (and described the approach of this PR)". I told it to use Builder and right now it is working through the massive refactoring :)

imobachgs:
Perhaps we could put Claude and Gemini to talk to each other :-) Claude recommended me to go for the builder pattern, although I have not asked it for an implementation. Actually, Claude toll me that there was an attempt to add i18n support [in 2024](https://github.com/clap-rs/clap/pull/5853). Let's see what is the outcome or your expermient with Gemini.

mvidner:
Closed in favor of #3594

PR #3531, opened 2026-05-21
refactor(web): unify authentication forms and complete SSH Public Keys support

dgdavid:
## Summary

What began as adding SSH public key support to the first-user workflow evolved into a broader a
uthentication form refactor. Supporting SSH keys cleanly required migrating both authentication forms to TanStack Form first, and once that work was underway, consolidating the flows into a single page became the most coherent UX direction.

The result is a unified authentication experience that reduces navigation overhead, improves consistency, and simplifies both validation and configuration handling.

Screen Shot 2026-05-25 at 10 37 32

---

Screen Shot 2026-05-25 at 10 38 38

### Unified authentication form (`authentication-form/`)

The previous `FirstUserForm`, `RootUserForm`, and `UsersPage` flows (including their routes) have been consolidated into a single `AuthenticationForm` page.

Both user sections are now visible and editable in one place, eliminating unnecessary navigation and making related authentication settings easier to manage together.

### SSH public key support for the first user

The first-user section now supports SSH public keys through an `ArrayField` implementation with several SSH-specific improvements:

- Newline-aware paste handling via `splitPasteOn=/\r?\n/`, allowing multi-line key blocks to paste correctly
- Middle truncation of long key entries through `maxEntryWidth=60` for improved readability
- SSH key validation that rejects malformed entries and private keys

Screen Shot 2026-05-25 at 10 41 01

### `ArrayField` enhancements

`ArrayField` was extended to support more flexible entry handling and better usability:

- Added `splitPasteOn` prop to customize paste splitting behavior
  - Defaults to whitespace/comma splitting
  - Supports newline-only splitting for values that legitimately contain spaces, such as SSH public keys
- Added `maxEntryWidth` prop to truncate long entries using PatternFly’s `Truncate` component while preserving full values in accessibility labels
- Added defensive null checks around `value` handling for edge-case stability

#### `PreservedValueField`

Extracts the preserve/edit toggle behavior used for hashed passwords into a reusable component shared across both authentication sections.

#### `SuggestionsTextField`

Introduces a `TextInput` backed by a native HTML `datalist`, integrated with TanStack Form through `useFieldContext`.

The component is registered through `useAppForm` as `field.SuggestionsTextField`, restoring functionality that had previously been reverted at https://github.com/agama-project/agama/pull/3490

### Root authentication mode redesign

Root authentication is now configured through a single `AuthMode` dropdown:

- None
- Password
- SSH Public Key
- Both

[...]

PR #3578, opened 2026-06-02
fix(web): add redirects for refactored authentication and system routes

dgdavid:
## Summary

This PR introduces redirects for legacy routes that were consolidated in recent refactoring (_System_ in #3482, _Authentication_ in #3531). Without these redirects, accessing the old URLs would result in 404 errors. While direct access to these URLs is not a common scenario at this stage of development, maintaining redirects improves robustness, preserves user experience, and follows good routing practices.

## Changes

Uses `<Navigate>` with `replace`  option to avoid polluting browser history, ensuring that the back button behaves intuitively after a redirect. 

- /hostname -> /system  (`HostnamePage` merged into `SystemPage`)
- /users/first, /users/first/edit, /users/root/edit ->  `/users` (`FirstUserForm` and `RootUserForm` unified into `AuthenticationForm`)

## Future Considerations

- This pattern should be reused for any future route change/consolidation.
- After certain time, these redirects can be safely removed.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jun  2 17:45:07 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Add redirections for legacy routes to preserve backward
+  compatibility after route refactors (gh#agama-project/agama#3578).
+
+-------------------------------------------------------------------
 Mon Jun  1 13:51:58 UTC 2026 - José Iván López González <jlopez@suse.com>
 
 - Create new LVM volume groups using all the availabe space on the
--- a/web/src/router.tsx
+++ b/web/src/router.tsx
@@ -21,7 +21,7 @@
  */
 
 import React from "react";
-import { createHashRouter, Outlet } from "react-router";
+import { createHashRouter, Navigate, Outlet } from "react-router";
 import App from "~/App";
 import Protected from "~/Protected";
 import ErrorPage from "~/components/core/ErrorPage";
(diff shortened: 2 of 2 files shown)


imobachgs (commented):
I am not against but I have a question: how do we know when we can drop them? I mean, if a user has a bookmark, how does we he/she know that the bookmark must be outdated?

dgdavid:
> I am not against but I have a question: how do we know when we can drop them? I mean, if a user has a bookmark, how does we he/she know that the bookmark must be outdated?

A good question indeed.

Honestly, I simply wanted to keep that door open just in case the feature/practice will face resistance. I have no problem keeping them there forever and ever, but a reasonable window time to drop the redirections could be between versions of SLE (I'm thinking from 16.1 to  16.2, but could be from 16.x to 17.x). I'm not too worried with that at this time, but more in define the mechanism of keeping navigation safe just in case we have to change more complex routes in the future.

Another topic is how to announce some routes are no longer available, which I would not talk about in this phase.

imobachgs (approved):
I am not convinced about the benefits right now, but I understand that it is a good thing to have it defined for the future in case it is needed.

dgdavid:
> I am not convinced about the benefits right now, but I understand that it is a good thing to have it defined for the future in case it is needed.

This is the exact point where I am too.

If you wish, we can postpone this for the time we feel is more needed, though.

dgdavid:
> > I am not convinced about the benefits right now, but I understand that it is a good thing to have it defined for the future in case it is needed.
> 
> This is the exact point where I am too.
> 
> If you wish, we can postpone this for the time we feel is more needed, though.

On the other hand, it's true that these few lines of code does not hurt.

dgdavid:
Ok, merging it as kind of _documentation of what to do if a route has to change_ in the future. 

Thanks!

PR #3580, opened 2026-06-03
feat(web): improve feedback for download logs action

dgdavid:
## Problem

Triggering a log download from the installer options menu looked like nothing happened. There was no feedback until the browser eventually started the download, which could take a while since log collection happens server-side.

## Solution

Introduced a `download` utility and a `DownloadFeedback` component that wrap any downloadable trigger with toast alert feedback:

- An **info** alert is shown immediately when the download is requested, warning the user it may take a while.
- A **success** alert replaces it once the browser starts the download, including the generated filename so the user knows what to look for in their downloads folder.
- The success alert auto-dismisses after a configurable timeout. Both alerts are manually dismissible.
- The filename includes a timestamp generated at click time (e.g. `agama-logs-2024-01-15T10-30-00-000Z.tar.gz`), reflecting when collection started rather than when it finished.

`InstallerOptionsMenu` adopts `DownloadFeedback` for the "Download logs" entry. The component wraps the entire menu rather than just the dropdown item, as PatternFly unmounts dropdown children on close. Wrapping only the item would cause the pending alert to vanish as soon as the menu closes.

### Limitations

The success alert is tied to the lifecycle of `DownloadFeedback`. If the user navigates to another page while the download is being prepared, the component unmounts and the eventual success alert is lost. The download itself completes normally, but the user gets no confirmation feedback.

Addressing this properly requires alert state that survives navigation, which in practice means either a global UI state store or a dedicated notification context mounted at the app root. The project does not have either of these yet and evaluating the tradeoffs is out of scope for this PR.

As a short-term solution, the success alert could be omitted until a global UI state solution is in place, only if some kind of consistence between doing nothing and navigating elsewhere is desired.  `DownloadFeedback` already accepts a `successTimeout` prop and can be straightforwardly extended with a flag to opt out of the success alert entirely, making it easy to enable later once the infrastructure exists.

## Testing

- Added unit tests for `download` utility and `DownloadFeedback` component.
- Tested manually.

## Screenshots

Screenshot_2026-06-03_12-50-43

Screenshot_2026-06-03_12-53-02


ancorgs:
Maybe we can take the opportunity to add a sentence explaining what the logs are. Similar to what #3581 does for the config.

dgdavid:
Due to limitation commented in the PR description, in a short meeting we have decided to:

* Use only the info toast, having there all the information (@ancorgs will take care).
* Drop the automatic dismissing of the info toast., to give users chance to read all the information.

The success toast could be introduced at some point when the ui has the corresponding mechanism to keep it alive even after users navigation.

ancorgs:
I added a commit slightly modifying the look and the behavior of the toast.

Now it initially looks like this:

download1

And it turns into this when the download starts but only if the previous one has not been closed:

download2

web/package/agama-web-ui.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Jun  4 14:26:23 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Improve the installation settings UI by showing and explaining the
```
dgdavid, line 4: Since we forgot this in its corresponding PR, #3581, I'm adding it here.

dgdavid (commented):
@ancorgs changes LGTM, but I can't approve the PR since I'm the author.

ancorgs (approved):
LGTM

PR #3372, opened 2026-04-10
Change the save dir for user-defined scripts.

ldzhong:
## Problem

The scripts cannot be executed from /var/log/agama-installation on a hardened system.

- [*bsc1261787*](https://bugzilla.suse.com/show_bug.cgi?id=1261787)
- https://github.com/agama-project/agama/pull/3371

## Solution

change /var/log/agama-installation to /var/lib/agama/

## Testing

- Adapted the unit test.

PR #3583, opened 2026-06-04
Support the old location in the `agama-scripts` service (SLE-16)

imobachgs:
- Fix the `agama-scripts.sh` to check the new location first and fall back to the old location for backward compatibility
- Addresses the issue where #3372 only changed to the new location without providing a fallback

The script now:

1. Respects SCRIPTS_DIR if set from the environment
2. Tries /var/lib/agama/scripts/init (new location) first
3. Falls back to /var/log/agama-installation/scripts/init (old location)
4. Defaults to the new location if neither exists

This ensures backward compatibility with the Agama version included in SLES 16.0 and Leap 16.0.


joseivanlopez (approved):
LGTM

PR #3577, opened 2026-06-02
feat(rust): task-based configuration system for agama-manager

joseivanlopez:
## Summary

- https://trello.com/c/iJhKbkyJ/5768-8-1266276-sles-161-unattended-doesnt-start
  - based on https://bugzilla.suse.com/show_bug.cgi?id=1266276

This PR introduces a task-based execution system for the configuration phase in agama-manager. Instead of running configuration steps sequentially, the new implementation uses a `TaskManager` to execute steps asynchronously with explicit dependency tracking.

## Motivation

The previous sequential implementation of `SetConfigAction::run()` had several limitations:

- No visibility into which configuration steps were running
- No ability to express dependencies between steps
- Configuration steps blocked even when they could run in parallel
- No structured way to track task progress and status
- Difficult to detect when the configuration actually finished (see [bsc#1266276](https://bugzilla.suse.com/show_bug.cgi?id=1266276)).

## The `task_manager` module

- New `TaskManager` for coordinating asynchronous task execution
- `TaskBuilder` with fluent API for defining tasks and dependencies
- `TaskError` newtype for type-safe error handling
- `TaskMetadata` for tracking task state (name, scope, description)
- Event emission system for task lifecycle (started, finished, failed)
- Support for task dependencies via `depends_on(&[TaskId])`

### SetConfigAction Refactoring

**Async Task-Based Execution**
- Migrated from sequential `run()` to task-based asynchronous execution
- Each configuration step now runs as an independent task
- Tasks can execute in parallel when dependencies allow

**Task Dependencies**
- Software configuration depends on: files, storage, bootloader, and NTP
- Storage configuration depends on: S390 (when available)
- Pre-installation scripts depend on: files configuration
- SELinux configuration integrated into software task (executes after SetConfig completes)

## Tasks

- [x] Add the task manager
- [x] Add dependencies between tasks
- [x] Add the scope to the tasks
- [x] Expose the tasks through the ~/api/status~ API.
- [x] Emit events
- [x] Add new event to notify that a task was created
- [x] Add a new event to notify that the configuration was finished (with an ID)
- [x] Adapt the web UI to track tasks
- [x] Adapt the monitor to track tasks
- [x] Adapt the `agama-autoinstall` service
- [ ] Code clean-up

## Follow-up

There are a bunch of changes that we would like to do once this PR is tested and merged:

- Get rid of the `TaskRunner` (not needed anymore).
- Remove the `run_in_background` helper.
- Make the manager responsible for retrieving and setting the resolvables
[...]

PR #3585, opened 2026-06-04
Set the software resolvables from the manager service

imobachgs:
This is a follow-up of #3577.

## Problem

Although #3577 improved the workflow of the `SetConfig` action, it does not fix [bsc#1266276](https://bugzilla.suse.com/show_bug.cgi?id=1266276) in all cases. It might happen that, for instance, the `bootloader` service asks the manager to add some packages/patterns (resolvables) for installation. The problem is that there is no guarantee that it happens within the `SetConfig` time frame.

## Solution

The manager is now responsible for setting the list of resolvables. It asks each module about its requirements and tells the software service about them.

## Benefits

- Only the `agama-manager` package depends on `agama-software`.
- More predictable workflow.
- The solver runs once for each `SetConfig` action. In the past, it ran several times.


rust/agama-storage-client/src/message/bootloader.rs
```diff
@@ -18,7 +18,10 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
-use agama_utils::{actor::Message, api::bootloader};
+use agama_utils::{
+    actor::Message,
+    api::{bootloader, software::Resolvable},
```
joseivanlopez, line 23: We should extract `Resolvable` type from the api. But it can be done in the follow-up.

joseivanlopez (approved):
LGTM

PR #3587, opened 2026-06-05
[SLE-16.0] Do not loose the repository password (bsc#1258701)

lslezak:
## Problem

- A custom password-protected installation repository cannot be used, libzypp fails to log in to the remote server.
- https://bugzilla.suse.com/show_bug.cgi?id=1258701

## Solution

- When creating the `Repository` object in Ruby the password in the URL is lost. Using that URL later causes failures.

## Testing

- Tested manually
- The system can be properly installed from a password protected FTP server
- The password is also correctly saved into the target system, the password-protected repository can be accessed also in the installed system

### DIFF
--- a/service/lib/agama/software/repository.rb
+++ b/service/lib/agama/software/repository.rb
@@ -60,7 +60,13 @@ module Agama
             "priority" => priority
           )
 
-          repo_id ? find(repo_id) : nil
+          return nil unless repo_id
+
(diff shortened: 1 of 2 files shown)


ancorgs (approved):
LGTM

mvidner:
@lslezak please share some howto for setting up the testing FTP server

PR #3590, opened 2026-06-05
[SLE-16.0] Added `plymouth.ignore-serial-consoles` boot option on ppc64le

lslezak:
## Problem

- Unable to login if both plymouth and gdm are installed
- https://bugzilla.suse.com/show_bug.cgi?id=1249569

## Solution

- It was suggested to add the `plymouth.ignore-serial-consoles` boot option on Power

## Testing

- Unfortunately not tested because of lack of hardware

### DIFF
--- a/live/config-cdroot/fix_bootconfig.ppc64le
+++ b/live/config-cdroot/fix_bootconfig.ppc64le
@@ -97,7 +97,7 @@ fi
 
 menuentry "Install $label" --class os --unrestricted {
   echo 'Loading kernel...'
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

ancorgs (approved):
Theoretically, in the rescue system that is not needed since we are only adding it to the installer with the purpose of getting it copied over to the target system.

Anyway, it generally LGTM (assuming the argument is indeed persisted to the installed system which I trust it is).

PR #3592, opened 2026-06-05
Improve tasks management

joseivanlopez:
- [ ] Get rid of `run_in_backgroup`
- [ ] Drop `TaskRunner`
- [x] Add the correct scope for DASD and zFCP tasks
- [x] Remove DASD thread in the D-Bus service
- [ ] Extract `Resolvable` type from api
- [ ] Add task for actions like ProbeStorage
- [ ] Improve tasks manager (only run tasks and emit events)


joseivanlopez:
Close in favor of https://github.com/agama-project/agama/pull/3602

PR #3594, opened 2026-06-05
Translatable CLI help: with Clap::Builder

mvidner:
- https://trello.com/c/8qXxSYVq/5790-3-cli-make-help-translatable

This is the second attempt, with clap::Builder. For the first one, see #3576 (with clap::Derive)

## Problem

The help for the `agama` command-line interface (collected in the [CLI reference](https://agama-project.github.io/docs/user/reference/cli)) is not translated.

The solution is not as simple as wrapping the strings in `gettext` calls because the strings are written as _comments_ in the Rust source code (and collected with the _derive_ feature of the `clap` package)

## Solution

Use a LLM (gemini-3.1-pro-preview) to translate our Derive usage into the Builder API. (Then watch it fail hopelessly as it tries to untangle the correct usage of doubly nested "\n" and "\\n" escaping... fix that manually)

Using Builder, the declarations and parsing looks a bit uglier, but IMHO bearable, and there is no additional helper code. 

**`agama-web-server serve`** also has some short CLI help but I consider it an implementation detail, not worth translating.

The **clap built-in** messages like "Usage:" are not translated :-/

NOTE: Clap help strings are of type StyledStr which means that they can use terminal beautifiers (colors, bold...) and by translating them we throw the **styling** away. I did not investigate whether it is possible to bring the styling back.

## Testing

Tested manually:

- [x]  compared the outputs of "cargo xtask markdown, completions, manpages" between Derive and Builder.
- [x] In my agama-weblate fork, the strings are added, yay! https://github.com/mvidner/agama-weblate/commit/35308a8d9c2c7b1e2da8feb0b415e24f5d44aaa3

## Screenshots

```console
$ LANG=cs_CZ ./target/debug/agama |& head -n3
Agama rozhraní příkazové řádky

Usage: agama [OPTIONS] <COMMAND>
$ LANG=cs_CZ ./target/debug/agama --help |& head -n8
Agama rozhraní příkazové řádky

Tento program umožňuje prohlížet nebo měnit nastavení programu Agama, spravovat instalační profily,
spouštět instalaci, sledovat průběh instalace atd.

Další informace získáte pomocí příkazu "help".

Usage: agama [OPTIONS] <COMMAND>
```

## Documentation

- [x] rust/package/agama.changes
- schema not affected
- [x] the CLI docs changes only slightly, see https://github.com/agama-project/agama-project.github.io/pull/172
- no new documentation needed IMHO, it all ties into existing automation


.github/workflows/weblate-update-pot.yml
```diff
@@ -10,7 +10,7 @@ on:
       # run on every change in master, the action actually checks if there is
       # any change in the translations, if there is no change it does not upload
       # the files to Weblate so it should be OK to run it quite often
-      - master
```
imobachgs, line 13: A left-over?

.github/workflows/weblate-update-pot.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           path: agama-weblate
-          repository: ${{ github.repository_owner }}/agama-weblate
```
imobachgs, line 74: Left-over?

rust/agama-cli/src/auth.rs
```diff
@@ -57,28 +60,60 @@ impl AuthHTTPClient {
...
+pub fn build_auth_cmd() -> Command {
+    // TRANSLATORS: CLI help for: agama auth
+    let about = gettext("Authenticate with Agama's server");
+    Command::new("auth")
+        .subcommand_required(true)
+        .arg_required_else_help(true)
+        .about(&about)
+        // TRANSLATORS: CLI help for: agama auth (details)
+        .long_about(make_long(&about, &gettext("\
+                             Unless you are executing this program as root, you need to authenticate with Agama's server \
```
imobachgs: You can use a "raw string" if you do not want to escape quotes and so on. But you would loose the indentation:

```rust
        .long_about(make_long(&about, &gettext(r#""
Unless you are executing this program as root, you need to authenticate with Agama's server
for most operations. You can log in by specifying the root password through the "auth login"
command. Upon successful authentication, the server returns a JSON Web Token (JWT) which is
stored to authenticate the following requests.

If you run this program as root, you can skip the authentication step because it
automatically uses the master token at /run/agama/token. Only the root user must have access
to such a file.

You can logout at any time by using the "auth logout" command, although this command does
not affect the root user.""#)))
```

Alternatively, we could use consts (see [my review comment](https://github.com/agama-project/agama/pull/3594#pullrequestreview-4459604810)).

imobachgs (commented):
I have not finished the review yet, but I was wondering if we could use constants for long texts. For instance:

```rust
const HELP: &str = gettext_noop(
    r#"
Some help

... with multiple lines
"#,
);
```

Of course, we need to mark the `gettext_noop` function as const:

```gettext_noop
pub const fn gettext_noop(text: &str) -> &str {
    text
}
```

Would it help to make the code more readable?

mvidner:
@imobachgs Please see https://github.com/agama-project/agama/pull/3594/changes/7e6d0a402004a3796be40b0f7e48d2895a308526 for my take on the readability with the long strings: now they are
- indented only 8 spaces
- not breaking the flow of other `Command` declarations

That commit  is only for `agama auth`. Should I apply it to the other commands?

imobachgs:
> @imobachgs Please see [7e6d0a4](https://github.com/agama-project/agama/commit/7e6d0a402004a3796be40b0f7e48d2895a308526) for my take on the readability with the long strings: now they are
> 
> * indented only 8 spaces
> * not breaking the flow of other `Command` declarations
> 
> That commit is only for `agama auth`. Should I apply it to the other commands?

It looks good. Yes, please, you can apply the same approach to other commands.

Thanks!

rust/agama-cli/src/auth.rs
```diff
@@ -57,28 +60,60 @@ impl AuthHTTPClient {
...
+    // TRANSLATORS: CLI help for: agama auth
+    let about = gettext("Authenticate with Agama's server");
+    // TRANSLATORS: CLI help for: agama auth (details)
+    let long_about = make_long(&about, &gettext("\
+        Unless you are executing this program as root, you need to authenticate with Agama's server \
+        for most operations. You can log in by specifying the root password through the \"auth login\" \
+        command. Upon successful authentication, the server returns a JSON Web Token (JWT) which is \
+        stored to authenticate the following requests.\n\
+        \n\
+        If you run this program as root, you can skip the authentication step because it \
```
jreidinger: I would mention run this program **locally** as root to avoid confusion.

rust/agama-cli/src/auth.rs
```diff
@@ -57,28 +60,60 @@ impl AuthHTTPClient {
...
 
-    match subcommand {
-        AuthCommands::Login => login(auth_client, read_password()?).await,
-        AuthCommands::Logout => logout(auth_client),
-        AuthCommands::Show => show(&auth_client.api.base_url),
+    match sub_matches.subcommand() {
+        Some(("login", _)) => login(auth_client, read_password()?).await,
+        Some(("logout", _)) => logout(auth_client),
+        Some(("show", _)) => show(&auth_client.api.base_url),
+        _ => Ok(()),
```
jreidinger, line 116: will it behave same as before that it writes help if you do not specify subcommand?

rust/agama-cli/src/commands.rs
```diff
@@ -35,114 +33,135 @@ pub enum Format {
...
-    },
+pub fn build_finish_cmd() -> Command {
+    Command::new("finish")
+        // TRANSLATORS: CLI help for: agama finish
+        .about(gettext("Finish the installation"))
+        .arg(
+            Arg::new("method")
+                .value_name("METHOD")
+                .value_parser(value_parser!(FinishMethod))
+                // TRANSLATORS: CLI help for: agama finish <METHOD>
```
jreidinger, line 117: I would probably also mention here for translators to not mess up with `\x20`. From experience I would prefer to not there formatting at all and split it into smaller paragraphs and do formatting outside of gettext so translators cannot mess it up.

jreidinger: also with this manual wrapping you keep it to translators to make lines reasonable long, which can be problem.

rust/agama-cli/src/auth.rs
mvidner: Yes, because we use these calls, and I have verified that it works:
https://docs.rs/clap/latest/clap/struct.Command.html#method.subcommand_required
https://docs.rs/clap/latest/clap/struct.Command.html#method.arg_required_else_help

rust/agama-cli/src/commands.rs
mvidner: Good points but fortunately you are wrong on both counts :-)
1. `\x20` is a [string literal syntax][x]. Both rustc and xgettext will put a plain space to the string. I used it to preserve line leading space in a context where leading spaces are stripped: after [a `\` line continuation][bs].
Which brings me to your second point:
2. Yes the lines are long, but it is fine, _clap_ takes care of line wrapping so translators need not wrap them.

```console
$ grep -A5 METHOD agama.pot
#. TRANSLATORS: CLI help for: agama finish <METHOD>
#: agama-cli/src/commands.rs:118
msgid ""
"What to do after finishing the installation. Possible values:\n"
"\n"
"   stop - do not reboot and the Agama backend continues running.\n"
```

[x]: https://doc.rust-lang.org/reference/expressions/literal-expr.html#8-bit-escapes
[bs]: https://doc.rust-lang.org/reference/expressions/literal-expr.html#string-continuation-escapes

rust/agama-cli/src/auth.rs
mvidner: Umm... okay... but I consider it a minor point and would rather not change the string now :)

jreidinger: well, if we do not change it now before string freeze, next opportunity will be in 16.2

imobachgs: I agree with @jreidinger here. You can adjust the text in a following PR if you prefer to close this one. But it would be good to have it fixed for 16.1.

imobachgs (approved):

PR #3574, opened 2026-06-02
refactor(web): migrate and redesign storage partition form

dgdavid:
[Back in April](https://github.com/agama-project/agama/pull/3300), TanStack Form was introduced as a strategic dependency to improve form management across the project. Following the successful migration and redesign of the network connection form, several other forms were migrated as a preparatory exercise before tackling the storage forms, arguably the most complex area of the Agama Web UI codebase.

This effort goes beyond a straightforward framework migration. It also provides an opportunity to revisit aspects of the existing forms that have proven to need further refinement, both from a maintenance perspective and in terms of usability and user experience. It is also an opportunity to extract reusable patterns and components that can be shared across similar storage-related forms.

Although fairly large in scope, this PR focuses on a single goal: migrating and redesigning the form to create and edit a partition as the first step toward modernizing the storage forms. Besides improving the form itself, the work begins assembling the reusable pieces and conventions that will support the future migration of other storage-related forms such as `LogicalVolumePage` and `FormattableDevicePage`. .

Mockups at https://gist.github.com/dgdavid/0e2cb4636997e0f537e534741ec0eb63#file-l-third-iteration-mockups-md

---

:hourglass:  Description to be completed. Meanwhile, you can check the plan originally written to make this AI assisted, https://gist.github.com/dgdavid/0e2cb4636997e0f537e534741ec0eb63

PR #3603, opened 2026-06-08
First pull request for the new feature branch for storage forms

ancorgs:
This pull request substitutes #3574 by rebasing the content and re-setting the target to a new feature branch.

## Description of the original pull request

Back in April](https://github.com/agama-project/agama/pull/3300), TanStack Form was introduced as a strategic dependency to improve form management across the project. Following the successful migration and redesign of the network connection form, several other forms were migrated as a preparatory exercise before tackling the storage forms, arguably the most complex area of the Agama Web UI codebase.

This effort goes beyond a straightforward framework migration. It also provides an opportunity to revisit aspects of the existing forms that have proven to need further refinement, both from a maintenance perspective and in terms of usability and user experience. It is also an opportunity to extract reusable patterns and components that can be shared across similar storage-related forms.

Although fairly large in scope, this PR focuses on a single goal: migrating and redesigning the form to create and edit a partition as the first step toward modernizing the storage forms. Besides improving the form itself, the work begins assembling the reusable pieces and conventions that will support the future migration of other storage-related forms such as `LogicalVolumePage` and `FormattableDevicePage`. .

Mockups at https://gist.github.com/dgdavid/0e2cb4636997e0f537e534741ec0eb63#file-l-third-iteration-mockups-md

---

:hourglass:  Description to be completed. Meanwhile, you can check the plan originally written to make this AI assisted, https://gist.github.com/dgdavid/0e2cb4636997e0f537e534741ec0eb63

## Extras

Apart from the original content of #3574, this includes:

- A rebase on top of the new feature branch (just branched from master)
- A couple of fixes in the form behavior
- Quite some new wording

## Intentionally excluded in this pull request (a.k.a. TODO)

- Clean non-visible validation errors (eg. size when a partition is being reused).
- Fix a known error in the storage main screen when a partition is being reused (it displays the default filesystem type instead of the reused one).
- Improve the texts of "Can be dynamically adjusted based on..." to reflect the current status more closely.


dgdavid (approved):
Thanks a ton for continuing and nailing this one, @ancorgs!  

The code you added looks good to me. Most likely, we'll keep improving and simplifying it with each iteration, but IMHO it's a really solid starting point. Your demo yesterday afternoon also highlighted some really helpful wording you achieved. I still think that "_Use current_" in the partition options is a bit redundant and noisy, but overall the form is clear and intuitive, which should help users understand it without too much cognitive effort.  

From my side, you can go ahead with this. Both the follow-up mentioned in the description and any other refinements we identify later can be addressed little by little as progressive enhancements.

dgdavid:
While working in LV counterpart I've found two things that worth checking. Recording them below 

## No data loss warning when reusing a partition for swap

Screenshot_2026-06-10_19-59-01


## Switching File System to "Default" even if "Current" is available.

To achieve this, I changed from reusing a partition for swap to reuse it for root, which make the File System selector mount with "Default" and its corresponding warning when maybe would be better to choose "Current" since it is available and it is the safer option.

| Selected | Available |
|-|-|
| Screenshot_2026-06-10_19-59-23 | Screenshot_2026-06-10_20-11-49 |

ancorgs:
> No data loss warning when reusing a partition for swap

Well, by definition no "real data" is stored in the swap space. So I think that is reasonable.

> Switching File System to "Default" even if "Current" is available.
> 
> To achieve this, I changed from reusing a partition for swap to reuse it for root,

So the current filesystem in the partition is swap?

> which make the File System selector mount with "Default" and its corresponding warning when maybe would be better to choose "Current" since it is available and it is the safer option.

Actually if the existing filesystem is swap then it should not be selectable for root. So the problem there is not that "default" was chosen by default but that "current" was allowed.

dgdavid:
> > No data loss warning when reusing a partition for swap
> 
> Well, by definition no "real data" is stored in the swap space. So I think that is reasonable.
> 
> > Switching File System to "Default" even if "Current" is available.
> > To achieve this, I changed from reusing a partition for swap to reuse it for root,
> 
> So the current filesystem in the partition is swap?
> 
> > which make the File System selector mount with "Default" and its corresponding warning when maybe would be better to choose "Current" since it is available and it is the safer option.
> 
> Actually if the existing filesystem is swap then it should not be selectable for root. So the problem there is not that "default" was chosen by default but that "current" was allowed.

I think will be better to address a meeting to talk about this, because most probably I didn't explain well these two unexpected situation to me.

dgdavid:
> 
> I think will be better to address a meeting to talk about this, because most probably I didn't explain well these two unexpected situation to me.

JFTR, @ancorgs and me clarified it in a meeting where I could show the real problem, already fixed at https://github.com/agama-project/agama/pull/3609 in commits https://github.com/agama-project/agama/pull/3609/commits/6c0a856b57379f6ddee15d3836bc2774a8cf7db4 and https://github.com/agama-project/agama/pull/3609/commits/fb3d1a5bc035c1e070cf92c1ac8d00bb40640d63


localhost_8080_ - 2026-06-11T170735 658

dgdavid:
> While working in LV counterpart I've found two things that worth checking. Recording them below
> [...]
> ## Switching File System to "Default" even if "Current" is available.
> [...]

This one has been also fixed as part of #3609 in commits https://github.com/agama-project/agama/pull/3609/commits/cee265c014c1f13dc18d6a34bcf3a49d5550a920 and https://github.com/agama-project/agama/pull/3609/commits/60dd03ecf57fed4155bb420dae634fdcac7a1cba

PR #3586, opened 2026-06-05
[Prototype] Header redesign: simplifying actions and reducing clutter

dgdavid:
## Background

In PR #3009, Agama introduced a major UI redesign that established an Overview page as the central hub and added breadcrumb navigation to almost all pages. As part of this change, a "Review and Install" button was added to the header with the intention of easing the transition to the new UI by making the action highly visible.

This decision, while well-intentioned, quickly revealed itself to be problematic:

- **Poor first breadcrumb design**: The shortcut button diverted attention from the more important first breadcrumb item, which received only an icon with no visual hint that it was a clickable link to return to Overview.
- **Wording complaints**: The original "Install" label felt inaccurate to some people, leading to complaints. It was renamed to "Review and Install" for accuracy, but this only consumed more space without satisfying critics.
- **Space constraints**: What was intended as a temporary transition aid now clutters the header, occupying space increasingly needed for other features, some already there and others incoming:
  - Theme selector
  - Localization (L10n) selector (currently only visible on product selection and overview pages, but needed everywhere)
  - Potential battery indicator for laptop installations
  - ...

## The Problem

The current header design has fundamental issues:

1. **Duplicated functionality**: Both the first breadcrumb item and the "Review and Install" button serve the same purpose (navigate to Overview), creating confusion
2. **Inconsistent header height**: The product logo is only rendered on the Overview page, causing the header height to vary between pages
3. **Poor use of breadcrumb patterns**: The first breadcrumb item uses only an icon, failing to leverage the familiar web breadcrumb convention of showing the hierarchy clearly
4. **Wasted space**: The "Review and Install" button consumes significant horizontal space in an increasingly crowded header

## Proposed Solution

This PR addresses these issues by:

1. **Making the product visible everywhere**: Display the product logo and name on all pages, not just Overview
2. **Proper breadcrumb hierarchy**: Show "Overview" as the first breadcrumb item, making it clear that it's clickable and the top-level page
3. **Removing the redundant button**: Eliminate the "Review and Install" button entirely, using the breadcrumb as the single, clear path back to Overview

### Visual Hierarchy

The new design positions the product name above the breadcrumb navigation:

```
[Logo] SUSE Linux Enterprise Server 16.1 Beta
       Overview > Authentication
```

Screenshot_2026-06-05_08-41-44

This layout:
- Maintains brand visibility with the full product name
- Makes the breadcrumb hierarchy immediately clear
- Uses a descriptive term ("Overview") that communicates the page's purpose
[...]

PR #3604, opened 2026-06-09
feat(web): add dark and high-contrast styles

dgdavid:
> [!IMPORTANT]
>
> **Reviewers**: this is better to be tested by yourself since adding screenshots would be a hard, not effective task.
>
> Also, **bear in mind that the icon/toggle final position and proper alignment will be addressed in https://github.com/agama-project/agama/pull/3586**

# TL;DR

Introduce a full dark mode and a clean, maintainable styling system built on [PatternFly 6.5.1](https://www.patternfly.org/releases/release-highlights). Users can easily switch between Light, Dark, System, and Contrast modes, with their preferences remembered and fully accessible. Shared styles have been refactored into utilities, backed by a curated `--agm-t--*` token API and class-based theming on `<html>`. The system also supports per-product branding and theme-aware logos, creating a solid foundation for theming that can evolve over time.

Related to https://jira.suse.com/browse/PED-14955 (private link)

localhost_8080_ - 2026-06-09T095758 798

localhost_8080_ - 2026-06-09T095909 811

## In depth

A dark theme has been one of the longest-missing pieces of the Agama web UI sop far. It was not for lack of interest: it kept slipping behind other priorities, and there was never quite enough time to do it properly. Doing it properly mattered, because a real theming system is is more than superficial styling. It needs solid foundations underneath, and for a long time those foundations were a single, ever-growing stylesheet full of one-off PatternFly tweaks :see_no_evil: 

In retrospect, the delay has been somewhat beneficial for the project. The web UI is much more stable now, and PatternFly has matured: its design-token system and built-in dark theme have been part of v6 for a while, and high contrast was finalized in 6.5 (it was a beta in 6.4) to meet enhanced WCAG AAA contrast across both light and dark. So instead of hand-building (and forever maintaining) a custom dark palette, Agama could stand on PatternFly's styles and add only what makes Agama feel like Agama. This PR finally delivers dark mode, and uses the moment to clean up the styling foundations and put in place a long-term foundation: a compact, understandable set of appearance controls, paired with the flexibility for each product to give the installer its own identity.

### Visual System Enhancements

- Users can switch between **Light, Dark, and System** (which follows the operating system), along with a **contrast** option. The selected preference is remembered across sessions, and the screen does not flash the wrong theme while loading.
- The default **dark theme matches the Agama look** (the deep "Pine" green from the https://brand.suse.com/design-language#color and [Agama website](https://agama-project.github.io/)), rather than a generic gray palette.
- **Products can re-skin the installer** (colors and even a dark version of their logo) by providing a small configuration file, without any product-specific hard-coding. Including a dark logo version is especially important to comply with SUSE's branding guidelines and avoid the "don'ts" listed at https://brand.suse.com/suse-logo#suse-logo.
- The appearance control is **fully keyboard-accessible and announced to screen readers**, while **high-contrast mode remains accessible** regardless of how a product customizes the rest of the theme.

### Under the hood

The styles moved from one large, hard-to-maintain `index.scss` to a tidy, documented set of partials and a curated token system, all on top of the PatternFly 6.5.1 upgrade. A product (or Agama itself) now adjusts a handful of readable role tokens instead of dozens of internal PatternFly variables, and the system maps each one onto everything it should affect.

## Summary

- Modularizes `index.scss` and introduces a **curated `--agm-t--*` token API** as the single surface for theming.
- Adds a **two-axis appearance system** (color scheme + contrast) with an accessible in-app control, persistence, and no flash on load, built on **PatternFly's built-in themes**.
- Lets each **product restyle Agama** via a small CSS file, plus a coherent Pine dark canvas for the default product and theme-aware logos.
- Upgrades PatternFly to **6.5.1**.

## What's included

- **PatternFly 6.5.1** upgrade (PF 6.5 relocated its stylesheets under `dist/`; imports updated). No component code changes were required.
- **SCSS modularization**: `index.scss` split into `tokens/`, `themes/`, `base/`, `components/`, `utilities/` partials assembled with `@use`/`@forward`. Dead classes removed; shared inline styles refactored into utility classes.
- **Curated token API** (`--agm-t--*` roles): `surface` (+`raised`/`floating`), `text` (+`subtle`, drives regular icons), `border`, `brand` (+`hover`/`clicked`), `brand--text` (on-brand), `link`, `action--background--color--hover`, `focus`, and `status--{info,caution,danger,success,custom}`. Each maps onto all the relevant PatternFly tokens, so setting a role propagates everywhere (page, cards, masthead, modals, menus, borders, button text, alerts, focus ring, labels).
- **AppearanceSettings** control: a `routine`-icon button opening a popover with two `ToggleGroup`s. Persists to `localStorage`, applies `pf-v6-theme-*` classes to `<html>` via `AppearanceProvider`, follows the OS on "System", and ships an inline flash-prevention script so the first paint is correct.
- **Per-product appearance**: a product ships `src/assets/products/<product-id>.css` (co-located with its logo) overriding the role tokens; `useProductAppearance` injects it at runtime. A documented `example.css` lists every role.
- **Default dark theme**: a coherent Pine canvas matching the Agama website, expressed through the surface roles.
- **Theme-aware logos**: `ProductLogo` prefers a `<id>-dark.svg` on dark with a light fallback.

#### Decisions taken

- **Leverage PatternFly's built-in themes, not hand-rolled token lists.** PF 6.5 ships full `pf-v6-theme-dark` and `pf-v6-theme-high-contrast` token sets (~438 tokens each) plus their combination. Agama only overlays the brand/surface tokens it cares about; PF owns the rest. This supersedes the original plan's hand-rolled `data-agm-*` approach.
- **Class-based theming on `<html>`.** Required so portaled content (modals, popovers) inherits the theme; also means "System" auto-switching needs JS (`matchMedia`), handled by the provider.
- **The theme-scoping rule (the key correctness fix).** Color/surface tokens are always theme-scoped (light under `:not(.pf-v6-theme-dark)`, dark under `.pf-v6-theme-dark`); only theme-independent tokens (typography, sizing, radius) sit at a bare `:root`. Our overrides previously sat at `:root` and load after PF, which clobbered PF's dark theme (washed-out surfaces, dark links). This rule is the fix.
- **`var(--agm-t--ROLE, <default>)` mapping pattern.** Each PF token reads its role with a fallback: unset keeps the stock value (PF's own per-theme raw token, no circular refs), set lets the role win. This is what makes "say nothing, get the default" work for both Agama and products.
- **High contrast is PF's by default.** All color mappings are scoped away from high contrast, so PF's accessible HC theme is never overridden by Agama or a product on body text/surfaces. Products may *opt in* to branding only the HC accent (`brand`/`link`), per HC variant, because light-HC is on white and dark-HC on black so a single value rarely contrasts on both.
- **Full brand family + brand-subtle.** PF wires `text`/`icon`/`border` brand independently in dark (and the `brand--subtle` family separately), so the brand role maps the whole family; otherwise radios, secondary buttons, and selected toggles stayed PF blue.
- **Popover, not Dropdown, for AppearanceSettings.** A Dropdown only manages keyboard focus for menu items, leaving the `ToggleGroup`s unreachable. A Popover traps focus so Tab/Space/Enter/Escape work. Each option has a descriptive `aria-label` (e.g. "Dark color scheme") so the two "System" options are unambiguous, and PF's `aria-pressed` conveys selection.
- **Own `usePersistedState` hook** (not a library) so the flash-prevention script can own the `localStorage` format; swappable later.
- **Coherent Pine via surface roles.** The Pine dark canvas is a few surface tokens (canvas, raised, floating, border) at the top of the dark block; PF tokens reference them, so it is tunable in one place and overridable per product.
- **Co-locate product CSS with logos** in `src/assets/products/`; webpack routes it to `dist/assets/appearance/` so the runtime path is unchanged.

#### CSS platform and feature choices

- **Class-based theming** is used instead of relying solely on `prefers-color-scheme` or `light-dark()`. A user's explicit choice (Light, Dark, or Contrast, including overriding the OS setting) cannot be expressed with media queries or the CSS `light-dark()` function alone, since those only follow the operating system's preference. The style is therefore applied via a class on `<html>` (consistent with PatternFly), while `matchMedia` is used only to follow the OS when the "System" option is selected.
- **`@layer` usage deferred** Putting Agama styles in a CSS cascade layer above PatternFly would only win if *all* PatternFly CSS were layered too. It is not: react-core imports ~151 unlayered component CSS files, and unlayered rules always beat any `@layer`. Layering just `patternfly-base` would actually regress our overrides. A full layering pass would need a postcss build transform (and would unlock `revert-layer` as a clean fallback mechanism), so it is documented as an optional follow-up rather than done here. We keep the proven source-order cascade (Agama imported last).
- **No CSS system colors / `forced-colors` for the contrast axis.** An approach was initially explored to implement the high-contrast axis by redefining contrast-critical tokens using [CSS system color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/system-color) keywords (`Canvas`, `CanvasText`, etc.) in `forced-colors` mode. This was later discarded, as the contrast axis aligns with PatternFly’s own high-contrast theme (finalized in 6.5, WCAG AAA), and OS-level forced-colors behavior is handled by the platform and PatternFly rather than being reimplemented.

#### Discarded along the way

- **Hand-rolled dark/contrast token lists**: replaced by leveraging PF's built-in themes once we found them.
- **`@layer` cascade migration**: PF imports ~151 unlayered component CSS files via react-core, so layering only Agama would regress overrides; not free, kept out of scope.
- **A single HC accent value for both light and dark HC**: unreadable on one surface; switched to per-variant HC values.
- **`reactuse`** for the persisted-state hook: deferred in favor of the small in-house hook.

#### Postponed

- **Validation**: WCAG/contrast audit of dark accents, `forced-colors` mode, screen-reader pass, FOUC confirmation, and PF 6.5 component QA (tables/Select/masthead).
- **AppearanceSettings radio-group semantics** (arrow-key roving) as an enhancement over the current accessible toggle-button pattern.
- **`InlineMenuToggle` extraction** across 6 storage editors (from the inline-style audit), and a `60dwh` invalid-unit typo.
- **First-iteration polish**: the default SUSE dark label is slightly low-contrast (green on green tint); a non-blocking edge case where the body background can show an unthemed strip under extreme product reskins.

---


web/src/components/core/AppearanceSettings.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const settings = (
+    <Stack hasGutter>
+      <StackItem>
+        <Content component="small" id={colorSchemeId}>
+          {colorSchemeLabel}
+        </Content>
+        <ToggleGroup aria-labelledby={colorSchemeId}>
+          <ToggleGroupItem
+            // TRANSLATORS: short label for the option shown in the UI
+            text={_("System")}
```
dgdavid: In an internal meeting, @joseivanlopez said that "System" alone might be not fully understand in certain cases. Thus open the question if the __Popover_ should have a small footer explaining what _System_ is here

_System makes style follow your OS or browser preferences_ or similar.

Thoughs?

imobachgs (approved):

web/src/components/core/AppearanceSettings.tsx
dgdavid: @joseivanlopez 

Finally I have chosen _Automatic_, which is the label Firefox uses for its settings and a word we also uses in forms. IMHO, this one fits well, since it well express the final behavior (automatically chosen based on system).

localhost_8080_ - 2026-06-09T140217 480

---

localhost_8080_ - 2026-06-09T140504 132

---

Screenshot_agma-live-installer_2026-06-09_14:06:24

dgdavid: Find changes at commit https://github.com/agama-project/agama/pull/3604/commits/29bfc61ddb423ce65024a9f6fca0beb02f42fff7

web/src/components/core/AppearanceSettings.tsx
```diff
@@ -0,0 +1,159 @@
...
+        </ToggleGroup>
+      </StackItem>
+      <StackItem>
+        <Divider />
+        <Text component="small" textStyle={["textColorSubtle", "fontSizeXs"]}>
+          <Interpolate
+            sentence={
+              // TRANSLATORS: explains the "Automatic" appearance option; %s is
+              // replaced by the "Automatic" option name (shown quoted)
+              "%s honors the browser and system preferences"
```
ancorgs: This string is not enclosed in `_()`

dgdavid: Fixed, thanks

PR #3605, opened 2026-06-09
Add mkfsOptions and mountOptions to the forms

ancorgs:
This is how the form looks initially.
nada

This is how it looks for a new filesystem.
new

And this how it would look like when mounting an existing filesystem.
reused

## TODO

- Fix the "blinking" in which the initial values are visible while submitting.
- Ensure the options (especially the `mkfs` ones) are correctly used by libstorage-ng during commit.


jreidinger (approved):

dgdavid (approved):
LGTM, thanks.

PR #3608, opened 2026-06-10
Fix new susereg

jreidinger:
## Problem

additional url is not respected when registering product with new suseconnect-ng version.

bsc: https://bugzilla.suse.com/show_bug.cgi?id=1268015

## Solution

Always send all parameters like we do in 16.1

## Testing

- ENOTIME to test. Setup is not trivial.


lslezak (approved):
LGTM

PR #3609, opened 2026-06-10
refactor(web): migrate logical volume form to TanStack Form

dgdavid:
## Motivation

The storage [partition form was previously migrated to TanStack Form](https://github.com/agama-project/agama/pull/3603). This PR applies the same approach to the logical volume form while extracting the shared components and validation logic used by both forms. This reduces duplication, improves consistency, and also lays the groundwork for migrating `FormattableDevicePage` next.

## Summary

- Migrates `LogicalVolumePage` to a TanStack Form-based form under `storage/logical-volume-form/`. The old page held its state in eight separate `useState` hooks and kept the filesystem and size fields in sync with the mount point through manual effects that the form switched on and off. The new form drives that syncing through TanStack Form listeners and subscriptions instead (field/form listeners for the name auto-fill, subscriptions for the filesystem and size fields).
- Extracts the storage form building blocks (mount point, filesystem selector, filesystem label, additional filesystem settings, and size fields) plus the validation/presentation helpers into `storage/shared/`, now reused by both the partition and logical volume forms.
- Deletes the old `LogicalVolumePage.tsx` and points the add/edit routes at the new form.

## Implementation notes

- **Shared components are presentation-only.** Validation stays in each form's  `fields.ts`, which reuses the pure helpers in `storage/shared/validation-helpers.ts`.
- **Committed mount point pattern** is preserved and now lives inside the shared `MountPointField` (updates on mount, suggestion select, and blur).
- **lvName auto-fill:** the logical volume name is derived from the mount  point following the same heuristic the backend implements in
  [`Y2Storage::Planned::LvmLv`](https://github.com/yast/yast-storage-ng/blob/a4f6631bc244aadfe40e47ca8959aad2870d74e8/src/lib/y2storage/planned/lvm_lv.rb#L96-L105)  (`/` → `root`, `/home` → `home`, `swap` → `swap`, anything else → no suggestion) and stops once the user edits it, using TanStack listeners +  `isDirty` (pattern documented in `form/conventions.md`).
- `withFrozenQuery` + `useFormSubmit` are used exactly as in the partition form.

## ⚠️ Deviations from the plan, **review closely**

These three points diverge from the [original migration plan](https://gist.github.com/dgdavid/57e71f36360198c3b1ba2938675fa41c#file-2-the-inital-claude-plan-md) and deserve extra reviewer attention:

1. **`SizeFields` was not actually "generic" as the plan assumed.** Its automatic-size solving relied on partition-only route params and hooks  (`useParams` collection/index, `usePartitionable`), so it could not be reused as-is on the logical volume route. It was added a dedicated commit ([`make SizeFields size-solving injectable`](https://github.com/agama-project/agama/pull/3609/commits/637075b3b75ba5064e1684f47cf9a046c64a91f3)) that lifts the solving behind a `useSolvedSizes` prop: each host form now supplies its own. The partition form keeps its solving in `partition-form/use-solved-sizes.ts`, and the logical volume form has its own in `logical-volume-form/use-solved-sizes.ts`.
   **Please sanity-check both solving paths**, especially the LV one (it builds a sparse model via `configModel.logicalVolume.add/edit` and reads the solved  size back via `findByName` / `findVolumeByMountPath`).

2. **`FilesystemAdditionalFields` was extracted too**, which the plan did not call out. As a side effect the logical volume form now exposes the filesystem  **label, mount options, and format (mkfs) options** that the old page lacked.
   This is an intentional feature gain for parity with the partition form, but it does change the LV form's surface, so confirm it is desired.

3. **`target` convention differs from the old page.** The new form uses an empty  string for "new logical volume" (matching the partition form's `name`  convention), whereas the old page used the `"new"` sentinel. The mapping to the config model (`name` set only when reusing) lives in `buildPayload` /  `toFormValues` — worth a careful read to confirm round-tripping for both  create and edit.

## Follow-up: Transformation Logic Extraction

After completing thmigration, a follow-up refactoring extracted data transformation logic into dedicated, testable modules. This improves code organization and enables reuse across storage forms.

### What Changed

**New modules created:**
- `shared/transformations.ts` - Generic helpers for filesystem/size transformations
- `partition-form/transformations.ts` - Partition-specific transformations  
- `logical-volume-form/transformations.ts` - Logical volume-specific transformations

**Extracted from Form.tsx:**
- `buildPayload()` - form values → ConfigModel
- `toFormValues()` - ConfigModel → initial form values
- `useSolvedSizes()` - backend size calculation
- Consolidated `use-solved-sizes.ts` into transformations modules

## Follow-up: Code organization (queries, constants, validations)

A second follow-up round completed the form directory layout so every concern has a predictable home, in both storage forms:

- **`queries.ts`** (2655ec509): the route-param + config-model query hooks that feed each form moved out of `Form.tsx` into a dedicated module.
- **Single source for shared constants** (6741af9aa): `FILESYSTEM_TYPE`,  `FILESYSTEM_ACTION`, and `SIZE_MODE` now live once in `shared/fields.ts`;  the form-local `fields.ts` modules import and re-export them instead of  keeping copies in sync by convention.
- **`validations.ts`** (a4b9868cf): submit validation moved out of `fields.ts` into its own module, leaving `fields.ts` as the pure field vocabulary  (types, constants, defaults).
- **Documentation** (7268cf260): the form directory modules and the shared fields re-export pattern are documented in `form/conventions.md`.

## Follow-up: UX fixes after manual testing

Manual testing of the logical volume form surfaced a batch of issues; some of the fixes also apply to the partition form. One commit per issue.

### Logical volume form

- **Simpler name field** (d3e8a474b): label is now just "Name" (the form context already says it is a logical volume), with a self-explanatory hint:  "Name for the logical volume. E.g., root, home, lv0".
- **Name suggestion mirrors the backend heuristic** (5b7e164f6):  `lvNameFromMountPoint` now matches what the installer would pick on its own: `/` suggests `root` (previously empty), and non-absolute mount points suggest nothing instead of being used verbatim.
- **Logical volume selector mirrors the partition selector** (19e106b85): the "New logical volume" option explains where it will be created, and each reusable logical volume shows its size, filesystem label, and a "Use current …" description of the data it holds.
- **Read-only field when nothing is left to reuse** (c4cf6e530): when the volume group exists but has no reusable logical volumes, a read-only field ("New logical volume. There are no available existing logical volumes on …") replaces the pointless one-option dropdown, mirroring the partition form.
- **No selector at all for new volume groups** (8ab9c4c72): when the volume group is being created in the same configuration, every logical volume is necessarily new, so the field is not rendered. Unlike disks (which always
  exist), a new volume group has nothing to reuse by definition.
- **"Current" preselected when reusing a volume with data** (49d102810): selecting an existing logical volume with a filesystem now switches the file system selector to "Current" instead of leaving "Default" (which implied
  formatting), exactly as the partition form already did.

### Both forms

- **"Current" preselected when editing a reused device** (ad0362343): choosing "Current" stores `{reuse: true}` with no filesystem type, but `toFormValues` only preselected "Current" when a type was present, so re-editing such a configuration wrongly showed "Default". Both forms now  honor an explicit `reuse: true`.
- **Data loss notice always shown when a reused device is formatted** (049e85619, 6c0a856b5, fb3d1a5bc): the notice was missing when the mount point constrains the filesystem to a single type (e.g. swap, where a read-only field replaces the dropdown) and when the reused device's filesystem is incompatible with the mount point. It now renders in both variants. `ReadOnlyField` learned to nest dependent content (like `DropdownField`) so the notice stays visually attached to the field.
- **"Current" restored when keeping the filesystem becomes possible again** (60dd03ecf): the automatic downgrade of "Current" to "Default" (e.g. after switching the mount point to swap) was permanent; switching back to a compatible mount point kept "Default". The reuse-vs-format intent is now tracked in the `filesystemAction` field, updated only by the user's own selections, so the form can restore "Current" without ever overriding a deliberate format choice.

### Refactoring and documentation

- **Shared `FilesystemFields`** (cee265c01): both forms carried copy-paste identical filesystem logic (volume template lookup, compatible filesystem calculation, options building, auto-reset effect). It now lives once in
  `storage/shared/FilesystemFields.tsx`, driven by a single `reusedDevice` prop; each form keeps a thin wrapper that resolves which device is being reused. As a side effect this fixed a bug: the partition form's "Current"
  option description named the whole disk instead of the partition.
- **Control fields pattern documented** (c54ce5781): fields like `committedMountPoint` and `filesystemAction` exist only to coordinate the form's behavior and never reach the payload. The pattern and its conventions are now described in `form/conventions.md`, and the fields are marked as control fields in `shared/fields.ts`. Follow-up noted there: evaluate a naming marker (underscore prefix or suffix).

## Initial screenshots

localhost_8080_ - 2026-06-10T202927 762
localhost_8080_ - 2026-06-10T203024 766
localhost_8080_ - 2026-06-10T203105 563

## Test plan

- [x] `npm run check-types`
- [x] `npm run eslint`
- [x] `npm run test` (storage + form suites: 740 passing)
- [ ] Manual: create a new logical volume on an existing volume group
- [ ] Manual: create a logical volume on a new volume group (no source field)
- [ ] Manual: reuse an existing logical volume (no name/size fields; "Current"
      filesystem option preselected)
- [ ] Manual: reuse and format (data-destruction notice)
- [ ] Manual: edit an existing logical volume (values round-trip, "Current"
      stays preselected)
- [ ] Manual: filesystem auto-reset when switching to an incompatible mount
      point, and "Current" restored when switching back to a compatible one
- [ ] Manual: data loss notice with single-type mount points (e.g. swap)
- [ ] Manual: name field shows "Name" with the new hint; `/` suggests `root`;
      auto-fill stops after manual edit
- [ ] Manual: logical volume selector options carry descriptions (new + reuse)
- [ ] Manual: read-only source field when no logical volumes are left to reuse
- [ ] Manual: keyboard navigation and screen-reader labels

---

Related to https://github.com/agama-project/agama/pull/3603

---


joseivanlopez:
Some details I see after playing with the LV form:

* The title of the page does not include the name of the VG: Storage >  > Configure logical volume
* The explanation of Automatic size mentions partitions: "Minimum: 512 MiB (default size for generic partitions)"
* The explanation for Default file system mentions partitions: "XFS (default file system for generic partitions)"

web/src/components/storage/logical-volume-form/Form.tsx
```diff
@@ -0,0 +1,297 @@
...
+  config,
+}: LogicalVolumeFormContentQuery) {
+  const { id } = useParams();
+  const volumeGroupIndex = Number(id);
+  const navigate = useNavigate();
+  const addLogicalVolume = useAddLogicalVolume();
+  const editLogicalVolume = useEditLogicalVolume();
+
+  // Mount points already in use, excluding the current one when editing.
+  const usedMountPoints = useMemo(() => {
```
joseivanlopez, line 136: I think this can be extracted to a shared hook that receives `config` and `mountPath`.

web/src/components/storage/logical-volume-form/Form.tsx
```diff
@@ -0,0 +1,297 @@
...
+  LogicalVolumeFormContent,
+);
+
+/**
+ * Page shell for the logical volume create/edit flow.
+ *
+ * Owns the breadcrumbs and the volume group null-check (Resource Not Found).
+ */
+export default function LogicalVolumeForm() {
+  const volumeGroupConfig = useVolumeGroupConfig();
```
joseivanlopez, line 265: Nothing to do now, but propably we need to use `withFronzenQuery` for `LogicalVolumeForm`. Otherwise, the form is rendered everytime the config changes.

web/src/components/storage/logical-volume-form/Form.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+  const breadcrumbs: BreadcrumbProps[] = [
+    // TRANSLATORS: breadcrumb label for the storage configuration section.
+    { label: _("Storage"), path: STORAGE.root },
+  ];
+
+  if (volumeGroupConfig) {
+    breadcrumbs.push(
+      { label: volumeGroupConfig.name },
```
joseivanlopez, line 274:
```suggestion
      { label: volumeGroupConfig.name ?? volumeGroupConfig.vgName },
```

web/src/components/storage/logical-volume-form/Form.tsx
```diff
@@ -0,0 +1,297 @@
...
+ */
+export default function LogicalVolumeForm() {
+  const volumeGroupConfig = useVolumeGroupConfig();
+
+  const breadcrumbs: BreadcrumbProps[] = [
+    // TRANSLATORS: breadcrumb label for the storage configuration section.
+    { label: _("Storage"), path: STORAGE.root },
+  ];
+
+  if (volumeGroupConfig) {
```
joseivanlopez, line 272: NP: if a volume group config does not exist, then we should early return with `ResourceNotFound`. Again, nothing to change now, but propably I would have two differerent XContent components.

web/src/components/storage/logical-volume-form/LogicalVolumeSourceFields.tsx
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * The volume group as it exists in the system, or `undefined` when the
+     * volume group is new (not yet created).
+     */
+    volumeGroup?: System.Device;
+  },
+  render: function Render({ form, availableLogicalVolumes, volumeGroup }) {
+    // The volume group does not exist yet: every logical volume is necessarily
+    // new, so there is nothing to choose and the field is not rendered.
+    if (!volumeGroup) return null;
```
joseivanlopez, line 77: The form directly decices whether to render `LogicalVolumeNameField` and `SizeFields`. But this compoment decides itself. I would move the control to the form and change the props to always receive a `volumeGroup`.

web/src/components/storage/logical-volume-form/validations.ts
```diff
@@ -0,0 +1,162 @@
...
+    filesystem: requiredString(fields.filesystem, _("Select a filesystem type")),
+    filesystemLabel: optionalFilesystemLabel(fields.filesystemLabel),
+  };
+}
+
+function validateSizeFields(fields: FormFields): FieldsValidationResult<SizeFields> {
+  // Size only applies when creating a new logical volume.
+  if (isReusingLogicalVolume(fields.target)) return {};
+
+  if (fields.sizeMode === SIZE_MODE.FIXED) {
```
joseivanlopez, line 88: All this code for size validation is repeated at partition validations.

web/src/components/storage/shared/FilesystemLabelField.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+const FilesystemLabelField = withForm({
+  ...sharedDefaultOptions,
+  props: {} as {
+    /** Optional override for the field label. */
+    label?: React.ReactNode;
+  },
+  render: function Render({ form, label }) {
+    return (
+      <form.AppField name="filesystemLabel">
```
joseivanlopez, line 48: BTW, maybe the `name` of the fields migth be a prop. It would make the field even more reusable because it is not attached to a specific value name of the form state. Each form could use a different value name for this field. Just an idea, nothing to change ;).

web/src/components/storage/DeviceSelectorModal.tsx.back
```diff
@@ -0,0 +1,661 @@
+/*
```
joseivanlopez, line 1: Leftover file?

web/src/components/storage/SearchedDeviceMenu.tsx.back
```diff
@@ -0,0 +1,388 @@
+/*
```
joseivanlopez, line 1: Leftover?

web/src/components/storage/SearchedVolumeGroupMenu.tsx.back
```diff
@@ -0,0 +1,275 @@
+/*
```
joseivanlopez, line 1: Leftover?

web/src/components/storage/logical-volume-form/Form.tsx
joseivanlopez: Maybe this is the culprit of the refreshing when the form is submitted.

joseivanlopez (commented):
I have added some comments. Nothing serious. Great work!

web/src/components/storage/logical-volume-form/Form.tsx
dgdavid: We can have a look. There are some things I didn't watched because lack of storage internal knowledge. So, I really appreciate the helpt to review and refine all this.

web/src/components/storage/logical-volume-form/Form.tsx
dgdavid: Did I forget to frozen that form? I have to check it more carefully when in front of the laptop (using the smart phone know, too small screen checking the code)

dgdavid: But Suarez we can do it today or next week. I mean, something shouldn't block the next release, it's kind of enhancement in the context of the migration we're trying to finis.

web/src/components/storage/logical-volume-form/Form.tsx
dgdavid: Arg! It's not the first time you remind this to me. Sorry.

web/src/components/storage/logical-volume-form/Form.tsx
dgdavid: Wasn't the Resource not found thingy included in the migration? That's make me think I have to re-check the partition-form too, just in case.

web/src/components/storage/logical-volume-form/validations.ts
dgdavid: Then, something to move to _share_.

web/src/components/storage/shared/FilesystemLabelField.tsx
dgdavid: The more reusable the better. I did not see the use case, but find with me.

web/src/components/storage/DeviceSelectorModal.tsx.back
dgdavid: Indeed. Fortunately it's only one. I have s ton of .back diles untracked in my current tracking branch

web/src/components/storage/SearchedDeviceMenu.tsx.back
dgdavid: Yep, 😅

web/src/components/storage/SearchedVolumeGroupMenu.tsx.back
dgdavid: Indeed. Finally not only once :)

Bad news is that I have another PR on top of that, so these files will appear there too.

web/src/components/storage/logical-volume-form/transformations.ts
```diff
@@ -0,0 +1,430 @@
...
+ * //   }
+ */
+export function buildPayload(
+  values: typeof defaultOptions.defaultValues,
+): ConfigModel.LogicalVolume {
+  const isReuse = isReusingLogicalVolume(values.target);
+
+  return {
+    mountPath: values.mountPoint,
+    lvName: values.lvName,
```
ancorgs, line 205:
```suggestion
    lvName: isReuse ? undefined : values.lvName,
```

dgdavid: Please, go ahead with the changes you spot necessary. Just commit them, since I'll be off for a while (more than usual) and don't want to block the work. Feel free to jump in.

ancorgs:
@joseivanlopez found several small things. I also found some.

But since #3610 sits on top of this pull request, I think the best we can do is merge this into the feature branch and fix all the found issues either as part of #3610 or as a separate pull request.

So I'm summarizing all those things that would need to be addressed after the merge:

- Thin pools can be chosen to be re-used. I don't think that should be the case (it is already wrong in the master branch).
- **Inconsistent way of displaying the warning about missing content. The commit e15d8fab089892d on #3610 tries to solve that, but I'm not sure it is a good fix.**
- **When using a generic size or filesystem type, the string always contains "[...]for generic partitions".**
- **The LV name is sent even when reusing an LV (see [suggestion](#discussion_r3401758277))**
- **An empty breadcrumb is displayed for new VGs (see [suggestion](#discussion_r3400953375)).**
- Size validation is duplicated for partitions and LVs.
- Calculation of usedMountPoints should be extracted to a shared hook.
- **Some `.back` files slipped into this pull request**
- We need to recheck the usage of `withFronzenQuery` (see [comment](#discussion_r3400944222)).
- **Mount paths are not formatted in the example of the field**
- Several smaller notes in some parts of the code.

Similar items coming from previous pull requests (#3605, #3603 and #3610)

- **Some validations for fields that are not visible could prevent from saving. It happened with wrong sizes after setting "size" back to automatic.**
- In the storage main screen, if a device is reused the default filesystem for that mount point is displayed, instead of the type of the re-used filesystem.
- If a mountpoint is reusing a partition/lv and then another partition/lv is being defined for the same mount point (which will fail when saving the form), then the sentence "X GiB (default size for /whatever)" displays the wrong size (the one from the reused device instead of the real default value).
- The forms display the initial values when submitting, although the proper values are actually submitted.
- We discussed whether it would make sense to improve the texts of "Can be dynamically adjusted based on..." to reflect the current status more closely.
- There is no need for an specialized `FilesystemFields` in `formattable-device-form`.

Taking all that into account, I'm merging here to move on. Because it generally works and is a huge step forward in sharing code among the different forms.

dgdavid:
Fine with me, although 

> * We need to recheck the usage of `withFronzenQuery` (see [comment](#discussion_r3400944222)).


I would postpone this til everything else is done,  unless you've pretty clear how to fix it or it is quite straightforward.

PR #3610, opened 2026-06-10
refactor(web): migrate formattable device form to TanStack Form

dgdavid:
## Motivation

The storage [partition form](https://github.com/agama-project/agama/pull/3603) and [logical volume form](https://github.com/agama-project/agama/pull/3609) were previously migrated to TanStack Form. This PR completes the series by migrating `FormattableDevicePage`, the last of the three storage forms, onto the same shared infrastructure.

## Summary

- Migrates `FormattableDevicePage` to a TanStack Form-based form under `storage/formattable-device-form/`. The old page held its state in four separate `useState` hooks and kept the filesystem selector in sync with the mount point through a manual effect (`useAutoRefreshFilesystem`) that the form switched on and off. The new form derives the filesystem options from the committed mount point and only auto-resets the selection when it becomes incompatible, like the other two forms.
- This is the simplest of the three storage forms: there is no partition/logical volume selector (the whole device is always used) and no size fields, so the "current filesystem" comes from the device itself.
- Reuses the shared building blocks from `storage/shared/` (`MountPointField`, `FilesystemSelector`, `FilesystemAdditionalFields`), the shared validation helpers, and the shared transformation helpers. No changes to the shared components were needed.
- Deletes the old `FormattableDevicePage.tsx` and points the format-device route at the new form.

localhost_8080_ - 2026-06-11T010642 852

## Implementation notes

- **Same patterns as the sibling forms:** `withFrozenQuery` for refetch protection, `useFormSubmit` for the submit lifecycle, submit-only validation using the `shake()` pattern, and the committed mount point pattern via the shared `MountPointField`.
- **The form directory follows the module layout from `form/conventions.md`:** `fields.ts` (field vocabulary: types, constants, defaults, re-exporting the shared parts from `storage/shared/fields.ts`), `validations.ts` (the submit `validate` function), `transformations.ts` (`buildPayload`/`toFormValues` mapping to and from the device config, built on the shared transformation helpers), and `queries.ts` (data hooks).
- **Data hooks live in `queries.ts`** (`useDeviceModelFromParams`, `useDeviceFromParams`, `useUnusedMountPoints`), resolved from the `collection`/`index` route params via `createPartitionableLocation`, with a `ResourceNotFound` guard when the params don't map to an existing device.
- **No disabled fields.** The old page disabled the filesystem selector until a valid mount point was chosen; the new form keeps every field enabled and lets validation handle the dependency, per `form/conventions.md`.

## ⚠️ Deviations from the plan and behavior changes, **review closely**

These points diverge from the [original migration plan](https://gist.github.com/dgdavid/4124a9210aec0db48dcf837eadef8ad5#file-2-the-initial-claude-plan-md) or from the old page's behavior and deserve extra reviewer attention:

1. **The filesystem is no longer auto-selected when picking a mount point.** The old page's `useAutoRefreshFilesystem` switched the selector to "Current <fs>" (reuse) whenever the device had a filesystem compatible with the chosen mount point, and to the concrete default type otherwise. The new form starts at **"Default"** (automatic) and only auto-resets when the current selection becomes incompatible, matching the partition and logical volume forms. In practice this means a device with reusable data now **defaults to being formatted** unless the user explicitly picks "Current"; the data-destruction notice from the shared `FilesystemSelector` mitigates this, but please confirm the trade-off (consistency across forms vs. the old keep-data-by-default behavior) is acceptable.
2. **The form gains features the old page lacked**: the "Default" (automatic) filesystem option, and the additional filesystem settings (label moved there, plus mount options and format/mkfs options) behind the "Define more file system settings" checkbox. Intentional feature gain for parity with the other forms, same as in #3609.
3.  **Navigation on submit changes from the storage root to `navigate(-1)`**, aligning with the partition and LV forms instead of keeping the old page's behavior.

## Follow-up

The branch evolved after this description was written:

- **Deviation 1 no longer applies.** Since 071979863, the form preselects "Current" when the device holds a filesystem compatible with the mount point and there is no stored config asking to format. A device with reusable data is no longer formatted by default, and the old keep-data-by-default behavior is back. Now implemented through the `filesystemAction` control field instead of the old `useAutoRefreshFilesystem` effect, so a deliberate format choice survives mount point changes (60dd03ecf).

- **The shared components did change after all.** The filesystem fields logic (options, compatibility check, auto-reset/restore) was extracted into `storage/shared/FilesystemFields` (cee265c01) and this form now reuses it (e5b87d89b), so the three storage forms share one implementation.

- **The data-loss notice also covers content without a filesystem** (e.g. a disk used as LVM PV or holding a partition table): such a device cannot offer "Current", but formatting it still destroys data, so the notice keys off the device content rather than its filesystem (e15d8fab0).

- Note for manual testing: "Current" is only offered for filesystem types allowed by the product's volume templates (e.g. not for NTFS), as in the other storage forms and in the old page.

## Test plan

- [x] `npm run check-types`
- [x] `npm run eslint`
- [x] `npm run test` (full suite: 2117 passing, incl. 43 formattable device form tests across Form, validations, and transformations suites)
- [ ] Manual: format a drive with a mount point and a concrete filesystem
- [ ] Manual: format a drive with the "Default" filesystem (hint shows the resolved type)
- [ ] Manual: keep the existing filesystem ("Current" option; no label/format options)
- [ ] Manual: format a device with data (destruction notice shown)
- [ ] Manual: edit an already configured device (values round-trip, incl. label and options)
- [ ] Manual: filesystem auto-reset when switching to an incompatible mount point (e.g. swap)
- [ ] Manual: validation errors (empty/invalid/duplicate mount point, invalid label)
- [ ] Manual: keyboard navigation and screen-reader labels

---

Related to https://github.com/agama-project/agama/pull/3603 and https://github.com/agama-project/agama/pull/3609

---

Co-Authored-By: Claude Sonnet 4.5 [noreply@anthropic.com](mailto:noreply@anthropic.com)


dgdavid:
@ancorgs , 

Commit https://github.com/agama-project/agama/pull/3610/commits/0719798639efbf4249b456511424d48417c3ed27 allows the form to open like 

Screenshot_2026-06-12_00-34-54


and therefore achieve

Screenshot_2026-06-12_00-42-52

web/src/components/storage/formattable-device-form/FilesystemFields.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ * Filesystem fields for the formattable device form.
+ *
+ * Thin wrapper around the shared FilesystemFields. Unlike the partition or
+ * logical volume forms, there is no device selector here: the whole device is
+ * always the one being reused, so it is passed straight through.
+ */
+const FilesystemFields = withForm({
```
joseivanlopez, line 41: In see the point of having a custom `FilesystemFields` for *partition-form* and *logical-volume-form* because they have to subscribe to changes in the form state. But in this case we can directly use the shared `FilesystemFields`.

web/src/components/storage/formattable-device-form/Form.tsx
```diff
@@ -0,0 +1,241 @@
...
+  config,
+}: FormattableDeviceFormContentQuery) {
+  // Route params and mutations are not frozen: they don't feed defaultValues
+  // and must always reflect the current state.
+  const { collection, index } = useParams();
+  const navigate = useNavigate();
+  const setFilesystem = useSetFilesystem();
+
+  // Mount points already in use, excluding the current one when editing.
+  const usedMountPoints = useMemo(() => {
```
joseivanlopez, line 101: As in the logical volumes PR, this is candidate to become a shared hook.

web/src/model/storage/device.ts
```diff
@@ -57,4 +57,29 @@ function supportShrink(device: Device): boolean {
...
-export { isDrive, isVolumeGroup, isMd, isPartition, isLogicalVolume, deviceSystems, supportShrink };
+/**
+ * Whether the device currently holds data that formatting it would destroy.
+ *
+ * Relies on structural signals only: a filesystem of its own, logical volumes
+ * (when it is a volume group), a partition table with partitions, or detected
+ * installed systems. A device that holds content but reports none of these is
+ * treated as empty on purpose; that gap belongs to the backend, not here.
+ */
+function hasContent(device: Device): boolean {
```
joseivanlopez, line 68: This check is incomplete. Already discussed with @ancorgs. For now, we will always show a warning when formatting an existing device. But this and other fixes will be addressed in a separate PR.

dgdavid: Thanks

ancorgs (approved):
This seems to work as expected. So we agreed to merge into the feature branch and then do any further fix/improvement there.

PR #3611, opened 2026-06-11
Implement RFC 9457 Problem Details for HTTP error handling

imobachgs:
Implements [RFC 9457](https://datatracker.ietf.org/doc/html/rfc9457) Problem Details for HTTP APIs to improve error handling in Agama's HTTP API.

## Changes

- Replaces simple `{"error": "..."}` JSON responses with structured Problem Details format
- Uses tag URIs (`tag:agama.opensuse.org,2026:problems/...`) that don't require endpoints
- Enum-based approach for type safety with variant-specific fields
- Translation support via gettext for title and detail fields
- Shares types between client (agama-cli) and server (agama-utils::api)
- OpenAPI documentation with schemars and aide
- Status code only in HTTP response (not in JSON body) to avoid redundancy
- Do not validate the profile twice.

## Error Types

- `SchemaValidationFailed` (400) - with list of validation errors
- `InvalidJson` (400) - malformed JSON in request
- `Unauthorized` (401) - authentication failures
- `InternalError` (500) - generic server errors
- `Generic` - fallback using RFC 9457's `about:blank` type

## Follow-up

- Add a specific type for profile-related errors (e.g., AutoYaST conversion).
- Do all the checks on the server-side, even if it is not a valid JSON file.

## Example Response

```json
{
  "type": "tag:agama.opensuse.org,2026:problems/schema-validation-failed",
  "title": "Schema Validation Failed",
  "detail": "The provided configuration does not match the required schema",
  "errors": ["/network/hostname: is required"]
}
```

## Benefits

- **Localizable**: Error messages respect server locale (set via ConfigureL10n)
- **Structured**: Extension fields provide context (e.g., validation errors, file paths, D-Bus service names)
- **Type-safe**: Enum variants enforce correct fields for each error type
- **Consistent**: All API errors follow the same format
- **Standards-compliant**: Follows RFC 9457 with proper `application/problem+json` content type

### Authenticating

Before:

```shell
% ./target/debug/agama -- --insecure --host 192.168.122.10 auth login
> Please enter the root password: ********
Backend responded with code 401 and the following message:

"Authentication via PAM failed: Auth_Err (7)"
```

After:

```shell
% ./target/debug/agama -- --insecure --host 192.168.122.10 auth login
Error:
Unauthorized

Details:
Authentication failed: Auth_Err (7)
```

### Wrong format

```shell
% ./target/debug/agama --insecure --host 192.168.122.10 config load /srv/www/htdocs/agama/wrong.jsonnet 
Backend responded with code 400 and the following message:

"Profile validation failed: The profile is not a well-formed JSON file"
```

After (it checks locally whether it is, at least, a valid JSON file). The real validation happens in the "config load". Check the "follow-up" section:

```shell
% ./target/debug/agama --insecure --host 192.168.122.10 config load wrong.jsonnet
It is not a valid JSON file
```

### Invalid schema

Before:

```shell
% ./target/debug/agama --insecure --host 192.168.122.10 config load /srv/www/htdocs/agama/wrong.json
✗ The profile is not valid. Please, check the following errors:

	* true is not of type "string". /product/id
	* 1 is not of type "string". /product/mode

The profile is not valid
```

After:

```shell
% ./target/debug/agama --insecure --host 192.168.122.10 config load wrong.json
Error:
Schema validation failed

Details:
The provided configuration does not match the required schema

  - true is not of type "string". /product/id
  - 1 is not of type "string". /product/mode
```

### DIFF
--- a/rust/agama-lib/src/http/base_http_client.rs
+++ b/rust/agama-lib/src/http/base_http_client.rs
@@ -369,7 +369,7 @@ impl BaseHTTPClient {
             Err(_) => {
                 if let Ok(error_object) = serde_json::from_str::<serde_json::Value>(&text) {
                     if let Some(error) = error_object.get::<&str>("error") {
-                        ProblemDetails::internal_error(&error.to_string())
+                        ProblemDetails::internal_error(error.to_string())
                     } else {
                         ProblemDetails::internal_error(text)
                     }
--- a/rust/agama-lib/src/http/base_http_client.rs
+++ b/rust/agama-lib/src/http/base_http_client.rs
@@ -18,6 +18,7 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
+use agama_utils::api::ProblemDetails;
 use reqwest::{header, IntoUrl, Response};
 use serde::{de::DeserializeOwned, Serialize};
 use url::Url;
@@ -34,23 +35,8 @@ pub enum BaseHTTPClientError {
     InvalidURL(#[from] url::ParseError),
     #[error(transparent)]
     InvalidJSON(#[from] serde_json::Error),
-    #[error("Backend responded with code {} and the following message:\n\n{}", .0, format_backend_error(.1))]
-    BackendError(u16, String),
-}
-
-fn format_backend_error(error: &String) -> String {
-    let message: Result<serde_json::Value, _> = serde_json::from_str(error);
-
-    match message {
-        Ok(message) => {
-            if let Some(error) = message.get("error") {
-                error.to_string().replace("\\n", "\n")
-            } else {
-                format!("{:?}", error)
-            }
-        }
-        Err(_) => format!("{:?}", error),
-    }
+    #[error("{1}")]
+    Problem(u16, ProblemDetails),
 }
 
 /// Base that all HTTP clients should use.
@@ -377,6 +363,21 @@ impl BaseHTTPClient {
             .text()
             .await
             .unwrap_or_else(|_| Self::NO_TEXT.to_string());
-        BaseHTTPClientError::BackendError(code, text)
+
+        let problem = match serde_json::from_str::<ProblemDetails>(&text) {
+            Ok(problem) => problem,
+            Err(_) => {
+                if let Ok(error_object) = serde_json::from_str::<serde_json::Value>(&text) {
+                    if let Some(error) = error_object.get::<&str>("error") {
+                        ProblemDetails::internal_error(&error.to_string())
+                    } else {
+                        ProblemDetails::internal_error(text)
+                    }
+                } else {
+                    ProblemDetails::internal_error(text)
+                }
+            }
+        };
+        BaseHTTPClientError::Problem(code, problem)
     }
 }
--- a/rust/agama-lib/src/http/base_http_client.rs
+++ b/rust/agama-lib/src/http/base_http_client.rs
@@ -368,8 +368,8 @@ impl BaseHTTPClient {
             Ok(problem) => problem,
             Err(_) => {
                 if let Ok(error_object) = serde_json::from_str::<serde_json::Value>(&text) {
-                    if let Some(error) = error_object.get::<&str>("error") {
-                        ProblemDetails::internal_error(error.to_string())
+                    if let Some(error) = error_object.get("error").and_then(|e| e.as_str()) {
+                        ProblemDetails::internal_error(error)
                     } else {
                         ProblemDetails::internal_error(text)
                     }
--- a/rust/agama-server/src/server/web.rs
+++ b/rust/agama-server/src/server/web.rs
@@ -22,7 +22,7 @@
 
 use crate::profile::profile_service;
 use crate::server::config_schema;
-use crate::web::error::ProblemDetailsExt;
+use crate::web::error::{ProblemDetailsExt, ProblemDetailsResponse};
 use agama_lib::logs;
 use agama_manager::service::Error as ManagerError;
 use agama_manager::users::PasswordCheckResult;
@@ -192,7 +192,7 @@ fn get_status_docs(op: TransformOperation) -> TransformOperation {
         )
         .tag("System & Monitoring")
         .response_with::<200, Json<Status>, _>(|res| res.description("Status of the installation"))
-        .response_with::<500, Json<ProblemDetails>, _>(|res| {
+        .response_with::<500, ProblemDetailsResponse, _>(|res| {
             res.description("Internal server error")
         })
 }
@@ -219,7 +219,7 @@ fn get_system_docs(op: TransformOperation) -> TransformOperation {
         .response_with::<200, Json<SystemInfo>, _>(|res| {
             res.description("System information retrieved successfully")
         })
-        .response_with::<500, Json<ProblemDetails>, _>(|res| {
+        .response_with::<500, ProblemDetailsResponse, _>(|res| {
             res.description("Internal server error")
         })
 }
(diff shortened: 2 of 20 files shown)


rust/agama-server/src/server/web.rs
```diff
@@ -70,19 +71,32 @@ pub enum Error {
...
 impl Error {
+    /// Converts this error into RFC 9457 Problem Details
+    pub fn into_problem_details(self) -> ProblemDetails {
+        match self {
+            Error::ConfigSchema(e) => e.into_problem_details(),
+            Error::Json(e) => ProblemDetails::invalid_json(e.to_string()),
+            Error::Manager(e) => ProblemDetails::internal_error(e.to_string()),
+            Error::Questions(e) => ProblemDetails::internal_error(e.to_string()),
+            Error::MissingLanguageTag => {
+                ProblemDetails::generic("Missing Language Tag", "The language tag is required")
```
jreidinger, line 82: Should not it be localized?

rust/agama-server/src/web/auth.rs
```diff
@@ -52,10 +53,18 @@ pub enum AuthError {
 
 impl IntoResponse for AuthError {
     fn into_response(self) -> Response {
-        let body = json!({
-            "error": self.to_string()
-        });
-        (StatusCode::UNAUTHORIZED, Json(body)).into_response()
+        let problem = match self {
+            AuthError::MissingToken => {
+                ProblemDetails::unauthorized("Authentication token is missing")
```
jreidinger, line 58: Should not it be localized? and same for later

rust/agama-utils/src/api/problem_details.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+pub fn write_problem(
+    f: &mut std::fmt::Formatter<'_>,
+    title: &str,
+    detail: Option<&str>,
+    body: Option<&str>,
+) -> std::fmt::Result {
+    write!(f, "An error ocurred: {title}")?;
```
jreidinger: why "An error occurred " is not translated while details are?

rust/zypp-agama/zypp-agama-sys/src/bindings.rs
```diff
@@ -687,7 +687,7 @@ unsafe extern "C" {
     #[doc = " repository array in list.\n when no longer needed, use \\ref free_repository_list to release memory\n @param zypp see \\ref init_target\n @param[out] status (will overwrite existing contents)"]
     pub fn list_repositories(zypp: *mut Zypp, status: *mut Status) -> RepositoryList;
     pub fn free_repository_list(repo_list: *mut RepositoryList);
-    #[doc = " Adds repository to repo manager\n @param zypp see \\ref init_target\n @param alias have to be unique\n @param url can contain repo variables\n @param[out] status (will overwrite existing contents)\n @param callback pointer to function with callback or NULL\n @param user_data"]
+    #[doc = " Adds repository to repo manager\n @param zypp see \\ref init_target\n @param alias have to be unique\n @param url can contain repo variables\n @param priority repository priority, 0 uses libzypp's default\n @param[out] status (will overwrite existing contents)\n @param callback pointer to function with callback or NULL\n @param user_data"]
```
jreidinger, line 690: thanks for this update. it slipped when I do review of that extra_repositories

rust/agama-utils/src/api/patch.rs
```diff
@@ -33,9 +33,13 @@ pub struct Patch {
 }
 
 impl Patch {
-    pub fn with_update(config: &Config) -> Result<Self, serde_json::Error> {
+    pub fn with_update(config: Value) -> Result<Self, serde_json::Error> {
```
jreidinger: unnecessary Result as fn cannot return Error

rust/agama-server/src/server/web.rs
```diff
@@ -181,7 +192,7 @@ fn get_status_docs(op: TransformOperation) -> TransformOperation {
         )
         .tag("System & Monitoring")
         .response_with::<200, Json<Status>, _>(|res| res.description("Status of the installation"))
-        .response_with::<500, Json<ErrorResponse>, _>(|res| {
+        .response_with::<500, Json<ProblemDetails>, _>(|res| {
```
jreidinger: Here is comment from AI that makes sense for me:

Using Json<ProblemDetails> in .response_with calls forces Aide to document the response as application/json, which conflicts with ProblemDetailsExt::into_response setting
  application/problem+json. This could lead to client generator mismatches and should be replaced with a type that correctly sets the content type in the OpenAPI spec (e.g.,
  ProblemDetailsResponse with an appropriate aide content-type override). Also seen in lines L220, L247, L270, L305, etc.

rust/agama-utils/src/api/problem_details.rs
jreidinger: and typo "occurred"

rust/agama-lib/src/http/base_http_client.rs
```diff
@@ -377,6 +363,21 @@ impl BaseHTTPClient {
...
             .await
             .unwrap_or_else(|_| Self::NO_TEXT.to_string());
-        BaseHTTPClientError::BackendError(code, text)
+
+        let problem = match serde_json::from_str::<ProblemDetails>(&text) {
+            Ok(problem) => problem,
+            Err(_) => {
+                if let Ok(error_object) = serde_json::from_str::<serde_json::Value>(&text) {
+                    if let Some(error) = error_object.get::<&str>("error") {
+                        ProblemDetails::internal_error(error.to_string())
```
jreidinger: `error.to_string()` will use json quoting which I am not sure is what is wanted.
See this playground, maybe you want `as_str`
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9fa0b53b7aaefdff21a1a8ff46ad20ce

rust/agama-server/src/server/web.rs
imobachgs: I am not sure. In general, the "title" of the ProblemDetails is translated. However, that's not the case for the details. The reason is that, in some cases, the "details" are coming from 3rd party libraries and it is strange to translate some of them but not translate others.

rust/agama-server/src/web/auth.rs
imobachgs: Same as above.

rust/agama-utils/src/api/problem_details.rs
imobachgs: Because I forgot :-)

rust/agama-server/src/server/web.rs
imobachgs: Initially it was documented as `ProblemDetailsResponse` but it generated an invalid OpenAPI spec. I will try again anyway.

rust/agama-lib/src/http/base_http_client.rs
imobachgs: Good point, but the proposed solution (using `unwrap`) does not look good. I would do something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=15ab966433347f8545d685b7144a6837.

rust/agama-server/src/server/web.rs
imobachgs: It turned out that the solution was not as simple as using `ProblemDetailsResponse`. But now it should look better.

rust/agama-lib/src/http/base_http_client.rs
jreidinger: yeah, I do not propose to use unwrap. I just want to demonstrate issue with `to_string`. your solution is fine...also unwrap_or that falls back to to_string is fine for me.

rust/agama-utils/src/api/problem_details.rs
```diff
@@ -186,7 +186,8 @@ pub fn write_problem(
     detail: Option<&str>,
     body: Option<&str>,
 ) -> std::fmt::Result {
-    write!(f, "An error ocurred: {title}")?;
+    let first_line = gettext("An error occurred: %s").replace("%s", title);
```
jreidinger: I thought that for dynamic text we use i18n_format crate

jreidinger (approved):

jreidinger (approved):

PR #3612, opened 2026-06-11
Deprecate mkfsOptions in favor of mkfsExtraArguments

ancorgs:
## Problem

- https://trello.com/c/lt6dOvXm/5779-2-storage-form-2-format-options-mount-options-in-the-form-to-create-mount-a-partition

We find out the field `mkfsOptions` in the JSON API was highly confusing and not really translated into the correct actions.

## Solution

Deprecate the confusing field in favor of a new one that is more obvious and should work almost out of the box and without introducing ambiguities.

mkfsExtraArgs

## Testing

- Added a new unit tests
- Tested manually

## ToDo

- Open the corresponding pull request for documentation

### DIFF
--- a/web/src/openapi/config/storage.ts
+++ b/web/src/openapi/config/storage.ts
@@ -336,7 +336,7 @@ export interface Filesystem {
   /**
    * Options for creating the file system.
    */
-  mkfsOptions?: string[];
+  mkfsExtraArguments?: string[];
   /**
    * Options to add to the fourth field of fstab.
    */
--- a/web/src/openapi/config/storage.ts
+++ b/web/src/openapi/config/storage.ts
@@ -334,9 +334,9 @@ export interface Filesystem {
   path?: string;
   mountBy?: MountBy;
   /**
-   * Options for creating the file system.
+   * Additional arguments injected as-is to the mkfs command when creating the file system.
    */
-  mkfsExtraArguments?: string[];
+  mkfsExtraArguments?: string;
   /**
    * Options to add to the fourth field of fstab.
    */
--- a/doc/storage.md
+++ b/doc/storage.md
@@ -136,7 +136,6 @@ So it will be possible to define (or reuse) multi-device Btrfs file systems usin
   "metaDataRaidLevel": "..." ,
   "devices": [ ... ],
   "label": "...",
-  "mkfsOptions": { ... },
   "subvolumePrefix": "...",
   "subvolumes": [ ... ],
   "snapshots": ...,
--- a/rust/agama-lib/share/storage.model.schema.json
+++ b/rust/agama-lib/share/storage.model.schema.json
@@ -143,7 +143,7 @@
         "default": { "type": "boolean" },
         "type": { "$ref": "#/$defs/filesystemType" },
         "label": { "type": "string" },
-        "mkfsOptions": { "type": "array", "items": { "type": "string" } },
+        "mkfsExtraArguments": { "type": "string" },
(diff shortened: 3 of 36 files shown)


dgdavid (commented):
Changes in the web part looks good to me, but I would merge them once #3609 and #3610 are merged.

mvidner:
Playing stupid here...
When I see "**Additional** format arguments", I am thinking, additional to what?

Am I correct in assuming that the `mkfs` command will become this?
`mkfs -t ext4 $VALUE /dev/sda9 123456789`

I have checked [man mkfs](https://www.man7.org/linux/man-pages/man8/mkfs.8.html) and that indeed seems the only way. And now I wonder what the problem with mkfsOptions was.

ancorgs:
> And now I wonder what the problem with mkfsOptions was.

We modeled `mkfsOptions` and `mountOptions` in the same way. That is, both were lists of strings.

But then `mountOptions` was indeed managed as a list of strings by Agama -> Y2Storage -> libstorage-ng. Meaning it was separated by commas when needed (eg. writing it to fstab), any layer (Agama, Y2Storage or libstorage-ng) can add or remove options in special cases (eg. the infamous "_netdev" option), duplicates can be detected, etc.

But `mkfsOptions` was actually only handled as a list of strings by Agama. It was then passed to Y2Storage as `mkfsOptions.join(" ")`. That was confusing. Let's take this string "-l version=2 -K" as example. What is the correct list of options? Maybe `["-l version=2 -K"]` or maybe `["-l", "version=2", "-K"]` or maybe `["-l version=2", "-K"]`...

web/src/openapi/config/storage.ts
```diff
@@ -336,7 +336,7 @@ export interface Filesystem {
   /**
    * Options for creating the file system.
    */
-  mkfsOptions?: string[];
+  mkfsExtraArguments?: string[];
```
mvidner: Ha! A leftover I think
```suggestion
  mkfsExtraArguments?: string;
```

ancorgs:
> Am I correct in assuming that the `mkfs` command will become this? `mkfs -t ext4 $VALUE /dev/sda9 123456789`

Correct. Unfortunately the arguments of `mkfs` are completely different for each file-system type. Even the flag used to set the file-system label is different. So libstorage-ng will compose the `mkfs` call with certain arguments (which depend A LOT of the file-system type, the label, the weather...) and then this field allows to add additional/extra arguments. I didn't find a better way to express that in the label.

web/src/openapi/config/storage.ts
ancorgs: Caused because I did a manual update instead of actually updating the file from the JSON schema. Fixed.

mvidner:
OK, so this fixes one problem ("if the options are a list, where are the item boundaries?") but keeps another problem: "how do I know if my option is _additional_ or, uh, _core_? What are the core options anyway?"
It turns out that the core options are a complex topic not easily explained in the UI, 
but I hope that my worries turn out to be academic and not a problem for actual users.

mvidner (approved):
LGTM.
(I may have poked in too hard from a theoretical angle about the UI strings. No big deal)

ancorgs:
> I may have poked in too hard from a theoretical angle about the UI strings

That's fine. In fact, those are completely legit remarks which also worry me. But I didn't find a good way to explain all that in the documentation or in the UI in a practical way.

So I will merge for now, but I remain fully open to suggestions (we have a second chance when reviewing/merging the feature branch into master).

PR #3614, opened 2026-06-12
Fixes for storage forms

joseivanlopez:
Follow-up of https://github.com/agama-project/agama/pull/3609 and https://github.com/agama-project/agama/pull/3610.

* [x] Inconsistent way of displaying the warning about removing content.
* [x] When using a generic size or filesystem type, the string always contains "[...]for generic partitions".
* [x] The LV name is sent even when reusing an LV 
* [x] An empty breadcrumb is displayed for new VGs
* [x] Some .back files slipped into this pull request
* [x] Mount paths are not formatted in the example of the field
* [x] Some validations for fields that are not visible could prevent from saving. It happened with wrong sizes after setting "size" back to automatic.


ancorgs (approved):
LGTM

PR #3615, opened 2026-06-12
Print the status after the exiting the monitor

imobachgs:
Print the output of the "agama status" comment after exiting the monitor. The idea is that the  user knows what is going on after the last command. Take into account that the monitor runs when loading the configuration or when performing the installation.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Fri Jun 12 12:37:58 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
-- Print the status after running the "monitor", "config load" and
-  "install" commands (gh#agama-project/agama#3615).
+- Print the status after running the "monitor", "config load",
+  "config edit" and "install" commands (gh#agama-project/agama#3615).
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jun 12 12:37:58 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Print the status after running the "monitor", "config load" and
+  "install" commands (gh#agama-project/agama#3615).
```
mvidner: The PR description , and apparently the code, only mention `agama monitor` but NOT the other two commands

imobachgs: The monitor runs on those commands. See [config load](https://github.com/agama-project/agama/blob/ebd786ba09d71b64f9a3e6865bb1480c7bf48ed4/rust/agama-cli/src/config.rs#L221), [config edit](https://github.com/agama-project/agama/blob/ebd786ba09d71b64f9a3e6865bb1480c7bf48ed4/rust/agama-cli/src/config.rs#L253) and [install](https://github.com/agama-project/agama/blob/ebd786ba09d71b64f9a3e6865bb1480c7bf48ed4/rust/agama-cli/src/lib.rs#L184).

mvidner (approved):
LGTM

I have tested it with testing_using_container.sh

PR #3616, opened 2026-06-12
Unify the output of the validation command

imobachgs:
- Unify the output of the "agama validate" command when using the remote or the local one.
- Drop the "profile is valid" message.
- Reduce the number of validation attempts.

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -25,11 +25,10 @@ use agama_lib::{
     profile::{ProfileHTTPClient, ProfileValidator, ValidationOutcome},
     utils::FileFormat,
 };
-use agama_utils::api;
+use agama_utils::api::{self, ProblemDetails};
 use agama_utils::make_long;
 use anyhow::{anyhow, Context};
 use clap::{Arg, ArgAction, ArgMatches, Command};
-use console::style;
 use fluent_uri::Uri;
 use gettextrs::gettext;
 use tempfile::Builder;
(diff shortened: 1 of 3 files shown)


rust/agama-cli/src/config.rs
```diff
@@ -453,18 +404,13 @@ async fn edit(
...
-            return Err(anyhow!("The profile is not valid"));
-        }
-
+        validate(http_client, CliInput::Full(updated.clone())).await?;
         return Ok(serde_json::from_str(&updated)?);
     }
 
-    Err(anyhow!(
+    Err(anyhow!(gettext(
         "Ignoring the changes becase the editor was closed with an error code."
```
jreidinger: typo "becase"

rust/agama-cli/src/config.rs
```diff
@@ -268,54 +263,25 @@ async fn patch_config(
...
-    match validity {
-        ValidationOutcome::Valid => {
-            eprintln!("{} {}", style("\u{2713}").bold().green(), validity);
-        }
-        ValidationOutcome::NotValid(_) => {
-            eprintln!("{} {}", style("\u{2717}").bold().red(), validity);
-        }
-    }
-
     Ok(())
```
jreidinger, line 285: I am not sure if I get logic. local validation returns ValidationOutcome and remote one always return (). So how exactly it works when profile is not valid?

imobachgs: It uses the new problems API: it relies on the error coming from the API. And in the local case, it turns the `ValidationOutcome` into a `ProblemsDetails`. So the rest is handled in the same way.

rust/agama-cli/src/config.rs
imobachgs: Oh, and old one :-)

jreidinger (approved):

PR #3621, opened 2026-06-15
Unify "status" and "monitor" strings

imobachgs:
The output of the "agama status" and "agama monitor" command can be improved with better texts and some additional formatting.

* Improve and unify tests.
* `agama status` does not display issues if there are pending questions.
* Remove the "titles" ("Details", "Blocking issues", and "Open questions").
* Do not display the scope of the issues.
* Adjust text of the LoadConfigRetryQuestion dialog.

# Screenshots

image

image

image

### DIFF
--- a/rust/agama-cli/src/monitor/ui/content.rs
+++ b/rust/agama-cli/src/monitor/ui/content.rs
@@ -169,7 +169,7 @@ impl<'a> Content<'a> {
         });
 
         let lines = vec![Line::from(Span::styled(
-            gettext("Following issues needs to be addressed in the configuration:"),
+            gettext("Fix invalid settings before starting the installation:"),
             Style::default().add_modifier(Modifier::BOLD),
         ))];
         Paragraph::new(lines).render(title_area, buf);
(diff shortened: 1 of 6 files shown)


jreidinger (approved):

rust/agama-cli/src/monitor/ui/content.rs
```diff
@@ -169,7 +169,7 @@ impl<'a> Content<'a> {
         });
 
         let lines = vec![Line::from(Span::styled(
-            gettext("Following issues needs to be addressed in the configuration:"),
+            gettext("Fix invalid settings before starting the installation:"),
```
jreidinger, line 172: I am not sure if settings has to be invalid...it can be simply missing like missing product or root password. Maybe it can be simple `Fix settings before starting the installation:` ... and to be honest as rest of commands are `agama config` I would probably use configuration instead of settings.

rust/agama-cli/src/status.rs
```diff
@@ -73,13 +73,19 @@ impl InstallationEnum {
...
+            ),
+            Self::Succeeded => gettext("The installation finished successfully."),
+            Self::Ready => gettext("Ready to start the installation."),
+            Self::Installing => gettext("The installation is in progress."),
+            Self::Proposing => gettext("The installer is preparing an installation proposal."),
+            Self::Question => gettext(
+                "There are unanswered questions. Use the \"agama monitor\" command or the \
+                web user interface to answer them:",
+            ),
+            Self::Issues => gettext("Fix invalid settings before starting the installation:"),
```
jreidinger, line 88: similar like above. So both are related

rust/agama-cli/src/monitor/ui/content.rs
imobachgs: We agreed to use this string as we use basically the same in the UI.

PR #3622, opened 2026-06-15
feature(web): split dock for Terminal

dgdavid:
> [!NOTE]
>
>  This is a proof-of-concept for where the terminal should live in the web UI, following up on the discussion in #3617. It docks the terminal as a resizable panel at the bottom of the app, opened from the installer menu, with show/hide, minimize/restore, and a "needs more space" state on smaller screens. The terminal content is a mockup for now: the real xterm.js session and its backend transport are out of scope here, the focus is placement, layout, and interaction. 
> 
> Find below the  reasoning, including why not a tab, popup, or drawer.

---

## Reconsidering my own comment

In my own comment at https://github.com/agama-project/agama/pull/3617#issuecomment-4692789565 I argued to start simple: offer the terminal as a modal/popup the way other tools do, doubted there would be much back-and-forth between the terminal and the UI, and floated a "minimize" mechanism or an always-available drawer only as possible later enhancements. Having thought it through, I've changed my mind, both on the popup and on the drawer, for the reasons I lay out below.

On the drawer specifically: it is a legitimate option ([PatternFly's own Drawer guidelines](https://www.patternfly.org/components/drawer/design-guidelines) even list a terminal as a sample use case), but Agama doesn't use drawers today, and if the UI evolves in that direction I can see them being valuable for workflows closer to the web UI's core purpose (details, previews, edit views beside the main content). That makes the drawer a space worth keeping free, especially since drawers could not compose well when several features want one at once. Spending it on terminal access now would likely be a shot in the foot tomorrow, when a workflow better suited to it comes along. So I'd rather find the terminal a home that doesn't take that space.

## The case for a split

@lslezak makes a good point in #3617: when users use the terminal they could switch between it and the installer, run a command, check the result in the web UI, act on it, run another, and so on. That back-and-forth could be real **even for a single command**: users could need to keep part of the page in view while they type (a value to copy, an error to read, a setting to check). **Wanting both on screen at once, without one covering the other, is a reasonable thing to ask for.**

The most terminal-heavy work will still happen in the Agama CLI and monitor interfaces, which are built for exactly that and expose far more than the web UI ever will, so **I still see the in-UI terminal as a cherry on top rather than a first-class citizen of the web UI**. But that is no reason to do it poorly: the question is simply where it lives.

Drawer aside, the real choice is between @lslezak's separate browser tab and a split. Both let users keep the terminal and the installer on screen at once, without the constant open and close. I'd suggest the split, and I think it serves his concern better:

- **Everything stays in one place.** A separate tab is a second window to manage and a second connection to keep in sync, and it quietly breaks the feeling that the installer is one app. A split keeps the installer and the terminal in the same window.
- **No shortcut collisions.** Switching back from a tab leans on `Ctrl+Tab` (as the mockup suggests), but that is the browser's own shortcut for cycling tabs, so we would be fighting the browser. With a split there is nothing to switch to, so the problem disappears.
- **Both interfaces at once.** Nothing to flip between, which answers the concern directly.
- **It stays out of the way until wanted.** The terminal isn't a first-class control at the top of the UI; it lives quietly in the installer menu and only appears once a user reveals it. Someone who never needs it never sees it.
- **It respects the "convenience" framing.** Once revealed, a user can **minimize** it to a slim bar instead of hiding it, keeping it within reach without giving it prime real estate.
- **The session survives.** The panel lives above the page outlet, so navigating between pages doesn't throw the session away; hide and minimize both keep it alive.
- **It doesn't spend the drawer.** The drawer stays free for the detail/edit workflows that will actually want it.

localhost_8080_ - 2026-06-15T080342 208

---

localhost_8080_ - 2026-06-15T080846 307

---

https://github.com/user-attachments/assets/551882e7-d85f-4e09-bd9d-f84e47e44a8b

## Why at the bottom, not the side

One thing the research turned up: PatternFly lays its components out based on the width of the browser window, not the width of the area they actually sit in (it uses viewport media queries, not container queries). A terminal on the side would make the app narrower than the window while PatternFly still arranged everything as if it had the full width, so the header actions and content would overflow and overlap the terminal. Docking the split at the bottom avoids this entirely: the app keeps full window width, so its responsive layout keeps working, and only the height is shared.

## Accessibility

The chrome around the terminal (the show/hide menu entry, the minimize, restore, hide and resize controls) is built with proper labels and roles, so that part is fine for screen-reader and keyboard users.

The terminal surface itself is the open question. XTerm.js does offer a `screenReaderMode` option (off by default) that exposes DOM elements to support NVDA and VoiceOver, plus a `minimumContrastRatio` option for contrast. So the building blocks exist; we should enable and actually test them. If screen-reader mode proves unreliable in practice, a simple fallback is to mark the terminal viewport `aria-hidden` and point assistive-technology users to the CLI, which is the better tool for them anyway. **This is worth a real test pass before we commit** rather than assuming either way.

## A follow-up question: should the terminal be available during installation?

One thing I haven't settled is whether the terminal should stay reachable once the installation is actually running, and on the progress, finished, and failed screens that take over the layout. There is a good case for it: those are arguably the moments troubleshooting matters most, a step hangs, something fails, and a user wants to look around.

Today the panel lives alongside the regular pages, so it isn't shown once installation takes over the screen. Making it available there is doable, but it is a deliberate choice rather than something that happens for free (those screens are their own layouts), so I'd rather raise it than assume. Worth deciding before we wire the real terminal in.

---

Co-Authored-By: Claude Sonnet 4.5 [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/web/src/components/core/TerminalPane.tsx
+++ b/web/src/components/core/TerminalPane.tsx
@@ -24,9 +24,18 @@ import Text from "~/components/core/Text";
 import VisualTooltip from "~/components/core/VisualTooltip";
 import TerminalUnavailable from "~/components/core/TerminalUnavailable";
 import { useTerminal } from "~/context/terminal";
-import { _ } from "~/i18n";
+import { _, N_ } from "~/i18n";
 
 import textStyles from "@patternfly/react-styles/css/utilities/Text/text";
+
+// Terminal action strings reserved for the translation freeze.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const FUTURE_TRANSLATIONS = {
+  show: N_("Show terminal"),
+  hide: N_("Hide terminal"),
+  open: N_("Open terminal"),
+  close: N_("Close terminal"),
+};
 
 const DEFAULT_FONT_SIZE = 14;
 const MIN_FONT_SIZE = 8;
--- a/web/src/components/core/TerminalPane.tsx
+++ b/web/src/components/core/TerminalPane.tsx
@@ -181,13 +181,13 @@ export default function TerminalPane({ enoughSpace }: TerminalPaneProps) {
     </VisualTooltip>
   );
 
-  // Not enough room for a usable terminal: just explain it, keeping the Hide
-  // action so the panel can still be dismissed.
+  // Not enough room for a usable terminal: just explain it. The message offers
+  // its own Hide action, so the header needs no actions here.
   if (!enoughSpace) {
     return (
-      <TerminalShell actions={hideButton}>
+      <TerminalShell>
         <CardBody isFilled className="agm-terminal__body">
-          <TerminalUnavailable />
+          <TerminalUnavailable onHide={hide} />
         </CardBody>
       </TerminalShell>
     );
--- a/web/src/components/core/TerminalPane.tsx
+++ b/web/src/components/core/TerminalPane.tsx
@@ -211,9 +211,7 @@ export default function TerminalPane({ enoughSpace }: TerminalPaneProps) {
   const toggleLabel = isMinimized ? restoreLabel : minimizeLabel;
 
   // TRANSLATORS: one-line summary of what the terminal is for
-  const summary = _(
-    "Linux command-line with administrative privileges on the installer system.",
-  );
+  const summary = _("Linux command-line with administrative privileges on the installer system.");
 
   const actions = (
     <Flex
--- a/web/src/components/core/TerminalPane.tsx
+++ b/web/src/components/core/TerminalPane.tsx
@@ -1,0 +1,247 @@
+/*
+ * Copyright (c) [2026] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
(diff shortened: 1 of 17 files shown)


dgdavid:
> localhost_8080_ - 2026-06-15T080846 307

Of course, these texts needs to be rewording accordingly and there is missing the "Close" or "Hide" action in a primary button or link form below the text, like we are doing in other empty states.

web/src/components/core/TerminalPane.tsx
```diff
@@ -0,0 +1,247 @@
...
+  // terminal panel to a bar, keeping the session
+  const minimizeLabel = _("Minimize terminal");
+  // TRANSLATORS: tooltip and accessible name for the button that expands the
+  // collapsed terminal panel back to its full size
+  const restoreLabel = _("Restore terminal");
+  const toggleLabel = isMinimized ? restoreLabel : minimizeLabel;
+
+  // TRANSLATORS: one-line summary of what the terminal is for
+  const summary = _(
+    "Run commands on the installer system. Meant for inspecting hardware, advanced setup, or troubleshooting.",
```
dgdavid:
```suggestion
    "Linux command-line with administrative privileges on the installer system (use with caution).",
```

Image

dgdavid:
```suggestion
    "Linux command-line with administrative privileges on the installer system (for troubleshooting and advanced configuration).",
```

Image

dgdavid:
```suggestion
    "Linux command-line with root access on the installer (troubleshooting and configuration).",
```

Image

web/src/components/core/TerminalPane.tsx
```diff
@@ -0,0 +1,256 @@
...
```
dgdavid:
```suggestion
    "Linux command-line with administrative privileges on the installer system.",
```

ancorgs (approved):
LGTM

PR #3623, opened 2026-06-15
mark properly autoinst question as the one that needs string data

jreidinger:
## Problem

Failed load of autoinst profile is does not ask in monitor for new URL.

## Solution

Mark properly question as the one that needs string value back.


mchf (approved):
lgtm

PR #3624, opened 2026-06-15
Do not flood logs with useless errors when downloading default profile fails

mchf:
## Problem

Agama searches for default profiles on each start. Mainly those *.xml ones wrote useless error / exception reports into logs when not found.

## Solution

The error happened because  an internal function for fetching the xml file returned nil which was not handled anymore. Simply failing doesn't solve all cases we need - in case of non present default profiles we want to silently ignore without failing. So, a capability of ignoring the fail by checking YAST_SKIP_PROFILE_FETCH_ERROR env variable was added. This variable is set in (almost) all cases when agama-autoyast is invoked from the code. When this variable is set and no profile is found, then empty profile is used in subsequent processing to end the processing pipeline with success.

## Testing

- *Tested manually*

## Screenshots

See the discussion in the bug.


mchf:
Exception trace is gone:
agama-journalctl

but that "500 Internal server error" doesn't look good too. Problem is that we're in generic profile fetch area here. I'm going to investigate possible solution / improvements.

mchf:
to_agama conversion failed on empty profile -> agama-autoyast exited with error -> that caused 500 Internal server error

kobliha:
This seems to be related to https://github.com/yast/yast-autoinstallation/pull/882 and https://github.com/yast/yast-autoinstallation/commit/8f8237c0b80905a18d215afae56d6c66aa563dca

service/lib/agama/commands/agama_autoyast.rb
```diff
@@ -51,6 +51,16 @@ def initialize(url, directory)
       # Run the command fetching, checking, converting and writing the Agama configuration.
       def run
         profile = fetch_profile
+
+        if profile.nil? && ENV["YAST_SKIP_PROFILE_FETCH_ERROR"] != "1"
```
imobachgs: What about something like:

```
if profile.nil?
  if ENV["YAST_SKIP_PROFILE_FETCH_ERROR"] ...
end
```

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Sun Jun 28 14:59:51 UTC 2026 - Michal Filka <mfilka@suse.com>
+
+- AutoYast to Agama profile conversion doesn't crash when empty
```
```suggestion
- AutoYaST to Agama profile conversion doesn't crash when empty
```

service/lib/agama/commands/agama_autoyast.rb
mchf: re-worked

imobachgs (approved):

PR #3630, opened 2026-06-15
Make the JSON editor really read-only

lslezak:
## Problem

- When showing the JSON config the editor was not really read-only, although the `isReadOnly` flag was set.

## Solution

- It seems the read-only property needs to be set in the editor options when that is configured. After removing the whole `options` configuration the `isReadOnly` flag works. But we need that `options` so I moved the value there.

## Testing

- Tested manually, the editor does not allow to change the content, it shows a read-only hint when trying to do so.

## Note

- I added a custom message so it can be translated.

## Screenshots

image


dgdavid:
Closing in favor of #3630, which includes a cherry-pick from here for the translation.

PR #3634, opened 2026-06-15
web: Fix and unify some strings

ancorgs:
## Problem

After some adaptation in the logic of the LVM web UI, the texts were not really accurate.

On the other hand, we may soon may need even more improvements in that page.

Last but not least, during a recent review of the CLI we detected some strings that were equivalent but different.

## Solution

Fix the texts and prepare some new ones.


joseivanlopez (approved):
LGTM

joseivanlopez (approved):

PR #3638, opened 2026-06-15
Busy state calc

jreidinger:
## Problem

Monitor shows state as busy when there is question during probing or installation like password for encrpted devices.

## Solution

Make questions having higher priority then existence of progress in state detection.

## Testing

- *Tested manually*

## Screenshot

Screenshot From 2026-06-15 23-56-41


rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jun 17 08:03:10 UTC 2026 - Josef Reidinger <jreidinger@suse.com>
+
+- agama monitor: Properly report as waiting for user when
+  agama is busy and has some question to answer
```
```suggestion
  Agama is busy and has some question to answer
```

imobachgs (approved):

PR #3640, opened 2026-06-16
Add optional cert_path second argument to rmt example

rtamalin:
## Problem

Exercise the agama-suseconnect use of libsuseconnect.so in a way that simulates registering against an RMT using a self-signed certificate. This helped identify and verify the fix for https://bugzilla.suse.com/show_bug.cgi?id=1268017 as proposed in SUSE/connect-ng#400.

## Solution

If the optional second argument specifying a cert path has been provided, install it in the system CA store before calling announce_system().

## Testing

- Leveraged in conjunction with SUSE/connect-ng#390 framework to test the SUSE/connect-ng#400 fix.

### DIFF
--- a/rust/suseconnect-agama/examples/rmt.rs
+++ b/rust/suseconnect-agama/examples/rmt.rs
@@ -1,4 +1,4 @@
-use std::{env, process::exit};
+use std::{env, fs, path::Path, process::{exit, Command}};
 
 use suseconnect_agama::{
     activate_product, announce_system, create_credentials_file, ConnectParams, ProductSpecification,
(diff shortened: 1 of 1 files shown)


rust/suseconnect-agama/examples/rmt.rs
```diff
@@ -8,10 +8,18 @@ use url::Url;
 pub fn main() {
     tracing_subscriber::fmt::init();
     let args: Vec<String> = env::args().collect();
+
+    // Basic argument validation
+    if args.len() < 2 {
+        eprintln!("Usage: rmt <url> [<cert_path>]");
```
jreidinger, line 17: the usage is already below at line 19, so I think it will be better to adapt it there ( the usage and exit 1 is perfectly fine for me )

rust/suseconnect-agama/examples/rmt.rs
```diff
@@ -8,10 +8,18 @@ use url::Url;
...
+        exit(1);
+    }
+
     let Some(url) = args.get(1) else {
         eprintln!("Provide url as first parameter");
         return;
     };
+    let cert_path = args.get(2).cloned();
```
jreidinger: this line is probably not necessary. Having just below that let for cert should be enough. ( you do not need to clone value as it is not used anywhere ).

rust/suseconnect-agama/examples/rmt.rs
rtamalin: resolved

rust/suseconnect-agama/examples/rmt.rs
rtamalin: resolved

jreidinger (approved):

PR #3642, opened 2026-06-17
Add a unit test for parsing of a wrong files section

imobachgs:
Just add a unit test to probe the problem detected in [bsc#1268432](https://bugzilla.suse.com/show_bug.cgi?id=1268432).


imobachgs:
This behavior comes from serde. Instead of fighting with it, we decided [to do the validation first](https://github.com/agama-project/agama/pull/3648).

PR #3643, opened 2026-06-17
Do not download the Monaco editor parts from CDN

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1268433
- When opening the Monaco editor using the "Show configuration" menu option it downloads some JavaScript files dynamically from a 3rd party CDN server

Screenshot from the browser devtools:
agama-monaco-cdn

## Solution

- Bundle the editor locally

## Notes

- I have moved the configuration editor to a separate component so it can be lazily loaded. The Monaco editor is loaded only when the current configuration is displayed for the first time. Then it is kept in memory.

## Testing

- Tested manually

## TODO

- [x] There is 404 error for the `codicon.ttf` file, that is located in `node_modules/monaco-editor/esm/vs/base/browser/ui/codicons/codicon/codicon.ttf`, we need to copy this file.
- [x] The `index.js` file size in production build increased from ~1.7MB to ~5.5MB, that's a huge increase, try loading the `<CodeEditor>` component dynamically only when the editor is needed. The `index.css` size increased only slightly from 1.6MB to 1.7MB, that is OK.
- [x] Fix the failing unit tests :scream: 

## Screenshots

When opening the editor everything is loaded from the Agama server, the CDN is not used anymore:

agama-monaco-local3

### Slow network

When the network is slow a spinner is displayed while the Monaco editor is being loaded. This was simulated in the Chrome browser using the "Fast 4G" network throttling option in the devtools:

[agama-editor-2026-06-18_16.04.05.webm](https://github.com/user-attachments/assets/53050da0-2924-4916-bc60-ca5ac5ccf11b)

### DIFF
--- a/web/src/components/core/ConfigDialog.tsx
+++ b/web/src/components/core/ConfigDialog.tsx
@@ -20,44 +20,35 @@
  * find current contact information at www.suse.com.
  */
 
-import React, { useEffect, useState } from "react";
-import { Content, Flex, Spinner } from "@patternfly/react-core";
-import { CodeEditor, Language } from "@patternfly/react-code-editor";
+import React, { Suspense } from "react";
+import { Bullseye, Content, Flex, Spinner } from "@patternfly/react-core";
 import Popup from "~/components/core/Popup";
-import { useAppearance } from "~/context/appearance";
-import { ROOT } from "~/routes/paths";
-import { isoTimestamp } from "~/utils";
 import { _ } from "~/i18n";
 
 type ConfigDialogProps = {
   onClose: () => void;
 };
 
+// load the component and its dependencies dynamically when needed, do not
+// include it in the default index.js file, the Monaco editor is huge
+const ConfigEditor = React.lazy(
+  () =>
+    import(
+      /* webpackChunkName: "config-editor" */
+      "~/components/core/ConfigEditor"
+    ),
+);
+
 /**
  * Dialog showing the current installation configuration in JSON format, with
  * options to copy or download the content.
  */
 export default function ConfigDialog({ onClose }: ConfigDialogProps) {
-  const { isDark } = useAppearance();
-  const [config, setConfig] = useState<string | undefined>(undefined);
-  // CodeEditor appends the extension based on the language, so it must be omitted here.
-  const [downloadFileName, setDownloadFileName] = useState("agama-config");
-
-  useEffect(() => {
-    const loadConfig = async () => {
-      try {
-        const response = await fetch(ROOT.config);
-        const text = await response.text();
-        setConfig(JSON.stringify(JSON.parse(text), null, 2));
-        setDownloadFileName(`agama-config-${isoTimestamp()}`);
-      } catch (error) {
-        console.error("Failed to load config:", error);
-        setConfig("");
-      }
-    };
-
-    loadConfig();
-  }, []);
+  const fallback = (
+    <Bullseye>
+      <Spinner />
+    </Bullseye>
+  );
 
   return (
     <Popup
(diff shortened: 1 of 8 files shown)


web/src/components/core/ConfigDialog.tsx
```diff
@@ -23,6 +23,8 @@
 import React, { useEffect, useState } from "react";
 import { Content, Flex, Spinner } from "@patternfly/react-core";
 import { CodeEditor, Language } from "@patternfly/react-code-editor";
+import * as monaco from "monaco-editor/esm/vs/editor/editor.api.js";
```
dgdavid: I would extract all of this to a "CodeEditor" component at core, importing there PF CodeEditor as PFCodeEditor. Why? To avoid repeating this workaround in a different place if we end using the editor at other place.

web/src/components/core/InstallerOptionsMenu.tsx
```diff
@@ -33,10 +33,15 @@ import {
...
-import ConfigDialog from "~/components/core/ConfigDialog";
 import DownloadLogsFeedback from "~/components/core/DownloadLogsFeedback";
 import { _ } from "~/i18n";
 
+// load the component and its dependencies dynamically when needed, do not
+// include it in the default index.js file
+const ConfigDialog = React.lazy(
+  () => import(/* webpackChunkName: "code-editor" */ "~/components/core/ConfigDialog"),
+);
+
```
dgdavid: Great movement, thanks.

I had something similar (not doable at this moment, requires router changes) for the whole Agama, but this is a perfect case (if work) to use it in isolation.

web/webpack.config.js
```diff
@@ -40,6 +41,11 @@ const copy_files = [
   { from: "./src/assets/products/*.svg", to: "assets/logos/[name][ext]" },
   // Per-product appearance overrides, loaded at runtime by product id.
   { from: "./src/assets/products/*.css", to: "assets/appearance/[name][ext]" },
+  // copy the "codicon" font used by the Monaco editor
```
dgdavid: I have doubts about this. I'm working in a different PR, a kind of follow up of https://github.com/agama-project/agama/pull/3604 in which I introduce the SUSE Mono https://brand.suse.com/design-language#typography

So, I prefer using just our mono fonts that adding yet another one.

web/webpack.config.js
```diff
@@ -60,6 +66,10 @@ const plugins = [
   // run the TypeScript type checks in a parallel process (report the problems to the dev server in
   // the development mode)
   new ForkTsCheckerWebpackPlugin({ devServer: development }),
+  new MonacoWebpackPlugin({
+    // include only support for the JSON format in the Monaco editor
+    languages: ["json"],
```
dgdavid, line 66: Another good movement, thanks!

web/webpack.config.js
lslezak: This is rather a workaround to fix the 404 error , if you have a better fix / better implementation then use that.

web/webpack.config.js
lslezak: When the Monaco editor is bundled locally then Webpack would include support for everything (in separate chunks) but as many parts would never be used they would just take extra space.

web/webpack.config.js
dgdavid: We should look how to fix it, since Agama is able to apply the font we have set for mono font-family

Image

lslezak: The "codicon" font contains icons displayed in the search bar (showed after pressing `CTRL` + `F`).

web/src/components/core/ConfigEditor.tsx
```diff
@@ -0,0 +1,97 @@
...
+// import only the base editor to save some space
+import * as monaco from "monaco-editor/esm/vs/editor/editor.api.js";
+import { loader } from "@monaco-editor/react";
+import { useAppearance } from "~/context/appearance";
+import { ROOT } from "~/routes/paths";
+import { isoTimestamp } from "~/utils";
+import { _ } from "~/i18n";
+
+/**
+ * Dialog showing the current installation configuration in JSON format, with
```
dgdavid: NP: It isn't longer a dialog.

lslezak:
Closing, fixed in `master` in #3658

PR #3644, opened 2026-06-17
fix profile integration test

jreidinger:
Fix of integration tests to adapt them to new output of error reporting from CLI.


mvidner (approved):
Thanks, the fix looks good and the test passes for me

PR #3649, opened 2026-06-18
fix: avoid calls to YaST software modules

joseivanlopez:
After the changes introduced by https://github.com/agama-project/agama/commit/908d17fe566c4879f6aed154e07b85893eb9f350, the Agama D-Bus service is using again the *YaST software modules* (e.g., `Yast::Package`) to check whether a package is available in the repositories.  But software is now configured by *agama-software* (rust service) instead of YaST, so calling YaST to check for the availability of the packages would report wrong results. Moreover,  calling to YaST software takes the libzypp lock, so it prevents the usage of *zypper* in the live image, see https://bugzilla.suse.com/show_bug.cgi?id=1268344.

Furthermore, the D-Bus service currently checks the availability of optional packages beforehand, so an optional package is requested to be installed only if it is available. But this approach has some drawbacks:

* The availability of a package cannot be checked until a product is selected or registered.
* The availability of a package might not be realliable if there was any network issue while reading the repositories.

This PR solves these problems by avoiding to check the availability of the packages. Now, all the packages are supposed to be available, but a package can be requested as optional. As result, *agama-software* will install the optional package only if it is available, achieving the same result as before without the need of checking for the availability of packages in advance.

There is a especial treatement for the *os-prober* package, as it is documented in the code.

Requires https://github.com/yast/yast-storage-ng/pull/1430

### DIFF
--- a/service/bin/agamactl
+++ b/service/bin/agamactl
@@ -22,6 +22,10 @@
 
 $LOAD_PATH.unshift File.expand_path("../lib", __dir__)
 
+# Mock some YaST modules like Package.
+agama_y2dir = File.expand_path("../lib/agama/y2dir", __dir__)
+ENV["Y2DIR"] = [ENV.fetch("Y2DIR", nil), agama_y2dir].compact.join(":")
+
 # Set the PATH to a known value
 ENV["PATH"] = "/sbin:/usr/sbin:/usr/bin:/bin"
 
@@ -53,10 +57,6 @@ end
 # @param name [Symbol] Service name
 # @see ORDERED_SERVICES
 def start_service(name)
-  # Mock some YaST modules like Package.
-  agama_y2dir = File.expand_path("../lib/agama/y2dir", __dir__)
-  ENV["Y2DIR"] = [ENV.fetch("Y2DIR", nil), agama_y2dir].compact.join(":")
-
   logger = logger_for(name)
   service_runner = Agama::DBus::ServiceRunner.new(name, logger: logger)
   service_runner.run
--- a/service/bin/agamactl
+++ b/service/bin/agamactl
@@ -1,8 +1,7 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-#
-# Copyright (c) [2022] SUSE LLC
+# Copyright (c) [2022-2026] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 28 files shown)


service/lib/agama/y2dir/modules/Package.rb
```diff
@@ -27,9 +29,35 @@ def main
...
+  private
+
+    OS_PROBER = "os-prober"
+    private_constant :OS_PROBER
+
+    # Whether the product requires "os-prober" package.
+    #
+    # @return [Boolean]
+    def require_os_prober?
+      packages = @storage&.product_config&.mandatory_packages
```
jreidinger: I do not reach you on IRC in time. Problem here is other way. If e.g. TW adds os-prober to mandatory packages, then OsProber in bootloader is ok as it will stop earlier...but if it is in mandatory packages, then software require it and it will fail with solver issue. So TW will be uninstallable on s390 Port.

joseivanlopez: Right now the product is not able to require packages per architecture. I guess the easiest solution is to add a *os_prober* property to the *boot* section of the product, and Agama checks that property, the architecture and the bootloader to decided whether to install os-prober.

jreidinger: Ideally you would add os-prober to optional_packages and not mandatory ones, so it is selected only if available. So perfect fit for such package that is not available everywhere

service/bin/agamactl
```diff
@@ -54,6 +53,10 @@ end
 # @param name [Symbol] Service name
 # @see ORDERED_SERVICES
 def start_service(name)
+  # Mock some YaST modules like Package.
+  agama_y2dir = File.expand_path("../lib/agama/y2dir", __dir__)
+  ENV["Y2DIR"] = [ENV.fetch("Y2DIR", nil), agama_y2dir].compact.join(":")
```
jreidinger: well, this has to be done before any `require "yast"` is called. so ideally it should be as almost first thing done. so ideally around line 24.

service/lib/agama/y2dir/modules/PackagesProposal.rb
```diff
@@ -0,0 +1,55 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+
+# :nodoc:
+module Yast
+  # Replacement for the Yast::PackagesProposal module
+  class PackagesProposalClass < Module
+    def main
+      puts "Loading mocked module #{__FILE__}"
```
jreidinger: I would prefer to call common logging instead of puts...and that reminds me that it would nice to also log all calls into mocked modules, as it makes debugging easier.

service/package/rubygem-agama-yast.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jun 23 11:48:47 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Do not take zypp lock (bsc#1268344).
+- Avoid calling to YaST software and configure products to include
+  os-prober as optinal package.
```
jreidinger: typo optinal

jreidinger (approved):

PR #3657, opened 2026-06-22
fix(web): apply UI language and keymap changes without a page reload

dgdavid:
## Summary

- Changing the installer UI language (and keymap) from the localization dialog now updates the interface right away, without needing a manual page reload.
- The translation catalog is loaded as plain query data and applied to the global translation singleton on every provider render, so the visible texts always match the selected language.
- Language and keymap are now updated through a single `changeL10n` request.
- Adds regression tests covering a language change and the catalog staying in sync across re-renders.

## Motivation

When changing the UI language from the localization dialog (for example English to French), the interface did not update reliably:

- It frequently kept the **previously selected** language (not necessarily English).
- The result was **intermittent**: sometimes it worked, sometimes it did not, depending on timing.
- A manual page reload always applied the correct language.

The product descriptions did update, because they come **already translated from the backend** in the `system` query payload. Only the frontend UI strings (loaded from the `po.<lang>.js` files) were affected.

## Root cause

UI strings are translated through a global singleton (`agama`, used by the `_()` helper), which is updated by calling `agama.locale(catalog)`. Previously that call happened **as a side effect inside the translations query function** (`loadTranslations`), and the query was cached per language.

That made applying the active catalog depend on cache state and on the order in which asynchronous queries resolved, which is fragile in two ways:

1. **A cached language did not re-apply its catalog.** The query function only runs when the query actually fetches. With the cached query served as-is, `agama.locale(...)` was not called again, so the global catalog could stay on a different language than the one being rendered.
2. **An out-of-order load could leave the wrong language active.** The change flow invalidated all translation queries, which could refetch the previously active language while the provider had not re-rendered into the new one yet. That refetch reapplied the old catalog, leaving the global singleton pointing at the wrong language.

A full page reload worked because it loads everything once, in order, with the correct language.

## Fix

Separate **loading** the catalog from **applying** it:

- `loadTranslations` now returns the catalog as data (or `null` when there are no translations), instead of mutating the global singleton.
- `InstallerL10nProvider` applies the catalog with `agama.locale(catalog)` on every render, before the children read it. Because applying no longer depends on the query function running, the visible texts stay in sync with the rendered language regardless of cache state or query timing.
- Language and keymap changes go through a single `changeL10n` request that refreshes the `["system"]` query. The locale update flows back through the provider, which re-renders with the new language and applies the matching catalog. The previous unfiltered translations invalidation is removed.
- `changeLanguage` and `changeKeymap` are kept as thin wrappers over `changeL10n` to avoid duplicating the language-resolution logic.
- The `<html lang>` attribute is now kept in sync on the initial load too (via an effect), not only after a change, which helps assistive technologies.

## Do we still cache the translations?

Yes. This change does **not** reload the translations on every render or on every switch:

- Each language's catalog is fetched **once** and cached by React Query (per `["translations", <language>]`). With refetch-on-mount disabled and no more translation-query invalidation, switching back to an already loaded language reuses the cached catalog instead of fetching again.
- Loading a catalog is a dynamic `import()` of the language chunk, which the browser also caches. So even when React Query drops an inactive language from its cache (after the default timeout), switching back to it does not hit the backend; it reuses the already downloaded chunk.
- Applying the catalog on render (`agama.locale(catalog)`) does no network or disk work. It only points the global translation singleton at the catalog already held in memory, which is cheap.

In short: fetching is still cached and happens at most once per language; the only thing that now runs on each render is the in-memory "make this catalog the active one" step, which is what guarantees the texts match the selected language.

## Known caveat

> [!NOTE]
> 
> **Switching the language causes a brief blink**: the whole interface is intentionally re-rendered so every text picks up the new catalog (the provider remounts its children when the language changes). The transition is fast, so usually nothing more than a quick flash is noticeable. This only happens at the moment of the change, not during normal usage, and it is a **reasonable trade-off compared to the previous behavior, where the interface was left showing stale translations (the previously selected language), sometimes even mixed, until a full page reload**.

## Tests

- [x] `npm run test -- src/context/installerL10n.test.tsx`
- [x] `npm run test -- src/components/core/InstallerL10nOptions.test.tsx`
- [x] `npm run check-types`
- [x] `npm run eslint`
- [x] Manual: change the UI language from the localization dialog and confirm the interface updates immediately, on both the product selection and the overview pages, without a page reload.

### Regression tests added

`src/context/installerL10n.test.tsx`:

- applies the language reported by the backend
- applies the new translations when the language changes
- keeps the translations in sync with the language on every render (this last one fails against the previous side-effect approach and passes with the fix)

---

Co-Authored-By: Knut Anderssen <kanderssen@suse.com>

---

Related to https://trello.com/c/CpKVYDEp/ (protected link)

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jun 22 11:08:29 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Apply UI language and keymap changes right away, without needing
+  a page reload (gh#agama-project/agama#3657).
+
+-------------------------------------------------------------------
 Fri Jun 19 12:52:28 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Fine-tune and extend the Agama appearance system: improve font
--- a/web/src/components/core/InstallerL10nOptions.test.tsx
+++ b/web/src/components/core/InstallerL10nOptions.test.tsx
@@ -52,6 +52,7 @@ const tumbleweed: Product = {
 
 const mockChangeUIKeymap = jest.fn();
 const mockChangeUILanguage = jest.fn();
+const mockChangeUIL10n = jest.fn();
 const mockPatchConfigFn = jest.fn();
 const mockConfigureL10nActionFn = jest.fn();
 const mockStateFn: jest.Mock<Stage> = jest.fn();
@@ -90,6 +91,7 @@ describe("InstallerL10nOptions", () => {
       language: "de-DE",
       changeKeymap: mockChangeUIKeymap,
       changeLanguage: mockChangeUILanguage,
+      changeL10n: mockChangeUIL10n,
     });
   });
 
@@ -172,8 +174,7 @@ describe("InstallerL10nOptions", () => {
       await user.selectOptions(keymapSelector, "English (UK)");
 
       await user.click(acceptButton);
-      expect(mockChangeUIKeymap).toHaveBeenCalledWith("gb");
-      expect(mockChangeUILanguage).toHaveBeenCalledWith("es-ES");
+      expect(mockChangeUIL10n).toHaveBeenCalledWith({ language: "es-ES", keymap: "gb" });
     });
 
     it("allows reusing settings for the selected product", async () => {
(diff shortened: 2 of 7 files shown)


teclator (approved):
LGTM

PR #3658, opened 2026-06-22
Do not download the Monaco editor parts from CDN

lslezak:
Just porting #3643 to `master`, we need this in RC2 anyway...

### DIFF
--- a/web/src/components/core/ConfigDialog.test.tsx
+++ b/web/src/components/core/ConfigDialog.test.tsx
@@ -21,7 +21,7 @@
  */
 
 import React from "react";
-import { screen, waitFor } from "@testing-library/react";
+import { act, screen } from "@testing-library/react";
 import ConfigDialog from "./ConfigDialog";
 import { plainRender } from "~/test-utils";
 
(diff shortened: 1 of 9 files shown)


web/src/components/core/ConfigDialog.test.tsx
```diff
@@ -68,75 +45,22 @@ const renderConfigDialog = () => {
...
   it("calls onClose when the close button is clicked", async () => {
-    fetchSpy.mockResolvedValue({
-      text: () => Promise.resolve(JSON.stringify(mockConfig)),
-    } as Response);
+    const { user } = await act(async () => {
+      return renderConfigDialog();
+    });
 
-    const { user } = renderConfigDialog();
+    // const { user } = renderConfigDialog();
```
dgdavid: left over?

web/src/components/core/ConfigDialog.test.tsx
```diff
@@ -68,75 +45,22 @@ const renderConfigDialog = () => {
...
-      text: () => Promise.resolve(JSON.stringify(mockConfig)),
-    } as Response);
-
-    renderConfigDialog();
-
-    await waitFor(() => {
-      expect(screen.getByLabelText("Download filename")).toHaveTextContent(
-        "agama-config-2026-06-03T10-30-00-000Z",
-      );
+    await act(async () => {
```
dgdavid: I wonder why `act` is needed. I'll not block the PR because of this, just curious about its need.

dgdavid (approved):
LGTM, just two minor comments

web/src/components/core/ConfigDialog.test.tsx
lslezak: The `act()` is not strictly needed in this case. But the test prints a warning that it should be used. It complains because the editor component is loaded dynamically and the test might not be testing what you expect because the is DOM changes after the test is finished (the editor component is displayed later).

web/src/components/core/ConfigDialog.test.tsx
lslezak: Ah, overlooked that... thanks!

web/src/components/core/ConfigDialog.test.tsx
dgdavid: I have pulled the branch and tested it by myself for better understanding.

The warning means React caught the screen still changing after your test thought it  was done. The component shows a spinner first, then swaps in the editor once the lazy import loads, and that swap lands a moment after render returns. A plain assertion checks too early, before the swap, so React complains.

`findBy*`  is to me a btter way to fix it because it waits: it keeps
  re-checking until the element appears, and the await holds the test open through the moment the editor finishes loading, so the late update now happens while the test is still watching. This also fits React Testing Library's philosophy of mimicking the real user, who sees the spinner and then waits for the editor. Wrapping in `act()` instead reaches into React's internals to force-flush the loading, something a user never does. Thus, `findBy*` does both, removes  the warning and keeps the test written from the user's point of view.
 
I'll send a commit with such a change if you don't mind.

dgdavid (approved):
LGTM 👍

PR #3661, opened 2026-06-23
refactor(web): unify localization settings into a single form

dgdavid:
## Summary

- Replace the three standalone localization selector pages (language, keyboard, time zone) with a **single TanStack form** on the Localization page, matching the shape of `system-form`, `registration-form`, and `authentication-form`.
- Each selector is a `SearchableSelectField` typeahead: filter by name, territory / country, code and (for time zones) UTC offset; rows show the human name with the code de-emphasized, and the time zone's current local time + offset.
- Introduce `SearchableSelectField`, a reusable typeahead combobox that powers the three selectors (see its own section below).
- Show the explanatory text (installer language/keyboard adjustable from the top bar) at the top of the
  page, before the fields, instead of below the form.

localhost_8080_ - 2026-06-23T150501 169

## Motivation / context

The Localization page used three `Page.Section` blocks, each linking out to a separate full-page radio-list selector that committed on its own and navigated back. That diverged from the rest of the app, where settings live in one form
with a single submit. This unifies the experience.

## SearchableSelectField (new component)

A reusable typeahead combobox following the W3C "editable combobox with list autocomplete" ARIA pattern, introduced here and used by all three selectors:

- **Full-string search:** each option carries the complete text to match (`filterText`, label included), so it filters by pieces not shown in the visible label (codes, territory); the visual `description` is rendered but never searched.
- **Accent-insensitive matching:** the query and option text are sanitized (lowercased + diacritics stripped), so "ingles" matches "Inglés". **`normalizeQuery`:** an optional last-resort callback to rewrite the query before matching, for the rare case where a natural way of typing should match an option it otherwise would not (here: a typed "UTC+1" matches a zone whose offset is stored as "+1", while a bare "UTC" still finds only Coordinated Universal Time).
- Bracket and list punctuation is treated as a separator too, so a committed selection fed back as the
  query (paste / browser autocomplete) such as "Spanish (Spain)" still matches.
- **`onOpen`:** an optional callback fired when the list opens (here: refresh the time zones' shown local time on open instead of at page load).
- Inline validation error below the control (`aria-invalid` + helper text), debounced filtering, and only one list open at a time.

## Implementation

Best reviewed commit by commit: the history is granular and the messages are self-explanatory. The form lives in `src/components/l10n/l10n-form/` and reuses the existing form abstractions (`useFormSubmit`, `useAppForm`, `withFrozenQuery`).

## Screenshots

| Before | After |
|-|-|
| localhost_8080_ - 2026-06-23T150624 598 | localhost_8080_ - 2026-06-23T150501 169 |
| localhost_8080_ - 2026-06-23T150629 304 | localhost_8080_ - 2026-06-23T150533 144 |
| localhost_8080_ - 2026-06-23T150639 912 | localhost_8080_ - 2026-06-23T150537 290 |
| localhost_8080_ - 2026-06-23T150648 580 | localhost_8080_ - 2026-06-23T150511 197 |
| localhost_8080_ - 2026-06-23T151113 507 | localhost_8080_ - 2026-06-23T151139 271 |
| localhost_8080_ - 2026-06-23T151242 575 | localhost_8080_ - 2026-06-23T151306 473 |
| localhost_8080_ - 2026-06-23T151322 988 | localhost_8080_ - 2026-06-23T151339 541 |

## String freeze

No new user-facing strings: placeholders / no-results reuse the existing translated selector strings, and the required-field message reuses the generic `_("Value is required")`.

These reused strings are also inconsistent with the selector labels: the field is labelled "Language" but its placeholder and empty state talk about the "locale" (`_("Filter by language, territory or locale code")`, `_("None of the locales match the filter.")`), and likewise "Keyboard" drifts to "keymap". Once the freeze lifts it would be nice to use clearer, generic and
consistent wording, e.g. a single `_("Type to filter")` placeholder and a generic empty state like `_("No results found")`, so the supporting text no longer diverges from the label.

## Follow-ups (not blocking)

- Screen readers announce the code as part of the option's name; hide it from the spoken name, or expose it as a secondary (described-by) description instead.
- ~The code uses the disabled text color, whose dark-mode contrast is likely below WCAG; improve it (possibly via a dedicated Agama token).~ Done at #3662.
- Ask for a  backend change to structured keymap fields, so the keyboard row can show the base name and the code separately instead of parsing a localized string.
- Remove the now-unused `Timezone.utcOffset` model field.
- Post-freeze string work (see above): 
  - generic/consistent placeholders, a resting empty-state prompt, and dedicated per-field required messages.
  - Shorter info text: find a clearer, shorter message for the installer-settings note. It should straightforward, easy to read, while keeping the link to the top-bar option (or maybe just mentioning it now that they are close), and stay conditional on local vs non-local, e.g. _("To change the installer's language and 
  keyboard, use the [option] in the top bar.") when local and _("To change the installer's language, use 
  the [option] in the top bar.") otherwise.
- Decide whether the displayed time should follow the user's locale (currently  fixed to en-US 24h).

---

Assisted by Claude Sonnet 4.5


web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue Jun 23 11:49:50 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Unify the localization settings into a single form: replace the
+  separate language, keyboard and time zone selection pages with one
```
teclator: NP: line too long ;)

teclator (approved):
LGTM

web/package/agama-web-ui.changes
dgdavid: Fixed.

web/package/agama-web-ui.changes
teclator, line 1: Better, thanks! LGTM

PR #3663, opened 2026-06-24
Agama 22 fix inst finish

jreidinger:
## Problem

When there is in `inst.finish` parameter used, then installation ends prematurly as it does not wait for finish of installation task.

## Solution

This is just minimal quick fix as issue is ship stopper that move processing and executing finish to the last task of install task.

Note: more complex fix by @imobachgs is at https://github.com/agama-project/agama/compare/agama-22...fix-inst-finish?expand=1

## Testing

- *Tested manually*

### DIFF
--- a/rust/agama-manager/src/actions.rs
+++ b/rust/agama-manager/src/actions.rs
@@ -249,11 +249,11 @@ impl InstallAction {
                     .call(progress::message::SetStage::new(Stage::Finished))
                     .await
                     .map_err(TaskError::from_error)?;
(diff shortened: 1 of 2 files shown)


lslezak (approved):
Looks reasonable

imobachgs:
I will add a pull request with the other fix. I would not use the word "complex" (although it certainly is), but "complete". This PR is enough to fix the P1, but IIUC it does not fix problems like getting to the "failure" status or trying to configure the system when the software installation failed.

PR #3664, opened 2026-06-24
Agama22 full iso unattended fix

jreidinger:
## Problem

With full medium and unattended installation agama reports that product is missing. But it is caused that it does not add predefined repositories during initial run and report issue.

More detailed explanation is that previously software even in unattended mode compute proposal multiple times. At least twice as it compute its configuration and then storage ask for software. The first computation do not have available software info from /system as it was computed after initial configuration. But the second computation has it already, so predefined repositories are added to it during that second run and all works. The change that breaks it was https://github.com/agama-project/agama/pull/3585 which now cause only single run of proposal. The change itself is good, but reveal this limitation.

## Solution

Add more logging that was useful when debugging wanted state.
And separate computing system and setting predefined repositories. Now predefined repositories are available from start of software service and is added also during initial computation of proposal.
The patch goal is to be minimal for agama 22. For master follow up PR will be opened with more cleaning.

## Testing

- *Tested manually* - downloaded full iso from openqa they use for testing, then `cargo build --release` and with resulting binaries I modify that full iso - `sudo ~/prace/agama-devtools/live-iso/builder/agama-iso-builder --copy-root agama-web-server /usr/bin/agama-web-server --copy-root agama /usr/bin/agama SLES-16.1-Full-x86_64-Build48.1.install.iso`
- also TW build was manually tested just to be sure it is not completely broken now - https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/agama22-full-iso-unattended-fix/images/iso/

### DIFF
--- a/rust/agama-software/src/model.rs
+++ b/rust/agama-software/src/model.rs
@@ -89,6 +89,8 @@ pub trait ModelAdapter: Send + Sync + 'static {
         l10n: Handler<agama_l10n::Service>,
         progress: Handler<progress::Service>,
     ) -> Result<WriteIssues, service::Error>;
+
+    fn predefined_repositories(&self) -> Vec<Repository>;
 }
 
 /// [ModelAdapter] implementation for libzypp systems.
@@ -128,6 +130,10 @@ impl Model {
 impl ModelAdapter for Model {
     fn set_product(&mut self, product_spec: ProductSpec) {
         self.selected_product = Some(product_spec);
+    }
+
+    fn predefined_repositories(&self) -> Vec<Repository> {
+        self.predefined_repositories.clone()
     }
 
     async fn write(
--- a/rust/agama-software/src/model/state.rs
+++ b/rust/agama-software/src/model/state.rs
@@ -26,9 +26,12 @@ use std::collections::HashMap;
 use std::fmt;
 
 use agama_utils::{
-    api::{self, software::{
-        Config, PatternsConfig, ProductConfig, RepositoryConfig, SoftwareConfig, SystemInfo,
-    }},
+    api::{
+        self,
+        software::{
+            Config, PatternsConfig, ProductConfig, RepositoryConfig, SoftwareConfig, SystemInfo,
+        },
+    },
     kernel_cmdline::KernelCmdline,
     products::{ProductSpec, UserPatternSpec},
 };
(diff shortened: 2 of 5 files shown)


mvidner:
Thanks!

> * _Tested manually_

Please say how you tested it.

Also, unit tests now fail.

imobachgs:
I would like to describe why the old approach (`add_system_config`) does not work anymore. Not to mention that, at least for `master`, we might need to do some clean-up.

jreidinger:
@imobachgs it is in commit message body https://github.com/agama-project/agama/pull/3664/changes/cf019876e0eb1cc613785f42af3b6574f8c55f08 ( not sure if it is enough ). I think for cleaning we should not have with_system at all, as system and proposal is computed after initial computation of wanted state.

jreidinger:
@imobachgs regarding master I would like to do it as follow-up PR as we need tomorrow morning do agama-22 merge to master and release first to factory and then create PR to SLFO to rebuild RC1.

jreidinger:
@imobachgs thinking more carefully about reason I would say that main reason was your and @joseivanlopez change in software service as previously we depends on multiple calls of software and its solver. So basically we have software proposal and **then** storage one which has package requirements for software and cause recompute of wanted state. And that recompute cause that system was already initialized from previous call. And as result this second call of wanted state properly initialize predefined repositories, but it no longer happens and it needs to be done in first and only computation of software.

imobachgs:
In my opinion, we should fix it to do not need two calls. The solver should run after the resolvables are set. It is fine as a hotfix, but I think we should revisit it.

lslezak:
JFYI: It fixes the problem also in manual installation.

jreidinger:
@imobachgs I am not sure if I get your comment. The fix is exactly to not do it in two calls. We store in model predefined repositories and use it for first call. So now after fixed version even the first computation of wanted state contains repositories, which was not case before.
I plan to do cleaning PR after this hotfix is merged to master...mainly

1. for wanted state system should not be needed
2. system probably do not need to list all repositories. If needed, I think it should be in proposal and not system
3. refactor a bit computing wanted state as now it is really huge.

imobachgs:
OK, I understand now. I guess it will be fine after cleaning-up the code and removing the unneeded parts. Thanks!

lslezak (approved):
Looks OK to me

PR #3665, opened 2026-06-24
Break a systemd service dependency cycle

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1268901
- When systemd detects a cycle in the service dependencies then it deletes some services to break the cycle. But that services are missing in the boot sequence and that can have nasty side effects.

```
Jun 24 15:03:54 localhost systemd[1]: NetworkManager-initrd.service: Found ordering cycle: ignition-enable-network.service/start after ignition-fetch-offline.service/start after basic.target/start after sysinit.target/start after cryptsetup.target/start after systemd-ask-password-console.path/start after plymouth-start.service/start after initrd-nmtui.service/start after network-online.target/start after network.target/start after NetworkManager-initrd.service/start - after ignition-enable-network.service
Jun 24 15:03:54 localhost systemd[1]: NetworkManager-initrd.service: Job cryptsetup.target/start deleted to break ordering cycle starting with NetworkManager-initrd.service/start
```

## Solution

- Remove the `Before=plymouth-start.service` dependency which was used to start the interactive TUI tool before the splashscreen (otherwise it would not be visible).
- Instead of ordering the services quit plymouth before starting the TUI tool.

## Notes

- As written in the comment, it should be enough to use the `plymouth hide-splash` and `plymouth show-splash` pair in the pre/post scripts but unfortunately then the `nmtui` ignores pressing the `Enter` key and cannot be used. It is strange as the `dialog` tool used before running `nmtui` works fine.
- As a workaround quit the splash screen completely. As this boot option will be used rarely it is an acceptable solution.

## Testing

- Tested manually, no dependency cycle is reported and the `nmtui` tool works fine.

### DIFF
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Wed Jun 24 15:19:28 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
 
-- Break a systemd dependency cycle (systemd deletes some services
+- Break a initrd-nmtui systemd dependency cycle (systemd deletes some services
   to break the cycle, that services are then missing and it might
   trigger the emergency shell) (bsc#1268901)
 
--- a/live/src/agama-installer.changes
+++ b/live/src/agama-installer.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jun 24 15:19:28 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
+
+- Break a systemd dependency cycle (systemd deletes some services
+  to break the cycle, that services are then missing and it might
+  trigger the emergency shell) (bsc#1268901)
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


live/src/agama-installer.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jun 24 15:19:28 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
+
+- Break a systemd dependency cycle (systemd deletes some services
```
mvidner:
```suggestion
- Break a initrd-nmtui systemd dependency cycle (systemd deletes some services
```

I find it better to use specific strings in .changes that people may find in logs and duplicate bug bug reports

mvidner (approved):
LGTM, with an improvement for .changes

live/src/agama-installer.changes
teclator: NP: the suggestion made the line too long :P

PR #3667, opened 2026-06-25
fix(web): improve accessibility of the software patterns selection

dgdavid:
**TL;DR:** The pattern categories are now exposed as real, accessible sections (groups) labelled by their heading, so the category a pattern belongs to is conveyed programmatically instead of only visually.

## The long story

This came out of QA testing the patterns selection page (`software/patterns/select`): some patterns showed up with the same display name under different categories (**most likely a patterns-definition issue**), and it was hard to tell which one was the right one to pick, the one the tester was actually after, because nothing indicated which category each entry belonged to. A real screen-reader user could face the same ambiguity.

Looking into it surfaced that the categories were rendered as visual headings only, not as real, navigable sections: a screen reader user got headings to jump between but no programmatic grouping tying a category to its checkboxes. Exposing each category as a labelled group makes the intended pattern identifiable by its category and closes the underlying accessibility gap. Along the way we also hardened the checkbox ids so that even patterns sharing the exact same name (not just the same display name) across categories cannot break the markup.

### Semantic groups (a11y)

- Each category is wrapped in a `section` with `role="group"` and `aria-labelledby` pointing at its heading. This is the ARIA equivalent of a `fieldset`/`legend` for a set of related form controls, and it is announced when entering the category's checkboxes. `role="group"` (rather than a `region` landmark) keeps the landmark list clean while heading navigation still works.
- The `h3` headings are preserved.

### Unique checkbox ids

- The checkbox derived its DOM id, its label's `htmlFor` target, and its React key from the pattern name. If the backend lists the same pattern name under more than one category, those collide (duplicate ids make every label resolve
  to the first input; React keys clash).
- The DOM id and key are now derived from the category and position, while the **form field stays keyed by the pattern name** (the backend identity). The backend selects patterns by name, so duplicate entries represent the same selection and now stay consistent without breaking the markup. Distinct patterns that merely share a visible label remain independent and are told apart by their category group.

## Tests

* Updated existing tests (adding new examples for increase coverage)

## Screenshots

Not needed, visually all still the same despite these markup changes.

## Additional info: the collapsible sections experiment

While working on this, we prototyped making the categories **expandable/collapsible** (an accordion, all expanded by default). While it might look nice at first sight, on review we realized it carried enough trade-offs for not including it now:

- **ARIA semantics conflict / landmark proliferation.** PatternFly's `ExpandableSection` hardcodes `role="region"` (a landmark) on each panel. That duplicates the `role="group"` semantics from this PR and adds one landmark per  category, i.e. the landmark proliferation we deliberately avoided. Resolving it means giving up the cleaner grouping in favour of an accordion-region model.
- **Sticky header edge cases.** The category headers are sticky. Collapsing a category the user had scrolled past let its sticky header detach and jump out of view, requiring JS scroll compensation (`scrollIntoView` + `scroll-margin`). That logic depends on layout and the scroll container, is not unit-testable (jsdom has no layout), and is fragile across viewports.
- **Filtering interaction complexity.** Collapsing interacts awkwardly with the existing filter: matches can hide inside a collapsed category. Handling it well needs extra signaling (e.g. a highlighted per-category match indicator) and a
  decision on whether to force-expand on filter vs respect the user's choice, all of which add state and surface area for bugs.
- **Translation freeze leaves it incomplete.** A usable collapsible UX wants bulk controls near the filter ("Expand all" / "Collapse all", and "Expand categories matching filter"). Those are new strings, which the current translation freeze  blocks, so the feature would ship half-implemented (only a FIXME placeholder).
- **Extra state to maintain.** Per-category collapse state, default-expanded handling, and its coupling with filtering add ongoing complexity for a feature that is largely cosmetic given the existing filter already narrows the list.

Conclusion: collapsible categories read more like a low-priority nice-to-have for a next major version (if really useful only) than something to land during the RC phase. The filter we already have covers the core need of narrowing a long list. **This PR therefore ships only the semantic/accessibility fix, and the exploration is kept aside in case we revisit it later.**

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 25 11:41:21 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Expose the software pattern categories as accessible groups,
+  so a pattern can be identified by its category
+  (gh#agama-project/agama#3667).
+
+-------------------------------------------------------------------
 Tue Jun 23 11:49:50 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Unify the localization settings into a single form: replace the
--- a/web/src/components/software/patterns-form/Form.test.tsx
+++ b/web/src/components/software/patterns-form/Form.test.tsx
@@ -84,6 +84,16 @@ describe("SoftwarePatternsSelection", () => {
     expect(headings).toHaveLength(3);
   });
 
+  it("exposes each category as a group labelled by its heading", async () => {
+    installerRender(<SoftwarePatternsSelection />);
+    const groups = await screen.findAllByRole("group");
+    expect(groups).toHaveLength(3);
+    // Each category checkbox set is reachable as a group labelled by its name.
+    screen.getByRole("group", { name: /Graphical Environments/ });
+    screen.getByRole("group", { name: /Base Technologies/ });
+    screen.getByRole("group", { name: /Desktop Functions/ });
+  });
+
   it("renders each category containing its patterns in order", async () => {
     installerRender(<SoftwarePatternsSelection />);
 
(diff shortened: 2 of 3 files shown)


teclator (approved):
LGTM

PR #3668, opened 2026-06-25
fix(rust): use correct profile schema path

joseivanlopez:
## Problem

The relative path for the profile schema is wrong, so the installation path (*/usr/share/agama/schema*) is always used. This is not good when running Agama from sources.

## Solution

Calculate the profile schema path based on the source file path.

### DIFF
--- a/rust/agama-cli/src/config.rs
+++ b/rust/agama-cli/src/config.rs
@@ -143,7 +143,7 @@ fn build_config_generate_cmd() -> Command {
         // TRANSLATORS: CLI help for: agama config generate
         gettext("Generate and print a native Agama JSON configuration from any kind and location");
     // TRANSLATORS: CLI help for: agama config generate (details)
-    let long_about = make_long(&about, &gettext("\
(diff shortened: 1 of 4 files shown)


rust/agama-cli/src/config.rs
```diff
@@ -154,7 +157,9 @@ fn build_config_generate_cmd() -> Command {
         - URL (including AutoYaST specific schemes)\n\
         \n\
         For an example of Jsonnet-based profile, see\n\
-        https://github.com/openSUSE/agama/blob/master/rust/agama-lib/share/examples/profile.jsonnet"));
+        https://github.com/openSUSE/agama/blob/master/rust/share/examples/profile.jsonnet",
```
jreidinger: sadly this kills translations, so I worry we cannot do it after string freeze. ( btw it was bad idea to have it part as translatable string.

joseivanlopez: Ohh, right!

jreidinger (approved):

PR #3671, opened 2026-06-25
Software system refactor

jreidinger:
Follow up of https://github.com/agama-project/agama/pull/3664 to polish a bit fix with a bit of refactoring and factor out strange behavior when system is not set.


joseivanlopez (commented):
Looks good. Only changelog entry is missing.

joseivanlopez (approved):
LGTM

PR #3672, opened 2026-06-26
bsc#1268803: fix `inst.finish` and installation tasks handling

imobachgs:
## Problems

Fixes installation finalization to execute after all tasks complete, preventing premature termination when using the `inst.finish` kernel parameter. See [bsc#1268803](https://bugzilla.suse.com/show_bug.cgi?id=1268803).

Previously, `FinishAction` executed immediately after spawning tasks rather than waiting for them to complete. This moves finalization into a task that depends on all critical installation tasks.

This issue was solved with #3663 for Agama 22 but there were still some problems:

- `IPMI` was not reported properly in case of error.
- If a step of the installation failed (e.g., software installation), the rest of the process sitll runs (e.g., the configuration of the target system).

This change address those issues too.

## Solution

- Add `run_always()` API to TaskManager for tasks that execute regardless of dependency success/failure
- Regular tasks are now cancelled if dependencies fail
- Track succeeded/failed tasks separately (previously only "completed")
- Move `FinishAction` execution into a finalization task that waits for all installation tasks
- Execute `FinishAction` on success or notify IPMI on failure

## Open questions

* Now tasks run if the dependencies did it. Could it be a problem for SetConfig? If that's the case, we might change the default behavior to be `run_always`.

### DIFF
--- a/rust/agama-manager/src/actions.rs
+++ b/rust/agama-manager/src/actions.rs
@@ -251,13 +251,6 @@ impl InstallAction {
                     .call(progress::message::SetStage::new(Stage::Finished))
                     .await
                     .map_err(TaskError::from_error)?;
-                // Here we use the default Stop, so only an explicit kernel parameter does something.
-                // Its main use case is interactive installation, as auto-installation calls the explicit
-                // finish action with its own default.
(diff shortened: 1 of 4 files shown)


joseivanlopez (approved):
LGTM

PR #3674, opened 2026-06-26
refactor(web): make progress-tracking query waits explicit per page

dgdavid:
## Problem

The current progress tracking implementation has a design flaw where all pages automatically wait for the "common proposal keys" (i.e., `proposal` and `extendedConfig`) to refetch, even when those keys are not relevant to the operation being tracked.

This causes a bug with iSCSI discovery in which the progress backdrop displays "Refreshing data..." indefinitely because:

1. The discovery operation completes and signals the `iscsi` scope.
2. The progress tracking automatically waits for `COMMON_PROPOSAL_KEYS` to refetch.
3. But discovery doesn't trigger a proposal change, so those queries never refetch.
4. The backdrop never dismisses, blocking user interaction.

See https://bugzilla.suse.com/show_bug.cgi?id=1269224.

As proposals come and go, the assumption that every operation affects the proposal becomes increasingly incorrect. Moreover, the upcoming architectural changes where some proposals might disappear make this assumption even more problematic, see check https://trello.com/c/jTgiokbJ.

## Solution

The solution is to make query-key subscriptions fully explicit, so each page declares exactly the queries it depends on. In other words, remove COMMON_PROPOSAL_KEYS and replace the `ensureRefetched` prop (which adds to the default common keys) with an `awaitQueriesRefetch` prop (which specifies all keys explicitly, with no defaults).

While the proposed change introduces additional verbosity by requiring each consumer to explicitly specify the query keys it depends on, it also makes those dependencies explicit, resulting in a more robust and future-proof design.

Note: some flaws were detected in the progress tracking, but it is out of the scope of this PR, see https://trello.com/c/3WLwI010/2400-improve-progress-tracking.

---

Assisted by Claude Sonnet 4.5

### DIFF
--- a/service/lib/agama/dbus/storage/dasd.rb
+++ b/service/lib/agama/dbus/storage/dasd.rb
@@ -54,7 +54,6 @@ module Agama
           dbus_method(:SetConfig, "in serialized_config:s") do |serialized_config|
             configure(serialized_config)
           end
-          dbus_signal(:SystemChanged, "serialized_system:s")
           dbus_signal(:ProgressChanged, "serialized_progress:s")
           dbus_signal(:ProgressFinished)
           dbus_signal(:FormatChanged, "serialized_summary:s")
@@ -88,6 +87,21 @@ module Agama
           start_progress(1, _("Configuring DASD"))
           manager.configure(config_json)
 
+          # The "return if unchanged" guard has been removed from the methods below to always
+          # emit the corresponding signal.
+          #
+          # Since signals do not carry payloads yet, the UI cannot update the query cache
+          # directly and must refetch after receiving the signal. Without emitting the signal,
+          # the related queries are never invalidated and never refetched, leaving the progress
+          # overlay blocked indefinitely.
+          #
+          # The overlay intentionally waits until fresh data arrives before unblocking, since
+          # data can take time to appear after progress completes. Dismissing it
+          # earlier would cause flickering and leave users able to interact with stale data.
+          #
+          # It can be reverted (and UI progress adapted accordingly) when signals carry payloads
+          # that allow the UI to update the cache directly, removing the need to wait for a
+          # refetch as part of progress completion.
           update_serialized_system
           update_serialized_config
 
@@ -113,17 +127,16 @@ module Agama
         # Updates the system info if needed.
         def update_serialized_system
           serialized_system = serialize_system
-          return if self.serialized_system == serialized_system
+          # return if self.serialized_system == serialized_system
 
           # This assignment emits a D-Bus PropertiesChanged.
           self.serialized_system = serialized_system
-          self.SystemChanged(serialized_system)
         end
 
         # Updates the config info if needed.
         def update_serialized_config
           serialized_config = serialize_config
-          return if self.serialized_config == serialized_config
+          # return if self.serialized_config == serialized_config
 
           # This assignment emits a D-Bus PropertiesChanged.
           self.serialized_config = serialized_config
--- a/service/lib/agama/dbus/storage/dasd.rb
+++ b/service/lib/agama/dbus/storage/dasd.rb
@@ -127,7 +127,6 @@ module Agama
         # Updates the system info if needed.
         def update_serialized_system
           serialized_system = serialize_system
-          # return if self.serialized_system == serialized_system
 
           # This assignment emits a D-Bus PropertiesChanged.
           self.serialized_system = serialized_system
@@ -136,7 +135,6 @@ module Agama
         # Updates the config info if needed.
         def update_serialized_config
           serialized_config = serialize_config
-          # return if self.serialized_config == serialized_config
 
           # This assignment emits a D-Bus PropertiesChanged.
           self.serialized_config = serialized_config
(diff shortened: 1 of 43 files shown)


dgdavid:
> 1. Do network operations actually affect the proposal? Need to verify if `awaitFreshQueries: [SYSTEM_QUERY_KEY]` is enoguh or if `[PROPOSAL_QUERY_KEY, EXTENDED_CONFIG_QUERY_KEY, SYSTEM_QUERY_KEY]` is required

@teclator  :point_up:

teclator:
Network is some those settings which does not need a proposal but just the extended config. In fact, the extended config and the system config should be the same. Before the HTTP API v2 we had calls to modify the config without applying it, but currently when we update or set the config we are modifying and applying it which sync the extended config with the system config after the write, therefore it should be the same unless there is some write error.

web/src/hooks/use-track-queries-refetch.ts
```diff
@@ -66,12 +65,9 @@ import { useQueryClient } from "@tanstack/react-query";
...
  *   memoize queryKeys in the parent component if it changes frequently.
  *
- * @see {@link https://tanstack.com/query/latest/docs/react/reference/QueryObserver}
+ * @see {@link https://tanstack.com/query/latest/docs/reference/QueryCache#querycachesubscribe}
  */
-function useTrackQueriesRefetch(
-  queryKeys: readonly string[],
-  onSuccess: (startedAt: number, completedAt: number) => void,
-) {
+function useTrackQueriesRefetch(queryKeys: readonly string[], onSuccess: () => void) {
```
jreidinger, line 70: hmm, this method change signature to no longer add those startedAt and completedAt, but it is still referenced.
e.g. at

- https://github.com/agama-project/agama/pull/3674/changes#diff-eb65ae9498680eeda88340b408a16eeaad99718992275169b604f454eeec76c7R274
-  https://github.com/agama-project/agama/pull/3674/changes#diff-ddac3c29784bca0c354d0d8d20d88ed53312bd403d1541f31a7f6d840916aaf7R213 (multiple times there)

I kind of expect that strict typescript should complain there.

service/lib/agama/dbus/storage/dasd.rb
```diff
@@ -113,17 +127,16 @@ def register_callbacks
         # Updates the system info if needed.
         def update_serialized_system
           serialized_system = serialize_system
-          return if self.serialized_system == serialized_system
+          # return if self.serialized_system == serialized_system
```
jreidinger: dead code. I think it should not be in final code to merge.

web/src/hooks/use-track-queries-refetch.ts
joseivanlopez: I think you got confused with the diff. The `startedAt` and `completedAt` params were removed from the `onSuccess` callback, but `startedAt` is still used by the hook `useTrackQueriesRefetch`.

joseivanlopez: But, the example in the documentation was wrong. I have updated it.

service/lib/agama/dbus/storage/dasd.rb
joseivanlopez: We have kept those checks commented, because there is chance to recover them at some point, see comment https://github.com/agama-project/agama/pull/3674/changes#diff-46ac2f49bea7c4bb80acd8be8d32cb71c8690ec88c0e8219ad212896eb7a156fR90-R104

joseivanlopez: We did the same for similar methods in storage, iSCSI and zFCP.

jreidinger: to be honest I am not big fan of it, as it can be always recovered by git itself when you use dedicated git commit or even tag for it. But it is not critical for me.

web/src/hooks/use-track-queries-refetch.ts
jreidinger: yes, but it is used in other test at https://github.com/agama-project/agama/pull/3674/changes#diff-ddac3c29784bca0c354d0d8d20d88ed53312bd403d1541f31a7f6d840916aaf7R75
and then that mockRefetchCallback is called multiple times with `startedAt` and `completedAt` params which is wrong. see it links I posted above.

joseivanlopez: Ahh, somehow the links you pasted did not properly go to the proper place and I did not see the usage. I will fix it. Thanks!

service/lib/agama/dbus/storage/dasd.rb
joseivanlopez: Ok, so let's remove all them.

jreidinger (approved):

joseivanlopez:
### Bug introduced and fixed during the refactor

The initial commits in this PR inadvertently broke `extendedConfig` invalidation due to a circular import timing issue. `useProposalChanges` built its invalidation list from a module-level `COMMON_PROPOSAL_KEYS` array, but `EXTENDED_CONFIG_QUERY_KEY` was imported from a module that was still initializing at load time, causing its value to be captured as `undefined`. As a result, the invalidation loop invalidated `[undefined]` (effectively nothing), leaving `extendedConfig` never refetched after proposal changes.

This was discovered during manual testing of the storage page: the backdrop remained stuck on "Refreshing data..." while waiting for `extendedConfig`. It was fixed in 3628d8c by building the key list inside the event handler, where all imports have already been resolved.

Root cause: the `export * as …` barrels in `config.ts` re-export submodules that import back from `config.ts`, creating this initialization-order trap.

jreidinger (approved):
still looks good.

PR #3676, opened 2026-06-28
[DRAFT] [DO NOT REVIEW YET] refactor(web): migrate space policy page to the new form stack

dgdavid:
Replaces the space policy ToggleGroup plus info-popover table with a form built on latest shared form stack: one field per device. It uses a `DropdownField` when there is a real choice and a `ReadOnlyField` when the device is reused and there are no options. Constraints (mount point, or why a device can't be shrunk) move inline, in the same note style `SizeFields` at other forms already uses.

## Why this proposal

It favors usability and discoverability. Each device is a self-contained field, so its actions and the reasons behind them sit right next to it, easy to track. In the old table the user had to follow a row across columns to match a device with its actions, which is harder to scan and worse for accessibility. It is also a consistency move: it now uses the same field components, label and hint patterns, and keyboard and screen reader semantics, that latest updated storage forms. The tradeoff is that it may take more vertical room than the table, and that is a tradeoff we should be happy to make for clarity.

### Clearer available actions, no hidden information

Two problems with the old toggle:

- Disabled actions are hard to spot. Unavailable actions show up as disabled buttons, which users can only tell apart by color and contrast. That is easy to miss, and worse across themes or for low-vision users. Here, an action is either offered or not present at all; when a device has no choice it becomes a read only field.
- The reasons were hidden behind a tooltip. Important information, like why a device can't be shrunk, sat behind a hover tooltip. Now it is shown inline, right under the field, where it can't be missed.

### This is a starting point, not the final form

We will keep iterating on forms as part of evolving Agama and expect to find better patterns over time. One things is explicitly deferred:

- Wording. "Do not modify" and "Allow shrink" are kept as the former (untranslated) toggle labels to respect the string freeze. Once it is lifted we can adopt shorter, clearer wording, and wire proper translations (see the in-code note about the pre-existing i18n gap).

### Before

localhost_8080_ (1)

### After

localhost_8080_


dgdavid:
@ancorgs 

I'm closing this for now, as there are better plans for this topic. If those plans don't prosper, we can always resurrect this later.

PR #3684, opened 2026-06-30
fix(web): resolve duplicate Mode labels in system page a11y tree

dgdavid:
## Summary

- Fix duplicate "Mode" accessible names in SystemPage
- Use `aria-labelledby` to create composite accessible names without adding translations
- Result: "Mode" labels now have unique accessible names ("Hostname Mode" and "Time Synchronization Servers Mode")

## Context

@jknphy reported that Puppeteer automated tests were finding two elements when trying to select the "Mode" dropdown in SystemPage. The issue was that both the Hostname and NTP sections had dropdowns with the same accessible name "Mode", making them indistinguishable to:

- Screen reader navigation by form controls
- Voice control users ("click Mode" → ambiguous)
- Automated testing tools (Puppeteer)

Changing visible labels is not an option since they could sound redundant with the fieldset legend. Moreover, at this moment Agama is currently in string freeze phase.

## Solution

The fix leverages `aria-labelledby` to create composite accessible names by combining parent context (fieldset legends) with field labels:

**Changes to components:**

1. **Fieldset**: Added optional `legendId` prop
   - Allows fieldset legends to be referenced by descendant elements
   - When provided, adds `id` attribute to the `<legend>` element

2. **DropdownField**: Added optional `additionalLabelId` prop
   - Wraps label in `<span id="{fieldName}-label">` for referenceability
   - When `additionalLabelId` is provided, combines it with the label ID: `aria-labelledby="{additionalLabelId} {fieldName}-label"`
   - When not provided, works as before (no aria-labelledby)

3. **HostnameFields & NtpFields**: Wire up the legend and label IDs
   - Set `legendId` on Fieldset
   - Pass same ID as `additionalLabelId` to DropdownField

**Result:**
- Visible labels: "Mode" (unchanged - no new translations needed)
- Accessible names:
  - Hostname dropdown: "Hostname Mode"
  - NTP dropdown: "Time Synchronization Servers Mode"
- Both dropdowns are now uniquely identifiable by their accessible names

## Before

Screenshot_2026-06-30_11-05-08

## After

Screenshot_2026-06-30_11-04-41

## Related

This approach can be reused for other cases where:
- Fields have identical labels within the same page/form
- String freeze prevents changing visible labels
- Parent context (fieldset, section) can disambiguate

### DIFF
--- a/web/src/components/form/DropdownField.tsx
+++ b/web/src/components/form/DropdownField.tsx
@@ -46,6 +46,17 @@ type DropdownFieldProps<T> = {
   helperText?: React.ReactNode;
   isDisabled?: boolean;
   /**
+   * Optional additional ID to include in the composite accessible name.
+   *
+   * When provided, this ID will be combined with the field's label ID to
+   * create a composite aria-labelledby value. Useful for referencing parent
+   * elements like fieldset legends to disambiguate fields with identical labels.
+   *
+   * Example: passing a legend ID "hostname-legend" when label is "Mode"
+   * creates accessible name "Hostname Mode".
+   */
+  additionalLabelId?: string;
+  /**
    * Render prop for content that depends on the current value, such as
    * nested fields that appear when a specific option is selected.
    */
@@ -96,6 +107,7 @@ export default function DropdownField<T extends string>({
   options,
   helperText,
   isDisabled = false,
+  additionalLabelId,
   children,
 }: DropdownFieldProps<T>) {
   const field = useFieldContext<T>();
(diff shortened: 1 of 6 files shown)


web/src/components/form/DropdownField.tsx
```diff
@@ -45,6 +45,17 @@ type DropdownFieldProps<T> = {
...
   isDisabled?: boolean;
+  /**
+   * Optional additional ID to include in the composite accessible name.
+   *
+   * When provided, this ID will be combined with the field's label ID to
+   * create a composite aria-labelledby value. Useful for referencing parent
+   * elements like fieldset legends to disambiguate fields with identical labels.
+   *
+   * Example: passing a legend ID "hostname-legend" when label is "Mode"
+   * creates accessible name "Hostname Mode".
```
joseivanlopez: Is the example right?

dgdavid: Maybe too coupled to the change made here. I will update it for better illustration.

joseivanlopez (approved):
LGTM

dgdavid:
Post-merged: I've realized that same situation might happen with other fields. It is the case of "Password" input in the Authentication page. 

A follow-up PR for add support to all fields will be created soon.

PR #3685, opened 2026-07-01
feat(web): allow disambiguate  form fields accessible name

dgdavid:
## Motivation

As a follow-up to #3684, which this PR actually supersedes, adds a `useFieldLabel` hook that lets any form field adjust its accessible name in a flexible way, reusing existing text nodes as a prefix for its label, among other options, to disambiguate fields that share a generic label without introducing new translatable strings.

The previous PR only fixed two "Mode" dropdowns on the System page that shared the accessible name "Mode", making them indistinguishable to screen readers, voice control, and automated tests. The fix lived only in `DropdownField`, despite the same collision happening elsewhere. On the Authentication page, the administrator and root sections each render "Password", "Password confirmation" and "SSH Public Keys", so those names are ambiguous across sections. Changing the visible labels is not an option (they would read as redundant next to the section legend, and also Agama is in a translation freeze at this moment), so the disambiguation has to happen in the accessible name.

This PR generalizes the one-off approach into a reusable hook and rolls it out to every field that renders a label.

### `useFieldLabel` hook

Resolves a field's accessible name from three independent options:

- `aria-label`: a literal name (consumer owns it).
- `aria-labelledby`: referenced element ids (consumer owns it).
- `labelPrefixedBy`: one or more element ids whose text is placed before the field's own label, producing a name like "Hostname Mode". Unlike `aria-labelledby`, the visible label is kept in the name (WCAG 2.5.3).

An explicit `aria-labelledby` wins over `labelPrefixedBy`. With no options the hook is a no-op and the control keeps the label association `FormGroup` already provides.

### Field rollout

All label-bearing fields accept the options: `TextField`, `MaskedField`, `NumberField`, `EmailField`, `SuggestionsTextField`, `DropdownField`, `RadioGroupField`, `SearchableSelectField`, `ArrayField`, and `CheckboxField` (literal overrides only, since its visible label is already its accessible name).

Fields with their own naming shape are handled in place: the radio group names its `role="group"` container, and the searchable select and array field keep their existing default names and only refine them on request. The array field also prefixes its entries list so it matches its input.

### Consumers

- System page: the Hostname and NTP "Mode" dropdowns now derive their legend id from `useId` instead of a hand-picked string.
- Authentication page: the shared password and SSH-key fields are prefixed with their section legend, giving names like "Administrator account Password" and "Root account SSH Public Keys".

 ## Additional changes

- `SearchableSelectField` now types `placeholder`, `emptyPlaceholder` and `noResultsText` as `TranslatedString` instead of plain `string`, so untranslated literals in these props are rejected at compile time. Literal strings passed as props can slip past the translation linter and reach users untranslated, as happened in #3688 and #3689.


web/src/components/form/ArrayField.tsx
```diff
@@ -283,20 +301,48 @@ type ArrayFieldProps = {
...
-   * Can be a plain string or a ReactNode (e.g. `LabelText` with a suffix).
-   * When a ReactNode is passed, also provide `inputAriaLabel` so assistive
-   * technologies receive a plain-text version of the label.
+   * Either a translated string (wrapped with `_()`) or a ReactNode (e.g.
+   * `LabelText` with a suffix). Plain untranslated strings are rejected at the
+   * type level. In both cases the input's accessible name comes from the
+   * visible label via the FormGroup association.
    */
-  label: React.ReactNode;
+  label: TranslatedString | Exclude<React.ReactNode, string>;
```
lslezak, line 309: Thanks for using the `TranslatedString` type!  👍

web/src/components/form/SearchableSelectField.test.tsx
```diff
@@ -75,7 +76,7 @@ function TestForm({
       <form.AppField name="language">
         {(field) => (
           <field.SearchableSelectField
-            label="Language"
+            label={_("Language")}
             placeholder="Filter by language or territory"
             emptyPlaceholder="Choose an option"
             noResultsText="No matches"
```
lslezak: The placeholder and the other texts should be marked for translation as well I guess,,,

web/src/components/form/SearchableSelectField.test.tsx
```diff
@@ -109,7 +110,7 @@ function RequiredForm() {
       <form.AppField name="language">
         {(field) => (
           <field.SearchableSelectField
-            label="Language"
+            label={_("Language")}
             placeholder="Filter by language or territory"
             noResultsText="No matches"
```
lslezak: The same here, missing translations.

web/src/components/users/authentication-form/Form.test.tsx
```diff
@@ -163,6 +163,25 @@ describe("AuthenticationForm", () => {
...
+      mockRootUser = {
+        password: "secret123",
+        hashedPassword: false,
+        sshPublicKeys: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5 user@host",
+      };
+      const { user } = installerRender(<AuthenticationForm />);
+      await enableFirstUser(user);
+
+      screen.getByLabelText("Administrator account Password");
+      screen.getByLabelText("Administrator account Password confirmation");
```
lslezak, line 177: I'm thinking whether there should be some separator between the prefix and the label. Something like
`Administrator account - Password confirmation`.

I'm not familiar with the assistive technologies, but keep in mind that the translators translate those parts separately without any context. I'm a bit afraid that putting those separately translated parts together might not sound well in some cases in some languages. Adding a pause for the screen reader would be nice. I'm just not sure if the dash (`-`) is the right solution, maybe the assistive technologies expect something else, I do not know.

web/src/hooks/use-field-label.ts
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * @example
+ * // Full control by referencing other on-screen elements. Replaces the visible
+ * // label entirely, so use only when those elements fully describe the field.
+ * <field.NumberField label={_("Size")} aria-labelledby="partition-name size-unit" />
+ */
+export function useFieldLabel(fieldName: string, options: FieldLabelOptions = {}) {
+  const { "aria-label": ariaLabel, "aria-labelledby": ariaLabelledBy, labelPrefixedBy } = options;
+  const labelId = `${fieldName}-label`;
+  const prefixedLabelledBy = labelPrefixedBy ? `${labelPrefixedBy} ${labelId}` : undefined;
```
lslezak, line 113: Just a note: Do not forget about the RTL languages, although we do not use them (at least yet), I'm not sure if it would work correctly here.

Nothing we should solve now, just keep that in mind...

lslezak (commented):
Besides some missing translations it looks well.

web/src/components/users/authentication-form/Form.test.tsx
dgdavid: But we can't put a separator there in a clean way. Improvements to that regard has to wait till 16.2. By now, this is the best thing I  can do having in mind translation freeze. In fact, this was already shipped in previous PR.

web/src/hooks/use-field-label.ts
dgdavid: Sure, something to think about in 16.2 since I had no time in this version to test RTL more deeply. I will take notes.

web/src/components/form/SearchableSelectField.test.tsx
dgdavid: Sure, it is a test that reflect the third-party problem we discussed about detecting missing translations, but let's keep it updated (also I will have a look to the components usage)

lslezak (approved):
LGTM, thanks!

PR #3686, opened 2026-07-01
Move binding information to Settings and remove the dedicated edit link

teclator:
## Problem
 
Recently, the option to configure connection device binding was integrated directly into the main connection form, replacing the separate form that previously handled it. While the separate form and its link in the connections table were successfully removed, the dedicated "Binding" card section and the "Edit binding settings" link were mistakenly left in the connection details view, resulting in a dead link.

- https://trello.com/c/J7KxozJ8/5864-1268594-dead-link-in-network-connection-details
  
## Solution
 
This PR cleans up the connection details view and updates the layout:
  - **Relocates the read-only binding information** (e.g., "Connection is available to all devices") into the general **Settings** card.
  - **Removes the dedicated "Binding" section** and the dead "Edit binding settings" link.

## Screenshots

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/web/src/components/network/ConnectionDetails.tsx
+++ b/web/src/components/network/ConnectionDetails.tsx
@@ -22,7 +22,6 @@
 
 import React, { useState } from "react";
 import {
-  Content,
   DescriptionList,
   DescriptionListDescription,
   DescriptionListGroup,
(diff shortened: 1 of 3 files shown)


web/src/components/network/ConnectionDetails.tsx
```diff
@@ -453,17 +436,10 @@ export default function ConnectionDetails({ connection }: { connection: Connecti
...
-            <BridgeDetails connection={connection} />
-          ) : connection.vlan ? (
-            <VlanDetails connection={connection} />
-          ) : (
-            <BindingSettings connection={connection} />
-          )}
+          {connection.wireless && <NetworkDetails connection={connection} />}
+          {connection.bond && <BondDetails connection={connection} />}
+          {connection.bridge && <BridgeDetails connection={connection} />}
+          {connection.vlan && <VlanDetails connection={connection} />}
```
dgdavid, line 442: Wow! Great catch and nice improvment. Thanks!

teclator: Well, it differs in the way that before we only showed it if nothing of the first where shown ;) but showing the bind in the connection details does not hurt

dgdavid: Whatever, much clean code :)

dgdavid: And sure, having the Binding there is almost expected, so an improvement at the end.

dgdavid (approved):
LGTM

PR #3688, opened 2026-07-02
Missing translations

joseivanlopez:
## Problem

Some strings are not translated. Missing translatated strings should be detected by the linter, but there are two edge cases:

* Literal string directly passed as prop to a PF component.
* Literal strings marked for translation with `N_` but then it is not translated with `_()`.

Examples:
~~~
<PFComponent foo="bar" />

const LABEL = N_("Disk");

<Text>{LABEL}</Text>
~~~

https://bugzilla.suse.com/show_bug.cgi?id=1269514

## Solution

Properly mark the literal strings for translations when they are used in a prop and fix missing translations in storage UI components by converting translation-marked constants to type-safe translation functions.

Changes

* Replaced static translation-marked constants (FILESYSTEM_NAMES, PARTITIONABLE_SPACE_POLICIES, VOLUME_GROUP_SPACE_POLICIES) with dedicated translation functions (filesystemLabel, partitionableSpacePolicyLabel, volumeGroupSpacePolicyLabel).
* Fixed untranslated space policy labels in SpacePolicyMenu and SpaceActionsTable.
* Removed unused SIZE_UNITS constant.

Moving translation logic into functions ensures translations are always applied and prevents future instances of forgotten translations.

In network, the connection type labels were not marked for translation. Fortunatelly, there are already translations for that strings.

### DIFF
--- a/web/src/components/storage/SpaceActionsTable.tsx
+++ b/web/src/components/storage/SpaceActionsTable.tsx
@@ -149,20 +149,20 @@ const DeviceActionSelector = ({
       <FlexItem>
         <ToggleGroup isCompact>
           <ToggleGroupItem
-            text="Do not modify"
+            text={_("Do not modify")}
             buttonId="not-modify"
             isSelected={adjustedAction === "keep"}
             onChange={() => changeAction("keep")}
           />
           <ToggleGroupItem
-            text="Allow shrink"
+            text={_("Allow shrinking")}
             buttonId="resize"
             isDisabled={isResizeDisabled}
             isSelected={adjustedAction === "resizeIfNeeded"}
             onChange={() => changeAction("resizeIfNeeded")}
           />
           <ToggleGroupItem
-            text="Delete"
+            text={_("Delete")}
             buttonId="delete"
             isDisabled={isDeleteDisabled}
             isSelected={adjustedAction === "delete"}
--- a/web/src/utils/network.ts
+++ b/web/src/utils/network.ts
@@ -36,6 +36,7 @@ import {
   SecurityProtocols,
 } from "~/types/network";
 import { _, N_ } from "~/i18n";
+import type { TranslatedString } from "~/i18n";
 
 /**
  * Connection type constants.
@@ -66,8 +67,9 @@ const CONNECTION_STATE_LABELS: Record<ConnectionState, string> = {
 /**
  * Returns the translated label for a connection state.
  */
-// eslint-disable-next-line agama-i18n/string-literals
-const connectionStateLabel = (state: ConnectionState): string => _(CONNECTION_STATE_LABELS[state]);
+const connectionStateLabel = (state: ConnectionState): TranslatedString =>
+  // eslint-disable-next-line agama-i18n/string-literals
+  _(CONNECTION_STATE_LABELS[state]);
 
 /**
  * Returns true if the given connection type is virtual.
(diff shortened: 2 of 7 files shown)


web/src/components/storage/SpaceActionsTable.tsx
```diff
@@ -149,20 +149,20 @@ const DeviceActionSelector = ({
...
           <ToggleGroupItem
-            text="Do not modify"
+            text={_("Do not modify")}
             buttonId="not-modify"
             isSelected={adjustedAction === "keep"}
             onChange={() => changeAction("keep")}
           />
           <ToggleGroupItem
-            text="Allow shrink"
+            text={_("Allow shrinking")}
```
dgdavid, line 158: If I get it correctly, we already had "Allow shrinking" marked for translation somewhere. Similar for "Do not modify", right?

web/src/utils/network.ts
```diff
@@ -81,19 +83,21 @@ const isVirtual = (type: ConnectionType): boolean =>
...
  */
 const CONNECTION_TYPE_LABELS: Record<ConnectionType, string> = {
-  [CONNECTION_TYPE.ETHERNET]: "Ethernet",
-  [CONNECTION_TYPE.WIFI]: "Wi-Fi",
-  [CONNECTION_TYPE.LOOPBACK]: "Loopback",
-  [CONNECTION_TYPE.BOND]: "Bond",
-  [CONNECTION_TYPE.BRIDGE]: "Bridge",
-  [CONNECTION_TYPE.VLAN]: "VLAN",
-  [CONNECTION_TYPE.UNKNOWN]: "Unknown",
+  [CONNECTION_TYPE.ETHERNET]: N_("Ethernet"),
```
dgdavid, line 86: Just for double check it, where these already marked for translations somewhere?

web/src/utils/network.ts
```diff
@@ -66,8 +67,9 @@ const CONNECTION_STATE_LABELS: Record<ConnectionState, string> = {
 /**
  * Returns the translated label for a connection state.
  */
-// eslint-disable-next-line agama-i18n/string-literals
-const connectionStateLabel = (state: ConnectionState): string => _(CONNECTION_STATE_LABELS[state]);
+const connectionStateLabel = (state: ConnectionState): TranslatedString =>
+  // eslint-disable-next-line agama-i18n/string-literals
```
dgdavid, line 71: I wonder if @lslezak linter PR at https://github.com/agama-project/eslint-plugin-agama-i18n/pull/9 helps with that.

dgdavid (approved):
Just minor, not blocking comments. LGTM, thanks.

web/src/components/storage/SpaceActionsTable.tsx
dgdavid: Or are you taking the opportunity to improve them due the fact "Allow shrink" was never marked for translation?

joseivanlopez: Sadly, we don't have translations yet for "Allow shrink" and "Do not modify". They never were marked for translation. So, we can take the opportunity for fixing the string.

web/src/utils/network.ts
joseivanlopez: Yes. The same strings are used in other places, so we have translations.

web/src/components/storage/SpaceActionsTable.tsx
dgdavid: Cool, thanks.

PR #3695, opened 2026-07-06
fix(web): enable dropdown menu flip to prevent overflow

dgdavid:
## Problem

In the Storage page, when selecting "Create LVM volume group" from the installation device dropdown and then opening the volume group actions menu, the dropdown would extend beyond the visible screen area. This happened because MenuButton had flip behavior disabled.

Related to https://bugzilla.suse.com/show_bug.cgi?id=1262122 (protected link)

## Solution

Modified `MenuButton` component to enable PatternFly's flip behavior:

- Changed `enableFlip` from `false` to `true`
- Added `flipBehavior: ["bottom-start"]` to prefer bottom positioning but allow automatic repositioning when viewport boundaries are reached

This ensures all dropdown menus in the application built with MenuButton component stay fully visible regardless of their position on screen.

## Screenshots

**Before**: Menu extends beyond visible area

localhost_8080_ (3)

**After**: Menu properly repositioned within viewport

localhost_8080_ (2)

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul  6 08:28:06 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Fix dropdown menus overflowing beyond viewport boundaries
+  (bsc#1262122).
+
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #3696, opened 2026-07-06
refactor(web): add missing `TranslatedString` types in user-facing text props

dgdavid:
## Summary

- Types user-facing text props as `TranslatedString` instead of plain `string` across form, core, and page components, so untranslated literals are rejected at compile time.
- Marks three already-untranslated row-actions labels ("Actions for %s" in the connections, DASD, and zFCP tables) for translation and reported them in https://bugzilla.suse.com/show_bug.cgi?id=1271404
- Fixes a type bug in `SelectableDataTable`: `itemActionsLabel` was declared as `(d: T) => string | string`, which parses as a function type only, even though a plain value is also accepted at runtime.

## Motivation

Plain string literals passed as props slip through the translation linter and reach users in English only (related to https://bugzilla.suse.com/show_bug.cgi?id=1269514, protected link). The `TranslatedString` tagged type exists precisely to prevent that: it turns a runtime i18n bug into a compile-time error, since text must go through `_()` before a component accepts it.

The type was introduced in #2928 seven months ago (Agama v19), but applying it to component props was overlooked: most user-facing text props remained plain `string`. #3685 and #3688 recently started closing that gap; this PR extends the coverage to the remaining user-facing text props.

## What changed

One commit per component, so each change is easy to review and revert:

**Form components and validation**
- `CheckboxField`: `label`, `aria-label`
- `Fieldset`: `legend`
- `validation-helpers.ts`: all message parameters and return types, plus the two forwarding helpers in `connection-form/validations.ts`

**Core and layout components**
- `SimpleDropdown`: `label`
- `ResizeHandle`: `label`
- `Loading`: `aria-label`
- `SelectableDataTable`: `itemActionsLabel` and the `label` handed to a custom `itemActionsComponent`

**Page-specific components**
- `SoftwarePage`: section titles, button labels, empty-state texts
- `RegistrationCertificateQuestion`: description list terms
- `UnsupportedAutoYaST`: section title and description
- `TextWithLinkToL10n` (InstallerL10nOptions): `text`
- `SystemInformationSection`: item labels

**i18n helpers**
- `formatList` now preserves the input type, mirroring the existing `sprintf` typing: formatting translated strings yields `TranslatedString`, formatting technical values (device names, mount paths) stays plain `string`.

## Tests

- [x] `npm run check-types`
- [x] `npm run eslint`
- [x] `npm run tests`

## Related

- https://bugzilla.suse.com/show_bug.cgi?id=1269514 (private link)
- #2928 (introduced the `TranslatedString` type, Agama v19)
- #3685 (SearchableSelectField TranslatedString props)
- #3688 (label functions typed as TranslatedString)

---


web/src/components/network/ConnectionsTable.tsx
```diff
@@ -383,7 +385,9 @@ export default function ConnectionsTable() {
             },
           ].filter(Boolean);
         }}
-        itemActionsLabel={(c: Connection) => `Actions for ${c.id}`}
+        // FIXME: translate after the translation freeze and drop the casting
+        // (related to bsc#1269514)
+        itemActionsLabel={(c: Connection) => `Actions for ${c.id}` as TranslatedString}
```
ancorgs: I think we should already mark the strings for translation (in a more translators friendly way that with `${c.id}`), instead of adding a FIXME.

I would open a bug report explaining which texts were not marked for translation and their (hopefully low) relevance. Then I would submit the real fix to master.

Very likely we will get no translations for the strings in time for 16.1, but everything will be properly recorded and fixed in the code.

dgdavid: > in a more translators friendly way that with `${c.id}`

Sure, this is a quite old leftover. I would use sprintf or Interpolate component.

dgdavid: > I would open a bug report explaining which texts were not marked for translation and their (hopefully low) relevance. Then I would submit the real fix to master.

Fine with me if you all agree.

dgdavid: Done at https://github.com/agama-project/agama/pull/3696/commits/944ea9db8726310f27af77b21664ec162615ba25

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jul 14 12:38:47 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Require translated text in user-facing component props, ensuring
+  users get the interface in their language
+  (gh#agama-project/agama#3696).
+
```
ancorgs, line 7: Please, mention [bsc#1271404](https://bugzilla.suse.com/show_bug.cgi?id=1271404) in the changelog

dgdavid: Thanks, I overlooked that. Updated now.

ancorgs (approved):
LGTM

PR #3697, opened 2026-07-07
refactor(web): rename overview to installation terminology

dgdavid:
## Motivation

"Overview" was developer jargon from an earlier version of Agama when the main menu item was called "Overview". The current UI shows "Installation" in the breadcrumb and presents a "summary" of installation settings, so the internal code should align with this user-facing terminology.

This refactoring improves code clarity by making internal names match what users actually see in the interface.

### Directory and Component Renaming
- Moved all files from `src/components/overview/` to `src/components/installation/`
- Renamed main component: `OverviewPage` -> `InstallationPage`
- Updated internal component names: `OverviewPageContent` -> `InstallationPageContent`

### Path Constants (routes/paths.ts)
- Renamed `ROOT.overview` to `ROOT.summary` for better semantic clarity
  - The route path remains `/installation`
  - The constant name `summary` better reflects what users conceptually see
  - More intuitive when reading code like `navigate(ROOT.summary)`

### Backwards Compatibility

- Added redirect in router: `/overview` -> `/installation`
- Prevents 404s for bookmarked URLs or external links
- Uses `replace` to maintain expected back-button behavior

## Tests

- [x] TypeScript compilation passes (`npm run check-types`)
- [x] Authentication form tests pass
- [x] All renamed components render correctly
- [x] Breadcrumb navigation works
- [x] Form submit "Go to installation summary" link works
- [x] Legacy `/overview` redirect works


ancorgs (commented):
I like the part about backwards compatibility. But I'm not so sure about the rest.

On the one hand, I'm not sure whether we really need to rename everything to match the UI. Who knows whether we change the string "Installation" in the future. We also have other examples like "Authentication" in the UI vs `users` in the code. Are we going to change them all and keep them consistent with the UI strings over time? Looks overkill to me.

On the other hand, I don't see why we use `summary` for the route and `installation` for the components. I personally do not like `installation`(*) and I prefer the current `overview` or `summary`. But none of those two are actually used in the UI, which somehow defeats the purpose of this whole change.

So IMHO, the whole PR fixes something that I don't think is broken. :-)

(*) IMHO, "installation" doesn't look like a good name for one particular namespace of an installer.

joseivanlopez:
> I like the part about backwards compatibility. But I'm not so sure about the rest.

I like the idea of using */installation* URL instead of */overview*. IMHO, having URLs aligned with the terms used in the UI is good.

I agree with @ancorgs about renaming internal components. As he commented, *installation* does not seem a good namespace. We already have *xSummary* components, so probably *summary* is a better "namespace".

dgdavid:
So?

What should we do? Close PR and keep everything as it is, or turn it into minimal change for the route by picking _summary_ instead of _installation_?

joseivanlopez:
> So?
> 
> What should we do? Close PR and keep everything as it is, or turn it into minimal change for the route by picking _summary_ instead of _installation_?

IMHO, I would keep the URL change and `ROOT.summary`.

dgdavid:
After above reviews and a later conversation with @ancorgs, tclosing this in favor of #3722.

PR #3698, opened 2026-07-07
Introduced `MarkedString` type

lslezak:
## Problem

- Using `N_()` without `_()` later in a React component results in an untranslated text ([bsc#1269514](https://bugzilla.suse.com/show_bug.cgi?id=1269514))
## Solution

- Activate the new `agama-i18n/marked-string` rule implemented in https://github.com/agama-project/eslint-plugin-agama-i18n/pull/9

## Additional changes

- The improved `agama-i18n/string-literals` check can now detect the used type so it does not complain when an expression which returns `MarkedString` is used. That means we can remove the `eslint-disable-next-line agama-i18n/string-literals` comments.

## Notes

- While updating the `@agama-project/eslint-plugin-agama-i18n` npm updated also some eslint packages which require some changes in the indentation at several places.

### DIFF
--- a/web/eslint.config.mjs
+++ b/web/eslint.config.mjs
@@ -15,6 +15,8 @@ export default [
   ...neostandardConfig,
   js.configs.recommended,
   ...tsEslint.configs.recommended,
+  // TODO: enable TypeScript checks as well?
+  // ...tsEslint.configs.recommendedTypeChecked,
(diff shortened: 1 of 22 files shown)


dgdavid (approved):
Great i18n improvement that for sure will help the project to miss less translations. Thanks! 👍

Issue #3701, opened 2026-07-08
network: vlan configuration with incorrect flags


jeffmahoney:
I ran into an issue with a DHCP server failing to see any incoming traffic after an [wicked2nm](https://github.com/openSUSE/wicked2nm) migration (that uses the Agama network create). SSH and ICMP worked fine. I tracked it down to the `VLAN_FLAG_REORDER_HDR` (`1`) flag not being set on the VLAN interface. After setting it, the DHCP server worked correctly. Without this flag, VLAN interfaces will work fine for Unicast and Multicast traffic but will fail for Broadcast traffic. 

Why is this an issue for Agama? I had used wicked2nm to migrate a SLES 15 SP7 system to SLES 16 and it migrated a working Wicked configuration to one that didn't. wicked2nm uses the agama-network crate, which has a problem in how it transfers the vlan configuration to dbus/NetworkManager. It doesn't specify the flags field. Ideally this should be unspecified and it wouldn't be written in the final output. However, the NetworkManager developers have made a conscious choice to handle a missing `flags` field as the *old* default of `0` which is written as a `flags=0` value in the nmconnection file. I expect that using Agama directly to configure a VLAN interface will result in a similar failure.

teclator:
The related PR should be already merged, therefore, closing it.

PR #3702, opened 2026-07-08
agama-network: fix handling of flags for VLAN interfaces (#3701)

jeffmahoney:
vlan_config_to_dbus leaves the flags field unspecified which _should_ be fine, but is unfortunately not.  The NetworkManager developers have made a conscious decision to treat an absent flags field as the old default value of 0, and that value is written into the nmconnection file.

The problem is that the kernel defaults to a value to 1/VLAN_FLAG_REORDER_HDR.

Without this flag, software using AF_PACKET doesn't work, including tcpdump filters and dhcpd.

This fix adds the flags field with the kernel default value of 1.


teclator:
@jeffmahoney thanks for the contribution. I'm wondering if we should set it to 1 unconditionally or we should read and set it depending on the current value instead. Otherwise we could override the value isn't it? @cfconrad, what do you think?

teclator:
This PR has been replaced by #3705, thanks a lot!

PR #3703, opened 2026-07-09
Do not timeout when starting agama-web-server

teclator:
## Problem

Starting **agama-web-server** service could timeout because the storage service is taking too long for starting.

## Solution

Set the service **StarTimeoutStartSec** to infinity in order to not timeout when starting it

## Testing

- *Tested manually*


lslezak (approved):
LGTM

PR #3705, opened 2026-07-09
agama-network: fix handling of flags for VLAN interfaces (based on #3702 for fixing #3701)

teclator:
## Problem

VLAN flags are not handled correctly as reported by (#3701).

- Issue https://github.com/agama-project/agama/pull/3701
- Proposed solution (https://github.com/agama-project/agama/pull/3702)
- https://bugzilla.suse.com/show_bug.cgi?id=1271222

## Solution

Based on @jeffmahoney proposed solution, we have extended it in order to handle properly the VLAN flags in case them are modified not setting it always as 1 but using the current value instead.

## Testing

- *Added a new unit test*


rust/agama-network/src/nm/dbus.rs
```diff
@@ -1907,6 +1920,87 @@ mod test {
...
+        if let ConnectionConfig::Vlan(config) = connection.config {
+            assert_eq!(config.id, 100_u32);
+            assert_eq!(config.parent, "eth0");
+            assert_eq!(config.protocol, VlanProtocol::IEEE802_1Q);
+            assert_eq!(config.flags, Some(5));
+        } else {
+            panic!("Wrong connection type");
+        }
+
+        // Scenario 2: DBus does not have flags
```
imobachgs:
```suggestion
        // Scenario 2: D-Bus does not have flags
```

rust/agama-network/src/nm/dbus.rs
```diff
@@ -399,6 +399,12 @@ pub fn cleanup_dbus_connection(conn: &mut NestedHash) {
             ipv6.remove("gateway");
         }
     }
+
+    if let Some(vlan) = conn.get_mut("vlan") {
+        if !vlan.contains_key("flags") {
+            vlan.insert("flags", 1u32.into());
```
imobachgs: Why not `1_u32`? Just for consistency.

imobachgs (commented):
It looks good but, please, check the Clippy output.

rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jul  9 00:00:00 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Handle VLAN flags properly, defaulting to kernel default value
```
```suggestion
- Handle VLAN flags properly, using the kernel default value
```

rust/agama-network/src/nm/dbus.rs
teclator: Becase was originally implemented so by @jeffmahoney but for consistency the change looks good to me.

rust/agama-network/src/model.rs
```diff
@@ -1366,6 +1366,8 @@ pub struct VlanConfig {
     pub parent: String,
     pub id: u32,
     pub protocol: VlanProtocol,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub flags: Option<u32>,
```
cfconrad: Maybe we would like to give meaning full names like we have in nm
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/a6e971fb24e9acaba26186b318d1d3c56bfc0efe/src/libnm-core-public/nm-setting-vlan.h#L67

cfconrad (approved):
Good catch, thx.

only for reference, this is the doc from nm which state the inconsistency:
```
The default value of this property is NM_VLAN_FLAG_REORDER_HEADERS, but it used to be 0. 
To preserve backward compatibility, the default-value in the D-Bus API continues to be 0 
and a missing property on D-Bus is still considered as 0.
```
https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html

Regarding the flag naming, I'm fine with a number but maybe text looks better in other agama api's, up to you.

rust/agama-network/src/model.rs
teclator: Ok, added suggestion https://github.com/agama-project/agama/pull/3705/commits/2c6e23f476fe76be96b87c0ce2646b3c85923127

imobachgs (approved):

imobachgs (approved):

cfconrad (approved):
LGTM, thx!

teclator:
and thanks to @jeffmahoney for reporting and also for creating a PR for fixing it!

PR #3706, opened 2026-07-09
Provide early feedback when Agama start takes too long time

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1270014
- No visual feedback when initializing the Agama server takes too long time
- There is just an empty grey screen
agama-init

## Solution

- Generate a separate HTML page which just shows a loading indicator (currently just a simple `<Spinner>` component but I guess @dgdavid can improve this :smiley: )
- Improve the  GNOME kiosk script so if Agama does not start in the limit the script starts the Firefox browser displaying the loading page.
- Use the `file://` protocol for the loading page so it does not need a running web server (Agama is not ready yet, starting yet another web server is an overkill).
- When Agama finally accepts connections the browser with the loading page is restarted with the standard Agama page.

[agama-early-feedback.webm](https://github.com/user-attachments/assets/b1064417-71cd-43dc-aee9-743c38fcc1e8)

## Testing

- Tested manually
- The testing ISO image is available at https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/early-feedback/images/iso/

## Notes

- This is just a proof of concept and needs some improvements.
- This is rather a quick workaround, we should find a proper solution for the long storage initialization in 16.2.

## TODO

- [ ] Remove the testing `sleep` before merging this change!

### DIFF
--- a/live/live-root/root/.local/bin/gnome-kiosk-script
+++ b/live/live-root/root/.local/bin/gnome-kiosk-script
@@ -10,7 +10,7 @@ PREFS=$HOME/.mozilla/firefox/profile/user.js
 LANG_FILE=/sys/firmware/efi/efivars/PlatformLang-8be4df61-93ca-11d2-aa0d-00e098032b8c
 
 # file exists and is not empty
-if [ -s  "$LANG_FILE" ]; then
+if [ -s "$LANG_FILE" ]; then
   # skip the first 4 bytes (the EFI attributes), keep only the characters allowed in a language code,
   # especially remove the trailing null byte
   EFI_LANG=$(cat "$LANG_FILE" | tail -c +5 | tr -cd "[_a-zA-Z-]")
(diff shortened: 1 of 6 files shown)


dgdavid:
> * Generate a separate HTML page which just shows a loading indicator (currently just a simple `<Spinner>` component but I guess @dgdavid can improve this 😃 )

That separated HTML page served without webserver is a good idea. Also, opting for the spinner is the only thing we can provide at this point because of string freeze, but something to change as soon as possible for something more explicitly like _Starting the installer. This may take a moment..._

Reasoning, this is not the typical, usually quick, loading use case. The server is indeed not up, then displaying the right feedback, when possible, is more than desired.

Thanks a lot for adding this to avoid users get impatient.

imobachgs:
> > * Generate a separate HTML page which just shows a loading indicator (currently just a simple `<Spinner>` component but I guess @dgdavid can improve this 😃 )
> 
> That separated HTML page served without webserver is a good idea. Also, opting for the spinner is the only thing we can provide at this point because of string freeze, but something to change as soon as possible for something more explicitly like _Starting the installer. This may take a moment..._
> 
> Reasoning, this is not the typical, usually quick, loading use case. The server is indeed not up, then displaying the right feedback, when possible, is more than desired.
> 
> Thanks a lot for adding this to avoid users get impatient.

I am not sure about the overall approach. I would rather prefer the web server to start listening from the very beginning, so we do not need this page (we are working on that). However, I like this approach to display an error when the web server is not even able to start.

imobachgs:
#3720 introduced some changes to start listening as soon as possible, so this change is not needed anymore. However, we might consider something similar when `agama-web-server` crashes during initialization.

lslezak:
For handling a service crash the systemd service can use the [ExecStopPost](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExecStopPost=) option in the unit file.

lslezak:
Closing this, #3720 solves the problem in a better way.

Issue #561, opened 2023-05-08
Allow stopping the installation


joseivanlopez:
There is no way to stop the installation once it starts. Sometimes it does not make sense to continue with the installation, for example if there is an important storage commit error. Now, that kind of errors are reported to the user (see https://github.com/openSUSE/agama/pull/558), but the overall insallation process always continues.

Some open questions:

* How to cancel the installation (from D-Bus point of view)?
* How should the UI to offer the option for canceling the installation?
* How should the UI behave after canceling the installation?

mvidner:
From the viewpoint of the lower layers:

The **D-Bus specification/protocol** does not have an API for ending a service, it assumes a connection simply drops off the bus by its process ending or by closing its socket. Some services implement a timeout, other implement an explicit `Quit` method.
If we do this we may create a problem where the UI client keeps talking to us, activating another instance with a reset state and not realizing it. If the client wants to detect that, it should use `org.freedesktop.DBus.GetNameOwner`, save the service unique connection ID (`:1.42`) and compare it later.

**ruby-dbus** so far does not care how to disconnect (and perhaps reconnect), it assumes that a connection lives until your process ends (https://github.com/mvidner/ruby-dbus/issues/43). I think it is fine for Agama, just mentioning it for completeness.

dgdavid:
@joseivanlopez, @mvidner 

This looks a bit out-dated to me after latest API changes, but maybe it is not. The thing is that we have [another open issue about this topic](https://github.com/agama-project/agama/issues/3321), so maybe revisiting it is not a bad idea.

imobachgs:
Let's close this one and keep #3321 open (reported from outside the team).

PR #3709, opened 2026-07-10
Add cancel button to installation progress screen

mvarlese:
Fixes https://github.com/agama-project/agama/issues/3321.

Users can now abort an ongoing installation via a cancel button displayed during the installation progress. The button triggers a confirmation dialog before canceling to prevent accidental aborts.

Backend changes
Add CancelInstall action to API
Handler resets stage to Configuring and clears progress
Frontend changes
Add cancelInstallation() API function
Add cancel button with confirmation dialog to InstallationProgress
Dialog warns user that all ongoing operations will stop


dgdavid (commented):
Hi @mvarlese  👋 

Thanks a lot for the PR 🙏 It's great to see people contributing to the project 🎉 

Setting the backend aside for a moment, I'd like to share my thoughts on this "Cancel" feature from the UI point of view.

At first glance, this looks like a simple, low-risk feature that's ready to ship, especially after seeing your implementation. But it's actually more complicated than it seems, and that's one of the reasons why I think we haven't implemented it yet. We actually have already discussed some of the challenges in #561, especially [lslezack's comment](https://github.com/agama-project/agama/issues/561#issuecomment-4126356744).

Even though you've intentionally limited this PR to the installation progress screen, there are still a few important questions we should answer first:

*   **When should "Cancel installation" be available?** Should it always be available? Only during the configuration phase? Or only once the installation has actually started?
*   **When available during the configuration phase**, before anything is installed, what does the action actually do? Reboot? Shut down? In that case, "Cancel" isn't really the right term because there's nothing in progress to cancel.
* **When available during the installation phase**, as in this PR, things become more complicated. "Cancel installation" is probably a good label, although the confirmation dialog could be rewritten to be less repetitive and more informative. It might also need to account for the fact that the installation keeps running while the dialog is open, so by the time the user confirms, the system may already be in a different state than when the dialog first appeared. More importantly, what do we tell users about the state of their system after they cancel? As far as I know, Agama can't currently show a meaningful "Installation canceled" summary explaining what was completed and what wasn't. Partitioning may have finished, packages may be only partially installed, and the bootloader may or may not be configured. Without that information, users are left with a machine in an unknown state. It's also important to make it clear that **canceling does not restore the previous system**. Anything that has already been done, whether partitioning, package installation, or bootloader configuration, cannot be rolled back.

*   **What should happen immediately after canceling?** Powering off might be the safest option, but the user may simply want to reboot and try again. Either way, we could be handing them a system that won't boot. At the time of cancellation, we don't have enough information to make that decision for them. One option would be to let users choose in the confirmation dialog, for example "Cancel and reboot", "Cancel and power off", or "Continue installation". However, that also increases the cognitive load for what should ideally be a simple interaction.

None of this is a criticism of the PR itself. It's solid work, and having a working implementation is genuinely helpful because it gives us something concrete to discuss. But I'd rather answer these questions before merging, so we don't ship a "Cancel" button that looks useful but ends up creating new problems for users.

Even setting all of that aside, though, I'm also not convinced this is worth adding at all. As Ladislav pointed out, this feature addresses a _quite rare scenario_. I don't think most users will ever cancel an installation halfway through. You could argue it's mainly useful for advanced users, for example when working with virtual machines where there's no hardware power button. But those users already know how to reboot or power off the machine if needed. They'll also soon have access to a terminal, where they'll be able to do the same without any confirmation dialog. I think that's fine, because once someone chooses to operate from the terminal, it's reasonable to assume they understand the consequences of what they're doing.

There's also a more general UI consideration. This is an installer, not a general-purpose application. We should absolutely provide a polished experience and make common workflows as smooth as possible. At the same time, I think we should be careful about adding actions that somehow could _encourage_ uncommon or potentially unsafe workflows, no matter how easy they are to implement. To me, a "Cancel installation" button feels similar to adding a "Log out" action to an installer. It's technically possible, but it doesn't really fit the primary purpose of the application. In this case, I think keeping it simple is better, even though it could mean not having such an action at all.

All above said, if we want to give users a safe way out of the installer, I'd argue the safer place to do it is *before* installation actually starts, not during. At that point nothing destructive has happened yet, so there's no ambiguity about system state to worry about. We could offer something like "Exit" with a confirmation dialog letting the user choose between "Reboot" and "Power off." That sidesteps every question raised above: there's nothing to partially roll back, no undefined machine state to explain, and no need to guess whether the user wants to retry or walk away.

Anyway, if we decide to move forward with the feature, I have a couple more comments:

*   I would target this for 16.2, not only because it needs more discussion, but also because we're currently in string/UI freeze.
*   I also wouldn't use the proposed call to action for this. Instead, I'd place this action in the overflow menu in the top-right corner, alongside "Show configuration" and "Download logs".
  
  | Current proposal | Suggested proposal |
  |-|-|
  | Image | Image |

Of course, these are just my current thoughts on the topic. I may well have overlooked something, either in favor of or against the feature. I'd be interested to hear what the rest of the team thinks before we draw any conclusions.

Thanks again for the work you put into this. It's a great starting point for finally settling this discussion and potentially delivering something polished and meaningful for users.

mvarlese:
Hello @dgdavid ,
Thank you for taking the time for reviewing this PR and share with a lot of extremely good level of detail all your thoughts and considerations. Frankly speaking, I simply took an issue (open on the project itself) to potentially be of help. I had no idea there were still thoughts around whether to implement such a feature at all. Your view is certainly valid and I would never take your feedback as criticism, quite the opposite, I learnt quite a bit by just reading your comment above.
I am happy if the team can use (if at all) this code at a later stage or discard it (if the decision is not to implement the feature at all). For me it was mostly a learning exercise (with the hope to provide some help :)).
Once again, thank you for taking the time for a very thorough explanation around the topic!

joseivanlopez:
Juse in case it helps, I have asked Gemini whether common installers usually offer a cancel option, see https://gist.github.com/joseivanlopez/38d0b7881a01ab42ae5b8d20d735bb0a.

IMHO, having a way to cancel in the phase 1 (see gist) is very useful in several scenarios. And for the phase 2, Agama should offer at least a way to abort the installation when something goes wrong (e.g. formatting a partition fails, a package is not found, etc).

dgdavid:
> Juse in case it helps, I have asked Gemini whether common installers usually offer a cancel option, see https://gist.github.com/joseivanlopez/38d0b7881a01ab42ae5b8d20d735bb0a.
> 
> IMHO, having a way to cancel in the phase 1 (see gist) is very useful in several scenarios. And for the phase 2, Agama should offer at least a way to abort the installation when something goes wrong (e.g. formatting a partition fails, a package is not found, etc).

I see your point, which in the end, it seems more aligned with providing an "Exit" option during the configuration phase rather than a "Cancel" action during the installation phase.

A couple of other thoughts came to mind while reading your comment:

* Unless I'm wrong and things has changed a lot lately, most of these installers do not behave like Agama when it comes to partitioning. They apply partitioning changes as part of the workflow itself, rather than deferring them until the actual installation phase. This could help them to better manage the cancel action or related feedback after such an step, don't know.
* What you describe as _offer at least a way to abort the installation when something goes wrong_ is not really a cancel installation action. If the installation has already reached a point where it cannot proceed, there is nothing left to cancel. Instead, looks more as a way to give feedback to the user that the installation cannot continue and offering a _graceful_ way to leave the installer, most likely by presenting the above mentioned "Reboot" and "Power off" options as part of such a feedback. A good improvement, in any case.

joseivanlopez:
> * Unless I'm wrong and things has changed a lot lately, most of these installers do not behave like Agama when it comes to partitioning. They apply partitioning changes as part of the workflow itself, rather than deferring them until the actual installation phase. This could help them to better manage the cancel action or related feedback after such an step, don't know.

AFAIR, the last time I installed a macOS, the partitioning was done once the installation started (like Agama). And the same for other Linux installers like Anaconda or Calamares. 

> * What you describe as _offer at least a way to abort the installation when something goes wrong_ is not really a cancel installation action. If the installation has already reached a point where it cannot proceed, there is nothing left to cancel. Instead, looks more as a way to give feedback to the user that the installation cannot continue and offering a _graceful_ way to leave the installer, most likely by presenting the above mentioned "Reboot" and "Power off" options as part of such a feedback. A good improvement, in any case.

I was just mentinoning it because it is a kind of "cancel" (or abort) which is missing right now.

If time permits, I will collect a list of possible use cases for cancelling in both "phase 1" and "phase 2".

dgdavid:
> > * Unless I'm wrong and things has changed a lot lately, most of these installers do not behave like Agama when it [...]
> 
> AFAIR, the last time I installed a macOS, the partitioning was done once the installation started (like Agama). And the same for other Linux installers like Anaconda or Calamares.

Then, I was wrong or things has changed a lot lately. Or both :wink:

> 
>[...]
> 
> If time permits, I will collect a list of possible use cases for cancelling in both "phase 1" and "phase 2".

Cool, that might help.

In any case, let me be clear: I'm not against implementing these escape hatches if we can find a solution that is consistent, intuitive, pleasant, and, most importantly, non-disruptive.I'm just sharing the considerations I had in the past and the ones that came back to mind this morning while reading @mvarlese's PR. The only thing I don't support is making it a primary or call-to-action button. 

Also, I still think this is more of a nice-to-have than a high-priority feature. That doesn't mean we should never implement it or invest time in it, but we need to find the right slot among our priorities to work on it.

In my opinion, the user experience around this kind of interaction deserves more attention and care than an early release (unlike some other features, where I'm more willing to prioritize shipping sooner). That doesn't mean the result will be perfect or free of oversights, but it will at least be a more deliberate and thoughtfully designed solution.

In short, I'm not opposed to the feature _per se_. I'm simply asking that we think it through more carefully, including whether it actually makes sense during the installation phase and, if it does, how we can provide the right user interaction and feedback towards the confirmation dialog, as we did some time ago with the installation confirmation dialog.

dgdavid:
Now that this has been tracked in #3911, let’s close this PR and restart the work in a new one when this is picked up for development. Hope you don’t mind.

Thanks a ton for the effort. It was really useful in laying the groundwork and getting us started.

PR #3710, opened 2026-07-10
Fix "unexpected error" drive is undefined for VROC RAID

mvidner:
## Problem

"Unexpected error" during SLES 16.1 installation when manually selecting the secondary VROC RAID 1 disk

- https://bugzilla.suse.com/show_bug.cgi?id=1271120

## Solution

Allow MD devices that are partitionable disk devices to export the "drive" section.

## Testing

- Added a new unit test

## Screenshots

No

## Documentation

_TODO: .changes_

- Run: `git ls-files '*.md'`

[cli.md]: https://github.com/agama-project/agama-project.github.io/blob/main/docs/user/reference/cli.md

### DIFF
--- a/service/lib/agama/storage/devicegraph_conversions/to_json_conversions/drive.rb
+++ b/service/lib/agama/storage/devicegraph_conversions/to_json_conversions/drive.rb
@@ -38,7 +38,7 @@ module Agama
           # @param storage_device [Y2Storage::Device]
           # @return [Boolean]
           def self.apply?(storage_device)
-            storage_device.is?(:disk, :dm_raid, :multipath, :dasd) &&
+            (storage_device.is?(:disk, :dm_raid, :multipath, :dasd) || storage_device.is?(:md)) &&
               storage_device.is?(:disk_device)
           end
 
--- a/service/test/agama/storage/devicegraph_conversions/to_json_test.rb
+++ b/service/test/agama/storage/devicegraph_conversions/to_json_test.rb
@@ -31,6 +31,7 @@ describe Agama::Storage::DevicegraphConversions::ToJSON do
   before do
     mock_storage(devicegraph: scenario)
     allow_any_instance_of(Y2Storage::Partition).to receive(:resize_info).and_return(resize_info)
+    allow_any_instance_of(Y2Storage::LvmLv).to receive(:resize_info).and_return(resize_info)
   end
 
   subject { described_class.new(devicegraph) }
@@ -171,6 +172,18 @@ describe Agama::Storage::DevicegraphConversions::ToJSON do
         json = subject.convert
         wires = json.first[:multipath][:wireNames]
         expect(wires).to contain_exactly("/dev/sda", "/dev/sdb")
+      end
+    end
+
+    describe "for a devicegraph with BIOS RAID (such as VROC)" do
+      let(:scenario) { "lvm-over-hardware-raid.xml" }
+
+      it "exports BIOS RAID volumes with drive class and drive section" do
+        json = subject.convert
+        vols = json.select { |d| ["/dev/md/a", "/dev/md/b"].include?(d[:name]) }
+        expect(vols).to_not be_empty
+        expect(vols.map { |v| v[:class] }).to all(eq "drive")
+        expect(vols.map { |v| v.key?(:drive) }).to all(be true)
       end
     end
   end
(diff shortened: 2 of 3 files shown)


mvidner:
@ancorgs has asked for more logs in the bug but I thought "maybe the robots have enough of logs already" so I used gemini-3.5-flash and asked it:

> Use curl to read https://.../1271120 then look in . at the downloaded logs.                                                   
> Can we see in the logs why the bug happens?
                                                                                                                             
(Then I symlinked the git tree into ./agama/ . Then I advised it that ruby  -ry2storage ... is not enough and needs -ryast)                                                                                                                                                                   

What it found out looks good enough for me so I dare ask my team mates to verify it:
___
### Root Cause Analysis of Bug 1271120

1. System Logs Context
In the logs (journalctl_sles161u1.txt and dmesg.txt), we see that the system has an Intel VROC (IMSM) RAID configuration:
 * A metadata container /dev/md127 (sysfs md127, XML class Y2Storage::MdContainer).
 * Two partitionable RAID 1 volumes inside that container: /dev/md125 (/dev/md/VrocVolume00_0, sysfs md125) and /dev/md124
   (/dev/md/VrocVolume01_0, sysfs md124).

2. Backend Classification Mismatch
In yast2-storage-ng (the storage library Agama uses):
 * Standard software RAIDs (like /dev/md0) are of class Y2Storage::Md. They return is?(:software_raid) = true and is?(:disk_device) = false.
 * VROC BIOS RAID volumes (like /dev/md125) are of class Y2Storage::MdMember. They return is?(:software_raid) = false and is?(:disk_device) =
   true.

Because BIOS RAID volumes are partitionable target disks, the backend storage system (system.rb) includes them under candidate_drives rather than software RAIDs.

When the backend serializes the storage model to JSON (agama/service/lib/agama/storage/devicegraph_conversions/to_json_conversions/device.rb):
 1. Device Class Determination:

```rb
 1    def device_class
 2      return "partition" if storage_device.is?(:partition)
 3      return "logicalVolume" if storage_device.is?(:lvm_lv)
 4      return "volumeGroup" if storage_device.is?(:lvm_vg)
 5      return "mdRaid" if storage_device.is?(:software_raid)
 6
 7      "drive"
 8    end
```

   Because a VROC BIOS RAID volume returns false for is?(:software_raid), its class is set to "drive".

 2. Drive Properties Section Exporter (drive.rb):

```rb
 1    def self.apply?(storage_device)
 2      storage_device.is?(:disk, :dm_raid, :multipath, :dasd) &&
 3        storage_device.is?(:disk_device)
 4    end
```

   Because a BIOS RAID volume (Y2Storage::MdMember) returns false for :disk, :dm_raid, :multipath, and :dasd, Drive.apply? returns false.

   As a result, the backend serializes the VROC volume with "class": "drive", but entirely omits the "drive" section from its JSON structure.

3. Frontend Crash
When the user manually changes the installation target to the second VROC disk, the React frontend renders the DrivesTable.tsx and tries to get each device's description using driveTypeDescription(device) in device.tsx:

```js
 1 const driveTypeDescription = (device: System.Device): string => {
 2   if (device.drive.type === "multipath") { ... }
```

Because device.class is "drive", but the drive property is completely missing (undefined), the browser dereferences device.drive.type and crashes with:
TypeError: can't access property "type", e.drive is undefined

### Propose Strategies

We can fix this robustly on two fronts:

 1. Backend Fix (drive.rb):
   Allow MD devices that are partitionable disk devices to export the "drive" section so that the UI can retrieve their model name and details:

```rb
 1    def self.apply?(storage_device)
 2      (storage_device.is?(:disk, :dm_raid, :multipath, :dasd) || storage_device.is?(:md)) &&
 3        storage_device.is?(:disk_device)
 4    end
```

 2. Frontend Fix (device.tsx):
   Make driveTypeDescription robust by using optional chaining (device.drive?.type) to prevent crashes if drive properties are omitted:

```js
 1    const driveTypeDescription = (device: System.Device): string => {
 2      if (device.drive?.type === "multipath") { ... }
```

mvidner:
I also asked:

> OK. Can we extract a test case from the logs?  

We can extract a perfect test case that replicates this exact scenario.

  1. The Match Between Logs and Test Fixtures
  In the system logs:
   * VROC Containers: /dev/md127 and /dev/md126 (Y2Storage::MdContainer devices).
   * VROC Volumes: /dev/md125 (/dev/md/VrocVolume00_0) and /dev/md124 (/dev/md/VrocVolume01_0) (Y2Storage::MdMember devices).

  In the existing test suite, there is already an XML fixture lvm-over-hardware-raid.xml with the exact same layout:
   * IMSM Container: /dev/md/imsm0 (Y2Storage::MdContainer).
   * IMSM Volumes: /dev/md/b and /dev/md/a (Y2Storage::MdMember devices).

  This means we don't even need to craft a new XML file—we can use lvm-over-hardware-raid.xml to construct the test case.

mvidner:
The real fix was done in #3719

PR #3718, opened 2026-07-13
Fixed in the backend for handling VLAN flags

teclator:
It backports #3705 to SLE-16, the VLAN configuration is not supported in the UI therefore, only the backend is adapted.

- https://bugzilla.suse.com/show_bug.cgi?id=1271079

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jul 13 12:40:10 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Support VLAN flags configuration, defaulting to REORDER_HDR if not
+  specified (gh#agama-project/agama#3718, bsc#1271079).
+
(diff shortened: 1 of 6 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Jul 13 12:40:10 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Support VLAN flags configuration, defaulting to REORDER_HDR if not
```
imobachgs: I would most probably use the same entry than the other PR:

```
- Handle VLAN flags properly, using the kernel default value
  of VLAN_FLAG_REORDER_HDR when not specified
  (gh#agama-project/agama#3701, bsc#1271222).
```

imobachgs (changes requested):
In general, it looks good. Just a problem with the tests (it seems to be a `use` is missing).

imobachgs (approved):

PR #3719, opened 2026-07-13
Fix detection of hardware RAIDs in /api/system

ancorgs:
## Problem

The list of existing storage devices (read from the libstorage-ng devicegraph) is exported as part of `/api/system`. In that representation, the BIOS-based RAIDs (a.k.a. hardware RAIDs) are represented as "drives" since they are fully equivalent to a disk.

But the logic to detect when a drive is a BIOS RAID was wrong. The code actually checked whether it was a DM RAID, which is a legacy kind of RAID not even supported in recent versions of SUSE and/or openSUSE.

As a consequence, the UI was failing is shown here.

hardware-raid-wrong

## Solution

Use the correct method to check if the RAID is based on BIOS/hardware. That proper check returns `true` for the deprecated DM RAIDs but also for any other kind of BIOS RAID.

As a result, the UI works again and the RAID can be selected just as a regular disk.

hardware-raid

## Testing

- Added a new unit test
- Tested manually using a devicegraph from a bug report (as seen in the screenshots).

### DIFF
--- a/service/lib/agama/storage/devicegraph_conversions/to_json_conversions/drive.rb
+++ b/service/lib/agama/storage/devicegraph_conversions/to_json_conversions/drive.rb
@@ -38,7 +38,7 @@ module Agama
           # @param storage_device [Y2Storage::Device]
           # @return [Boolean]
           def self.apply?(storage_device)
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

PR #3720, opened 2026-07-13
Improve initialization process

joseivanlopez:
### Problem

The Agama web UI only appears once the Agama web server is ready. And the web server does not start until the storage D-Bus service finishes its startup. Sometimes, the storage D-Bus service takes long startup time. As consequence, the UI might take long time to appear.

* https://bugzilla.suse.com/show_bug.cgi?id=1265451
* https://bugzilla.suse.com/show_bug.cgi?id=1270014

### Solution

Asynchronously initialize the storage D-Bus service, so the web server is able to respond even though the D-Bus data is not loaded yet. The UI shows a spinner meanwhile the data is not ready.

### DIFF
--- a/rust/agama-storage-client/src/service.rs
+++ b/rust/agama-storage-client/src/service.rs
@@ -63,17 +63,10 @@ impl Starter {
     /// from each proxy to cache the current values.
     pub async fn start(self) -> Result<Handler<Service>, Error> {
         let storage_proxy = proxies::Storage1Proxy::new(&self.dbus).await?;
-        storage_proxy.config().await?;
-
         let bootloader_proxy = proxies::BootloaderProxy::new(&self.dbus).await?;
-        bootloader_proxy.config().await?;
-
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

joseivanlopez:
The current UI start-up process is the following:

* Firefox does not start until the web server is available.
* Once the web server is available, the UI shows up.
* The UI shows a blank page with a centered spinner, nothing else.
* The spinner stays until the web server responds with the requested information.
* The UI shows the product selection.

With the changes implemented in this card, the web server is available earlier than before because it does not wait for storage to start up. As consequence, the UI appears earlier too, and it keeps waiting (with spinner) until the storage information is available.

After discussing this with @dgdavid, we decided to keep the UI as it is for now. The spinner is enough to indicate the installer is loading. Anyway, in the future we can think on providing more contextual information, but we need to take into account some implications. For example, what language to use?

PR #3721, opened 2026-07-14
Improve hardware detection

imobachgs:
## Problems

* If calling to `lshw` fails or it is not possible to parse its output, Agama won't work.
* There are some troubles when trying to find out the CPU in s390 architectures and the amount of RAM.

## Solution

* Do not crash, when calling `lshw` or parsing its output fails. Just return empty hardware information.
* Improve model and processor detection on s390 architecture.
* Improve memory detection when the "memory" key is not present (but there is a "memory:0").

### Logs when starting Agama

```
jul 14 15:13:33 agama agama-web-server[10565]: Failed to read hardware information: Failed to parse lshw output: Error("control character (\\u0000-\\u001F) found while parsing a string", line: 2, column: 0)
```

### Logs during auto-installation

The following error was already there. This PR does not change anything on that regard.

```
Error:
Profile validation failed

Details:
Jsonnet evaluation failed:
RUNTIME ERROR: hw.libsonnet:36:3 Unterminated String
        profile.jsonnet:1:15-36 thunk <agama> from <$>
        profile.jsonnet:2:16-21
        profile.jsonnet:22:14-20        object <anonymous>
        Field "comment"
        Field "product"
        During manifestation
```

### s390 information

image

## Testing

- Added a unit test
- Tested manually

### DIFF
--- a/rust/agama-manager/src/hardware.rs
+++ b/rust/agama-manager/src/hardware.rs
@@ -252,8 +252,9 @@ impl HardwareNode {
         }
 
         for children in &mut self.children {
-            if let Some(result) = children.find_first_by_class_mut(class) {
-                return Some(result);
+            let result = children.find_first_by_class_mut(class);
+            if result.is_some() {
+                return result;
             }
         }
 
@@ -263,6 +264,8 @@ impl HardwareNode {
 
 impl From<&HardwareNode> for HardwareInfo {
     fn from(value: &HardwareNode) -> Self {
+        // Find the first "processor" node to get the CPU. Use the "product" key. If not present,
+        // fallback to "vendor" (like in s390x).
         let cpu = value
             .find_by_class("processor")
             .first()
--- a/rust/agama-manager/src/hardware.rs
+++ b/rust/agama-manager/src/hardware.rs
@@ -18,7 +18,7 @@
 // To contact SUSE LLC about this file by physical or electronic mail, you may
 // find current contact information at www.suse.com.
 
-use agama_utils::api::manager::HardwareInfo;
+use agama_utils::{api::manager::HardwareInfo, arch::Arch};
 use serde::Deserialize;
 use serde_with::{formats::PreferMany, serde_as, OneOrMany};
 use std::{
@@ -68,7 +68,14 @@ impl Registry {
 
     pub async fn read(&mut self) -> Result<(), Error> {
         match &self.source {
-            Source::System => self.read_from_system().await,
+            Source::System => {
+                self.read_from_system().await?;
+                // On s390x, enrich the system node with model info from read_values if needed
+                if Arch::is_s390() {
+                    self.read_s390_model().await.ok();
+                }
+                Ok(())
+            }
             Source::File(ref path) => self.read_from_file(path.clone()),
         }
     }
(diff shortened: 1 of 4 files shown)


rust/agama-manager/src/hardware.rs
```diff
@@ -98,6 +105,50 @@ impl Registry {
...
+    /// or "LinuxONE III LT1" from "IBM LinuxONE III LT1".
+    ///
+    /// Expected format: "8561 = IBM LinuxONE III LT1"
+    fn parse_s390_version(output: &str) -> Option<String> {
+        let line = output
+            .lines()
+            .find_map(|l| l.split_once('=').map(|(_id, model)| model.to_string()))?;
+
+        if let Some((_, version)) = line.split_once("IBM") {
+            Some(version.trim().to_string())
```
jreidinger, line 146: more compress version can be `Some(line.split_once("IBM")?.1.trim().to_string())`

rust/agama-manager/src/hardware.rs
```diff
@@ -191,16 +242,36 @@ impl HardwareNode {
...
+    ///
+    /// * `class`: class to search for (e.g., "disk", "processor", etc.).
+    pub fn find_first_by_class_mut(&mut self, class: &str) -> Option<&mut HardwareNode> {
+        if self.class == class {
+            return Some(self);
+        }
+
+        for children in &mut self.children {
+            if let Some(result) = children.find_first_by_class_mut(class) {
+                return Some(result);
```
jreidinger: I find that double Some a bit confusing. What about

```
let result = children.find_first_by_class_mut(class)
if result.is_some() {
  return result;
}
```

rust/agama-manager/src/hardware.rs
```diff
@@ -191,16 +242,36 @@ impl HardwareNode {
...
+    }
 }
 
 impl From<&HardwareNode> for HardwareInfo {
     fn from(value: &HardwareNode) -> Self {
         let cpu = value
             .find_by_class("processor")
             .first()
-            .and_then(|c| c.product.clone());
+            .and_then(|c| c.product.clone().or(c.vendor.clone()));
```
jreidinger, line 272: I think your future I will thank you if you write reason here e.g. with example input..and similar to memory. So we know "Why" we are doing it.

rust/agama-manager/src/hardware.rs
```diff
@@ -323,6 +394,34 @@ mod tests {
...
+        let model = Registry::parse_s390_version(output_with_prefix);
+        assert_eq!(model, Some("eServer zSeries 900".to_string()));
+
+        let invalid_output = "No equals sign here";
+        let model = Registry::parse_s390_version(invalid_output);
+        assert_eq!(model, None);
+
+        let empty_output = "";
+        let model = Registry::parse_s390_version(empty_output);
+        assert_eq!(model, None);
```
jreidinger, line 425: 👍

rust/agama-manager/src/hardware.rs
imobachgs: But less readable in my opinion.

jreidinger (approved):

PR #3722, opened 2026-07-14
refactor(web): serve installation summary at root path

dgdavid:
Based on @ancorgs's [comment](https://github.com/agama-project/agama/pull/3697#pullrequestreview-4692312501) and a later meeting with him, this supersedes #3697, keeping only the route/path change from it.

The installation summary (`OverviewPage`) is now served only at the root path `/` instead of `/overview`. It was already rendered at the index route, so this just drops the redundant `/overview` route, repoints the breadcrumb and post-submit summary links to `/`, and adds a `/overview` to `/` redirection for bookmarked URLs.

The reasoning is that almost all other nested routes (software, authentication, storage, iSCSI, etc.) are not prefixed with either `/overview` or `/installation`. From that perspective, having the **Installation** breadcrumb always navigate to `/`, with the installation summary shown by default whenever no higher-priority page (such as product selection) takes precedence, provides a simpler and more consistent routing model.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 14 16:59:31 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Align paths and navigation with the Installation breadcrumb
+  instead of the internal overview concept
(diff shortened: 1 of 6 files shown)


joseivanlopez (approved):
LGTM

PR #3724, opened 2026-07-15
Tune Tokio runtimes per binary

imobachgs:
Match each binary's Tokio runtime to its workload:

- **agama-web-server**: cap worker threads at `min(cores, 4)` instead of 32 — the installer's web server is low-concurrency and I/O-bound.
- **agama** and **agama-autoinstall**: switch to a `current_thread` runtime via `#[tokio::main(flavor = "current_thread")]` — short-lived, sequential, I/O-bound CLIs.

The multi-threaded runtime helper is now only used by the server (and `xtask`), so the `runtime` module moves from `agama-utils` to `agama-server`.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -6057,7 +6057,6 @@ version = "0.1.0"
 dependencies = [
  "agama-cli",
  "agama-server",
- "agama-utils",
  "clap",
  "clap-markdown",
  "clap_complete",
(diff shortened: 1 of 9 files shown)


imobachgs:
We decided to be conservative and just limit the max number of threads to 4 (#3729). We will consider the `flavor` change for 16.2.

PR #3726, opened 2026-07-15
service: More accurate model when reusing filesystems

ancorgs:
## Problem

When reusing an existing filesystem, the web UI shows a message like, for example,  "Current XFS at /dev/sb4". But the problem is that, in that example, "XFS" is not really the current file-system type (which will be kept), but the default file-system type that would be used if the device would actually be re-formatted.

It is just a "cosmetic" error, since the proper operation is actually performed.

## Solution

Fix the model to ensure it represents accurate information about reusing filesystems.

Added comments to make this better in the future (we already have a Trello card about it since a very long time).

## Testing

- Added a new unit tests
- Tested manually


ancorgs:
Closed in favor of the simpler #3732

PR #3727, opened 2026-07-15
feat(service): storage logical search operators

joseivanlopez:
## Storage logical search operators

Extends the storage search `condition` to support the logical operators **`and`**, **`or`**, and **`not`**, so device searches can combine multiple criteria instead of a single flat match.

Previously a `condition` accepted only a single leaf (`name`, `size`, or partition `number`). Now those leaves can be freely nested inside operator nodes, evaluated recursively across the full pipeline: JSON import/export, model conversions, config solvers, and config/model checkers. The schema and a full set of examples are updated accordingly.

### Operators

- **`and`** — array of conditions; all must match.
- **`or`** — array of conditions; at least one must match.
- **`not`** — single condition; matches when it does *not*.
- Operators nest arbitrarily and wrap the existing leaves (`name`, `size`, `number`).

### JSON examples

Match drives larger than 10 GiB, excluding `/dev/vdb`:

```json
{
  "search": {
    "condition": {
      "and": [
        { "size": { "greater": "10 GiB" } },
        { "not": { "name": "/dev/vdb" } }
      ]
    }
  }
}
```

Match either of two drives:

```json
{
  "search": {
    "condition": {
      "or": [
        { "name": "/dev/vda" },
        { "name": "/dev/vdb" }
      ]
    }
  }
}
```

Negate a size condition:

```json
{
  "search": {
    "condition": {
      "not": { "size": { "less": "10 GiB" } }
    }
  }
}
```

Nested operators (one of two drives, and not smaller than 1 GiB):

```json
{
  "search": {
    "condition": {
      "and": [
        { "or": [ { "name": "/dev/vda" }, { "name": "/dev/vdb" } ] },
        { "not": { "size": { "less": "1 GiB" } } }
      ]
    }
  }
}
```

Operators also apply to partition searches (using the `number` leaf):

```json
{
  "search": {
    "condition": {
      "and": [
        { "number": 1 },
        { "size": { "greater": "1 GiB" } }
      ]
    }
  }
}
```

### DIFF
--- a/rust/share/examples/storage/search_conditions.json
+++ b/rust/share/examples/storage/search_conditions.json
@@ -1,0 +1,117 @@
+{
+  "storage": {
+    "drives": [
+      {
+        "search": {
+          "condition": {
+            "and": [
+              { "size": { "greater": "10 GiB" } },
+              { "not": { "name": "/dev/vdb" } }
+            ]
+          }
+        }
+      },
+      {
+        "search": {
+          "condition": {
+            "or": [
+              { "name": "/dev/vda" },
+              { "name": "/dev/vdb" }
+            ]
+          }
+        }
+      },
+      {
+        "search": {
+          "condition": {
+            "not": { "size": { "less": "10 GiB" } }
+          }
+        }
+      },
+      {
+        "search": {
+          "condition": {
+            "and": [
+              {
+                "or": [
+                  { "name": "/dev/vda" },
+                  { "name": "/dev/vdb" }
+                ]
+              },
+              { "not": { "size": { "less": "1 GiB" } } }
+            ]
+          },
+          "ifNotFound": "skip",
+          "max": 1
+        },
+        "partitions": [
+          {
+            "search": {
+              "condition": {
+                "and": [
+                  { "number": 1 },
+                  { "size": { "greater": "1 GiB" } }
+                ]
+              }
+            }
+          },
+          {
+            "search": {
+              "condition": {
+                "or": [
+                  { "number": 1 },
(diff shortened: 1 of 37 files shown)


ancorgs (approved):
In general it looks good. But there is one aspect that I find kind of unexpected.

I somehow expected most of the classes at the `SearchConditions` namespace to have a common ancestor. I see several places in which the type of the accepted argument is basically a full list of the classes there and I also see several places where the exact type of a given condition is explicitly checked (eg. code like `is_a?(SearchConditions::Name)` or several `case` statements).

All that feels like there are too many components that are too aware of the existing conditions and how each of them work.

joseivanlopez:
> I somehow expected most of the classes at the `SearchConditions` namespace to have a common ancestor. I see several places in which the type of the accepted argument is basically a full list of the classes there and I also see several places where the exact type of a given condition is explicitly checked (eg. code like `is_a?(SearchConditions::Name)` or several `case` statements).

There is no base class yet because there is nothing in common. Each `SearchConditions::X` represents a node of the search tree of the config. And each node simply contains search values. Then, the solver needs to check the type of each node in order to know how to match a device with each condition.

> All that feels like there are too many components that are too aware of the existing conditions and how each of them work.

I think the solver is the only one that needs to check for the type of each search node. Note that the node itself hasn't logic for checking a device. It only contains values (e.g., name, fs type, etc). And the solver decides how to check that values with a device.

joseivanlopez:
Well, actually the the *converters* need to know the type of each search node tree too.

PR #3728, opened 2026-07-16
fix(web): prevent storage forms from flickering when changing mount point or filesystem

dgdavid:
## Summary

Changing the mount point or filesystem in the partition and logical volume forms caused the entire form to briefly disappear and show the page loading spinner. This happened because the automatic size hint is backed by a `useSuspenseQuery` whose query key depends on those values. Every key change triggered a new suspension that propagated to the page-level `Suspense` boundary.

This PR isolates the solver-dependent UI into its own component and wraps it in a local `Suspense` boundary. The query inputs are deferred using React's `useDeferredValue`, allowing the previous size hint to remain visible while the updated data is fetched. Since the solver is expected to respond quickly and the size hint is read-only guidance rather than editable state, briefly showing the previous value provides a smoother UX without compromising correctness.

## Implementation

Initially, it was explored a TanStack Query solution using [`placeholderData`/`keepPreviousData`](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function). Unfortunately, this does not work with `useSuspenseQuery`, as `placeholderData` is ignored and the hook still suspends whenever the query key changes.

The implementation therefore follows the same approach already used in the Overview page:

- Existing pattern:
  <https://github.com/agama-project/agama/blob/5c4ee6f4cf1355e26be2f1654b287329eb5f4d97/web/src/components/overview/OverviewPage.tsx#L23>
- React documentation for `useDeferredValue`:
  <https://react.dev/reference/react/useDeferredValue#showing-stale-content-while-fresh-content-is-loading>

Specifically, this PR:

- Extracts the automatic size note into its own component.
- Defers the mount point and filesystem values with `useDeferredValue`.
- Wraps only the solver-dependent content in a local `Suspense` boundary so the initial load remains localized while subsequent updates keep the previous content visible until the new result is available.

### Before

https://github.com/user-attachments/assets/97124d8c-c21a-4eed-8c54-1917c967f189

### After

https://github.com/user-attachments/assets/50357368-0258-4f05-ac8d-c3727e7331b4

## Documentation

The PR also documents this pattern in `web/src/components/form/conventions.md`, explaining:

- why the TanStack Query approach does not work with `useSuspenseQuery`,
- when `useDeferredValue` is the appropriate solution,
- and when this pattern should (and should not) be used in future forms.

---

Related to https://trello.com/c/RjALCqek (protected link)

---

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 16  07:06:02 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Prevent the partition and logical volume forms from flickering
+  when changing the mount point or file system
+  (gh#agama-project/agama#3728).
+
+-------------------------------------------------------------------
 Tue Jul 14 16:59:31 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Require translated text in user-facing component props, ensuring
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -3,7 +3,7 @@ Thu Jul 16  07:06:02 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Prevent the partition and logical volume forms from flickering
   when changing the mount point or file system
-  (gh#agama-project/agama#3728).
+  (gh#agama-project/agama#3728, bsc#1271524).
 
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
This certainly makes the user experience waaay smoother. LGTM

PR #3729, opened 2026-07-16
Limit the number of Tokio worker threads

imobachgs:
## Problem

The limit of Tokio worker threads is too high. For big machines, it can go up to 32 threads.

## Solution

Limit to 4 worker threads as it is more than enough.

## Testing

- Tested manually

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul 16 10:05:13 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Limit the number of Tokio worker threads to 4
+  (gh#agama-project/agama#3729).
+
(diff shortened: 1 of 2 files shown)


dgdavid (approved):

rust/package/agama.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Jul 16 10:05:13 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Limit the number of Tokio worker threads to 4
+  (gh#agama-project/agama#3729).
```
dgdavid: NP: isn't there any related bug report we can reference here?

imobachgs: We can use the old one when we set the limit to 32.

dgdavid (approved):

PR #3731, opened 2026-07-16
Do not set the dns search list for disabled IP protocols (bsc#1271481)

teclator:
### Summary

This PR resolves an issue where configuring a connection with disabled IP protocols failed in NetworkManager if a DNS search list was defined (`bsc#1271481`). Additionally, it corrects a conversion mismatch for `dnsSearchList` in the AutoYaST network reader.

- SUSE Bugzilla: [bsc#1271481](https://bugzilla.suse.com/show_bug.cgi?id=1271481)

### Problem
1. **NetworkManager Validation Failure:** NetworkManager validates settings and throws an error if IP-specific attributes (such as `dns-search`) are present on a protocol whose method is set to `disabled` (e.g., when disabling IPv6 entirely). The problem is not limited to what Agama serializes: when updating a connection, Agama merges its output on top of the connection already stored by NetworkManager, so contradictory values (e.g. a `dns-search` or `address-data` left over from a previous `manual` configuration) can be inherited from the existing connection and re-sent even though the method is now `disabled`.
2. **AutoYaST Conversion Error:** The AutoYaST parser was translating the `searchlist` key to snake_case (`dns_searchlist`) instead of the camelCase (`dnsSearchList`) expected by Agama's connection models, causing DNS search configurations from profiles to be ignored or improperly mapped.

### Solution
- **Rust (NetworkManager Client):**
  - Added `cleanup_disabled_ip_config` in `rust/agama-network/src/nm/dbus.rs`, called from `cleanup_dbus_connection`. When an `ipv4`/`ipv6` section has its method set to `disabled`, it drops every attribute except `method`. Because `cleanup_dbus_connection` runs *after* merging with the existing connection, this also removes contradictory values inherited from it, not just the ones Agama would serialize.
  - This centralizes the "disabled means empty" invariant in a single place, covering `dns-search` as well as `address-data`, `gateway`, `dns-data`, routes, etc.
- **Service (AutoYaST compatibility):**
  - Corrected the mapping from `dns_searchlist` to `dnsSearchList` in `service/lib/agama/autoyast/network_reader.rb` and `service/share/autoyast-compat.json`.

### Testing
- **Rust:** Added unit tests to `rust/agama-network/src/nm/dbus.rs`:
  - `test_ip_config_emptied_when_disabled` merges an updated connection that disables both protocols on top of an existing connection carrying `manual` values, and asserts that only the `method` attribute survives for both `ipv4` and `ipv6`.
  - `test_ip_config_kept_when_not_disabled` verifies that a non-disabled method keeps its configuration, including `dns-search`.
- **Service:** Updated AutoYaST integration unit tests (`connections_reader_test.rb` and `network_reader_test.rb`) to assert the correct mapping using `dnsSearchList`.
- **Manual Verification:** Performed manual installations ensuring NetworkManager correctly applies settings without errors when protocols are disabled.

### TODO

The current solution will fix the problem if the IPv{4,6} protocols are disabled but would be nice we validate also a profile with the method4 or method6 as disabled and conflicting configuration like the gateway{4,6} as well as addresses for the disabled protocol but that is by now out of the scope of this PR.

### DIFF
--- a/rust/agama-network/src/nm/dbus.rs
+++ b/rust/agama-network/src/nm/dbus.rs
@@ -452,21 +452,20 @@ fn ip_config_to_ipv4_dbus<'a>(
         ip_config.link_local4
     };
 
+    let method4 = ip_config.method4.unwrap_or(Ipv4Method::Auto);
+
     let mut ipv4_dbus = HashMap::from([
         ("address-data", address_data),
         ("dns-data", dns_data),
-        ("dns-search", ip_config.dns_searchlist.clone().into()),
         ("ignore-auto-dns", ip_config.ignore_auto_dns.into()),
-        (
-            "method",
-            ip_config
-                .method4
-                .unwrap_or(Ipv4Method::Auto)
-                .to_string()
-                .into(),
-        ),
+        ("method", method4.to_string().into()),
         ("link-local", (link_local as i32).into()),
     ]);
+
+    // Only set dns-search if IPv4 is not disabled to avoid NetworkManager errors
+    if method4 != Ipv4Method::Disabled {
+        ipv4_dbus.insert("dns-search", ip_config.dns_searchlist.clone().into());
+    }
 
     if !ip_config.routes4.is_empty() {
         ipv4_dbus.insert(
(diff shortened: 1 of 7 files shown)


imobachgs (approved):

imobachgs (approved):

rust/agama-network/src/nm/dbus.rs
```diff
@@ -411,6 +416,26 @@ pub fn cleanup_dbus_connection(conn: &mut NestedHash) {
...
 
+/// Empties the IP configuration when its method is "disabled".
+///
+/// When the IPv4 or IPv6 method is disabled, any other attribute (e.g. "address-data",
+/// "dns-search", "gateway", ...) is meaningless and could even be contradictory. As this runs
+/// after merging the current configuration with the given one, it drops any such value that could
+/// have been kept from the original connection, leaving only the "method" attribute.
+///
+/// * `ip`: IPv4 or IPv6 section of a connection.
+fn cleanup_disabled_ip_config(ip: &mut HashMap<&str, zbus::zvariant::Value>) {
```
jcronenberg, line 427: Thing is, some ip settings are allowed to be set even when disabled, so I am not sure if a blanket removal of everything really is the right call.

jcronenberg: But I don't want to block this, I think it's fine as is and should pretty much never have an actual impact.

teclator: > Thing is, some ip settings are allowed to be set even when disabled, so I am not sure if a blanket removal of everything really is the right call.

That should be taking into account then, and we should provide the same behavior although it is somehow unexpected to have it disabled and set some IP at least to me. Currently `nmcli` ask you about deleting the addresses, but that is something we would not do in case of unattended installation.

I will merge it as it is for fixing the current issue, but will start thinking in some alternative.

PR #3732, opened 2026-07-16
web: Avoid displaying a wrong reused filesystem type

ancorgs:
## Problem

When reusing an existing filesystem, the web UI shows a message like, for example,  "Current XFS at /dev/sb4". But the problem is that, in that example, "XFS" is not really the current file-system type (which will be kept), but the default file-system type that would be used if the device would actually be re-formatted.

It is just a "cosmetic" error, since the proper operation is actually performed.

## Solution

Omit the filesystem type. So instead of "Current XFS at /dev/sb4" the string "Current /dev/sdb4" would be used. The string was already there for other cases, so there is no impact in the translations and the meaning of the sentence basically stays.

It would also be possible to inspect the current system to calculate the real filesystem type that will be re-used and display that instead. But the implemented fix is much easier and good enough for the SLE RC phase (in which we want to be conservative with the changes).

## Testing

- Tested manually

## See also

Another solution was implemented in the model layer (see #3726), but we are not sure whether is formally correct, so we opted for this simpler solution for now.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Jul 16 15:10:34 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Do not display the type when reusing filesystems (bsc#1271008).
+
+-------------------------------------------------------------------
 Thu Jul 16  07:06:02 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Prevent the partition and logical volume forms from flickering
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

dgdavid:
> It would also be possible to inspect the current system to calculate the real filesystem type that will be re-used and display that instead. But the implemented fix is much easier and good enough for the SLE RC phase (in which we want to be conservative with the changes).

Can this be considered as a follow-up nicety for 16.2?

dgdavid:
> > It would also be possible to inspect the current system to calculate the real filesystem type that will be re-used and display that instead. But the implemented fix is much easier and good enough for the SLE RC phase (in which we want to be conservative with the changes).
> 
> Can this be considered as a follow-up nicety for 16.2?

Answer was in the change itself, https://github.com/ancorgs/agama/blob/7876db1671efba7170700f180f51f6d088ff3f5f/web/src/components/storage/utils/partition.tsx#L54-L57

PR #3745, opened 2026-07-21
fix(web): truncate long device names in storage tables

dgdavid:
## Summary
- Long device names (a multipath WWID, a named md array...) no longer overflow the 15ch-wide "Device" column in `DrivesTable`, `VolumeGroupsTable` and `MdRaidsTable`
- Adds `TruncatedDeviceName`, a small component that truncates the name, keeps the full name reachable for screen reader and mouse users, and adds zero extra keyboard tab stops

Related to https://trello.com/c/Egc8FPNf (protected link)

## The bug
The "Device" column has always been pinned to a 15ch-wide cell, but the name itself was never actually truncated to fit it. Short names like `sda` or `nvme0n1` are fine, but anything longer (a multipath device's WWID, a named RAID array) just overflows the column with no ellipsis and no way to see the rest.

## What we tried first: PatternFly's Truncate
PatternFly 6.3 improved its `Truncate` component with a "reactive" mode: instead of guessing a fixed character count, it measures the actual column width at runtime (via `ResizeObserver`) and truncates only as much as needed. That looked like the obvious fit here, so it went in first.

Two things ruled it out:

1. **A visible rendering gap.** In the middle-truncation mode, PF lays the leading and trailing text out as two CSS grid columns, and reserves a hardcoded 6ch minimum width for the leading one. When the actual (already-clipped) text is shorter than that minimum, the grid track doesn't shrink to match, leaving a visible empty gap between the ellipsis and the trailing characters. Confirmed visually, not just in theory.
2. **It solves a problem we don't have.** The whole point of the reactive mode is to adapt to a column whose width isn't known ahead of time. Our "Device" column width is fixed (`15ch`, set inline). We already know the truncation point before rendering, so paying for a `ResizeObserver` per cell just to re-derive a number we already have is wasted work, on top of the rendering bug.

So instead: reuse the fixed-length middle-ellipsis truncation Agama already has (`deviceBaseName(device, true)`, previously only used for menu/description text), applied directly to the table cell.

## Accessibility
A truncated name still needs to be available in full, so `TruncatedDeviceName`:
- Puts the full name in a visually-hidden span, so it becomes the element's accessible name for screen reader users with no interaction required
- Wraps it in a tooltip (`VisualTooltip`) that mirrors the full name visually for mouse users on hover

Neither adds a keyboard tab stop. These tables select via a radio group (`selectionMode="single"`), so today the whole table costs a single tab stop for selection: Tab lands on it once, arrow keys move between rows from there. Adding a stop per truncated name would multiply that by the number of long names in the table, friction for information a keyboard user reaching for the next control doesn't need at that moment.

**Known gap:** a sighted, keyboard-only user with no mouse and no screen reader has no way to reveal the full name today. Closing that fully would mean attaching a tooltip to the row's own selection control instead of the name, so it rides that control's existing tab stop. Not done here: PatternFly has no ready way to attach a tooltip to the checkbox/radio its `Td select={...}` shorthand renders, so it needs changes beyond this component (a candidate for its own follow-up PR).

## Playground screenshot

localhost_8080_ (15)

Screenshot_2026-07-21_10-05-17

## Follow-ups

- `deviceBaseName(device, true)` (and `baseName`, `deviceLabel`) take a bare `truncate?: boolean`, unreadable at the call site and with a truncation length that's a single hardcoded constant, not adjustable per caller. Planned as its own follow-up: replace it with an options object (`{ maxLength?: number }`), so e.g. this PR's `TruncatedDeviceName` could pass the table's actual 15-char width instead of borrowing the generic 17-char default used for menu text. A boolean can only ever mean yes/no; an options object also scales cleanly if another option shows up later (e.g. truncation position, a custom omission marker), as a new named field instead of another signature rewrite. Scoped separately since it touches ~15 call sites across menus, headers and forms, unrelated to the bug fixed here.
- Closing the sighted-keyboard-only gap noted above: attach the tooltip to each row's selection checkbox/radio instead of the name. Needs `SelectableDataTable` to stop delegating to `Td`'s `select` shorthand and render the checkbox/radio itself, since PatternFly doesn't expose a way to attach a tooltip to it otherwise. A bigger, riskier change to a shared component, so deferred rather than bundled here.


joseivanlopez (approved):
LGTM

PR #3746, opened 2026-07-21
refactor(web): improve storage device names truncation

dgdavid:
> [!NOTE] 
> Follow-up of https://github.com/agama-project/agama/pull/3745/, which
fixed long device names overflowing the storage tables. That fix added a
`truncate` boolean to the `baseName` / `deviceBaseName` / `deviceLabel`
helpers; this PR turns that ad-hoc flag into a small, explicit options API.
>
> #3745 must be merged first.

## Summary

- Replace the `truncate?: boolean` argument with a `{ truncate, maxLength, omission, tolerance }` options object, so call sites read clearly and can control the length, the marker, and how far a name may exceed the limit before it is shortened at all.
- Let `TruncatedDeviceName` receive `maxLength` as a prop, so each table sets it next to its own column width instead of the component hardcoding one.

## What changed

- **Options object.** `baseName`, `deviceBaseName` and `deviceLabel` take an optional `BaseNameOptions`:
  - `truncate` opts into shortening (replaces the old boolean).
  - `maxLength` is the length to truncate to; it defaults to the shared sentence-text length, so callers that want the previous behavior just pass `{ truncate: true }` and never repeat the constant.
  - `omission` is the middle marker (default `"…"`), now configurable; the truncation math accounts for its length so the result always stays within `maxLength`.
  - `tolerance` is how far a name may exceed `maxLength` and still be shown in full. It defaults to `omission.length + 2`, so truncation only kicks in when it saves clearly more width than the marker itself occupies (no swapping a few real characters for a similar-length marker).
- **`TruncatedDeviceName` takes `maxLength`.** The component no longer hardcodes a width; each storage table passes the value beside its column style, keeping the number and the width it fits within in one place.
- **Dropped a stale FIXME** pointing at PatternFly's `Truncate`, found unsuitable for the fixed-width table case in #3745.

### DIFF
--- a/web/src/components/storage/DrivesTable.tsx
+++ b/web/src/components/storage/DrivesTable.tsx
@@ -70,7 +70,7 @@ export default function DrivesTable({
   const columns = [
     {
       name: _("Device"),
-      value: (device: Storage.Device) => <TruncatedDeviceName device={device} />,
+      value: (device: Storage.Device) => <TruncatedDeviceName device={device} maxLength={13} />,
       sortingKey: "name",
       pfTdProps: { style: { width: "15ch" } },
     },
--- a/web/src/components/storage/DriveHeader.tsx
+++ b/web/src/components/storage/DriveHeader.tsx
@@ -99,5 +99,5 @@ const Text = (drive: ConfigModel.Drive): string => {
 };
 
 export default function DriveHeader({ drive, device }: DriveHeaderProps) {
-  return sprintf(Text(drive), deviceLabel(device, true));
+  return sprintf(Text(drive), deviceLabel(device, { truncate: true }));
 }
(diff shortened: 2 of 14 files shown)


dgdavid (commented):
This is an internal refactor for improving a helper without not visual user impact. Thus, changelog entry is actually not needed. The relevant one if the added in #3745

web/src/components/storage/DrivesTable.tsx
```diff
@@ -70,7 +70,7 @@ export default function DrivesTable({
   const columns = [
     {
       name: _("Device"),
-      value: (device: Storage.Device) => <TruncatedDeviceName device={device} />,
+      value: (device: Storage.Device) => <TruncatedDeviceName device={device} maxLength={13} />,
```
joseivanlopez, line 73: Why this fixed `maxLength`? I mean, why 13?

dgdavid: Because column width is 15ch and usually is better to truncate a bit below these "ch" units.

joseivanlopez (approved):
LGTM

PR #3749, opened 2026-07-21
fix(web): avoid a flash from Agama colors to a product's own palette

dgdavid:
## Problem

When a product ships a custom appearance skin (`assets/products/<id>.css`, see `doc/product_theming.md`), the browser paints with Agama's stock colors first and only switches to the product's palette once the product config
resolves from the backend and `useProductAppearance` injects the stylesheet link. Anything rendered before that (most visibly the full-screen loading spinner) shows the wrong colors, then flips. This is a visible flash on
every load/reload of a themed product.

Light/dark/high-contrast already avoid the equivalent problem: a small inline script in `index.html` applies the persisted choice before React boots. The product skin had no equivalent.

## Fix

- `use-product-appearance.ts` now mirrors the active product id to `localStorage` (`agm-product-id`), alongside the existing behavior of injecting/updating the `<link rel="stylesheet" href="assets/appearance/<id>.css">`.
- `index.html` gains a second pre-boot script, matching the existing light/dark one: it reads `agm-product-id` and, if present, creates that same `<link>` element (same id) before first paint. Once the app boots and the real product resolves, `useProductAppearance` finds the existing element by id and updates it in place rather than creating a duplicate.
- A missing/404 stylesheet is already harmless (documented in `doc/product_theming.md`), so an unthemed product or a first-ever visit behaves exactly as before.

Known, accepted trade-off: if the product changes between two loads, the pre-paint script briefly shows the *previous* product's colors until the real query resolves and swaps the stylesheet. Same category of trade-off the existing script already accepts for OS-level "system" preference changes between loads.

### Before

https://github.com/user-attachments/assets/ac171efa-181e-40e2-ba19-dbba7ac03c6d

### After

https://github.com/user-attachments/assets/1ab396e3-0795-4fb1-ab22-b6c0732cc7e0


joseivanlopez (approved):
LGTM

PR #3751, opened 2026-07-23
software: do not copy installation repositories to the target

### LINKED ISSUE #1458: Make creation of /etc/zypp/repos.d entries optional
Leap 16.0 is going to use openSUSE-repos for repo management by default, so repo definitions created by Agama are in our case redundant.

The package itself maintains entries under /etc/zypp.repos.d/(openSUSE:*|NVIDIA:*).repo 

We do have a rpmsave / archiving section for default repo definitions in %post, however it would be nice if we would not create these in the first place.

https://build.opensuse.org/projects/Base:System/packages/openSUSE-repos/files/openSUSE-repos.spec?expand=1

image

imobachgs:
Fixes #1458 and [bsc#1264277](https://bugzilla.suse.com/show_bug.cgi?id=1264277).

The product definition (`products.d`) can declare installation repositories, registered in libzypp as `agama-0`, `agama-1`, etc. so packages can be installed from them. These were being copied to the installed system by `SoftwareAction::Finish`, which is not desired.

This adds a `remove_installation_repos` step to `finish()` that removes installer-only repositories (aliases with the reserved `agama-` prefix, plus the Driver Update Disk repo) before the zypp config is copied to the target. It folds in the former `remove_dud_repo` step. libzypp deletes the `.repo` file and cached metadata on removal, so `copy_files` finds nothing left behind.

Failures removing a single repository are logged and skipped so the rest are still processed.

Here is the output of `zypper lr` after installing the system:

image


rust/agama-software/src/zypp_server.rs
```diff
@@ -55,6 +55,22 @@ use crate::{
 
 const GPG_KEYS: &str = "/usr/lib/rpm/gnupg/keys/gpg-*";
 
+/// Alias prefix reserved for the installation repositories created by Agama corresponding to the
+/// product definition (see `build_repo` in `model::state`). They are named `agama-0`, `agama-1`,
+/// etc. and must not be copied to the
```
joseivanlopez: NP: unexpected early line-break.

joseivanlopez (approved):
LGTM

PR #3753, opened 2026-07-23
refactor(web): rework SplitInfoLayout into SideBySideLayout, fix title/body alignment

dgdavid:
## TL;DR

Renames `SplitInfoLayout` to `SideBySideLayout` and fixes its markup (explicit grid areas, RTL-safe alignment, title paired with body) to replace its previous fragile implementation across all 7 screens it backs.

> [!NOTE]
> 
> No big impact, does not deserve changelog entry
> 

## Motivation
`SplitInfoLayout` backs seven screens: login, error, and every installation status screen. It was put together quickly to get their two-column look in place, and it's worth revisiting now that there's room to work on the details, because a few things are worth fixing/improving now:

- Alignment was applied by matching PatternFly's generated inline `style` attribute by substring (`[style*="Order-on-md: 3"]`). Any PF change to that internal value, or any reordering of the props, would silently break alignment with no type error and no test failure.
- The prop names (`firstRowStart`, `firstRowEnd`, `secondRowStart`, `secondRowEnd`) describe grid positions in the desktop layout only; the component's own docs needed a note admitting the names don't match what happens on small viewports, where everything just stacks.
- `secondRowEnd` had zero real callers and `iconSize` was never overridden, yet both stayed part of the API, plus a `rowSpan` ternary that existed only to compensate for `secondRowEnd` always being empty.
- Three separate consumers hardcoded `style={{ minInlineSize: "25dvw" }}` on their primary button, because the layout didn't size its own content column.
- On small viewports, the description rendered right-aligned while everything else centered: a desktop-only alignment utility was leaking into the mobile layout.
- The divider used `left: 50%`, a physical position that puts it on the wrong side of the content in an RTL locale.

## Summary
- Renames `SplitInfoLayout` to `SideBySideLayout` (component, props type, file, CSS class), with a semantic API (`icon`, `title`, `description`, `children`) replacing grid-coordinate prop names (`firstRowStart`, `secondRowEnd`, ...)
- Replaces `order`-based grid placement with an explicit `grid-template-areas`, so DOM order always matches reading order (icon, title, description, body) and no CSS selector has to reach into PatternFly's generated inline styles to apply alignment
- Title and description now share a single grid row with the body, with the body's top edge nudged down to match the title's line-box height, so the title reads as paired with the body's first line instead of floating above it
- Switches the divider and column alignment to logical properties, so the layout follows the document's writing direction instead of a hardcoded `left: 50%`
- Removes three consumer-level `style={{ minInlineSize: "25dvw" }}` hacks; the layout now owns that sizing itself
- Moves `InstallationFinished`'s reboot hint and `ServerError`'s "Check whether Agama server is running." message into the body column, pairing each with its related action instead of floating alone next to the title
- Drops redundant top margin and heading-styled text from `ErrorPage`'s error message, which were compensating for the old alignment

## Where to look
`SideBySideLayout` backs every screen touched here. No all screenshots attached, easiest is to check these live at both a narrow and a >=768px viewport width:

- **Login**  `#/login`. Directly navigable; also shown automatically whenever the session isn't authenticated.

  | Before | After |
  |-|-|
  | login_before | login_after |

- **Installation progress**  `#/installation/progress`. Directly navigable; also auto-rendered while an installation is running, on any URL under `/`.
- **Installation finished**  `#/installation/finished`. Directly navigable; also auto-rendered once installation completes. Check both the TPM-alert variant and the plain reboot-hint variant (`isGrub2WithTpm`) since the reboot hint moved column in this PR.

  | Before | After |
  |-|-|
  | complete_before | complete_after |

- **Installation exit**  `#/installation/exit`. Directly navigable, or reached via the reboot button from the "finished" screen.

  | Before | After |
  |-|-|
  | exit_before | exit_after |

- **Installation failed**  not a route; auto-rendered on any URL under `/` when the backend reports a failed installation stage. Easiest to trigger by forcing a backend failure.
- **Server error**  not a route; rendered whenever the D-Bus/HTTP connection to the installer closes unrecoverably. Trigger by stopping the `agama` service while the UI is open. The "Check whether Agama server is running." message moved column in this PR, same as the reboot hint above.

  | Before | After |
  |-|-|
  | error_before | server_error_after |

- **Error page**  not a route; it's the router's error boundary, only reached by an unhandled error inside a route loader under the protected tree. Lowest priority to check manually, hardest to trigger, and the change here (dropping a stray top margin and heading-sized text) is easy to verify by reading the diff instead.

  | Before | After |
  |-|-|
  | 404_before | 404_after |

## Follow-up

- Discuss whether `ErrorPage` should keep `variant="minimal"` (no masthead/header actions) or move to the standard page variant now that its layout has been reworked


joseivanlopez (approved):
LGTM

PR #3754, opened 2026-07-23
fix(web): reserve masthead space in Page's minimal layout

dgdavid:
Pages using the minimal layout (login, error pages, installation exit/failed) mount an empty masthead that collapses to near-zero height, so their content sits flush against the top of the viewport while the other edges already get PatternFly's page inset as spacing. Give the empty masthead that same inset as a minimum height.

> [!NOTE]
> 
> No big impact, does not deserve changelog entry
> 

## Screenshots

| Before | After |
|-|-|
| login_before |  |
| 404_before |  |
| exit_before |  |

### DIFF
--- a/web/src/assets/styles/components/_patternfly-overrides.scss
+++ b/web/src/assets/styles/components/_patternfly-overrides.scss
@@ -39,6 +39,17 @@
   }
 }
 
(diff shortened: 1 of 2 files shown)


joseivanlopez (approved):
LGTM

PR #3563, opened 2026-05-29
(POC): Dracut module for configuring HNV in the initrd

teclator:
This is just a revamp of https://github.com/agama-project/agama/pull/3221

- https://jira.suse.com/browse/PED-14534

# Add Hybrid Cloud Network (HCN) Support to Agama Live ISO

## Summary

This PR implements a new dracut module (`99hcnmgr`) that provides automatic Hybrid Cloud Network (HCN) configuration during the initramfs boot phase on IBM Power systems. This enables seamless network connectivity during live partition migration events on PowerPC (ppc64/ppc64le) architectures.

## Problem Statement

IBM Power systems use HCN (Hybrid Cloud Network) with bonded network interfaces to maintain connectivity during live partition migrations. When a partition migrates between hypervisors, the underlying network adapter changes (e.g., from SR-IOV to VNIC), but the bond interface ensures continuity. However, existing initramfs network configuration tools don't understand this PowerPC-specific device topology, requiring manual intervention.

## Solution

This PR introduces a dracut module that:

1. **Discovers HCN devices** from PowerPC device-tree (`/proc/device-tree`)
2. **Generates bonding configuration** automatically based on discovered topology
3. **Integrates with NetworkManager** via `nm-initrd-generator`
4. **Provides HCN-specific kernel parameters** (`rd.hcn.ip=`, `rd.hcn.route=`) for network configuration
5. **Persists network configuration** to the installed system
6. **Waits for device availability** after migration events (up to 3 minutes)
7. **Supports DNS configuration** via extended `rd.hcn.ip=` syntax

### Architecture

**Dracut Module Components:**

- **`module-setup.sh`**: Module registration and dependency declaration
  - PowerPC-only activation (`ppc64`/`ppc64le`) using standard `DRACUT_ARCH` with a fallback to `uname -m`
  - Clean systemd integration before NetworkManager starts
  - Declares explicit dependency on `"network-manager"` rather than a generic `"network"`
  - Only installs necessary binaries: `hcnmgr`, `hexdump`, `ofpathname`, `pseries_platform`, `awk`, `sed`
  - Bypasses the standard NetworkManager connection generator in built initramfs via post-install `sed` patches on NM services and hooks

- **`parse-hcnmgr.sh`**: Core discovery and configuration logic
  - Activates when `rd.hcn=1`, `rd.hcn.ip=`, or `rd.hcn.route=` is present (exits early if none found)
  - Scans device-tree for HCN-capable devices (PCI SR-IOV and VNICs)
[...]

PR #3755, opened 2026-07-23
HNV / HCN - fix multiple bonds defects (bsc#1272445)

teclator:
## Problem

This pull request addresses two critical defects that surface when configuring NetworkManager bonds from HCN (Hybrid Cloud Network) / HVN device-tree data inside the initrd.

### 1. Multiple IP/Route Assignments in Multi-Bond Setups
When more than one bond is discovered and the user passes `rd.hcn.ip=` or `rd.hcn.route=` arguments referencing different slave MACs, every IP and route configuration fell through and was attached to **every** bond instead of being mapped precisely to the bond owning that slave MAC.
* **Root Causes:**
  - The per-bond loop did not skip IP/route configurations meant for other bonds' slaves, falling back to an "unqualified config" and applying it anyway.
  - `FIRST_BOND` was computed incorrectly because `BOND_NAMES` is newline-separated, so `awk '{print $1}'` returned all bond names instead of just the first. This made the fallback dead code on multi-bond systems.
  - MAC matching was case-sensitive while slave MACs are stored lowercase in `MAPPINGS`. Passing uppercase MACs on the command line resulted in zero IP assignments (or wrong fallbacks).

### 2. Spurious Default DHCP Connection Created by NetworkManager
An extra default DHCP connection is created in `/run/NetworkManager/system-connections` alongside the correct HCN bond profiles. This default profile brings the HCN bond ports up independently via DHCP, which breaks the bond.
* **Root Cause:**
  - `NetworkManager-config-initrd.service` runs `nm-initrd-generator` before udev discovery completes and before `99hcn` has had a chance to create the HCN bond profiles. Since `nm-initrd-generator` does not understand `rd.hcn.*` parameters, it generates no connections early. Since `rd.neednet=1` is set, NetworkManager fabricates a default DHCP connection because `/etc/NetworkManager/system-connections` does not exist as a directory and no connections were yielded.

- **Bugzilla link:** [bsc#1272445](https://bugzilla.suse.com/show_bug.cgi?id=1272445)
- **Related Jira issue:** `jsc#PED-14533`
- **Requires**: `https://src.opensuse.org/GNOME/NetworkManager/pulls/9`
- **Original implementation:** `https://github.com/agama-project/agama/pull/3563`

### 1. Fix multiple IP/route assignments (Commit `7eadb57bad26`)
- Track other bonds' slaves (`OTHER_SLAVE_NAMES` / `OTHER_SLAVE_MACS`) and skip (`matched_other`) when an IP/route belongs to another bond's slave.
- Corrected the `FIRST_BOND` computation to be unquoted so `awk` sees a single record and returns the first token. This ensures unqualified fallback configs are correctly mapped only to the first bond.
- Implemented case-insensitive MAC matching using lowercased copies, keeping interface names matched as-is.

### 2. Regenerate NM connections after HCN (Commit `b36ae328b643`)
- Added `NetworkManager-config-initrd.service` to `After=` in `hcn-init-initrd.service` to explicitly sequence HCN after standard early generation.
- Added `regen_standard_connections()` in `parse-hcn.sh`, called in the success path right after HCN profiles are written to `/etc`:
  - **On NM ≥ 1.54:** Restarts `NetworkManager-config-initrd.service`.
  - **On Older NM:** Clears `/run/NetworkManager/system-connections/*` and re-runs the generator with `getcmdline`.
- Gating this on HCN profiles existing ensures the default DHCP fallback is only dropped when HCN has a valid configuration to replace it with.

## Testing

- **Added new integration tests:** Included BATS integration tests in `live/test/hcn/test_integration.bats` (implemented in commit `7eadb57bad26`).
- **Validated existing test suite:** Confirmed that all 15 existing HCN BATS integration tests pass successfully.
- **Tested manually**
```bash
initqueue:/root# cat /proc/cmdline 
BOOT_IMAGE=/boot/ppc64le/loader/linux splash=silent plymouth.ignore-serial-consoles TERM=linux  live.password=linux rd.break=initqueue rd.hcn.ip=10.2.2.65::10.2.0.1:255.255.255.0::2e-7a-3c-6a-1c-00:off rd.hcn.ip=10.2.2.66::10.2.0.1:255.255.255.0::2e-7a-3c-6a-1c-01:off ip=
initqueue:/root# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: env4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UNKNOWN group default qlen 1000
    link/ether 2e:7a:3e:73:69:04 brd ff:ff:ff:ff:ff:ff
    altname enx2e7a3e736904
3: env6: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond156efe56 state UNKNOWN group default qlen 1000
    link/ether 2e:7a:3e:73:69:06 brd ff:ff:ff:ff:ff:ff
    altname enx2e7a3e736906
4: env5: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond22f58d45 state UP group default qlen 1000
    link/ether 2e:7a:3e:73:69:05 brd ff:ff:ff:ff:ff:ff
    altname enx2e7a3e736905
5: enP32800p1s0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond22f58d45 state UP group default qlen 1000
    link/ether 2e:7a:3c:6a:1c:00 brd ff:ff:ff:ff:ff:ff
    altname enx2e7a3c6a1c00
6: enP32802p1s0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond156efe56 state UP group default qlen 1000
    link/ether 2e:7a:3c:6a:1c:01 brd ff:ff:ff:ff:ff:ff
    altname enx2e7a3c6a1c01
7: bond22f58d45: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 2e:7a:3c:6a:1c:00 brd ff:ff:ff:ff:ff:ff
    inet 10.2.2.65/24 brd 10.2.2.255 scope global noprefixroute bond22f58d45
       valid_lft forever preferred_lft forever
8: bond156efe56: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 2e:7a:3c:6a:1c:01 brd ff:ff:ff:ff:ff:ff
    inet 10.2.2.66/24 brd 10.2.2.255 scope global noprefixroute bond156efe56
       valid_lft forever preferred_lft forever
```
image


live/live-root/usr/lib/dracut/modules.d/99hcn/hcn-init-initrd.service
```diff
@@ -2,7 +2,7 @@
 Description=Hybrid Cloud Network config for NetworkManager (initrd)
 DefaultDependencies=no
 Wants=systemd-udev-trigger.service
-After=systemd-udev-trigger.service
+After=systemd-udev-trigger.service NetworkManager-config-initrd.service
```
aafeijoo-suse: `NetworkManager-config-initrd.service` has `Before=systemd-udevd.service systemd-udev-trigger.service`, so this should not be necessary... Besides, it only covers NetworkManager >= 1.54

teclator: I agree

live/test/hcn/test_hcn-cmdline.bats
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env bats
```
imobachgs, line 1: I did not know that it was possible to write those tests.

imobachgs (approved):
Great job!

PR #3757, opened 2026-07-24
Properly detect when an AutoYaST profile is not found

imobachgs:
## Problem

After #3624, `agama config generate` does not return an error when processing an AutoYaST profile that does not exist. Hence, the `agama-autoinstall` binary thinks the profile was successfully loaded from `label://OEMDRV/autoinst.xml` and stops there, not processing the rest of predefined locations.

See [bsc#1272439](https://bugzilla.suse.com/show_bug.cgi?id=1272439).

## Solution

* Return an error when the profile is not found.
* Still do not process the profile if it was not found (that was the original problem that #3624 mitigated).

## Testing

* Added a unit test.
* Manually tested.

### DIFF
--- a/service/bin/agama-autoyast
+++ b/service/bin/agama-autoyast
@@ -58,4 +58,7 @@ rescue Agama::Commands::CouldNotWriteAgamaConfig => e
   warn "Could not write the Agama configuration:\n\n"
   warn e.full_message
   exit 4
+rescue Agama::Commands::ProfileNotFound => e
+  warn "Could not find the AutoYaST profile"
+  exit 5
 end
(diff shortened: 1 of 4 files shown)


mchf:
The problem with returning an error in case of that predefined locations was that it caused "weird" internal error in agama web server in logs. There should be a note about that in the PR. Did you checks whether it happens with your fix or not?

imobachgs:
> The problem with returning an error in case of that predefined locations was that it caused "weird" internal error in agama web server in logs. There should be a note about that in the PR. Did you checks whether it happens with your fix or not?

No, there are no errors in the logs. See the screenshot below:

imagen

* It does not load the file from `OEMDRV` because it does not exist. That was the problem this pull request fixes. See that there is no backtrace.
* It continues with the rest of pre-defined locations. I loads the file in `file:///autoinst.xml`.

mchf (approved):
lgtm

PR #3758, opened 2026-07-24
fix: properly handle offline installations

imobachgs:
## Problem

If the network is not available, Agama does not work properly (#3341):

1. Online medium: Agama complains that it cannot read the product. That's actually true, but it does not allow to set up the network to read the repositories.
2. Offline medium: Agama complains that it cannot read the remote repositories, even if it does not need them for installing the product.

## Solution

- Bring back the method "Probe" action (it was there in pre-APIv2 times) to refresh the software metadata. Merge it with the existing "ProbeStorage" and "ProbeDASD". It would allow us to solve problem 1.
- Do not use the remote repositories when using the off-line medium.

The solution is quite conservative because we are in RC phase, so we wanted to explictly avoid introducing any API change (e.g., a mechanism to allow telling Agama to "ignore" something). Let's work on that for the 16.2 cycle.

## Technical details

- Replace the separate `Refresh`/`ProbeStorage`/`ProbeDASD` actions with a single generic `Probe` action (`Action::Probe { only: Option<Vec<Scope>> }`), used to refresh storage, DASD and software state through one HTTP action/CLI command
- Adapt `ManagerHTTPClient` and the web UI (`api.ts`, `model/action.ts`) to the new `Probe` action.
- Make sure the software service attempts to load the repositories on every `SetConfig` call, so the `software.load_source` issue is always kept up to date. It implies adding some delay on each `SetConfig` call.
- Skip the product's remote repositories entirely when a local installation repository (off-line/full media) is already present. This is an explicit trade-off: package updates that would only come from the remote repos won't be applied in that scenario.

Additionall, we did a small internal cleanup: consolidate the "well-known" predefined repository alias constants (`INSTALLATION_REPO_ALIAS`, `DUD_REPO_ALIAS`) in `agama-software::service`, and move `AGAMA_REPO_PREFIX` next to `build_repo()` in `model::state`, deriving the generated alias from that constant instead of a separate hardcoded literal.

## Testing

- Tested manually

## Follow up

- Allow to start a probing from the UI (on a separate PR).

### DIFF
--- a/rust/agama-utils/src/api/action.rs
+++ b/rust/agama-utils/src/api/action.rs
@@ -21,7 +21,7 @@
 use std::str::FromStr;
 
 use crate::{
-    api::{iscsi, l10n},
+    api::{iscsi, l10n, Scope},
     kernel_cmdline::KernelCmdline,
 };
 use schemars::JsonSchema;
@@ -39,6 +39,8 @@ pub enum Action {
     /// Performs a DASD probing on demand.
     #[serde(rename = "probeDASD")]
     ProbeDASD,
+    #[serde(rename = "probe")]
+    Probe { scopes: Option<Vec<Scope>> },
     #[serde(rename = "configureL10n")]
     ConfigureL10n(l10n::SystemConfig),
     #[serde(rename = "install")]
--- a/rust/agama-utils/src/api/action.rs
+++ b/rust/agama-utils/src/api/action.rs
@@ -34,13 +34,11 @@ pub enum Action {
     DiscoverISCSI(iscsi::DiscoverConfig),
     #[serde(rename = "activateStorage")]
     ActivateStorage,
-    #[serde(rename = "probeStorage")]
-    ProbeStorage,
-    /// Performs a DASD probing on demand.
-    #[serde(rename = "probeDASD")]
-    ProbeDASD,
+    /// Probes hardware and refreshes services (storage, DASD, software, etc).
+    ///
+    /// If `only` is not given, it probes everything.
     #[serde(rename = "probe")]
-    Probe { scopes: Option<Vec<Scope>> },
+    Probe { only: Option<Vec<Scope>> },
     #[serde(rename = "configureL10n")]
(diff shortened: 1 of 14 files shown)


web/src/model/action.ts
```diff
@@ -35,8 +37,8 @@ type ActivateStorage = {
   activateStorage: null;
 };
 
-type ProbeStorage = {
-  probeStorage: null;
+type Probe = {
+  probe: { only?: Scope[] | null };
```
jreidinger, line 41: I am surprised that I do not see anywhere change from ProbeStorage to Probe(only: ...). This feels strange.

imobachgs: Me too. They are not used in the UI (I confirmed with @joseivanlopez and @teclator). Now, they are used by the `agama probe` CLI. I guess we should extend "agama probe" to do the probing in iSCSI and zFCP too.

imobachgs: I will open the discussion about iSCSI and ZFCP. Additionally, I will keep this PR open until we update the UI (on a different PR against this branch, I guess). @dgdavid?

jreidinger (approved):

rust/agama-utils/src/api/action.rs
```diff
@@ -34,11 +34,11 @@ pub enum Action {
...
-    #[serde(rename = "probeStorage")]
-    ProbeStorage,
-    /// Performs a DASD probing on demand.
-    #[serde(rename = "probeDASD")]
-    ProbeDASD,
+    /// Probes hardware and refreshes services (storage, DASD, software, etc).
+    ///
+    /// If `only` is not given, it probes everything.
+    #[serde(rename = "probe")]
+    Probe { only: Option<Vec<Scope>> },
```
jreidinger, line 41: will it work without serde default that use None when param is omitted? because typescript signature used can cause omit of "only" parameter.

imobachgs: Yes, that's the reason to make it an `Option`.

jreidinger (approved):

PR #3759, opened 2026-07-24
feat(web): add a global live region for screen reader announcements

dgdavid:
## Summary

Screen reader announcements now go through one shared, predictable channel instead of each part of the UI rolling its own. That makes announcement behavior consistent everywhere instead of depending on how each component happened to implement it.

Announcements also keep working while a dialog is open, which they didn't reliably do before.

Three places already using ad-hoc announcements move over to the shared one: list add/remove/navigation feedback, the software patterns filter result count, and the hostname mode helper text. 

> [!NOTE]
> 
> **No new strings added: existing sentences are reused.**

## Follow-ups (separate PRs)

A few other screens still announce things the old way (the same pattern then hostname field used to follow). Moving them over is straightforward, but each one needs a quick call on wording and timing, so they're left for follow-up
PRs to keep this one focused:

- Password strength feedback
- The storage size mode helper text
- Table filtering feedback, starting with the DASD device table, though it's
  worth checking other filterable tables for the same pattern
- The installation progress indicator
- Hostname helper text: a screen reader user only hears the hostname helper text at the moment the mode changes, because that's when the announcement fires. Someone who tabs into that field later, without changing the mode,
won't hear it at all. The proper fix is to attach the helper text to the control itself, the same way a form field's error or hint text is normally wired up, so it's read whenever the control gets focus, not just at the moment of change. The shared dropdown component doesn't support that wiring yet.


joseivanlopez (approved):
LGTM

PR #3675, opened 2026-06-28
fix(web): polish keyboard focus ring and other UI rough edges

dgdavid:
## Summary

Follow-up polish addressing rough edges spotted after the recent look-and-feel work. Each fix targets a specific accessibility or consistency issue that affects usability and interaction methods.

- Make checkboxes and radios consistently sized across all forms
- Fix invisible loading spinners on disabled buttons
- Redesign button focus indicators to be fully visible and not clipped
- Improve skip-to-content link contrast in dark theme
- Fix form submission to scroll to alerts instead of page top

## Motivation

While implementing the dark theme and visual updates, several usability issues emerged:

1. **Invisible feedback**: Loading spinners disappeared when buttons were disabled, leaving users without visual confirmation that an action was processing
2. **Inconsistent sizing**: Radio buttons appeared at different sizes depending on which form they were in
3. **Clipped focus rings**: The outline-based focus indicator was frequently cut off by container overflow, making keyboard navigation harder to follow
4. **Low contrast**: The skip-to-content link (a critical accessibility feature) was nearly unreadable in dark theme
5. **Hidden alerts**: Form validation messages appeared below the viewport on pages with intro content, so users didn't see them

### Before

https://github.com/user-attachments/assets/1560a65c-53a8-46b1-9179-ed398051f559

https://github.com/user-attachments/assets/16705099-7f0d-47a1-bb4b-873bdfb64ac2

### After

https://github.com/user-attachments/assets/3a17421a-9718-416e-ad4c-9582c9b66676

https://github.com/user-attachments/assets/922d9ce5-30a2-4ef7-81d2-80f3d1a8a65f

## Related PRs

This builds on the look-and-feel work in:
- #3604 (dark theme foundation)
- #3607 (color tokens)
- #3613 (component updates)
- #3647 (PatternFly overrides)
- #3662 (theme refinements)

---

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

PR #3761, opened 2026-07-26
refactor(web): rework the page shell and fix the heading outline

dgdavid:
The user interface has gone through a few incarnations, and the shell carried something from each of them. Most of what this pull request removes was perfectly reasonable when it was written, then simply stopped being used along the way:

- The sticky action bar and its action helpers served the full-page selectors of the older layout. Pages eventually moved their actions into the forms they submit, and the last three users disappeared with the unified localization form (#3661). The sticky header helper became unused in that same pull request. It had been written for the layout replaced by the interface revamp (#3009).
- Scrolling the page to the top after submitting was how alerts were brought into view until the focus ring work (#3675) started scrolling the alert itself instead, which also works when the alert sits below the fold.
- The card passthroughs date back to the first version of the component (#1540): the header passthrough lost its last user in early 2025 (#1901), and the body passthrough when the storage page was rearranged (#2767).

Since nothing was actually broken, there was never much reason to revisit any of it. Now that the interface has settled, with a stable layout, pages following the same structure, and forms sharing one stack, it is a good moment to clean up the shell. Doing that also lets the accessibility fixes below happen at the root instead of page by page.

## Fixes included

**Heading outline.** A page title is a first-level heading while sections were third-level ones, so every page with sections skipped a level. Four pages that replace their content with a message also rendered a second first-level heading. Sections are now second-level headings, the surrounding headings move accordingly, and no page skips a level.

**Section naming.** A section without a name still rendered a `section` element, which ends up in the landmark list with nothing to identify it. One page had already worked around that by asking for a different element, a good sign the component was making the wrong choice. Sections now become landmarks only when they have a name and remain plain groupings otherwise. It is no longer possible to ask for both a visible title and an invisible label: the invisible one used to win, leaving the visible heading and the announced name out of sync.

**Clearer props.** The minimal variant inherited every header, progress, and questions prop while ignoring all of them, yet did not accept two props the standard variant does not even have. Each variant now declares only the props it accepts. The card passthrough becomes a named prop, the negative prop for skipping the progress monitor becomes a positive one, and the decision about when switching product makes sense moves out of the layout and into the menu that exposes the action.

## Notes for the reviewer

- Commits are meant to be read in order.
- Sections are no longer full height by default, since every call site with an opinion explicitly asked for that behaviour. The five that relied on the old default now request it by name, preserving today's appearance. Whether they should keep it is a visual decision that's easier to make while looking at those pages.
- The zFCP controllers section named itself with an invisible label while repeating the same text in bold in its body. That text now becomes the section title, and the controller becomes its description, so the information stays the same with a real heading. Its size remains unchanged: the level follows the page outline, the size follows the design.
- The decisions behind these changes are documented in a short conventions guide next to the component, so the next page does not have to rediscover them. It joins the existing conventions for forms and tests, with the intention of eventually bringing them together into an Agama style guide. Feedback on the wording is as welcome as feedback on the code.
- The move to the layout namespace is mechanical: one import change per page.

## Things worth checking
- [ ] Heading outline per page: one first-level heading, no skipped levels
- [ ] Storage, iSCSI, zFCP, Software, and the unsupported profile question still look correct, as they are the sections that keep full height
- [ ] The zFCP page shows the controllers section with its new heading
- [ ] "Skip to content" still lands on the page content
- [ ] The change product entry hides on product selection and during and after installation
- [ ] The header progress monitor still hides on the installation and finished pages

## Follow-ups (not part of this pull request)

- Split the two page layouts into their own modules, an internal refactoring worth reviewing separately.
- Decide whether the five full-height sections should keep that treatment.
- Merge the conventions documents into a single Agama style guide once the forms, tests, and pages conventions have settled.
- Retire the submit, cancel, and back actions the shell still exposes. The first two are redundant because pages on the form stack use the buttons provided by the form, while the third overlaps with breadcrumbs and the browser's back navigation. All three are marked `@deprecated` here so editors strike through their call sites and the count only goes down.


teclator (approved):
LGTM!

PR #3767, opened 2026-07-27
feat(service): storage search by partitions

joseivanlopez:
## Storage search: `partitions` condition

Adding a new `partitions` search leaf so partitionable devices can be matched by the **presence or properties of their partitions**. Implemented end-to-end: JSON schema, from/to JSON conversions and config solvers.

No production change was needed in the model layer (used by the UI) because it never inspects the condition tree.

### The `partitions` condition

A `partitions` leaf can appear in the search condition of `drives` and `mdRaids` (not `partitions`, `logicalVolumes` or `volumeGroups` — they have no partitions). It takes one of two shapes:

```jsonc
// Partitioned vs. unpartitioned
{ "search": { "condition": { "partitions": "any" } } }
{ "search": { "condition": { "partitions": "none" } } }

// At least one partition bigger than 10 GiB
{ "search": { "condition": { "partitions": { "any": { "size": { "greater": "10 GiB" } } } } } }

// No formatted partition at all
{ "search": { "condition": { "partitions": { "none": { "filesystem": "any" } } } } }

// Every partition formatted as xfs
{ "search": { "condition": { "partitions": { "all": { "filesystem": { "type": "xfs" } } } } } }

// At least 2 partitions / at most 3 partitions
{ "search": { "condition": { "partitions": { "count": { "min": 2 } } } } }
{ "search": { "condition": { "partitions": { "count": { "max": 3 } } } } }
```

`count` can be restricted to the partitions matching a condition — e.g. "between 2 and 5 partitions bigger than 10 GiB":

```jsonc
{
  "search": {
    "condition": {
      "partitions": {
        "count": {
          "condition": { "size": { "greater": "10 GiB" } },
          "min": 2,
          "max": 5
        }
      }
    }
  }
}
```

And it composes with the device-level operators — e.g. "a disk bigger than 100 GiB with an ext4 partition":

```jsonc
{
  "search": {
    "condition": {
      "and": [
        { "size": { "greater": "100 GiB" } },
        { "partitions": { "any": { "filesystem": { "type": "ext4" } } } }
      ]
    }
  }
}
```

### Refactor: search matchers

The condition evaluator (`config_solvers/search_matchers.rb`) was a single module that had grown to match three different subjects (the device, its filesystem and each of its partitions). It is now split into small classes under `config_solvers/search_matchers/`:


rust/share/storage.schema.json
```diff
@@ -988,6 +990,97 @@
...
+        { "required": ["min"] },
+        { "required": ["max"] }
+      ],
+      "properties": {
+        "condition": {
+          "description": "Optional condition; omit to count all partitions.",
+          "$ref": "#/$defs/partitionSearchCondition"
+        },
+        "min": {
+          "description": "Minimum number of matching partitions (must be >= 1).",
```
imobachgs, line 1073: np: for a human oriented description, I would use "must be greater than 0" or "must be > 0".

imobachgs (approved):
Nice work. LGTM.

PR #3768, opened 2026-07-27
refactor(web): reworks product not found screen

dgdavid:
## TL;DR

When the installer cannot get hold of the product to install, the overview no longer leaves users at a dead end: it explains the situation, points at the network settings, and offers to look again once a connection is in place, reporting progress while it does. The software page gets the same treatment when it cannot read the repositories. Wording is reused from what is already translated, so nothing waits on the string freeze.

## Main changes 

Gives users a way out when the installer cannot get hold of the product to install, instead of leaving them on a screen that never recovers.

- The bare "Product not found" empty state on the overview now gets the same treatment the software page already used for this situation: an icon, the explanation, the *"This might be due to network connectivity."* hint, and a **Go to network settings** link.
- That screen also offers a **Reload** action. Pointing at the network settings was not enough on its own: once a connection is set up, nothing makes the installer look for the product again, so users were left on the same dead end they started from. Reloading asks the backend to read the software information again.
- The software page offers the same two actions when it cannot read the repositories. It is the same problem seen from the other side, one showing up on the online medium and the other on the offline one, and until now that page reported the failure and stopped there.
- The install button is not rendered at all. It used to be shown disabled, offering an action that can never succeed in this state.
- The page intro goes away too. Both *"Take a moment to review the installation settings..."* and *"When ready, click on the [install] button..."* describe a flow that is not available, and the second one points at a button that is no longer there.
- The system information card stays: it is how users identify the machine the installer is running on, and that is still useful (arguably more useful) when something went wrong. Its **IPs** field is now not rendered when there is no address to show, instead of rendering an empty value.

## Depends on

Built on top of https://github.com/agama-project/agama/pull/3758, which brings back the action used to read the software information again. This one should land after it.

## Wording: reused because string freeze

Every sentence here is a string that already exists in the catalog. That is a deliberate constraint of the translation freeze, not a claim that these are the right words. They were written for other contexts and they show it:

> **Product not found**
> The product was not found in the repositories so it is not possible to proceed with the installation.
> This might be due to network connectivity.
> **Reload**  | **Go to network settings**

"The product was not found in the repositories" leans on installer vocabulary ("repositories") that means little to someone who just wants to install a system, and "This might be due to network connectivity" hedges twice ("might", "connectivity") right where the user needs a clear next step. **Reload** is the weakest of the three: it is borrowed from a screen where it means "reload the page", and here it means something rather different.

Once the freeze lifts, something more direct would serve better. For example:

> **Cannot find the product to install**
> The installer could not download the product from the network. Check the connection and try again.
> **Try again** | **Check network settings**

Worth a pass from someone with a UX writing hat on, rather than picking the above.

## A note on the network action looking like a button

**Go to network settings** is rendered with a button appearance, even though it is a link and stays one in the markup: an anchor pointing at its destination, keyboard and screen reader behaviour untouched.

The reason is purely aesthetic. It sits right next to **Reload**, which is a real button, and a button followed by a piece of underlined text read as two unrelated things rather than as the pair of choices they are. Giving both the same appearance makes them read as one set.

This is not a claim that links should generally look like buttons. Where an action stands on its own, it stays a plain link, which is why **Go to registration** on the software page is untouched. The installer does already take this liberty elsewhere, for instance in the actions accompanying the connections list, so it is worth agreeing on when it is warranted rather than deciding case by case. That belongs in the layout conventions.

## Follow-ups

- **Two alerts at once.** The software page can now render its own alert for unreadable repositories alongside the general one for the remaining issues. Worth checking whether both can really show at the same time and how that looks, since stacking alerts is something we would rather avoid. Whatever we settle on, the rule belongs in the layout conventions.
- **Fixed sentence versus reported description.** The empty state shows a fixed sentence rather than the description the backend sends for this situation. Worth deciding which should win: the reported description is closer to the real cause and may carry details the fixed sentence cannot (which product, which repository), but it is also written for a different context.

## Test plan

- [ ] Boot the installer with a product that cannot be found (for instance, with no network) and open the overview.
- [ ] The empty state explains the situation, links to the network settings and offers to reload.
- [ ] No install button and no page intro are shown.
- [ ] Set up a connection, come back and reload. The overview recovers on its own, without reloading the browser.
- [ ] While reloading, the button is busy and cannot be pressed again. The rest of the page stays usable.
- [ ] On a medium without network access, the software page reports that the repositories could not be read and offers the same two actions.
- [ ] With a product available, the overview behaves as before: intro, settings, system information, and a working install button with its confirmation dialog.
- [ ] The software page still shows its own unavailable states (missing registration, missing product, no patterns).

## Screenshots

> [!IMPORTANT]
> 
> All screenshots in the PR were took with fake data. They might be not fully accurate or even being a bit out-dated (lkike the Reload button, now labeled "Try again") Better to test manually

### Before

localhost_8080_ (24)

### After

localhost_8080_ (28)

localhost_8080_ (29)

localhost_8080_ (32)

localhost_8080_ (31)

## Progress on the empty state, in its own commit

https://github.com/agama-project/agama/pull/3768/commits/bb07a3a1fca9810d869b93d7c71e65a87298fc44 adds the usual progress overlay to the "Product not found" screen, so that while the installer looks for the product it says which step it is on. Until then the screen only turned its own actions busy, which shows that something is happening but not what, and shows nothing at all when the search was started from somewhere else.

It is deliberately a commit of its own, touching one file, so it can be dropped if the team would rather not have it.

The part worth highlighting is where it is mounted. Every other screen asks for this overlay at page level, which covers everything on that page for as long as the operation runs. **The overview cannot afford that**: it would be covered on every software operation, including the many the user can carry on working through. Mounting it on the empty state instead means it only ever exists while the product is missing, and by then there is nothing else to do on that screen. No CSS or new mechanism is involved, it is the same component the pages use.

The software page is untouched, since it already asks for the overlay at page level. Both of them wanting it would put two overlays on the same operation.

### Before

localhost_8080_ (29)

### After

localhost_8080_ (33)

## A screen that could never recover, found in review

Worth calling out because the reasoning is easy to repeat elsewhere.

Both the retry action and that overlay originally waited for the issues to be fetched again before treating the operation as over, which is the usual way of not showing stale content once something finishes. Here it was the wrong thing to wait for. The backend only reports issues that differ from the ones it already holds, so a read that changes nothing reports nothing, the issues are never fetched again, and the wait never ends.

That is exactly the case the button exists for. With the product still missing, the button stayed disabled and the overlay stayed up, and the only way out was reloading the browser: the opposite of what this PR set out to fix.

https://github.com/agama-project/agama/pull/3768/commits/5d53db06f762045c09a842725789b51d9add7be5 makes it wait for the operation itself instead. Data may arrive slightly after the screen unblocks, so the previous content can linger for a moment, which is a far better trade than getting stuck. Anywhere else waiting on a query that is only refreshed when something actually changed deserves the same look.

---

Related to https://github.com/agama-project/agama/issues/3341

---

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -2,7 +2,10 @@
 Mon Jul 27 12:41:09 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Show a proper empty state, pointing to the network settings,
-  when the product to install is not found
+  when the product to install is not found, and offer to read the
+  software information again once the connection is in place
+  (gh#agama-project/agama#3768).
+- Offer the same actions when the repositories cannot be read
   (gh#agama-project/agama#3768).
 
 -------------------------------------------------------------------
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jul 27 12:41:09 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Show a proper empty state, pointing to the network settings,
+  when the product to install is not found
+  (gh#agama-project/agama#3768).
+
+-------------------------------------------------------------------
 Mon Jul 27 11:39:41 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Adapt to use the "Probe" action of the API (gh#agama-project/agama#3341).
--- a/web/src/components/core/IssuesAlert.tsx
+++ b/web/src/components/core/IssuesAlert.tsx
@@ -27,9 +27,19 @@ import { _ } from "~/i18n";
 
 import type { Issue } from "~/model/issue";
 
-export default function IssuesAlert({ issues }: { issues: Issue[] }) {
+export type IssuesAlertProps = {
+  /** Issues to report. Nothing is rendered when the list is empty. */
+  issues: Issue[];
+  /**
+   * Optional buttons or links offering a way to deal with the issues, rendered
+   * below the description.
+   */
+  actions?: React.ReactNode;
+};
+
+export default function IssuesAlert({ issues, actions }: IssuesAlertProps) {
   if (isEmpty(issues)) return;
-  const props: Partial<AlertProps> = { isInline: true, variant: "warning" };
+  const props: Partial<AlertProps> = { isInline: true, variant: "warning", actionLinks: actions };
 
   if (issues.length === 1) {
     return <Alert {...props} title={issues[0].description} />;
--- a/web/src/components/core/UnavailableState.tsx
+++ b/web/src/components/core/UnavailableState.tsx
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) [2026] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+import React from "react";
+import {
+  Content,
+  EmptyState,
+  EmptyStateActions,
+  EmptyStateBody,
+  EmptyStateFooter,
+} from "@patternfly/react-core";
+import { Icon } from "~/components/layout";
+import Link from "~/components/core/Link";
+
+const EmptyStateIcon = () => <Icon name="apps_outage" />;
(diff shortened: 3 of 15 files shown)


dgdavid:
After a conversation with @lslezak, looks like better wording is already translated and can be brought back. Still having room for UX improvements, though, but would help to release better UI by now.

The software page used to have a screen for repositories that could not be loaded, with a link to retry, which is the very problem this PR is solving. It was removed at some point, and its texts went out of use with it. Going out of use does not throw the translations away, though: they are kept aside at the end of each translation file, and putting a text back in the code picks its translations up again on its own, with nothing for translators to do.

Two of them are worth taking back, described at the time as *"link for retrying failed repository load"* and *"progress message"*:

| text | Spanish | in all 13 languages we ship |
| --- | --- | --- |
| `Try again` | *Volver a intentar* | yes |
| `Loading the installation repositories...` | *Cargando los repositorios de instalación...* | yes |

They can be seen at the end of [`web/es.po` in agama-weblate](https://github.com/agama-project/agama-weblate/blob/master/web/es.po#L6343-L8534), and each of the other languages has its own.

This matters most for the reload action. **Try again** was written for retrying a failed repository load, which is somehow what the button does, while **Reload** is taken from a screen where it means reloading the browser page. Swapping them replaces the weakest wording in this PR with one that fits, in every language, without waiting for the freeze to lift.

**Loading the installation repositories...** is likewise a better fit than the generic *"Loading data..."* now used for the aria-label of the disabled button while the reply is on its way.

What stays untouched is the text of the alert on the software page. That one is the description the backend reports for the issue, and it is not frontend responsability to replace, however well another sentence might read.

imobachgs (approved):

PR #3772, opened 2026-07-29
feat(service): extend storage search conditions

joseivanlopez:
## Storage search: `driver`, `boss` and `id` conditions

The last three leaf conditions of the storage search, implemented end-to-end: JSON schema, from/to JSON conversions and config solvers.

As in the previous iterations, the model layer (used by the UI) needed no production change: it never inspects the condition tree.

### `driver` and `boss` (drives only)

`driver` is a plain scalar, so several values are expressed with the existing `or` operator.

```jsonc
// Any drive handled by the ahci kernel driver
{ "search": { "condition": { "driver": "ahci" } } }
```

A device reports a *list* of drivers, so `driver` matches if **any** of them is the given one.

The leaf `boss`  is a boolean and indicates whether the device is a BOSS (Boot Optimized Storage Solution). Use false to match any other device.

Composing with the operators — "a not BOSS disk bigger than 100 GiB":

```jsonc
{
  "search": {
    "condition": {
      "and": [
        { "boss": false },
        { "size": { "greater": "100 GiB" } }
      ]
    }
  }
}
```

### `id` (partitions only)

Accepts any of the partition ids managed by libstorage-ng (`esp`, `bios_boot`, `linux`, `swap`, `lvm`, `raid`, `ntfs`, …).

```jsonc
// The EFI system partition
{ "search": { "condition": { "id": "esp" } } }

// A formatted Linux partition
{
  "search": {
    "condition": {
      "and": [
        { "id": "linux" },
        { "filesystem": "any" }
      ]
    }
  }
}
```

It works anywhere a partition condition works: regular partition searches, partitions marked for deletion, and the `partitions` quantifiers of a `drives`/`mdRaids` search — e.g. "a disk with no ESP":

```jsonc
{ "search": { "condition": { "partitions": { "none": { "id": "esp" } } } } }
```

### Refactor: per-device search converters

`from_json_conversions/search.rb` was a single generic converter: any device could build any condition tree, and only the JSON schema stopped `{ "id": "esp" }` on a drive or `{ "driver": "sd" }` on a logical volume.

It is now one converter per device (`DriveSearch`, `MdRaidSearch`, `VolumeGroupSearch`, `LogicalVolumeSearch`, `PartitionSearch`), each declaring the leaves and sort criteria it accepts through `WithX` mixins — mirroring the structure already used by `config_solvers/search_matchers/`. An unsupported leaf or sort criterion now raises instead of being silently accepted, and the copy-pasted `and`/`or`/`not` handling that lived in three different places is gone.


mvidner:
@joseivanlopez I don't see any Bugzilla or Jira reference and there is no .changes entry

EDIT: ah, this is targetting `storage-search`, not `master`... but still, I should not have to *guess* the context.

service/lib/agama/storage/config_solvers/search_matchers/with_driver.rb
```diff
@@ -0,0 +1,56 @@
...
+          # Whether any of the drivers of the given device matches the condition node.
+          #
+          # @param node [Configs::SearchConditions::Driver]
+          # @param device [Y2Storage::BlkDevice]
+          #
+          # @return [Boolean]
+          def match_driver?(node, device)
+            return true unless node.driver
+
+            device.driver.include?(node.driver)
```
ancorgs: Are we sure `#driver` exist for all the possible devices (honest question, I have not checked)?

OTOH, I see that similar code at yast2-storage-ng uses `device.driver&.include?(whatever)` so I wonder whether it wouldn't be safer to do the same.

ancorgs:
I kind of remember that the three fields (disk driver, disk transport and partition id) are kind of controversial. So I will add comments with my findings about all of them. Then we can decide whether to adapt the name, whether to modify the check or whether to keep them as-is.

ancorgs:
### Aspects to consider about driver

Take into account `#driver` (name in singular) is actually an array of values coming from `hwinfo`.  And there is a very similar attribute called `#driver_modules`. Both typically look very similar but with subtle differences. See some examples of the differences at the description of https://github.com/openSUSE/hwinfo/pull/169

On the other hand, the pull request linked above shows that both fields are actually expected to contain all the involved drivers and that `hwinfo` really cares about that, so it is a trustworthy field with a clear (non-opinionated) meaning. :+1: I feel comfortable exposing it in its current form.

But some questions still arises, should we check for matches in both fields or just checking `#driver` is ok? Should we expose both fields?

Just for reference, AutoYaST allows to use both `driver` or `driver_modules` as key for an `skip_list`entry. But the user must specify the concrete property and match it with the concrete value.

```xml
<skip_list config:type="list">

    <!-- An entry like this could potentially match -->
     <listentry>
         <skip_key>driver</skip_key>
         <skip_value>iscsi_tcp</skip_value>
     </listentry>

    <!-- An entry like this could also potentially match -->
     <listentry>
         <skip_key>driver_modules</skip_key>
         <skip_value>iscsi-tcp</skip_value>
     </listentry>

    <!-- This cannot match -->
     <listentry>
         <skip_key>driver</skip_key>
         <skip_value>iscsi-tcp</skip_value>
     </listentry>

</skip_list>
```

ancorgs:
### Aspects to consider about transport

As I interpret [this comment](https://bugzilla.suse.com/show_bug.cgi?id=1176140#c22) (and the rest of the conversation) there might be more than one way to check the transport of a device.

yast2-storage-ng relies on libstorage-ng detection, which basically calls `lsscsi` and [process the output](https://github.com/openSUSE/libstorage-ng/blob/master/storage/SystemInfo/CmdLsscsi.cc#L78) to turn the returned string into an entry of the enum defined by libstorage-ng itself.

It would be good if we could find some kind of authoritative list of "data transports" for Linux and make sure what libstorage-ng does is actually the proper detection mechanisms instead of an opinionated value from an opinionated list of possible transports.

ancorgs:
### Aspects to consider about id

The partition id reported by libstorage-ng is actually an abstraction of the real partition ids. It offers a common set of values no matter whether the partition lives in a MS-DOS, GPT or DASD partition table.

We already had a problem [in the past](https://bugzilla.suse.com/show_bug.cgi?id=1211107) when that set of values was somehow redefined to reflect GPT ids more precisely. As a partial fix to that bug a function was introduced in libstorage-ng to somehow identify all the ids that used to be reported as LINUX in the past but that now have another value. Apparently that function was never used in YaST, so the change in behavior is likely still a real problem although the bug report was recently closed as part of a clean-up.

Taking all that into account, it must be noted that exposing the partition ID is not actually exposing the GPT IDs, but only a subset of them (referenced by some convenient names introduced by libstorage-ng) and is subject to the impact of similar changes in the future.

We may also consider offering somehow a way to filter any linux-related partition (that is, somehow exposing the `is_linux_partition` that was introduced to fix the originally reported bug).

service/lib/agama/storage/config_solvers/search_matchers/with_driver.rb
joseivanlopez: > Are we sure `#driver` exist for all the possible devices (honest question, I have not checked)?

Yes, the *drive* section always resolves a `BlkDevice` which defines `#driver`.

> OTOH, I see that similar code at yast2-storage-ng uses `device.driver&.include?(whatever)` so I wonder whether it wouldn't be safer to do the same.

According to its documentation, `BlkDevice#driver` always returns an Array (https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/blk_device.rb#L665). But I think using the *safe navigation operator* would be better. I see nothing in our code to ensure an Array if `hwinfo` returns nothing for it.

rust/share/storage.schema.json
```diff
@@ -895,6 +898,29 @@
...
+          "description": "Matches devices handled by the given kernel driver, as reported by hwinfo. A device can be handled by several drivers; the condition matches if any of them is the given one. Do not confuse with driver_modules, which values usually use underscore (virtio-pci vs virtio_pci) and subfix (ahci vs ahci_mod).",
+          "type": "string",
+          "examples": ["ahci", "sd"]
+        }
+      }
+    },
+    "searchConditionDellBoss": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["dellBoss"],
```
ancorgs: I'm not sure where `dellBoss` is a good name for the attribute. Mentioning a concrete vendor looks weird. What about making it simply `boss`? It would still be pretty concrete (is the name of the technology) and looks more future-proof in case Dell manages to turn BOSS into a widely used standard (I think that's their intention).

We could even go with a more generic name instead of using the concrete technology name. Something like `bootOptimized`. But I'm not sure. It's probably safer to stick to `boss` for now.

ancorgs (approved):
LGTM

PR #3773, opened 2026-07-30
refactor(web): move the storage view from a provider to the URL

dgdavid:
## TL;DR

The storage view goes back to the URL, where it lived until a scroll problem forced it into a context that every page has carried since. The scroll problem has an answer now, so the context goes away and reloading or sharing a storage link works again.

## Problem

The storage page keeps track of how it is being read: which sections are expanded, which tab is open. That belongs in the address bar, and it used to live there.

It moved into a context in October, because writing a search param sent the page back to the top on every expand and collapse, which is unbearable on a page this long. That solved the jumping, and it came with a cost:

- Reloading returns a collapsed page, and the reader starts again.
- A storage link cannot point at anything in particular, so explaining a layout means describing which sections to open.
- Every page in the installer carries the storage view, including the ones with nothing to do with storage, for as long as the session lasts.

## Solution

A search param write can say two things about itself: leave the scroll alone, and replace the current history entry rather than adding one. That is what makes the address bar usable for this. Sections open and close, the URL keeps up, the reader stays where they were, and the back button still goes back to the previous page instead of undoing toggles one at a time. Confirmed against a running installer.

The rest follows from it. Once the address describes the view, nothing needs to hold it, so the context and the provider wrapped around the application both go. The view now lives and ends with the page it belongs to instead of lasting the whole session. A reload rebuilds it because the address still says what it was, and a link carries it to whoever receives it.

Doing that takes three small hooks: one for a single value, one for a set of values packed into a single param, and one to clear several at once for a reset action. Each area keeps its own param names together, so nothing in the hooks knows about drives or tabs.

Which is the point of building it this way rather than the quickest way. Tables want the same thing for their filters and sort order, and whatever page comes next will want it too. They get a tested mechanism to reuse, and storage is simply its first caller.

## Details

A leftover goes too. Entering storage from the old sidebar used to reset the view on purpose, and the sidebar was removed in January. The half that did the resetting stayed behind, waiting for a signal that nothing has sent since.

Links into the storage forms need no work: they already carry the search along, through an option added when this state first lived in the URL and left untouched since. Opening a partition form takes the view with it, and Cancel or a save comes back to it through history.

### Coming back to storage

History brings the view back, which covers Cancel and the browser back button. Arriving through a plain link, such as the one on the overview page, opens storage fresh. That is what the sidebar used to force deliberately, and it is now what happens when a link carries no view with it.

### Where this is written down

Knowing when to put state in the URL is a decision, not a detail, and it will come up again on the next page. It is written down in `doc/ui/guidelines`, a new index intended to become the single place to look for interface conventions.

The conventions written earlier still sit next to the code they describe. The index lists them rather than moving them, so nothing is lost and they can be migrated one at a time, when each is next touched.

## Tests

A view can now be described by its address, so cases that used to need clicking through the interface are set up directly, and storage components no longer need anything wrapped around them to render.

- [x] Unit tests for the new tools, including deep links and multi-value params
- [x] Component tests: a section opens expanded when its address says so
- [x] `npm run test`, `npm run eslint`, `npm run check-types`
- [x] Manual: expanding and collapsing keeps the reading position, and reloading restores the expanded sections
- [ ] Manual: both tab strips, volume group sections, and "Reset to defaults" clearing the view in one step
- [ ] Manual: leaving a form and coming back restores the previous view

## Follow-up

Three of them, each on its own:

- **Tables.** Filters and sort order are the same kind of state, and putting them in the address makes a filtered table shareable and reload-proof. Free text needs a small pause before it is written, so it is worth doing per table rather than all at once.
- **Reading position.** Returning to a page still lands at the top. That is a different mechanism from anything here: nothing records where the reader was. It belongs in the page shell, so it waits for the shell rework to land.
- **Links that carry no view.** Should the overview link reopen whatever was last looked at, rather than opening storage fresh? The URL cannot answer that alone: the params belong to the storage address and leave with it, which is what keeps every other page's address clean. Something would have to remember the last view, so this one comes back to wanting a store. A much smaller one than the provider it replaces, since the URL stays the source of truth and the memory only speaks when a link arrives carrying nothing. Opinions welcome.


teclator (approved):
LGTM

PR #3775, opened 2026-07-31
Clean /system endpoint (licenses, proxy, software)

jreidinger:
## Problem

We would like to reduce API of /system API endpoint before release of SLES16.1. We would like to keep then backward compatibility as people can start using it.

## Solution

It removes the following elements:

- proxy section completely as it is not used and even not in web UI type script definition
- licenses section completely as webUI uses product license id and then ask for /license/<id>?lang= without using /system
- software->repositories as it is not displayed anywhere and not used.
- the `icon` from software patterns (not used).

## Testing

- adapted unit tests that failing
- *Tested manually* using devel iso with opensuse https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/system_cleaning/images/iso/ . Note: as it touched repositories and licenses, it is probably worth to test also full SLES16.1 iso which I cannot due to bandwidth issue.


joseivanlopez (approved):
LGTM

PR #3785, opened 2026-08-03
feat(web): keep the tables sort order in the URL

dgdavid:
## TL;DR

Follow-up of https://github.com/agama-project/agama/pull/3773/, which put the storage proposal UI state in the address and wrote the guideline saying filters and sort order belong there too. This PR address half of the sort order half, for the four tables that hold it in a local reducer.

## Summary

- Add `useSortedByParam`, a hook that keeps the sorting of a table in a search param named by the caller, on top of the generic hooks from #3773.
- Use it in the DASD, zFCP, iSCSI targets and network connections tables, which drop the sorting from their reducers.

## What changed

- **Columns are addressed by name, not by position.** The param carries the column identifier and, when descending, the direction: `?s=channel:desc`. A position would point at the wrong column the day one is inserted, breaking links people saved. A param naming a column that does not exist, or one that is malformed, reads as the table default, so an outdated link still opens the table.
- **One param, not two.** A sort click changes column and direction together. Ascending is where sorting starts, so it is left out (`?s=status`), and sorting that matches the table default removes the param altogether.
- **`sortingId` on `SelectableDataTableColumn`.** A string `sortingKey` already names a field; a function does not, so those columns declare an identifier. The DASD channel column and all four sortable connections columns need one.
- **The table component stays free of the router.** The hook does the URL work, and each page passes its own param name from its own module (`storage/ui-state-params.ts`, new `network/ui-state-params.ts`). That is what lets `DrivesTable`, `MdRaidsTable` and `VolumeGroupsTable` keep their local sorting inside `DeviceSelectorModal`: a picker opened over a page can never write to that page's param.

## Tests

- [x] `npm run test -- src`, `npm run eslint`, `npm run check-types`
- [x] Sort each of the four tables, reload, and get the same order
- [x] Paste the address in a second tab and get the same order
- [x] Sort back to the default column and check the param leaves the address
- [x] Edit the param by hand to name a column that does not exist and check the table still opens, sorted by its default
- [x] Open the device selector over a sorted storage page, sort inside it, close it, and check the page behind is untouched

## Notes

Filters are next, one PR per table, plus a layer that remembers the view of a page across visits.

---


teclator (approved):
LGTM

PR #3787, opened 2026-08-03
feat(web): keep the table filters in the URL

dgdavid:
## TL;DR

Follow-up of https://github.com/agama-project/agama/pull/3785/, which put the sort order of these same four tables in the address. This PR does the other half, the filters, and finishes what the guideline written in #3773 asked for.

Fltering a table now survives a reload and travels in a copied link, because the filters live in the address instead of a reducer. A single hook, `useFilterParams`, gives a table its filters and takes care of everything that makes URL-backed filtering fiddly, so each table needed only to say which filters it has.

## Problem

Filter a table, reload, and the full list is back. Send someone the link and they see everything, not what you were looking at. The back button leaves the page rather than undoing the filter. Every table solved it the same way, a reducer holding a bag of filters, so the state was private to a component that nobody else could ask.

## Solution

`useFilterParams` owns a table's whole filter:

```tsx
const { filters, setFilter, resetFilters, hasActiveFilters } = useFilterParams({
  status: choiceFilter(["active", "offline", "read_only"]),
  formatted: choiceFilter(["yes", "no"]),
  minChannel: textFilter(),
  maxChannel: textFilter(),
});
```

That is the entire wiring of the DASD table, the busiest of the four. The address reads `/storage/dasd?status=offline&minChannel=0.0.0150`.

Four decisions live inside the hook rather than in each table, which is the point of it: a table cannot get them wrong, and a fifth table gets them for free.

- **Typed text waits for a pause** (300 ms) before it reaches the address, so the history does not grow an entry per letter and the back button does not undo letters. The field itself keeps up with the keyboard, and so do the rows.
- **Clearing drops every param in one update**, along with any keystroke still waiting. Both are needed: separate updates each start from the params of their own render and undo one another, and a pending keystroke would otherwise land right after the clearing and put the filter back.
- **A value the filter does not offer reads as the default**, so an outdated link, or one edited by hand, still opens the table.
- **A filter at its default writes no param at all**, so one view has one address.

## Details

- **The address says what it holds.** A filter is declared once and that name is the param, so there are no constants to keep in sync and no way for two filters to collide. Reading `?status=offline&minChannel=0.0.0150` in a bug report tells you what to change, and since unknown values fall back to the default, changing it by hand is safe. The params carrying composed values were renamed to match (`sortBy=channel:desc`, `expanded=d0,d2`, `settingsTab`, `resultTab`), replacing the abbreviations of earlier PRs in this stack.
- **The reducers shrank or disappeared.** zFCP and connections held nothing but filters, so both are gone. DASD and iSCSI keep what describes an action being prepared: the selection, and a pending format request. Filters describe how the table is being looked at, which is what the address is for.
- **One dead filter removed.** `ConnectionsFilters.device` had a type and a predicate but no control ever set it, so the filtering it described could not be reached.
- **Re-renders were measured, not assumed.** Counted with a React `Profiler` around the DASD table, before and after:

  | Action | Before | After |
  | --- | --- | --- |
  | Typing 8 characters | 17 | 17 |
  | Sorting by a column | 4 | 4 |
  | Picking a filter | 5 | 7 |
  | The pause after typing | 1 | 3 |
  | Clearing every filter | 2 | 4 |

  Typing is the one that would have hurt, and it does not change: the field stays on local state while the keys come in. What costs two extra re-renders is writing to the address, because that goes through the router. Sorting shows the same price, 4 re-renders per click, and it is identical code before and after this PR, so the cost belongs to keeping state in the URL rather than to this hook. 

## Tests
- [x] `npm run test -- src`, `npm run eslint`, `npm run check-types`
- [ ] Filter each of the four tables, reload, and get the same rows
- [x] Paste the address in a second tab and get the same view
- [x] Type in a text filter and check the rows follow immediately while the address updates once, when typing stops
- [x] Press back after typing and leave the page, rather than undoing letters
- [x] Put a filter back to "All" and check its param leaves the address
- [x] Clear all filters and check the address is left clean, with any unrelated param untouched
- [x] Edit a param by hand to name something that does not exist and check the table opens unfiltered
- [x] Open the device selector over a filtered storage page and check the page behind is untouched

## Notes

Guideline updated at `doc/ui/guidelines/url-state.md`: how to choose between the hooks, the naming rule, the default-writes-nothing rule, and the one trap worth knowing, that a value being typed has to stay local until the address actually holds it.

Follow-up in this line: the software patterns search box, which is the same idea outside a table.

---


teclator (approved):
LGTM once the PR is up to date and ready for be merged

PR #3788, opened 2026-08-03
feat(web): keep the patterns search in the URL

dgdavid:
Last piece of the line started in #3785 (table sorting) and #3787 (table filters), which is this PR's base. Same idea, applied outside a table.

## Problem

The patterns page holds the longest list in the installer, which is precisely why it grew a search box. Filter it, reload, and the search is gone. Someone helping a user cannot say "open this link" and have them land on the same short list.

## Solution

The search now lives in the address, `/software/patterns?search=gnome`, using the hook the four tables already use. One filter, so the page only says it has one:

```tsx
const { filters, setFilter, resetFilters } = useFilterParams({ search: textFilter() });
```

Nothing about the page changes shape. The list, the results badge, the screen reader announcement and the sticky category headers all read the value the user is typing, exactly as before. Only the address lags, by the same short pause the tables use, so the history does not grow an entry per letter. Clearing the field empties both the field and the address, including a keystroke still waiting for that pause.

## Details

This is the first URL-backed filter on a page that also holds an unsaved form: the checkboxes carry a pattern selection that has not been submitted yet, and every search is now a navigation. It replaces the current entry and stays on the same route, so the page is never remounted and the selection is safe, but that is worth an assertion rather than an assumption. There is a test that checks a pattern, types a search, and checks the pattern is still checked.

Nothing on master prompts on navigation today, so there is nothing else to work around here. A guard that prompts on unsaved changes would have to compare paths before it blocks, or it would fire while somebody is typing in a search box. That belongs to the guard whenever it arrives.

## Tests

- [x] `npm run test -- src`, `npm run eslint`
- [x] Type in the search, reload, and get the same short list
- [x] Paste the address in a second tab and get the same view
- [x] Press back after typing and leave the page, rather than undoing letters
- [x] Scroll to the middle of the list, type, and check the position does not jump
- [x] Check a pattern, search for another one, uncheck it, accept, and check the selection saved is the one intended
- [x] Clear the search and check the full list and a clean address come back

---


teclator (approved):
LGTM once it is ready for being merged

PR #3790, opened 2026-08-03
HCN config (carry over  device independent options)

teclator:
## Problem

`parse-hcn` builds a kernel command line **from scratch** and hands only the arguments it
derives from `rd.hcn.*` to `nm-initrd-generator`:

```
bond=bond333e80f5:enP32775p1s0,env6:mode=active-backup,... ip=bond333e80f5:dhcp rd.route=...
```

Every other network option the user passed is therefore invisible to the generator.

The generator run NetworkManager performs on its own (`NetworkManager-config-initrd.service`)
is **not** a fallback for those options: the `ip=hcn` marker written by `hcn-cmdline.sh`
(bsc#1272445, jsc#PED-14534) is what stops that run from producing any connection, which is
exactly why the marker exists. So everything that run would have parsed into a connection is
discarded, and nothing else picks it up later.

The practical consequence: on an HCN boot, `nameserver=192.168.1.1` does **nothing**. The
same applies to `rd.peerdns=`, `rd.net.timeout.dhcp=`, `rd.net.dhcp.retry=`,
`rd.net.dhcp.vendor-class=` and `rd.net.dhcp.dscp=`.

While verifying the generator behaviour, two further pre-existing defects turned up. They
are unrelated to the carry-over and are kept in their own commit.

### `rd.net.timeout.carrier` silently reset

`nm-initrd-generator` **unconditionally** writes `15-carrier-timeout.conf` to its
`--run-config-dir`. `parse-hcn` left that at the default `/run/NetworkManager/conf.d`, so the
HCN run overwrote the file NetworkManager had already generated from the *real* command line,
resetting a user supplied `rd.net.timeout.carrier` back to the 10 s default.

### Bond lookup never matched in multi-bond setups

`fixup_nm_connections()` tested bond membership with:

```sh
strstr " $BOND_NAMES " " ${master:-$controller} "
```

`BOND_NAMES` is **newline** separated, but the needle is space delimited. With a single bond
the string is `" bond1 "` and the check works; from two bonds on it is `" bond1\nbond2 "`,
which contains neither `" bond1 "` nor `" bond2 "`. The `elif` fallback branch therefore
never matched at all as soon as there was more than one bond.

- **Bugzilla link:** [bsc#1272445](https://bugzilla.suse.com/show_bug.cgi?id=1272445)
- **Related Jira issues:** `jsc#PED-14533`, `jsc#PED-14534`

---

### 1. Carry the device independent network options over (commit `385557495`)

New `carry_over_cmdline()` in `parse-hcn.sh`, appended to `NEW_ARGS` just before the
generator is invoked. It copies verbatim the options that **name no device**, so they need
no rewriting and cannot create a stray connection:

| Parameter | Effect on the bond connection |
|-----------|-------------------------------|
| `nameserver=<ip>` (repeatable) | `ipv4.dns` / `ipv6.dns` |
| `rd.peerdns=0` | `ignore-auto-dns=true` |
| `rd.net.timeout.dhcp=<seconds>` | `dhcp-timeout` |
| `rd.net.dhcp.retry=<count>` | multiplies `dhcp-timeout` |
| `rd.net.dhcp.vendor-class=<id>` | `dhcp-vendor-class-identifier` |
| `rd.net.dhcp.dscp={CS0\|CS4\|CS6}` | `dhcp-dscp` |

**Deliberately left out**, with the reasoning recorded in the code and in both docs so it
does not have to be rediscovered:

- `rd.net.dhcp.client-id=`, `bootdev=`, `rd.ethtool=`, plus the `vlan=` / `bridge=` / `team=`
  stacked devices. Each names a device, and the user names a bond *port*, so the reference
  would have to be rewritten to its bond first. Worse, `nm-initrd-generator` creates a **full
  connection** for any device named in these (verified: `eth9`, `eth8`, `eth7` each produced
  their own `.nmconnection`), which this module would then persist to
  `/etc/NetworkManager/system-connections`. The stacked devices additionally need a way to be
  addressed from `rd.hcn.ip`, which the current syntax does not offer.
- `rd.net.dns`, `rd.net.dns-backend`, `rd.net.dns-resolve-mode`, `rd.net.timeout.carrier`.
  These produce a global configuration file rather than a connection, do not depend on the
  bonds having been resolved, and NetworkManager's own run already wrote them to
  `/run/NetworkManager/conf.d`.

An earlier draft of this work did implement the port-to-bond rewriting via a
`bond_for_device()` helper. It was dropped as disproportionate: ~30 lines of helper plus three
rewriting loops, serving three niche options, each of which risks persisting a stray
connection to the installed system.

### 2. Fix the config clobbering and the bond lookup (commit `4305363b2`)

- Add `HCN_RUNTIME_CONF_DIR=/run/hcn/conf.d` and pass `-r "$HCN_RUNTIME_CONF_DIR"` to the
  generator, next to the connections directory that is already isolated the same way. The HCN
  run can no longer clobber NetworkManager's `15-carrier-timeout.conf`.
- Add a flat, space separated `BOND_LIST` alongside `BOND_NAMES` and use it for the `strstr`
  substring check. `FIRST_BOND` then simplifies from `$(echo $BOND_NAMES | awk '{print $1}')`
  to `${BOND_LIST%% *}`, dropping a `shellcheck disable` with it.

---

## Documentation

- `README.md`: new "Standard network parameters" section with the carried-over table, an
  example, and an explicit "Not carried over" list.
- `README.md`: new "Host name" note clarifying that the `rd.hcn.ip` host name field reaches
  `/proc/sys/kernel/hostname` (parity with a normal `ip=` boot) but **not** `/etc/hostname` —
  that is the separate standalone `hostname=` parameter handled by `99agama-cmdline`. This
  distinction had caused confusion.
- `ARCHITECTURE.md`: new "Carrying over the rest of the network command line" section
  covering the same ground plus the `--run-config-dir` rationale.
- Both step lists ("How It Works", "Sequential Boot Process") updated for the new step.

---

## Testing

- **Unit/integration:** 65 bats tests green
  (`test_parse-hcn.bats`, `test_hcn-cmdline.bats`, `test_integration.bats`). Three new
  `carry_over_cmdline` tests cover the full set, a partial set (guarding against empty `opt=`
  arguments for unset options) and the empty case.
- **Lint:** `shellcheck -s sh -e SC3043,SC1091,SC2086,SC2018,SC2019,SC2154 parse-hcn.sh`
  clean — the same exclusion set the file already carried, nothing new introduced.
- **End-to-end against the real generator** (NetworkManager 1.56.1):

  ```bash
  mkdir -p /tmp/e2e/c /tmp/e2e/r
  /usr/libexec/nm-initrd-generator -c /tmp/e2e/c -r /tmp/e2e/r -- \
    bond=bond333e80f5:enP32775p1s0,env6:mode=1,miimon=100,fail_over_mac=2,primary=enP32775p1s0 \
    ip=bond333e80f5:dhcp \
    nameserver=192.168.1.1 rd.peerdns=0 rd.net.timeout.dhcp=60 rd.net.dhcp.retry=3
  ```

  `bond333e80f5.nmconnection` gets `dns=192.168.1.1;`, `ignore-auto-dns=true` and
  `dhcp-timeout=180` (60 x 3 retries), with exactly the three expected connection files
  (`bond333e80f5`, `enP32775p1s0`, `env6`) and no strays.
- **Multi-bond lookup bug reproduced directly**, confirming no bond matches from two bonds on
  with `BOND_NAMES` and all of them match with `BOND_LIST`.

### Not yet validated

Real **ppc64le hardware**. In particular the carrier-timeout fix rests on
`NetworkManager-config-initrd.service` running before `hcn-init-initrd.service`; that is read
from the unit ordering, not observed on a live boot.

### Note on the host name

An earlier iteration of this branch also created `/run/NetworkManager/initrd` before the
generator call, on the assumption that the generator does not create its `--initrd-data-dir`
and silently drops the host name. **That turned out to be wrong and was removed.** Against
NetworkManager 1.56.1 the generator creates the directory itself, recursively
(`-i /tmp/h/deep/i` produced `/tmp/h/deep/i/hostname`), and the ordering already holds:
`hcn-init-initrd.service` is `Before=dracut-initqueue.service` while `nm-run.sh` is an
`initqueue/settled` hook, so `parse-hcn` always runs first and the file is in place when the
hook reads it. The host name already worked; there was no bug. The docs now describe that
path as existing behaviour rather than as a fix.


live/live-root/usr/lib/dracut/modules.d/99hcn/parse-hcn.sh
```diff
@@ -287,6 +289,39 @@ EOF
...
+#
+# Prints the arguments to append to the generated command line, each one
+# preceded by a space.
+carry_over_cmdline() {
+  local opt val
+
+  for opt in nameserver rd.peerdns rd.net.timeout.dhcp rd.net.dhcp.retry \
+    rd.net.dhcp.vendor-class rd.net.dhcp.dscp; do
+    for val in $(getargs "$opt"); do
+      info "parse-hcn: keeping $opt=$val"
```
aafeijoo-suse, line 358: Be careful with this call to `info()` here. When `DRACUT_SYSTEMD` is set, `info()` logs to stdout, hence corrupting the output of `carry_over_cmdline()` and the content of `NEW_ARGS`.
This is currently working with your hcn module because `hcn-init-initrd.service` does not set `Environment=DRACUT_SYSTEMD=1`, like all dracut systemd services do. Also, if `dracut-state.sh` is sourced somewhere, this environment variable may be set. So, to be safe, I would remove this `info()`, or replace it with `warn()`, which prints to stderr... but this is not really a warning...

live/live-root/usr/lib/dracut/modules.d/99hcn/parse-hcn.sh
```diff
@@ -287,6 +289,39 @@ EOF
...
+# which this module would then persist to /etc.
+#
+# Prints the arguments to append to the generated command line, each one
+# preceded by a space.
+carry_over_cmdline() {
+  local opt val
+
+  for opt in nameserver rd.peerdns rd.net.timeout.dhcp rd.net.dhcp.retry \
+    rd.net.dhcp.vendor-class rd.net.dhcp.dscp; do
+    for val in $(getargs "$opt"); do
```
aafeijoo-suse, line 357: Just a warning: `getargs()` for options without values (i.e., booleans) returns an empty string, so these options would be lost. I see on the set above that all the options expect a value, `rd.peerdns` is usually called via `rd.peerdns=0`, so this is not a problem now. But keep that in mind if you introduce new kernel command line options here in the future.

live/live-root/usr/lib/dracut/modules.d/99hcn/parse-hcn.sh
```diff
@@ -252,7 +254,7 @@ EOF
     if [ -n "$mapping_info" ]; then
       found_master=${mapping_info% *}
       found_ifname=${mapping_info#* }
```
aafeijoo-suse, line 287: This is just an observation, unrelated to the current changes. I see `$found_master` a `$found_ifname` are not being reset in each `for con in "$conn_dir"/*.nmconnection; do` iteration. A connection without `$mapping_info`, or that does not meet the elif condition, inherits the previous iteration's values and is treated as "HCN-related".

live/live-root/usr/lib/dracut/modules.d/99hcn/hcn-init-initrd.service
```diff
@@ -10,13 +10,19 @@ ConditionKernelCommandLine=!rd.hcn=0
...
+Environment=DRACUT_SYSTEMD=1
 Type=oneshot
 ExecStart=/usr/bin/parse-hcn
 SyslogIdentifier=hcn-init-initrd
+StandardInput=null
 StandardOutput=journal+console
 StandardError=journal+console
+# parse-hcn waits up to 3 minutes for the HCN devices to show up in sysfs,
+# which is longer than the default start timeout
+TimeoutStartSec=300
```
aafeijoo-suse, line 25: Hmm... I think 5 minutes timeout for the whole run of parse-hcn brings a problem.
- `get_dev_hcn()` has a local loop that can wait up to 3 minutes **per device-tree node**. With every device, the counter starts again from zero...

https://github.com/agama-project/agama/blob/4fb6a75225669f5c429c05b5b00f1a634d0b27b0/live/live-root/usr/lib/dracut/modules.d/99hcn/parse-hcn.sh#L46-L47

- If we have 2 devices => this may take up to 6 minutes. But, with `TimeoutStartSec=300` (seconds) systemd will kill hcn-init-initrd.service at 5 minutes mid-run and the unit fails hard.

My suggestion: keep `TimeoutStartSec=300`, but instead of having a `local wait=12`, create a global counter available for all `get_dev_hcn()` calls, using a global variable of your choice for that (e.g. `HCN_WAIT_RETRIES` or so). This way, you will never exceed 180 seconds.

aafeijoo-suse (commented):
Unrelated to current changes, since I had to take a look at `get_dev_hcn()`, I realized that `devname=` should be reset in each iteration, `ofpathname` can assign a value to it but no `/sys/class/net/$devname` yet? See:

https://github.com/agama-project/agama/blob/4fb6a75225669f5c429c05b5b00f1a634d0b27b0/live/live-root/usr/lib/dracut/modules.d/99hcn/parse-hcn.sh#L56-L74

aafeijoo-suse (approved):

jreidinger (approved):

PR #3791, opened 2026-08-03
refactor(web): migrate the product selection form to the new forms stack

dgdavid:
## TL;DR

The product selection form now follows the conventions the migrated forms already share, and the screen it renders looks the way it is meant to again. Along the way, radios and checkboxes across the whole app sit properly aligned with their labels.

## Problem

Product selection is one of the forms still waiting to be migrated, so it does not follow the conventions agreed for forms. It keeps the chosen product, the chosen mode and the license acceptance in separate pieces of local state wired together by hand, inside the page that renders them, and it has no form directory of its own. Anyone moving between forms meets a different shape here than everywhere else, and the parts worth finding, such as the license acceptance or a single product option, are buried in a long page.

Reusing the shared submit action was not an option either: it could not take a richer label, nor extra disabled and loading conditions, so this screen built its own.

## Solution

Adopt the agreed forms conventions here, and lean on the pieces the stack already offers instead of local equivalents.

- The form holds the selection, the mode and the license acceptance, the same way the other migrated forms do, so the rules between those values sit in the form rather than in the page.
- The form gets its own directory, with the product option and the license acceptance as components of their own, matching the layout every migrated form uses.
- Submitting and cancelling now use the shared form buttons instead of the ones this screen built for itself.
- The button that opens the license dialog becomes a standalone component, since both the page and the form show it.
- The shared submit and cancel buttons gain what this screen needed to be able to reuse them: their own content, the remaining button properties, and extra loading or disabled conditions combined with the one already tracked. Any form with a richer action can now reach for them rather than rolling its own.

## Details

The selection rules are unchanged: a product with a license cannot be submitted until the license is accepted, switching products clears a previous acceptance, and the submit label still names what is about to happen.

**Product options.** Every step inside an option is now the same size, from the product name to its details and between the details themselves. A product with no mode left to offer no longer reserves room for details it has nothing to put in, and the modes of the selected product are set apart from its description, so the further choice reads as one. The option also carries less markup than it needs to say the same thing.

**Actions.** The submit and cancel buttons are back at the large size, and the submit label takes the room its own text needs, so a long product name wraps evenly instead of breaking into short lines on a narrow window.

## Changes reaching beyond this screen

Three of these are app-wide on purpose, since the underlying behavior was not specific to product selection:

- Radios and checkboxes keep their control aligned with the label at any label size. PatternFly places the control using an offset derived from the default label font size, so the alignment drifts as soon as a label renders bigger. Screens with large option labels, such as boot device selection, look different for this reason, and better.
- Large buttons no longer underline their label while hovered, focused or held down. An underline reads as a link on something that already looks like a button, and the color and background changes carry the state. Inline link buttons keep their underline.
- Products can tune the space between a radio label and the content below it through a new appearance role, rather than reaching for PatternFly tokens. Unset, everything keeps the spacing it has.

## Tests

- [x] Selecting a product without a license and confirming
- [x] Selecting a product with a license: submit stays disabled until the license is accepted
- [x] Switching between two products with licenses clears the acceptance
- [x] A product with modes: picking a mode, and the label naming the choice
- [x] Changing the already selected product, including cancelling out of it
- [ ] Single product available, with and without modes
- [x] Boot device selection and the other screens with radio options, for the alignment and spacing changes
- [x] Dark theme and high contrast, for the button label change

---


teclator (approved):
LGTM, and can be merged once the conflicts are solved

PR #3794, opened 2026-08-04
Fix AutoYaST support

imobachgs:
## Problem

While checking the new version of the SUSE documentation, I found a set of problems in the AutoYaST support:

- There are some omissions in the compatibility specification.
- Some properties are not handled properly.

## Solution

- Fix the import of some network attributes (`lladdr`, `wireless_auth_mode`, `wireless_mode`).
- Use `fcp_lun` instead of `lun` in zFCP.
- Fix and extend the compatibility specification.

## Testing

- Updated unit tests
- Tested manually


teclator (approved):
LGTM, although we already commented that the support for VLAN and bridge configuration is missing and pending, will try to add it tomorrow.

PR #3797, opened 2026-08-05
doc(web): improve styling guide and fix some rough edges

dgdavid:
## TL;DR

A product can restyle the whole installer from one stylesheet it ships itself.
This documents how, end to end, and fixes four things in Agama that only a
non-default palette makes visible.

Two documents, three screenshots, four small UI fixes. Reviewing the guide is
the bulk of it.

## Problem

Writing a product stylesheet meant reading Agama's code. The role names lived in
a SCSS comment, the selectors that scope a color to light or dark had to be
inferred, contrast was advice rather than something anyone could check, and the
limits were found by trying. Nothing said how the file reaches an ISO, or how to
use a typeface Agama does not bundle.

## Solution

- **`doc/product_theming.md`** now walks the whole job: the roles and where they
  live, a stylesheet small enough to copy, the steps in order, a contrast
  checklist with targets, the accessibility rules a palette can break, and what
  to do when no role covers what the brand needs. Three screenshots show how far
  this goes, from colors and a logo alone to a pixel typeface and square
  corners.
- **`doc/product_theming_packaging.md`** (new) covers the served paths and where
  their names come from, shipping a font, the RPM and the single line of KIWI
  that carry it, and the ways this fails silently.
- **`example-advanced.css`** gains a worked `@font-face` example.

The guide is written to be followed by hand. There is a prompt for drafting with
an AI, kept as one clearly optional section whose output is a starting point,
not a result.

## Vocabulary

Two words were doing three jobs, so the documents now say:

- **stylesheet**, the file a product writes and ships,
- **color scheme**, light or dark, which the appearance menu names and the
  person installing picks,
- and no **theme** and no **skin**, since a theme is expected to move the layout
  too and this feature never does.

## Four fixes in Agama itself

A default palette resolves every family and color to the same value, which hides
these. None is a product's problem to solve.

- **Plain headings did not take the heading family.** PatternFly assigns it
  through `Title`, so a heading written as a plain element kept the body family.
- **The breadcrumb trail rendered in two typefaces**, since only its last item
  is a heading. It now uses one family throughout and sizes itself against the
  viewport, giving way before it wraps.
- **A summary icon was stranded on its own line** when a long title wrapped, and
  the issue icon came from a different component at a different size. Both now
  come from the same component, and the row no longer wraps.
- **The header logo dropped below the trail** when the trail wrapped. The row no
  longer wraps and the text column may shrink below its longest line.

Two carry a judgement worth a second opinion. The summary row is top-aligned
now, so the icon tracks the first line of a wrapped title; centered may still be
right when nothing wraps. And the trail's `clamp(size--200, 1.6vw, size--400)`
has Agama's own body size as its floor, so it never renders smaller than today,
though its leading items shrink about 10% on a narrow header. The viewport is
the measure rather than the masthead because PatternFly lays the masthead out as
a subgrid from the xl breakpoint on, and a container query would break that
layout at exactly 1200px.

## Follow-up, not in this PR

The stylesheet is named after the product `id:` and the logo after its `icon:`,
and the two diverge in practice: `openSUSE_Leap` ships `Leap16.svg`, so neither
name is derivable from the other. Looking the logo up by `id:` as well would
drop that field from the API and from every product definition, at the cost of
products that share a logo shipping their own copy. That is a product definition
API change, so it belongs in its own discussion.

## Testing

- The documents were exercised by writing three complete stylesheets against
  them, measuring every checklist row with `check-contrast.js` in light and
  dark, and looking at the result in the running installer. Those three are the
  screenshots in the guide; none of them ships.
- `Summary` had no tests and has them now, covering the icon it picks per state
  and the loading placeholders.
- `npm run test`, `npm run check-types` and `npm run eslint` pass.
  `example-advanced.css` passes stylelint.
- Worth a look in the running installer, since none of the code changes shows in
  a default palette: the overview with a title long enough to wrap, a two-line
  and a three-line trail, and the header at 1024x768, at 1200px and at 200%
  zoom.


dgdavid:
Converted to a draft mode since it lacks of some bits initially shaped as a material for the blog that will be landing here instead.

dgdavid:
Re-ready for review, althoug PR description wording might be adapted a bit. 

It now affect to the code as well for fixing things found when updating the documentation by creating two skin for fake products.

The part needs more attention is the packaging stuff. Please, help to shape it.

doc/product_theming.md
```diff
@@ -4,70 +4,322 @@ The Agama web UI can be restyled for a product without touching application
 code: colors, logo, and a handful of other roles are set in a single
 per-product file, loaded at runtime after Agama's own styles.
 
+This document is about designing and writing that file.
+[product_theming_packaging.md](product_theming_packaging.md) covers getting it
+onto a medium, which a product can do from its own package without contributing
+anything to Agama.
```
ancorgs: I guess this means "with no need to change Agama source code or style files". I mean, is really not about contributing or not.

doc/product_theming.md
```diff
@@ -4,70 +4,322 @@ The Agama web UI can be restyled for a product without touching application
...
+[product_theming_examples.md](product_theming_examples.md) walks through two
+skins end to end, from the prompt each started with to the problems each ran
```
ancorgs: I would clarify what "from the prompt each started with" means. I know the documentation kind of propose to use a IA (claude) prompt to bootstrap the creation process. But at this point in the document it is really not that clear what that "prompt" means.

doc/product_theming.md
dgdavid: Sure. Good catch.

doc/product_theming.md
dgdavid: And now is even outdated since there are three examples. Will fix it later.

doc/product_theming.md
```diff
@@ -1,124 +1,573 @@
...
+brand is the [Fedora Project's](https://docs.fedoraproject.org/en-US/project/brand/),
+borrowed as a friendly nod to a neighbour: Agama's interface is built on
+PatternFly, the open source design system Red Hat publishes, and Red Hat
+sponsors Fedora. No Fedora logo, artwork or name appears in the result, and
+Fedora is a trademark of Red Hat, Inc.
+
+| Light | Dark |
+| --- | --- |
+| ![The installation overview dressed as Blue Sombrero, light: blue links and icons on white, Agama's own typeface, corners and sizes](images/theming/blue-sombrero-light.png) | ![The same screen in dark: the same blue hue as pale ink on a night-blue page](images/theming/blue-sombrero-dark.png) |
+
```
dgdavid, line 79:
```suggestion

> [!NOTE]
> **Why not just use the brand's darkest blue for the dark page?**
>
> On its own it would work as a page background: pale text on it reads
> comfortably. The trouble comes one step up. In dark mode, cards and dialogs
> sit _lighter_ than the page, so a page that starts at the brand's darkest blue
> pushes them into a mid blue, and there the secondary text and the brand color
> itself stop having enough contrast to be read.
>
> A dark page has to leave room above itself. Taking that same blue lower gives
> it that room, and everything stacked on top still reads.

```

ancorgs (approved):
LGTM

PR #3798, opened 2026-08-05
Fix and extend the AutoYaST compatibility reference for scripts

imobachgs:
## Problem

The AutoYaST compatibility reference omits the `chrooted` element apart from other small problems
in the scripts mapping.

## Solution

Extend the reference and fix the wrong mapping.


dgdavid (approved):
Approving it assuming no changes entry is needed here.

dgdavid (approved):
Re-approved

PR #3795, opened 2026-08-04
feat: extend storage search

joseivanlopez:
Merge feature branch extending the storage search options. Includes the following PRs:

* https://github.com/agama-project/agama/pull/3727
* https://github.com/agama-project/agama/pull/3747
* https://github.com/agama-project/agama/pull/3767
* https://github.com/agama-project/agama/pull/3772

Related to https://bugzilla.suse.com/show_bug.cgi?id=1273493.

The associated documentation is added by https://github.com/agama-project/agama-project.github.io/pull/180.

PR #3800, opened 2026-08-05
fix: adapt storage system

joseivanlopez:
## Problem

The serialized information about devices exported for *system* and *proposal* requires some improvements:

* Property `driver` is in singular, but it refers to a list of drivers.
* Property `dellBoss` should be renamed to `boss`, following the same syntax as the searches in the config document, see https://github.com/agama-project/agama/pull/3772.
* Property `active` from `block` section is unused. Note that the documents should only contain information actually used by clients or usable for completing a config.
* Property `id` is missing in the `partition` section. This property is useful because it is used as a search condition, see https://github.com/agama-project/agama/pull/3772.

## Solution 
Adapt the exported information of the storage *system* and *proposal*:

* Rename `driver` to `drivers`.
* Rename `dellBoss` to `boss`.
* Remove unused property `active` from `block` section.
* Add `id` to `partition` section.


mvidner:
It is obvious to you as the author, but as a reviewer of an "easy" PR, I had to search for the context:
https://github.com/agama-project/agama/pull/3795

If you had used the PR template, you could have filled in "**Problem:** after merging #3795, Agama broke"

rust/package/agama.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Wed Aug  5 15:08:46 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Fix storage system and proposal schemas
+  (gh#agama-project/agama#3800):
+  - Rename "drive" as "drives".
```
mvidner:
```suggestion
  - Rename "driver" to "drivers".
```

rust/package/agama.changes
```diff
@@ -1,3 +1,13 @@
...
+  - Rename "dellBoss" as "boss".
```
mvidner:
```suggestion
  - Rename "dellBoss" to "boss".
```

joseivanlopez:
> It is obvious to you as the author, but as a reviewer of an "easy" PR, I had to search for the context: #3795
> 
> If you had used the PR template, you could have filled in "**Problem:** after merging #3795, Agama broke"

You are totally right. Sorry for the mingy description. I have improved it.

rust/package/agama.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Wed Aug  5 15:08:46 UTC 2026 - José Iván López González <jlopez@suse.com>
+
+- Fix storage system and proposal schemas
+  (gh#agama-project/agama#3800):
+  - Rename "drive" to "drives".
```
mvidner: You have fixed the preposition but kept the confusion about drives vs drivers, please fix as appropriate :)

mvidner (changes requested):
Thanks for the improved description, but one conceptual mistake remains

mvidner (approved):
Thanks, LGTM (but I haven't exercised the code... the included tests seem enough)

PR #3807, opened 2026-08-07
remove volume id and ppc64 hacks

wfeldt:
## Problem

There's a number of no longer needed hacks in the Agama obs project files.

## Solution

- install media volume id can be set explicitly in the kiwi file; remove volid related xorriso hacks
- ppc64 is supported natively in kiwi; remove ppc64 workaround from boot config scripts

## Note

With this patch the volume id is hard coded and contains the product name (openSUSE). That's maybe not wanted for SLES. The way forward would be either to drop any product reference there or have a more elaborate kiwi config file where the elements are replicated for each product (via `<preferences>`).

## Testing

- manually tested

## Todo

1. s390x not yet ready
2. the xorriso wrapper script has been hijacked to do complete unrelated things; these code blobs should move elsewhere so we can get rid of the xorriso wrapper


teclator (commented):
Just miss the changelog entry referencing the PR at least

imobachgs (approved):

wfeldt:
Separated out the ppc64 part, as volid handling is still undecided.

- https://github.com/agama-project/agama/pull/3826

PR #3816, opened 2026-08-11
refactor(web):  remove the dead core/EmptyState and stale consumers

dgdavid:
The `core/EmptyState.tsx` wrapper had two consumers, but neither was reachable from any route or component, so nothing in the running application ever rendered it.

It had also stopped doing anything beyond PatternFly's own component, thus t is not worth keeping neither, the wrapper nor the unused components. 

> [!NOTE]
>
> No visual impact, just more clean-up. It does not deserve entry in the changes file.


joseivanlopez (approved):
LGTM

PR #3817, opened 2026-08-12
Do not run software SetConfig in background

imobachgs:
## Problem

When running the `agama config load` command, it may happen that the monitor exits before the software service starts to refresh the repositories.

## Solution

After the introduction of the `TaskManager`, the call to the software `SetConfig` already happens in the background. Spawning a task in the the in the software service causes that the software task (in the manager) exits while the repositories are still not read.

## Testing

- Tested manually
- A testing ISO is [available here](https://build.opensuse.org/projects/systemsmanagement:Agama:branches:wait-for-software-set-config/packages/agama-installer:openSUSE/repositories/images/binaries).

## Follow-up

- Work on the web UI to reduce the amount of time the user is waiting after selecting a product.

### DIFF
--- a/rust/agama-software/src/service.rs
+++ b/rust/agama-software/src/service.rs
@@ -285,18 +285,10 @@ impl Service {
             return Ok(());
         };
 
-        let model = self.model.clone();
-        let progress = self.progress.clone();
-        let issues = self.issues.clone();
-        let l10n = self.l10n.clone();
-        let state = self.state.clone();
-        let events = self.events.clone();
         tracing::info!("Synchronizing the wanted software state");
-
-        tokio::task::spawn(async move {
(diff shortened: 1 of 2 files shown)


joseivanlopez (commented):
Is the software service blocked while updating the proposal? If so, is there any risk of deadlock?

imobachgs:
> Is the software service blocked while updating the proposal? If so, is there any risk of deadlock?

Yes, it is blocked. Actually, it was already blocked anyway, because the `zypp` operations run on a single thread. The difference is that, for a short period of time, the progress had not started and nobody noticed.

I ran the testing image and it seems to work just fine.

joseivanlopez (approved):
LGTM

dgdavid (approved):

PR #3822, opened 2026-08-14
feat(web): allow  to browse network devices with details when binding a connection

dgdavid:
## Problem

When a connection is bound to a device, the form offers a dropdown listing device names with their MAC address underneath. That is enough to pick between two devices. It is not enough to tell which of six is the right one: nothing there says which devices are connected, which already carry an address, or which are Wi-Fi rather than Ethernet.

## Solution

The dropdown keeps working exactly as before for a quick pick, and gains a "Browse with details..." entry at the end. It opens a dialog listing the same devices in a table, with the rest of what is known about each: type, IP addresses and state, next to the name and MAC address.

The table can be sorted by column, and starts in the same order as the dropdown the user came from. Nothing is committed until the dialog is confirmed, so browsing costs nothing: cancelling or pressing Escape leaves the form untouched.

The dialog follows the storage device selector: the same table with radio rows and sortable headers, a confirmation button that names what it will do ("Use enp1s0") instead of a bare "Confirm", and the same hint beside it while nothing is picked. Someone who has chosen a disk during installation meets a dialog that behaves the way they already learned.

The entry is a general capability of the shared dropdown field rather than something specific to this form, so other selectors can offer more than their listed values without rebuilding it. It sits inside the list on purpose: a second control beside the field would do one job with two elements and put a tab stop in everyone's path for help that is not always needed. That reasoning is written down as a new pattern in the form conventions, so the next selector facing the choice does not have to work it out again.

Network device states are now translatable, which they were not before.

## Accessibility

- The browse entry is a real option in the listbox, not a button smuggled into it, so it is announced and reached exactly like the values around it.
- ARIA does not allow an option to declare that it opens a dialog. The visible label ends in an ellipsis, and screen reader users additionally hear "Opens a dialog", a sentence the dropdown field supplies itself so every selector says it the same way.
- Focus opens on the row of the device already in use, so it is what a screen reader announces first and where the arrow keys start. It returns to the dropdown toggle when the dialog closes.
- The column currently sorting the table reports its direction.

One known limitation is not addressed here: PatternFly names the row selection radios "Select row 1", "Select row 2" and so on, with no way for a caller to override it, so a screen reader user arrowing through the radio group hears positions instead of device names. It is tracked upstream in https://github.com/patternfly/patternfly-react/issues/11912, and every table in Agama that selects a row shares it. Working around it locally would mean reimplementing PatternFly's selection cell, so this waits for the fix.

## Screenshots

<details>
<summary>Click to show hide real and mockups screenshots</summary>

---

#### Before

localhost_8080_ - 2026-08-14T080729 315

#### After

localhost_8080_ - 2026-08-14T160207 656

localhost_8080_ - 2026-08-14T160214 820

#### When more data is available (mock-ups)

localhost_8080_ - 2026-08-14T171933 490
localhost_8080_ - 2026-08-14T171943 002

</details>

## Follow-up

Two ports of the same model, both disconnected, still look alike here. Telling them apart needs link speed, the socket the port sits in, or the hardware model, and none of those exists in the API yet. Same for asking a device to blink its port light.

Storage names its confirmation button by what the pick does to the current choice ("Keep sda", "Change to sdb"). This dialog only says "Use enp1s0": the selector preselects a device, so "Keep" would greet someone who has chosen nothing, and editing a connection cannot change its device anyway. Worth converging once it can.

A filter over the table was built and then removed, to keep this first iteration the same shape as the storage dialog. Reverting the last commit brings it back if these lists prove long in practice.

Letting people choose which columns to see would suit a dialog whose whole point is detail, but the choice needs somewhere to live, and the installer has nowhere to keep it yet beyond the appearance menu holding its own setting. So the column chooser waits, along with anything else worth remembering per dialog, per table or per page.

## Testing

- Unit tests cover the new dropdown entry, the dialog (rendering, ordering, picking, confirming and cancelling) and the wiring between them, including that confirming a device keeps the interface and MAC fields in sync.
- `npm run test -- src`, `npm run check-types` and `npm run eslint` pass.
- Worth a manual keyboard pass in the running installer: open the device dropdown on an Ethernet or Wi-Fi connection with the binding set to name or MAC address, arrow down to the browse entry, and check the trip through the dialog and back.


teclator (approved):
LGTM!, great work, thanks!

dgdavid:
> LGTM!, great work, thanks!

Thanks a ton!

Before merging, I'm interested in your view about the mentioned follow-ups too. Would be possible to extend the offered information at some point sooner? Asking because the selector makes a lot of sense with more data, although now it can help since it shows a bit more, like IP address.

PR #3826, opened 2026-08-18
remove ppc64 hacks from boot config scripts

wfeldt:
## Problem

There are some ppc64 related hacks in kiwi boot config scripts dating back to the time where ppc64 was not supported natively in kiwi.

## Solution

- ppc64 is supported natively in kiwi; remove ppc64 workarounds from boot config scripts

## Testing

- manually tested

## Todo

- s390x not yet ready


teclator (approved):
LGTM

PR #3827, opened 2026-08-20
Own kernel

jreidinger:
## Problem

Agama lacks ability for products to define its own kernel flavor if needed.
- https://jira.suse.com/browse/PED-16833
- https://trello.com/c/cnvDHtLu/5968-feature-agama-products-should-be-allowed-to-choose-the-kernel-flavour

## Solution

Implemented it with new key "kernel" in software section. Note: as it is in `products.d` we can change it in future as we do not keep backward compatibility for products.yaml

## Testing

- *Added a new unit test*


mchf (approved):
lgtm

PR #3834, opened 2026-08-25
feat(web): dress the whole medium, not just one product at a time

dgdavid:
## Problem

[Get Agama dressed for success](https://agama-project.github.io/blog/2026/08/11/dressed-for-success) introduced product skins: drop a CSS file named after your product id, set a handful of `--agm-t--*` roles, and Agama wears your colors, your logo, your typography. 

That works beautifully as long as "the thing being branded" is a product. But a medium is not always one product, and it is never *only* products:

- Multi-product media (the openSUSE ISO, or any vendor ISO offering several flavors) would need the same CSS copied once per product id, and kept in sync forever.
- The **product selection and login screens**  appear *before* any product is picked, so no product skin is loaded yet. The first screen users see was the one screen that stayed stock. Not the best first impression for a branded medium.
- Some branding is not about the product at all. It is the vendor, the OEM, the hospital, the lab: one visual identity across everything on the medium.

So this PR takes the wardrobe up one size.

### 1. Brand skins

A second, broader scope: a single `brand.css` installed at `assets/appearance/brand.css` on the medium. Agama does not ship it; a branding package does. If it is not there, nothing changes at all.

It is linked statically from `index.html`, which buys two things:

- It is in effect **at first paint**, including on the product selection screen.
- It is loaded **before** the runtime product link, which is appended to the `<head>` later. So when both exist, the product skin still wins, role by role: a product can override just its accent color and inherit the rest of the brand.

The resulting cascade, in plain terms:

```
product skin  >  brand skin  >  Agama defaults
```

Nobody has to choose. A vendor sets the house style once; a product refines it where it matters.

### 2. No product selected means no product skin

Agama remembers the last product's appearance so returning users do not get a flash of stock colors before the skin loads. That memory now gets cleared when there is no active product, so whatever is underneath (the brand skin, or Agama's own defaults) applies again instead of a stale product's palette lingering.

### 3. Corner radius, now per shape size

Skins already had `--agm-t--border-radius`, one knob for the whole UI. Real brands rarely work that way: they define a *shape scale*, usually a large radius for big containers, a medium one for cards, a small one for buttons and inputs. PatternFly happens to split radius into exactly three tiers under the hood, so the role set now exposes them:

| Role | Reaches |
| --- | --- |
| `--agm-t--border-radius` | **master**: all three tiers at once (unchanged) |
| `--agm-t--border-radius--small` | buttons, inputs, selects, tables, tooltips, menus, pills |
| `--agm-t--border-radius--medium` | cards, popovers, expandable sections |
| `--agm-t--border-radius--large` | modal dialogs, login card (plus switch/slider thumbs and calendar dates, which PatternFly puts in the same tier) |
| `--agm-t--button--border-radius` | buttons only, plus the toggles that read as buttons (split-button children, primary and secondary menu toggles) |

The last one is the narrow end of a chain that already existed: `--agm-t--control--border-radius` covers buttons, text inputs and selects together, which is all most brands need, and the button role splits buttons back out for the ones that round their buttons more than their fields. The login screen below is where that split shows most plainly, since it puts a button and a password field side by side.

**Fully backwards compatible.** Each tier falls back to the master, which falls back to PatternFly's stock value; the button role falls back to the control role, and that to the small tier. A skin that only sets `--agm-t--border-radius` behaves exactly as it does today; a skin that sets nothing looks exactly as it does today. Already-deployed skins keep working untouched, including the split-button and menu-toggle shapes, which keep following the pill token rather than being quietly rerouted through the control role.

### 4. Documentation

`doc/product_theming.md` and `doc/product_theming_packaging.md` become `doc/theming.md` and `doc/theming_packaging.md`, since they now cover both scopes rather than products only. The design document gains the two scopes and how they compose, why the selectors must stay at zero specificity, and a section on writing a brand appearance; the packaging one gains where `brand.css` goes, how to ship it (including the virtual provide that keeps two branding packages off the same path), and how to drop it onto a running installer. The AI prompt at the end picks the scope in one placeholder, with the logo paragraph split out so a brand appearance can skip it.

`example-advanced.css` gains a worked example of the three radius tiers, and the role index at the top of `_semantic.scss` documents them.

## Testing

One brand proves a skin is possible. Two, picked to break in different places, show the role set is not shaped around a single vendor. Both are built from openly licensed design tokens, measured with `web/scripts/check-contrast.js` against the surfaces each color actually lands on, in each theme separately, with any value that falls short replaced by another from the same system's own ramp.

**[Material Design 3](https://m3.material.io/)** is the interesting case for shape. Material assigns a different corner radius per component, so the three radius tiers land exactly on its scale: 4dp for inputs and menus, 12dp for cards, 28dp for dialogs. It also defines only one status color, `error`, so the other four are generated the way Material does it, one tonal palette each read at the same tone. That is why every status in it lands within a tenth of a point of the others: same tone, same contrast, by construction.

<details>

---

localhost_8080_ - 2026-08-25T130329 611
localhost_8080_ - 2026-08-25T130350 078
localhost_8080_ - 2026-08-25T130355 776
localhost_8080_ - 2026-08-25T130421 164

localhost_8080_ - 2026-08-25T130307 763

localhost_8080_ - 2026-08-25T130408 319
localhost_8080_ - 2026-08-25T130403 948
localhost_8080_ - 2026-08-25T130415 506

</details>

**[Acorn](https://acorn.firefox.com/latest/home/acorn-aRSAh0Sp)**, Firefox's design system, stresses a different corner. Its palette is written in OKLCH rather than hex, its accent changes hue between schemes (blue in light, cyan in dark), and several of its tokens are `color-mix()` recipes rather than fixed colors. All of it expressible: converted, composited, and set through the same roles as everything else.

<details>

---

localhost_8080_ - 2026-08-25T132406 869
localhost_8080_ - 2026-08-25T132436 546
localhost_8080_ - 2026-08-25T132440 071
localhost_8080_ - 2026-08-25T132446 438
localhost_8080_ - 2026-08-25T132418 687
localhost_8080_ - 2026-08-25T132424 667
localhost_8080_ - 2026-08-25T132429 836
localhost_8080_ - 2026-08-25T132452 396
</details>

Both turned up the same weak spots, which is the useful result. Each specifies a caution yellow, a subtle border, and a disabled text color that fall short of WCAG AA as published, because each leans on an exemption the theming guide deliberately does not take. And in each, the first draft put a status color directly on top of the link color, the exact collision the accessibility notes warn about.

Material also found a real gap, and it is fixed here rather than written down for later. Its buttons are fully rounded while its text fields are not, and until now both were driven by the same role, so a skin could have one or the other. There is a narrower role for that now, and Material's shape scale goes in whole.

Neither skin ships with Agama. They exist here as the exercises that validated the role set, built from public documentation and using no logos or wordmarks.

## Trying it

Drop a `brand.css` into the served assets and reload:

```
scp brand.css root@<installer>:/usr/share/agama/web_ui/assets/appearance/brand.css
```

Then reload the browser. No rebuild, no Agama release. As the blog post put it: your branding stays in your repository and ships on your schedule.

## Checklist for reviewers

- [ ] Brand skin applies on the product selection screen (before any product is chosen).
- [ ] With both files present, the product skin wins where it sets a role and inherits the brand elsewhere.
- [ ] Going back to product selection drops the previous product's colors.
- [ ] An existing skin that sets only `--agm-t--border-radius` renders identically to before.
- [ ] Both light and dark, plus high contrast.

---

### DIFF
--- a/doc/theming.md
+++ b/doc/theming.md
@@ -1,33 +1,39 @@
-# Creating a product appearance
+# Creating a product or brand appearance
 
-The Agama web UI can be restyled for a product without touching application
-code: colors, logo, and a handful of other roles are set in a single per-product
-file, loaded at runtime after Agama's own styles.
+The Agama web UI can be restyled without touching application code: colors,
+logo, and a handful of other roles are set in a single CSS file, loaded after
+Agama's own styles. The same roles and the same file format restyle a single
+product or a whole medium; see [The two kinds of
+appearance](#the-two-kinds-of-appearance).
 
-Two things meet in that file, and this document keeps them apart. The **product
-appearance** is what a product writes: the palette, the logo and a few related
-roles that make the installer read as that product. The **color scheme** is
-light or dark, and it belongs to whoever is installing, who picks it in the
-appearance menu or inherits it from their system, alongside a separate contrast
-setting. A product writes the stylesheet and never picks the scheme, so that
-stylesheet has to say how its colors behave in both.
+Two things meet in that file, and this document keeps them apart. The
+**appearance** is what a product or a brand writes: the palette, the logo and a
+few related roles that make the installer read as that product or that vendor.
+The **color scheme** is light or dark, and it belongs to whoever is installing,
+who picks it in the appearance menu or inherits it from their system, alongside
+a separate contrast setting. The stylesheet never picks the scheme, so it has to
+say how its colors behave in both.
 
 This document is about designing and writing that file.
-[product_theming_packaging.md](product_theming_packaging.md) covers getting it
-onto a medium, which a product ships from its own package: no change to Agama's
-source or style files is involved.
+[theming_packaging.md](theming_packaging.md) covers getting it onto a medium,
+which both kinds ship from their own package: no change to Agama's source or
+style files is involved.
 
 **The short version.** Copy `example.css`, set only the roles the brand needs,
 add a light and a dark logo. Measure every contrast pair the checklist lists, in
 both color schemes, with the script rather than the eye. Then look at the result
 in a running installer, because half of what goes wrong has no number attached.
 If the brand needs something no role covers, propose the role instead of
-reaching past it: that contract is the whole reason a product stylesheet
-survives an Agama upgrade.
+reaching past it: that contract is the whole reason either kind of stylesheet
+survives an Agama upgrade. A brand appearance skips the logo and is otherwise
+written exactly the same way.
 
+- [The two kinds of appearance](#the-two-kinds-of-appearance), for one product
+  or for a whole medium, and how they combine
 - [What a product can change](#what-a-product-can-change) and [what it
   cannot](#what-a-product-cannot-change)
 - [The roles](#the-roles), and [the steps](#steps) in order
+- [Brand appearances](#brand-appearances), the one that covers a whole medium
 - [Contrast checklist](#contrast-checklist), plus [brand colors that cannot be a
   fill](#when-the-brand-color-cannot-be-a-fill) and [elevation in dark
   mode](#elevation-in-dark-mode-and-why-tooltips-differ)
@@ -38,6 +44,62 @@ survives an Agama upgrade.
 - [Optional: starting from an AI draft](#optional-starting-from-an-ai-draft),
   and why [the second round is the real
   one](#the-second-round-is-the-real-one)
+
+## The two kinds of appearance
+
+Both kinds set the same roles, in the same file format. What differs is how far
+they reach:
+
+- A **product appearance** restyles Agama for one product, and applies once that
+  product is selected. It is named after the product id, and it is the only one
+  that can change the logo.
+- A **brand appearance** restyles Agama for a whole medium: whatever product is
+  selected, and before any product is selected at all. It is a single
+  `brand.css`, and it covers two cases a product appearance cannot. A vendor, an
+  OEM or an institution whose medium offers several products would otherwise
+  copy the same file once per product id and keep the copies in sync forever.
+  And the product selection screen, the first screen anyone sees, comes before
+  any product is picked, so no product appearance is loaded there yet.
+
+They are not alternatives: a medium can carry both, and they combine.
+
+### Precedence
+
+Three layers, resolved **role by role**, not file by file:
+
+```
+product appearance  >  brand appearance  >  Agama default
+```
+
+Every role carries its own default in `_semantic.scss`, so a role nobody sets
+keeps Agama's value. A role the brand sets and the product does not is the
+brand's. A role the product sets is the product's, whatever the brand said.
+
+The consequence worth planning for: a product appearance that sets only its
+accent color inherits the brand's surfaces, borders and radii, not Agama's. That
+is usually what a medium wants. When it is not, the product file has to set
+those roles too. A partial file does not replace the file below it; only the
+roles it sets do.
+
+The same resolution happens independently in each color scheme, so a product may
+override the dark palette and inherit the brand's light one.
+
+### Keep the selectors at zero specificity
+
+Precedence works because both files declare their roles at the same specificity,
+which leaves the load order to decide. The examples get this right by wrapping
+the scheme condition in `:where()`:
+
+```css
+:root:where(.pf-v6-theme-dark) {
+  /* ... */
+}
+```
+
+`:where()` contributes no specificity, so the block competes on equal terms with
+a plain `:root`. Writing `:root.pf-v6-theme-dark` instead raises the specificity
+and lets a brand appearance beat a product one, which breaks the order above.
+Copy the selectors from the examples rather than writing your own.
 
 ## What a product can change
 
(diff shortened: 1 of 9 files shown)


doc/theming.md
```diff
@@ -1,33 +1,39 @@
-# Creating a product appearance
```
dgdavid, line 1: NOTE: document has been renamed from `product_theming.md` to just `theming.md` If PR get merged, worth updating  links in https://github.com/agama-project/agama-project.github.io/blob/main/blog/2026-08-11-dressed-for-success.mdx

doc/theming_packaging.md
```diff
@@ -1,20 +1,21 @@
-# Shipping a product appearance on a medium
+# Shipping an appearance on a medium
```
dgdavid, line 1: NOTE: document has been renamed from `product_theming_packaging.md` to just `theming_packaging.md` If PR get merged, worth updating  links in https://github.com/agama-project/agama-project.github.io/blob/main/blog/2026-08-11-dressed-for-success.mdx

doc/theming.md
lslezak: What about creating a compatibility symlink until the blog post is updated?

dgdavid: Is that possible?

dgdavid: Ah, sure. Forget the question, I was thinking in a completely different think (HTTP redirection).

I do not think it worth it. But as you, the reviewers, wants.

lslezak: Ah, you can add a symlink into Git, but GitHub does not redirect to the target so it does not work in the  GitHub web UI as expected.

doc/theming_packaging.md
```diff
@@ -127,9 +160,9 @@ that side of the line.
   load and the UI keeps the Agama defaults, which is what makes this safe to
   add late in a medium's build.
 - **A stale precompressed sibling wins.** The server prefers a `.gz` next to a
-  file when the browser accepts gzip, so a leftover `<product-id>.css.gz` from an
-  earlier build keeps being served after the plain file is updated. Ship both or
-  neither.
+  file when the browser accepts gzip, so a leftover `<product-id>.css.gz` (or
+  `brand.css.gz`) from an earlier build keeps being served after the plain file
+  is updated. Ship both or neither.
```
lslezak:
```suggestion
  is updated. Ship both or none.
```

doc/theming_packaging.md
```diff
@@ -95,6 +112,22 @@ install -Dm644 MyProduct.svg %{buildroot}%{_datadir}/agama/web_ui/assets/logos/M
 install -Dm644 MyProduct-dark.svg %{buildroot}%{_datadir}/agama/web_ui/assets/logos/MyProduct-dark.svg
 ```
 
+A branding package ships the one file instead:
+
+```
```
lslezak: Use the `rpm-spec` language specification for syntax highlighting.
```suggestion
```rpm-spec
```

doc/theming_packaging.md
```diff
@@ -95,6 +112,22 @@ install -Dm644 MyProduct.svg %{buildroot}%{_datadir}/agama/web_ui/assets/logos/M
...
+%install
+install -Dm644 brand.css %{buildroot}%{_datadir}/agama/web_ui/assets/appearance/brand.css
+```
+
+Since a medium has room for exactly one brand appearance, branding packages
+should declare a virtual provide and conflict on it, so two of them fail at
+install time instead of racing for the same path:
+
+```
```
lslezak:
```suggestion
```rpm-spec
```

doc/theming.md
```diff
@@ -39,6 +45,62 @@ survives an Agama upgrade.
...
+  OEM or an institution whose medium offers several products would otherwise
+  copy the same file once per product id and keep the copies in sync forever.
+  And the product selection screen, the first screen anyone sees, comes before
+  any product is picked, so no product appearance is loaded there yet.
+
+They are not alternatives: a medium can carry both, and they combine.
+
+### Precedence
+
+Three layers, resolved **role by role**, not file by file:
```
lslezak, line 68: You can include [mermaid](https://mermaid.js.org/) charts in the GitHub markdown:

```mermaid
flowchart LR
  b1(Product appearance) --> b2(Brand appearance) --> b3(Agama default)
```

(Use "Edit" or Copy Markdown" in the three-dots menu to see the source text.

web/src/assets/products/example-advanced.css
```diff
@@ -76,16 +76,31 @@
...
+        more than its text fields; a fully rounded button is 999px.
+
+     Anything left unset falls through to the layer above it and, in the end,
+     to PatternFly's stock radius. Below: 6px everywhere, except roomier modal
+     dialogs, squarer text fields and fully rounded buttons. */
   --agm-t--border-radius: 6px;
-  --agm-t--control--border-radius: 6px;
+  --agm-t--border-radius--large: 12px;
+  --agm-t--control--border-radius: 2px;
+  --agm-t--button--border-radius: 999px;
```
lslezak, line 103: Why 999px and not 50%?

lslezak (commented):
Just some minor suggestions.

doc/theming.md
```diff
@@ -39,6 +45,62 @@ survives an Agama upgrade.
...
+
+```
+product appearance  >  brand appearance  >  Agama default
+```
```
dgdavid:
```suggestion

```mermaid
flowchart LR
  b1(Product appearance) --> b2(Brand appearance) --> b3(Agama default)
```
```

web/src/assets/products/example-advanced.css
dgdavid: That's based on Material Design.  In any case, it is an example, let's keep it.

lslezak (approved):
LGTM

PR #3837, opened 2026-08-26
Stop blocking `agama config generate` on AutoYaST conversion questions

imobachgs:
## Problem

`agama config generate` can use an AutoYaST profile as its source. When the
conversion finds unsupported elements, a question used to be raised through
the HTTP questions API. That question is only rendered by the web UI, so a
plain `agama config generate` invocation (e.g. from a script or terminal)
would hang forever waiting for an answer nobody could give.

## Solution

Move the "what should we do about unsupported elements?" decision out of the
conversion path and into the only component that actually needs to ask it:
`agama-autoinstall`. `agama config generate` now just reports the problems
and keeps writing the configuration to stdout, unchanged from a scripting
point of view.

### `agama-autoyast` (Ruby)

- No longer talks to the questions HTTP API. `Agama::AutoYaST::ProfileReporter`
  is removed.
- If unsupported elements are found (and `ay_check` is enabled), it writes
  them to `unsupported.json` in the working directory, alongside
  `autoinst.json`.

### `/profile/autoyast` (agama-server)

- Returns `{ "profile": <config>, "unsupported": [...] }` instead of just the
  raw configuration. `unsupported` is empty when no problems were found. The
  new `AutoyastConversionResult`/`UnsupportedElement` types live in
  `agama-lib::profile` so they can be shared by the server, the CLI and
  `agama-autoinstall`.

### `agama config generate` (agama-cli)

- Prints unsupported elements to **stderr** and keeps writing the
  configuration to **stdout**, so existing pipelines/scripts keep working.
- No question is raised from the CLI anymore.

### `agama-autoinstall`

- `ConfigLoader` gains an `agama-lib` HTTP client and fetches AutoYaST URLs
  directly via `ProfileHTTPClient::from_autoyast` instead of shelling out to
  `agama config generate`. Non-AutoYaST URLs and predefined locations keep
  using the existing subprocess-based pipeline.
- If the conversion reports unsupported elements, `UserQuestions` raises a
  `Continue`/`Abort` question (same `QuestionSpec`/HTTP questions mechanism
  used for `should_retry`), so the web UI's generic questions dialog handles
  it without any frontend changes. Choosing "Abort" is treated as a load
  failure, following the existing retry path.

## Testing

- Updated `service/test/agama/commands/agama_autoyast_test.rb` for the new
  `write_unsupported`/`check?` behavior.
- `cargo test` / `cargo build` across `agama-lib`, `agama-server`,
  `agama-cli`, and `agama-autoinstall`.


teclator (approved):
LGTM

PR #3838, opened 2026-08-26
Silence clippy::result_large_err for axum Response error handlers

imobachgs:
## Problem

`cargo clippy` flags several `agama-server` handlers with `clippy::result_large_err`:

`axum::response::Response` (`http::Response<axum::body::Body>`) is 128 bytes, exactly at clippy's default `large-error-threshold`. A number of handlers return `Result<T, Response>` (or `Result<Response, Response>`) to short-circuit with a pre-built HTTP response, which trips the lint.

## Solution

Add `#[allow(clippy::result_large_err, reason = "...")]` to the affected functions instead of boxing the response. This matches how axum itself handles the identical situation in its own examples (see [tokio-rs/axum#3824](https://github.com/tokio-rs/axum/pull/3824)): `Response` is the natural short-circuit error type for handlers/middleware, and the extra bytes are negligible compared to the cost of the HTTP request itself. Boxing would just move the cost around for no real benefit.

### Functions annotated

- `agama-server/src/profile/web.rs`: `validate`
- `agama-server/src/server/web.rs`: `get_status`, `get_system`, `get_extended_config`, `get_config`, `put_config`, `patch_config`, `get_proposal`, `get_issues`, `get_questions`, `ask_question`, `update_question`, `get_license`, `run_action`, `get_storage_model`, `set_storage_model`, `solve_storage_model`, `check_password`

## Testing

- `cargo clippy --workspace --all-targets -- -D clippy::result_large_err` passes with no remaining `result_large_err` warnings.
- `cargo build -p agama-server` succeeds.

### DIFF
--- a/rust/agama-server/src/profile/web.rs
+++ b/rust/agama-server/src/profile/web.rs
@@ -164,6 +164,11 @@ impl ProfileBody {
     }
 }
 
+#[allow(
+    clippy::result_large_err,
+    reason = "Response is used to short-circuit with a pre-built HTTP response; the extra \
+              bytes are negligible per-request cost (see tokio-rs/axum#3824)"
+)]
 async fn validate(body: String) -> Result<(), Response> {
     let profile = ProfileBody::from_string(body);
     let profile_content = profile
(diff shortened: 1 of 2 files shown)


dgdavid (approved):

teclator (approved):
LGTM

PR #3839, opened 2026-08-26
Do not cancel remaining tasks when network configuration fails

imobachgs:
## Problem

- [bsc#1276875](https://bugzilla.suse.com/show_bug.cgi?id=1276875)

If network configuration cannot be applied, Agama will not execute the remaining tasks
(including reading repositories and so on). The real problem happens in cases where you
try to, for instance, apply a wireless configuration to a device that does not exist.

## Solution

For 16.1, we will just log the error and not propagate the error, so the remaining tasks are not
canceled. For 16.2, we should go for registering issues instead.

## Testing

- Tested manually

### DIFF
--- a/rust/agama-manager/src/actions.rs
+++ b/rust/agama-manager/src/actions.rs
@@ -634,9 +634,7 @@ impl SetConfigAction {
                 .run(|| async move {
                     if let Err(error) = handler.update_config(network_config).await {
                         tracing::error!("Failed to update the network configuration: {error}");
-                    }
-
-                    if let Err(error) = handler.apply().await {
+                    } else if let Err(error) = handler.apply().await {
                         tracing::error!("Failed to apply the network configuration: {error}");
                     }
                     Ok(())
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM

rust/agama-manager/src/actions.rs
```diff
@@ -632,11 +632,13 @@ impl SetConfigAction {
...
-                    handler
-                        .update_config(network_config)
-                        .await
-                        .map_err(TaskError::from_error)?;
-                    handler.apply().await.map_err(TaskError::from_error)?;
+                    if let Err(error) = handler.update_config(network_config).await {
+                        tracing::error!("Failed to update the network configuration: {error}");
+                    }
+
+                    if let Err(error) = handler.apply().await {
```
teclator: In case of update error, does it makes sense to apply at all?

teclator (approved):
LGTM

PR #3841, opened 2026-08-28
Rename question classes and improve backward-compatible answer matching

imobachgs:
## Problem

Question "class" identifiers were inconsistent (mixed dotted/underscore
naming, e.g. `software.package_error.provide_error`), not compatible with
previously used names, and their `data` map keys mixed snake_case/camelCase
with no tolerance in `AnswerRule` matching.

## Solution

- Rename all question classes to consistent snake_case identifiers (Rust,
  Ruby, web UI, and docs updated accordingly).
- Add a centralized `resolve_class()` map next to `AnswerRule` so answer
  rules/profiles written against the old class names keep matching.
- Make `AnswerRule::answers_to` tolerant of both snake_case and camelCase
  `data` key naming.

## Testing

- `cargo build --workspace` / `cargo test --workspace`
- `bundle exec rspec` for the affected Ruby specs
- `npm test` (jest) for the affected web components

### DIFF
--- a/rust/agama-security/src/service.rs
+++ b/rust/agama-security/src/service.rs
@@ -226,7 +226,7 @@ impl Service {
         let labels = [gettext("Trust"), gettext("Reject")];
         let msg = gettext("Trying to import a self-signed certificate. Do you want to trust it and register the product?");
 
-        let question = QuestionSpec::new(&msg, "registration.certificate")
+        let question = QuestionSpec::new(&msg, "self_signed_regcert")
             .with_owned_data(certificate.to_data())
             .with_actions(&[
                 ("Trust", labels[0].as_str()),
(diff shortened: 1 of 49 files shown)


rust/agama-security/src/service.rs
```diff
@@ -226,7 +226,7 @@ impl Service {
         let labels = [gettext("Trust"), gettext("Reject")];
         let msg = gettext("Trying to import a self-signed certificate. Do you want to trust it and register the product?");
 
-        let question = QuestionSpec::new(&msg, "registration.certificate")
+        let question = QuestionSpec::new(&msg, "selfSignedRegcert")
```
jreidinger: this name is kind of inconsistent. I would expect something like `selfSignedRegCert`

jreidinger (approved):

PR #3844, opened 2026-08-28
Adjust default encryption derivation function based on the bootloader

ancorgs:
## Problem

See https://bugzilla.suse.com/show_bug.cgi?id=1258955

When encrypting a device, the user can specify the key derivation function algorithm to use (PBKDF2, ARGON2 or ARGON2i).

When the user omits that configuration detail, then PBKDF2 was used as a default fallback for all situations to ensure compatibility with all bootloaders and disk layouts. In fact, the default value was configurable by product, but no product used that capability so the hardcoded fallback PBKDF2 was always used in the end.

When using a BLS-capable bootloader like systemd-boot, there is no reason to use PBKDF2. Using it may be considered as reducing security for no reasons and can be perceived as a step back compared to the systemd-boot support present in the YaST installer.

## Solution

This change how the default KDF is calculated in case the user omits it. Instead of using a product-based configuration (that no product was using anyways). The KDF to use is now based on the bootloader that will be installed.

There is still room for improvement (using Grub2 should not imply that PBKDF2 is used for every single device), but this change is very safe and backwards compatible with the previous behavior. If Grub2 is used (which almost always happens in SLE, for example) absolutely nothing changes. Only if a different bootloader is used (which in SLE can only be done activating a tech preview) then the KDF is adjusted.

## Testing

- Added a new unit test
- Manually tested:
  - Verified that SLES keeps the behavior in all scenarios except the tech preview
  - Verified that now TW uses the default encryption KDF (that should be Argon2i) when using systemd-boot
  - Verified that TW keeps using PBKDF2 when using grub2 (basically all scenarios except EFI).
- Additionally, this is available for any further testing https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/fix_enc/images/iso/

### DIFF
--- a/service/lib/agama/storage/config_solvers/encryption.rb
+++ b/service/lib/agama/storage/config_solvers/encryption.rb
@@ -19,16 +19,19 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "agama/storage/config_solvers/base"
-
 module Agama
   module Storage
     module ConfigSolvers
       # Solver for the encryption configs.
       #
-      # The encryption configs are solved by assigning the default encryption values defined by the
-      # productd, if needed.
-      class Encryption < Base
+      # The encryption configs are solved by assigning the default encryption values required by the
+      # used bootloader, if needed.
+      class Encryption
+        # @param bootloader_config [Storage::BootloaderConfig]
+        def initialize(bootloader_config)
+          @bootloader_config = bootloader_config
+        end
+
         # Solves all the encryption configs within a given config.
         #
         # @note The config object is modified.
@@ -43,6 +46,12 @@ module Agama
 
       private
 
+        # @return [Storage::BootloaderConfig]
+        attr_reader :bootloader_config
+
+        # @return [Config]
+        attr_reader :config
+
         def solve_encryptions
           config.supporting_encryption.each { |c| solve_encryption(c) }
         end
@@ -51,9 +60,7 @@ module Agama
         def solve_encryption(config)
           return unless config.encryption
 
-          encryption = config.encryption
-          encryption.method ||= default_encryption.method
-          solve_encryption_values(encryption)
+          solve_encryption_values(config.encryption)
         end
 
         def solve_physical_volumes_encryptions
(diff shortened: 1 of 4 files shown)


service/lib/agama/storage/config_solvers/encryption.rb
```diff
@@ -71,24 +78,14 @@ def solve_physical_volumes_encryption(config)
...
-          config.label ||= default_encryption.label
-          config.cipher ||= default_encryption.cipher
-          config.key_size ||= default_encryption.key_size
-        end
-
-        # Default encryption defined by the product.
-        #
-        # @return [Configs::Encryption]
-        def default_encryption
-          @default_encryption ||= config_builder.default_encryption
```
joseivanlopez: `ConfigBuilder#default_encryption` was only used for this. I think we can remove it.

service/lib/agama/storage/config_solvers/encryption.rb
```diff
@@ -71,24 +78,14 @@ def solve_physical_volumes_encryption(config)
...
+          bootloader_type = bootloader_config.type
+          # The bootloader type is always known because bootloader_config is already solved,
+          # but this is defensive code to ensure backwards compatibility even with the tests
+          return if bootloader_type && !bootloader_type.is?(:grub2)
 
-          config.password ||= default_encryption.password
-          config.pbkd_function ||= default_encryption.pbkd_function
-          config.label ||= default_encryption.label
-          config.cipher ||= default_encryption.cipher
-          config.key_size ||= default_encryption.key_size
```
joseivanlopez, line 84: Aren't these default values needed?

service/lib/agama/storage/config_solvers/encryption.rb
```diff
@@ -51,9 +60,7 @@ def solve_encryptions
         def solve_encryption(config)
           return unless config.encryption
 
-          encryption = config.encryption
-          encryption.method ||= default_encryption.method
```
joseivanlopez, line 55: The dafault `method` is not assigned here but `#solve_physical_volumes_encryptions` still uses it. Is that ok?

service/lib/agama/storage/config_solvers/encryption.rb
ancorgs: No. They are actually never used.

Back then we probably just added every single possible attribute. But they are never defined anywhere by any product or used for anything.

service/lib/agama/storage/config_solvers/encryption.rb
ancorgs: Yes. The config conversor from JSON always assigns a method. It cannot actually be blank when solving has started.

ancorgs: Oh. I see what you mean. It can/must be removed from there as well. Let me double check.

joseivanlopez (approved):
LGTM

PR #3846, opened 2026-08-28
Added "SLES for NVIDIA" Kiwi profile

lslezak:
## Details

- Defined the Kiwi profile for the "SLES for NVIDIA" product
- Inherit the SLES defaults, just use `kernel-nvidia` instead of `kernel-default` and install the new product definition and the branding (defined in https://github.com/agama-project/agama-sles-nvidia)
- Disable self-update for it


teclator (approved):
The changes LGTM

PR #3851, opened 2026-08-31
CI: validate the *.kiwi file

lslezak:
## Problem

- We do not validate the Kiwi XML file in CI, that could help to find the issues sooner.

## Solution

- Validate the Kiwi XML files with `kiwi-ng image info`

## Testing

- Added a new CI test, see [example run](https://github.com/agama-project/agama/actions/runs/33394848242/job/99496810088#step:6:5)

### DIFF
--- a/.github/workflows/ci-live-kiwi.yml
+++ b/.github/workflows/ci-live-kiwi.yml
@@ -39,7 +39,7 @@ jobs:
         run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && zypper ref
 
       - name: Install the required tools
-        run: zypper --non-interactive install --no-recommends python3-kiwi git
+        run: zypper --non-interactive install --no-recommends python3-kiwi git make
 
       - name: Checkout repository
         uses: actions/checkout@v7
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-live-kiwi.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: zypper modifyrepo -d repo-non-oss repo-openh264 repo-update && zypper ref
+
+      - name: Install the required tools
+        run: zypper --non-interactive install --no-recommends python3-kiwi git
+
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Validate the Kiwi image definition
+        run: kiwi-ng image info --description live/src
```
mvidner: NP: I am not a fan of hiding useful commands like this in the Action YAML file. IMHO these should be easily runnable from my development shell (I prefer via `make check`)

Related: having to use `yq` to extract a shell script before checking it with shellcheck, in [ci-cache-cleanup.yml](https://github.com/agama-project/agama/blob/58f41f7e9bb98a7abf6f63b63715c86c6f2115aa/.github/workflows/ci-cache-cleanup.yml#L12)

mvidner (approved):
Thanks!
I have a suggestion to make it better, but it is not blocking.

mvidner (approved):

PR #3853, opened 2026-08-31
fix(web): default registration form to the effective registration URL

imobachgs:
## Problem

* [bsc#1277099](https://bugzilla.suse.com/show_bug.cgi?id=1277099)

The registration form only pre-filled the "Server URL" field (and selected
"Custom" as the registration server) when the user had explicitly typed a
custom URL before. When the URL came only from the `inst.register_url` boot
argument, the effective registration URL used by Agama was invisible to the
user, both before submitting the form and after a failed registration
attempt.

## Fix

The backend already exposes the effective, merged registration URL via
`GET /api/extended_config` (`product.registrationUrl`), regardless of
whether it came from the user-set configuration or from
`inst.register_url`. This change makes the registration form use that
endpoint (via a new `useExtendedProduct()` hook) to compute its defaults,
instead of the plain `/api/config`-backed `useProduct()`.

No backend changes were needed.

## Testing

- Added a test asserting the form defaults to the URL exposed via the
  extended configuration.
- Full frontend test suite and `tsc --noEmit` pass.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -3,7 +3,8 @@ Mon Aug 31 15:53:32 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Default the registration form to the effective registration URL
   (e.g., the one set via the inst.register_url boot argument),
-  instead of only showing it when the user typed a custom one.
+  instead of only showing it when the user typed a custom one
+  (bsc#1277099).
 
 -------------------------------------------------------------------
 Fri Aug 28 13:08:30 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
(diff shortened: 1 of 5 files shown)


dgdavid (approved):
LGTM 👍

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Aug 31 15:53:32 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Default the registration form to the effective registration URL
```
dgdavid, line 4: Missing bsc#1277099 reference

dgdavid (changes requested):
Missing bug reference in changelog

dgdavid (approved):
👍

PR #3855, opened 2026-09-01
fix(web): deal with unknown localization settings

dgdavid:
## Problem

* [bsc#1277313](https://bugzilla.suse.com/show_bug.cgi?id=1277313)

Agama now accepts a profile with an unknown timezone or locale and registers an installation issue for it, which is the right approach. The web UI does not cope with that case: the overview crashes with "Cannot read properties of undefined" and takes the whole page down.

The backend drops the localization proposal as a whole as soon as one of locale, keymap or timezone is unknown, and the summary read it anyway. On top of that, the issue scope was typed as `localization` while the backend sends `l10n`, so no localization issue was ever noticed by the UI.

| Overview | Page |
|-|-|
| localhost_8080_ - 2026-09-01T154623 979 | localhost_8080_ - 2026-09-01T154618 810 |

## Solution

The overview and the localization form now read the extended configuration, which always carries the three values, instead of the proposal. The summary reports "Invalid settings" like its neighbours do, and the localization page shows which setting is not recognized.

A nice side effect: the form keeps the recognized language and keyboard selected, so the user only has to fix the one field that is actually wrong instead of picking the three again.

| Overview | Page |
|-|-|
| localhost_8080_ - 2026-09-01T154413 338 | localhost_8080_ - 2026-09-01T154416 620 |

## Details

* `model/issue.ts` used a scope name the backend does not send.
* New `hooks/model/config/l10n.ts`, following the `useProduct` / `useExtendedProduct` pair added in #3853.
* The summary falls back to the configured id when the system reports no matching entry, so an unknown value shows as itself rather than as an error page.

## Testing

* New cases for the invalid settings path in the summary and the page, and for the fallback to raw ids.
* `npm run test`, `npm run eslint` and `npm run check-types` pass.
* Checked against a running installer with `timezone: "Europe/ndorra"` in the profile.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Sep  1 14:20:00 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Cope with unknown localization settings: report them as an
+  installation issue instead of crashing the overview, and keep the
+  recognized language, keyboard and time zone selected so only the
+  wrong one has to be fixed (bsc#1277313).
+
+-------------------------------------------------------------------
 Mon Aug 31 15:53:32 UTC 2026 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Default the registration form to the effective registration URL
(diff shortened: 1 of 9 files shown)


imobachgs (approved):

joseivanlopez (approved):
LGTM

PR #3858, opened 2026-09-02
Use camelCase for ieee8021x in Agama Network HTTP API

teclator:
## Problem

The network connection setting key `"ieee-8021x"` was defined in kebab-case, which deviated from our agreed design convention of using `camelCase` for all JavaScript and JSON objects in the Agama HTTP API.

## Solution

Consistent with our agreed API style guide, the serialized key was updated to camelCase (`"ieee8021x"`):
- Renamed the serialization/deserialization target to `"ieee8021x"`.
- Retained the old `"ieee-8021x"` key as an alias (`alias = "ieee-8021x"`) to ensure seamless backward compatibility with any older client configurations.

## Testing

- Added two new unit tests in `rust/agama-utils/src/api/network/settings.rs`:
  - `test_network_connection_ieee8021x_serialization` to verify serialization/deserialization of the camelCase key.
  - `test_network_connection_ieee8021x_alias` to verify backward-compatible deserialization of the kebab-case key.

### DIFF
--- a/rust/package/agama.changes
+++ b/rust/package/agama.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Sep  2 11:31:32 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
+
+- Use camelCase consistently for the JSON API keys renaming the
+  network connection "ieee-8021x" key to "ieee8021x", still
+  accepting the old name for backward compatibility.
+
+-------------------------------------------------------------------
 Fri Aug 28 15:32:30 UTC 2026 - Martin Vidner <mvidner@suse.com>
 
 - update runtime dependencies to:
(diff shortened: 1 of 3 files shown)


rust/package/agama.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Sep  2 11:31:32 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
```
```suggestion
Wed Sep  2 09:31:32 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
```

dgdavid:
This PR has a failing check that could be related with the merge of master in next.

I will address as soon as possible, but do not expect to have it fixed in less than 3 hours. I'm not on front of the laptop at this point.

imobachgs (approved):

PR #3859, opened 2026-09-02
refactor(web): rework the Popup component and its call sites

dgdavid:
## TL;DR

Popup implied more than it stated: which kind of dialog it is, whether it is open, which of its children are footer buttons. This rework turns each of those into something the caller states, and closes three gaps on the way: dialogs the user may walk away from now offer a close button and answer to Escape, every dialog carries a name screen readers can read, and Enter answers the questions asking for a value.

## Problem

Popup has been around since the very first screens and has gone through several rounds of UI changes since. It still does its job well, and a review of its call sites turned up a handful of details that the moves left behind: conventions that no longer apply, a couple of habits nobody had a reason to question, and three small gaps that were easy to miss because nothing complains about them.

A dialog the user may simply walk away from, like the license or the language settings, offered no close button and ignored Escape, so the only way out was the Cancel button in the footer. A dialog rendering no title was announced by screen readers as an unnamed dialog, since it pointed its accessible name at a heading that was never there. And in the questions asking for a password or a location, pressing Enter did nothing useful.

The other findings are traps rather than bugs, ways a call site could go wrong with nothing saying so. Footer content was chosen by comparing each child against a marker component, so buttons wrapped in a fragment, produced by a map or returned from another component quietly landed in the body, next to the content. A dialog could be left unnamed, or mounted without saying whether it is open, in which case it rendered nothing at all and looked broken. Dialogs with no buttons still got an empty footer, and a type suppression was hiding a real mismatch.

## Solution

Three fixes land where the user meets them: dialogs you may walk away from offer a close button and honor Escape, every dialog carries a name, and Enter answers any question that renders a form, with the very action its primary button carries. Which kind of dialog each one is now follows from the closing callback: with it the dialog is dismissible, without it the flow waits for an answer, as installer questions do.

The rest is the component saying out loud what it used to imply. Footer buttons travel in their own slot, so they can no longer scatter into the body and a dialog with none gets no footer. Being open is stated instead of guessed. The action vocabulary keeps its installer meaning (confirm, cancel, dangerous, ancillary), sheds a layer that added nothing to it, and gains the one it lacked: a closing action that reuses the dialog's own callback. The documentation follows, describing what a caller decides rather than how the buttons are painted.

The seven question components each repeated the same wiring. They now share a question dialog that takes the question and the callback answering it, and every question is left with little more than its own content.

Best read commit by commit: each one moves a single decision and is green on its own.

## Accessibility

- Every dialog carries an accessible name, checked at compile time rather than in an inspector.
- The dangling name reference is gone, so screen readers no longer follow it to nothing.
- Escape closes what can be closed and leaves must-answer dialogs alone, which is what the two roles promise.
- Enter answers questions with a form, so filling a password no longer requires leaving the keyboard to find the button.

## Follow-up

The dialogs with no size set are still content-sized, which makes their widths inconsistent. Deciding on a default is a visual call and deserves its own change rather than riding along here.

Renaming the component to Dialog, matching what ARIA and PatternFly call it, is correct in principle and pure churn in practice. Not now, and the same goes for reviewing the wrappers built on top of it, which this change only migrates.

## Testing

- The behavior was pinned first: the dialog suite gained characterization tests through accessible queries (name, description, the dismissal contract) before anything moved, including the two quirks this change removes.
- New tests cover the question dialog end to end (naming, actions, answering, disabled actions and the Enter path) and the Enter behavior of the questions that collect a value.
- `npm run test -- src`, `npm run check-types` and `npm run eslint` pass.
- Worth a manual pass in the running installer: Escape and the close button on the license and language dialogs, Enter on the encrypted device and package questions, and a screen reader on a dialog that renders no title.

---

Co-Authored-By: Claude <noreply@anthropic.com>


teclator (approved):
LGTM

PR #3861, opened 2026-09-02
32 bit tw

jreidinger:
## Problem
Tumbleweed still supports via ports i586 and armv6 and armv7, while agama is not built against it.

- https://trello.com/c/2i3cOxom/5986-2-drop-firefox-dependency-from-i586-armv6-and-armv7
- https://bugzilla.suse.com/show_bug.cgi?id=1266309

## Solution

Suseconnect is build only for 64 bit architectures, so make registration support available only on such architectures.
Adapt kiwi file to be able to create 32bit media.

## Testing

- *Tested manually* in repository https://build.opensuse.org/project/show/systemsmanagement:Agama:branches:32_bit_tw using qemu. Image is built, but grub2 cannot pass booting to linux kernel.


mvidner (changes requested):
The code looks good but please add the missing .changes entries

mvidner (approved):

PR #3862, opened 2026-09-02
fix(web): offer installer l10n settings before product selection

dgdavid:
## TL;DR

Offer the option to apply the installer language and keyboard to the product to install right from the start, including while the product is still being chosen.

## Problem

The dialog behind the language and keyboard button in the top bar lets users apply the chosen settings to the product to install too. Until a product was selected, that checkbox was hidden and a note took its place, saying the change would affect only the installer interface.

localhost_8080_ - 2026-09-03T090402 920

There was a good reason for hiding it. The installer had nowhere to keep a localization choice made before the product selection, so it dropped it.

That is no longer true. A configuration update is merged into what is already stored, and picking a product is just another update, so an earlier choice survives it. This comes from the config-based API (9cc8b2dba) and the merging logic in #2951, both landed back in [Agama 19](https://agama-project.github.io/blog/2026/03/20/agama-19).

## Solution

The checkbox is now part of the dialog in every case, checked by default as before. The note about affecting only the installer interface, along with the component that decided between the two, is gone.

Fixes [bsc#1277718](https://bugzilla.suse.com/show_bug.cgi?id=1277718) (protected link).

| The dialog sent | Resulting settings |
|-|-|
| localhost_8080_ - 2026-09-03T085734 058 | localhost_8080_ - 2026-09-03T085748 839 |
|localhost_8080_ - 2026-09-03T090902 007 | localhost_8080_ - 2026-09-03T090655 426 |

## Details

The checkbox is described by a note saying the product's own language and region settings may offer more options. That note matters most precisely while the product is being chosen, since the installer list is short and users would otherwise assume it is everything on offer, so it is always shown. What changes is where it sends them: with a product selected it links to the language and region page, and before that, when the page cannot be reached yet, it says the options turn up once a product is chosen. **Pointing at a page nobody can open yet would only raise the question of where that page is.**

> [!NOTE]
> 
> The wording still says "the selected product", which might, or might not, reads a bit ahead of itself during product selection. Left untouched, but happy to reword it if you prefer.

Nothing on the backend side pins down that localization set before product selection survives the selection, so the manual check below is what confirms it.

## Tests

Every dialog variant (language and keyboard, language only, keyboard only) now covers the no-product-selected case: the checkbox is offered, checked by default, and the chosen settings reach the product configuration on accept. The link to the language and region page is still expected to be absent at that point.

- [x] `npm run test`, `npm run eslint` and `npm run check-types` pass
- [x] Boot the installer, open the language and keyboard dialog from the product selection screen, pick a language with the checkbox left checked, then select a product and confirm the language and region page reflects the choice
- [x] Repeat with the checkbox unchecked and confirm the product settings stay untouched
- [ ] Repeat once a product is selected and confirm nothing changed there

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Sep  3 08:13:38 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Offer applying the installer language and keyboard to the product
+  to install also while the product is being selected (bsc#1277718).
+
+-------------------------------------------------------------------
 Wed Sep  2 14:47:52 UTC 2026 - David Diaz <dgonzalez@suse.com>
 
 - Close dialogs with Escape or a close button, answer questions by
--- a/web/src/components/core/InstallerL10nOptions.test.tsx
+++ b/web/src/components/core/InstallerL10nOptions.test.tsx
@@ -289,11 +289,26 @@ describe("InstallerL10nOptions", () => {
         mockProduct(undefined);
       });
 
-      it("does not allow reusing setting", async () => {
-        await renderAndOpen();
-        const dialog = screen.getByRole("dialog");
-        expect(within(dialog).queryByRole("checkbox")).toBeNull();
-        screen.getByText(/This will affect only the installer interface/);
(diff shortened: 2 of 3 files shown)


teclator (approved):
LGTM

PR #3863, opened 2026-09-03
Sanitize web dependencies

### LINKED ISSUE #3899: Reduce Node/JavaScript and Rust dependencies
Before the next release, it would be a good idea to reduce the number of dependencies in the JavaScript and Rust code bases. The aim is to reduce our exposure to security bug reports.

Check for:

- Unused dependencies.
- Multiple version of the same dependency.

### Links

- [cargo-machete](https://github.com/bnjbvr/cargo-machete "‌") or [cargo-udeps](https://crates.io/crates/cargo-udeps "‌").
- [knip](https://knip.dev/ "‌") (not tested)

### Review

- [https://github.com/agama-project/agama/pull/3852](https://github.com/agama-project/agama/pull/3852 "smartCard-inline")
- [https://github.com/agama-project/agama/pull/3856](https://github.com/agama-project/agama/pull/3856 "smartCard-inline")
- [https://github.com/agama-project/agama/pull/3863](https://github.com/agama-project/agama/pull/3863 "smartCard-inline")

### Acceptance criteria

- [ ] Check and reduce JavaScript/Node dependencies
- [x] Check and reduce Rust dependencies
- [ ] Document the tools we used, so the next time it could be automated

joseivanlopez:
## Summary

This PR reorganizes web dependencies to keep OBS package builds as minimal as possible, reducing CVE exposure and security maintenance burden.

## Changes

- Moved test frameworks and tools to `optionalDependencies`
- Moved linters and code quality tools to `optionalDependencies`
- Moved development tools to `optionalDependencies`
- Moved build tools to `devDependencies`
- Added CI automation: Knip tool to detect unused dependencies and files
-  Removed unused files

By making development and testing dependencies optional, production builds only include the packages strictly necessary to run and build the application, significantly reducing the attack surface and security maintenance overhead.


web/knip.json
```diff
@@ -0,0 +1,20 @@
...
+    "src/po/po.*.js",
+    "svgo.config.js"
+  ],
+  "ignoreDependencies": [
+    "typescript-eslint"
+  ],
+  "webpack": {
+    "config": ["webpack.config.js"]
+  },
+  "ignoreExportsUsedInFile": true
```
dgdavid: When I was playing with that library, I found more useful these exports to be documented instead. Dont  remember exactly the details, but it was using the `@internal` TSDoc tag explaining why such a export existed (to be used in tests, for example).

https://gist.github.com/dgdavid/2e11356603443b5bc946cbec950478d6

web/package.json
```diff
@@ -95,9 +94,11 @@
   "dependencies": {
     "@material-symbols/svg-400": "^0.40.2",
     "@material-symbols/svg-700": "^0.40.2",
+    "@monaco-editor/react": "^4.7.0",
```
dgdavid, line 99: Please, check this do not defeat https://github.com/agama-project/agama/pull/3658/changes

web/package.json
```diff
@@ -25,12 +25,11 @@
   "devDependencies": {
```
mvidner, line 25: I forgot to mention:
`devDependencies` should contain only what is needed in the OBS for building the RPM.
What is needed in CI and for local development should go to `optionalDependencies` so that we ship as few dependencies as needed.
https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#optionaldependencies

web/package.json
```diff
@@ -43,7 +42,6 @@
     "@types/webpack-env": "^1.18.8",
     "@typescript-eslint/eslint-plugin": "^8.56.1",
     "@typescript-eslint/parser": "^8.56.1",
-    "ajv": "^8.18.0",
```
mvidner, line 46: Nice catch, I remember `ajv` from one of my CVE bugs 👍

mvidner (changes requested):
This is good, but I want MORE, and it unfortunately conflicts with your work so far.

Also, I really want `optionalDependencies` in SLE-16.0 (not necessarily right now for GMC, but definitely later)

web/package.json
lslezak: It's OK, that is already used at https://github.com/agama-project/agama/pull/3658/changes#diff-2f189285027ee814c50aae51e0a1127e16d03c902aee39349d31b094d2fefb59R28 I just forgot to add explicitly into `packages.jpon`. It was pulled in by some other dependency so it worked fine but could break in the future.

.github/workflows/ci-web.yml
```diff
@@ -36,6 +36,40 @@ on:
...
+        cache: 'npm'
+        cache-dependency-path: 'web/package-lock.json'
+
+    - name: Install dependencies
+      run: npm ci
+
+    - name: Run knip
+      # Only check for unused files and dependencies, not exports (which are often
+      # intentional API surface). This keeps CI focused on catching real issues.
+      run: npx knip --exclude exports,types
```
lslezak: I'd probably move the step to the `frontend_build` job defined below, we already run checks like `npm run check-types` there.

The point is that running `knip` takes just ~2 seconds but setting up the environment (Git checkout, nodeJS and Npm setup) takes much more time. Let's run all quick checks together.

lslezak (commented):
Move to the already existing CI job?

web/package.json
dgdavid: Perfect, thanks for the double check.

web/knip.json
joseivanlopez: All test-only exports already have @internal comments properly documented. I found 3 functions that are exported specifically for testing:

1. getIpAddresses in src/hooks/model/system/network.ts
   - Marked with @internal
   - Comment explains: "Extracted and exported for testing purposes only"
2. getNetworkStatus in src/hooks/model/system/network.ts
   - Marked with @internal
   - Detailed explanation of why it was extracted for testing
3. parsePasteEntries in src/components/form/ArrayField.tsx
   - Marked with @internal Exported for testing only.

All of these follow the pattern from the gist you referenced - they're internal helper functions extracted from hooks/components to enable proper testing, and they're clearly documented as test-only exports.

web/package.json
joseivanlopez: The Situation

PR #3658 (already merged) did the following:
1. Added monaco-editor-webpack-plugin to bundle Monaco editor locally
2. Configured the code to use loader.config({ monaco }) to tell the Monaco loader to use local files instead of CDN

Current PR #3863 is adding @monaco-editor/react to package.json because:
- The code is already using it (see ConfigEditor.tsx line 28: import { loader } from "@monaco-editor/react")
- It was just missing from the dependency declaration

Why It's Safe

The @monaco-editor/react package is just a React wrapper/utility library for Monaco editor. It provides:
- The loader utility to configure where Monaco loads from
- React component wrappers

It doesn't automatically download from CDN. The code explicitly configures it to use the local bundle:

~~~js
import { loader } from "@monaco-editor/react";
// ...
// avoid downloading the monaco editor parts from the CDN, load the locally bundled files
loader.config({ monaco });
~~~

This configuration (already in master since PR #3658) ensures that @monaco-editor/react uses the locally bundled Monaco files created by monaco-editor-webpack-plugin, not the CDN.

Conclusion

Adding @monaco-editor/react to package.json is the correct fix - it declares a dependency that's already being used in the code and properly configured to work with local bundles. It does not reintroduce CDN downloads.

joseivanlopez (commented):
Good catch! Based on the analysis:

**OBS build** (
> build
> webpack

webpack compiled successfully) only needs webpack and build tools. The following newly added dependencies should go to `optionalDependencies`:

1. **`@jest/globals`** - Used only in test files
2. **`gettext-parser`, `glob`, `jed`** - Used by `share/po/po-converter.js` to convert translations. Since `po.*.js` files are pre-compiled and committed to git, these are only needed for local translation updates
3. **`knip`** - Dependency analysis tool, runs only in CI

The translation workflow is:
- Developers update `.po` files locally (needs po-converter dependencies)
- Compiled `po.*.js` files are committed
- OBS build just bundles the existing `po.*.js` files ✅

I'll move these 5 packages to `optionalDependencies` to keep OBS builds lean.

web/package.json
joseivanlopez: Good catch! Based on the analysis:

**OBS build** (
> build
> webpack

webpack compiled successfully) only needs webpack and build tools. The following newly added dependencies should go to `optionalDependencies`:

1. **`gettext-parser`, `glob`, `jed`** - Used by `share/po/po-converter.js` to convert translations. Since `po.*.js` files are pre-compiled and committed to git, these are only needed for local translation updates
2. **`knip`** - Dependency analysis tool, runs only in CI

The translation workflow is:
- Developers update `.po` files locally (needs po-converter dependencies)
- Compiled `po.*.js` files are committed
- OBS build just bundles the existing `po.*.js` files ✅

I'll move these packages to `optionalDependencies` to keep OBS builds lean.

.github/workflows/ci-web.yml
joseivanlopez: Good point!

web/knip.json
dgdavid: Then, why ignore exported but not used stuff? I would expect not face an export that is not needed.

web/package.json
dgdavid: Great! Nice to have it fully reasoned, although I just wanted to avoid going against previous work accidentally. Great to see everything is okish.

web/knip.json
joseivanlopez: See the analysis here https://gist.github.com/joseivanlopez/8b0aa72e4693cf11be62098f135a7548.

web/package.json
mvidner: Thanks! But... This has only removed 5 of the old `devDependencies`, I was hoping for many more like all the eslint and its plugins...
I am testing a more aggressive approach right now, will post my proposal shortly.

mvidner: It is in https://github.com/agama-project/agama/commit/578265334d4c40b5a72e54f4aa056f3e9384827f on https://github.com/agama-project/agama/compare/fix-js-deps-mvidner?expand=1

mvidner: It is kinda working (except `knip` part :scream: ) but now my `node_modules.obscpio` has changed size from 128 MB to 162 MB :scream: so I guess I need to add a feature to omit the optional deps to https://github.com/openSUSE/obs-service-node_modules

mvidner: A quick-and-dirty node_modules.py patch reduces the obscpio to 81 MB. But will it build??

mvidner:
Now osc build fails, I need to bring the CI test builds and the actual OBS builds closer together :frowning_face: 

```sh
cd ~/obs/systemsmanagement:Agama:Devel/agama-web-ui
cp ~/svn/agama/web/package/* .
# this uses --omit=optional, but master has nothing there, it omits nothing, succeeds
osc build

# edit _service to change from master to <param name=revision>fix-js-deps
osc service manualrun
rm *.tgz
# this uses --omit=optional, but they are still present in node_modules.obscpio
# and it fails! ajv-keywords somehow sneaks in and wants more modules...
osc build
```

mvidner:
Another important thing: I want this for the 16.1 branch. Maybe not right now for GMC, but eventually, as an update.
But we can target `master` now and backport it later.

PR #3860, opened 2026-09-02
Fix UI changelog entry

teclator:
## Problem

There was an extra blank line in the changelog file (`web/package/agama-web-ui.changes`), which violates the expected RPM/packaging format.

## Solution

Remove the redundant empty line between the two changelog entries in `web/package/agama-web-ui.changes`.

PR #3864, opened 2026-09-03
refactor(web): rework the web UI page shell, dialogs and URL state

dgdavid:
## TL;DR

Merge the accumulated work in `next` into `master`. Every change here already went through its own review.

## Big picture

The work falls into four parts, plus a backend rename.

**The URL now describes what you are looking at.** Expanded storage sections, active tabs, table filters and sort order, and the software patterns search all live in the address bar instead of in React state. Reloading a page or sharing a link reopens the same view. This arrived through a chain of PRs that built the hooks first (`useSearchParamState`, `useSortedByParam`, `useFilterParams`) and then moved one table at a time, and it retired the storage UI-state provider. The reasoning is written down in the new `doc/ui/guidelines/url-state.md`, indexed from a new `doc/ui/guidelines.md`.

**The page shell was reworked.** `core/Page` was a single 530-line module holding several layouts. It is now `web/src/components/layout/` with one module per layout (`StandardLayout`, `MinimalLayout`, `SideBySideLayout`), documented in `layout/conventions.md`. Along the way the heading outline was fixed, the minimal layout stopped shifting content by not reserving masthead space, and `SplitInfoLayout` became `SideBySideLayout` with its title and body actually aligned.

**Dialogs and accessibility.** `core/Popup` got a contract: every dialog has a name, `isOpen` is required, the action vocabulary is explicit, dismissible dialogs close with Escape and a close button, and Enter answers a question that renders a form. A global live region was added so the UI can announce things to screen reader users. Dead components (`core/EmptyState`, `WifiNetworksList`, `UsedSize`) were removed rather than carried along.

**Forms and network.** The product selection form was rebuilt on the new form stack (TanStack Form), the device selector grew a dialog that shows network devices with their details instead of a bare list, and new connections stopped being bound to the loopback device. `form/conventions.md` records the conventions.

**Backend.** The network HTTP API key `ieee-8021x` was renamed to `ieee8021x` for camelCase consistency. The old name is still accepted, so this is not a breaking change.

Two bug fixes are also user-visible: installer language and keyboard settings can now be applied to the product to install while the product is still being chosen (bsc#1277718), and the loopback binding fix above.

## Included pull requests

Layout and page shell:

- #3753 refactor(web): rework SplitInfoLayout into SideBySideLayout, fix title/body alignment
- #3754 fix(web): reserve masthead space in Page's minimal layout
- #3761 refactor(web): rework the page shell and fix the heading outline
- #3762 refactor(web): move each page layout to its own module

URL as UI state:

- #3773 refactor(web): move the storage view from a provider to the URL
- #3785 feat(web): keep the tables sort order in the URL
- #3787 feat(web): keep the table filters in the URL
- #3788 feat(web): keep the patterns search in the URL

Dialogs, accessibility and cleanup:

- #3759 feat(web): add a global live region for screen reader announcements
- #3816 refactor(web): remove the dead core/EmptyState and stale consumers
- #3859 refactor(web): rework the Popup component and its call sites

Forms and network:

- #3791 refactor(web): migrate the product selection form to the new forms stack
- #3821 fix(web): stop binding new connections to the loopback device
- #3822 feat(web): allow browsing network devices with details when binding a connection
- #3862 fix(web): offer installer l10n settings before product selection

Backend and housekeeping:

- #3858 Use camelCase for ieee8021x in Agama Network HTTP API
- #3860 Fix UI changelog entry

## Addendum: two fixes added after the ISO test

> [!WARNING]
> 
> Testing the `next` image turned up a layout problem that does not reproduce on the development server: the header was oversized and the spacing was wrong across the whole interface. Tracking it down produced the two commits below, which belong to no pull request listed above.

- 08d0d66 fix(web): stop a stray BOM from killing PatternFly's base reset

  Production builds compress the Sass output, and Dart Sass marks a stylesheet containing non-ASCII characters with a byte-order mark rather than a `@charset` rule. PatternFly's base stylesheet gets one, and concatenating the extracted stylesheets into a single `index.css` moves that mark into the middle of the file, where it stops being skippable whitespace and becomes part of the selector that follows it. That selector then matches nothing, and the rule it silently disabled was the reset zeroing the margins and padding of headings, paragraphs, lists and much else, so browser defaults came back everywhere: a 10.72px margin on the masthead heading, 18px on card titles, and so on down the page. Only production builds compress, which is why the development server never showed it. Sass is now told not to emit the mark, which costs nothing because the page already declares UTF-8 through a meta tag.

- a9c5e74 fix(web): give the font size scale an absolute 200 step

  Unrelated to the symptom above, found in the same pass. Every step of the font size scale is expressed in `rem` except the 200 step, which used `em`. Custom properties are inherited unresolved, so that `em` was measured against whichever element read the token rather than against the root where it is declared. Since the small, body and body-small sizes all resolve through the 200 step, elements asking for those sizes were told to keep the size they already had, which turned the matching PatternFly utilities into no-ops and let sizes compound with nesting. The step is now absolute like its neighbours.

Worth re-running the manual pass on a fresh image, since the first fix changes spacing on every screen.

## Tests

Each PR carries its own tests; the suite grew by roughly a thousand lines of new test files (URL state hooks, announcer, layouts, Popup, device selector dialog).

Since this touches most of the interface, please give it a manual pass too. The `next` branch image is built at

https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/next/images/iso/agama-installer.x86_64-openSUSE.iso

(note that branch images are rebuilt and dropped regularly, so that link is only good while this PR is open; anyone reading later might face a dead link)

- [x] `npm run test`, `npm run eslint` and `npm run check-types` pass
- [x] Boot the installer and walk product selection, storage, software, network and users, checking no layout regressed
- [ ] Reload a page with expanded storage sections, a filtered or sorted table, and a patterns search, and confirm the view comes back
- [ ] Open and dismiss dialogs with Escape and the close button, and answer a question with Enter
- [ ] Post a network connection using the old `ieee-8021x` key and confirm it is still accepted

---

Co-Authored-By: Claude <noreply@anthropic.com>


dgdavid:
Sadly, in my manual tests with a downloaded ISO I see a lot of extra spacing everywhere

image

Investigating it.

dgdavid:
> Sadly, in my manual tests with a downloaded ISO I see a lot of extra spacing everywhere
> 
> ...

Although it does not happen locally :man_shrugging: 

localhost_8080_ - 2026-09-03T155627 482

dgdavid:
@imobachgs, @teclator 

I have added two commits here after [manually testing the ISO and found some weird spacing](https://github.com/agama-project/agama/pull/3864#issuecomment-5526620633) Check the "Addendum" note in the PR description to understand why.

Let's see, when images rebuild, if these fix the problem faced by me.

dgdavid:
> @imobachgs, @teclator
> 
> I have added two commits here after [manually testing the ISO and found some weird spacing](https://github.com/agama-project/agama/pull/3864#issuecomment-5526620633) Check the "Addendum" note in the PR description to understand why.
> 
> Let's see, when images rebuild, if these fix the problem faced by me.

Verified on a running ISO.

Rather than wait for a new image, I tested the fix directly on the ISO that was already running. The fix removes exactly three bytes from the built `index.css`, so I removed those same three bytes on the live system and reloaded the page. Same file, same server, same browser, only those bytes different.

| Before | After |
|-|-|
|  |  |

That rule clears the margins and padding of headings, paragraphs, lists and much else, which is why the whole interface looked off rather than one component.

This confirms the diagnosis and that removing the byte restores the layout. The other half, that the build no longer produces the byte in the first place, I checked by compiling the three stylesheets the app imports and confirming the byte is gone. **A fresh image is still worth a manual pass, since this changes spacing on every screen.**

imobachgs (approved):

PR #3865, opened 2026-09-04
Fix panic on missing profile in /private/profile endpoints

imobachgs:
## Problem

* [bsc#1278738](https://bugzilla.suse.com/show_bug.cgi?id=1278737)

`POST /api/private/profile/evaluate` or `.../validate` with an empty/`{}` body panicked the request handler (`.expect("Missing profile")`) instead of returning an error.

## Fix
- Add `ProfileError::MissingProfile` and return it as a proper error instead of panicking.
- Fix the error→HTTP-status mapping: `ProfileError` relied on `ProblemDetails::generic()`, which always maps to 500 despite comments claiming these were 400 client errors. Added a real `ProblemDetails::BadRequest` (400) variant and used it where appropriate.
- Fix `server::web::Error::Profile`, which was discarding `ProfileError`'s per-variant mapping and collapsing every profile error into a generic 500.

## Tests
- Unit tests for `retrieve_profile()`'s "nothing given" cases and the new error mapping.
- New `tests/profile_service.rs` integration tests hitting `/private/profile/validate` and `/private/profile/evaluate` over HTTP, asserting 400 instead of a crash.

Full `agama-server`/`agama-utils` test suite, `cargo fmt`, and `cargo clippy` pass with no new warnings.

### DIFF
--- a/rust/agama-server/src/profile/web.rs
+++ b/rust/agama-server/src/profile/web.rs
@@ -70,35 +70,48 @@ pub enum ProfileError {
     InvalidJson(#[from] serde_json::Error),
     #[error("{0}")]
     BadRequest(String),
+    #[error("No profile was given. Provide one of: path, url or profile (request body).")]
+    MissingProfile,
 }
 
-impl IntoResponse for ProfileError {
-    fn into_response(self) -> Response {
-        let problem = match self {
+impl ProfileError {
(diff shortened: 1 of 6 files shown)


joseivanlopez (approved):
LGTM

PR #3866, opened 2026-09-04
Remove dead code from agama-server

imobachgs:
## Problem
Three modules in `agama-server` had no callers anywhere in the workspace:

- `src/dbus.rs`: `DBusObjectChangesStream`/`ObjectsCache`, unused.
- `src/error.rs`: a crate-level `Error` enum, unused and shadowed in practice by an unrelated same-named type in `server::web::Error` that is actually used.
- `src/bootloader.rs`: declares `pub mod web;` for a `src/bootloader/web.rs` file that does not exist. It was never even wired into `lib.rs`, so it was never compiled as part of the crate.

## Fix
Remove the three files and their module declarations. Also drop the `pin-project` and `tokio-stream` dependencies, which were only used by the removed `dbus.rs` (confirmed unused via `cargo machete`).

## Tests
`cargo check`/`cargo test`/`cargo clippy`/`cargo fmt --check` all pass with no new warnings for `agama-server`.


mvidner (approved):
LGTM

PR #3867, opened 2026-09-04
Add support for Tumbleweed for riscv64

andreas-schwab:


live/config-cdroot/fix_bootconfig.riscv64
```diff
@@ -0,0 +1,156 @@
+#! /bin/bash
```
imobachgs, line 1: Note: the code in this file is basically the same than `aarch64` (but renaming the paths and so on to `riscv64`). /cc @jreidinger

jreidinger:
@andreas-schwab if you think it is ready for test build of iso, I can change target branch to test branch in agama and trigger our CI and create OBS testing project for you with your changes?
In general change looks good to me. Maybe that bootconfig can be simplified as condition for efi is probably not needed as risc support only efi, right? and not sure if systemd-boot supports risc

andreas-schwab:
aarch64 only supports efi as well.

systemd-boot does support riscv.

https://download.opensuse.org/ports/riscv/tumbleweed/appliances/iso/agama-installer.riscv64-openSUSE.iso

PR #3868, opened 2026-09-04
Add a web terminal to Agama's web UI

imobachgs:
> [!NOTE]
> For reviewers: the [doc/web-terminal.md](https://github.com/agama-project/agama/blob/web-terminal/doc/web-terminal.md) contains a description of the implementation. Moreover, you can [download the testing ISO](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/branches:/web-terminal/images/iso/).

## Problem

Agama users connected to the web UI have no way to interact with the underlying system without switching to a different tty (local) or connecting through SSH (remote).

## Solution

- **Backend** (`agama-server`): a new `GET /api/terminal/ws` endpoint spawns a shell attached to a pty (`pty-process`) for each connection. Binary frames carry raw keystrokes/output; text frames carry small JSON control messages (resize, exit). No server-side session state: closing the socket ends the shell, matching the "no persistence" error-handling requirement.
- **Frontend**: the previously-mocked terminal panel (`TerminalDock`/ `TerminalPane`) is now wired to a real `@xterm/xterm` instance connected to the backend above. The terminal is reachable from the "More options" menu ("Open terminal"/"Close terminal"). The panel offers two actions besides closing: minimizing (collapses to a header bar, session stays alive) and closing (ends the session).

## Testing

- Backend: new integration test suite (`rust/agama-server/tests/terminal.rs`) against a real bound server — running commands, creating files, resizing, auth rejection, and verifying the shell is reaped after disconnect.
- Frontend: unit tests for the terminal context, `TerminalDock`, `TerminalPane`, `InstallerOptionsMenu`, and a dedicated `useTerminalSession` hook test (mocked `xterm`/`WebSocket`).
- `cargo test`/`clippy`, `npx jest`, `eslint`, `stylelint`, and `tsc` all pass.

## Documentation

- Added a short section to `rust/WEB-SERVER.md` documenting the new endpoint.

## Screenshots

<details>
<summary>`agama config edit` running in the terminal</summary>

imagen
</details>


web/src/assets/styles/components/_agama-custom.scss
```diff
@@ -363,14 +363,14 @@ nav.agm-breadcrumb > ol > li > a {
   display: none;
 }
 
+// Hosts the xterm.js instance. Unlike a plain `<pre>`, xterm.js manages its
+// own internal scrolling and text rendering, so this only needs to give it a
+// definite size and a background matching its own theme (set when the
+// terminal is created; see `useTerminalSession`) for the padding around it.
```
dgdavid, line 369: I'd drop the `Unlike a plain <pre>` part since it's mostly contextual commentary about the previous mocking approach rather than something needed to explain the implementation. 

```suggestion
// Hosts the xterm.js instance. Give it a definite size and a background
// matching its theme (set when the terminal is created; see
// `useTerminalSession()` for the surrounding padding).
```

Also, we could avoid hard-coding the terminal background here and use one of the existing semantic surface tokens. `--agm-t--surface--color` is used for page and card backgrounds, while `--agm-t--surface--color--raised` is intended for secondary surfaces. I'm not sure either is the perfect semantic fit for a terminal, though. Perhaps this is a good case for introducing a dedicated terminal background token if the terminal is intentionally distinct from the surrounding surface.

web/src/components/core/InstallerOptionsMenu.tsx
```diff
@@ -126,6 +128,14 @@ export default function InstallerOptionsMenu({ hideLabel = false }: InstallerOpt
                 {/* TRANSLATORS: menu entry to download the installer logs as an archive */}
                 <ItemContent icon="archive" text={_("Download logs")} />
               </DropdownItem>
+              <DropdownItem key="toggle-terminal" onClick={toggleTerminal}>
+                {/* TRANSLATORS: menu entry that opens or closes the terminal
```
dgdavid, line 132: Will this translator comment apply to both translations below?

web/src/hooks/use-terminal-session.ts
```diff
@@ -0,0 +1,205 @@
...
+ * the real terminal) — the terminal still connects right away, and attaches
+ * to the container as soon as one becomes available.
+ *
+ * If the WebSocket disconnects unexpectedly, a new one is opened (a new
+ * backend shell), reusing the same terminal instance and scrollback; per the
+ * terminal's error handling rules, there is no attempt to recover the
+ * previous shell. Reconnection attempts back off exponentially (capped) so a
+ * persistently failing connection (e.g., an expired token or a backend that
+ * is down) does not hammer the server with an unbounded, un-throttled retry
+ * loop.
```
dgdavid, line 69: I would like to see this in action.

web/src/components/core/InstallerOptionsMenu.tsx
lslezak: Unfortunately not, you have to repeat the comment for each string. And more over this is a wrong place, the comment must be immediately before the translated string otherwise it is not assigned to it. See below.

web/src/components/core/InstallerOptionsMenu.tsx
```diff
@@ -126,6 +128,14 @@ export default function InstallerOptionsMenu({ hideLabel = false }: InstallerOpt
...
+                    panel, starting or ending its shell session. */}
+                <ItemContent
+                  icon="terminal"
+                  text={isTerminalOpen ? _("Close terminal") : _("Open terminal")}
```
lslezak, line 136: To correctly place the translator's comments use this syntax:

```suggestion
                  text={isTerminalOpen ? _("Close terminal") : _("Open terminal")}
                  text={
                    isTerminalOpen
                      ? /* TRANSLATORS: menu entry that closes the terminal
                       panel, starting or ending its shell session. */
                        _("Close terminal")
                      : /* TRANSLATORS: menu entry that opens the terminal
                       panel, starting or ending its shell session. */
                        _("Open terminal")
                  }
```

web/src/components/core/InstallerOptionsMenu.tsx
```diff
@@ -126,6 +128,18 @@ export default function InstallerOptionsMenu({ hideLabel = false }: InstallerOpt
...
                 <ItemContent icon="archive" text={_("Download logs")} />
               </DropdownItem>
+              <DropdownItem key="toggle-terminal" onClick={toggleTerminal}>
+                <ItemContent
+                  icon="terminal"
+                  text={
+                    isTerminalOpen
+                      ? /* TRANSLATORS: menu entry that closes the terminal, closing the session */
+                        _("Close terminal")
+                      : /* TRANSLATORS: menu entry that closes the terminal */
```
dgdavid:
```suggestion
                      : /* TRANSLATORS: menu entry that opens the terminal */
```

web/src/components/core/TerminalToggleButton.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    await user.click(button);
+    screen.getByRole("button", { name: "Close terminal" });
+
+    await user.click(screen.getByRole("button", { name: "Close terminal" }));
+    screen.getByRole("button", { name: "Open terminal" });
+  });
+
+  it("accepts custom props", () => {
+    plainRender(<TerminalToggleButton data-testid="custom-terminal-toggle" />);
+    screen.getByTestId("custom-terminal-toggle");
```
dgdavid: NP: Nothing wrong here, but I would use another attr/prop for this test for not opening the door to someone else thinking we are using data-testid. Maybe "autofocus" or or so.

dgdavid (approved):
Web part LGTM 👍 

I would like to see it in action before merging, if possible.

dgdavid:
After manually testing it and before a talk with @imobachgs, I have done some follow-up work to make the terminal usable with the keyboard alone.

A terminal grabs every key, <kbd>Tab</kbd> included, since shells use <kbd>Tab</kbd> for completion. So once a keyboard user landed in it, the only way out was exit or <kbd>Ctrl</kbd>+<kbd>D</kbd> (after commit https://github.com/agama-project/agama/pull/3868/commits/5ff08e1f2c00a69805edd97c23251abb9a2d3b61): you had to kill the session just to go look at another page. Pointer users click away and come back to it untouched.

Now <kbd>Escape</kbd> then <kbd>Tab</kbd> moves the focus out, the same sequence embedded code editors use (Monaco, CodeMirror, Ace). <kbd>Escape</kbd> alone still reaches the shell; only a <kbd>Tab</kbd> right after it is intercepted, and that pair means nothing to a shell. A short hint under the terminal spells it out, and it doubles as the terminal's accessible description so screen reader users hear it on focus.

Getting in is fixed too: a skip link to the terminal appears whenever it is open, reopening the panel first if minimized. It is told apart from "skip to content" by an icon and bolder text. Opening the terminal already put you straight into the shell, and it still does, but picking any other entry in the options menu used to drop the focus at the top of the page when the menu closed; it now goes back to the menu toggle.

---

Related to (found while doing research after realizing about the terminal focus trap):

* https://github.com/xtermjs/xterm.js/discussions/5306
* https://github.com/xtermjs/xterm.js/discussions/4569

dgdavid:
localhost_8081_

localhost_8081_ (1)

---


localhost_8081_ (5)
localhost_8081_ (4)

---

localhost_8081_ (8)
localhost_8081_ (7)

dgdavid:
localhost_8081_ (3)

doc/web-terminal.md
```diff
@@ -0,0 +1,235 @@
...
+## Keyboard accessibility
+
+A terminal has to take over almost every key, Tab included (shells use it to complete words). Left
+as-is, that is a keyboard trap: once focused, someone not using a pointer has no way back to the
+rest of the interface, which WCAG forbids ([SC 2.1.2, "No Keyboard
+Trap"](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html)).
+
+The way out is **Escape, then Tab**, which is the same sequence embedded code editors use for this
+(Monaco, CodeMirror, Ace). Escape still reaches the shell as usual; only a Tab typed right _after_
+an Escape is intercepted, moving the focus to the panel around the terminal instead. That sequence
```
dgdavid: This isn't fully accurate, and **I owe an apology**, since I was the one who assumed it was without doing thorough research. I've done it now, and each editor handles it differently:

- **CodeMirror**: <kbd>Escape</kbd> then <kbd>Tab</kbd>. https://codemirror.net/examples/tab/
- **Monaco**: <kbd>Ctrl</kbd>+<kbd>M</kbd> toggles whether <kbd>Tab</kbd> moves focus. https://github.com/Microsoft/monaco-editor/wiki/Monaco-Editor-Accessibility-Guide#tab-trapping
- **Ace**: with `enableKeyboardAccessibility: true` (since 1.17.0), <kbd>Enter</kbd> starts editing and <kbd>Escape</kbd> exits. https://github.com/ajaxorg/ace/blob/master/CHANGELOG.md#1170-2023-04-12

I was going to suggest reusing Monaco's shortcut for consistency, but that doesn't work either: the PatternFly `CodeEditor` we use doesn't rely on it. It wraps Monaco in its own focusable container, where <kbd>↓</kbd> or <kbd>Space</kbd> enters the editor and <kbd>Shift</kbd>+<kbd>Tab</kbd> leaves it ([6.5.1 source](https://github.com/patternfly/patternfly-react/blob/@patternfly/react-code-editor@6.5.1/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx)). So none of these is a convention we already follow, and the terminal needs its own shortcut.

The terminal is the tricky part, because almost every key is real input for the shell or the running program:

- <kbd>Escape</kbd> then <kbd>Tab</kbd> isn't as meaningless to a shell as it seems: in bash, <kbd>Esc</kbd> works as [Meta](https://www.gnu.org/software/bash/manual/html_node/Introduction-and-Notation.html), and Meta+<kbd>Tab</kbd> [completes from history](https://www.gnu.org/software/bash/manual/html_node/Commands-For-Completion.html).
- <kbd>Shift</kbd>+<kbd>Tab</kbd> is sent to the program, and some tools use it.
- <kbd>Ctrl</kbd>+<kbd>M</kbd> is Firefox's "Mute tab", and Firefox is the browser we ship in the ISO.
  <details>
  <summary>Click to show/hide partial screenshot of https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly </summary>
  
   ---  
  Image
   </details>
With all that, my new proposal is <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd> to move focus out of the terminal:

- **It's free in the terminal.** xterm.js sends nothing for <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+letter ([Keyboard.ts](https://github.com/xtermjs/xterm.js/blob/master/src/common/input/Keyboard.ts)), so intercepting it takes nothing away from users. This is also why desktop terminals put their own actions on <kbd>Ctrl</kbd>+<kbd>Shift</kbd>, e.g. copy with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>C</kbd>.
- **It's free in Firefox.** Of all the <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+letter combinations, Firefox only leaves F, L, and U unused (https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly). I discarded U because on Linux it's the usual way to type a character by its Unicode code (e.g. <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>U</kbd>, `e9`, <kbd>Space</kbd> gives é), so the input method may take it before the page does. I also discarded F because terminal users associate it with Find.
- **And, at least in English, the L can be seen as "leave"** 😛

Ideally, we'd go one step further. Leaving already moves focus to the terminal panel, but tabbing *into* it lands straight in the shell. Ace natively, and PatternFly by wrapping Monaco, put a stop before the editor instead, where <kbd>Enter</kbd> goes in. This isn't only what others do: it's an established accessibility pattern, similar to how the [W3C grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/) lets users press <kbd>Enter</kbd> to interact with a widget inside a cell and <kbd>Escape</kbd> to leave it. It works well for keyboard users, who could move past the terminal without ever getting into the shell, and for screen reader users, since that stop can announce how it works before anything reaches the shell.

However, I'm aware of the trade-off: we need a working terminal, but we also need to invest time in other parts of the application, and this has already taken quite a lot of time.

Thus, if you agree, we could replace <kbd>Escape</kbd> then <kbd>Tab</kbd> with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd> and keep everything else as it is: focus still moves to the terminal panel, and the text under the terminal would read "<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd> to move focus out". Since the terminal input already points to that text through `aria-describedby`, screen reader users will hear the new shortcut when landing in the terminal without any extra change. We'd also need to update the same claim in the `useTerminalSession` docs and in `doc/web-terminal.md`.

Let me know what you think, and sorry again for shipping a commit with misleading and under-researched information.

dgdavid: @imobachgs ☝️

doc/web-terminal.md
```diff
@@ -0,0 +1,235 @@
+# The web terminal
+
+Agama's web UI can open a real shell on the system being installed, running as root, right inside
+the browser. It exists for troubleshooting, working around something the installer does not expose,
+mointoring, etc.
```
joseivanlopez:
```suggestion
monitoring, etc.
```

doc/web-terminal.md
joseivanlopez: Great research. Thanks a lot. I vote for your plan: a) use `Ctrl`+`Shift`+`L` and b) create a follow-up for improving how to focus the the terminal.

rust/agama-server/src/terminal/web.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Control message sent to the client to report that the shell exited.
+///
+/// Exactly one of `code` or `signal` is set: `code` for a normal exit
+/// (`exit`, `exit N`, Ctrl-D — any way the shell terminates on its own,
+/// whatever the resulting number), `signal` when it was killed by an
+/// unhandled signal instead (a crash, an out-of-memory kill, an external
+/// `kill`, ...).
+#[derive(Debug, Serialize)]
+#[serde(tag = "type", rename_all = "camelCase")]
+enum ServerMessage {
```
joseivanlopez, line 87: Why is this a enum? In case of enum, I would expect 2 variants, for example `ExitCode`, `ExitSignal`.

imobachgs: To accomodate future messages. About using two variants, I considered that too, but the proposed solution was good enough. I can change It anyway.

doc/web-terminal.md
imobachgs: Your proposal looks good. But, please, create a separate issue for futher improvements. I do not want to get stuck in implementing this change.

dgdavid: > create a separate issue for futher improvements. I do not want to get stuck in implementing this change.

Probably I was not clear, but this is exactly my intention: just change the shortcut and leave it as it is, with a no priority follow-up issues for the rest.

rust/agama-server/src/terminal/web.rs
joseivanlopez: Ok, it was just out of curiosity :)

imobachgs: No, no, it was a good point. The change might be cheap. And, to be consistent, the ResizeMessage could be a variant of an enum too.

joseivanlopez:
The code looks good to me, as much as I can understand it :-). I was playing with the terminal and I have 2 questions. But none of them are blocking this PR, it is just a matter of curiosity or ideas for the future (if you think it really makes sense).

The first one is related to the fact that the terminal is killed with a page reload. It is basically the only element that disappears when your reload. Now, thanks to the info tracked in the URL (tabs, open menus, etc), we can properly recover the last position in the UI. And this brings me to the second idea.

Can we track the status of the terminal in the URL (open, minimized, size)? If so, maybe it helps to keep the terminal status after reloading. I know there are more challenges in the backend size in order to reconnect to the web socket. As said, it is only an idea, nothing to do now. 

BTW, image we can print the terminal as it was before reloading the page. Would it make sense to show it at least with the "Connection lost" message?

joseivanlopez (approved):
LGTM

Approving, although I guess you are going to add some small changes.

doc/web-terminal.md
joseivanlopez: @dgdavid I would suggest you to create this follow-up directly in our agama project at Github, for example in the backlog. Thanks!

dgdavid (approved):
Also LGTM 👍 but you might want to check and test the latest two commits I have added before merging. 

Thanks!

imobachgs:
> The code looks good to me, as much as I can understand it :-). I was playing with the terminal and I have 2 questions. But none of them are blocking this PR, it is just a matter of curiosity or ideas for the future (if you think it really makes sense).
> 
> The first one is related to the fact that the terminal is killed with a page reload. It is basically the only element that disappears when your reload. Now, thanks to the info tracked in the URL (tabs, open menus, etc), we can properly recover the last position in the UI. And this brings me to the second idea.
> 
> Can we track the status of the terminal in the URL (open, minimized, size)? If so, maybe it helps to keep the terminal status after reloading. I know there are more challenges in the backend size in order to reconnect to the web socket. As said, it is only an idea, nothing to do now. 

Hi,

Yes, we can. We need to implement some kind of persistence in the backend to keep the terminal running after the websocket is closed. It should not be that hard. We could create a follow up issue to implement It if we think it is worth It.

> BTW, image we can print the terminal as it was before reloading the page. Would it make sense to show it at least with the "Connection lost" message? 
> 

Well, if I reload the page, I assume that the connection is lost. But if we know that we are reloading, we can do that too.

Thank you for the review!

imobachgs:
> Also LGTM 👍 but you might want to check and test the latest two commits I have added before merging. 
> 
> Thanks!

Thanks for the review and the changes! I will do some small changes before merging, so we have plenty of time to do some more testing.

imobachgs:
@joseivanlopez @dgdavid I have changed a few things:

- Use an enum for client messages.
- Use two variants for server messages (`Exit` and `Killed`).

Please, check the update.

joseivanlopez (approved):

PR #3869, opened 2026-09-04
Fix changes file

imobachgs:
## Problem

#3864 introduce a changes entry after the top.

## Solution

Relocate the line.


mvidner:
The problem is much bigger, in agama-web-ui.changes too

PR #3870, opened 2026-09-04
Translation workflow files for the SLE-16.1 branch

lslezak:
## SLE-16.1 branch supports

- Automatically create pull requests for the updated SLE-16.1 translations
- The CI definition was actually copied from the 16.0 jobs and modified to handle the SLE-16.1 branch

## To do

I'm quite busy now so I took the simpler and quicker path, later we should do some refactoring and cleanup (share the code).

- [ ] Unify and share the tasks, avoid running the very same jobs just for different branches


joseivanlopez (approved):
LGTM

PR #3872, opened 2026-09-06
Convert the live/ tests to BATS format

lslezak:
## Problem

- We use an openSUSE Tumbleweed based container for running the Live ISO tests
- But for the SLE-16.1 branch we should use some something more stable and more close to SLE16.1 which is openSUSE Leap 16.1
- It turned out that we actually can run the test without any container

## Convert the live ISO tests to BATS format

- Use the [Bats](https://bats-core.readthedocs.io/en/stable/index.html) (Bash Automated Testing System) which is a tiny (~100kB!) framework designed for testing Bash scripts
- This removed the Ruby+RSpec dependency
- Bats is included in openSUSE Leap / Tumbleweed (`sudo zypper in bats`), the tests can be easily used locally
- Also the dependency to `agama-cli` package has been dropped, the `agama download` call is mocked. Now it possible to run the tests in plain Ubuntu CI VM without any container
- Additionally run the tests in `live/test/hcn` subdirectory (we have tests which we do not run regularly!? :scream: )
- Even after running additional tests the CI run is more than 2x faster! :+1: 

## Notes

- The unit tests are not used in the SLE-16.1 image, actually they are not even packaged to the source tarballs.
- That means it is safe to merge this change to the SLE-16.1 branch, it won't change anything from the OBS POV, the source package will stay untouched.
- I added few small fixes to the existing HCN tests:
  - Do not use fixed hardcoded directories in `/tmp` (problematic when running by multiple users, more over it is a security issue)
  - Properly fail the whole testsuite if a test fails

## Testing

- The updated CI workflow [is green](https://github.com/agama-project/agama/actions/runs/33962810633/job/101297547727#step:5:5).


jreidinger (approved):
looks good to me...can we run test also during build? so it verify also other archs, etc?

PR #3879, opened 2026-09-08
Disable local media 16.1

### LINKED ISSUE #3902: [QE][Build 65.2] Full repo was removed after installation
Bugzilla: [bsc#1277466](https://bugzilla.suse.com/show_bug.cgi?id=1277466 "‌")

The code that handles the software installation has been completely rewritten in 16.1. Initially, Agama did not remove any repository, but then we got [bug#1264277](https://bugzilla.suse.com/show_bug.cgi?id=1264277 "‌").

In Tumbleweed and openSUSE, the list of repositories is defined in a package, so we had duplicated packages. This behavior does not affect registered SLES systems, as you get the repositories through SCC services.

Checking Leap 16.0 behavior, Agama used to disable the local repository. I tried the offline medium in an x86_64 machine, so it is not 100% the same case.

So following Jiri suggestion, we could keep but disable the _removable installation repositories_. If the user define a custom repository, it should remain untouched.

Check what “removable” meant in 16.0.

---

## Review

- Pull Request: [https://github.com/agama-project/agama/pull/3879](https://github.com/agama-project/agama/pull/3879 "smartCard-inline")

jreidinger:
## Problem

Full medium repo is removed after installation which is not much user friendly in air gap scenario. 16.0 only disable such repository.

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1277466
trello: https://trello.com/c/pOJll4fa/5996-3-sles161-p2-1277466-qebuild-652-full-repo-was-removed-after-installation

## Solution

Disable local only repos instead of removal of them.

## Testing

- *Added a new unit test*

### DIFF
--- a/rust/agama-software/src/zypp_server.rs
+++ b/rust/agama-software/src/zypp_server.rs
@@ -689,11 +689,17 @@ impl ZyppServer {
     ///
     /// This covers the installation repositories aliased `agama-0`, `agama-1`,
     /// etc. (see `build_repo` in `model::state`) as well as the Driver Update
-    /// Disk repository.
+    /// Disk repository. The exception is local repositories, as they are only disabled
+    /// so it can be easily readd later for offline scenarios (bsc#1277466)
     fn remove_installation_only_repos(&self, zypp: &zypp_agama::Zypp) -> ZyppServerResult<()> {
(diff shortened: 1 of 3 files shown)


lslezak (approved):
LGTM

PR #3883, opened 2026-09-09
feat(web): enhance ports selection

teclator:
## Problem

The ports of a bond or a bridge are typed by hand into a free-text field, whose only help is a line listing the device names: *"Available devices: enp1s0 and enp2s0"*. Getting `enp197s0f3u1u1` right is on the user, a typo is silently accepted, and the list does not scale past a handful of NICs.

The VLAN parent field already solves the picking half of this, with a dialog showing each device with its details. Ports deserve the same, except that a bond or a bridge takes several of them.

- https://trello.com/c/loaOB90k/
- related to https://bugzilla.suse.com/show_bug.cgi?id=1279417

## Solution

Build the field on the multi-select field added in #3927: the devices found in the system are offered in a searchable list, and a name can still be written out for a device that is not there yet. The line listing the device names gives way to a helper text naming those two ways of filling the field in.

- Each option is the device name with its hardware address underneath, and no more. Everything in an option is read out with it, so the details that would make a row long to hear are left to the dialog. The list is searched by the same two things it shows.
- The foot of the list leads to a dialog listing the same devices in a table, for the details a list cannot hold and for picking several at once. It opens with the devices already listed as ports picked, so it serves for dropping a port as much as for adding one. Confirming with nothing picked drops them all.
- The dialog shows two of the details the API reports since #3884: the driver under the device type, and a **Link** column with the speed, or *"No link"* for a card with no cable in it. Both, like the "Used by" column before them, only appear when some device has something to put in them. The column sorts the way it reads rather than on the raw speed, since an unplugged card may still report the speed it last ran at. The addresses, the widest column and the one saying least about which card a row is, come last.
- Devices another bond or bridge already uses are marked in the "Used by" column of the dialog and still offered, since moving a port from one controller to another is legitimate.
- The field settles down once it is left alone: past three ports the rest is counted rather than listed, and the text box and the button emptying the field are only there while the field has focus. Everything comes back on focus or on a click anywhere in the control.
- Names are still typed and pasted freely, since a port may be a device that is not plugged in yet, or one that only shows up once the installed system boots. Names the dialog knows nothing about are left untouched by what is picked in it, because it never asked about them.
- Neither the loopback device nor the bond or bridge being configured are offered, and a name matching the latter, which only a typed value can be, is refused by `validatePorts` when the form is submitted: a device cannot be a port of itself.

Master is merged in, which is what made the field and the details possible. The picker this branch had built in the meantime, a button beside the field and the `addOn` slot in `ArrayField` that held it, is dropped in favour of the field and its footer entry, so there is no second tab stop next to the input.

Two things outside the ports field proper:

- The web `Device` type gains the `speed`, `carrier`, `driver`, `bus_path` and `mtu` fields that #3884 added on the server side. Only the first four are shown here; `ConnectionDetails` is the natural home for the rest, and is left for a follow-up.
- `formatNumber` in `i18n.ts` wraps `Intl.NumberFormat` around `agama.language`, as `formatList` already does for `Intl.ListFormat`. Without it, "2.5 Gb/s" followed the browser's locale rather than the installer's.

## Testing

- Unit tests for `PortsField` covering the list, the names written out by hand, the trip through the dialog and the self-port check; those of `DeviceSelectorModal` extended for the new columns and their sorting; new ones for `formatLinkSpeed`, `deviceLinkLabel`, `deviceLinkRank`, `formatNumber` and the ports validation.
- `npm run check-types`, `npm run eslint` and `jest src/` are clean.
- A manual pass over the new field is still pending.

## Screenshots

<details>
<summary><b>📷 Picking the ports of a bond (expand to view)</b></summary>
<br />

**1. The field with the list open** — the options offering the devices found in the system, each with its hardware address, and *"Browse with details..."* at the foot of the list.

_(screenshot)_

**2. Searching the list** — the text typed narrows it down, by device name or by hardware address.

_(screenshot)_

**3. The dialog behind the foot of the list** — the same devices in a table, with the driver under the type, the **Link** column, the state, which bond or bridge already uses each device, and the addresses last. Several devices picked.

_(screenshot)_

**4. The ports the dialog wrote back** — the field listing what was picked.

_(screenshot)_

**5. The field at rest with more than three ports** — the rest counted rather than listed, with no text box and no clear button until the field is focused again.

_(screenshot)_

**6. A name written out for a device the system does not report** — taken as it is, and kept whatever the dialog is used for afterwards.

_(screenshot)_

**7. The bond named among its own ports** — refused on submit, with the port it is talking about marked.

_(screenshot)_

</details>


web/src/components/form/ArrayField.tsx
```diff
@@ -682,98 +701,152 @@ export default function ArrayField({
...
+  };
+
+  /**
+   * Replaces the whole list, the way an add-on picker showing what is already
+   * listed does.
+   *
+   * Values go through the same normalization and duplicate rules as a typed
+   * entry. What is announced is the resulting list rather than what changed,
+   * since entries may just as well have been dropped. The draft is left alone:
+   * the caller decides whether the text the user was typing is still relevant.
```
dgdavid: NP: I have read mostly the same doc three times already. Wondering if it could be specified only at one place.

dgdavid: Also, wonder if it works as expected: the picker not deleting something it does not know about.

web/src/components/network/connection-form/PortsSelector.tsx
```diff
@@ -0,0 +1,195 @@
...
+  } as PortsSelectorProps,
+  render: function Render({ form, kind }) {
+    const isBond = kind === "bond";
+    const name = isBond ? "bondPorts" : "bridgePorts";
+    const controllerField = isBond ? "bondIface" : "bridgeIface";
+    // TRANSLATORS: label for the bond or bridge ports field.
+    const label = isBond ? _("Bond ports") : _("Bridge ports");
+    // TRANSLATORS: accessible name of the button opening the dialog for picking
+    // the ports of a bond or a bridge among the devices found in the system.
+    const pickLabel = isBond ? _("Select bond ports") : _("Select bridge ports");
```
dgdavid: I'm not sure this is the right approach. It couples that "PortSelector" thingy to bond and bridge, when it could be less coupled for tomorrow needs if all of this were defined by the caller. By all of this I mean its name, its label, its picker label. 

What is More, PortSelector is not a good name according with form/conventions. PortField it is. Here you are not rendering just the selector, this the internal PortPicker (that I would name DevicePicker or InterfacePicker or simply Picker).

web/src/components/network/connection-form/PortsSelector.tsx
```diff
@@ -0,0 +1,195 @@
...
+  // Devices already used elsewhere are still offered, since moving a port from
+  // one controller to another is legitimate. The dialog only grows a column
+  // for it when there is something to tell.
+  const hasPortsInUse = devices.some((d) => portOf(d));
+
+  return (
+    <>
+      <VisualTooltip content={label}>
+        <Button
+          variant="control"
```
dgdavid: I believe `plain` variant would fit better for this control. Worth testing it.

web/src/components/network/connection-form/PortsSelector.tsx
dgdavid: Even further, the whole ArrayField + "Addon" combination should be a generic avaialble Field at web/src/components/form with zero knowledge about entries but providing the resuable behavior of a form needing a multi-entry values with a picker. It's doable, I think.

web/src/components/form/ArrayField.tsx
teclator: It is questionable. It allows to select and deselect from the devices known leaving the unknown to be handled by the user. We could append the unknown ones to the known devices in order to manage all from the dialog although there will be no info for them.

dgdavid: I don't think so. The dialog should not contain these values. Remember this is a special feature. Users still being able to delete these from the input.

Let's talk about it. But let's not make things more complex than needed.

dgdavid: Also remember, this is an input text **not** a dropdown. The so called until now _addon_ is an aid to pick among existent stuff, nothing else.

dgdavid: And it is something for which a different helper text can help. A wording problem.

**Enter device names or select from available devices.**

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,195 @@
...
+                const controller = controllerOf(device.name, connections);
+                return controller === controllerIface ? undefined : controller;
+              };
+
+              return (
+                <field.ArrayField
+                  label={label}
+                  helperText={
+                    // TRANSLATORS: helper text for the bond or bridge ports field.
+                    _("Pick the devices to use as ports, or type the name of one not listed yet.")
```
dgdavid: As discussed yesterday in a meeting, I think this is misleading because, besides being a rather complex helper text describing a feature that most users won’t need,  _not listed yet_ suggests that devices can be added to the list later. **The picker is only there as an aid to select devices the system already knows about**, while the input also allows any device name.

Then, I would go for something simpler and easy to parse


```suggestion
                    _("Enter device names or select from available devices.")
```

web/src/components/form/ArrayField.tsx
dgdavid: Check https://github.com/agama-project/agama/pull/3883/changes#r3976643621

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -91,38 +120,45 @@ export default function DeviceSelectorModal({
...
         <Stack>
           <span>{device.name}</span>
-          <Text textStyle={["textColorSubtle", "fontSizeXs"]}>{device.macAddress}</Text>
+          <Text textStyle={["textColorSubtle", "fontSizeXs"]}>
+            {isPresent(device)
+              ? device.macAddress
+              : // TRANSLATORS: said of a network device named in the
+                // configuration although the system does not report it, e.g. a
+                // bond port typed by hand or a card plugged in later.
+                _("Not present yet")}
```
dgdavid: As said, I do not agree with that. I have the feeling we are mixing things here to make more visible a feature that will be used sporadically. This needs more discussion, sorry.

dgdavid (changes requested):

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: Read https://github.com/agama-project/agama/pull/3883#discussion_r3976303124 and https://github.com/agama-project/agama/pull/3883#discussion_r3976643621

web/src/components/form/ArrayField.tsx
teclator: That was original implementation, the unknown was not touched and only the known where seleted and handled by the picker.

web/src/components/network/connection-form/PortsSelector.tsx
teclator: Agreed it's doable, and I think that's the right end state. I'd rather not build it in this PR though.

Half of it is already there: `ArrayField` lives in `components/form` and knows nothing about entries. Its `addOn` hands the caller the current `entries` and a `setEntries` that replaces them, and `mergePicked` (also in `ArrayField`) carries the one rule that isn't caller-specific — a picker must not drop what it never offered.

What's left to generalize is the picker itself, and that's where I get stuck on value: a generic one needs the item type, the identity key, the row renderer and the labels as parameters, which is `DeviceSelectorModal` rebuilt to serve a single caller. I'd be guessing at the seams rather than seeing them.

So I'd wait for the second picker add-on — the storage device lists look like the likely one — and factor it out then, when there's a real second shape to factor against. That also keeps this PR in line with what you said further down the thread: the add-on is an aid to pick among existing stuff, nothing else.

I've written the pattern down in `components/form/conventions.md` (pattern 8, "Add-on beside a field that takes typed values") so the next one has the rules to follow instead of rediscovering them.

teclator:
@dgdavid ready for another look. What moved since your review:

**The dialog no longer touches what it does not offer.** I had tried appending the unknown names as stand-in rows so everything could be managed from one place; that's out. `mergePicked(entries, offered, picked)` in `ArrayField` carries the rule now: a name the dialog was never given is kept whatever the user does in there. Deleting it stays a job for the input, as you said.

**Renamed after the conventions.** `PortsSelector` → `PortsField`, and the inner one is `DevicePicker`. The field takes `name`, `controllerField`, `label` and `pickLabel` from the caller, so it knows nothing about bonds or bridges — `BondFields` and `BridgeFields` just pass different labels.

**Helper text is your wording**, verbatim: *"Enter device names or select from available devices."* Mine implied devices could be added to the list later, which was wrong.

**Trimmed the repeated JSDoc** in `ArrayField` you'd read three times.

On the generic `ArrayField` + picker Field: I answered in the thread — I think it's the right end state but I'd rather factor it against a real second caller than guess at the seams now. The rules are written down in `components/form/conventions.md` (pattern 8) meanwhile, including why the stand-in rows were dropped.

One thing worth your eyes because it came from me and not from the ticket: **confirming the dialog with nothing picked clears the list.** The button reads "Use no device" in that state, so it isn't a silent wipe, and it seemed wrong to make the dialog able to add and drop ports but not the last one. Happy to drop it if you'd rather the dialog never emptied the field.

The screenshots in the description predate all of this — the helper text and button they show are stale.

abonillabeeche:
Nice work on the device picker — the "Used by" column especially solves a real pain point. While testing this locally I noticed the `carrier`/`speed` fields from #3884 aren't consumed yet in `DeviceSelectorModal`, so link state and speed (the two things hardest to get without opening every connection's Details) still aren't shown. Opened #3910 on top of this branch to close that gap: adds a `deviceLinkLabel()` helper and a "Link" column next to State ("Down" / "Up" / "Up (%d Mb/s)" / "Unknown" when the device reports nothing).

teclator:
Thanks @abonillabeeche, and thanks for taking the time to build #3910 rather than just filing the gap — that's the more useful kind of review.

Some context on the sequencing, so it doesn't look like an oversight: the extra device details and the picker that consumes them were originally written together in one branch, and split into two PRs on purpose to keep each reviewable. #3884 is the backend half — it's mine, and it's still open. The consumer side, "Link"/speed columns included, was always meant to land right after it, once #3884 was merged and the fields were actually there to read. Consuming them from this PR would have meant merging a frontend that depends on an API that hasn't been agreed on yet.

Original implementation before the split into two PRs


The design itself also has some history worth pointing at: browsing the devices with their details, and picking from that same dialog, was discussed and mocked by @dgdavid in #3822 back in June. The mock-ups in that description are pretty much where this is heading:

Device dialog mock-up from #3822, showing the extra columns

So none of this is a disagreement about *what* to show — we want the same thing. It's only about ordering. Rather than have two PRs doing the same work, shall we compare notes? If your `deviceLinkLabel()` and the column layout are further along than what I had queued, I'd sooner take yours and drop mine; the main thing to settle is that it lands on top of #3884 once the field names there are final, since they may still change in review.

One detail from #3822 worth carrying over either way: the columns should stay quiet when there's nothing to say. This PR only grows the "Used by" column when some device actually is a port of something, and I'd keep "Link" to the same rule.

web/src/components/form/multi-select-field/MultiSelectField.tsx
```diff
@@ -901,7 +901,10 @@ export default function MultiSelectField({
           </HelperTextItem>
           {hasAnyError && (
             <HelperTextItem variant="error">
-              {sift([...fieldErrors, ...entryErrors]).join(". ")}
+              {/* A rule refusing the whole field and the mark pointing at the
+                  value that broke it are often worded the same, and hearing
+                  the same sentence twice says nothing the first one did not. */}
+              {unique(sift([...fieldErrors, ...entryErrors])).join(". ")}
```
dgdavid: Is there chances to have duplicated fieldErrors? why? Do you have an example?

web/src/components/network/connection-form/BondFields.test.tsx
```diff
@@ -78,19 +79,21 @@ describe("BondFields", () => {
...
     await user.type(input, "enp1s0{enter}");
     await user.type(input, "br0{enter}");
 
-    expect(await screen.findByText("enp1s0")).toBeInTheDocument();
-    expect(await screen.findByText("br0")).toBeInTheDocument();
+    // Looked up among the committed values: the list of devices the field
+    // offers is still open behind them, naming the same devices.
+    const entries = within(screen.getByRole("listbox", { name: "Bond ports entries" }));
+    entries.getByRole("option", { name: "enp1s0" });
+    entries.getByRole("option", { name: "br0" });
```
dgdavid: I wonder if former screen.findByText didn't work and why. If it works, I would keep it since looks to me near to what a user does "find by scanning text"

web/src/components/network/connection-form/BondFields.tsx
```diff
@@ -21,12 +21,11 @@
  */
 
 import React from "react";
-import { sprintf } from "sprintf-js";
+import PortsField from "./PortsField";
```
dgdavid, line 24: In storage there is a shared/ directory to put the "common" fields. In network we only have the connection-form right now. So it is ok this living there, but I will keep an eye to move it to shared place as soon as it is consumed by other form, if any new arrive in the future.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { useConnections, useDevices } from "~/hooks/model/system/network";
+import { connectionTypeLabel, controllerOf } from "~/utils/network";
+import { _ } from "~/i18n";
+
+import type { TranslatedString } from "~/i18n";
+import type { FormFields } from "./fields";
+import type { Device } from "~/types/network";
+
+/** The loopback device is never a port of anything. */
+const LOOPBACK = "lo";
```
dgdavid: I would say we have used LOOPBACK thingy in other places too. If so, I would put it available in a single place every component/util can consume instead of repeating it all the time.

```zsh
❯ rg "LOOPBACK" web 
web/src/utils/network.ts
50:  LOOPBACK: "loopback",
116:  [CONNECTION_TYPE.LOOPBACK]: N_("Loopback"),

web/src/types/network.ts
131:  LOOPBACK = 0,

web/src/types/network.test.ts
33:  it("returns false for ETHERNET, WIFI, and LOOPBACK types", () => {
36:    expect(isVirtual(CONNECTION_TYPE.LOOPBACK)).toBe(false);

web/src/components/network/connection-form/VlanFields.tsx
95:            types: [CONNECTION_TYPE.VLAN, CONNECTION_TYPE.LOOPBACK],

web/src/components/network/connection-form/VlanFields.test.tsx
49:  type: CONNECTION_TYPE.LOOPBACK,

web/src/components/network/connection-form/Form.test.tsx
44:  type: CONNECTION_TYPE.LOOPBACK,
```

dgdavid: Any chances of unifying all these? Something that can query all of them from a single place? I mean, The "lo", the "0" and the "loopback"?

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+/** Names of the given devices. */
+const names = (devices: Device[]): string[] => devices.map((d) => d.name);
+
+/**
+ * Ports left after a dialog offering `offered` answered with `picked`.
```
dgdavid: NP:  A bit hard to parse

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+  label: TranslatedString;
+  /**
+   * Title of the dialog listing the devices with their details, e.g. "Select
+   * bond ports".
+   *
+   * The field label that said what the devices are for is no longer in sight
+   * once the dialog covers the form, and the entry that opened it reads the
+   * same on every field, so the dialog cannot name itself.
+   */
+  title: TranslatedString;
```
dgdavid: NP: I would use a more descriptive name for this prop. When I was reading the call to this fields I asked myself what the title is for in a field. Now I got it, but its DX might be improved.

web/src/components/network/connection-form/VlanFields.tsx
```diff
@@ -90,6 +90,11 @@ const VlanFields = withForm({
             // TRANSLATORS: label for the VLAN parent device field.
             _("Parent device")
           }
+          title={
```
dgdavid: NP: the same here, a better prop name would be nice.

web/src/utils/network.ts
```diff
@@ -99,6 +100,66 @@ const DEVICE_STATE_LABELS: Record<DeviceState, MarkedString> = {
...
+ *
+ * A rate of a gigabit or more is given in Gb/s, the unit such cards are sold
+ * and talked about in ("10 Gb/s", not "10000 Mb/s"), keeping a decimal for the
+ * rates that fall between two whole gigabits, e.g. "2.5 Gb/s".
+ */
+const formatLinkSpeed = (mbps: number): TranslatedString => {
+  if (mbps < 1000) {
+    // TRANSLATORS: link speed of a network device in megabits per second. %d is
+    // replaced by the number, e.g. "100 Mb/s".
+    return sprintf(_("%d Mb/s"), mbps) as TranslatedString;
```
dgdavid: I would say sprintf return a TranslatedString

Image

Then, why the "as TranslatedString"

web/src/utils/network.ts
```diff
@@ -99,6 +100,66 @@ const DEVICE_STATE_LABELS: Record<DeviceState, MarkedString> = {
...
+  }
+
+  // TRANSLATORS: link speed of a network device in gigabits per second. %s is
+  // replaced by the number, e.g. "2.5 Gb/s".
+  return sprintf(_("%s Gb/s"), formatNumber(mbps / 1000)) as TranslatedString;
+};
```
dgdavid, line 123: I wonder why xbytes was not an option for this https://github.com/miraclx/xbytes#examples

Just becuase the "/s" part?

web/src/utils/network.ts
```diff
@@ -99,6 +100,66 @@ const DEVICE_STATE_LABELS: Record<DeviceState, MarkedString> = {
  */
 const deviceStateLabel = (state: DeviceState): TranslatedString => _(DEVICE_STATE_LABELS[state]);
 
+/**
+ * Formats a link speed given in Mb/s.
```
dgdavid: NP: nothing ensures the speed is given in Mb/s. The param, although called mbps, it is just a number.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    options={options}
+                    allowCustomEntries
+                    // A bond or a bridge over a handful of ports is common
+                    // enough that the field would otherwise be a wall of
+                    // names; past three the rest is counted instead, and
+                    // comes back whenever the field has focus.
+                    entriesThreshold={3}
+                    // At rest the ports are all the field has to say, so
+                    // neither the text box nor the button emptying it is kept
+                    // on screen. Both come back on focus.
```
dgdavid: I believe this is actually part of the MultiSelectField documentation.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+              return (
+                <>
+                  <field.MultiSelectField
+                    label={label}
+                    options={options}
+                    allowCustomEntries
+                    // A bond or a bridge over a handful of ports is common
+                    // enough that the field would otherwise be a wall of
+                    // names; past three the rest is counted instead, and
+                    // comes back whenever the field has focus.
```
dgdavid: It is ok to explain why, but maybe it is not fully needed such a documentation here since the prop is  (or should be) properly documented in the MultiSelectField.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    entriesThreshold={3}
+                    // At rest the ports are all the field has to say, so
+                    // neither the text box nor the button emptying it is kept
+                    // on screen. Both come back on focus.
+                    collapseInputOnBlur
+                    hideClearAllOnBlur
+                    helperText={
+                      // TRANSLATORS: helper text for the ports field of a bond
+                      // or a bridge, naming the two ways of filling it in.
+                      _("Choose the devices found in the system, or enter other names.")
```
dgdavid: No, please. Use a shorter helper text. What about the one I used for the MultiSelectField PR, "Choose devices or enter their names"


```suggestion
                      _("Choose devices or enter their names.")
```


Image


@ancorgs , what's your view?

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    // which of the ports it was talking about, and stays quiet
+                    // until then, as validation does everywhere else.
+                    validateOnSubmit={(value) =>
+                      value === controllerIface
+                        ? // TRANSLATORS: error shown on a port naming the bond
+                          // or bridge being configured. %s is the name of that
+                          // device, e.g. "bond0".
+                          sprintf(_("%s cannot be a port of itself"), value)
+                        : undefined
+                    }
```
dgdavid: I would say it is the first time I see a validateOnSubmit in a field itself. If if remember correctly, this is responsability of the form through validations.ts, right?

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                            // leading to a dialog listing the same devices with
+                            // more information about each of them. The trailing
+                            // ellipsis hints that more follows instead of a
+                            // device being picked right away.
+                            label: _("Browse with details..."),
+                            onSelect: () => setIsDialogOpen(true),
+                            opensDialog: true,
+                          }
+                        : undefined
+                    }
```
dgdavid, line 170: Is that possible? Facing a dropdown without entries here? If so, we should add some hints to users in dropdowns too.. which are supposed to always provide something. Do you think this is a real scenario?

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                      available.length > 0
+                        ? {
+                            // TRANSLATORS: entry at the end of the device list,
+                            // leading to a dialog listing the same devices with
+                            // more information about each of them. The trailing
+                            // ellipsis hints that more follows instead of a
+                            // device being picked right away.
+                            label: _("Browse with details..."),
+                            onSelect: () => setIsDialogOpen(true),
+                            opensDialog: true,
```
dgdavid, line 167: This one is my fault! I leak such a poor prop name in the MultiSelectField PR. It should be rename to something easy to guess while reading, `srHint` or so. I would take care of that.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+                // The type and the driver are searchable although the option
+                // does not show them: the dialog does, and a user who knows
+                // which driver their card runs on should not have to open it to
+                // find out which device that is.
+                filterText: sift([
+                  device.name,
+                  device.macAddress,
+                  connectionTypeLabel(device.type),
+                  device.driver,
+                ]).join(" "),
```
dgdavid: Are you sure about that? I'm not. Letting users to filter by something it is not in the screen is not a good pattern IMHO.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+              const hasPortsInUse = available.some((d) => portOf(d));
+
+              const options = available.map((device) => ({
+                value: device.name,
+                label: device.name,
+                // The hardware identifier, and nothing else: every word here is
+                // read out with the option, so a row telling the whole story
+                // takes longer to hear than the list takes to read. What tells
+                // two cards apart is one entry away, in the table behind the
+                // foot of the list.
```
dgdavid: Thanks for clarifying it here. I actually think that it is pretty undertandable even without comment, but just in case someone is tempted to add something in the future. I would change it a bit, though 


```suggestion
            // Keep this to the hardware identifier: putting all details in
            // the dropdown currently creates an accessibility problem, as
            // too much information is harder to follow when read aloud. We
            // also want to keep the common case simple. For details, the
            // component already offers "Browser with details..."
```

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -149,11 +234,10 @@ export default function DeviceSelectorModal({
     <Popup
       isOpen
       variant="medium"
-      // TRANSLATORS: title of the dialog for picking a network device
-      title={_("Select a network device")}
+      title={title}
       // Focus starts on the picked device, so its row is what the user hears
       // and sees first, and the arrow keys move from there.
-      elementToFocus={initialDevice ? "input[type=radio]:checked" : undefined}
+      elementToFocus={pick ? "input:checked" : undefined}
```
dgdavid, line 247: I need more time to have a look to this by manually testing, since I do not understand right not its purpose.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -86,19 +140,52 @@ export default function DeviceSelectorModal({
...
+            // TRANSLATORS: table column telling which bond or bridge already
+            // uses a network device as one of its ports.
+            name: _("Used by"),
+            value: (device: Device) => portOf(device) || "-",
+          },
+        ]
+      : []),
+    // Last, and on purpose: it is the widest column and the one that says the
+    // least about which card a row is, so it does not come between the columns
+    // that do.
```
dgdavid: NP: It is ok to me, but do actually need such an explanation?

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -51,26 +81,50 @@ const deviceAddresses = (device: Device): string =>
...
+    if (isMultiple) return [];
+    const firstDevice = first(devices);
+    return firstDevice ? [firstDevice] : [];
+  };
+  const [selection, setSelection] = useState<Device[]>(selected ?? defaultSelection());
+
+  // The link is best-effort: a device reports it or it does not, and a column
+  // no device can fill is a column of dashes. It is only added when at least
+  // one device has something to say, the same way "Used by" is.
+  const hasLink = devices.some((device) => deviceLinkLabel(device));
```
dgdavid: I tend to not like this kind of behavior, displaying a column conditionally in that way. Then users could get confused trying to guess what they have to do to see a column they saw before but not later.

Now we need someone to tip the scale, @ancorgs please have a look.

Also, please do not uses "-" meaning nothing here. Simple leave the cell empty. Empty means empty. BTW, with this approach it could still being "a column of dashes" when just one device has such a link.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -51,26 +81,50 @@ const deviceAddresses = (device: Device): string =>
...
+ * rather than devices, such as the ports of a bond, may well hold a name no
+ * device answers to; the dialog says nothing about it and the caller keeps it
+ * (see `mergePicked` in `PortsField`).
  */
 export default function DeviceSelectorModal({
+  title,
   devices,
   selected,
+  selectionMode = "single",
+  portOf,
```
dgdavid, line 112: Why make this a prop of the dialog? I see the point, but I believe it should not be a prop of the dialog but something else. No idea at this moment, sorry, just the feeling it is misplaced. I will review it again tomorrow with fresh mind.

web/src/components/network/connection-form/PortsField.tsx
```diff
@@ -0,0 +1,231 @@
...
+type PortsFieldProps = {
+  /** Form field holding the names of the ports. */
+  name: Extract<keyof FormFields, `${string}Ports`>;
+  /**
+   * Form field holding the name of the controller the ports belong to.
+   *
+   * Watched rather than read once: the field is filled while the form is being
+   * used, and a controller cannot be a port of itself.
+   */
+  controllerField: Extract<keyof FormFields, `${string}Iface`>;
```
dgdavid, line 62: Which is the purpose of thi? Are you triggering validations before submit?

dgdavid (changes requested):
I have done a first review to not keep you blocked with this. I will do another review with a fresh mind when possible.  Despite the comments, thanks a lot for driving this one in such an agile way.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -76,9 +76,31 @@ export type DeviceSelectorModalProps = {
...
+ * Addresses of a device, one per line.
+ *
+ * An address is a single unbreakable word, and an IPv6 one is long enough to
+ * outgrow the column it sits in: run two of them together on one line and the
+ * cell cuts the second one off mid-prefix. A line each, and leave to breaking
+ * the address the cases where even that is not enough.
+ */
+const deviceAddresses = (device: Device): React.ReactNode => {
+  const addresses = device.addresses || [];
+  if (addresses.length === 0) return "-";
```
dgdavid: No please, no "-" for meaning empty. Just leave the cell empty.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -76,9 +76,31 @@ export type DeviceSelectorModalProps = {
...
+    <Stack>
+      {addresses.map((address) => {
+        const text = formatIp(address);
+        return (
+          <Text key={text} textStyle="textBreakWord">
+            {text}
+          </Text>
+        );
+      })}
+    </Stack>
```
dgdavid: I'm not fully sure to want display ALL addresses, one per line, in a table. I t could be a nightmare one one device belongs to more than 3 or 4. But go ahead with this right not opening a follow-up to find s different way to deal with this. Evaluating to hide them behind a "N more" link, which needs a good keyboard navigation support. Also we have in mind, in the future, to allow users choose which columns to see. But all of these are follow-ups, not belongs to this PR.

What I would like to yo do and manually test here is to keep using this breakword thingy but to drop the stack and keeping the join to let the UI decide how many fit in a line to save some space.. Somthing like

```ts
<Text textStyle="breakWord">{formatList(
  (deviceAddress || []).map((address) => formatIp(address))
)}</Text>
```

Note that this example also uses `src/i18n#formatList` instead of `join(", ")`.

It should fix the _problem_ of ipv6 addresses cut without requiring a line per IP. If so, I would got with this until we can research and release another round of improvements for columns like this.

dgdavid (changes requested):
Remember, an empty cell is just empty. No "-" needed. Also, I would not request a line per IP without first test a less _aggressive_ approach. I put an example (not tested)

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: This apply to all columns.

web/src/components/form/multi-select-field/MultiSelectField.tsx
teclator: Only through the ports field, which checked a port against its own controller on top of \`validatePorts\` doing the same, so the two said "br0 cannot be a port of itself" one after the other. That per-value check is gone now (see the thread below), and with it the only way to produce a duplicate, so I have reverted this.

web/src/components/network/connection-form/BondFields.test.tsx
teclator: It did not, and the reason was mine to fix: the list of devices is still open behind the committed values, naming the same devices, so the text is there twice. Closing it first brings `findByText` back:

```js
await user.type(input, "enp1s0{enter}");
await user.type(input, "br0{enter}");
await user.keyboard("{Escape}");

await screen.findByText("enp1s0");
await screen.findByText("br0");
```

web/src/components/network/connection-form/BondFields.tsx
teclator: Agreed. `connection-form/` is the only consumer today; the day a second form wants it, a `shared/` next to it is the right move.

web/src/components/network/connection-form/PortsField.tsx
teclator: Dropped altogether: the filter now goes through `CONNECTION_TYPE.LOOPBACK`, which is what `VlanFields` already excludes by, so there is no "lo" of this field own.

```js
const available = devices.filter(
  (d) => d.type !== CONNECTION_TYPE.LOOPBACK && d.name !== controllerIface,
);
```

teclator: They are three different things wearing the same word: `"lo"` is an interface name the kernel happens to give it, `0` is NetworkManager device type, and `"loopback"` is our own connection type. The first is the one that does not belong anywhere, and it is gone. Unifying the other two means deciding whether `CONNECTION_TYPE` and the `DeviceType` enum are one thing, which is a job of its own, but I agree it is worth doing.

web/src/components/network/connection-form/PortsField.tsx
teclator: Rewritten:

```
Ports left after the dialog answered with `picked`.

Only the `offered` ports follow the pick. The dialog was never told about
the others, names the user wrote for devices the system does not report, so
it has nothing to say about them and they stay.
```

web/src/components/network/connection-form/PortsField.tsx
teclator: Renamed to `dialogTitle`, here and in the callers.

web/src/components/network/connection-form/VlanFields.tsx
teclator: Renamed in `DeviceSelector` too, so both fields take `dialogTitle`.

web/src/utils/network.ts
teclator: You are right, and it is our own doing: `src/sprintf.d.ts` redeclares `sprintf` as `<T extends string | TranslatedString>(format: T, ...) => T`. Both casts removed.

web/src/utils/network.ts
teclator: It formats bytes, and these are bits per second:

```js
> xbytes(1000, { bits: true })
"8.00 Kb"        // it read 1000 as bytes and multiplied by 8
> xbytes(125000000, { bits: true })
"1.00 Gb"        // so we would feed it mbps * 1e6 / 8 to get bits back
```

On top of the round trip, the number comes out with a hardcoded decimal point and two fixed decimals ("1.00 Gb", not "1 Gb"), and the decimal separator is the whole reason `formatNumber` exists: a browser in English while the installer runs in German would write 2.5 where every other number on the screen says 2,5. The `/s` would be ours anyway.

web/src/utils/network.ts
teclator: True, it is just a number. Written down in the doc where it can be read:

```
The number is taken to be in Mb/s, the unit the kernel and NetworkManager
report a link speed in, which is also what `Device.speed` holds.
```

web/src/components/network/connection-form/PortsField.tsx
teclator: Removed. Same for the one above it.

web/src/components/network/connection-form/PortsField.tsx
teclator: Removed as well: the props are documented in `MultiSelectField`, and repeating the reasoning at each call site only gives it two places to go stale.

web/src/components/network/connection-form/PortsField.tsx
teclator: Taken as suggested: `_("Choose devices or enter their names.")`.

web/src/components/network/connection-form/PortsField.tsx
teclator: Right, and it is gone. `validatePorts` in `validations.ts` is the only thing checking it now.

What the field added was a mark on the offending value, so the user could see which port the message meant. Not worth a second place where validation lives, and it is what made the message read twice (the thread in `MultiSelectField`).

web/src/components/network/connection-form/PortsField.tsx
teclator: It is possible, just rare: a machine whose only network device is the loopback still reaches this form, and then the list has nothing but the values the user types. I kept the guard because a dialog opening on an empty table is a worse landing than no entry at all, but I have no strong feelings, and if the rule is that a dropdown always offers something then this should be an empty-state hint in `MultiSelectField` rather than a caller deciding. Happy either way.

web/src/components/network/connection-form/PortsField.tsx
teclator: No problem. We only pass `opensDialog` here, so whatever you rename it to I will follow.

web/src/components/network/connection-form/PortsField.tsx
teclator: Fair enough, and changed: `filterText` is now the name and the MAC address, which is exactly what the option shows.

web/src/components/network/connection-form/PortsField.tsx
teclator: Taken almost word for word, thanks.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: It only moves the initial focus: the dialog opens with the caret on the row that is already picked, instead of on the first thing in the modal, so a keyboard user hears where they stand and the arrow keys carry on from there. The change from the previous line is that the multiple case uses checkboxes rather than radios, hence `input:checked` instead of `input[type=radio]:checked`, and `pick` is just "the first picked device, if any" (nothing is picked when a dialog for several opens empty, and then focus falls back to the dialog).

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Cut down to one line.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: The dashes are gone, in this column and in every other one: an empty cell now means nothing to say.

On the conditional column itself I will wait for @ancorgs. For what it is worth, the two cases differ: "Used by" appears only when the caller passes `portOf`, so it is per form and stable while the dialog is open, whereas "Link" depends on the devices in front of the user. Both stay put for as long as the dialog is up, so a column never comes and goes under the user hands, but your point about the same dialog looking different on two machines stands.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Left as it is for now, waiting for your second look. The reason it ended up here: the dialog cannot work out which controller already uses a device, that is the form business (it depends on the connections and on the controller being edited), and it is also what decides whether the column is worth drawing at all.

web/src/components/network/connection-form/PortsField.tsx
teclator: No, nothing is validated before submit. The field watches the controller name so the list of offered devices keeps up with it: type `br0` in the device name field and `br0` stops being offered as one of its own ports. Refusing a name typed by hand is `validatePorts` job, on submit.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Done, here and everywhere else in the table.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Went with your version, and it does fix it. One line, wrapped where the column ends, `formatList` instead of `join(", ")`:

```jsx
const deviceAddresses = (device: Device): React.ReactNode => (
  <Text textStyle="textBreakWord">
    {formatList((device.addresses || []).map((address) => formatIp(address)))}
  </Text>
);
```

I will open the follow-up for the "N more" idea and the column chooser.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Applied to all of them: the Link column and "Used by" no longer fall back to a dash either.

web/src/components/network/connection-form/BondFields.test.tsx
dgdavid: Great! This is closer what a user does. Thanks.

web/src/components/network/connection-form/PortsField.tsx
dgdavid: And later in validation, how do you manage user didn't enter "lo" by hand?

Does worth having a `isLoopback?` util able to receive either, a device or string?

web/src/components/network/connection-form/PortsField.tsx
dgdavid: Did such a mark survive to the drop? I mean, once the form is submitted, it is ok to see the "offending" value marked somehow. I think it is how is done in other fields (IP maybe?)

web/src/components/network/connection-form/PortsField.tsx
dgdavid: Not really sure about the empty state. Neither about displaying something like "Nothing to select" because it is depends on the configuration of the field (if it accepts free input, like this, or not)

Keep it as it is by now, and let's create a follow-up to improve the situation at the dropdown level.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: Thanks for the clarification. I overlooked the comment about the line, sorry. Was a bit tired while doing the review.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: But the thing about "Used by" worries me less than the other one, which looks more unstable. As requrested, give me more time to think about it.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: Sure, keep it as it is now, but I would like to find a better approach. Since it will be just an internal change, you can move this on as it is.

web/src/components/network/connection-form/PortsField.tsx
dgdavid: Ok, same than the partOf: keep it as it is now, but let's explore if there is a better alternative in the future. I'm a bit lost with the Extract type too, but since it works any internal improvement, if any, will be transparent to the UI so we can address confidently later.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: Great, nice. 

Also, as I said to @joseivanlopez in another review, I tend to not type these returns (`: React.ReactNodw`) because most of the time TypeScript is able to infert it. And it is not always a React.ReactNode

image


https://www.totaltypescript.com/jsx-element-vs-react-reactnode

web/src/components/network/connection-form/PortsField.tsx
dgdavid: I mean, something in the line of

```ts
/** Name the kernel gives the loopback device. */
const LOOPBACK_IFACE = "lo";

/**
 * Returns true if the device is the machine's loopback interface.
 *
 * A string is treated as a device name; a `Device` may also identify the
 * loopback by its reported connection type.
 */
const isLoopback = (device: Device | string): boolean => {
  if (typeof device === "string") return device === LOOPBACK_IFACE;

  return device.type === CONNECTION_TYPE.LOOPBACK || device.name === LOOPBACK_IFACE;
};
```

web/src/components/network/connection-form/PortsField.tsx
dgdavid: As discussed in a meeting, let go tin that way by now but this reveals a bigger problem: if there are no interfaces to pick, the whole network UI _might be_ useless for those who are not able to guess what the installed machine will have. Rather complex scenario. In my view, when there are no devices, Netwok page should render a big empty state as already is done in Software.

dgdavid: I mean, an informative empty state allowing to "continue anyway". But this deserves a different PR

dgdavid: Although this is more complex that is seems and a very strange case. So, let's keeps things simple by now.

dgdavid: Also, as commented in such a meeting, this is more the field of auto-agama. So, UI should have the mentioned EmptyState to protect users and for rather complex scenario power users should use auto-agama or open the terminal and work from there.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
dgdavid: In a meeting, we have decided to show it always while selecting a port. No matter if it is fully empty.

web/src/components/network/connection-form/PortsField.tsx
dgdavid: If it works, let's continue as it is and revisit it when time permits.

web/src/components/network/connection-form/PortsField.tsx
teclator: Taken almost verbatim, thanks: `LOOPBACK_IFACE` and `isLoopback(device | string)` now live in `~/utils/network` (77cef0baf).

On the hand-typed `lo`: nothing stopped it before, it went straight into the connection. There is now a rule refusing it, next to the one refusing a controller as a port of itself:

```ts
const portError = (port: string, iface: string): TranslatedString | undefined => {
  if (iface && port === iface) return sprintf(_("%s cannot be a port of itself"), iface);
  if (isLoopback(port)) return _("The loopback device cannot be a port");

  return undefined;
};
```

The three places that asked "is this the loopback" go through the helper now: the devices the ports field offers, the ones `Form.tsx` will bind a connection to, and the rule above. `VlanFields` still excludes by type through the `DeviceSelector` filter, where no name written by hand can reach it, so I left it as it was; say the word if you would rather have one way in everywhere.

web/src/components/network/connection-form/PortsField.tsx
teclator: It did not survive the drop. It does now (8d7553c76): once the form is refused, the port the message is about is marked, the way the DNS servers field marks the address that broke it, and it stays quiet until that first submit.

Validation is still in one place. The rule and the mark both come from the same `portError` in `validations.ts`, so the two cannot drift into wording the same problem differently, and the field only asks which of its values broke it:

```tsx
validateOnSubmit={(value) => portError(value, controllerIface)}
```

Putting the two fields side by side turned up two more differences, both fixed:

- the field said which value was wrong but left the user to remove them one at a time, while the DNS one ends its error with a link taking out every marked entry at once. The multi select field now offers the same link, worded identically (773c01cb4).
- a value written by hand is drawn outlined, so its red read lighter than the filled red of the DNS chips. An entry in error is now filled whatever the caller asked for, since being wrong is the louder news (9b52fbdcf).

And since the rule and the mark are worded the same, the field no longer reads that one sentence twice.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
```diff
@@ -161,17 +157,17 @@ export default function DeviceSelectorModal({
...
-        ]
-      : []),
+    {
+      // TRANSLATORS: table column telling whether a network device has a cable
+      // plugged in and, when it does, how fast the link is.
+      name: _("Link"),
+      // The link is best-effort: a device reports it or it does not, and the
+      // ones that do not leave the cell empty. The column stays either way,
+      // since one that comes and goes with the data leaves the user wondering
+      // what they did to lose it.
```
dgdavid: NP: "The column stasys.... lose it." is already told in the dialog documentation. I would drop from here.

web/src/components/network/connection-form/PortsField.test.tsx
```diff
@@ -255,15 +255,19 @@ describe("PortsField", () => {
     within(deviceRow("enp2s0")).getByText("No link");
   });
 
-  it("leaves out the details no device reports", async () => {
+  // A column that comes and goes with the data leaves the user wondering what
+  // they did to lose it, so it stays and the cells are the empty ones.
```
dgdavid: NP: Again, clarification already stated. What is more, I would not test all those "even when device..." I would, instead, just test the dialog mounts the expected columns, enough IMHO

web/src/components/network/connection-form/Form.tsx
```diff
@@ -86,7 +88,7 @@ type FormValues = typeof defaultOptions.defaultValues;
  * machine, so binding a connection to it would produce a profile that cannot
  * reach anything.
  */
-const UNBINDABLE_DEVICES = ["lo"];
+const UNBINDABLE_DEVICES = [LOOPBACK_IFACE];
```
dgdavid: Leftover, right?

web/src/components/form/multi-select-field/MultiSelectField.tsx
dgdavid: the unique again?

web/src/components/network/connection-form/validations.ts
```diff
@@ -27,6 +27,10 @@
  * form/validation-helpers.ts. The exported validate function is wired into the
  * form's onSubmitAsync validator, following the submit-only validation
  * convention.
+ *
+ * A single rule is exported on its own, `portError`, since the ports field
+ * marks the offending port with the very same sentence the form refuses it
+ * with.
```
dgdavid: NP: Not sure if this is really needed. It is clearly exported because it is used/needed somewhere. Something easy to guess.

web/src/components/form/multi-select-field/MultiSelectField.tsx
```diff
@@ -904,7 +921,26 @@ export default function MultiSelectField({
...
+                    )}
+                  >
+                    {(text) => (
+                      <Button variant="link" isInline onClick={clearInvalid}>
+                        {text}
+                      </Button>
+                    )}
+                  </Interpolate>
+                </>
+              )}
```
dgdavid: Worth having a look if this logic, which If I'm not wrong is exactly the same than the written in ArrayField, should/can be moved to the shared primitives. `clearInvalid`, for example, looks a good candidate for src/components/forms/primitives/entry-helpers.

dgdavid (commented):
It mostly looks good. I have added some NP comments, please check them.

web/src/components/network/connection-form/DeviceSelectorModal.tsx
teclator: Dropped, the docblock keeps it. ce16080b4

web/src/components/network/connection-form/PortsField.test.tsx
teclator: Done in dadfa1f2d: both `even when…` tests are gone, replaced by one that asserts the set of column headers the dialog mounts. It no longer depends on what the mocked devices report, which was the point the two tests were making the long way round.

web/src/components/network/connection-form/validations.ts
teclator: Dropped. ce16080b4

web/src/components/network/connection-form/Form.tsx
teclator: Not quite, it was still feeding `exclude={{ devices: … }}` on both selectors. But you are right that it was a second answer to a question `isLoopback` already answers, and the weaker one: it matched by name only, while `isLoopback` also knows the type the system reports.

bbbaef8bb derives the excluded names from `isLoopback`, so the criterion lives in one place, and `LOOPBACK_IFACE` goes back to being private to `utils/network`.

web/src/components/form/multi-select-field/MultiSelectField.tsx
teclator: The inner one dedups the marks the values carry against each other; this one dedups those marks against the rules refusing the whole field, which is the case the comment above it was about. Two different jobs reading as one repeated call, though.

85293ff8e names the list (`errorMessages`) and computes it next to the others, so what it is for shows. The `sift` went with it, both lists it joins are already sifted.

web/src/components/form/multi-select-field/MultiSelectField.tsx
teclator: Good catch. 8571dfeab moves the split and the sentence to `primitives/entry-helpers`, next to `pasteAnnouncement` and for the same reason: the outcome is the same in both fields, so it is worded the same. The two were not quite identical, they even disagreed on the word, one removed `entries` and the other `values`. They now both say `%d invalid entries removed.`, which is already translated.

It stays pure (`clearInvalid(entries, errorFor) -> { kept, announcement }`), leaving each field to store what it keeps and to announce it its own way. Covered in `entry-helpers.test.ts`.

dgdavid (approved):
I haven't had time to test it manually myself, which would be ideal for a better review. I have had, however, a lot of conversations with @teclator who has developed this and has show me the feature working. Thus, approving it.

PR #3884, opened 2026-09-09
feat(rust): report more details about the network devices

teclator:
## Problem

The devices reported by the network service carry little more than a name, a MAC address and a state. That is not enough to tell two NICs apart: nothing says which one has a cable plugged in, which one is the 10G card, or where each one sits. NetworkManager knows all of it, we just do not ask for it.

- https://trello.com/c/loaOB90k/
- related to https://bugzilla.suse.com/show_bug.cgi?id=1279417

## Solution

Add `speed`, `carrier`, `driver`, `busPath` and `mtu` to the device type, all optional and all best-effort: a device that does not report a property leaves it unset instead of failing the whole build.

- `speed` comes from the `Device.Wired` D-Bus interface, which only wired devices implement. It reads `0` while the link is down, which means "unknown" and not "0 Mb/s", hence the `Option`.
- `carrier` is derived from the `InterfaceFlags` bitmask, which works for every device type, instead of the per-type `Carrier` property that NetworkManager deprecated.
- The device stream watches `InterfaceFlags`, `Mtu` and the `Device.Wired` properties too, so plugging a cable in mid-session refreshes what is reported instead of leaving a stale "no link".

Nothing consumes the new fields yet. The UI side is a separate pull request.

## Target payload

Example of `/api/system` output under `"network"` showing the new fields for an Ethernet device:

```json
{
  "network": {
    "devices": [
      {
        "name": "enp197s0f3u1u1",
        "type": "ethernet",
        "macAddress": "60:7D:09:7C:E4:39",
        "ipConfig": {
          "addresses": ["192.168.1.20/24", "fe80::1af5:39f9:d3fd:25e2/64"],
          "nameservers": ["192.168.1.1"],
          "ignoreAutoDns": false,
          "gateway4": "192.168.1.1",
          "routes4": [
            {
              "destination": "192.168.1.0/24",
              "metric": 100
            },
            {
              "destination": "0.0.0.0/0",
              "nextHop": "192.168.1.1",
              "metric": 100
            }
          ],
          "routes6": [
            {
              "destination": "fe80::/64",
              "metric": 1024
            }
          ],
          "linkLocal4": "default"
        },
        "connection": "Ethernet",
        "state": "connected",
        "speed": 1000,
        "carrier": true,
        "driver": "r8152",
        "busPath": "pci-0000:c5:00.3-usb-0:1.1:1.0",
        "mtu": 1500
      }
    ]
  }
}
```

## Testing

- Added unit tests for the two conversions: carrier out of the flags bitmask, and a speed of `0` mapping to `None`.
- Tested manually with a machine with two NICs, one of them unplugged.


abonillabeeche:
Thanks  - looks good to me. I've added a PR on top of yours to complement, tested briefly and added a video with a preview.

rust/agama-network/src/nm/streams/devices.rs
```diff
@@ -138,6 +143,11 @@ impl DeviceChangedStream {
             {
                 return Some(NmChange::DeviceUpdated(path));
             }
+            "org.freedesktop.NetworkManager.Device.Wired"
```
imobachgs, line 146: What about wireless devices?

imobachgs (commented):
It looks good. Just a question if we should implement similar behavior for wireless devices.

imobachgs (approved):
Sorry, I was looking to the wrong D-Bus spec. It looks good but, please, add a reference to the changes entry.

PR #3905, opened 2026-09-10
Improve service CI in 16.1

lslezak:
## Problem

- The CI in the SLE-16.1 fails

## Solution

- Fix the unit test so it properly works in the 16.x environment.
- Use the 16.0 container (16.1 is not published yet, will be updated later.)
- Cleanup, do not install not needed packages like `suseconnect-ruby-bindings`

## Testing

- The CI is green now

PR #3908, opened 2026-09-10
[SLE-16.0] Do not read /etc/YaST2/control.xml file from the system in test

lslezak:
Backporting the test fix from #3905 to SLE-16.0.


joseivanlopez (approved):
LGTM

PR #3910, opened 2026-09-10
feat(web): show link state and speed in the device picker

abonillabeeche:
## Problem

DeviceSelectorModal (added here) shows a device's MAC address, type, IP addresses and NetworkManager connection state, but not whether it actually has a cable plugged in or how fast the link is. That is exactly the information that is hardest to get by copy/pasting from the Details of other connections, and the original motivation for this feature.

gh#agama-project/agama#3884 adds `carrier` and `speed` to the `Device` API type (best-effort, from NetworkManager's `Device.Wired` interface and `InterfaceFlags`), but leaves them unconsumed: "the UI side is a separate pull request."

## Solution

- Add `carrier`/`speed` to the frontend `Device`/`APIDevice` types.
- Add a `deviceLinkLabel()` helper in `utils/network.ts`: "Down" when there is no carrier, "Up (%d Mb/s)" when the speed is known, "Up" otherwise, "Unknown" when the device does not report a carrier at all.
- Add a "Link" column to `DeviceSelectorModal` using it, next to the existing "State" column.

Opened against `enhance_ports_selection` rather than `master` since it only makes sense on top of this branch's `DeviceSelectorModal`, and depends on #3884 for the backend fields to actually be populated at runtime.

## Testing

- Extended `DeviceSelectorModal.test.tsx` with the "Down"/"Up"/"Up (Mb/s)"/"Unknown" cases.
- Manually verified in a browser against fixture devices covering all four link states.

### DIFF
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Sep 10 12:39:17 UTC 2026 - Alejandro Bonilla <abonilla@suse.com>
+
+- Show the link state and, if up, the speed of each device offered
+  when picking the ports of a bond or a bridge, so a card can be
+  told apart from another without checking its details elsewhere
+  (bsc#1279417).
+
+-------------------------------------------------------------------
 Mon Sep  7 15:00:00 UTC 2026 - Knut Anderssen <kanderssen@suse.com>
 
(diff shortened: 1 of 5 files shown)


abonillabeeche:
https://github.com/user-attachments/assets/2bde4a11-1e14-4c3d-b0e0-f582501a0d6d

PR #3915, opened 2026-09-10
Use license.txt for English in license.opensuse

lkocman:
## Problem

The license text for license.opensuse was empty. 

Claude really think it's the bug is that we used license.en.txt instead of license.txt for English (which is used for lookup). This should fix it.

image (4)


mchf:
Relevant piece of code is https://github.com/agama-project/agama/blob/master/rust/agama-utils/src/licenses.rs#L176

mchf (approved):
Assuming that english is default language then it really seems that code looks fore license.txt file. So LGTM

jreidinger:
It targets master aka 16.2 is it correct or it should be 16.1 branch?

lkocman:
16.1 PR is here, sorry https://github.com/agama-project/agama/pull/3923

PR #3917, opened 2026-09-11
web: fix package submission; clean up RPM .spec

mvidner:
## Problem

obs-service-node_modules has changed in an incompatible way

- https://github.com/agama-project/agama/actions/runs/34576340312/job/103189541429
- https://github.com/openSUSE/obs-service-node_modules/issues/51

## Solution

- Adapt to that
- also, mostly unrelated, clean up the spec file

## Testing

- Tested manually: `osc service manualrun node_modules; osc build` still works

## Screenshots

No

## Documentation

The changes file is enough

### DIFF
--- a/web/package/_service
+++ b/web/package/_service
@@ -17,6 +17,8 @@
   </service>
   <service name="node_modules" mode="manual">
     <param name="cpio">node_modules.obscpio</param>
+    <!-- obs-service-node_modules version 3 needs this -->
+    <param name="legacy-container">true</param>
     <param name="output">node_modules.spec.inc</param>
     <param name="source-offset">1000</param>
   </service>
--- a/web/package/agama-web-ui.changes
+++ b/web/package/agama-web-ui.changes
@@ -1,6 +1,8 @@
 -------------------------------------------------------------------
 Fri Sep 11 08:46:42 UTC 2026 - Martin Vidner <mvidner@suse.com>
 
+- web: update _service to work with obs-service-node_modules 3
+  (gh#openSUSE/obs-service-node_modules#51)
 - web: remove leftovers from spec file
 
 -------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


imobachgs (approved):

imobachgs:
@mvidner Does it apply to `SLE-16.1` branch too?

mvidner:
> Does it apply to `SLE-16.1` branch too?

tl-dr: Yes

Good point: I want it as preparation for the dependency reduction (#3863) but the compatibility actually depends on the GH action that submits git to OBS: which version of `obs-service-node_modules` are [`obs-staging-web.yml`][web] and [`obs-service-shared.yml`][shared] using. And whether these Actions are used in SLE-16.1 at all.

[web]: https://github.com/agama-project/agama/blob/SLE-16.1/.github/workflows/obs-staging-web.yml
[shared]: https://github.com/agama-project/agama/blob/SLE-16.1/.github/workflows/obs-service-shared.yml

And since the Action is enabled and it is using a Tumbleweed container, the answer is:

Yes it applies to SLE-16.1

imobachgs:
> > Does it apply to `SLE-16.1` branch too?
> 
> tl-dr: Yes
> 
> Good point: I want it as preparation for the dependency reduction (#3863) but the compatibility actually depends on the GH action that submits git to OBS: which version of `obs-service-node_modules` are [`obs-staging-web.yml`](https://github.com/agama-project/agama/blob/SLE-16.1/.github/workflows/obs-staging-web.yml) and [`obs-service-shared.yml`](https://github.com/agama-project/agama/blob/SLE-16.1/.github/workflows/obs-service-shared.yml) using. And whether these Actions are used in SLE-16.1 at all.
> 
> And since the Action is enabled and it is using a Tumbleweed container, the answer is:
> 
> Yes it applies to SLE-16.1

OK, good, just wanted to clarify this because I guess we should backport this to the `SLE-16.1` branch then.

PR #3927, opened 2026-09-15
feat(web): add a multi-select form field with search and custom entries

dgdavid:
## Problem

A form field holding a list of values had only `ArrayField`, which takes whatever the user types and offers nothing. That is right for addresses or domains, which the installer cannot know in advance. It is wrong for setups such as bond and bridge ports, where the values are devices the backend already lists: the user had to type names they could not see, spelled exactly right, with no way to find out what was available without leaving the form.

The other half of the gap is on the other side: `SearchableSelectField` offers a filtered list, but takes a single value.

## Solution

Add a searchable field that lets users select multiple values from a predefined list, with optional support for custom entries.

### Implementation details

`MultiSelectField` offers the known values in a list, filters them as the user types, and holds as many of them as the user picks. It takes values written out as well when the caller sets `allowCustomEntries`, so the form can accept values outside the predefined list without giving up the list for everything else.

The committed values live inside the control, ahead of the text box, and the text box follows the last one on its line. Beyond `entriesThreshold` the rest are summarized as "N more", which expands and collapses, so a field holding a dozen values does not push the rest of the form down the page.

Callers can add a `footerEntry`, the entry pattern added in <https://github.com/agama-project/agama/pull/3822> and already in use by `DropdownField`, to reach details the list cannot offer in a clean, semantic and valid way. It is pinned to the foot of the list, so scrolling or filtering never takes the way out of sight.

As the user types, the best match is picked out, so Enter or Tab takes it without a trip through the list. Nothing is ever typed into the box on their behalf: "eth" never silently becomes "eth0", the text stays exactly as written, and nothing is picked out while the user is deleting. `autoHighlight={false}` turns it off where guessing would get in the way.

Two things around the field are worth naming here; the commits tell the rest.

The pieces the field needs were already written, half in `ArrayField` (committed values, and the draft handling: trim, normalize, validate, split a paste, skip duplicates) and half in `SearchableSelectField` and `DropdownField` (the filter, the footer entry). They move out ahead of the field rather than being copied, so there is one of each. Two side effects of sharing them: the mark showing where the keyboard is in a list now draws the product's focus ring in both places, where before it was a hardcoded blue from a token that no longer exists in one and nothing at all in the other; and both fields describe a paste with the same sentences, the ones `ArrayField` already had, so no translation is lost.

`components/form/` gains three levels, because twenty-five flat modules said nothing about which of them a form may reach for, which belong to the fields, and which belong to one field alone, to the point that four files had taken a `multi-select-` prefix to fake the namespace they were missing. The top level stays the public surface, the fields and the helpers for writing a form; `primitives/` holds what sits below a field and is shared by two of them; `multi-select-field/` holds what belongs to this field alone. Renames only, no behavior change.

> [!WARNING]
> 
> The field is not adopted anywhere yet. Bond ports, bridge ports and any other form come in their own pull requests, so this one is the component and the reasoning behind it.

## Accessibility

The field follows the APG editable combobox with list autocomplete, in its automatic selection variant.

- One tab stop. Real DOM focus stays in the text box; the options, the committed values, the "N more" summary and the button emptying the field are all pointed at with `aria-activedescendant`. Nothing else in the control is in the tab order.
- One navigation mode at a time: the list, or the committed values, never both. The rules are written down as a contract in `multi-select-field/use-keyboard.ts`, which is where the whole keyboard lives, since that is the part easiest to break by accident.
- The field can never hold the user in: with nothing to commit, Tab always moves focus on.
- The list is a multiselectable listbox, and every option says whether it is one of the chosen values.
- What the user cannot see is said: how many options are left after filtering, a value added or removed, a value already held, text refused because it matches no option, the field emptied. Sighted users get the same information as instructions beside the field.
- The options are not checkboxes. PatternFly's menu item with a checkbox renders a real `<input>` and turns the item into a `<label>`, which costs the option its `role="option"` and its `aria-selected`, taking the listbox and the keyboard model with it. A native checkbox inside an option is invalid HTML and tells assistive technology nothing it does not already have. The reasoning is written into the form conventions so the next field facing it does not have to work it out again.
- ARIA does not allow an option to declare that it opens a dialog, so the footer entry's label ends in an ellipsis and screen reader users additionally hear "Opens a dialog". That sentence comes from the shared component, so every field says it the same way.

One deliberate departure is worth naming: a Tab that commits something keeps focus in the field, against the convention that Tab moves from one component to another. It is for the sake of agility, a field taking many values is filled a good deal faster when the key confirming a value is the one already under the finger, and it is what `ArrayField` already does, so both fields take Tab the same way. It is written down at the prop, with the APG reference, to be revisited if users find it gets in their way.

## Screenshots

> ![WARNING]
> 
> These are from a local mockup, bond ports has not been touched in this PR.

### Empty

localhost_8081_ (21)

### Empty an open

localhost_8081_ (23)

### Closed with selected values

localhost_8081_ (24)

### Open with selected values

localhost_8081_ (25)

### Closed with more than 3 values selected (threshold configurable)

localhost_8081_ (26)

## Testing

- 66 unit tests cover the field: filtering, adding and removing, custom entries, the "N more" summary, the footer entry, every row of the keyboard contract, and every announcement.
- Unit tests travel with the pieces that moved: the paste helpers with the function that covers them, `ArrayField` unchanged otherwise.
- `npm run test`, `npm run eslint` and `npm run check-types` pass.
- Tested by hand in the browser, keyboard only, against a throwaway form driving the real field.

Worth a manual pass with Orca and NVDA: whether the filter outcome is announced too often when typing fast, and whether the footer entry's hint reads well after an ellipsis label, and how a paste reads in both fields now that they share the sentences.

## Follow-up

- Adopting the field in the bond ports and bridge ports forms.
- `placeholder` is offered because `SearchableSelectField` has one, but this field has helper text and instructions and does not need it. Worth dropping if nothing asks for it.
- The Tab default above, once there is feedback from real use.

## Documentation

The form conventions gain a section on fields holding several values: which of the two to reach for, what they share, and why the options are not checkboxes.

They also gain one on the field library itself, which the conventions had never described: the three levels of `components/form/` and who imports what, when a module has earned a place in `primitives/`, when a field has earned a directory, and how the modules inside one are named. The chapter on code organization had said it was about forms; it now covers the fields they are built from as well.

---


imobachgs (approved):
It looks good. Great work :wink:

PR #3935, opened 2026-09-15
Fix storage install

### LINKED ISSUE #3912: Error while formatting partitions produces hard-to-read error message
How to reproduce:

- Navigate to the new volume / partition dialog
- Add additional FS-format options (some wrong / misslleped ones) ;)
- Start installation

Image

You get similar error message like this

Image

I'd love to see something a bit easier to read

### LINKED ISSUE #3916: Installation still continues after failed Storage actions
Even if formatting file system for "/" failed and I decide not to continue with Storage actions, the installer jumps on the next step which is installation of packages. And as there is no disk to install to, it uses /mnt which is the inst-sys memory. Then the system crashes.

joseivanlopez:
## Problem

Once the installation starts, the UI shows a question when there is a storage commit error and the user has to decide whether to continue with the installation or not. But selecting "no" only cancels the remaing storage actions and continues with the installation. 

Moreover, the details of the commit error are not properly formatted, making them difficult to read.

## Solution

Generate a D-Bus error when the user decides to cancel on a commit error. This allows to properly cancel the task for installing storage and all the dependant tasks. As result, the installer goes directly to the "Installation failed" view when the user decides to not continue with the installation.

A new specialized question component is added for handling storage commit errors in the web UI.

<details>
<summary>Show/Hide</summary>

commit-error

commit-error-details

</details>


dgdavid:
> commit-error-details

What a difference with the current status you showed me this afternoon. I guess `<pre>` did the trick, but in any case: cool! 🆒

web/src/components/questions/ExpandableTechnicalSection.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    screen.getByText(/Error: Failed to commit/);
+    screen.getByText(/Caused by:/);
+    screen.getByText(/Storage device not found/);
+    screen.getByText(/Invalid configuration/);
+  });
+
+  it("shows expand/collapse toggle buttons", () => {
+    plainRender(<ExpandableTechnicalSection text="Error details" />);
+
+    screen.getByText("Show technical details");
```
dgdavid, line 52: NP: if I remember correctly, the code we saw had the `(English only)` clarification. If so, which is the reason to drop it?

web/src/components/questions/ExpandableTechnicalSection.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+  it("renders nothing when text is undefined", () => {
+    const { container } = plainRender(<ExpandableTechnicalSection text={undefined} />);
+
+    expect(container.firstChild).toBeNull();
+  });
+
+  it("renders nothing when text is empty", () => {
+    const { container } = plainRender(<ExpandableTechnicalSection text="" />);
+
+    expect(container.firstChild).toBeNull();
```
dgdavid, line 73: Not sure right now (doing a review from mobile phone) but I would say we usually use the expect(container).toBeEmpty or similar

web/src/components/questions/StorageCommitErrorQuestion.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * @param question - the question to be answered
+ * @param answerCallback - the callback to be triggered on answer
+ */
+export default function StorageCommitErrorQuestion({
+  question,
+  answerCallback,
+}: {
+  question: Question;
+  answerCallback: AnswerCallback;
+}): React.ReactNode {
```
dgdavid, line 43: NP: I'm not a fan to use the return type here, but let typescript to infer it. Why? Because it is good doing so and, most important, is it a React.Node or a JSX.Element?

What I mean is that inference works better in components return type.

web/src/components/questions/ExpandableTechnicalSection.test.tsx
dgdavid: Nevermind, We probably saw an outdated version of the file since we were looking it directly in the PR. So, looks like finally we drop such a suffix.

dgdavid (approved):
Web part LGTM 👍

imobachgs (approved):

web/src/components/questions/StorageCommitErrorQuestion.tsx
joseivanlopez: Ah, I didn't know that TS is able to infer the return type of a function. Cool. I will consider it from now on. Thanks!

dgdavid (approved):
LGTM

PR #3950, opened 2026-09-17
refactor(rust): replace #[async_trait] with native async-fn-in-traits

imobachgs:
## Problem

`agama-utils`'s `Actor`/`MessageHandler` traits used `#[async_trait]` (boxed futures) on 171 implementation sites across 18 crates, even though neither trait is ever used as a `dyn` trait object anywhere in the workspace (verified via a workspace-wide grep) - the boxing was pure overhead.

## Solution

Switched `Actor`/`MessageHandler` to `trait_variant::make(Send)`, which lets implementors keep writing plain `async fn` while requiring `Send` futures (needed since actors run inside `tokio::spawn`), without boxing. Removed `#[async_trait]` from all 171 call sites and the dependency from the 7 crates that no longer needed it.

Traits that are genuinely used as `dyn` objects (`Adapter`, `Watcher`, the `*Client` traits, `ModelAdapter`) were left untouched.

## Results

Clean `cargo build --workspace --all-targets -j 8`, sccache disabled, mold linker on, 3 runs each vs. a worktree of the base commit:

| Run | master | this branch |
|---|---|---|
| 1 | 224.9s | 147.7s |
| 2 | 238.4s | 149.3s |
| 3 | 234.4s | 178.9s |
| **Average** | **232.6s** | **158.6s** |

**~32% faster clean build.**

## Links

- Check [Announcing `async fn` and return-position `impl Trait` in traits](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/) for further information.


joseivanlopez (approved):
LGTM

PR #3951, opened 2026-09-17
fix(web): avoid package name being split in two parts

### LINKED ISSUE #3948: bug#1279489 - agama-24: name of package being installed is split on real screens
https://bugzilla.suse.com/show_bug.cgi?id=1279489

With certain resolutions there are problems displaying the package names in the progress screen for installation. I think the culprit CSS was already identified but I'm not sure if we merged the fix. The Trello card and the bugzilla entry contain no details about that.

dgdavid:
While software is being installed, the screen names the package going in right now. In some resolutions that name was shown torn in two, its last characters far from the rest with a wide gap between them.

image

The name now stays in one piece, and the line takes the room the screen gives it instead of the room the name asks for, so nothing moves as one package follows another. Names too long for the line are shortened in the middle, as expected.

Fix #3948 and https://bugzilla.suse.com/show_bug.cgi?id=1279489 (protected link)


dgdavid:
Closing in favor of #3952, which needs to land first.

Once that’s merged, the next step is to merge `SLE-16.1` into `master`. 

> [!WARNING] 
> Be aware that this won’t be a clean merge: the files touched by `SLE-16.1` are no longer the same ones, as `master` has since reorganized these internals.
> 
> That reorganization is what will make it easier to bring further improvements and features to users, so the conflict is worthwhile. It does mean, however, that the progress screen changes will need to be re-applied manually on top of the new layout components.

PR #3953, opened 2026-09-17
products: support multiple licenses per product

imobachgs:
## Problem

- A product could only reference a single license (`ProductSpec.license`). Some products need to require accepting more than one license (e.g., an EULA plus a separate open source license).
- Separately, `/licenses/{id}` accepted a client-supplied `?lang=` query param to pick the language, which is inconsistent with the rest of Agama, where the system language is controlled by the `ConfigureL10n` action instead of per-request client input.

## Solution

- `ProductTemplate`/`ProductSpec`/`Product` gain a `licenses: Vec<String>` field. The old `license: Option<String>` is kept but marked deprecated.
- Production `products.d/*.yaml` (SLES, SLES4SAP, Leap) migrated from `license` to `licenses`.
- Frontend: `Product.licenses?: string[]` added (`license` deprecated); `LicenseDialog`/`ProductSelectionPage`/`ProductOption`/`Form` updated to read `licenses` (first entry for now — full multi-license acceptance UX is future work).

Implementing this required first simplifying how the manager serves license content, since it used to be keyed by a client-supplied. As part of that:

- `licenses::Registry` now reads/caches content for a single language (the system language, refreshed on `ConfigureL10n`) instead of resolving language per-request.
- `/licenses/{id}` no longer accepts `?lang=`; it always returns the license in the current system language, consistent with the rest of Agama.
- License `name` is now extracted from the text (first paragraph) alongside `body`.
- `SystemInfo.licenses` exposes the list of available licenses (id + name) via `GET /api/system`.

## Follow-up

- UI support for accepting more than one license per product.
- Removing the deprecated `license` attribute (`ProductTemplate`/`ProductSpec`/`Product.license`, frontend `Product.license`).


rust/agama-utils/src/api/manager/system_info.rs
```diff
@@ -45,8 +48,13 @@ pub struct Product {
     pub icon: String,
     /// Registration requirement
     pub registration: bool,
-    /// License ID
+    /// License ID.
+    ///
+    /// Deprecated: use `licenses` instead.
```
joseivanlopez: Do we really need to keep it as deprecated? We control the definition of the products.

rust/agama-utils/src/licenses.rs
```diff
@@ -182,40 +198,44 @@ impl Registry {
         };
 
         let license_path = self.path.join(id).join(file_name);
-        let body = std::fs::read_to_string(license_path)?;
+        let text = std::fs::read_to_string(license_path)?;
+        let (name, body) = Self::parse_content(&text);
         Ok(LicenseContent {
             id: id.to_string(),
+            name,
```
joseivanlopez, line 205: NP: maybe the name should be an Option.

imobachgs: I do not think so, you should always have a name. It is true that if the license file does not follow the format we expect, the name could be rather long. But we can improve that later.

rust/agama-utils/src/api/manager/system_info.rs
imobachgs: No, we do not need it. I just wanted to be conservative because we are already pretty late. I will get rid of it because it is exposed through SystemInfo.

joseivanlopez (approved):
LGTM

PR #3955, opened 2026-09-18
Set default bootloader for MicroOS to systemd-boot

foursixnine:
Small update to set the correct bootloader: https://bugzilla.suse.com/show_bug.cgi?id=1277929


ancorgs (approved):
LGTM

PR #3957, opened 2026-09-18
Fix blocked storage

### LINKED ISSUE #3939: Installation gets blocked after refreshing UI if storage asks a question
The installation can get blocked during the installation phase.

If commiting the storage changes fails, the installer shows a question and the user can decide whether to continue or abort. If the UI is refreshed (F5) at this point (without answering the question), the the installation process is blocked and cannot be recovered.

This is related to https://bugzilla.suse.com/show_bug.cgi?id=1270283.

### Review

* https://github.com/agama-project/agama/pull/3957
* Follow-up for 16.2: https://github.com/agama-project/agama/issues/3961

joseivanlopez:
## Problem

The installation can get blocked while rescanning devices or while committing the changes.  In both cases, there is a deadlock if a question is raised and any data is requested before answering the question (for example, refreshing the UI with F5 or running a CLI command like `agama monitor`).

This happens because some storage actions like `ActivateStorage` and `InstallStorage` are not executed in a separate Tokio task, blocking the main service task.

## Solution

A temporary and low-risky solution was provided now for SLE 16.1 GMC:

1) Avoid blocking while performing storage activation by executing the activation call in a separate Tokio task, similar to probe.
2) Avoid blocking while installing by rejecting any request to storage if the service is currently installing. 

A more generic solution will be implemented for 16.2.


dgdavid (approved):
Thanks Iván. Web part LGTM 👍

imobachgs (approved):

PR #3958, opened 2026-09-18
Adapt the storage section of the CC template

ancorgs:
## Problem

So far we have used a storage section in the JSON profile with no clear origin and rationale.

## Solution

Replace the storage section with one that looks closer to the specifications verified by the auditors.

- No separate `/boot`
- Encrypted LVM with separate `/`, `/var`, `/var/log/audit`, `/home` and `swap`

Regarding sizes, this follows an approach based on  https://gitlab.suse.de/security/common-criteria-sle15sp2/-/blob/master/autoyast/autoyast-cc_sles12_x86_64_fde.xml (SUSE internal link), although it does not look as a very smart usage of the space.

See https://github.com/agama-project/agama/pull/3959 for an alternative.

## Testing

- Profile generation tested manually


live/live-root/usr/share/cc-setup/agama-template.json
```diff
@@ -48,51 +48,50 @@
...
-            "size": {
-              "min": "10 GiB"
+              "path": "/var/log/audit",
+              "type": "btrfs",
+              "mountOptions": ["acl", "nodev", "noexec", "nosuid"]
             }
+          },
+          {
+            "name": "home",
+            "size": { "min": "512 MiB", "max": "2 GiB" },
```
jreidinger, line 88: is it good idea to have maxed root and limited home just to 2GiB? that looks quite small.

ancorgs: It's inspired by this profile (SUSE internal link) https://gitlab.suse.de/security/common-criteria-sle15sp2/-/blob/master/autoyast/autoyast-cc_sles12_x86_64_fde.xml

There the size of `/home` is 1GiB (like that, statically) and only `/` grows.

Of course, we can do better, I just wanted to not deviate much from that original profile.

I personally would better limit `/` (to something like 100 GiB) and would allow `/var` and `/home` to grow.

ancorgs: If I would have to device myself about sizes of those logical volume I would propose something like https://github.com/agama-project/agama/pull/3959

lslezak:
Closing, I have merged the alternative version in #3959.

PR #3959, opened 2026-09-18
Adapt the storage section of the CC template - version 2

ancorgs:
## Problem

So far we have used a storage section in the JSON profile with no clear origin and rationale.

## Solution

Replace the storage section with one that looks closer to the specifications verified by the auditors.

- No separate `/boot`
- Encrypted LVM with separate `/`, `/var`, `/var/log/audit`, `/home` and `swap`

This is an alternative to https://github.com/agama-project/agama/pull/3958 which takes a different approach regarding the size of the mentioned logical volumes. It creates an LVM volume group that is as big as possible (using the whole disk), but it is more conservative with the size of the logical volumes. So all the extra LVM space is left unassigned so it can distributed after installation.


lslezak (approved):
I like this approach more. Let's merge it, we can improve it later if needed.

I have tested it and it works well.

ancorgs:
Apparently my previous `push -f` didn't do what I expected (my fault, for using debatable techniques like `amend`) and this was merged with slightly different sizes. I will make a pull request to fix it (basically making numbers even lower since the idea is to leave quite some space for the administrator to be distributed later).

PR #3960, opened 2026-09-18
feat(web): allow asking for multiple licenses

dgdavid:
Since #3953, a product can require accepting more than one license. The product selection form now shows one checkbox per license, each naming the license it accepts and linking to its text. The selection can be confirmed only once all of them are accepted, and choosing another product clears them.

Products with a single license look and read exactly as before.

> [!NOTE]
> Screenshots were taken with mocked licenses: Leap 16.1 was made to require two of them for testing

localhost_8081_ (40)
localhost_8081_ (41)
localhost_8081_ (32)
localhost_8081_ (33)
localhost_8081_ (42)

## Details

- **Wording under the translation freeze.** With several licenses, each checkbox reuses the existing text, "I have read and accept the [license] for %s", with the license name where the product name was. No new strings are needed. The cost is that every link reads "license", although each checkbox is still named after its own license, for screen readers too.
- **License names.** They come from the backend, which takes the first paragraph of each license text, so they read as the license authors wrote them (e.g. "LICENSE AGREEMENT openSUSE® Leap 16.1"). A license without a name falls back to the product name, which gives exactly today's text.
- **License dialog.** It is now titled with the license name instead of the product name.
- **Current selection.** A product with several licenses offers one full-width button per license, labelled with its name and wrapping when the name is long. A single license keeps "View license".
- **Language.** Licenses are requested without a language, as #3953 expects: the backend answers in the system language, which already follows the interface one. The notice for a license not available in the interface language stays.

## One checkbox for all the licenses

Another design was proposed along the way, and from the UX side it is still the recommended one: a single, properly worded checkbox covering all licenses, each named as a link to its own tab in one shared dialog. Users would find every license in one place and accept them in one clear step, instead of working through a list of almost-identical checkboxes, and the form would stay just as easy to use and understand with five licenses as with one.

The decision to go with one checkbox per license was taken independently of the translation freeze. The freeze only adds a hurdle to the alternative in 16.1, as it needs new strings for the checkbox and the dialog. With an exception to the freeze it could still be implemented now, or it can be reconsidered later, if it is reconsidered to bring users that smoother experience.

## Tests

Unit tests cover one checkbox per license, confirming only with all of them accepted, clearing them when switching products, licenses without a name, and the current selection card.

I also checked it manually on an ISO, after making Leap 16.1 list two licenses in `/usr/share/agama/products.d` (adding `license.final`) and restarting `agama-web-server`:

- A product with two licenses shows two checkboxes, each naming its license, and cannot be confirmed until both are accepted.
- Each link opens its own license, titled with its name.
- Choosing another product and coming back clears both checkboxes.
- A product with a single license looks as before.
- A product without licenses shows no checkbox and no "License acceptance required" label.

## Follow-ups

Once the translation freeze is over, in master:

- A proper text for products with several licenses, so each link reads as the license it opens.
- A label grouping the checkboxes.
- The current selection card listing the licenses under an "Accepted licenses" heading.

The product selection form in `master` has been reworked since, so bringing this forward means adapting it there rather than a plain merge.


imobachgs (approved):

web/package/agama-web-ui.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Sep 18 14:30:33 UTC 2026 - David Diaz <dgonzalez@suse.com>
+
+- Ask to accept each license of a product requiring several
+  (gh#agama-project/agama#3960).
```
imobachgs: Please, add jsc#PED-16988 as a reference.

imobachgs: Nevermind, I did it.

PR #3965, opened 2026-09-21
Adjust Common Criteria sizes to leave more unused space

ancorgs:
## Problem

#3959 modified the storage section of the profile to create an LVM setup in which not all the space was necessarily consumed during installation (so the user can later distribute any extra space as needed).

But while adjusting those initial sizes and updating the pull request I must have messed too much with `amend` and `push -f` and the final sizes that were merged were a bit too much.

## Solution

This pull request adjust the sizes to the values I really wanted to include. Sorry for the noisze.


lslezak (approved):
LGTM, thanks!

PR #3966, opened 2026-09-21
service: Handle libstorage-ng errors when checking for resizing

ancorgs:
## Problem

Agama relies on libstorage-ng to check whether a given storage device can be resized.

If the filesystem is corrupt, libstorage-ng can throw an exception during that check.

Most of those exceptions are handled by the libstorage->Agama callbacks mechanism and transformed into an Agama question.

But, as reported at [bsc#1272827](https://bugzilla.suse.com/show_bug.cgi?id=1272827) sometimes that can produce a crash on the Agama storage process. Making it very hard to continue with the installation.

The error is in fact a bit more severe than the bug report suggests. It can affect any installation with a corrupted filesystem, not only those cases in which remote disks (like zFCP) are involved.

## Solution

Handle the corresponding exception, so corrupted devices are never candidate to be resized.

With that, the installation can proceed since the storage service does not longer crashes.

## Testing

- Added new unit tests probing the device is reported as not resizable with a pre-existing default generic explanation ("_Shrinking is not supported by this device_").
- Manually tested simulating a corrupted filesystem :+1:


joseivanlopez (approved):
LGTM

